PR #1473, opened 2019-10-24
Unify to prefix messages for signing

### LINKED ISSUE #1470: Users should be able to sign with the same format
### Expected Behavior

We should only use prefixed signatures like https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign

### Current Behavior

We currently use both prefixed and non-prefixed signatures

nambrot:
### Description

Up till this point, we had two different ways of signing messages:
1. The official Ethereum way by prefixing a message with a special ethereum prefix and hashing that (https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign). We built tooling around that with `Signatures.sol` and use it to for example authorize signing keys.
2. The more "low-level" way of signing the message directly, mostly used for `Attestations`.

This PR moves the latter to the former so that we have a predictable way of signing messages. This will especially come in handy if we decide in the future to sign attestations with a key stored in a node (vs. explicitly specifying the private key right now).

### Tested

- Unit tests

### Related issues

- Fixes #1470 

### Backwards compatibility

- This is not backwards compatible for Attestations, so should ideally go in with all the other breaking changes


packages/contractkit/src/test-utils/ganache.setup.ts
```diff
@@ -3,7 +3,18 @@ import * as ganache from '@celo/ganache-cli'
 import * as path from 'path'
 
 const MNEMONIC = 'concert load couple harbor equip island argue ramp clarify fence smart topic'
-
+export const ACCOUNT_PRIVATE_KEYS = [
```
timmoreton, line 6: what are these for?

nambrot: They are the corresponding private keys for the accounts in ganache. They are the same from the protocol tests. Sometimes we need access to the private keys to do some of these actions, like comparing if signatures are equivalent. Though this would also meanz that we can now get ridd of the private keys in the protocol tests since we can just "natively" sign

jmrossy (approved):
The /utils package changes LGTM. Don't have enough context to review the rest

packages/utils/src/signatureUtils.ts
```diff
@@ -1,6 +1,44 @@
...
+  const prefix = '\x19Ethereum Signed Message:\n' + messageLength(message)
+  return Web3Utils.soliditySha3(prefix, message)
+}
+
+export function hashMessage(message: string) {
+  return Web3Utils.soliditySha3({ type: 'string', value: message })
+}
+
+// Uses a native function to sign (as signFn), most commonly `web.eth.sign`
+export async function signMessageNatively(
```
asaj: Maybe `signMessageNoPrefix`?

packages/utils/src/signatureUtils.ts
```diff
@@ -28,10 +66,18 @@ export function serializeSignature(signature: Signature) {
...
-  return serializedV + serializedR + serializedS
+  return '0x' + serializedV + serializedR + serializedS
+}
+
+export function parseSignature(message: string, signature: string, signer: string) {
+  return parseSignatureForMessageHash(hashMessageWithPrefix(message), signature, signer)
 }
 
-export function parseSignature(messageHash: string, signature: string, signer: string) {
+export function parseSignatureForMessageHash(
```
asaj: `parseSignatureNoPrefix`?

packages/contractkit/src/utils/signing.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  signMessage,
+  signMessageNatively,
+} from '@celo/utils/lib/signatureUtils'
+import { testWithGanache } from '../test-utils/ganache-test'
+import { ACCOUNT_PRIVATE_KEYS } from '../test-utils/ganache.setup'
+
+// This only really tests signatureUtils in @celo/utils, but is tested here
+// to avoid the web3/ganache setup in @celo/utils
+testWithGanache('Signing', (web3) => {
+  it('signs a message the same way natively and with an explicit private key', async () => {
```
asaj: `signs a message the same way via RPC and with an explicit private key`

packages/contractkit/src/utils/signing.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    const nativeSignature = await signMessageNatively(message, account, web3.eth.sign)
+    parseSignature(message, serializeSignature(nativeSignature), account)
+
+    const simulatedSignature = await signMessage(message, ACCOUNT_PRIVATE_KEYS[0], account)
+    parseSignature(message, serializeSignature(simulatedSignature), account)
+
+    expect(nativeSignature).toEqual(simulatedSignature)
+  })
+
+  it('signs a message that was hashed the same way natively and with an explicit private key', async () => {
```
asaj: I don't understand what this is testing that's different from the previous test?

packages/utils/src/signatureUtils.ts
nambrot: The native signing does prefix the message?

packages/contractkit/src/utils/signing.test.ts
nambrot: Added a clarifying comment. Basically the difference is that if message is a hex-string, an implementation might not properly calculate the length of the message in the prefix

```
// This test checks that the prefixing in `signMessage` appropriately considers hex strings
// as bytes the same way the native RPC signing would
```

packages/utils/src/signatureUtils.ts
nambrot: I did change `signMessageForMessageHash` to `signMessageNoPrefix`

packages/utils/src/signatureUtils.ts
```diff
@@ -1,6 +1,55 @@
...
+  return {
+    sign: async (message: string) =>
+      Promise.resolve(
+        serializeSignature(signMessage(message, privateKey, privateKeyToAddress(privateKey)))
+      ),
+  }
+}
+
+export function signMessage(message: string, privateKey: string, address: string) {
+  return signMessageNoPrefix(hashMessageWithPrefix(message), privateKey, address)
```
asaj: Hmm maybe my previous suggestion was misguided because this line is really confusing:
```signMessageNoPrefix(hashMessageWithPrefix```

File: packages/utils/src/signatureUtils.ts [hunk 7]
```diff
@@ -28,10 +77,14 @@ export function serializeSignature(signature: Signature) {
   const serializedV = signature.v.toString(16)
   const serializedR = signature.r.slice(2)
   const serializedS = signature.s.slice(2)
-  return serializedV + serializedR + serializedS
+  return '0x' + serializedV + serializedR + serializedS
+}
+
+export function parseSignature(message: string, signature: string, signer: string) {
+  return parseSignatureNoPrefix(hashMessageWithPrefix(message), signature, signer)
```
asaj: Ditto

asaj (approved):
LGTM pending naming fix

PR #1474, opened 2019-10-24
Re-work locked gold requirements for validators and groups

asaj:
### Description

This PR re-works locked gold requirements to validators and groups. Specifically, validators and groups now must adhere to these requirements until de-registration, after which the requirements are lifted. Validators and groups are prevented from deregistering if they were in a group/had members in their group recently.

### Tested

Unit tests
### Other changes

- Prohibit unlocking of gold while voting in governance
- Move proof-of-possession precompile to common `UsingPrecompiles` contract


packages/protocol/contracts/governance/Validators.sol
```diff
@@ -451,8 +371,8 @@ contract Validators is
    */
   function _updateValidatorScore(address validator, uint256 uptime) internal {
     address account = getLockedGold().getAccountFromValidator(validator);
-    require(isValidator(account), "isvalidator");
-    require(uptime <= FixidityLib.fixed1().unwrap(), "uptime");
+    require(isValidator(account));
+    require(uptime <= FixidityLib.fixed1().unwrap());
```
nambrot, line 370: I'm not sure if I asked this before, but do we not provide these strings? In my editor it is actually a lint violation

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -71,13 +68,17 @@ contract Validators is
     address group;
   }
 
-  // Stores the membership history of a validator.
+  // Stores a the per-epoch membership history of a validator, used to determine which group
```
kevjue: nit:  typo

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -524,21 +434,29 @@ contract Validators is
...
   function deregisterValidator(uint256 index) external nonReentrant returns (bool) {
     address account = getLockedGold().getAccountFromActiveValidator(msg.sender);
     require(isValidator(account));
+
+    // Require that the validator has not been a member of a validator group for
+    // `validatorLockedGoldRequirements.duration` seconds.
     Validator storage validator = validators[account];
     if (validator.affiliation != address(0)) {
-      _deaffiliate(validator, account);
+      require(!groups[validator.affiliation].members.contains(account));
```
kevjue, line 445: Would this assertion always fail, or will the validator always be removed from the group's affiliation list before `deregisterValidator` is called?

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -596,12 +516,12 @@ contract Validators is
     require(commission <= FixidityLib.fixed1().unwrap(), "Commission can't be greater than 100%");
     address account = getLockedGold().getAccountFromActiveValidator(msg.sender);
     require(!isValidator(account) && !isValidatorGroup(account));
-    require(meetsValidatorGroupBalanceRequirements(account));
-
+    uint256 lockedGoldBalance = getLockedGold().getAccountTotalLockedGold(account);
```
kevjue, line 511: Should these two lines be changed to use `meetsAccountLockedGoldRequirements` instead?

kevjue (approved):
LGTM.  Just a few nits.

packages/protocol/contracts/governance/Validators.sol
asaj: The validator needs to be removed from the group's membership list before `deregisterValidator` is called. Furthermore, it needs to have been removed from this list more than `validatorLockedGoldRequirements.duration` seconds ago.

The reason we need to check to see if it's currently a member of the group with which it's affiliated is because the `lastRemovedFromGroupTimestamp` isn't relevant information if the validator is still a member of a group.

packages/protocol/contracts/governance/Validators.sol
asaj: No, only because the group hasn't been registered yet, so the requirement is `0`. If we did this check after registration (i.e. moved this line to the end of the function) then we could use `meetsAccountLockedGoldRequirements`

PR #1480, opened 2019-10-25
[Wallet] Remove QR debouncing to improve responsiveness

jmrossy:
### Description

Remove QR debouncing to improve responsiveness. It seems we no longer need it, possible due to the QR lib upgrade.

### Tested

Scanned QR codes a bunch of times

### Backwards compatibility

Yes


jeanregisser (approved):
Ship it! 👍

PR #1486, opened 2019-10-25
[Wallet] Add fiat exchange rate polling

annakaz:
### Description

Poll fiat exchange rate from Notification Service and store daily exchange rates. This will be used in the wallet/blockchain api to display transactions in their local currency at the time of the transaction, rather than their present rate. 

### Tested

notification-service:
Ran locally with a short fiat_exchange_polling_interval to confirm that fiat exchange rates are being fetched. Ran jest tests

blockchain-api:
Ran locally, and ran and updated tests with types

### Other changes

Added `LocalCurrencyCode` to @celo/utils to be used across wallet, notification-service and blockchain-api to ensure supported currencies are available across all required packages
Note that that it is copied across packages (imports commented out) as we must first republish @celo/utils to npm

### Related issues

- Notification service side of #834, blockchain-api changes to build on top of this

### Backwards compatibility

Yes


annakaz:
Closing in favor of directly using a different exchange rate API rather than saving data in firebase

PR #1487, opened 2019-10-25
Remove consistently failing integration sync test

asaj:
### Description

Removes an end-to-end test that we no longer rely on. This test fails consistently which habituates us to failing master builds, which can hide underlying real issues.

### Tested

These are tests

### Other changes

None


nambrot (approved):
Ironically, the test is still failing?

nategraf:
So I actually know why this test is failing (the `alfrajoresstaging` environment was deployed with a different epoch size than exists in the `.env.alfajorestaging` file at master) at the moment and am working on a fix. If I do fix it, would it be worth reinstating or has it been historically unreliable?

nategraf:
I've created https://github.com/celo-org/celo-monorepo/pull/1504 to address the standing issue with this test if we want to keep it.

nategraf:
We decided to go with #1504 to fix the test rather than nixing it.

PR #1490, opened 2019-10-25
Adding a parameter to limit variation to the median price

### LINKED ISSUE #529: Reports SNBAT move the oracle price too quickly
### Expected Behavior

Rate limiting on the median, e.g. no more than 5% a day.

### Current Behavior

No rate limiting

mrsmkl:
### Description

Added a parameter to limit variation to the reported prices. There is now a new variable where the current "median" is stored, and it is changed each time a new report is made.

Alternative approaches I can think of:
* setting a window for possible median values each day / time period
* preventing oracles from changing their individual values too much
* change the value linearly when reading

### Tested

Added tests.

### Related issues

- Fixes #529


packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -24,8 +28,11 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
   mapping(address => SortedLinkedListWithMedian.List) private timestamps;
   mapping(address => mapping(address => bool)) public isOracle;
   mapping(address => address[]) public oracles;
+  mapping(address => uint) private limitedMedianRate;
```
asaj: `uint256`

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -24,8 +28,11 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
...
+  mapping(address => uint) private limitedMedianRateTimestamp;
```
asaj: ditto

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -79,6 +90,34 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
     emit ReportExpirySet(_reportExpirySeconds);
   }
 
+  /**
+   * @notice Sets the maximum change rate for median.
+   * @param rate Maximum median chaneg rate as unwrapped Fraction.
```
asaj: "change"

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -79,6 +90,34 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
...
+  function setMaxMedianChangeRatePerDay(uint256 rate) public onlyOwner returns (bool) {
+    return _setMaxMedianChangeRatePerDay(rate);
+  }
+
+  /**
+   * @notice Sets the maximum change rate for median.
+   * @param rate Maximum median chaneg rate as unwrapped Fraction.
+   * @return True upon success.
+   */
+  function _setMaxMedianChangeRatePerDay(uint256 rate) private returns (bool) {
```
asaj: Why do we need this private function? Can we fold it into the public function?

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -179,9 +217,33 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
...
     emit OracleReported(token, msg.sender, now, value, DENOMINATOR);
+    recomputeRate(token);
+  }
+
+  function recomputeRate(address token) internal {
+    uint256 originalMedian = limitedMedianRate[token];
     uint256 newMedian = rates[token].getMedianValue();
+    if (limitedMedianRateTimestamp[token] != 0 && maxMedianChangeRatePerDay.unwrap() != 0) {
+      uint256 td = now.sub(limitedMedianRateTimestamp[token]);
+      if (td > 1 days) td = 1 days;
```
asaj: Why cap `td` at one day?

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -179,9 +217,33 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
...
+      FixidityLib.Fraction memory timeFraction = FixidityLib.newFixedFraction(td, 1 days);
+      FixidityLib.Fraction memory maxChangeRate = timeFraction.multiply(maxMedianChangeRatePerDay);
```
asaj: Is this exactly what we want? I would've thought `maxMedianChangeRatePerDay` would compound

e.g. if `maxMedianChangeRatePerDay` was 10%, and the median rate was 1 at t=0, at t=5 it could be set to 1 * 1.1^5 or 1 / 1.1^5

packages/protocol/contracts/stability/SortedOracles.sol
mrsmkl: I think it's good idea to have some upper limit for the change. But I don't know how often they are supposed to be updated.

packages/protocol/contracts/stability/SortedOracles.sol
mrsmkl: Yeah, it won't compound now, so if you update more frequently, you get more change.
Perhaps we have to implement exponent function?

mrsmkl: Looks like there is already a precompile for that

mrsmkl: I think there is a precompile that could be useful, going to test that.

packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -54,10 +59,45 @@ contract UsingPrecompiles {
...
+            case 0 {
+              error := 1
+            }
+            default {
+              returnNumerator := mload(returnDataPosition)
+              returnDenominator := mload(add(returnDataPosition, 32))
+            }
         }
     }
+    if (error == 1) {
```
asaj: Why is this necessary?

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -160,9 +189,55 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
       address(0)
     );
     emit OracleReported(token, msg.sender, now, value, DENOMINATOR);
+    recomputeRate(token);
+  }
+
+  function fracMulExp(FixidityLib.Fraction memory a, FixidityLib.Fraction memory b, uint256 exp)
```
asaj: Perhaps include in the fixidity lib?

packages/protocol/contracts/stability/SortedOracles.sol
asaj: I'm not totally sure I agree this is necessary, It seems to me that a limit on the rate of change per unit time is sufficient

packages/protocol/test/stability/sortedoracles.ts
```diff
@@ -42,6 +43,62 @@ contract('SortedOracles', (accounts: string[]) => {
     })
   })
 
+  describe('#setMaxMedianChangeRatePerDay', () => {
```
asaj: Update to `PerSecond`

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -160,9 +189,55 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
...
+  function recomputeRate(address token) internal {
+    uint256 originalMedian = limitedMedianRate[token];
     uint256 newMedian = rates[token].getMedianValue();
+    if (limitedMedianRateTimestamp[token] != 0 && maxMedianChangeRatePerSecond.unwrap() != 0) {
+      uint256 td = now.sub(limitedMedianRateTimestamp[token]);
+      if (td > 1 days) td = 1 days;
+      uint256 maxChange = fracMulExp(
+        FixidityLib.fixed1(),
+        maxMedianChangeRatePerSecond.add(FixidityLib.fixed1()),
+        td
```
asaj, line 234: This doesn't feel safe to do as the exponents could grow very large...

packages/protocol/contracts/common/UsingPrecompiles.sol
mrsmkl: oops, got some error and didn't realize the precompile was implemented in ganache

packages/protocol/contracts/stability/SortedOracles.sol
mrsmkl: So need to add something to make sure it won't overflow?

packages/protocol/contracts/stability/SortedOracles.sol
mrsmkl: fixidity lib cannot inherit UsingPrecompiles...

packages/protocol/contracts/stability/SortedOracles.sol
mrsmkl: Added some checks, now it shouldn't be able to overflow.

mcortesi (commented):
It seems to me, that the *intent* of the PR is to limit the variation of the median price, but in particular for short time periods.

For "long" time periods, it probably doesn't matter, nor we can anticipate those values. So, why limit it?

And if you check this for short time periods, we eliminate the whole overflow/checks issue

What do you think? @asaj @mrsmkl @rcroessmann

mrsmkl:
I think that's a good idea, changed the behavior to that.

mcortesi:
Still thinking about this...
I'm not convinced that this whole pr is a good a idea...

we are defining an artificial limit to the exchange rate variation; but since that limit doesn't exist outside the system (any centralized exchange); we create the opportunity to game the system and profit from arbitrage in a non productive way for us.

In particular, taking the Max possible rate `0.000005` here are the possible variations:

Period |  Max Variation
--|--
10 seconds |  0,00500%
30 seconds | 0,02%
1 minute |  0,03%
1 hour |  1,82%
1 day |  54,03%
15 day |  65096,04%
30 day | 42505134,29%
1 year | 3,02E+68

Here we can see that if we use this max

 1. it doesn't make sense the check the limit after 15 days
 2. Stating that in can only change 1.84% in hour is probably unreal, specially in crypto.

mrsmkl:
If we change the max period from year to for example one day, then we get a bit different results. Then, if max is 0.0005, then the max change in hour would be 600%
Still need to think about if there is any advantage over just having some period of time before an oracle can post a new report. Or perhaps a period until the actual median is updated after a report.

mcortesi:
Yeah, the more i see the `SortedOracles` design, the more questions/doubts i have...

Basically, when asking for the `medianRate` (which imo is the most important method); it will get median from all rates, it won't check if they are expired.

Expiring only happens when somebody calls the `removeExpired` method.

Then i wonder what's the "mechanics" for it...

If we have a long/medium expiration time, for example 1 hour.. we'll be obtaining a rate considering information from last minute and last 59 minute, without making any distinction. Both will be equally interesting to us; when arguably something that was informed an hour ago should not be as relevant as information from a minute ago.

That means that it should be short... for example 2 minutes; and we need to call removeExpired constantly.

Then there's the other problem... we have "hundreds" of oracles reporting in a 2 minute window... which thinking in terms of block tx space not sure if it's ideal. Maybe not hundreds then ;)

But "tens of oracles" reporting... and you don't want the exchange rate to be change everytime a new report arrives... but changes happen due to a median computation, which it we assume a majority honest and up-to-date reports, it should be quite stable...

so, i'm a bit confused about what are we looking here. But it is important, since this controls the usd/gold exchange rate which is used in several parts, like rewards for validators...

cmcewen:
closing after checking with sami

PR #1492, opened 2019-10-25
[Wallet] Fetch tobin tax for exchanges

### LINKED ISSUE #1249: Wallet should calculate tobin tax for gold exchanges
### Expected Behavior

The contract should be queried to get the actual fee

### Current Behavior

The UI is all done for #955 but the fee is hardcoded

annakaz:
### Description

Fetch Tobin tax for gold/dollar exchanges in wallet to replace hardcoded value. The tobin tax is a fee (configured to 0.5%) levied on selling gold to the reserve when the reserve is low to prevent a run on the reserve

### Tested

Ran on device, exchanged dollars and gold and confirmed that Tobin tax was fetched
Added jest test for fetching tobin tax

### Related issues

- Fixes #1249 

### Backwards compatibility

Yes


packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -87,14 +90,23 @@ class ExchangeReview extends React.Component<Props> {
   componentDidMount() {
     const { makerToken, makerAmount } = this.getConfirmationInput()
     this.props.fetchExchangeRate(makerAmount, makerToken)
+    this.props.fetchTobinTax(makerAmount, makerToken)
```
jmrossy, line 136: In the time btwn when this is dispatched, and when the value comes in, the old value will still live in the store and be shown in the UI right? Does the UI block until that's in? Or can a user proceed with the incorrect value? 
Ditto concern for `fetchExchangeRate` above. We may need a solution like what we do for fees, where things block while those are refreshed.

jmrossy (changes requested):

packages/mobile/src/exchange/saga.ts
```diff
@@ -36,19 +43,70 @@ const LARGE_DOLLARS_SELL_AMOUNT_IN_WEI = new BigNumber(1000 * 100000000000000000
...
+      yield call(getConnectedAccount)
+
+      // Using native web3 contract wrapper since contractkit
+      // hasn't yet implemented tobin tax interface
+      const reserve = yield call([
+        contractKit._web3Contracts,
+        contractKit._web3Contracts.getReserve,
+      ])
+
+      const tobinTaxFraction = yield call(reserve.methods.getOrComputeTobinTax().call)
```
annakaz, line 60: Note that the Reserve contract lacks a view function for the Tobin tax so we use .call() that executes the transaction without sending it

Looked into how `web3.eth.contracts` handles method calls for a view function vs for a state-changing function that is only called- both are eventually resolved to just "calls", so we aren't making the node do more work this way. See implementation of functions on contracts here https://github.com/ethereum/web3.js/blob/v1.2.0/packages/web3-eth-contract/src/index.js#L685
cc @cmcewen

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -113,6 +116,7 @@ export class ExchangeReview extends React.Component<Props, State> {
       makerAmount = getTakerAmount(inputAmount, exchangeRate)
     }
     this.props.fetchExchangeRate(makerToken, makerAmount)
+    this.props.fetchTobinTax(makerAmount, makerToken)
```
jmrossy: I think my old comment about this got buried so posting again:

In the time btwn when this is dispatched, and when the value comes in, the old value will still live in the store and be shown in the UI right? Does the UI block until that's in? Or can a user proceed with the incorrect value?
Ditto concern for fetchExchangeRate above. We may need a solution like what we do for fees, where things block while those are refreshed.

And something else I just noticed, a `get` method like this should be pure. It's a bad place to dispatch these actions (fetch tobin tax and fetch exchange rate). These should be in a componentDidMount of DidUpdate

annakaz: Updated to dispatch the actions in componentDidMount- had originally written it this way to avoid having to pass makerToken/makerAmount within componentDidMount, but agree that it's unexpected to dispatch an action in this getter

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: Good point- created an issue to track this here https://github.com/celo-org/celo-monorepo/issues/2472

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: Created an issue to track the UI blocking- https://github.com/celo-org/celo-monorepo/issues/2472

jmrossy (approved):

PR #1494, opened 2019-10-27
Validators: uses Ethereum address for proof of possession

### LINKED ISSUE #449: Validators SBAT submit their BLS signed ethereum address as proof-of-possession
### Expected Behavior

Validators prove possession of their private key by signing their ethereum address

### Current Behavior

Validators prove possession of their private key by signing their BLS public key

kobigurk:
### Description

Currently, the proof of possession is used with the BLS public key, opening the possibility for front-running attacks. This changes the behaviour such that the Ethereum address is used.

### Tested

Runs end-to-end tests with the new method.

### Other changes

Upgrades to the latest Blake2x-based hashing algorithm in bls-zexe.

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain/issues/449

### Backwards compatibility

Not backwards compatible.


yarn.lock
```diff
@@ -4647,6 +4647,46 @@
     source-map-support "^0.5.7"
     tslib "^1.9.3"
 
+"@stablelib/binary@^0.7.2":
```
nategraf, line 4650: Is there a requirement for these new packages for this PR?

nategraf (approved):
👍

yarn.lock
kobigurk: I need the @stablelib/blake2xs package, which I guess pulls all of these.

kevjue (approved):
LGTM!

PR #1497, opened 2019-10-28
[Wallet] Improve error handling around account creation and keystore ops

### LINKED ISSUE #1535: Improve error handling around account creation and keystore retrieval

jmrossy:
### Description

- Improve error handling around account
- Better error message on invite screen when account creation fails
- Improve logging in keystore utils
- Send keystore errors to Sentry

### Tested

Ran through invite flow and import flow, no regressions

### Issues
 - Fixes #1535

### Backwards compatibility

Yes


packages/mobile/src/web3/saga.ts
```diff
@@ -167,47 +176,32 @@ export function* assignAccountFromPrivateKey(key: string) {
...
-            'Importing same account as current one'
-          )
+          Logger.warn(TAG + '@assignAccountFromPrivateKey', 'Attempted to import same account')
         } else {
           Logger.error(TAG + '@assignAccountFromPrivateKey', 'Error importing raw key')
           throw e
         }
       }
       yield call(web3.eth.personal.unlockAccount, account, pincode, UNLOCK_DURATION)
+      web3.eth.defaultAccount = account
```
jeanregisser, line 189: Do you know why this was previously done only when `!zeroSyncMode`?

packages/mobile/src/utils/keyStore.ts
```diff
@@ -1,13 +1,34 @@
...
+import Logger from 'src/utils/Logger'
+
+const TAG = 'utils/keystore'
 
 export async function setKey(key: string, value: string) {
-  return RNSecureKeyStore.set(key, value, { accessible: ACCESSIBLE.WHEN_UNLOCKED })
+  try {
+    Logger.debug(TAG, `Setting key ${key} in keystore`)
+    await RNSecureKeyStore.set(key, value, {
+      accessible: ACCESSIBLE.AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY,
```
jeanregisser: Why changing the `accessible` option now?

jeanregisser (approved):
Looks good! 👍 
I've left 2 questions.

packages/mobile/src/web3/saga.ts
jmrossy: It's still only done in !zerosyncmode, I just improved the organization.
It's think it's because we don't use web3 to sign txs in zero sync mode, not 100% sure tho

packages/mobile/src/utils/keyStore.ts
jmrossy: I was just experimenting before I realized this only affects iOS. This option seems right to me, WDYT?

packages/mobile/src/web3/saga.ts
jeanregisser: ok 👍

packages/mobile/src/utils/keyStore.ts
jeanregisser: This has a subtle implication when restoring the phone from a backup. 

If using one of the options with the `_THIS_DEVICE_ONLY` suffix, the keychain item won't be backed up. Which means users will lose their mnemonic if they restore their phone (though I know they should have manually backed up their key). But the `AsyncStorage` content will be restored.

This doesn't sound right to me as is and we should make a careful consideration about this (and same for the geth data folder and `AsyncStorage` content).

More about this:
https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlockthisdeviceonly

jmrossy: Yep that was intentional, I felt that the backup key should not be synced to Apple's servers.

jeanregisser: Will circle back on this directly on Slack.

PR #1499, opened 2019-10-28
Disable indexer in webapp pod

### LINKED ISSUE #1441: Blockscout seems to sometimes be running behind
Refresh on integration-blockscout -- sometimes the data is stale, sometimes up to date.  Is a web replica caching something?

jcortejoso:
### Description

Setting the variable `DISABLE_INDEXER=true` in the web process so some internals works as expected.

### Tested

Testing locally and in an online environment.

### Related issues

- Fixes #1441 

### Backwards compatibility

No problem


packages/helm-charts/blockscout/templates/blockscout-web.deployment.yaml
```diff
@@ -43,6 +43,7 @@ spec:
           - name: PORT
             value: "4000"
 {{ include "celo.blockscout-env-vars" .  | indent 10 }}
+{{ include "celo.blockscout-env-vars-web" .  | indent 10 }}
```
nambrot: Do you expect a lot more blockscout env vars just for the web pod? Otherwise, can we just move it into this file?

jcortejoso: I do not expect more, at least for now. We can add directly and if we add more then we can switch again to this approach.

nambrot (approved):

PR #1506, opened 2019-10-29
Fix compile error during local clean build

nategraf:
### Description

During a local clean build of the monorepo I consistently got and error saying that the type of `Web3EthAbi` has no `default` field. It seems that module does not have type support anyway so I've added a `@ts-ignore` annotation, which fixes the build.

I am not sure why, but I have gotten this error occasionally over time on clean builds.


timmoreton (approved):
If it works for you.. :-)

PR #1507, opened 2019-10-29
Move to forno terminology

tkporter:
### Description

Use `forno` for our zero sync ingress

### Tested

Created ingresses for each non-prod & prod environment

### Other changes

n/a

### Related issues

n/a

### Backwards compatibility

We'll need to make sure to use `forno` for all future deployments. The old name is still up & will be taken down once we move other services over


annakaz (approved):
LGTM!

PR #1534, opened 2019-10-30
[Wallet] Fix yarn dev command for running android

jmrossy:
### Description

`yarn dev` is broken since the RN upgrade to 0.61
This is basically because `react-native run-android` runs the bundler from the monorepo root instead of the mobile project. This is a simple workaround for now.


jeanregisser (approved):
Ship it 👍 🚢

i1skn:
Just one little downside - when app loads for the first time packager is not ready yet and you need to reload the app.

jmrossy:
> Just one little downside is that the first time the app loads - packager is not ready yet and you need to reload the app.

Strange, for me I don't need to reload the app. I just need to wait for the bundler to be ready

PR #1537, opened 2019-10-30
Fix e2e on CI

martinvol:
### Description

* Added retries for flaky `yarn`,  `yarn build` and `yarn detox`.
* Added pidcat for geth logging.
* Doesn't get the CI red if e2e fails while we investigate a detox issue: https://github.com/wix/Detox/issues/1723

### Tested

itself.

### Other changes

Deleted legacy scripts related to e2e.

### Related issues

-


.circleci/config.yml
```diff
@@ -202,8 +202,8 @@ jobs:
             #   echo "Cache found, just run post-script."
             #   yarn postinstall
             # fi
-            yarn
-            yarn build
+            yarn || yarn
```
nategraf, line 206: Is there any chance we can figure out why the `yarn` command is flaky? Retries will only get us so far and increase the time to CI reporting real failures. I understand if you don't have time now, but creating an issue with what you know and a TODO here pointing to that issue would be helpful.

nategraf: Oh, so are the failures because of wix/Detox#1723? if so a reference inline to this file would be good

jmrossy (approved):

.circleci/config.yml
martinvol: We do those retries in many places, for example here: https://github.com/celo-org/celo-monorepo/blob/8fb3ff57d64e375104ccef3aa87fcfa062c639cd/.circleci/config.yml#L93
It is because because sometimes npm has a latency problem and the whole build fails. The retry is only executed if the jobs fails. Also the build command, as it requires network resources it's not guaranteed it's 100% deterministic.

The problems related to detox is in `Run test itself`, and it has the inline reference indeed :)

Issue #1440, opened 2019-10-23
Celotool command to create flattened artifacts and deploy them to blockscout


timmoreton:

diminator:
also see #1186

diminator:
closed in favor of #1186

PR #1540, opened 2019-10-30
[Celotool] Integration with truffle blockscout plugin

### LINKED ISSUE #1186: Developers SBAT register the protocol contracts source code and ABIs in Blockscout easily
### Expected Behavior

In order to visualize the source code & abi of Smart Contracts in BlockScout, it's necessary to register that for each individual contract using the API. It involves too flatten the source code to embedded all the Solidity dependencies.

Would be nice to have a script to facilitate all this process for each protocol release.

### Current Behavior

Manual registration of source code

jcortejoso:
### Description

Added new command `bin/celotooljs.sh deploy initial verify-contracts -e berlintestnet001` --all contracts-- or `bin/celotooljs.sh deploy initial verify-contracts -e berlintestnet001 --contract RegistryProxy` to verify the deployed contracts in the online blockscout for that environment. The contract abis have to be available at `packages/protocol/build/<network_name>`

### Tested 

Used the command to validate the contracts in a testnet just after deploying the contracts.

### Related issues

- Fixes #1186 #1440

### Backwards compatibility

No compatibility problems


packages/celotool/src/cmds/deploy/initial/verify-contracts.ts
```diff
@@ -0,0 +1,50 @@
...
+export const describe = 'verify the celo smart contracts in blockscout'
+
+export const builder = (argv: yargs.Argv) => {
+  return argv
+    .option('skipClusterSetup', {
+      type: 'boolean',
+      description: 'If you know that you can skip the cluster setup',
+      default: false,
+    })
+    .option('contract', {
```
asaj: Where is this used?

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
```diff
@@ -0,0 +1,47 @@
+import * as yargs from 'yargs'
+import { InitialArgv } from '../../deploy/initial'
+import { switchToClusterFromEnv } from 'src/lib/cluster'
+import { getBlockscoutUrl } from 'src/lib/endpoints'
+import { execCmd } from 'src/lib/utils'
+
+export const command = 'verify-contracts'
+
+export const describe = 'verify the celo smart contracts in blockscout'
```
asaj: How is this command different than the initial one?

packages/protocol/truffle-config.js
```diff
@@ -146,7 +146,7 @@ if (argv.truffle_override || !(argv.network in networks)) {
 }
 
 module.exports = {
-  plugins: ['truffle-security'],
+  plugins: ['truffle-security', 'truffle-plugin-blockscout-verify'],
```
asaj, line 149: Super cool!

asaj (commented):
For my own understanding, what exactly does it mean for a contract to be "verified" in this case?

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
diminator: maybe better to use have only one command `verify-contracts` and use the `contract` arg as either:
- a single contract address
- a list of contract addresses
- all contracts

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const builder = (argv: yargs.Argv) => {
+  return argv
+    .option('skipClusterSetup', {
```
diminator: specified but not used

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+}
+
+interface VerifyContractsInitialArgv extends InitialArgv {
+  skipClusterSetup: boolean
+  contract: string
+}
+
+export const handler = async (argv: VerifyContractsInitialArgv) => {
+  await switchToClusterFromEnv()
```
diminator: is this where you want to use `skipClusterSetup`?

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
```diff
@@ -0,0 +1,47 @@
...
+  // Check if blockscout is deployed and online?
+  const blockscoutUrl = getBlockscoutUrl(argv)
```
diminator: not specified as arg

diminator (changes requested):

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
jcortejoso: No difference at all. Really it does not makes sense to have an `upgrade` command for this, so I will remove. If we want to "rerun" the verification process we can use `initial` again, and the contracts already verified will be skipped.
I have not added a `destroy` command to this. It is possible to clean the blockscout database with the contract code if somebody finds this useful, but I do not find very useful to unverify a contract, but if you see otherwise we can add.

packages/celotool/src/cmds/deploy/initial/verify-contracts.ts
jcortejoso: It is not included as a `yarn` task, but you can run for example `bin/celotooljs.sh deploy initial verify-contracts -e integration --contract Attestations --verbose` to only validate the `Attestations` contract. I find quite useful having this option bc some contract may fail to get verified in blockscout and help with debugging without iterating through all of them. 

If you are referring to `skipClusterSetup`, you are right it is not being used. Thanks!

packages/protocol/truffle-config.js
jcortejoso: 🆒

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
jcortejoso: So if you do not pass any `--contract` or you use the keyword `--contract all`, you will verify all the available contracts in `protocol/build/<network>/contracts/`. We can change the syntax to allow a list or do not use the keyword `all` for processing all. If you pass one contract it will try to verify only that contract (useful for debugging mostly).

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
jcortejoso: K8s cluster is only used to make a portForwardAnd to the rpc interface of a geth node, so I think we do not need `skipClusterSetup`. My fault.

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
jcortejoso: The blockscout url can be get from the env using the function `getBlockscoutUrl`, so we do not need to provide that as an arg. The url is propagated to the plugin as an argument so the plugin can use it.

packages/celotool/src/cmds/deploy/upgrade/verify-contracts.ts
diminator: list is not required, but if `contract != all` then we need to pass the contract address to the script

aaitor (approved):
lgtm

diminator (approved):

PR #1555, opened 2019-10-31
[Wallet] Create new carousel component

### LINKED ISSUE #1552: Create new education carousel
For verification:
https://www.figma.com/file/9lBnIzsABhO8oXE0d0c8Km/Onboarding-and-Verification?node-id=10%3A557

And also:
https://docs.google.com/document/d/1eEFVtUzp8s2IG8GfMqg2Tw8Ew-lzdMFOReUEjOVCxOo/edit

jmrossy:
### Description

This is the first of many PRs that will build the new optional verification experience.

This adds a new carousel for user education:
https://www.figma.com/file/9lBnIzsABhO8oXE0d0c8Km/Onboarding-and-Verification?node-id=233%3A1559

Remove unused verify input screen
Create new screen for first stage of new verification designs

### Tested

In app (android, still TODO iOS but the lib is cross-platform)

### Related issues

- Fixes #1552 

### Backwards compatibility

Yes

wa-carousel


packages/mobile/src/verify/VerificationLoadingScreen.tsx
```diff
@@ -0,0 +1,58 @@
...
+        text: 'Test 2',
+      },
+      {
+        text: 'Test 3',
+      },
+    ]
+    return (
+      <SafeAreaView style={styles.container}>
+        <ScrollView contentContainerStyle={styles.scrollContainer}>
+          {/* <DevSkipButton nextScreen={Screens.WalletHome} /> */}
```
i1skn, line 30: As far as I understand, `DevSkipButton` would not be visible in production builds. So, I'm curious about the reason to comment it.

jmrossy: Good question, it won't be. it was just a bit in the way as I was dev-ing this screen. Consider everything except the carousel still WIP

packages/mobile/src/verify/VerificationLoadingScreen.tsx
```diff
@@ -0,0 +1,58 @@
...
+      {
+        text: 'Test 2',
+      },
+      {
+        text: 'Test 3',
+      },
+    ]
+    return (
+      <SafeAreaView style={styles.container}>
+        <ScrollView contentContainerStyle={styles.scrollContainer}>
```
jeanregisser, line 29: If there's only the `ScrollView` in this screen (i.e. nothing outside of it), it's nicer on iOS to have the `SafeAreaView` be inside the `ScrollView` so the content doesn't get clipped when overscrolling up or down.

jeanregisser (approved):
Nice! 👍 

One thing in the screenshot is the shadow seems off compared to the figma design.

i1skn:
Other than the comment I left regarding `DevSkipButton` it LGTM 👍 !

jmrossy:
> Nice! 👍
> 
> One thing in the screenshot is the shadow seems off compared to the figma design.

Yeah, shadows in React Native are very limited on android. I've improved it in the later PRs but still doesn't look perfect.

PR #1556, opened 2019-10-31
[Wallet] Protect Backup Key and Safeguards with PIN

### LINKED ISSUE #1401: Backup flow should request pin confirmation before showing backup key
### Expected Behavior

Get user pin and confirm it before showing the backup key.
This won't be too hard for the typical case where we run a light node but it's tricky for the zero sync mode. See web3/saga.ts

### Current Behavior

No pin required

i1skn:
### Description

When a user tries to open Backup Key or Safeguards screens
she should be asked to confirm PIN before proceeding.

See the video - https://drive.google.com/open?id=1--tRRChcnbivpPvIftAdF0_6M89xIq5z
### Tested

Ran on Android device

### Other changes

This PR introduces `APP/NAVIGATE_PIN_PROTECTED` action, which can be used to navigate to a screen, which should be protected by PIN.

### Related issues

- Fixes #1401 partially, but I'm gonna create another issue to address Forno (Zero Sync) case.

### Backwards compatibility

Yes


packages/mobile/src/app/actions.ts
```diff
@@ -96,3 +117,20 @@ export const setAnalyticsEnabled = (enabled: boolean): SetAnalyticsEnabled => ({
...
+export const navigatePinProtected = (
+  routeName: string,
+  params?: NavigationParams
+): NavigatePinProtected => ({
+  type: Actions.NAVIGATE_PIN_PROTECTED,
+  routeName,
+  params,
+})
+
+export const startPinVerification = () => ({
```
jmrossy: You should type these action creator to use the interface you defined above:
`export const startPinVerification = (): StartPinVerification => ({`

packages/mobile/src/backup/BackupSocialIntro.tsx
```diff
@@ -3,30 +3,41 @@ import colors from '@celo/react-components/styles/colors'
...
 import { Namespaces } from 'src/i18n'
 import backupIcon from 'src/images/backup-icon.png'
 import { headerWithBackButton } from 'src/navigator/Headers'
-import { navigate, navigateHome } from 'src/navigator/NavigationService'
+import { navigateHome } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { RootState } from 'src/redux/reducers'
 
+interface StateProps {
+  doingPinVerification: boolean
```
jmrossy, line 22: I don't think this screen needs this pin check. It's redundant because the user needs to enter their pin to reach this far.

packages/mobile/test/schemas.ts
```diff
@@ -9,6 +9,7 @@ export const vNeg1Schema = {
     numberVerified: false,
     error: null,
     dismissErrorAfter: null,
+    doingPinVerification: false,
```
jmrossy: This should go in the latest schema version (v3). The old ones cannot be modified.

jmrossy (commented):
Nice work!! Just a few comments to address

packages/mobile/src/backup/BackupSocialIntro.tsx
i1skn: Actually not, because I've disabled the check only if they are going to the introduction. This, to my opinion, makes more sense - so the user can read about the feature before entering the PIN. What do you think? Also I've covered this case in the video.

packages/mobile/src/app/saga.ts
```diff
@@ -130,8 +141,30 @@ export function handleDeepLink(deepLink: string) {
   // Other deep link handlers can go here later
 }
 
+export function* navigatePinProtected(action: NavigatePinProtected) {
+  const zeroSyncMode = yield select(zeroSyncSelector)
+  try {
+    if (!zeroSyncMode) {
+      const pincode = yield call(getPincode, false)
+      yield put(startPinVerification())
+      const account = yield call(getConnectedAccount)
```
jmrossy: This line is potentially problematic. It means that the navigation won't proceed until the app is done syncing, which seems like an unnecessary requirement for this. Recommend using `getAccount` instead

jmrossy (approved):
Ship it! 💯

PR #1558, opened 2019-10-31
Add EpochRewards smart contract to calculate epoch rewards and payments

asaj:
### Description
This PR adds `EpochRewards.sol`, a smart contract for calculating the actual epoch reward to voters and payments to validators.

### Tested

Unit and end-to-end tests.

### Other changes

None
### Related issues

- Fixes #TODO

### Backwards compatibility

Not backwards compatible


packages/contractkit/src/kit.ts
```diff
@@ -89,6 +89,7 @@ export class ContractKit {
       this.contracts.getReserve(),
       this.contracts.getStableToken(),
       this.contracts.getValidators(),
+      // this.contracts.getEpochRewards(),
```
nategraf: Nit: Unlike in `packages/contractkit/src/contract-cache.ts` this seems to be the only commented out call here. Maybe forgo adding it.

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,384 @@
...
+
+/**
+ * @title Contract for calculating epoch rewards.
+ */
+contract EpochRewards is Ownable, Initializable, UsingPrecompiles, UsingRegistry {
+
+  using FixidityLib for FixidityLib.Fraction;
+  using SafeMath for uint256;
+
+  uint256 constant GENESIS_GOLD_SUPPLY = 600000000000000000000000000;
```
nategraf: Nit: I had to plug this separate out all the zeros to figure out how much this is :P
```suggestion
  uint256 constant GENESIS_GOLD_SUPPLY = 600000000000000000000000000; // 600 million Gold
```

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,384 @@
...
+  uint256 constant GOLD_SUPPLY_CAP = 1000000000000000000000000000;
```
nategraf:
```suggestion
  uint256 constant GOLD_SUPPLY_CAP = 1000000000000000000000000000; // 1 billion Gold
```

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,384 @@
...
+
+  struct RewardsMultiplierAdjustmentFactors {
+    FixidityLib.Fraction underspend;
+    FixidityLib.Fraction overspend;
+  }
+
+  struct RewardsMultiplierParameters {
+    RewardsMultiplierAdjustmentFactors adjustmentFactors;
+    FixidityLib.Fraction max;
+  }
```
nategraf: This is not particularly each to parse. Could you add some comments here as to what these are. Also why is `RewardsMultiplierAdjustmentFactors` a separate struct from `RewardsMultiplierParameters`?

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,384 @@
...
+  function getTargetGoldTotalSupply() public view returns (uint256) {
+    uint256 timeSinceInitialization = now.sub(startTime);
+    if (timeSinceInitialization < SECONDS_LINEAR) {
+      // Pay out half of all block rewards linearly.
+      uint256 linearRewards = GOLD_SUPPLY_CAP.sub(GENESIS_GOLD_SUPPLY).div(2);
+      uint256 targetRewards = linearRewards.mul(timeSinceInitialization).div(SECONDS_LINEAR);
+      return targetRewards.add(GENESIS_GOLD_SUPPLY);
+    } else {
+      // TODO(asa): Implement block reward calculation for years 15-30.
+      return 0;
```
nategraf, line 231: An `assert` or failing `require` may be appropriate here.

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,384 @@
...
+  struct RewardsMultiplierParameters {
+    RewardsMultiplierAdjustmentFactors adjustmentFactors;
+    FixidityLib.Fraction max;
+  }
+
+  struct TargetVotingYieldParameters {
+    FixidityLib.Fraction target;
+    FixidityLib.Fraction adjustmentFactor;
+    FixidityLib.Fraction max;
+  }
```
nategraf: Can you add comments on each of these structs. That would help make this easier to understand on first pass.

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -418,14 +415,16 @@ contract Validators is
...
+      FixidityLib.Fraction memory totalPayment = FixidityLib.newFixed(maxPayment).multiply(
+        validators[account].score
+      );
       uint256 groupPayment = totalPayment.multiply(groups[group].commission).fromFixed();
       uint256 validatorPayment = totalPayment.fromFixed().sub(groupPayment);
       getStableToken().mint(group, groupPayment);
       getStableToken().mint(account, validatorPayment);
+      return totalPayment.fromFixed();
     }
+    return 0;
```
nategraf: Could you explain this return value somewhere?

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -119,6 +125,25 @@ describe('governance tests', () => {
     return blockNumber % epochSize === 0
   }
 
+  const assertDifferenceCloseTo = (
```
nategraf: Based on the way this is used a signature of `assertAlmostEqual(actual: BigNumber, expected: BigNumber, delta: BigNumber = new BigNumber(10).pow(12).times(5))` would be clearer. The call site might be `assertAlmostEqual(currentBalance.minus(previousBalance), expected)`

packages/protocol/contracts/governance/EpochRewards.sol
asaj: Yeah, not my finest work here not documenting these. Let me know if you think it's more readable now.

packages/protocol/contracts/governance/Validators.sol
asaj: 🤦‍♂

packages/protocol/migrationsConfig.js
```diff
@@ -26,6 +26,22 @@ const DefaultConfig = {
...
+    },
+    rewardsMultiplierParameters: {
+      max: 2,
+      adjustmentFactors: {
+        underspend: 1 / 2,
+        overspend: 5,
+      },
+    },
+    targetVotingGoldFraction: 2 / 3,
+    maxValidatorEpochPayment: '1000000000000000000',
```
rcroessmann: Did we decide to increase this from 75k to 100k cUSD?

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,569 @@
...
+                {
+                  from: nonOwner,
+                }
+              )
+            )
+          })
+        })
+      })
+
+      describe('when the parameters are the same', () => {
```
nategraf, line 278: It may be worth testing that each factor can be adjusted separately (i.e. if one if the same and the other is different this call works.)

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,569 @@
...
+        // Assert equal to 9 decimal places due to fixidity imprecision.
+        assert.equal(expected.dp(9).toFixed(), actual.dp(9).toFixed())
+      })
+    })
+  })
+
+  describe.only('#updateTargetVotingYield()', () => {
+    const randomAddress = web3.utils.randomHex(20)
+    const totalSupply = new BigNumber(129762987346298761037469283746)
+    const reserveBalance = new BigNumber(2397846127684712867321)
```
nategraf, line 458: Could you note whether or not these are arbitrary? (So future editors know how important it is to preserve them)

packages/protocol/contracts/governance/EpochRewards.sol
nategraf: Those comments are very helpful 👍

nategraf (approved):

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -119,6 +125,23 @@ describe('governance tests', () => {
     return blockNumber % epochSize === 0
   }
 
+  const assertAlmostEqual = (
```
m-chrzan: Maybe move this to utils.ts?

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -119,6 +125,23 @@ describe('governance tests', () => {
...
+    actual: BigNumber,
+    expected: BigNumber,
+    delta: BigNumber = new BigNumber(10).pow(12).times(5)
+  ) => {
+    if (expected.isZero()) {
```
m-chrzan: Why is zero a special case?

packages/contractkit/src/contract-cache.ts
```diff
@@ -4,6 +4,7 @@ import { AccountsWrapper } from './wrappers/Accounts'
 import { AttestationsWrapper } from './wrappers/Attestations'
 import { BlockchainParametersWrapper } from './wrappers/BlockchainParameters'
 import { ElectionWrapper } from './wrappers/Election'
+// import { EpochRewardsWrapper } from './wrappers/EpochRewards'
```
m-chrzan, line 7: Is there a reason to keep this commented out code (here and a few spots below)? Maybe a TODO would be better?

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,399 @@
...
+/**
+ * @title Contract for calculating epoch rewards.
+ */
+contract EpochRewards is Ownable, Initializable, UsingPrecompiles, UsingRegistry {
+
+  using FixidityLib for FixidityLib.Fraction;
+  using SafeMath for uint256;
+
+  uint256 constant GENESIS_GOLD_SUPPLY = 600000000000000000000000000; // 600 million Gold
+  uint256 constant GOLD_SUPPLY_CAP = 1000000000000000000000000000; // 1 billion Gold
```
m-chrzan: We could use the `ether` suffix to make these a little more readable (https://solidity.readthedocs.io/en/v0.5.12/units-and-global-variables.html#ether-units).

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,399 @@
...
+  }
+
+  /**
+   * @notice Sets the target voting gold fraction.
+   * @param value The percentage of floating gold voting to target.
+   * @return True upon success.
+   */
+  function setTargetVotingGoldFraction(uint256 value) public onlyOwner returns (bool) {
+    require(value != targetVotingGoldFraction.unwrap() && value < FixidityLib.fixed1().unwrap());
+    targetVotingGoldFraction = FixidityLib.wrap(value);
```
m-chrzan, line 137: Consider reordering these lines and using Fixidity comparison operators to reduce `unwrap`s.

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,399 @@
...
+    return (
+      params.max.unwrap(),
+      params.adjustmentFactors.underspend.unwrap(),
+      params.adjustmentFactors.overspend.unwrap()
+    );
+  }
+
+  /**
+   * @notice Sets the target voting gold fraction.
+   * @param value The percentage of floating gold voting to target.
```
m-chrzan: Nit: Capitalization of "gold" is inconsistent throughout the file. Here it's lowercase, in the comment to `struct RewardsMultiplierParameters ` it's uppercase. I'd vote for "Gold" since "Celo Dollars" is uppercase and that feels natural.

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,399 @@
...
+    uint256 max,
+    uint256 underspendAdjustmentFactor,
+    uint256 overspendAdjustmentFactor
+  )
+    public
+    onlyOwner
+    returns (bool)
+  {
+    require(
+      max != rewardsMultiplierParams.max.unwrap() ||
```
m-chrzan, line 177: Similar as above, consider first wrapping and using Fixidity comparisons.

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,399 @@
...
+  function setTargetVotingYieldParameters(
+    uint256 max,
+    uint256 adjustmentFactor
+  )
+    public
+    onlyOwner
+    returns (bool)
+  {
+    require(
+      max != targetVotingYieldParams.max.unwrap() ||
```
m-chrzan, line 204: Ditto.

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -0,0 +1,399 @@
...
+    uint256 remainingSupply = GOLD_SUPPLY_CAP.sub(totalSupply.add(targetGoldSupplyIncrease));
+    uint256 targetRemainingSupply = GOLD_SUPPLY_CAP.sub(targetSupply);
+    FixidityLib.Fraction memory ratio = FixidityLib.newFixed(remainingSupply).divide(
+      FixidityLib.newFixed(targetRemainingSupply)
+    );
+    if (ratio.gt(FixidityLib.fixed1())) {
+      FixidityLib.Fraction memory delta = ratio.subtract(FixidityLib.fixed1()).multiply(
+        rewardsMultiplierParams.adjustmentFactors.underspend
+      );
+      FixidityLib.Fraction memory r = FixidityLib.fixed1().add(delta);
```
m-chrzan: Wondering if these variables could get better names to make the logic more readable. E.g. `ratio` -> `remainingToTargetRatio` makes it clear in further computations what that variable is.

m-chrzan (commented):
Reviewed up to EpochRewards.sol. 

Would be great to get a clear explanation of how this works in the docs, though probably out of scope of this PR if we want to get it in soon.

packages/protocol/migrationsConfig.js
asaj: Fixed

packages/protocol/test/governance/epochrewards.ts
asaj: Done, albeit in a lazy way since getting this in ASAP is high priority and this feels low risk

packages/celotool/src/e2e-tests/governance_tests.ts
asaj: Because if the value is zero we don't need to worry about rounding errors

packages/contractkit/src/contract-cache.ts
asaj: Just following the existing pattern

packages/protocol/contracts/governance/EpochRewards.sol
asaj: Will that allow us to compare value to the previous targetVotingGoldFraction? (we will have overwritten it)

Was trying to follow the checks/effects/interactions pattern. Doing it in this order also saves gas in the case where value doesn't change the targetVotingGoldFraction

packages/protocol/contracts/governance/EpochRewards.sol
asaj: Same answer, if we wrap first (and overwrite storage) we can't do this check anymore

packages/protocol/contracts/governance/EpochRewards.sol
asaj: Ditto

packages/protocol/contracts/governance/EpochRewards.sol
m-chrzan: Ah right. I guess the other option would have been to declare a new Fixidity fraction in memory, but that could cost more gas and would definitely add code noise.

packages/protocol/contracts/governance/test/EpochRewardsTest.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.5.8;
+
+import "../EpochRewards.sol";
+import "../../common/FixidityLib.sol";
```
m-chrzan: Is this import necessary?

packages/protocol/migrations/20_elect_validators.ts
```diff
@@ -117,13 +117,15 @@ async function registerValidator(
   ).toString('hex')
   const publicKeysData = publicKey + blsPublicKey + blsPoP
 
+  console.log('locking gold for validator registration')
```
m-chrzan: Mega-nit: I think by convention we capitalize these debug logs in migrations. Here and below.

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,576 @@
...
+
+  const targetVotingYieldParams = {
+    initial: toFixed(new BigNumber(1 / 20)),
+    max: toFixed(new BigNumber(1 / 5)),
+    adjustmentFactor: toFixed(new BigNumber(1 / 365)),
+  }
+  const rewardsMultiplier = {
+    max: toFixed(new BigNumber(2)),
+    adjustments: {
+      underspend: toFixed(new BigNumber(1 / 2)),
```
m-chrzan, line 61: Will we always want `underspend`  < 1 and `overspend` > 1? If so, should these conditions be required in the contract?

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,576 @@
...
+      overspend: toFixed(new BigNumber(5)),
+    },
+  }
+  const targetVotingGoldFraction = toFixed(new BigNumber(2 / 3))
+  const targetValidatorEpochPayment = new BigNumber(10000000000000)
+  const exchangeRate = 7
+  const randomAddress = web3.utils.randomHex(20)
```
m-chrzan: Naming suggestion: would `mockStableAddress` be clearer?

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,576 @@
...
+
+  describe('#setTargetValidatorEpochPayment()', () => {
+    describe('when the payment is different', () => {
+      const newPayment = targetValidatorEpochPayment.plus(1)
+
+      describe('when called by the owner', () => {
+        let resp: any
+
+        beforeEach(async () => {
+          resp = await epochRewards.setTargetValidatorEpochPayment(newPayment)
```
m-chrzan: Nit: I think it's conventional, when unit testing a particular method, to have the call be in an `it`, even if you could deduplicate it in a `beforeEach`. `beforeEach` is more about the setup needed before the call is made.

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,576 @@
...
+
+  describe('#setTargetVotingGoldFraction()', () => {
+    describe('when the fraction is different', () => {
+      const newFraction = targetVotingGoldFraction.plus(1)
+
+      describe('when called by the owner', () => {
+        let resp: any
+
+        beforeEach(async () => {
+          resp = await epochRewards.setTargetVotingGoldFraction(newFraction)
```
m-chrzan: Ditto here and elsewhere.

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,576 @@
...
+      it('should return a percentage of the active votes', async () => {
+        const expected = fromFixed(targetVotingYieldParams.initial).times(activeVotes)
+        assertEqualBN(await epochRewards.getTargetEpochRewards(), expected)
+      })
+    })
+  })
+
+  describe('#getTargetTotalEpochPaymentsInGold()', () => {
+    describe('when a StableToken exchange rate is set', () => {
+      // Hard coded in EpochRewardsTest.sol
```
m-chrzan: Would reasoning about this code/writing potential future tests be clearer if we allowed setting `numValidators` in the mock contract and did that here?

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,576 @@
...
+          .integerValue(BigNumber.ROUND_FLOOR)
+        await mockGoldToken.setTotalSupply(totalSupply)
+      })
+
+      it('should return one plus 10% times the underspend adjustment', async () => {
+        const actual = fromFixed(await epochRewards.getRewardsMultiplier())
+        const expected = new BigNumber(1).plus(
+          fromFixed(rewardsMultiplier.adjustments.underspend).times(0.1)
+        )
+        // Assert equal to 7 decimal places due to fixidity imprecision.
```
m-chrzan, line 430: Should this be pulled out to a new assertion helper?

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -0,0 +1,576 @@
...
+          fromFixed(rewardsMultiplier.adjustments.overspend).times(0.1)
+        )
+        // Assert equal to 7 decimal places due to fixidity imprecision.
+        assert.equal(expected.dp(7).toFixed(), actual.dp(7).toFixed())
+      })
+    })
+  })
+
+  describe('#updateTargetVotingYield()', () => {
+    const randomAddress = web3.utils.randomHex(20)
```
m-chrzan: Similar here as near the beginning of this file, might be worth renaming this to `mockReserveAddress` or similar for context.

packages/protocol/test/governance/epochrewards.ts
asaj: Good call

m-chrzan (approved):
LGTM pending outstanding comments.

packages/protocol/migrations/20_elect_validators.ts
asaj: Removed, had added for debugging

packages/protocol/test/governance/epochrewards.ts
asaj: Actually, after thinking about this more, I think this doesn't strictly need to be true, wdyt @rcroessmann?

rcroessmann: I think we should enforce both. Especially the condition overspend > 1 seems necessary as the protocol could end up trying to spend more than the remaining cGLD for overspend <= 1.

PR #1560, opened 2019-10-31
Avoid re-encrypting key files with yarn keys:encrypt command

nategraf:
### Description

Currently the `yarn keys:encrypt` command encrypts files, which changes them and is a bit of annoying. This PR is a quick fix to that.


martinvol:
Can you add a couple of comments to explain how this works? I'm struggling to get it.

nategraf:
I added some comments and a usage string 👍

jeanregisser:
Cool! this was annoying me too 👍 
From what I see, we now have to delete the encrypted files we want to update.

Next step would be to detect which encrypted files need to be updated. 

But anyway already a nice improvement 🎉

tkporter (approved):

martinvol (approved):

PR #1565, opened 2019-11-01
[Wallet] Update link styles and Implement VerificationEducationScreen

### LINKED ISSUE #1521: Replace green link color with underlined text
### Expected Behavior

text should be the same color as other text in-paragraph, but underlined to indicate link status

### Current Behavior

Currently, bold green is used for inline links (eg celo.org/build.wallet, Terms page in settings)

NOTE: taking this on as part of optional verifiaction

jmrossy:
### Description

- Implement new Link styles and break out text buttons into a separate component
- Clean up related fonts and styles
- Implement VerificationEducationScreen

### Tested

See screenshot for education screen.
Checked all other affected screens, design is as intended

### Related issues

- Fixes #1521 

### Backwards compatibility

Yes

wa-ver-edu


packages/mobile/src/account/Account.tsx
```diff
@@ -81,6 +81,10 @@ export class Account extends React.Component<Props, State> {
     const verified = await isPhoneNumberVerified(phoneNumber)
     this.setState({ verified })
     this.setState({ version: DeviceInfo.getVersion() })
+    navigate(Screens.BackupQuiz, {
```
jmrossy: Mistake, will remove

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -1,21 +1,66 @@
...
+          {/** TODO use new icon when it's ready */}
+          <NuxLogo testID="VerificationEducationIcon" />
+          <Text style={styles.h1} testID="VerificationEducationHeader">
+            {t('education.header')}
+          </Text>
+          <Text style={styles.body}>{t('education.body1')}</Text>
+          <Text style={styles.body}>{t('education.body2')}</Text>
+          <Link onPress={this.onPressLearnMore}>{t('education.learnMore')}</Link>
         </ScrollView>
+        <>
```
i1skn, line 48: Do you think we can remove the fragment here?

i1skn (commented):
Overall LGTM and great work 👍 ! Just one little question.

packages/mobile/src/verify/VerificationEducationScreen.tsx
jmrossy: Actually, we do need this fragment because the container uses a justifyCOntent: space-between property to force the buttons to the bottom. We don't want any additional spacing btwn the buttons so they need to be grouped :)

i1skn: 👍

i1skn (approved):

jeanregisser (approved):
All good :+1:

Issue #501, opened 2019-08-14
Devs SBAT create governance proposal to upgrade changed contracts


timmoreton:
### Expected Behavior

Detect changes, deploy new contracts, create a governance proposal that repoints proxies at new contracts.  

### Current Behavior

Governance now owns Proxy contracts, so `yarn run upgrade` no longer works to upgrade contract changes on specified network. (The script at `packages/protocol/scripts/truffle/upgrade.ts` is outdated, as it assumes the `MultiSig` contract is deployed. However, we don't deploy it anymore, so the script breaks.)

yorhodes:
2 options:
1. use hotfixes
2. use governance proposals

Analysis:
- hotfixes have fewer timing dependencies compared to governance proposal phases
- hotfixes might seem centrally unilateral from an optics perspective

Concrete steps:
1. migrate functionality from protocol upgrade script to contractkit for constructing proxy logic update transaction
2. add functionality to contractkit Governance wrapper for <encoding/hashing> transaction and submitting <proposal/hotfix>
3. add validator commands to CLI for interacting with governance <hotfix/proposals>
4. migrate functionality from protocol upgrade script to <celotool> for detecting changes between network bytecode and local bytecode and performing upgrade via CLI/contractkit

yorhodes:
closing due to #2142 being merged

PR #1570, opened 2019-11-03
Support governance proposals/hotfixes in contractkit

### LINKED ISSUE #672: Create ContractWrapper for Governance
### Expected Behavior

Create ContractWrapper for Governance

yorhodes:
### Description

Adds interfaces to `src/wrappers/Governance` for interacting with proposals/hotfixes
Adds `src/governance/proposals` for constructing proposal/hotfix inputs

### Tested

Added `src/wrappers/Governance.test` to test e2e proposal functionality
Hotfix tests disabled until #1737 is fixed

### Other changes
- Touches #501
- Fixes #672

Adds `timeTravel` to `test-utils/test-ganache`
Updates, simplifies, and dedupes type converters in `src/wrappers/BaseWrapper`
Migrates util functionality to `/packages/utils/src/address`
Adds positional `args` to `src/explorer/block-explorer/CallDetails` interface


packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -27,6 +27,18 @@ export function toBigNumber(input: string) {
   return new BigNumber(input)
 }
 
+function trimLeading0x(input: string) {
```
mcortesi: there's a function in @celo/utils address.ts or if not, it should live there

packages/utils/src/collections.ts
```diff
@@ -22,3 +22,17 @@ export function zip3<A, B, C>(as: A[], bs: B[], cs: C[]) {
 export function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
   return value !== null && value !== undefined
 }
+
+// https://stackoverflow.com/questions/33355528/filtering-an-array-with-a-function-that-returns-a-promise
+export function mapAsync<T, U>(
```
mcortesi: there's `concurrentMap` for this. As a rule of thumb, we should be aware of running an unbounded amount of concurrent functions. There's always some bound to consider.

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +136,298 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
       stageDurations: res[4],
     }
   }
+
+  getProposalMetadata: (proposalID: BigNumber) => Promise<ProposalMetadata> = proxyCall(
```
mcortesi: add doc comments

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +136,298 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+    (res) => ({
+      proposer: res[0],
+      deposit: toBigNumber(res[1]),
+      timestamp: toBigNumber(res[2]),
+      transactionCount: toBigNumber(res[3]).toNumber(),
+    })
+  )
+
+  getProposalTransaction: (
+    proposalID: BigNumber,
```
mcortesi: probably it's nice for this to be a number, not bigumber... or support both

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +136,298 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+    txIndex: number
+  ) => Promise<Transaction> = proxyCall(
+    this.contract.methods.getProposalTransaction,
+    tupleParser(parseNumber, parseNumber),
+    (res) => ({
+      value: toBigNumber(res[0]),
+      destination: res[1],
+      // @ts-ignore string[] bytes type
+      data: toBuffer(res[2]),
```
mcortesi: i have a 

```ts
export function parseBytes(input: string): Array<string | number[]> {
  return input as any
}
```
 we can have a `toBuffer` to accept string[], instead of the ts-ignore. it's a common thing this ts-ignore, we should encapsulate it in one place.

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +136,298 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+    })
+  )
+
+  isApproved: (proposalID: BigNumber) => Promise<boolean> = proxyCall(
+    this.contract.methods.isApproved, 
+    tupleParser(parseNumber), 
+    identity
```
mcortesi: you can skip this argument

packages/contractkit/src/wrappers/Governance.ts
mcortesi: there's a `NumberLike` type

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +136,298 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+    const encoded = this.getTransactionsEncoded(transactions)
+    return keccak256(
+      this.kit.web3.eth.abi.encodeParameters(
+        ['uint256[]', 'address[]', 'bytes', 'uint256[]'],
+        [encoded.values, encoded.destinations, encoded.data, encoded.dataLengths]
+      )
+    ) as Buffer
+  }
+
+  buildTransactionsFromJSON = (jsonTransactions: JSONTransaction[]) =>
```
mcortesi: this can work.. but doesn't provide much type safety; since it's highly dynamic.
When thinking about proposal, my idea, was that we can add a method to `CeloTransaction` to "obtain" either the tx json... or convert it in a proposal.

or maybe have a proposal builder where we add CeloTransactions.

wdyt?

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +136,298 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+      this.kit,
+      this.contract.methods.revokeUpvote(lesserID.toString(), greaterID.toString())
+    )
+  }
+
+  async approve(proposalID: BigNumber) {
+    const proposalIndex = await this.getDequeueIndex(proposalID)
+    return toTransactionObject(
+      this.kit,
+      this.contract.methods.approve(proposalID.toString(), proposalIndex)
```
mcortesi: we should use `toFixed()` for BigNumbers. Sometimes, toString() returns `1e10` format

mcortesi (commented):
left a few comments.

Governance wrapper is missing all the doc strings

packages/utils/src/collections.ts
yorhodes: how should I think about the bound? to me it seems like any number of `call()`s can be safely concurrent

packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -27,13 +30,25 @@ export function toBigNumber(input: string) {
...
 
+export function toBuffer(input: string | string[]) {
+  return Buffer.from(trimLeading0x(input as string), 'hex')
+}
+
+export function parseBuffer(buf: Buffer) {
+  return '0x' + buf.toString('hex')
+}
+
+export function parseBytes(input: Buffer): Array<string | number[]> {
```
mcortesi: this already exists in master

packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -27,13 +30,25 @@ export function toBigNumber(input: string) {
   return new BigNumber(input)
 }
 
+export function toBuffer(input: string | string[]) {
+  return Buffer.from(trimLeading0x(input as string), 'hex')
+}
+
+export function parseBuffer(buf: Buffer) {
```
mcortesi: we probably need better names for these functions. we have a buffer -> string and a buffer -> bytes

packages/contractkit/src/wrappers/Governance.test.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  beforeAll(async () => {
+    accounts = await web3.eth.getAccounts()
+    kit.defaultAccount = accounts[0]
+    governance = await kit.contracts.getGovernance()
+    registry = await kit._web3Contracts.getRegistry()
+    lockedGold = await kit.contracts.getLockedGold()
+    accountWrapper = await kit.contracts.getAccounts()
+
+    await concurrentMap(4, accounts.slice(0, 4), async (account) => {
```
mcortesi, line 36: it's strange to see this, since these accounts are already registered as validators?

maybe you can register them as validators here... idk

packages/contractkit/src/wrappers/TransactionBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export interface JSONTransaction {
+  value: NumberLike
+  celoContractName: CeloContract
+  methodName: string
+  args: any[]
+}
+
+type Web3Method = Contract['methods'][keyof Contract['methods']]
+export class TransactionBuilder {
```
mcortesi: I would call this ProposalBuilder even it's just txs. They are also partial tx for Proposal only

packages/contractkit/src/wrappers/TransactionBuilder.ts
```diff
@@ -0,0 +1,103 @@
+import { keccak256 } from 'ethereumjs-util'
```
mcortesi: Move this outside wrappers. Take a look about the Metadabuilder that nam did fro identity. I think there's an identity folder for it. Maybe have a governance folder

packages/contractkit/src/wrappers/TransactionBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+  methodName: string
+  args: any[]
+}
+
+type Web3Method = Contract['methods'][keyof Contract['methods']]
+export class TransactionBuilder {
+  public transactions: Transaction[]
+  private kit: ContractKit
+
+  constructor(_kit: ContractKit, _transactions: Transaction[] = []) {
```
mcortesi: typescript tip:
```ts
constructor(private kit: ContractKit, private transactions: Transaction[]) {}
```
and remove the var declaration.

also maybe `private readonly`

packages/contractkit/src/wrappers/TransactionBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+    return keccak256(
+      this.kit.web3.eth.abi.encodeParameters(
+        ['uint256[]', 'address[]', 'bytes', 'uint256[]'],
+        [encoded.values, encoded.destinations, encoded.data, encoded.dataLengths]
+      )
+    ) as Buffer
+  }
+
+  get encoded() {
+    return encodedTransactions(this.transactions)
```
mcortesi: probably encodedTransactions belongs outside the wrapper file

mcortesi: also maybe create a `.toJSON()` method instead.

packages/contractkit/src/wrappers/TransactionBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  appendCeloTx(tx: CeloTransactionObject<any>) {
```
mcortesi: add another argument for tx arguments, same as `.send()` has within CeloTransactionObject. The user should be able to specify `value` for example. And not the `from`

mcortesi (commented):
more comments

packages/contractkit/src/wrappers/TransactionBuilder.ts
mcortesi: thinking this a bit more.

We should be able to extract: `data` and `to` from the `CeloTransactionObject`. same applies to the web `TransactionObject`.
as an optional extra argument we still need the `value`.

You shouldn't use `defaultParams` directly. This are optional and not required.

I suggest we add an accessor in CeloTransactionObject for `to` and `encodedAbi`.

How to implement those?

abi it's easy, just call `tx.encodeAbi()`. 

for  `to` it's tricky, the internal `txo` doesn't expose it publicly, but i've looked at the code and you should be able to do `txo._parent.options.address`

This is the actual web3 code:
```js
Contract.prototype._processExecuteArguments = function _processExecuteArguments(args, defer) {
    var processedArgs = {};

    processedArgs.type = args.shift();

    // get the callback
    processedArgs.callback = this._parent._getCallback(args);

    // get block number to use for call
    if(processedArgs.type === 'call' && args[args.length - 1] !== true && (_.isString(args[args.length - 1]) || isFinite(args[args.length - 1])))
        processedArgs.defaultBlock = args.pop();

    // get the options
    processedArgs.options = (_.isObject(args[args.length - 1])) ? args.pop() : {};

    // get the generateRequest argument for batch requests
    processedArgs.generateRequest = (args[args.length - 1] === true)? args.pop() : false;

    processedArgs.options = this._parent._getOrSetDefaultOptions(processedArgs.options);
    processedArgs.options.data = this.encodeABI();

    // add contract address
    if(!this._deployData && !utils.isAddress(this._parent.options.address))
        throw new Error('This contract object doesn\'t have address set yet, please set an address first.');

    if(!this._deployData)
        processedArgs.options.to = this._parent.options.address;

    // return error, if no "data" is specified
    if(!processedArgs.options.data)
        return utils._fireError(new Error('Couldn\'t find a matching contract method, or the number of parameters is wrong.'), defer.eventEmitter, defer.reject, processedArgs.callback);

    return processedArgs;
};
```
https://github.com/ethereum/web3.js/blob/ba00dbab7cb0000a76fe67dbaf246e23ad974271/packages/web3-eth-contract/src/index.js#L720

Also, `to` might be overrided by the defaultParams... but i wonder if we should allow that in the CeloTxObject

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +118,369 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+    // is upvoter targeting a valid proposal?
+    if (ZERO_BN.isLessThan(proposalID)) {
+      const proposalIdx = queue.findIndex((qp) => qp.id.isEqualTo(proposalID))
+      // is target proposal in queue?
+      if (proposalIdx !== -1) {
+        const weight = await this.getVoteWeight(upvoter)
+        queue[proposalIdx].upvotes = queue[proposalIdx].upvotes.plus(weight)
+        searchID = proposalID
+      } else {
+        throw new Error(`Proposal ${parseNumber(proposalID)} not in queue`)
```
mcortesi: maybe `parse` fails and you can't throw the error ;)

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +118,369 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+
+  /**
+   * Applies provided upvoter's upvote to given proposal.
+   * @param proposalID Governance proposal UUID
+   * @param upvoter Address of upvoter
+   */
+  async upvote(proposalID: NumberLike, upvoter: Address) {
+    const exists = await this.proposalExists(proposalID)
+    if (!exists) {
+      throw new Error(`Proposal ${parseNumber(proposalID)} does not exist`)
```
mcortesi: same

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +118,369 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+    const enc = encodedTransactions(transactions)
+    return toTransactionObject(
+      this.kit,
+      this.contract.methods.executeHotfix(enc.values, enc.destinations, enc.data, enc.dataLengths)
+    )
+  }
+}
+
+export function encodedTransactions(transactions: Transaction[]) {
+  if (transactions.length === 0) {
```
mcortesi: this doesn't seem to be a check that belong to this fuction, but to the propose() method

packages/contractkit/src/wrappers/TransactionBuilder.ts
yorhodes: can't seem to access `_parent` on the txo,
I came up with a "pipelined" design, let me know what you think

packages/utils/src/address.ts
```diff
@@ -3,11 +3,14 @@ import { privateToAddress, toChecksumAddress } from 'ethereumjs-util'
 export type Address = string
 
 export function eqAddress(a: Address, b: Address) {
-  return a.replace('0x', '').toLowerCase() === b.replace('0x', '').toLowerCase()
+  return trimLeading0x(a).toLowerCase() === trimLeading0x(b).toLowerCase()
 }
 
+export function trimLeading0x(input: string) {
```
jmrossy: This is a duplicate of this: https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/signatureUtils.ts#L147

Though this is actually easier to find, so please migrate the of the other one over to this. Ditto for `eqAddress ` above and `areAddressesEqual` in sigUtils

packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -27,17 +30,26 @@ export function toBigNumber(input: string) {
...
+}
+
+type bytes = Array<string | number[]>
+export function parseBytes(input: Buffer | string): bytes {
+  return input as any
+}
+
 /** Parse string -> int */
 export function toNumber(input: string) {
   return parseInt(input, 10)
```
jmrossy: I realize you didn't write this but can you make this use BigNumber which is more robust and used for number parsing elsewhere?
cc @mcortesi

packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -27,17 +30,26 @@ export function toBigNumber(input: string) {
...
 }
 
 export function parseNumber(input: NumberLike) {
```
jmrossy: What's the difference of this and toNumber above?

jmrossy (changes requested):

packages/contractkit/src/test-utils/jest.setup.js
```diff
@@ -0,0 +1,2 @@
+let one_ms = 1000
+jest.setTimeout(10 * one_ms)
```
mcortesi: do we want this for every test?

packages/utils/src/address.ts
```diff
@@ -3,11 +3,19 @@ import { privateToAddress, toChecksumAddress } from 'ethereumjs-util'
...
 export function eqAddress(a: Address, b: Address) {
-  return a.replace('0x', '').toLowerCase() === b.replace('0x', '').toLowerCase()
+  return trimLeading0x(a).toLowerCase() === trimLeading0x(b).toLowerCase()
+}
+
+export function trimLeading0x(input: string) {
+  return input.startsWith('0x') ? input.slice(2) : input
+}
+
+export function prependLeading0x(input: string) {
```
mcortesi: there was an `ensure0x()` function which does that. And i'll argue that's a better name for it... since this works either if you have or you don't have the 0x already

packages/contractkit/src/governance/proposals.ts
```diff
@@ -0,0 +1,105 @@
...
+import { TransactionObject, Tx } from 'web3/eth/types'
+
+import { concurrentMap } from '@celo/utils/lib/async'
+
+import { CeloContract } from '../base'
+import { ContractKit } from '../kit'
+import { bufferToString, CeloTransactionObject, stringToBuffer } from '../wrappers/BaseWrapper'
+import { Proposal, ProposalTransaction } from '../wrappers/Governance'
+
+export interface CeloProposalTransactionJSON {
```
mcortesi: you can skip the `Celo` if you want.. we are already inside Celo/contractKit... but anyway you want works

packages/contractkit/src/governance/proposals.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (parsedTx == null) {
+        throw new Error(`Unable to parse ${transaction} with block explorer`)
+      }
+      return { callDetails: parsedTx.callDetails, value: parsedTx.tx.value }
+    })
+  }
+}
+
+type TxParams = Pick<ProposalTransaction, 'to' | 'value'>
+export class ProposalTransactionFactory {
```
jmrossy: Just as a general design point (but not a blocking issue), this Java style use of Classes without an actual need for classes is not really typical in Javascript. If every method is static, why have a class at all? More common would be to just make functions and export them. @mcortesi WDYT?
Related talk I'd recommend: https://www.youtube.com/watch?v=QyJZzq0v7Z4

packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -22,30 +25,39 @@ export abstract class BaseWrapper<T extends Contract> {
   }
 }
 
-/** Parse string -> BigNumber */
-export function toBigNumber(input: string) {
-  return new BigNumber(input)
-}
+export const numberLikeToBigNumber = (input: NumberLike) => new BigNumber(input)
```
jmrossy: Nit: there's no need for us to define a NumberLike type, `BigNumber.Value` is exactly this

packages/mobile/src/identity/verification.ts
```diff
@@ -48,6 +26,29 @@ import { web3 } from 'src/web3/contracts'
 import { getConnectedAccount, getConnectedUnlockedAccount } from 'src/web3/saga'
 import { privateCommentKeySelector } from 'src/web3/selectors'
```
jmrossy: Why the space here? I recommend letting VSCode organize the imports for you as the linter requires them to be alphabetized

packages/mobile/src/invite/saga.ts
```diff
@@ -43,6 +40,10 @@ import { addLocalAccount, web3 } from 'src/web3/contracts'
 import { getConnectedUnlockedAccount, getOrCreateAccount } from 'src/web3/saga'
 import { zeroSyncSelector } from 'src/web3/selectors'
 
+import { trimLeading0x } from '@celo/utils/src/address'
```
jmrossy: ditto

packages/contractkit/src/explorer/block-explorer.ts
```diff
@@ -93,23 +84,23 @@ export class BlockExplorer {
...
+    const argKeys = Array.from(Array(parameters.__length__).keys())
     delete parameters.__length__
-    Object.keys(parameters).forEach((key) => {
-      if (Number.parseInt(key, 10) >= 0) {
-        delete parameters[key]
-      }
+    const args = argKeys.map((argKey) => {
+      const arg = parameters[argKey]
+      delete parameters[argKey]
+      return arg
```
mcortesi: not sure i follow this logic change... the Array.from(Array....) not sure what it does

packages/contractkit/src/explorer/base.ts
```diff
@@ -10,6 +11,23 @@ export interface ContractDetails {
   jsonInterface: ABIDefinition[]
 }
 
+export interface CallDetails {
+  contract: string
+  function: string
+  parameters: Record<string, any>
+  args: any[]
```
mcortesi: what is args??

packages/contractkit/src/explorer/block-explorer.ts
```diff
@@ -93,23 +84,23 @@ export class BlockExplorer {
...
+    // remove number keys from parameters
+    const argKeys = Array.from(Array(parameters.__length__).keys())
     delete parameters.__length__
-    Object.keys(parameters).forEach((key) => {
-      if (Number.parseInt(key, 10) >= 0) {
-        delete parameters[key]
-      }
+    const args = argKeys.map((argKey) => {
+      const arg = parameters[argKey]
+      delete parameters[argKey]
```
mcortesi: style... we have a map that in it's closure thing modifies another object (parameters....) 
i would try using a forEach

packages/contractkit/src/wrappers/Governance.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    expect(config.concurrentProposals).toEqBigNumber(expConfig.concurrentProposals)
+    expect(config.dequeueFrequency).toEqBigNumber(expConfig.dequeueFrequency)
+    expect(config.minDeposit).toEqBigNumber(minDeposit)
+    expect(config.queueExpiry).toEqBigNumber(expConfig.queueExpiry)
+    expect(config.stageDurations.Approval).toEqBigNumber(expConfig.approvalStageDuration)
+    expect(config.stageDurations.Referendum).toEqBigNumber(expConfig.referendumStageDuration)
+    expect(config.stageDurations.Execution).toEqBigNumber(expConfig.executionStageDuration)
+  })
+
+  describe.only('Proposals', () => {
```
mcortesi: .only()

packages/protocol/migrations/20_elect_validators.ts
```diff
@@ -157,9 +157,9 @@ async function registerValidator(
 }
 
 module.exports = async (_deployer: any, networkName: string) => {
-  if (networkName === 'development') {
-    return
-  }
+  // if (networkName === 'development') {
```
mcortesi: remove?

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -1,21 +1,103 @@
...
+export interface ProposalMetadata {
+  proposer: Address
+  deposit: BigNumber
+  timestamp: BigNumber
+  transactionCount: number
+}
+
+export type ProposalTransaction = Pick<Transaction, 'to' | 'input' | 'value'>
+export type ProposalParams = Parameters<Governance['methods']['propose']>
+export class Proposal {
```
mcortesi: i think it's better to have 
```ts
type Proposal = ProposalTransaction[]
```
and just an utility function `toParams`.. which probably you only need on the wrapper, so it can be just part of it

a Class "feels" to much just to wrap an array

packages/contractkit/src/governance/proposals.ts
```diff
@@ -0,0 +1,96 @@
...
+import { Proposal, ProposalTransaction } from '../wrappers/Governance'
+
+export interface ProposalTransactionJSON {
+  contract: CeloContract
+  function: string
+  args: any[]
+  value: string
+}
+
+export class ProposalUtility extends Proposal {
```
mcortesi: seems weird for an utility class to extend from a Value class... i think composition work better

packages/contractkit/src/governance/proposals.ts
```diff
@@ -0,0 +1,96 @@
...
+) {
+  const to = tx.defaultParams && tx.defaultParams.to ? tx.defaultParams.to : params.to
+  const value = tx.defaultParams && tx.defaultParams.value ? tx.defaultParams.value : params.value
+  if (!to || !value) {
+    throw new Error("Transaction parameters 'to' and/or 'value' not provided")
+  }
+  return proposalTxFromWeb3Txo(tx.txo, { to, value: valueToString(value) })
+}
+
+export async function proposalTxFromJSONAndKit(tx: ProposalTransactionJSON, kit: ContractKit) {
```
mcortesi: style... imo the other approach was better.. but it just a matter of taste...

something like
```ts
const builder = newProposalBuilder(kit)
const proposal = builder
   .addTx(ts, params) //celo is the default
   .addTx(ts2, params2)
   .addWeb3Tx(ts3, params3)
   .build()

console.log(proposal.hash())
console.log(proposal.toJson(kit))
console.log(proposal.rawTransactions)


// to read a proposal from a json.
proposal = Proposal.fromJson(kit, json) // no need to specify that kit is a paramenter in function name


```

packages/contractkit/src/explorer/base.ts
yorhodes: if parameters is a map, args is ordered list

packages/contractkit/src/explorer/block-explorer.ts
```diff
@@ -93,15 +94,14 @@ export class BlockExplorer {
...
     }
 
-    const parameters: Record<string, any> = abi.decodeParameters(
-      matchedAbi.inputs!,
-      encodedParameters
-    )
+    const parameters = abi.decodeParameters(matchedAbi.inputs!, encodedParameters)
 
-    // remove number & number keys
+    const args = new Array(parameters.__length__)
```
nambrot, line 100: Leave a comment about this transformation as it is not particularly obvious

packages/contractkit/src/explorer/block-explorer.ts
```diff
@@ -8,6 +8,7 @@ export interface CallDetails {
   contract: string
   function: string
   parameters: Record<string, any>
+  args: any[]
```
nambrot: What's the difference between parameters and args? The latter being ordered?

packages/contractkit/src/governance/proposals.ts
```diff
@@ -0,0 +1,88 @@
...
+  args: any[]
+  value: string
+}
+
+export class ProposalUtility {
+  constructor(private readonly kit: ContractKit, public readonly proposal: Proposal) {}
+
+  get hash(): Buffer {
+    const paramsEncoded = this.kit.web3.eth.abi.encodeParameters(
+      ['uint256[]', 'address[]', 'bytes', 'uint256[]'],
```
nambrot: This seems like a magical constant that  should be extracted and named

packages/contractkit/src/governance/proposals.ts
```diff
@@ -0,0 +1,88 @@
...
+      GovernanceWrapper.toParams(this.proposal)
+    )
+    return keccak256(paramsEncoded) as Buffer
+  }
+
+  async json(): Promise<ProposalTransactionJSON[]> {
+    const contractDetails = await obtainKitContractDetails(this.kit)
+    const blockExplorer = new BlockExplorer(this.kit, contractDetails)
+
+    return concurrentMap(1, this.proposal, async (transaction) => {
```
nambrot: Why constrain yourself to a single process?

packages/contractkit/src/governance/proposals.ts
```diff
@@ -0,0 +1,88 @@
...
+        function: parsedTx.callDetails.function,
+        args: parsedTx.callDetails.args,
+        value: parsedTx.tx.value,
+      }
+    })
+  }
+}
+
+type ProposalTxParams = Pick<ProposalTransaction, 'to' | 'value'>
+export class ProposalBuilder {
```
nambrot, line 49: I see that these two technically do two different things, but what do you think about merging ProposalUtility and ProposalBuilder?

packages/contractkit/src/wrappers/Governance.test.ts
```diff
@@ -0,0 +1,251 @@
...
+      builder.addWeb3Tx(registry.methods.setAddressFor(...r), {
+        to: registry._address,
+        value: '0',
+      })
+    )
+    return new ProposalUtility(kit, builder.proposal)
+  }
+
+  const verifyRepointResult = (repoints: Repoint[]) =>
+    concurrentMap(1, repoints, async (repoint) => {
```
nambrot: Why constraint concurrency to 1?

packages/contractkit/src/wrappers/Governance.test.ts
```diff
@@ -0,0 +1,251 @@
...
+    const repoints: Repoint[] = [
+      [CeloContract.Random, '0x0000000000000000000000000000000000000001'],
+      [CeloContract.Escrow, '0x0000000000000000000000000000000000000002'],
+    ]
+    const proposalID = new BigNumber(1)
+
+    let proposal: Proposal
+    beforeAll(() => (proposal = registryRepointProposal(repoints).proposal))
+
+    const proposeFn = async (proposer: Address) => {
```
nambrot: Slightly more concise:

```ts
const proposeFn = (proposer: Address) =>
      governance.propose(proposal).sendAndWaitForReceipt({ from: proposer, value: minDeposit })
```

Though I certainly concede that it is a taste question

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -72,4 +148,399 @@ export class GovernanceWrapper extends BaseWrapper<Governance> {
...
+    const dequeue = await this.getDequeue()
+    const index = dequeue.findIndex((d) => d.isEqualTo(proposalID))
+    if (index === -1) {
+      throw new Error(`Proposal ${proposalID} not in dequeue`)
+    }
+    return index
+  }
+
+  // TODO: merge with SortedOracles/Election findLesserAndGreater
+  // proposalID is zero for revokes
```
nambrot: I see thee appeal to use this method for both votes and revokes, but it IMO made for a quite confusing reading. I would recommend for extracting of much of the logic into helper functions and then have separate functions for votes/revokes.

packages/utils/src/address.ts
```diff
@@ -9,49 +9,36 @@ import {
 export type Address = string
 
 export function eqAddress(a: Address, b: Address) {
-  return stripHexLeader(a).toLowerCase() === stripHexLeader(b).toLowerCase()
+  return trimLeading0x(a).toLowerCase() === trimLeading0x(b).toLowerCase()
+}
+
+export function trimLeading0x(input: string) {
```
nambrot: It seems to me that virtually all use cases of `trimLeading0x` are to create a `Buffer`. What do you think about a `hexToBuffer` method?

packages/protocol/contracts/governance/Governance.sol
```diff
@@ -848,21 +869,29 @@ contract Governance is
...
+   * @return Whitelist tally
    */
-  function isHotfixPassing(bytes32 hash) public view returns (bool) {
+  function hotfixWhitelistValidatorTally(bytes32 hash) public view returns (uint256) {
     uint256 tally = 0;
     uint256 n = numberValidatorsInCurrentSet();
     for (uint256 idx = 0; idx < n; idx++) {
       address validator = validatorAddressFromCurrentSet(idx);
-      if (hotfixes[hash].whitelisted[validator]) {
+      if (isHotfixWhitelistedBy(hash, validator)) {
```
nambrot, line 881: This might be off topic and I don't fully remember, but I believe the validator set contains the addresses of the validating keys, but is that the keys we expect validators to whitelist with? cc @asaj

yorhodes: will fix in a followup PR

yorhodes: Opened https://github.com/celo-org/celo-monorepo/issues/1832

yorhodes:
Closing for merge of #2139

PR #1575, opened 2019-11-04
Support claim signatures and support Keybase claims

### LINKED ISSUE #1407: Add address to claim metadata
### Expected Behavior

* Claim metadata should include the `address`.
* Displaying claim metadata should verify this matches the address used to sign the metadata.

### Current Behavior

It doesn't.

### LINKED ISSUE #1408: ContractKit: Create and verify a keybase claim
### Expected Behavior

- [x] CLI function to add a claim on a Keybase username to claims metadata

- [ ] CLI function to verify, given a URL or as part of verifying a claim metadata file, a Keybase claim

nambrot:
### Description

This PR adds the ability to claim Keybase accounts. It also requires claims to be signed by the claiming account, and that is enforced during the parsing of metadata.

A user can claim a keybase account by running:
`celocli account:claim-keybase ./namtest.json --username nambrot --from 0x13897F87A0aB80824fa85600b935dAF579b69157`

If the keybase CLI is installed, we will automatically attempt to upload the proof to the keybase file system, otherwise, we'll print manual instructions such as:

> Proving a keybase claim requires you to publish the signed claim on your Keybase file system to prove ownership. We saved it for you under verify-0x13897F87A0aB80824fa85600b935dAF579b69157.json. It should be hosted in your public folder at .well-known/celo//verify-0x13897F87A0aB80824fa85600b935dAF579b69157.json, so that it is available under https://nambrot.keybase.pub/.well-known/celo/verify-0x13897F87A0aB80824fa85600b935dAF579b69157.json

### Tested

- Ran locally

### Related issues

- Fixes #1407 
- Fixes #1408


packages/cli/src/commands/account/claim-keybase.ts
```diff
@@ -0,0 +1,131 @@
...
+    username: string,
+    address: string
+  ) {
+    const signedClaim = { payload: claim, signature }
+    try {
+      cli.action.start(`Attempting to automate keybase proof`)
+      const publicFolderPrefix = `/keybase/public/${username}/`
+      await this.ensureKeybaseFilePathToProof(publicFolderPrefix)
+      const fileName = proofFileName(address)
+      const tmpPath = `/tmp/${fileName}`
```
mcortesi: use os tmp dir? https://nodejs.org/docs/latest-v10.x/api/os.html#os_os_tmpdir

packages/cli/src/utils/exec.ts
```diff
@@ -0,0 +1,52 @@
+import { spawn, SpawnOptions } from 'child_process'
+
+export function execCmd(
+  cmd: string,
+  args: string[],
+  options?: SpawnOptions & { silent?: boolean }
+) {
+  return new Promise<number>(async (resolve, reject) => {
```
mcortesi: you  don't need the async here

packages/cli/src/utils/exec.ts
```diff
@@ -0,0 +1,52 @@
+import { spawn, SpawnOptions } from 'child_process'
+
+export function execCmd(
```
mcortesi, line 3: at this point... having so many instances of this fn, shouldn't we move it to celo-utils?

packages/contractkit/src/identity/claims/keybase.ts
```diff
@@ -0,0 +1,65 @@
...
+  username: t.string,
+})
+export type KeybaseClaim = t.TypeOf<typeof KeybaseClaimType>
+
+export const keybaseFilePathToProof = `.well-known/celo/`
+export const proofFileName = (address: Address) => `verify-${address}.json`
+export const targetURL = (username: string, address: Address) =>
+  `https://${username}.keybase.pub/${keybaseFilePathToProof}${proofFileName(address)}`
+
+export async function verifyKeybaseClaim(claim: KeybaseClaim, signer: Address) {
```
mcortesi: output type seems weird: `error | void` ? maybe `error | true`?

packages/utils/src/address.ts
```diff
@@ -11,3 +11,5 @@ export const privateKeyToAddress = (privateKey: string) => {
     '0x' + privateToAddress(Buffer.from(privateKey.slice(2), 'hex')).toString('hex')
   )
 }
+
+export { toChecksumAddress } from 'ethereumjs-util'
```
mcortesi, line 15: i don't see ethereumsjs-utils in utils pkg.json...

packages/contractkit/src/identity/claims/keybase.ts
mcortesi: also, add a comment specifying the output format interpretation

packages/cli/src/utils/cli.ts
```diff
@@ -59,3 +59,8 @@ export function failWith(msg: string): never {
   console.error(msg)
   return process.exit(1)
 }
+
+export async function binaryPrompt(promptMessage: string) {
+  const resp = await cli.prompt(promptMessage)
```
mcortesi: maybe add a fixed "[y/yes, n/no]" as a prompt suffix?

mcortesi: i don't if oclif has it, but a good library for this kind of things is `inquirer.js`

mcortesi (approved):
lefts some comments, but in general everything is good.

Maybe we can add tests to the cli commands, they don't seem to be as "small" as usual.
For testing cli commands, i've added support for them in another pr which is about to be merged. Basically you can run a command using ganache/devchain same as contractkit

packages/cli/src/utils/exec.ts
nambrot: Not sure what I was thinking

packages/cli/src/utils/exec.ts
nambrot: We should, but I'm not sure I want to be the person transforming the codebase, nor prescribe which version should be the one to rule them all

packages/contractkit/src/identity/claims/keybase.ts
nambrot: You know how much I struggle with how to express unsuccessful computation. I wanted to avoid throwing and immediately catching errors, so I opted for the soft version of that which is that a successful completion yields no/undefined return. Made the return type explicit (string | undefined)

packages/utils/src/address.ts
nambrot: https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/package.json#L25 ?

mcortesi: sorry i was looking at the wrong package.json

packages/cli/src/commands/account/claims.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    const generateEmptyMetadataFile = () => {
+      writeFileSync(emptyFilePath, IdentityMetadataWrapper.fromEmpty(account))
+    }
+
+    const readFile = () => {
+      return IdentityMetadataWrapper.fromFile(emptyFilePath)
+    }
+
+    test('account:create-metadata cmd', async () => {
+      const newFilePath = `${tmpdir()}/newfile.json`
```
mcortesi, line 32: don't you need to remove the file if already present?

packages/cli/src/commands/account/claims.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      await CreateMetadata.run(['--from', account, newFilePath])
+      const res = JSON.parse(readFileSync(newFilePath).toString())
```
mcortesi, line 34: something that you can do is use jest [snapshot testing](https://jestjs.io/docs/en/snapshot-testing) which is a nice way of doing a quick test that asserts the content remains the same over different runs

mcortesi (approved):
great tests!! ;)

packages/cli/src/commands/account/claims.test.ts
nambrot: Yeah, but I'm just assuming that it does not exist :)

packages/cli/src/commands/account/claims.test.ts
nambrot: I'm gonna leave that for a future PR :)

mcortesi:
But the second time you run the test it will

PR #1578, opened 2019-11-04
Optimized Attestation view calls and removal of the reveal TX

### LINKED ISSUE #1491: Developers should be able to use actionable attestations
### Expected Behavior

With the move to moving the reveal off-chain, we now need to fetch the attestation service URL from the metadata registered on the Accounts contract instead.

### Current Behavior

On the contract wrapper, we use `getActionableAttestations` to give consumers a single interface through at the end which they have all the relevant information to make progress on their attestations. It currently contains the data encryption key for the `reveal` transaction.

### LINKED ISSUE #1545: Users should not be able to reference a reveal tx
### Expected Behavior

Once we moved completely to the attestation-service, we should remove the reveal transaction

### Current Behavior

We have a `reveal` tx

nambrot:
### Description

This PR tries to reduce the number of view calls a client has to make to receive attestations. It does so by introducing two view calls:

`Attestations#getCompletableAttestationStates`
It directly returns the attestation issuers for which a user can complete attestations for (instead of having to iterate over all selected issuers themselves)

`Accounts#batchGetMetadataURL`
Fetches the metdata URLs for a set of accounts (instead of having to do N view calls)

This PR also removes any mention of the `reveal` transaction 

### Tested

- [x] on `namoffchainreveal`
- [x] unit tests

### Related issues

- Fixes #1491 
- Fixes #1545


packages/utils/src/collections.ts
```diff
@@ -18,6 +18,16 @@ export function zip3<A, B, C>(as: A[], bs: B[], cs: C[]) {
   return res
 }
 
+export function compact<A>(as: Array<A | null>): A[] {
```
jmrossy: Seems like this could be done with `.filter`?

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -179,50 +176,50 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
+          // TODO: Once we have status indicators, we should check if service is up
+          // https://github.com/celo-org/celo-monorepo/issues/1586
+          return {
+            blockNumber,
+            issuer,
+            attestationServiceURL: attestationServiceURLClaim.url,
+          }
+        } catch (error) {
+          console.error(error)
+          return null
```
jmrossy, line 241: Consider allowing the error to propagate

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -350,35 +347,23 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
-        Buffer.from(publicKey.slice(2), 'hex'),
-        Buffer.from(phoneNumber, 'utf8')
-      ).toString('hex')
-
-    return toTransactionObject(
-      this.kit,
-      this.contract.methods.reveal(
-        PhoneNumberUtils.getPhoneHash(phoneNumber),
-        encryptedPhone,
+  async requestAttestationFromIssuer(
```
jmrossy: Maybe `reveal` is a better name here? or something else. I find this is confusing given that requesting an attestation is already a step in the verification process

asaj:
Do you have any information on whether or not this actually makes things faster?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -484,6 +436,48 @@ contract Attestations is
...
+    */
+  function getCompletableAttestations(bytes32 identifier, address account)
+    external
+    view
+    returns (uint32[] memory, address[] memory, uint[] memory, bytes memory)
+  {
+    AttestedAddress storage state = identifiers[identifier].attestations[account];
+    address[] storage issuers = state.selectedIssuers;
+
+    uint num = 0;
```
asaj: uint256 everywhere

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -484,6 +436,48 @@ contract Attestations is
 
   }
 
+  /**
+    * @notice Returns the state of all attestations that are completable
+    * @param identifier Hash of the identifier.
+    * @param account Address of the account.
+    * @return ( blockNumbers[] - Block number of request/completion the attestation,
+    *           issuers[] - Address of the issuer,
+    *
```
asaj: formatting

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -238,46 +230,6 @@ contract Attestations is
     delete state.unselectedRequests[msg.sender];
   }
 
-  /**
```
asaj, line 225: Document this in the PR desrcriptinn. If anything, maybe this should be the title of the PR

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -228,6 +228,36 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
+   * @return (stringLengths[] - the length of each string in bytes
+              data - all strings concatenated
+             )
+   */
+  function batchGetMetadataURL(address[] calldata accountsToQuery)
+    external
+    view
+    returns (uint[] memory, bytes memory)
+  {
+    uint totalSize = 0;
```
asaj: uint256

asaj:
Do you have any information on whether or not this actually makes things faster?

Also please be sure to follow the style guide.

packages/utils/src/collections.ts
nambrot: Great point, originally filter didn't properly get rid of the type, but i can use the custom defined type guard below

packages/contractkit/src/wrappers/Attestations.ts
nambrot: I was thinking about it, but it seemed to me that a validator not or incorrectly registering an attestation service url should not result in an error for a client who just wants to complete attestations. Thus, I favored returning the attestations that are completable, by intent of the function.

packages/contractkit/src/wrappers/Attestations.ts
nambrot: I'm always open to naming advice, I opted for `requestFromIssuer` because that's what it represents imo (vs. requesting from the smart contract). `reveal` made more sense when it was on-chain as we encrypted the number to the validator, but now we are explicitly requesting from a validator's service

nambrot:
We have no concrete information on whether this makes things actually faster or not. However theoretically, we move from 2*N + 1 view calls to 1 view call (where N = 3). It might be a negligent number once we improve light client performance, but in the meantime this seems like an easy win.

My bad on `uint` vs `uint256`, it's been a while since I've been allowed to use all 256 bits :)

packages/attestation-service/config/config.json
```diff
@@ -0,0 +1,21 @@
+{
```
asaj, line 1: Didn't review this

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -179,50 +176,50 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
   }
 
   /**
-   * Returns an array of attestations that can be completed, along with the issuers public key
+   * Returns an array of attestations that can be completed, along with the issuers attestation service url
```
asaj: nit: issuer's

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -179,50 +176,50 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
-    const issuers = await this.contract.methods.getAttestationIssuers(phoneHash, account).call()
-    const issuerState = Promise.all(
-      issuers.map((issuer) =>
-        this.contract.methods
-          .getAttestationState(phoneHash, account, issuer)
-          .call()
-          .then(parseAttestationInfo)
-      )
+
+    const result = await this.contract.methods.getCompletableAttestations(phoneHash, account).call()
```
asaj, line 218: Please define a helper method that parses this into a more readable object

packages/contractkit/src/wrappers/Attestations.ts
asaj: +1 to two "request"s being confusing. In some sense, we are revealing our phone number here to the issuer. Alternatively, maybe the original request is more of a commitment to request attestations?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -238,46 +230,6 @@ contract Attestations is
...
-   * @notice Reveal the encrypted phone number to the issuer.
-   * @param identifier The hash of the identifier to be attested.
-   * @param encryptedPhone The number ECIES encrypted with the issuer's public key.
-   * @param issuer The issuer of the attestation.
-   * @param sendSms Whether or not to send an SMS. For testing purposes.
-   */
-  function reveal(
```
asaj: 🕺

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -228,6 +228,36 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
+  function batchGetMetadataURL(address[] calldata accountsToQuery)
+    external
+    view
+    returns (uint256[] memory, bytes memory)
+  {
+    uint256 totalSize = 0;
+    uint256[] memory sizes = new uint256[](accountsToQuery.length);
+    for (uint256 i = 0; i < accountsToQuery.length; i++) {
+      sizes[i] = bytes(accounts[accountsToQuery[i]].metadataURL).length;
+      totalSize += sizes[i];
```
asaj: safemath everywhere

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -484,6 +436,49 @@ contract Attestations is
...
+    * @notice Returns the state of all attestations that are completable
+    * @param identifier Hash of the identifier.
+    * @param account Address of the account.
+    * @return ( blockNumbers[] - Block number of request/completion the attestation,
+    *           issuers[] - Address of the issuer,
+    *           stringLengths[] - The length of each metadataURL string for each issuer,
+    *           stringData - All strings concatenated
+    *         )
+    */
+  function getCompletableAttestations(bytes32 identifier, address account)
```
asaj, line 417: style guide violation

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -228,6 +228,36 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
   }
 
+  /**
+   * @notice Getter for the metadata of multiple accounts.
+   * @param accountsToQuery The addresses of the accounts to get the metadata for.
+   * @return (stringLengths[] - the length of each string in bytes
+   *          data - all strings concatenated
+   *         )
+   */
+  function batchGetMetadataURL(address[] calldata accountsToQuery)
```
asaj, line 209: style guide violation

packages/protocol/contracts/common/interfaces/IAccounts.sol
```diff
@@ -21,5 +21,9 @@ interface IAccounts {
   function getDataEncryptionKey(address) external view returns (bytes memory);
   function getWalletAddress(address) external view returns (address);
   function getMetadataURL(address) external view returns (string memory);
+  function batchGetMetadataURL(address[] calldata)
```
asaj, line 23: Doesn't fit on one line?

nambrot: Doesn't look like it does?

packages/protocol/contracts/identity/Attestations.sol
nambrot: Talked offline, looks like both ways are acceptable

packages/protocol/contracts/common/Accounts.sol
nambrot: Talked offline, looks like both ways are acceptable

packages/contractkit/src/wrappers/Attestations.ts
nambrot: I like commit a lot, but many related structs and events do make more sense as `request` for that transaction. Here are the steps currently:

- request - "commits" to a number of attestations, we have that commitment as an "unselected request" right now
- selectIssuers - "selects" on an "unselected request", and persists the actual attestations with their corresponding issuers
- requestFromIssuer - advances an attestation by "revealing" the phone number to an issuer, prompting an issuer to send the SMS
- complete - completes the attestation by posting the signed message from the SMS.

I still don't quite like `reveal` but given my lack of alternative suggestions, I'll move it to reveal

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -1,29 +1,26 @@
 pragma solidity ^0.5.3;
 
-import "openzeppelin-solidity/contracts/utils/ReentrancyGuard.sol";
-import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import 'openzeppelin-solidity/contracts/utils/ReentrancyGuard.sol';
```
asaj: I thought we went with double quotes?

asaj (approved):

packages/protocol/contracts/common/Accounts.sol
nambrot: Fixed with an upstream merge

jmrossy (approved):

Issue #1319, opened 2019-10-12
Wallet should support LRD


nityas:
### Expected Behavior

Support for Liberian Dollars

### Current Behavior

No support for Liberian Dollars

PR #1581, opened 2019-11-05
[BlockchainApi] Use new third party currency api 

jeanregisser:
### Description

Use a new currency API (https://currencylayer.com) which supports LRD (Liberian Dollar) among many others.

### Tested

Updated and ran tests

### Other changes

- Changed caching policy
  - cache rates for the current day for 12 hours
  - cache rates for past days indefinitely (they should never change)

### Related issues

- #1319

### Backwards compatibility

Yes


packages/blockchain-api/src/config.ts
```diff
@@ -1,9 +1,12 @@
 import dotenv from 'dotenv'
+import * as secretsFile from './secrets.json'
 
 // Load environment variables from .env file
 dotenv.config()
 
+export const DEPLOY_ENV = (process.env.DEPLOY_ENV as string).toLowerCase()
 export const EXCHANGE_RATES_API = (process.env.EXCHANGE_RATES_API as string).toLowerCase()
+export const { EXCHANGE_RATES_API_ACCESS_KEY } = (secretsFile as any)[DEPLOY_ENV]
```
jmrossy: Don't we need a JSON.Parse here?

jmrossy (approved):
Once question below but otherwise the changes look good!

packages/blockchain-api/src/config.ts
jeanregisser: TypeScript does it for us because we're using the `resolveJsonModule` flag config.

Issue #540, opened 2019-08-14
Developers SBAT use *Contract as opposed to Truffle.Contract<*Instance> type in tests


timmoreton:
### Expected Behavior

We use the `*Contract` types generated by typechain everywhere in our tests

### Current Behavior

We use the `Truffle.Contract<*Instance>` type instead

jeamick:
Hello @timmoreton ,

Could you please give details about which test are you talking about ?
I would be interested to look a bit at this issue if possible.

Thanks
JM

timmoreton:
@jeamick Hey! @asaj can give you some more context

jeamick:
hey @timmoreton  and @asaj ,
By a quick research I found some examples of what you are describing in this part of the repository:

https://github.com/celo-org/celo-monorepo/tree/master/packages/protocol

are you talking about this ?

By the way is there a reason why it was like this ?
Convenience?

Thanks
JM

asaj:
Happy to provide more context!

This is really just a readability issue, and not a functional difference. Typechain creates contract specific types for us, so we might as well use them, but the types should be functionally the same, just with a more readable name.

PR #1585, opened 2019-11-05
Adjust e2e transfer and governance tests to match new fee distribution and eliminate ProposerFraction

nategraf:
### Description

Fee distribution updated in https://github.com/celo-org/celo-blockchain/pull/563 requires updates to the e2e transfer test. This PR provides those changes.

### Tested

Tested with e2e and existing unit tests.

### Other changes

* Only restart the geth nodes between different sync options to avoid a fragile test condition in which we only test transfers from a fresh state.
* Shorten the delay on waiting for the sync node to update.
* Check gas usage on test transfers, because the tests may erroneous pass or fail a test condition if the gas usage changes without the test being updated.

### Related issues

- Related to #540

### Backwards compatibility

Introduces a breaking change to contract kit by removing proposer fraction. Test changes are compatible with blockchain clients which include https://github.com/celo-org/celo-blockchain/pull/563

Depends on https://github.com/celo-org/celo-blockchain/pull/563


nategraf:
Marking `do not merge` until celo-org/celo-blockchain#563 is merged

.circleci/config.yml
```diff
@@ -522,7 +522,7 @@ jobs:
           command: |
             set -e
             cd packages/celotool
-            ./ci_test_transfers.sh checkout master
+            ./ci_test_transfers.sh checkout victor/gpm
```
asaj: I'm still not convinced it's better to repoint both PRs instead of just repointing one.

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -360,6 +367,13 @@ describe('governance tests', () => {
         assert.equal(expected.toFixed(), currentVotes.minus(previousVotes).toFixed())
       }
 
+      // Returns the gas fee base for a given block, which is distributed to the governance contract.
+      const gasFeeBase = async (blockNumber: number): Promise<BigNumber> => {
```
asaj: nit: consider `blockBaseGasFee`

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -72,19 +65,6 @@ class InflationManager {
       .setInflationParameters(toFixed(rate).toString(), updatePeriod)
       .sendAndWaitForReceipt({ from: this.validatorAddress })
   }
-
-  resetInflation = async () => {
```
asaj, line 76: Was this not used, or..?

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -207,6 +182,14 @@ describe('Transfer tests', function(this: any) {
     }
     await initAndStartGeth(hooks.gethBinaryPath, fullInstance)
 
+    // Install an arbitrary address as the goverance address to act as the infrastructure fund.
+    // This is chosen instead of full migration for spead and to avoid the need for a governance
```
asaj: speed

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -442,8 +443,8 @@ describe('Transfer tests', function(this: any) {
         before(`start geth on sync: ${syncMode}`, () => startSyncNode(syncMode))
 
         describe('Transfer CeloGold >', () => {
-          const GOLD_TRANSACTION_GAS_COST = 29180
-          describe('gasCurrency = CeloGold >', () => {
+          const GOLD_TRANSACTION_GAS_COST = 30005
```
asaj, line 466: Why did this change?

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -542,7 +543,9 @@ describe('Transfer tests', function(this: any) {
         describe('Transfer CeloDollars', () => {
           describe('gasCurrency = CeloDollars >', () => {
             testTransferToken({
-              expectedGas: 189456,
+              // TODO(nategraf): Initializing a stable token account costs 5k additional gas.
+              // The first time this transfer is done, which is for the full node, this will be seen.
+              expectedGas: syncMode === 'full' ? 190303 : 175303,
```
asaj: 5k or 15k?

asaj: Can we restart with clean nodes so we don't have this funky check?

asaj (approved):
Looks great! Mostly nits

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: It wasn't used.

.circleci/config.yml
nategraf: My intention is to merge celo-org/celo-blockchain#563, point this back to master, wait for CI to pass, then merge this one. I simply want to see that the e2e tests pass (or fail) on this PR while the other is still pending.

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: It was out of date and being ignored. I adjusted each of the gas costs based on master then added a step step to enforce that they are accurate.

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: Restarting with clean nodes takes longer and reduces test coverage because it only considers transfers to fresh accounts. I'll go ahead and split this off into it's own test case though.

packages/celotool/src/e2e-tests/governance_tests.ts
nategraf: Done

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: Done

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: I should really leave math to the computers

PR #1607, opened 2019-11-06
Deployment fixes for attestation-service

nambrot:
### Description

This PR includes various fixes to make the attestation-service actually operational. Before, I mostly tested that the service would run, but only with #1578 I was able to actually test it end-2-end. Specifically, we know have a helm-hosted postgres database all attestation service instances share. I could not make a dynamic password work, so for now, it is hardcoded to `password`, which should be fine for our testnets as the postgres service is a `ClusterIP` service.

### Tested

- [x] on `namoffchainreveal`


packages/celotool/src/lib/attestation-service.ts
```diff
@@ -43,6 +44,10 @@ function helmParameters() {
     )}"`,
     `--set geth.validators="${fetchEnv(envVar.VALIDATORS)}"`,
     `--set domain.name=${fetchEnv(envVar.CLUSTER_DOMAIN_NAME)}`,
+    `--set global.postgresql.postgresqlDatabase=AttestationService`,
```
jmrossy, line 47: Is a shared global db the long term plan or a stop gap?

packages/celotool/src/lib/attestation-service.ts
```diff
@@ -31,6 +31,7 @@ function helmParameters() {
     `--set domain.name=${fetchEnv(envVar.CLUSTER_DOMAIN_NAME)}`,
     `--set celotool.image.repository=${fetchEnv('CELOTOOL_DOCKER_IMAGE_REPOSITORY')}`,
     `--set celotool.image.tag=${fetchEnv('CELOTOOL_DOCKER_IMAGE_TAG')}`,
+    `--set mnemonic="${fetchEnv('MNEMONIC')}"`,
```
tkporter: nit: `envVar.MNEMONIC`

tkporter (approved):

packages/celotool/src/lib/attestation-service.ts
nambrot: I'd say since this is only used for our testnets it is an acceptable solution for now? We currently use it only for a mutex around attestations so I think it's fine for now? Will change if requirements do

PR #1616, opened 2019-11-06
[BlockchainApi] Hotfix secrets not present when deploying to App Engine

jeanregisser:
### Description

Hotfix following #1581

Secrets were missing once deployed to App Engine.

This is because dynamically required files don't get copied to the dist folder.
So here we use a static import and create a dummy secrets.json for environments that are not trusted.

### Tested

Ran the tests with and without being able to decrypt `secrets.json.enc`
Ran locally.
Deployed to integration successfully.

### Other changes

N/A

### Backwards compatibility

Yes

### DIFF
--- a/packages/blockchain-api/package.json
+++ b/packages/blockchain-api/package.json
@@ -14,7 +14,8 @@
     "build": "tsc -p .",
     "clean": "rm -rf dist coverage",
     "gcp-build": "npm run build",
-    "deploy": "./deploy.sh"
+    "deploy": "./deploy.sh",
+    "postinstall": "test -f src/secrets.json || echo \"{}\" > src/secrets.json"
   },
   "dependencies": {
     "@celo/contractkit": "0.1.6",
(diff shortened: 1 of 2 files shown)


jmrossy (approved):

packages/blockchain-api/package.json
```diff
@@ -14,7 +14,8 @@
     "build": "tsc -p .",
     "clean": "rm -rf dist coverage",
     "gcp-build": "npm run build",
-    "deploy": "./deploy.sh"
+    "deploy": "./deploy.sh",
+    "postinstall": "test -f src/secrets.json || echo \"{}\" > src/secrets.json"
```
martinvol, line 18: The way to do this is contemplated here: https://github.com/celo-org/celo-monorepo/blob/master/scripts/create_key_templates.sh (although it's not well documented)

jeanregisser: Great thanks, gonna change it to this then!

martinvol: 🙌

PR #1621, opened 2019-11-06
Decentralized secret templates

jeanregisser:
### Description

Following @martinvol's [comment](https://github.com/celo-org/celo-monorepo/pull/1616#discussion_r343225088), I wanted to use the existing template mechanism to place the `secrets.json` for blockchain-api.
I noticed however that all templates were in the same folder and `secrets.json` was already used for the mobile project.

I took this as an opportunity to decentralize this a little bit by using a filename convention.
Whenever you need a template for some encrypted file listed in `scripts/key_placer.sh`, add it next to the `xxx.enc` file.

For instance if you have `packages/my_project/secrets.json` listed in `scripts/key_placer.sh`, you can create `packages/my_project/secrets.json.template` with the public template content you want and it will be automatically placed if you don't have access to decryption keys.

Let me know what you think.

### Tested

Ran `yarn keys:decrypt` with and without having access to our keyring.
Templates are correctly used when I don't have access to our keyring.

### Other changes

N/A

### Backwards compatibility

Yes


package.json
```diff
@@ -20,7 +20,6 @@
     "report-coverage": "yarn run lerna run test-coverage",
     "test:watch": "node node_modules/jest/bin/jest.js --watch",
     "postinstall": "yarn run lerna run postinstall && patch-package && yarn keys:decrypt",
-    "preinstall": "bash scripts/create_key_templates.sh",
```
martinvol, line 23: Not really sure if this is a good practice, but the reason this is in the preinstall is in case some dependency needs this for yarning or build. In case what I just said doesn't make any sense, then just 🚢

martinvol (approved):

jmrossy (approved):

package.json
jeanregisser: It totally makes sense and I was also concerned by this when I made the change.
I think it's ok for now and we can revisit if it causes an issue later.

PR #1622, opened 2019-11-06
Prettify Solidity files

nambrot:
### Description

As a constant violator of the Solidity Style Guide (@asaj can attest to this fact), I told myself: "No more!". Hence this PR adds `prettier-solidity` which will prettify our `.sol` files so that future generations shall no longer need to live in fear of style guide violations!

It was actually incredibly easy, there was only 1 place where I needed to add a `// prettier-ignore`

### Tested

- CI

### Backwards compatibility

- I believe they should be

### DIFF
--- a/packages/protocol/contracts/common/linkedlists/AddressSortedLinkedListWithMedian.sol
+++ b/packages/protocol/contracts/common/linkedlists/AddressSortedLinkedListWithMedian.sol
@@ -1,15 +1,13 @@
 pragma solidity ^0.5.3;
 
-import "openzeppelin-solidity/contracts/math/SafeMath.sol";
-import "./AddressLinkedList.sol";
-import "./SortedLinkedListWithMedian.sol";
-
+import 'openzeppelin-solidity/contracts/math/SafeMath.sol';
+import './AddressLinkedList.sol';
+import './SortedLinkedListWithMedian.sol';
 
 /**
  * @title Maintains a sorted list of unsigned ints keyed by address.
  */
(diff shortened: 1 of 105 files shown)


asaj (commented):
❤️ 

Can we do some configuration in prettierrc? For example, double instead of single quotes? You can see an example of overrides for solidity here: https://github.com/prettier-solidity/prettier-plugin-solidity

packages/protocol/contracts/common/linkedlists/AddressSortedLinkedListWithMedian.sol
```diff
@@ -155,16 +139,15 @@ library AddressSortedLinkedListWithMedian {
...
-  )
+  function getElements(SortedLinkedListWithMedian.List storage list)
     public
     view
     returns (address[] memory, uint256[] memory, SortedLinkedListWithMedian.MedianRelation[] memory)
   {
     bytes32[] memory byteKeys = list.getKeys();
     address[] memory keys = new address[](byteKeys.length);
     uint256[] memory values = new uint256[](byteKeys.length);
+    // prettier-ignore
```
asaj, line 150: Why do we need this?

nambrot: Prettier was unable to produce the line below, instead trying to put everything on a single line except `keys.length` on the next line, which then violated out max-line-length on our lint, so this seemed like the best trade-off

nambrot:
Screwed up with my git-fu, reopened at #1623

PR #1623, opened 2019-11-06
Prettify Solidity files

nambrot:
### Description

As a constant violator of the Solidity Style Guide (@asaj can attest to this fact), I told myself: "No more!". Hence this PR adds `prettier-solidity` which will prettify our `.sol` files so that future generations shall no longer need to live in fear of style guide violations!

It was actually incredibly easy, there was only 1 place where I needed to add a `// prettier-ignore`

### Tested

- CI

### Backwards compatibility

- I believe they should be


nambrot:
@asaj I looked at the linked override at https://github.com/prettier-solidity/prettier-plugin-solidity#vscode, I wasn't sure if that's what we wanted, for example a width of 80 (instead of the 100) that we have been using. Both numbers seemed appropriate, so I kept 100 as it was less churn. Same with tabs/spaces, just kept the default

asaj:
> @asaj I looked at the linked override at https://github.com/prettier-solidity/prettier-plugin-solidity#vscode, I wasn't sure if that's what we wanted, for example a width of 80 (instead of the 100) that we have been using. Both numbers seemed appropriate, so I kept 100 as it was less churn. Same with tabs/spaces, just kept the default

+1 to sticking with our defaults here

asaj (approved):

PR #1624, opened 2019-11-07
Allow validator groups to update their commission

### LINKED ISSUE #1374: Users should be able to set and query the commission rate on groups
### Expected Behavior

Add function to set commission for a group on `Validators.sol`

Users should be able to query the group commission rate and, if they are the owner of the group, set the commission rate through the CLI.

### Current Behavior

Users cannot  query or set commission rates.

asaj:
### Description

This PR adds a function to allow validator groups to update their commission, as well as a corresponding command in celocli.

### Tested

Contract and contractkit unit tests.

### Other changes
- Remove out of date CLI example

### Related issues

- Fixes #1374 

### Backwards compatibility

Technically yes, though the smart contract was changed.


packages/contractkit/src/wrappers/Validators.test.ts
```diff
@@ -86,6 +86,16 @@ testWithGanache('Validators Wrapper', (web3) => {
...
   })
 
+  test('SBAT updateCommission', async () => {
+    const groupAccount = accounts[0]
+    await setupGroup(groupAccount)
+    await (await validators.updateCommission(new BigNumber(0.2))).sendAndWaitForReceipt({
+      from: groupAccount,
+    })
+    const commission = (await validators.getValidatorGroup(groupAccount)).commission
+    expect(commission.toFixed()).toEqual('0.2')
```
nategraf: Could you add an `expect` for commission either in this test or the `setupGroup` function to ensure that the is what you expect before the change (i.e. not `0.2`)

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -651,6 +652,23 @@ contract Validators is
     return true;
   }
 
+  /**
+   * @notice Updates a validator group's commission.
+   * @param commission Fixidity representation of the commission this group receives on epoch
+   *   payments made to its members.
```
nategraf: I note about the range may be helpful.
```suggestion
   *   payments made to its members. In the range 0 to 1 inclusive.
```

nategraf (approved):
Nice!

packages/cli/src/commands/validatorgroup/commission.ts
```diff
@@ -0,0 +1,28 @@
...
+  static examples = [
+    'commission --from 0x47e172F6CfB6c7D01C1574fa3E2Be7CC73269D95 --commission 0.1',
+  ]
+
+  async run() {
+    const res = this.parse(ValidatorGroupCommission)
+
+    this.kit.defaultAccount = res.flags.from
+    const validators = await this.kit.contracts.getValidators()
+    const tx = await validators.updateCommission(new BigNumber(res.flags.commission))
```
mcortesi: can you add "checks" ?

packages/contractkit/src/wrappers/Validators.test.ts
mcortesi: there's a  `expect().toEqBigNumber` in contractKit that you can use (no need to import or anything)

mcortesi (approved):

PR #1629, opened 2019-11-07
[Wallet] Set default gradle properties

jeanregisser:
### Description

Set default gradle properties (JVM memory and enabling parallel).
This is so the app compiles fine even if the developer didn't set `GRADLE_OPTS` in their environment.

### Tested

App compiles fine.

### Other changes

N/A

### Backwards compatibility

Yes.


cmcewen:
Will this overwrite the values we set in CI? circle has different limits so just want to make sure it'll still work okay

martinvol (commented):
Maybe you can delete the references to this env variable in the CircleCI testes.
Here: https://github.com/celo-org/celo-monorepo/blob/a5cc546a4f2c8b50711d4db9d2ff8c43cac6dbe8/.circleci/config.yml#L16
and here: https://github.com/celo-org/celo-monorepo/blob/a5cc546a4f2c8b50711d4db9d2ff8c43cac6dbe8/.circleci/config.yml#L32

jeanregisser:
@cmcewen apparently gradle.properties have precedence over `GRADLE_OPTS` according to https://docs.gradle.org/current/userguide/build_environment.html

jmrossy (approved):

PR #1631, opened 2019-11-08
Sentry infrastructure

### LINKED ISSUE #1279: Infrastructure SBAT support for sentries
Needed for kubernetes and VM deployments.

tkporter:
### Description

Infrastructure setup for sentries!

The general setup is:
* 1 sentry per proxied validator
* If there are N validators and M (where 0 <= M <= N) of those are to be proxied, the validators with indices 0 through M - 1 will be proxied. M is specified as `PROXIED_VALIDATORS` in a .env file, if it's not specified M = 0 is assumed.
* A proxied validator specifies which sentry to use by using the cmd line flag `--instanbul.sentries=[internal enode];[external enode]`
* Sentries have their own account type

#### For Kubernetes:

Sentries are pretty much the same as a normal tx node. I created a template that is used by both tx nodes and sentries, and I also created a service template that is used now by tx nodes, sentries, and validators.

An internal service (ClusterIP) is created for each sentry node that allows TCP/UDP traffic on port 30503 (the default internal sentry port). The sentry service internal IP address is used by a validator when communicating with the sentry because a sentry pod can be deleted & given a new internal IP.

If the testnet is internal (ie STATIC_IPS_FOR_GETH_NODES=false), the IP address of the internal sentry service is also used as the sentry’s IP address for its “external” interface. If testnet is external, the TCP / UDP load balancers are created the same as before, with the sentry using the load balancer IP address for its “external” interface. The external IP address the proxy uses is the one that celotool has allocated for the validator.

#### For VM testnets:

Validators that are proxied do not have an access config, which is what assigns a VM an external IP address. One is created & then destroyed temporarily during the 1 time setup in the startup script so that things can be downloaded from the external internet. These internal validators are on their own subnet that can only communicate with GCP services (so monitoring & logging can continue as expected). Specifying an IP range is required when creating a subnet, and I don’t think there’s a straightforward way to dynamically choose this IP range, so it’s just hardcoded as `10.25.0.0/24`. This IP range can’t conflict with an existing subnet, so this means only 1 VM testnet can really be used on a particular VPC (which feels fine because we already have separate VPCs for VM testnets).

Similar to what I did for helm charts, sentries and tx-nodes now use the same full node code. Sentries use port 30503 for their internal interface & can only send/receive traffic on that port internal to the VPC.

### Tested

Ran an internal kubernetes testnet, an external kubernetes testnet, and a VM-based testnet. Made sure that port-forwarding & contract deployment still works when the VM validator 0 does not have an external IP address. Manually restarted VM docker containers to make sure they're robust enough to handle a restart without breaking things.

### Other changes

I noticed that the docker containers weren’t robust enough to account for restarting, so I fixed that. Now if a container crashes, it’ll be auto-restarted & everything should continue perfectly.

I also removed references to the “og” setup in the helm charts, and did some small celotool refactors.

### Related issues

- Fixes #1279 

### Backwards compatibility

Everything should be backwards compatible!


timmoreton (approved):
Mega! nice simplifications.

aaitor:
This has been tested and deployed in [Berlintestnet002](https://berlintestnet002-ethstats.celo-networks-dev.org/). With the tests made by @jcortejoso and I everything looks been working nicely.
We are going to continue doing some additional tests in the following days and make a final review as soon as the celo-blockchain and celo-monorepo PRs are merged in master.
Nice work @tkporter !

jcortejoso:
Nice PR @tkporter. @aaitor and me were testing today in a new network to understand better the new setup and were able to do nicely. We saw a small issue if you deploy the network initially without proxied nodes (no `PROXIED_VALIDATORS` variable), and the adding them and trying to `celotool deploy upgrade testnet`, giving this error:

```{ Error: Command failed: kubectl scale StatefulSet berlintestnet002-sentry --replicas=0 --namespace berlintestnet002
Error from server (NotFound): statefulsets.apps "berlintestnet002-sentry" not found

    at ChildProcess.exithandler (child_process.js:294:12)
    at ChildProcess.emit (events.js:198:13)
    at ChildProcess.EventEmitter.emit (domain.js:448:20)
    at maybeClose (internal/child_process.js:982:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:259:5)
  killed: false,
  code: 1,
  signal: null,
  cmd:
   'kubectl scale StatefulSet berlintestnet002-sentry --replicas=0 --namespace berlintestnet002' }
error Command failed with exit code 1.
```

If it is supposed to support adding proxied validator after initial deployment I can test and patch the PR to allow it. If it nos a supported usercase and it is not something we expect I guess it is ok.

Kudos for the PR!

tkporter:
@jcortejoso Thanks! Was that testnet deployed initially with this branch? If it was deployed initially without this branch that makes sense because the resource doesn't exist yet (which is another oversight, it's probably a good idea to allow scaling down to fail so that existing testnets can just be `upgrade --reset` without having to destroy and then initial)

If it was deployed initially with the branch then that's definitely an issue, I was hoping it'd default to 0 replicas but still create the statefulset resource

jcortejoso:
Hi @tkporter. To reproduce I think we have to create the testnet from another branch (thus statefulset `${tesnet}-sentry` wont exist), and then upgrade the testnet package with option `--reset`, and the conflicting line would be `https://github.com/celo-org/celo-monorepo/blob/010583d335f28773f6c825a0e40ad6ab77057fd0/packages/celotool/src/lib/helm_deploy.ts#L658` in that situation. 

As I seen you have just fixed it allowing fail. Nice!

packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -445,15 +445,13 @@ export async function deleteStaticIPs(celoEnv: string) {
...
-    const [output] = await execCmd(
-      `kubectl delete pvc --selector='component=validators' --namespace ${celoEnv}`
-    )
-    console.info(output)
-
-    const [outputTx] = await execCmd(
-      `kubectl delete pvc --selector='component=tx_nodes' --namespace ${celoEnv}`
-    )
-    console.info(outputTx)
+    const componentLabels = ['validators', 'tx_nodes', 'sentry']
```
nambrot: How about doing them in parallel?

packages/celotool/src/lib/vm-testnet-utils.ts
```diff
@@ -339,15 +266,22 @@ export async function generateAndUploadSecrets(celoEnv: string) {
...
     await uploadSecrets(celoEnv, secrets, `tx-node-${i}`)
   }
   // Validators
   const validatorCount = parseInt(fetchEnv(envVar.VALIDATORS), 10)
   for (let i = 0; i < validatorCount; i++) {
-    const secrets = generateValidatorSecretEnvVars(AccountType.VALIDATOR, i)
+    const secrets = generateNodeSecretEnvVars(AccountType.VALIDATOR, i)
     await uploadSecrets(celoEnv, secrets, `validator-${i}`)
   }
+  // Proxies
```
kevjue, line 291: What do you think of adding an option to specify the percentage of validators that are proxied?  So that would allow for a testnet with some proxied validator and some standalone ones.

tkporter: The way it's set up you can set how many of the validators you want to be proxied, is that ok for what you were thinking? I imagine introducing percentages might slightly obfuscate what's going on

So for example
```
VALIDATORS=4
PROXIED_VALIDATORS=3
```
Would have 75% of the validators be proxied

packages/helm-charts/testnet/templates/validators.statefulset.yaml
```diff
@@ -61,20 +61,56 @@ spec:
...
+
             celotooljs.sh generate public-key --mnemonic "$MNEMONIC" --accountType load_testing --index 0 > /root/.celo/bootnodeEnodeAddress
             echo -n "Generating Bootnode enode address for the validator: "
             cat /root/.celo/bootnodeEnodeAddress
+
             [[ "$BOOTNODE_IP_ADDRESS" == 'none' ]] && BOOTNODE_IP_ADDRESS=${{ .Release.Namespace | upper }}_BOOTNODE_SERVICE_HOST
             echo `cat /root/.celo/bootnodeEnodeAddress`@$BOOTNODE_IP_ADDRESS:30301 > /root/.celo/bootnodeEnode
             echo -n "Generating Bootnode enode for the validator: "
             cat /root/.celo/bootnodeEnode
+
```
nambrot, line 82: This is becoming a pretty sizable script, consider extracting it into a proper files and then using helm to inject it

nambrot (approved):
Mostly skimmed and don't know much about the terraform stuff

packages/terraform-modules/testnet/modules/bootnode/startup.sh
```diff
@@ -43,7 +43,7 @@ echo "Pulling bootnode..."
 docker pull $GETH_BOOTNODE_DOCKER_IMAGE
 
 echo "Starting bootnode..."
-docker run -p 30301:30301/udp --name bootnode --net=host -d $GETH_BOOTNODE_DOCKER_IMAGE /bin/sh -c "\
+docker run -p 30301:30301/udp --name bootnode --net=host --restart=always -d $GETH_BOOTNODE_DOCKER_IMAGE /bin/sh -c "\
```
jcortejoso, line 46: I would say that if `--network=host`, the expose `-p` option does not affect in any way.

tkporter:
Closing in favor of #2073, which included this PR

Issue #1523, opened 2019-10-29
Verification V2 (Optional Verification)


nityas:
### Expected Behavior

Full Spec: https://docs.google.com/document/d/15oOHJlJk-yk19JRMZLO-Wm-iEB6tfgW4XcTtnNsJuEk/edit?pli=1#

Figma for onboarding: https://www.figma.com/file/9lBnIzsABhO8oXE0d0c8Km/Onboarding-and-Verification?node-id=10%3A557

Figma for tx flow: https://www.figma.com/file/yipmtvjgUreKhZyuK2slFg/Transaction-Flow%3A-Verification-and-Contact-Permissioning?node-id=0%3A1

Includes the Following (feel free to split up):
- invite code optional
- new verification flow
- post-onboarding prompts to re-verify

### Current Behavior

What is the current behavior?

PR #1635, opened 2019-11-08
[Wallet] Tie new verification screens into sagas and cleanup old code

### LINKED ISSUE #1350: Replace Verification Success screen with animation
### Expected Behavior

the user should see this animation: https://lottiefiles.com/share/MKKyBN after verification succeeds

### Current Behavior

Success screen with rings logo and timer

jmrossy:
### Description

- Remove files and strings from old verification flow
- Add fail modal to handle failures case
- Remove some unrelated dead code (e.g. BottomButton)
- Implement success screen incl. new animation
- Add a migration for the new state schema

### Tested

Tested verification flow for success + fail cases.
More testing still in progress

### Related issues

- #1523
- Fixes #1350 

### Backwards compatibility

Yes, added a migration for the state changes


packages/mobile/src/identity/contactMapping.test.ts
```diff
@@ -39,7 +38,6 @@ describe('Import Contacts Saga', () => {
     await expectSaga(doImportContacts)
       .provide([
         [call(getConnectedAccount), null],
-        [call(waitForUserVerified), true],
```
cmcewen, line 42: Why don't we have to wait for this?

packages/mobile/src/verify/VerificationInputScreen.tsx
```diff
@@ -2,93 +2,229 @@ import Link from '@celo/react-components/components/Link'
...
 import { Screens } from 'src/navigator/Screens'
 import { RootState } from 'src/redux/reducers'
-import { currentAccountSelector } from 'src/web3/selectors'
+import Logger from 'src/utils/Logger'
+
+const TAG = 'VerificationInputScreen'
 
 const CodeInput = withTextInputPasteAware(TextInput)
 
+function getRecodedAttestationValue(attestationCode: AttestationCode, t: i18n.TranslationFunction) {
```
cmcewen, line 44: You could use the global i18n.t and then you don't have to pass in the t value and could move this function to a different file

cmcewen: Also typo - "recorded"

packages/mobile/src/verify/VerificationInputScreen.tsx
```diff
@@ -2,93 +2,229 @@ import Link from '@celo/react-components/components/Link'
...
   }
 }
 
 class VerificationInputScreen extends React.Component<Props, State> {
   static navigationOptions = { header: null }
 
+  interval: number | undefined
+
   state: State = {
+    timer: 60,
```
cmcewen, line 157: I think maybe moving the timer to a sub-component that only counts in its own state would improve performance

packages/mobile/src/verify/VerificationInputScreen.tsx
```diff
@@ -2,93 +2,229 @@ import Link from '@celo/react-components/components/Link'
...
     codeInputValues: [],
+    isCodeSubmitting: [],
```
cmcewen: the name leads me to believe this is a bool - maybe `codeSubmittingStatuses`?

packages/mobile/src/verify/VerificationInputScreen.tsx
```diff
@@ -131,29 +268,38 @@ class VerificationInputScreen extends React.Component<Props, State> {
             {t('input.body2')}
           </Text>
           <Text style={styles.codeHeader}>{t('input.codeHeader1')}</Text>
```
cmcewen, line 263: Not to be too fancy but I believe you could do a map here of `[0,1,2]` and add one for the codeHeader to be more concise

packages/mobile/src/verify/VerificationLoadingScreen.tsx
```diff
@@ -5,77 +5,83 @@ import colors from '@celo/react-components/styles/colors'
...
+  componentDidUpdate() {
+    if (this.props.verificationStatus === VerificationStatus.Done) {
+      navigate(Screens.VerificationSuccessScreen)
+    } else if (this.props.verificationStatus === VerificationStatus.RevealingNumber) {
+      navigate(Screens.VerificationInterstitialScreen)
+    }
+  }
+
+  componentWillUnmount() {
+    BackHandler.removeEventListener('hardwareBackPress', this.handleBackButton)
```
cmcewen, line 68: So I'm not sure here exactly how this works, but I don't know if unmount is called if you push another screen onto the stack, so if that happens we could have multiple listeners at the same time. My intuition is that the latest one would take precedence, but just wanted to flag

packages/mobile/src/verify/VerificationSuccessScreen.tsx
```diff
@@ -1,38 +1,57 @@
...
+  static navigationOptions = { header: null }
+
+  state = {
+    isTextWhite: false,
+  }
+
+  componentDidMount() {
+    setTimeout(() => {
+      navigate(Stacks.AppStack)
+    }, 7800)
```
cmcewen: Why this number? Is this the animation length? Looks like there is an `onAnimationFinish` prop: https://github.com/react-native-community/lottie-react-native/blob/master/docs/api.md#component-api

cmcewen (commented):
woot woot woot!

packages/mobile/src/identity/contactMapping.test.ts
jmrossy: It's unrelated. it was only there because of the saga interference bug, which was mitigated when we moved import into a separate onboarding screen. So this is dead code now.

packages/mobile/src/verify/VerificationInputScreen.tsx
jmrossy: Actually Recoded is what I meant here, as in recoded into base64.
Also FYI isn't new code, it's from the old Verifying.tsx file, though I'm open to moving it to a new file. I kept it here cause it's kind of a hack specifically for this screen to show the same code in the UI as the user sees in their SMS

packages/mobile/src/verify/VerificationInputScreen.tsx
jmrossy: Good idea!

packages/mobile/src/verify/VerificationInputScreen.tsx
jmrossy: Will rename

packages/mobile/src/verify/VerificationLoadingScreen.tsx
jmrossy: Good catch. I'll look into it and see if there's a better solution

packages/mobile/src/verify/VerificationInputScreen.tsx
jmrossy: Wouldn't that cause a rerender of the child components every time because a new array is created?

packages/mobile/src/verify/VerificationSuccessScreen.tsx
jmrossy: Love it! Didn't realize that existed but should have guessed.

packages/mobile/src/verify/VerificationInputScreen.tsx
cmcewen: ohhhhh. gotcha gotcha. not important but maybe then just `getBase64AttestationValue` or even just `getDisplayAttestationCode` since it also might return a string with code accepted

packages/mobile/src/verify/VerificationInputScreen.tsx
cmcewen: this is an interesting question. i believe react diffs on what is returned from render, so assuming that the component was exactly the same with the same props, i think it would not rerender. i guess to ensure this we could create that array outside of the component, and then also give each coderow and text a key to make extra sure

i think in cases where you create inline functions and pass as props, it will rerender because the functions are different

packages/mobile/src/verify/VerificationInputScreen.tsx
jmrossy: Have this made but waiting on word from Nitya. Will circle back on this.

cmcewen (approved):

PR #1643, opened 2019-11-08
Fix support for validator key rotation, add end-to-end test

### LINKED ISSUE #1627: End-to-end tests for validator key rotation
### Expected Behavior

End-to-end tests of validators authorizing validating keys, updating BLS keys, etc.

### Current Behavior

No end-to-end tests of this.

asaj:
### Description

This PR fixes support for validator key rotation, and tests it in the end-to-end governance tests. 

### Tested

- Updates the governance end-to-end tests to test that key rotation works by using the account as the validator signer in epoch N, an authorized validator signer in epoch N+1, and a second authorized validator signer in epoch N+2

### Other changes
- Refactor `publicKeysData` logic to split out the various components, not store the proof-of-possession.
- Updated end-to-end governance tests to ensure no IBFT round changes by inferring the random proposer order and asserting blocks are proposed in that order 
- Fix small bug in end-to-end tests where private keys written to the same file
- Update `account:authorize` CLI command so that we don't assume account and signing keys are managed by the same node, add `account:proof-of-possession` command
- Make name optional in `account:register`
- Small changes to naming in `Accounts.sol`
- Reorder functions in `Accounts.sol`
- Move some logic from `protocol` into `utils`
 
### Related issues

- Fixes #1627 

### Backwards compatibility

Not backwards compatible


packages/cli/src/utils/command.ts
```diff
@@ -4,14 +4,24 @@ import { IArg, ParseFn } from '@oclif/parser/lib/args'
...
-    throw new CLIError(`${input} is not a public key`)
+    throw new CLIError(msg)
   }
 }
+
+const parseEcdsaPublicKey: ParseFn<string> = (input) => {
+  parseBytes(input, 64, `${input} is not an ECDSA public key`)
+}
+const parseBlsPublicKey: ParseFn<string> = (input) => {
+  parseBytes(input, 48, `${input} is not a BLS public key`)
```
kobigurk: Can we pull these numbers from a common place? Especially since they might change soon.

packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -123,16 +177,20 @@ contract UsingPrecompiles {
...
-   * @param proofOfPossessionBytes The public key and signature of the proof of possession.
+   * @param sender The address signed by the BLS key to generate the proof of possession.
+   * @param blsKey The BLS public key that the validator is using for consensus, should pass proof
+   *   of possession. 48 bytes.
+   * @param blsPop The BLS public key proof-of-possession, which consists of a signature on the
+   *   account address. 96 bytes.
    * @return True upon success.
    */
-  function checkProofOfPossession(address sender, bytes memory proofOfPossessionBytes)
+  function checkProofOfPossession(address sender, bytes memory blsKey, bytes memory blsPop)
```
kobigurk, line 133: I'm very happy you were able to separate it into two values! Previously, this caused the contract to have more than the maximum code size. How did you get around it?

packages/utils/src/bls.ts
```diff
@@ -35,3 +36,21 @@ export const blsPrivateKeyToProcessedPrivateKey = (privateKeyHex: string) => {
...
+  const blsPrivateKeyBytes = blsPrivateKeyToProcessedPrivateKey(privateKeyHex.slice(2))
+  return blsPrivateKeyBytes
+}
+
+export const getBlsPublicKey = (privateKeyHex: string) => {
+  const blsPrivateKeyBytes = getBlsPrivateKey(privateKeyHex)
+  return '0x' + bls12377js.BLS.privateToPublicBytes(blsPrivateKeyBytes).toString('hex')
+}
+
+export const getBlsPoP = (address: string, privateKeyHex: string) => {
```
kobigurk, line 53: We should probably verify address is a valid address.

packages/contractkit/src/wrappers/Accounts.ts
```diff
@@ -69,44 +71,75 @@ export class AccountsWrapper extends BaseWrapper<Accounts> {
...
+    )
   }
 
   /**
    * Authorizes an address to sign consensus messages on behalf of the account.
-   * @param account Address of the active account.
-   * @param validationSigner The address of the signing key to authorize.
+   * @param signer The address of the signing key to authorize.
+   * @param proofOfSigningKeyPossession The account address signed by the signer address.
+   * @param proofOfBlsKeyPossession Proof-of-possession generated for the corresponding BLS key. Only needed if a validator has been registered.
```
kobigurk: This is currently not implemented?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -85,102 +133,191 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
-   * @param voter The address of the vote signing key to authorize.
+   * @param signer The address of the signing key to authorize.
    * @param v The recovery id of the incoming ECDSA signature.
    * @param r Output value r of the ECDSA signature.
    * @param s Output value s of the ECDSA signature.
-   * @dev v, r, s constitute `voter`'s signature on `msg.sender`.
+   * @dev v, r, s constitute `signer`'s signature on `msg.sender`.
    */
-  function authorizeVoteSigner(address voter, uint8 v, bytes32 r, bytes32 s) external nonReentrant {
+  function authorizeVoteSigner(address signer, uint8 v, bytes32 r, bytes32 s)
```
kobigurk, line 142: Is this where the BLS rotation will be implemented?

asaj: Nope! This is actually unrelated, and allows you to specify a different ECDSA key for voting in governance in validator elections.

BLS rotation is implemented in `Validators.updateBlsKey`

packages/protocol/contracts/common/UsingPrecompiles.sol
asaj: We upped the max code size 😄

packages/contractkit/src/wrappers/Accounts.ts
asaj: Whoops! This was accidentally left over from the coupled-key implementation. BLS key rotation now happens in `Validators`

packages/protocol/contracts/common/UsingPrecompiles.sol
kobigurk: :D

packages/protocol/contracts/common/Accounts.sol
kobigurk: OK! I understand it better now.

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -85,102 +133,211 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
+   * @dev Fails if the `signer` is not an account or currently authorized attestation signer.
    * @return The associated account.
    */
-  function activeAttesttationSignerToAccount(address accountOrAttestationSigner)
-    external
-    view
-    returns (address)
-  {
-    address authorizingAccount = authorizedBy[accountOrAttestationSigner];
+  function activeAttesttationSignerToAccount(address signer) external view returns (address) {
```
nambrot: Typo

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -73,6 +79,48 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
     setWalletAddress(walletAddress);
   }
 
+  /**
```
nambrot, line 82: Why did you move the functions up here? I thought external functions go above public ones?

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -128,31 +136,115 @@ describe('governance tests', () => {
...
-      }
-      await initAndStartGeth(context.hooks.gethBinaryPath, groupInstance)
-      allValidators = await getValidatorGroupMembers()
-      assert.equal(allValidators.length, 5)
+      const groupPrivateKey = await getValidatorGroupPrivateKey()
+      const rotation0PrivateKey =
+        '0xa42ac9c99f6ab2c96ee6cae1b40d36187f65cd878737f6623cd363fb94ba7087'
+      const rotation1PrivateKey =
+        '0x4519cae145fb9499358be484ca60c80d8f5b7f9c13ff82c88ec9e13283e9de1a'
+      const additionalNodes: any[] = [
```
nategraf: `GethInstanceConfig` from `src/e2e-tests/util` would be more specific
```suggestion
      const additionalNodes: GethInstanceConfig[] = [
```

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -34,7 +37,7 @@ describe('governance tests', () => {
   let goldToken: any
   let registry: any
   let validators: any
-  let accounts: AccountsWrapper
+  let accounts: any
```
nambrot, line 43: Why are we not using the Wrapper anymore?

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -128,31 +136,115 @@ describe('governance tests', () => {
...
+  ) => {
+    const validator = (await validatorWeb3.eth.getAccounts())[0]
+    const signer = (await signerWeb3.eth.getAccounts())[0]
+    await unlockAccount(validator, validatorWeb3)
+    await unlockAccount(signer, signerWeb3)
+    const pop = await (await newKitFromWeb3(
+      signerWeb3
+    ).contracts.getAccounts()).generateProofOfSigningKeyPossession(validator, signer)
+    const validatorKit = newKitFromWeb3(validatorWeb3)
+    const validatorAccounts = await validatorKit._web3Contracts.getAccounts()
```
nambrot: What about using the ContractKit wrapper? That would avoid you having to estimate gas etc.

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -501,10 +657,10 @@ describe('governance tests', () => {
         // Assert equal to 10 decimal places due to rounding errors.
```
nategraf:
```suggestion
         // Assert equal to 9 decimal places due to rounding errors.
```

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -73,6 +79,48 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
+
+  /**
+   * @notice Setter for the data encryption key and version.
+   * @param dataEncryptionKey secp256k1 public key for data encryption. Preferably compressed.
+   */
+  function setAccountDataEncryptionKey(bytes memory dataEncryptionKey) public {
+    require(dataEncryptionKey.length >= 33, "data encryption key length <= 32");
+    accounts[msg.sender].dataEncryptionKey = dataEncryptionKey;
+    emit AccountDataEncryptionKeySet(msg.sender, dataEncryptionKey);
+  }
```
nategraf, line 122: The most common compression for public keys is 257 bits = 33 bytes if I am not mistaken. https://cryptobook.nakov.com/asymmetric-key-ciphers/elliptic-curve-cryptography-ecc#ecc-keys

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -85,102 +133,211 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
+    address signer,
+    bytes calldata ecdsaKey,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external nonReentrant {
+    Account storage account = accounts[msg.sender];
+    authorize(signer, v, r, s);
+    account.signers.validator = signer;
+    require(getValidators().updateEcdsaKey(msg.sender, signer, ecdsaKey));
```
nategraf: I think changing the signer with the previous overload could break the invariant that stored `ecdsa` key correspond to the authorized signing address.

nategraf: Does this need `require(!getValidators().isValidator(msg.sender));`

```suggestion
    require(!getValidators().isValidator(msg.sender));
    require(getValidators().updateEcdsaKey(msg.sender, signer, ecdsaKey));
```

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -85,102 +133,211 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
-  function activeAttesttationSignerToAccount(address accountOrAttestationSigner)
-    external
-    view
-    returns (address)
-  {
-    address authorizingAccount = authorizedBy[accountOrAttestationSigner];
+  function activeAttesttationSignerToAccount(address signer) external view returns (address) {
+    address authorizingAccount = authorizedBy[signer];
+    if (authorizingAccount != address(0)) {
+      require(accounts[authorizingAccount].signers.attestation == signer);
```
nategraf, line 218: If I am not mistaken, if this `require` ever fails, there is a bug. It may be worth adding a comment here on that.

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -85,102 +133,211 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
+    }
+  }
+
+  /**
+   * @notice Returns the account associated with `signer`.
+   * @param signer The address of the account or previously authorized vote signer.
+   * @dev Fails if the `signer` is not an account or previously authorized vote signer.
+   * @return The associated account.
+   */
+  function voteSignerToAccount(address signer) external view returns (address) {
```
nategraf, line 249: When would someone want to use this instead of `activeVoteSignerToAccount`? Also If I am not mistaken, this will return an account which is authorized as a validator or attester, which is confusing because the name mentions `vote`. You might consolidate the next three functions into one `signerToAccount` function

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -420,31 +423,38 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
+  function authorize(address authorized, uint8 v, bytes32 r, bytes32 s) private {
+    require(isAccount(msg.sender) && isNotAccount(authorized) && isNotAuthorizedSigner(authorized));
 
     address signer = Signatures.getSignerOfAddress(msg.sender, v, r, s);
-    require(signer == current);
+    require(signer == authorized);
 
-    authorizedBy[previous] = address(0);
-    authorizedBy[current] = msg.sender;
+    authorizedBy[authorized] = msg.sender;
```
nategraf, line 427: It seems that with this change, it seems once an authorization is replaced, the previously authorized address can never again become an authorized signer. Is that the intention? If so could you add some comments to that effect were relevant.

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -420,31 +423,38 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
-   * @param previous The previous authorized address.
+   * @notice Authorizes some role of of `msg.sender`'s account to another address.
+   * @param authorized The address to authorize.
    * @param v The recovery id of the incoming ECDSA signature.
    * @param r Output value r of the ECDSA signature.
    * @param s Output value s of the ECDSA signature.
    * @dev Fails if the address is already authorized or is an account.
    * @dev v, r, s constitute `current`'s signature on `msg.sender`.
    */
-  function authorize(address current, address previous, uint8 v, bytes32 r, bytes32 s) private {
```
nategraf, line 441: In my pass through this file I didn't see a way to clear authorizations (i.e. set `account.signer.* = address(0)`) Did I miss it? If not, that seems like an important function to add.

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -470,46 +477,82 @@ contract Validators is
...
+    onlyRegisteredContract(ACCOUNTS_REGISTRY_ID)
+    returns (bool)
+  {
     require(isValidator(account));
     Validator storage validator = validators[account];
-    _updatePublicKeysData(validator, publicKeysData);
-    emit ValidatorPublicKeysDataUpdated(account, publicKeysData);
+    require(_updateEcdsaKey(validator, signer, ecdsaKey));
+    emit ValidatorEcdsaKeyUpdated(account, ecdsaKey);
     return true;
```
nategraf, line 538: Why does this function return a bool if it reverts on failure?

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -249,39 +411,39 @@ describe('governance tests', () => {
...
-          (await validators.methods.getValidator(validator).call({}, blockNumber))[3]
+          (await validators.methods.getValidator(validator).call({}, blockNumber)).score
         )
         const previousScore = new BigNumber(
-          (await validators.methods.getValidator(validator).call({}, blockNumber - 1))[3]
+          (await validators.methods.getValidator(validator).call({}, blockNumber - 1)).score
         )
-        assert.isNotNaN(score)
-        assert.isNotNaN(previousScore)
+        assert.isFalse(score.isNaN())
```
nambrot, line 409: Looks like `chai` has a function for this `assert.isNotNaN(score)`

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -258,27 +260,32 @@ contract Validators is
...
+    returns (bool)
+  {
+    address account = getAccounts().activeValidatorSignerToAccount(msg.sender);
     require(!isValidator(account) && !isValidatorGroup(account));
     uint256 lockedGoldBalance = getLockedGold().getAccountTotalLockedGold(account);
     require(lockedGoldBalance >= validatorLockedGoldRequirements.value);
     Validator storage validator = validators[account];
-    _updatePublicKeysData(validator, publicKeysData);
-    validator.publicKeysData = publicKeysData;
+    address signer = getAccounts().getValidatorSigner(account);
```
nategraf, line 283: Is this not always equal to `msg.sender` by virtue of having obtained `account` via `activeValidatorSignerToAccount`?

packages/cli/src/commands/account/proof-of-possession.ts
```diff
@@ -0,0 +1,28 @@
...
+  static description = 'Generate proof-of-possession to be used to authorize a signer'
+
+  static flags = {
+    ...BaseCommand.flags,
+    signer: Flags.address({ required: true }),
+    account: Flags.address({ required: true }),
+  }
+
+  static examples = [
+    'proof-of-possession --account 0x5409ed021d9299bf6814279a6a1411a7e866a631 --signer 0x6ecbe1db9ef729cbe972c83fb886247691fb6beb',
```
nambrot, line 16: As naming czar, I wonder what you think about `prove-possession` as our commands are generally actions/verbs?

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -706,19 +749,39 @@ contract Validators is
...
+   * @notice Returns the validator BLS key.
+   * @param signer The account that registered the validator or its authorized signing address.
+   * @return The validator BLS key.
+   */
+  function getValidatorBlsKeyFromSigner(address signer)
+    external
+    view
+    returns (bytes memory blsKey)
+  {
+    address account = getAccounts().validatorSignerToAccount(signer);
```
nategraf: Is there a reason this is `validatorSignerToAccount` instead of `activeValidatorSignerToAccount`?

packages/contractkit/src/wrappers/Accounts.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  const registerAccountWithLockedGold = async (account: string) => {
+    if (!(await accountsInstance.isAccount(account))) {
+      await accountsInstance.createAccount().sendAndWaitForReceipt({ from: account })
+    }
+    await lockedGold.lock().sendAndWaitForReceipt({ from: account, value: minLockedGoldValue })
+  }
+
+  const getParsedSignatureOfAddress = async (address: string, signer: string) => {
+    const addressHash = web3.utils.soliditySha3({ type: 'address', value: address })
+    console.log('account', address)
```
nambrot: Should these be here?

packages/contractkit/src/wrappers/Accounts.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    console.log('addressHash', addressHash)
+    const signature = await web3.eth.sign(addressHash, signer)
+    return parseSignature(addressHash, signature, signer)
+  }
+
+  beforeAll(async () => {
+    accounts = await web3.eth.getAccounts()
```
nambrot, line 43: in `test-utils/ganache.setup.ts` we have the list of accounts as well

packages/contractkit/src/wrappers/Accounts.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    lockedGold = await kit.contracts.getLockedGold()
+    accountsInstance = await kit.contracts.getAccounts()
+  })
+
+  const setupValidator = async (validatorAccount: string) => {
+    const publicKey = await addressToPublicKey(validatorAccount, web3)
+    await registerAccountWithLockedGold(validatorAccount)
+    await validators
+      .registerValidator(
+        // @ts-ignore
```
nambrot: Should these fit on a single line?

packages/contractkit/src/wrappers/Accounts.ts
```diff
@@ -9,11 +14,12 @@ import {
   toTransactionObject,
 } from '../wrappers/BaseWrapper'
 
-enum SignerRole {
-  Attestation,
-  Validation,
-  Vote,
+export interface Signature {
```
nambrot: You should be able to import this from ` '@celo/utils/lib/signatureUtils`

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -85,102 +133,211 @@ contract Accounts is IAccounts, ReentrancyGuard, Initializable, UsingRegistry {
...
+    emit ValidatorSignerAuthorized(msg.sender, signer);
   }
 
   /**
-   * @notice Check if an address has been authorized by an account for voting or validating.
-   * @param account The possibly authorized address.
-   * @return Returns `true` if authorized. Returns `false` otherwise.
+   * @notice Authorizes an address to sign consensus messages on behalf of the account.
+   * @param signer The address of the signing key to authorize.
+   * @param ecdsaKey The ECDSA key corresponding to `signer`.
```
nambrot: I'm not too familiar with the terminology, but does `ECDSA key` always imply the public key? In my naive reading I would have assumed private key

packages/protocol/test/governance/validators.ts
```diff
@@ -1016,53 +1068,116 @@ contract('Validators', (accounts: string[]) => {
...
-      describe('when the public keys data is too long', () => {
+      describe('when the public key is not 48 bytes', () => {
         it('should revert', async () => {
           // @ts-ignore Broken typechain typing for bytes
-          await assertRevert(validators.updatePublicKeysData(newPublicKeysData + '00'))
+          await assertRevert(validators.updateBlsKey(newBlsPublicKey + '01', newBlsPoP))
         })
       })
 
-      describe('when the public keys data is too short', () => {
```
nategraf, line 1062: It would be helpful to also have a check to make sure we revert if the PoP is invalid (e.g using `newBlsPoP.slice(0, -2) + '01'`)

packages/contractkit/src/wrappers/Accounts.ts
```diff
@@ -168,25 +228,14 @@ export class AccountsWrapper extends BaseWrapper<Accounts> {
...
-    [SignerRole.Attestation]: this.contract.methods.authorizeAttestationSigner,
-    [SignerRole.Validation]: this.contract.methods.authorizeValidationSigner,
-    [SignerRole.Vote]: this.contract.methods.authorizeVoteSigner,
-  }
-
-  private async authorizeSigner(role: SignerRole, account: Address, signer: Address) {
-    const sig = await this.getParsedSignatureOfAddress(account, signer)
-    // TODO(asa): Pass default tx "from" argument.
-    return toTransactionObject(this.kit, this.authorizeFns[role](signer, sig.v, sig.r, sig.s))
+  parseSignatureOfAddress(address: Address, signer: string, signature: string) {
```
nambrot, line 226: Optional Nit: is `signer` also an `Address`?

packages/contractkit/src/wrappers/Validators.test.ts
```diff
@@ -52,12 +48,17 @@ testWithGanache('Validators Wrapper', (web3) => {
...
 
   const setupValidator = async (validatorAccount: string) => {
+    const publicKey = await addressToPublicKey(validatorAccount, web3)
     await registerAccountWithLockedGold(validatorAccount)
     // set account1 as the validator
     await validators
       .registerValidator(
         // @ts-ignore
-        publicKeysData
+        publicKey,
```
nambrot: Optional Nit: Fit on one line?

packages/protocol/contracts/common/Accounts.sol
asaj: @nambrot 

This was just function re-ordering, so I'm not going to address in this PR if it's okay

packages/protocol/contracts/common/Accounts.sol
asaj: The intended invariant is slightly different, which is that if you're a registered validator, the stored `ecdsa` key corresponds to the validator signer (either the account key or an authorized validator signer)

asaj: Actually, we want to `require(getValidators().isValidator(msg.sender));`, but `updateEcdsaKey` will do this check for us

packages/protocol/contracts/common/Accounts.sol
asaj: Yeah that is the intention, will add comments.

packages/protocol/contracts/common/Accounts.sol
asaj: Nice eye, indeed we do not support this. I think that's *okay*, but perhaps not ideal.

packages/protocol/contracts/governance/Validators.sol
asaj: Just following a common solidity pattern

packages/protocol/contracts/governance/Validators.sol
asaj: Not necessarily, `msg.sender` could be the account address

packages/protocol/contracts/governance/Validators.sol
asaj: Hmm, I'm not certain it matters in this case but I feel that `validatorSignerToAccount` here is safer. Wdyt?

packages/protocol/test/governance/validators.ts
asaj: Since we actually don't have this precompile in ganache we can't check this yet 🙁

packages/protocol/contracts/common/Accounts.sol
nategraf: Oh I misread this anyway. It's correct.

packages/protocol/contracts/common/Accounts.sol
nategraf: Ok I see now. The first overload is for if `msg.sender` is not a validator, and the second is for if they are. Thanks for clarifying.

packages/protocol/contracts/governance/Validators.sol
nategraf: I don't understand. Is it not true that `x == getValidatorSigner(activeValidatorSignerToAccount(x))`?

packages/protocol/contracts/governance/Validators.sol
nategraf: I don't think it is safer, but on second look I aggree that is generally doesn't matter. One thing is though that it might be assumed that the input needs to be the validators authorized signer for validation, but the result would be the same if the input signer was authorized for voting or attestation instead.

packages/utils/src/address.ts
```diff
@@ -1,7 +1,35 @@
...
+  privateToPublic,
+  pubToAddress,
+  sha3,
+  toChecksumAddress,
+} from 'ethereumjs-util'
+import Web3 from 'web3'
 
 export type Address = string
 
+export async function addressToPublicKey(address: string, web3: Web3) {
```
nambrot: The `utils` package currently does not specify `web3` as a dependency, thus generally what I did in `signatureUtils` is to pass the `web3.eth.sign` method as a parameter instead. That also highlights to the consumer of this function that there should be a web3 provider that can actually sign (vs. a web3 parameter that is used to make view calls for example).

nambrot: Also would this make sense in signatureUtils instead?

packages/utils/src/address.ts
```diff
@@ -1,7 +1,35 @@
...
+  const msg = new Buffer('dummy_msg_data')
+  const data = '0x' + msg.toString('hex')
+  // Note: Eth.sign typing displays incorrect parameter order
+  const sig = await web3.eth.sign(data, address)
+
+  const rawsig = fromRpcSig(sig)
+
+  const prefix = new Buffer('\x19Ethereum Signed Message:\n')
+  const prefixedMsg = sha3(Buffer.concat([prefix, new Buffer(String(msg.length)), msg]))
```
nambrot: Probably use `hashMessageWithPrefix` in signatureUtils?

packages/utils/src/address.ts
```diff
@@ -12,4 +40,9 @@ export const privateKeyToAddress = (privateKey: string) => {
   )
 }
 
+export const privateKeyToPublicKey = (privateKey: string) => {
+  return toChecksumAddress(
```
nambrot, line 18: Did you mean to call toChecksumAddress here?

packages/cli/src/utils/command.ts
asaj: Added constants in `@celo/utils/src/bls`

packages/protocol/contracts/common/Accounts.sol
asaj: Ah so actually upon revisiting, this is not the case. If I authorize A as my attestation signer, and then B, and then call `activeAttestationSignerToAccount(A)`, we would hit this line.

packages/protocol/contracts/common/Accounts.sol
asaj: +1 to `signerToAccount`

packages/protocol/contracts/governance/Validators.sol
asaj: That will be true if `x` is an authorized signer, but not if `x` is an account address.

packages/protocol/contracts/common/Accounts.sol
asaj: I think the more important constraint is `view` functions go last, but ultimately we should just turn linting back on for this

packages/celotool/src/e2e-tests/governance_tests.ts
asaj: Because we need to call functions at particular blocks

packages/celotool/src/e2e-tests/governance_tests.ts
asaj: Yeah that's what we used to use, and it's dangerous because the `BigNumber` representation of `NaN` is not considered `NaN` by chai

packages/cli/src/commands/account/proof-of-possession.ts
asaj: I feel that's a bit misleading, since we're not actually proving possession here, but merely *generating* the pop. Wdyt?

packages/contractkit/src/wrappers/Accounts.test.ts
asaj: No they should not

packages/contractkit/src/wrappers/Accounts.test.ts
asaj: Leaving as-is if it's all the same to you

packages/utils/src/address.ts
asaj: Yeah, any reason not to?

packages/protocol/contracts/governance/Validators.sol
nategraf: Oh I see now. Thanks for explaining that.

packages/protocol/contracts/common/Accounts.sol
nategraf: I agree that this can wait until we see if anyone is missing it.

nategraf (approved):
Thanks for addressing all my comments 👍

packages/celotool/src/e2e-tests/governance_tests.ts
nambrot: @mcortesi Would be nice to expose that functionality on `proxyCall`?

packages/cli/src/commands/account/proof-of-possession.ts
nambrot: sgtm, only thing I might add is `generate-pop`?

packages/utils/src/address.ts
nambrot: I thought checksums only apply to addresses, not public keys

packages/utils/src/signatureUtils.ts
```diff
@@ -25,6 +27,34 @@ export interface Signer {
   sign: (message: string) => Promise<string>
 }
 
+export async function addressToPublicKey(
+  signer: string,
+  signFn: (message: string, signer: string) => Promise<string>
+) {
+  const msg = new Buffer('dummy_msg_data')
+  const data = '0x' + msg.toString('hex')
+  // Note: Eth.sign typing displays incorrect parameter order
```
nambrot, line 36: This note probably should be elsewhere now?

packages/utils/src/signatureUtils.ts
```diff
@@ -25,6 +27,34 @@ export interface Signer {
...
+  const sig = await signFn(data, signer)
+
+  const rawsig = ethjsutil.fromRpcSig(sig)
```
nambrot, line 39: Optional Nit: Didn't know about `fromRpcSig`!! Could we streamline either that or our implementation (from below =`parseSignature`) here?

nambrot (approved):
Just nits!

.circleci/config.yml
```diff
@@ -524,7 +524,7 @@ jobs:
           command: |
             set -e
             cd packages/celotool
-            ./ci_test_transfers.sh checkout master
+            ./ci_test_transfers.sh checkout asaj/key-rotation-plus-enode
```
jmrossy, line 528: do you intend to keep these pointing to your branch? Do you fix set it back in a follow-up PR once tests pass?

packages/utils/src/address.ts
```diff
@@ -12,12 +12,18 @@ export const privateKeyToAddress = (privateKey: string) => {
...
-export const privateKeyToPublicKey = (privateKey: string) => {
-  return '0x' + privateToPublic(Buffer.from(privateKey.slice(2), 'hex')).toString('hex')
+export const publicKeyToAddress = (publicKey: string) => {
+  return toChecksumAddress(
+    '0x' + pubToAddress(Buffer.from(publicKey.slice(2), 'hex')).toString('hex')
+  )
 }
 
 export { toChecksumAddress } from 'ethereumjs-util'
+export { isValidAddress } from 'ethereumjs-util'
```
jmrossy, line 47: Why are these re-exported here?
@nambrot FYI this conflicts with the changes you made in your Attestations svc PR

packages/utils/src/address.ts
```diff
@@ -12,12 +12,18 @@ export const privateKeyToAddress = (privateKey: string) => {
...
+export const privateKeyToPublicKey = (privateKey: string) => {
+  return toChecksumAddress(
+    '0x' + privateToPublic(Buffer.from(privateKey.slice(2), 'hex')).toString('hex')
+  )
 }
 
-export const privateKeyToPublicKey = (privateKey: string) => {
-  return '0x' + privateToPublic(Buffer.from(privateKey.slice(2), 'hex')).toString('hex')
+export const publicKeyToAddress = (publicKey: string) => {
+  return toChecksumAddress(
```
jmrossy, line 24: Leaving similar feedback I left for Nam on his PR that touches this:
Would be nice these methods confirmed the leading chars are '0x' before stripping them. So happens @yorhodes has added utils in this file for just this
https://github.com/celo-org/celo-monorepo/pull/1570

packages/utils/src/signatureUtils.ts
jmrossy: Yeah, utils/address seems like a better place

packages/utils/src/signatureUtils.ts
```diff
@@ -25,6 +27,34 @@ export interface Signer {
...
+    rawsig.s
+  )
+
+  const computedAddr = ethjsutil.pubToAddress(pubKey).toString('hex')
+  assert(eqAddress(computedAddr, signer), 'computed address !== signer')
+
+  return '0x' + pubKey.toString('hex')
+}
+
+export function eqAddress(a: Address, b: Address) {
```
jmrossy: There's an `areAddressesEqual` below in this file AND @yorhodes adds another one in his PR here: #1570 
I'll leave it to you two to coordinate but pretty sure we don't need another util method for this ;)

jmrossy (commented):
Some small comments

.circleci/config.yml
asaj: Yeah, you need to do it this way in order to run the end-to-end tests properly

packages/utils/src/address.ts
asaj: They're used elsewhere, allows us to only depend on`@celo/utils` in those places rather than also `ethereumjs-utils`

packages/utils/src/address.ts
asaj: `signatureUtils` has this functionality, suggest he use that.

packages/utils/src/signatureUtils.ts
asaj: Must have been a merge conflict? This was supposed to move to address.ts

PR #1645, opened 2019-11-09
Fix invite script by moving to contractkit

cmcewen:
### Description

Invite script was broken so just decided to move to celotool + contractkit

### Tested

Ran locally and got a text with a valid invite

### Backwards compatibility

Shouldn't break anything


packages/celotool/src/cmds/account/invite.ts
```diff
@@ -12,39 +14,102 @@ export const describe = 'command for sending an invite code to a phone number'
...
+  const one = new BigNumber(10).pow(decimals.toNumber())
+  return one.times(value)
 }
 
 export const handler = async (argv: InviteArgv) => {
-  console.log(`Sending invitation code to ${argv.phone}`)
+  const phone = argv.phone
+
+  console.log(`Sending invitation code to ${phone}`)
+  await execCmd('gcloud config set project celo-testnet')
```
asaj, line 33: I believe we have a utility function that switches you to the correct project based on the `.env` file

packages/celotool/src/cmds/account/invite.ts
```diff
@@ -12,39 +14,102 @@ export const describe = 'command for sending an invite code to a phone number'
...
   const cb = async () => {
-    await execCmd(
-      `yarn --cwd ../protocol run invite -n ${argv.celoEnv} -p ${argv.phone} -f ${argv.fast}`
+    const kit = newKit('http://localhost:8545')
+    const account = (await kit.web3.eth.getAccounts())[0]
+    console.log(`Using account: ${account}`)
+    kit.defaultAccount = account
+
+    // TODO(asa): This number was made up
+    const verificationGasAmount = new BigNumber(10000000)
```
asaj: s/verification/attestation everywhere

packages/celotool/src/cmds/account/invite.ts
```diff
@@ -12,39 +14,102 @@ export const describe = 'command for sending an invite code to a phone number'
...
+  )
   const cb = async () => {
-    await execCmd(
-      `yarn --cwd ../protocol run invite -n ${argv.celoEnv} -p ${argv.phone} -f ${argv.fast}`
+    const kit = newKit('http://localhost:8545')
+    const account = (await kit.web3.eth.getAccounts())[0]
+    console.log(`Using account: ${account}`)
+    kit.defaultAccount = account
+
+    // TODO(asa): This number was made up
```
asaj, line 44: Feels dangerous but I guess it's been working for us

packages/celotool/src/cmds/account/invite.ts
```diff
@@ -12,39 +14,102 @@ export const describe = 'command for sending an invite code to a phone number'
...
+    const [goldToken, stableToken, attestations, escrow] = await Promise.all([
+      kit.contracts.getGoldToken(),
+      kit.contracts.getStableToken(),
+      kit.contracts.getAttestations(),
+      kit.contracts.getEscrow(),
+    ])
+    const verificationFee = new BigNumber(
+      await attestations.attestationRequestFees(stableToken.address)
+    )
+    const goldAmount = verificationGasAmount.times(gasPrice).toString()
```
asaj: We don't need to send gold anymore at all, do we?

packages/celotool/src/cmds/account/invite.ts
cmcewen: Key is only present in celo-testnet so I added a comment

packages/celotool/src/cmds/account/invite.ts
cmcewen: I'm not sure if anyone uses this + pays gas in gold so I just added a comment to remove it - just hesitant to do it when not everyone will have a chance to look at this

packages/celotool/src/cmds/account/invite.ts
```diff
@@ -12,39 +14,96 @@ export const describe = 'command for sending an invite code to a phone number'
...
+
+    await stableToken.approve(escrow.address, stableTokenEscrowAmount).sendAndWaitForReceipt()
+    const expirySeconds = 60 * 60 * 24 * 5 // 5 days
+
+    console.log(
+      `Transferring ${goldAmount} Gold, ${stableTokenInviteAmount} StableToken, and escrowing ${stableTokenEscrowAmount} StableToken`
+    )
+    await Promise.all([
+      // TODO: remove if no one is paying for gas with gold
+      goldToken.transfer(temporaryAddress, goldAmount).sendAndWaitForReceipt(),
```
jmrossy, line 81: Yeah, i think your comment is right. The invite doesn't actually need to send gold.

packages/celotool/src/cmds/account/invite.ts
```diff
@@ -12,39 +14,96 @@ export const describe = 'command for sending an invite code to a phone number'
...
   const cb = async () => {
-    await execCmd(
-      `yarn --cwd ../protocol run invite -n ${argv.celoEnv} -p ${argv.phone} -f ${argv.fast}`
+    const kit = newKit('http://localhost:8545')
+    const account = (await kit.web3.eth.getAccounts())[0]
+    console.log(`Using account: ${account}`)
+    kit.defaultAccount = account
+
+    // TODO(asa): This number was made up
+    const attestationGasAmount = new BigNumber(10000000)
```
jmrossy, line 45: Pretty sure we can cut this and the gold related parts too

jmrossy (approved):

PR #1649, opened 2019-11-10
Verifier release build and remove some unused env variables

cmcewen:
### Description

Verifier release build was failing - also found some unused env variables.

### Tested

Did a build of the verifier app and was able to run it and send texts

### Other changes

Unused env variables

### Backwards compatibility

No issues


packages/mobile/src/config.ts
```diff
@@ -26,8 +26,6 @@ export const CELO_VERIFIER_START_MINING_LINK = 'celo://verifier/start'
 export const CELO_FAUCET_LINK = 'https://celo.org/app'
 export const CELO_TERMS_LINK = 'https://celo.org/terms'
 
-export const DEFAULT_COUNTRY = Config.DEFAULT_COUNTRY || null
```
jmrossy, line 29: Why remove this? In fact, if anything we should set it to Liberia fro the pilot right?

cmcewen: We don't use it anymore anywhere - searching for DEFAULT_COUNTRY doesn't show anything

jmrossy: Ah good catch. The phone number input component still supports it though:
packages/react-components/components/PhoneNumberInput.tsx

So should we add it back? I think it would be good to have

jmrossy (approved):

Issue #1620, opened 2019-11-06
Enforce consistency between the network in .env and the db in json


jmrossy:
### Expected Behavior

Enforce consistency btwn the network in the .env files and firebase databases in the google-services.json files 

### Current Behavior

The firebase DB specified in the google-services.json files can be out of sync and overwritten when running yarn

jmrossy:
I fixed this when I created the run_app.sh script

PR #1651, opened 2019-11-11
[Wallet] Hide get invite code link for pilot build and firebase db fixes

jmrossy:
### Description

- Hide get invite code link for pilot build
- Fix firebase dbs in google service json files
- Fix small bug in keyplacer script

### Tested

Checked invite screen

### Related issues

- #1620

### Backwards compatibility

Yes


scripts/key_placer.sh
```diff
@@ -74,9 +74,7 @@ for file_path in "${files[@]}"; do
 
   # When encrypting ensure the plaintext file exists and the encrypted file does not.
   if [[ $1 == "encrypt" ]]; then
-    if [[ -f "$encrypted_file_path" ]]; then
-        continue
-    elif [[ ! -f "$file_path" ]]; then
+    if [[ ! -f "$file_path" ]]; then
```
i1skn, line 77: I assume, that before we were overriding encrypted configs if they are already exists and now we will override them, correct? If yes - should we update the comment on line 75?

jmrossy: Yes, that's the intended behavior :)
And I'll update the comment, good catch!

i1skn (approved):

PR #1652, opened 2019-11-11
[wallet] Doesn't calculate fee on empty balance

### LINKED ISSUE #1426: Fee calculation fails if account balance is 0
### Expected Behavior

No user displayed error. Just log an error and set fee to 0.

### Current Behavior

This error:
fees/saga/estimateFeeSaga/Error estimating fee:Returned error: gas required exceeds allowance or always failing transaction in Error: Returned error: gas required exceeds allowance or always failing transaction

martinvol:
### Description

Prevents the app from updating fees if the balance is zero, due web3 known issue.

### Tested

Added test case and mocked the others.

### Related issues

- Fixes #1426

### Backwards compatibility

-


packages/mobile/src/fees/saga.ts
```diff
@@ -27,6 +28,16 @@ const placeholderSendTx: BasicTokenTransfer = {
 
 export function* estimateFeeSaga({ feeType }: EstimateFeeAction) {
   Logger.debug(`${TAG}/estimateFeeSaga`, `updating for ${feeType}`)
+
+  const balance = yield select(stableTokenBalanceSelector)
+  if (balance === '0') {
+    Logger.warn(`${TAG}/estimateFeeSaga`, 'Error estimating fee')
```
jmrossy: Can we get a more specific warning message here?

packages/mobile/src/fees/saga.ts
```diff
@@ -27,6 +28,16 @@ const placeholderSendTx: BasicTokenTransfer = {
 
 export function* estimateFeeSaga({ feeType }: EstimateFeeAction) {
   Logger.debug(`${TAG}/estimateFeeSaga`, `updating for ${feeType}`)
+
+  const balance = yield select(stableTokenBalanceSelector)
+  if (balance === '0') {
```
jmrossy, line 40: Might be good to parse the balance and check it here, in case it's null or empty string

jmrossy (commented):
Nice! Just a few small suggestions

packages/mobile/src/fees/saga.ts
martinvol: What should happen if the balance it's null or empty?

jmrossy: The same thing :)

jmrossy (approved):

PR #1663, opened 2019-11-12
Add AccountClaim to Metadata

### LINKED ISSUE #1576: Users should be able to claim other addresses
### Expected Behavior

### Current Behavior

What is the current behavior?

nambrot:
### Description

This PR adds the ability for accounts to claim control of other accounts. Such claim can be verified by finding the corresponding claim in the other account's metadata.

### Tested

- [x] Unit tests
- [ ] e2e on `namoffchainreveal`

### Other changes

- Extracted common `io-ts` types into `utils`

### Related issues

- Fixes #1576


packages/utils/src/address.ts
```diff
@@ -12,4 +12,12 @@ export const privateKeyToAddress = (privateKey: string) => {
   )
 }
 
+export const publicKeyToAddress = (publicKey: string) => {
+  return '0x' + pubToAddress(Buffer.from(publicKey.slice(2), 'hex')).toString('hex')
```
jmrossy, line 16: Pretty sure web3 has a util for this already.

packages/utils/src/address.ts
```diff
@@ -12,4 +12,12 @@ export const privateKeyToAddress = (privateKey: string) => {
...
+}
+
+export const privateKeyToPublicKey = (privateKey: string) => {
+  return '0x' + privateToPublic(Buffer.from(privateKey.slice(2), 'hex')).toString('hex')
```
jmrossy, line 20: ditto

jmrossy (approved):
Lgtm, just two nits below

packages/utils/src/address.ts
nambrot: Can you link it? I couldn't see it in particular? Also, I think we have generally tried to avoid depending on web3 beyond web3-utils in `utils`

jmrossy: Yeah good point, nvm

mcortesi (approved):

PR #1665, opened 2019-11-12
Point end-to-end governance test back to master

asaj:
### Description
Points end-to-end tests back to master after https://github.com/celo-org/celo-monorepo/pull/1558


kevjue (approved):
LGTM!

PR #1676, opened 2019-11-12
[Wallet] Limit QR code scanner to 1 code per second

### LINKED ISSUE #1524: Issues when sending/requesting money from an account from a QR code
iOS 13.2, iPhone 8

### Expected Behavior

Everything works normally when sending/requesting from an account that has been scanned. 

### Current Behavior

For some reason sending/requesting money from someone in my contacts is pretty frictionless, but when I tried doing the same with an account that was scanned (in my example, my own account) the app started behaving weirdly. I'm wondering if the camera from the previous view is still consuming resources?

1. Request / Send call to action not showing up all the time

IMG_3903

IMG_3901

2. New account balance didn't change for a little bit (I have 10.04 cUSD)

IMG_3902

3. When the Send/Request call to action eventually did show (this was not deterministic), my autosuggest covered it

IMG_3900

4. When I tried to send/request while the testnet reminder banner was still there, I was then unable to "swipe" it away

IMG_3899

5. I could not use the back arrow on the navigation bar to back out of the process-- I had to quit the app

6. General slowness when changing between "amount" input and "for" input

i1skn:
### Description

QR code scanner would emit scanned codes multiple times for the same code. This PR would limit it to scan 1 code per second. So, essentially after the code scanned - 1 second should be enough to leave the QR code scan screen.

### Tested

Ran on iPhone.

### Other changes

Disable Dark Mode in iOS, because it font colors were inverted in it.

### Related issues

- Fixes #1524

### Backwards compatibility
Yes

### DIFF
--- a/packages/mobile/ios/Podfile.lock
+++ b/packages/mobile/ios/Podfile.lock
@@ -491,22 +491,19 @@ DEPENDENCIES:
   - Yoga (from `../../../node_modules/react-native/ReactCommon/yoga`)
 
 SPEC REPOS:
-  https://cdn.cocoapods.org/:
+  https://github.com/cocoapods/specs.git:
     - Analytics
     - boost-for-react-native
-    - Fabric
-    - Firebase
-    - FirebaseCore
-    - FirebaseInstanceID
-    - GoogleUtilities
-  https://github.com/cocoapods/specs.git:
     - Crashlytics
+    - Fabric
     - Firebase
     - FirebaseAnalytics
     - FirebaseAnalyticsInterop
     - FirebaseAuth
     - FirebaseAuthInterop
+    - FirebaseCore
     - FirebaseDatabase
+    - FirebaseInstanceID
     - FirebaseMessaging
     - FirebaseStorage
     - GoogleAppMeasurement
--- a/packages/mobile/src/qrcode/QRScanner.tsx
+++ b/packages/mobile/src/qrcode/QRScanner.tsx
@@ -23,19 +23,35 @@ interface DispatchProps {
   handleBarcodeDetected: typeof handleBarcodeDetected
 }
 
+interface State {
+  isScanningEnabled: boolean
+}
+
 type Props = DispatchProps & WithNamespaces & NavigationFocusInjectedProps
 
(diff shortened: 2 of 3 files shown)


packages/mobile/ios/Podfile.lock
```diff
@@ -491,22 +491,19 @@ DEPENDENCIES:
...
 SPEC REPOS:
-  https://cdn.cocoapods.org/:
+  https://github.com/cocoapods/specs.git:
     - Analytics
     - boost-for-react-native
-    - Fabric
-    - Firebase
-    - FirebaseCore
-    - FirebaseInstanceID
-    - GoogleUtilities
```
jmrossy, line 501: Looks like GoogleUtilities is now removed from this list? And Firebase too. The same thing happened on in my podfile.lock when I ran `bundle exec pod install`. Not sure if it will have any affect. @jeanregisser ?

packages/mobile/src/qrcode/QRScanner.tsx
```diff
@@ -23,19 +23,35 @@ interface DispatchProps {
...
 
+  // This method would be called multiple times with the same
+  // QR code, so we need to catch only the first one
   onBardCodeDetected = (rawData: any) => {
-    Logger.debug('QRScanner', 'Bar code detected')
-    this.props.handleBarcodeDetected(rawData)
+    if (this.state.isScanningEnabled) {
+      this.setState({ isScanningEnabled: false })
+      Logger.debug('QRScanner', 'Bar code detected')
+      this.props.handleBarcodeDetected(rawData)
```
jmrossy: You should move this code to the callback function of setState, otherwise I still think this can run more than once. 
Also please clear the timeout in the componentWillUnmount just in case

i1skn: Have addressed that

packages/mobile/ios/Podfile.lock
jeanregisser: Looks like it's fine for now. I think it's still some leftover from when you ran it with a newer CocoaPods version.
We'll upgrade soon.

jmrossy (approved):

PR #1685, opened 2019-11-13
Update latest proxy changes from celo-blockchain

jcortejoso:
### Description
Update with the latest changes from https://github.com/celo-org/celo-blockchain/tree/sentry_v1. Particularly the rename from `sentry` to `proxy` and new parameters for the proxy options.

### Tested

Tested in a k8s deployment. No tested nothing related with terraform.

### Other changes

This PR considers that a new variable `VALIDATOR_ADDRESS` is added to `.env.proxy-${rid}`, where the value corresponds with the etherbase address of the `VALIDATOR-${RID}`
Also need to update the `CELOTOOL_DOCKER_IMAGE_TAG` with a valid commit with the changes of this branch.


tkporter (approved):
Thank you!

Issue #1702, opened 2019-11-13
Developers SBAT read % of C Labs attestation requests completed by a validator


clrblmt:
### Expected Behavior

Read on-chain data to calculate % of C Lab attestations requests completed during TGCSO. @nambrot can help with read details. 

Post % to leaderboard metrics sheet. This number will be used to calculate final BTU multiplier boost for TGCSO. Calculations in sheet. 

*TGCSO = The Great Celo Stake Off

### Current Behavior

Cannot read % of attestations completed by a validator.

mrsmkl:
Currently, blockscout stores the requested and fulfilled attestations to the account table, and they can be seen from the Celo info page.
They are not currently  reflected in the leaderboard score in real time, I was thinking that the multipliers would be added after the challenge is over. This can be done by making a DB query that will update the values of `attestation_multiplier` for competitors. But now that we have the data, it should be easy enough to do it however we want to.

PR #1706, opened 2019-11-14
Change the event we emit when selecting issuers

### LINKED ISSUE #1694: Attestations contract should emit events to track validator performance
### Expected Behavior

### Current Behavior

What is the current behavior?

nambrot:
### Description

To accomodate https://github.com/celo-org/celo-monorepo/issues/1702, we are changing the event we emit in `selectIssuers` so that we can keep track of attestations involving validators. This will slightly increase the gas usage of the tx.

### Tested

- Unit tests

### Related issues

- Fixes #1694 

### Backwards compatibility

- Yes


packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -98,7 +98,7 @@ contract Attestations is
   event AttestationIssuersSelected(
     bytes32 indexed identifier,
     address indexed account,
-    uint256 attestationsRequested,
+    address indexed issuer,
```
asaj, line 101: nit: change event name to the singular, i.e. `AttestationIssuerSelected`

asaj (approved):

PR #1707, opened 2019-11-14
Quickload for Homepage on low connectivity

aaronmgdr:
### Description

The Current Homepage of Celo.org has a large video that is loaded and played inline this does not work well for much of celos target areas of use. This Changes the behavior to only load the video if we can determine that the network connection is good and the device memory is not terrible, otherwise it keeps the new still SVG graphic in place.

SERVE

### Tested

on Firefox and chrome so far.  while a few chrome only APIs are used there are fallbacks which mean the experience will not be hampered 

### Other changes

This also fixes an issue with Blue Banner where it would cut off long text on mobile by dynamically getting the real height of the banner and setting the Header to be offset by that amount rather than by a constant amount. 
### Related issues

- Fixes


packages/web/src/home/version3/HomeCover.tsx
```diff
@@ -165,99 +201,11 @@ const styles = StyleSheet.create({
...
     flexDirection: 'column',
     alignItems: 'stretch',
     justifyContent: 'center',
     maxWidth: '100vw',
   },
-  letsMake: {
-    textAlign: 'center',
-    zIndex: 1,
-  },
-  mask: {
```
aaronmgdr, line 186: this was unused code that was duplicated in TextAnimation

packages/web/package.json
```diff
@@ -49,6 +49,7 @@
     "memoize-one": "^5.1.1",
     "mobile-detect": "^1.4.2",
     "net": "^1.0.2",
+    "network-speed": "^2.0.3",
```
aaronmgdr, line 55: less than a K
https://bundlephobia.com/result?p=network-speed@2.0.4

cmcewen (approved):

aaronmgdr:
lets ship it

PR #1708, opened 2019-11-14
[Wallet] Prompt user to restart app when turning off Forno a second time

### LINKED ISSUE #1422: “Failed to sync” screen is shown after 2 mins when in Geth sync mode
**Frequency:** 60%
**App version:** Staging build v1.5.0 (103645729), (101548577)
**Repro on:**  Samsung Galaxy A8+ (8.0.0), LG nexus 5 (6.0.1) 
 
**Repro Steps:** 
1) Launch the app. 
2) Tap on the ‘cog’ icon. 
3) Tap on ‘Celo Lite’ option. 
4) Turn off celo lite toggle. 
5) Proceed for payment by selecting any celo user from the + tab. 
6) Observe. 

**Investigation:** [Random] Error “Could not calculate fee” is shown while user proceeding for payment after turning off the celo lite toggle. 
**Current Behavior:** Sorry something went wrong error message is shown. 
**Expected Behavior:** User should be able to send payment without getting any error after turning off the toggle 
**Attachment:**[ Android_SomthingWentWrong.mp4](https://drive.google.com/file/d/1yIcYfDIqGlq8M2_Hf5m666Q41C9VUkci/view?usp=sharing) (Bug observe at 1:02 sec)

### LINKED ISSUE #1745: When on Celo Lite mode, should not see 'Connecting to Celo...'
### Expected Behavior

When Lite mode is enabled, would expect to see the 'Bad Connection' banner(if there is no internet connection) or no banner(if there is a connection)

### Current Behavior

when enabling celo-lite, the 'Connecting to Celo...' banner is displayed

cc @annakaz

annakaz:
### Description

Currently, when users start geth twice in one app session (for example, Forno -> geth -> Forno -> geth, or geth -> Forno -> geth), the app fails to switch to geth the second time. Now, the app shows a modal prompting users to restart the app when they are starting geth for the second time in a session. When the app restarts geth works fine. 

For some reason starting geth a second time in a single app session leads to a minimum gas price error, because it cannot find the Gold Token address as geth does not think it is deployed. Had done some geth debugging with help from Kevin but no clear solution, so adding this modal as a workaround for now.

Also fixes an issue where switching on Forno leads to the app thinking it's disconnected until the app is restarted. This was not caught before because the disconnect behavior was conflated with long block times. 

### Tested

On pilot: 
Started in Forno, sent a transaction, turned it off, turned it on, turned it off, modal pops up- restarted app and sent transaction in geth mode. 

### Other changes

Remove some unused icons
Added snapshot tests for GethAwareButton, FeeExchangeIcon

### Related issues

- Fixes #1422 
- Fixes https://github.com/celo-org/celo-monorepo/issues/1745

### Backwards compatibility

Yes


packages/mobile/src/geth/saga.ts
```diff
@@ -43,6 +43,7 @@ export function* waitForGethConnectivity() {
 
 function* waitForGethInstance() {
   const zeroSyncMode = yield select(zeroSyncSelector)
+  Logger.error(TAG, `zeroSync mode: ${zeroSyncMode}`)
```
jmrossy: why `.error` here?

packages/mobile/src/web3/saga.ts
```diff
@@ -372,7 +376,7 @@ export function* addAccountToWeb3Keystore(key: string, currentAccount: string, p
     )
     yield put(setAccountInWeb3Keystore(account))
   } catch (e) {
-    Logger.debug(TAG + '@addAccountToWeb3Keystore', 'Failed to import raw key')
+    Logger.debug(TAG + '@addAccountToWeb3Keystore', `Failed to import raw key: ${e}`)
```
jmrossy: Should we use `.error` here which can take `e` as the 3rd param? It formats the exception messages nicely

packages/mobile/src/web3/saga.ts
```diff
@@ -417,17 +421,29 @@ export function* ensureAccountInWeb3Keystore() {
 }
 
 export function* switchToGethFromZeroSync() {
+  Logger.debug(TAG + 'Switching to geth from zeroSync..')
```
jmrossy: , not +

packages/mobile/src/web3/saga.ts
```diff
@@ -417,17 +421,29 @@ export function* ensureAccountInWeb3Keystore() {
...
     yield call(ensureAccountInWeb3Keystore)
 
-    // Ensure web3 is fully synced using new provider
-    yield call(waitForWeb3Sync)
+    if (gethAlreadyStartedThisSession) {
+      // If geth is started twice within the same session,
+      // there is an issue where it cannot find deployed contracts.
+      // Restarting the app fixes this issue.
+      yield call(navigate, Screens.WalletHome)
+      Logger.debug(TAG + 'Showed message')
```
jmrossy: ditto

packages/mobile/src/account/CeloLite.tsx
```diff
@@ -1,62 +1,139 @@
...
+  showModal = () => {
+    this.setState({ modalVisible: true })
+  }
+
+  hideModal = () => {
+    this.setState({ modalVisible: false })
+  }
+
+  onPressRestartModal = () => {
+    this.props.toggleZeroSyncMode(false)
```
jmrossy, line 60: The naming has me a bit confused here. Any reason we dispatch this action here instead of restarting the app as the name says? I see below that the restart logic is in the saga instead, any reason we don't have it here?

jmrossy (commented):
Mostly looks good! Just a few small comments

packages/mobile/src/geth/saga.ts
annakaz: Removed this log 👍

packages/mobile/src/account/CeloLite.tsx
annakaz: Good point. I updated the name to `onPressToggleWithRestartModal`. 

The restart logic is in the saga because we have to wait for geth to be turned on before restarting the app. Putting the restart logic here would require dispatching an action after the syncMode is done switching, and listening for that to know when to restart.

annakaz:
Added snapshot tests for some unrelated components (GethAwareButton, FeeExchangeIcon) to increase code coverage

packages/mobile/src/account/CeloLite.tsx
jmrossy: 👍

jmrossy (approved):

PR #1714, opened 2019-11-14
Deploy new version of Celostats and minnor change from celo-blockchain

jcortejoso:
### Description

This PR includes the changes to deploy the latest changes over celostats, particularly the ones from https://github.com/celo-org/ethstats-server/pull/6. These changes include multiple UX additions and improvements, some performance tweaks, and particularly important the addition of the lists for authorized and banned nodes to submit to ethstats.
Also 

### Tested

Tested on a testnet (berlintestnet001), using celotool+helm chart.

### Other changes
Included some changes (removing `--miner.verificationpool` option) from https://github.com/celo-org/celo-blockchain/pull/572.
I removed the the references in celotool to this option (for helm and terraform) but I can have missed something.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility
Currently we need to complete the `.env` variable `ETHSTATS_TRUSTED_ENODES` with the enode address of the authorized nodes. I think we plan to change this to use the etherbase account of the authorized nodes, but not sure.
The enode address for the nodes deployed in a testnet can be obtained using:
```
namespace="mytestnet"
for pod in $(kubectl get pods -n $namespace -l 'component in (tx_nodes, validators)' --template '{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}'); do
  kubectl exec -n $namespace -it $pod -c geth -- geth --exec "admin.nodeInfo['enode'].split('//')[1].split('@')[0]" attach | tr -d '"'
done
```
I can move this logic to celotool if we plan to use the enodes for authorizing for longer time.


.env
```diff
@@ -12,22 +12,24 @@ CLUSTER_DOMAIN_NAME="celo-networks-dev"
...
+BLOCKSCOUT_DOCKER_IMAGE_TAG="8be64adeccbb8f67dea3b6653600021577e544c5"
 BLOCKSCOUT_WEB_REPLICAS=3
 BLOCKSCOUT_DB_SUFFIX=
 
 ETHSTATS_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/ethstats"
-ETHSTATS_DOCKER_IMAGE_TAG="0ffe524c625ea59e4492dc92c2e638689c36e4b0"
+ETHSTATS_DOCKER_IMAGE_TAG="49fe674cb57f5d11174f4608f7469b03a69ebea5"
+ETHSTATS_TRUSTED_ENODES="ec86feaa5738d806cbe5431cfef97ce147c6ca4efa815f5684008c2f9455d8d39a8540e3fa19c151b3c9f5db89e86cb2565a05a4dbfb2d13f30834a01297e964,0eec0157fbe47a02899d5a5acb2c7309a9ffe4a02e97dcf841e9b809fb7608fc7b1aecf58ee4a8ce0b6ff26e3b7343b7373f553fc83b43596f8f58911afef9ba,b81a149fc0dbf96e9b5e36554896afabe442d318e1861d949e11fb9cbd05ae67236dc08904601a4033cb66e253f37fa335c6fa3d0ff2aa247b3134a904a90df4,cefb1c211db59c17d587cf7f14cb62bb7e46693f44e5d571d8224cdee72819ed22b1a28b13a64efab3a97493669adcae69b8cc93e52c9a8202d3a9fedfbe4a9e,ac1b214fcbb1ab6374ed8a17380199aea3e1088afbb67bc1cd5ee2e7f5b986f3255f8531ae01d5dbf0c248fbe4137204b180c81a4436ee4ac808a659b5621cc7,59cbf3ef26eb6e1c82694c22cf8f10657d30f746eefc3ae3539f3f938b8894fcdce062bc3b56b2ec996d9f603574bfe6396918a40bdf7b74a1fa3c10e4342394"
```
tkporter: Rather than putting these in the .env files I think it'd be best to generate these inside celotool and without directly pulling them from kubernetes pods so that this would work for VM testnets as well.

So when we set the helm parameters for ethstats, we could generate the enodes of all `envVar.TX_NODES` and `envVar.VALIDATORS` similar to how we generate the bootnode enode here: https://github.com/celo-org/celo-monorepo/blob/master/packages/celotool/src/lib/vm-testnet-utils.ts#L390

We generate the keys in Kubernetes & VM testnets the same way (ex: https://github.com/celo-org/celo-monorepo/blob/master/packages/helm-charts/testnet/templates/validators.statefulset.yaml#L61) so this would work for both types of testnets

tkporter (commented):
Thanks for cleaning up verification pool url! Could you also remove it from the .env files?

packages/celotool/src/lib/ethstats.ts
```diff
@@ -29,13 +30,26 @@ export async function upgradeHelmChart(celoEnv: string) {
...
+  const mnemonic = fetchEnv(envVar.MNEMONIC)
+  const publicKeys = []
+  const txNodes = parseInt(fetchEnv(envVar.TX_NODES), 0)
+  const validatorNodes = parseInt(fetchEnv(envVar.VALIDATORS), 0)
+  publicKeys.push(getAddressesFor(AccountType.TX_NODE, mnemonic, txNodes))
+  publicKeys.push(getAddressesFor(AccountType.VALIDATOR, mnemonic, validatorNodes))
+
+  publicKeys.push(fetchEnv(envVar.ETHSTATS_TRUSTED_ENODES).split(','))
+  return publicKeys.reduce((accumulator, value) => accumulator.concat(value), [])
+}
```
jcortejoso: @tkporter I adapted to the latest changes Dimi is finishing for using etherbase accounts. Should be ready soon (need to update the commits here). Meanwhile do you think this way is ok for getting the new addresses? It is concatenated the addresses from `getAddressesFor` with the ones from the env `ETHSTATS_TRUSTED_ENODES` (this one for external nodes submitting to celostats).

packages/helm-charts/ethstats/templates/ethstats.deployment.yaml
```diff
@@ -26,19 +26,40 @@ spec:
       - name: ethstats
         image: {{ .Values.ethstats.image.repository }}:{{ .Values.ethstats.image.tag }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
+        command:
+        - /bin/sh
+        - -c
+        args:
+        - |
+          cat <<'EOF' > /celostats-server/lib/utils/config.js
```
tkporter, line 34: This feels a little weird to generate a javascript file here, what do you think about either passing in the addresses as env variables here or maybe putting this into a .json file that's read by a .js file in ethstats?

diminator:
@jcortejoso The latest https://github.com/celo-org/celo-blockchain/pull/543 uses addresses `0xc0FFee` rather than `enode://`

packages/helm-charts/ethstats/templates/ethstats.deployment.yaml
timmoreton: raised as https://github.com/celo-org/celo-monorepo/issues/1837

timmoreton (approved):
Merging -- please come back

PR #1716, opened 2019-11-14
Reset isRedeemingInvite on rehydrate

cmcewen:
### Description

If the app is force quite while redeeming an invite, it will open again with the spinner spinning but no way to paste an invite code.

### Tested

Restarted the app during an invite redeem


i1skn (approved):
LGTM 👍

Issue #981, opened 2019-09-13
CLabs SBAT operate a single oracle with Brownian motion


timmoreton:
- [x] resurrect/build an Oracle that submits prices 
- [x] compute private CSV of prices 
- [x] in migrations, fund Oracle account and add its address to the whitelist of approved oracles
- [x] set it up to be deployed as a cronjob (talk to Trevor for what to do about VM testnets)

PR #1724, opened 2019-11-14
[cli] Add local-signing capability + commands for oracle reporting

yerdua:
### Description

This includes commands to report and get rates for a given token.

It also makes it possible to add private keys to the cli in order
to do local signing of transactions. Two things were needed for
this:
  1. Storing the private key info somewhere, since the contractkit
     instance is new for each running of a command.
  2. Closing the connection properly when the web3 provider has
     been wrapped with a CeloProvider (this is what allows multiple
     keys)

### Tested

Tested manually on a testnet called `yerduaoracletest`:

- ran `yarn celocli localaccounts:add ~/.oracle_key` to make it possible to sign transactions by the oracle account
- ran `yarn celocli oracle:report StableToken --numerator 1.08 --from $ORACLEACCOUNT` and saw that the transaction succeeded
- ran `yarn celocli oracle:rates StableToken` and saw the value reported above in the output

### Related issues

- Part of #981 

### Backwards compatibility

Functionality is added, not changed or removed, so it's backwards-compatible.


packages/cli/src/base.ts
```diff
@@ -46,6 +47,7 @@ export abstract class BaseCommand extends Command {
     if (this.requireSynced) {
       await requireNodeIsSynced(this.web3)
     }
+    addKeysToKit(this.kit, this.config.configDir)
```
mcortesi: kit and web3 are lazy variables, they are only created if somebody access them. But adding this line, we automatically add them to all commands

i think is better to add this to `get kit()`

packages/cli/src/commands/oracle/report.ts
```diff
@@ -0,0 +1,64 @@
...
+      const multiplier = new BigNumber(10).pow(numerator.decimalPlaces()).toNumber()
+      numerator = numerator.multipliedBy(multiplier)
+      denominator = denominator.multipliedBy(multiplier)
+    }
+
+    await displaySendTx(
+      'sortedOracles.report',
+      await sortedOracles.report(
+        res.args.token,
+        numerator.toNumber(),
```
mcortesi, line 56: this should be `toFixed()`. same on the other line... we could lose precision with toNumber

packages/cli/src/commands/oracle/report.ts
```diff
@@ -0,0 +1,64 @@
...
+        denominator.toNumber(),
+        res.flags.from
+      )
+    )
+    this.log(
+      `Reported oracle value of ${numerator.div(denominator).toNumber()} ${
```
mcortesi: toFixed

packages/cli/src/commands/oracle/report.ts
```diff
@@ -0,0 +1,64 @@
...
+import { BaseCommand } from '../../base'
+import { displaySendTx } from '../../utils/cli'
+import { Flags } from '../../utils/command'
+
+export default class ReportPrice extends BaseCommand {
+  static description =
+    'Report the price of Celo Gold in a specified token (currently just Celo Dollar, aka: "StableToken")'
+
+  static args = [
+    {
```
mcortesi, line 13: unsure if we need this option know... since there's only one token.. .maybe make it optional, and use a default value

mcortesi (changes requested):
The code assumes that the user will have their private key as file stored in his file system without any kind of encryption; which is unsafe.

So, two things:

1) why do we need local signing? can't we just assume that the node owns that account already?
2) if we need local signing, i think we should try use geth key-storage. I don't know mucho of the details here, so my suggestion is to avoid local signing all together.

yerdua:
> The code assumes that the user will have their private key as file stored in his file system without any kind of encryption; which is unsafe.

Good point! I felt pretty uncertain about how to handle this safely, so I appreciate this callout.

> 
> So, two things:
> 
> 1. why do we need local signing? can't we just assume that the node owns that account already?

The idea in my larger task (to have an oracle running for baklava), is to have a lightweight kubernetes cronjob that doesn't run a node itself, and instead sends signed transactions to the FORNO node. The container run by the cronjob would just have the npm version of `celocli`, the value to report, and the key for the oracle.

> 2. if we need local signing, i think we should try use geth key-storage. I don't know mucho of the details here, so my suggestion is to avoid local signing all together.

I don't know anything about geth key-storage, but just given the name, I assume that wouldn't be an accessible thing when all you have in a container is the cli. What about having the private key be an optional argument and not adding it permanently to the cli's config?

I guess this brings up a larger question of what exactly the CLI is for. Do we _want_ it to include the ability to send transactions to any publicly accessible node, or only the only the user themself is running? Personally, I lean toward the first option. 

That said, if this is too complex a question for now, I can put this functionality in celotool instead.

mcortesi:
> > The code assumes that the user will have their private key as file stored in his file system without any kind of encryption; which is unsafe.
> 
> Good point! I felt pretty uncertain about how to handle this safely, so I appreciate this callout.
> 
> > So, two things:
> > 
> > 1. why do we need local signing? can't we just assume that the node owns that account already?
> 
> The idea in my larger task (to have an oracle running for baklava), is to have a lightweight kubernetes cronjob that doesn't run a node itself, and instead sends signed transactions to the FORNO node. The container run by the cronjob would just have the npm version of `celocli`, the value to report, and the key for the oracle.
> 
> > 1. if we need local signing, i think we should try use geth key-storage. I don't know mucho of the details here, so my suggestion is to avoid local signing all together.
> 
> I don't know anything about geth key-storage, but just given the name, I assume that wouldn't be an accessible thing when all you have in a container is the cli. What about having the private key be an optional argument and not adding it permanently to the cli's config?
> 
> I guess this brings up a larger question of what exactly the CLI is for. Do we _want_ it to include the ability to send transactions to any publicly accessible node, or only the only the user themself is running? Personally, I lean toward the first option.
> 
> That said, if this is too complex a question for now, I can put this functionality in celotool instead.

My two cents:

I think ideally is nice for the cli to support local signing tx; but we need to have a secure way of handling the private keys, and that seems though. That's why i would search for other options first.

Something that might work is using environment variables to pass information. Again, probably not ideal, but that's what we do in other places. And for servers (k8s containters) we usually pass that information with env variables at deployment time.

packages/cli/src/commands/oracle/report.ts
yerdua: I can't use `toFixed()` here without also changing `contractkit`. The function in the wrapper expects the type of `numerator` and `denominator` to be numbers, and specifically, integers. Lines 46-49 do some processing of the initial values to convert a float value into a relationship between integers.

For example, if the command line args are: `--numerator 1.081 --denominator 1`, they'll become `numerator = 1081`, `denominator = 1000`.

Given this, do you still think there's risk of loss of precision?

mcortesi: oh, nevermind then. maybe put a TODO comment

mcortesi (approved):

PR #1731, opened 2019-11-14
Make styling more consistent in validator quick start and add password prompt to account:unlock

nategraf:
### Description

I ran through the Validator getting start doc and here are some edits, to the doc itself and the CLI, I thought might make it a little nicer.

Most important here is that I've added a password prompt option to the `account:unlock` command. Asking our users to enter their password on the command line breaks security best practices as it will be accessible in `~/.bash_history`

Additionally I added a verbose flag to the `node:synced` command to give more info on the sync progress.

### Tested

I used the new command options to run through the guide

### Backwards compatibility

100%


tkporter (approved):
🤤

PR #1734, opened 2019-11-14
Fix flaky end-to-end transfer, protocol unit tests

asaj:
### Description

This PR fixes three flaky tests:

- The end-to-end transfer test, in which we do not always estimate gas correctly for setting the inflation params, causing our transaction to revert and the tests to fail.
- The epoch rewards unit tests, in which we have tests that depend on block time but we are not traveling precisely to the correct block time
- The stabletoken unit tests, which is checking the wrong value in one of the events 

### Tested

Tests pass

### Other changes

- Minor typo


packages/protocol/test/governance/epochrewards.ts
```diff
@@ -396,16 +396,23 @@ contract('EpochRewards', (accounts: string[]) => {
...
     beforeEach(async () => {
-      await timeTravel(timeDelta.toNumber(), web3)
       targetEpochReward = await epochRewards.getTargetEpochRewards()
       targetEpochReward = targetEpochReward.plus(
         await epochRewards.getTargetTotalEpochPaymentsInGold()
       )
     })
 
+    const timeTravelToDelta = async () => {
+      const currentTime = new BigNumber((await web3.eth.getBlock('latest')).timestamp)
```
nambrot, line 406: Is time that large that we require `BigNumber`?

nambrot (approved):

PR #1736, opened 2019-11-15
Update genesis block to include round in aggregated signature

asaj:
### Description

This PR updates the genesis block in accordance with https://github.com/celo-org/celo-blockchain/pull/599

### Tested

End-to-end tests.

### Other changes

None

### Related issues

- Fixes #https://github.com/celo-org/celo-blockchain/issues/595

### Backwards compatibility

Not backwards compatible

### DIFF
--- a/packages/celotool/src/lib/generate_utils.ts
+++ b/packages/celotool/src/lib/generate_utils.ts
@@ -158,22 +158,37 @@ export const generateGenesisFromEnv = (enablePetersburg: boolean = true) => {
 const generateIstanbulExtraData = (validators: Validator[]) => {
   const istanbulVanity = 32
   const blsSignatureVanity = 192
-
   return (
     '0x' +
     repeat('0', istanbulVanity * 2) +
     rlp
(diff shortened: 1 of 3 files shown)


packages/celotool/src/lib/generate_utils.ts
```diff
@@ -158,22 +158,37 @@ export const generateGenesisFromEnv = (enablePetersburg: boolean = true) => {
...
   const istanbulVanity = 32
   const blsSignatureVanity = 192
-
   return (
     '0x' +
     repeat('0', istanbulVanity * 2) +
     rlp
       // @ts-ignore
       .encode([
+        // Added validators
```
nambrot, line 167: Love the comments

nambrot (approved):

packages/celotool/src/lib/generate_utils.ts
kobigurk: Came here to say the same!

Issue #1625, opened 2019-11-07
Validators should be able to use Twilio with the attestation service


nambrot:
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

PR #1738, opened 2019-11-15
Support blacklisting country codes for the attestation service

### LINKED ISSUE #1446: Validators SBAT whitelist/blacklist locales by text provider they dont want to serve
### Expected Behavior

Some locales might be prohibitive to serve as a text provider due to cost or other reason and thus should be blacklist/whitelist-able

### Current Behavior

We sent a SMS regardless

nambrot:
### Description

The primary objective of this PR is to support #1446, to allow validators to blacklist locales they do not want to serve, primarily for cost reasons if some locales are too prohibitive for them to serve. 

### Tested

- Against `namoffchainreveal`

### Other changes

- This PR refactors to be able to support #1625 soon
- It also removes the assumption that the attestation key is the same as the account key
- It handles concurrency better by properly locking onto the row in the table to prevent multiple text messages being sent by two requests at the same time.

### Related issues

- Fixes #1446 

### Backwards compatibility

- It is not backwards compatible with deployed attestation services, but since those are non-existent this should be fine


packages/attestation-service/README.md
```diff
@@ -0,0 +1,33 @@
+# Attestation Service
+
+This is the attestation service that validators run on the Celo network to provide attestations of phone numbers to accounts.
```
jmrossy: Nit, consider:
`A service run by validators on the Celo network to send SMS messages, enabling attestations of user phone numbers.`

packages/attestation-service/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Configuration
+
+You can use the following environment variables to configure the attestation service:
+
+- `DATABASE_URL` - The URL under which your database is accessible, currently supported are `postgres://`, `mysql://` and `sqlite://`
+- `CELO_PROVIDER` - The URL under which a celo blockchain node is reachable, i.e. something like `https://integration-forno.celo-testnet.org`
+- `ACCOUNT_ADDRESS` - The address of the account on the `Accounts` smart contract
```
jmrossy, line 11: Can we get this from contract kit?

packages/attestation-service/README.md
```diff
@@ -0,0 +1,33 @@
...
+- `APP_SIGNATURE` - The hash with which clients can auto-read SMS messages on android
+- `SMS_PROVIDERS` - A comma-separated list of providers you want to configure, we currently support:
+
+`nexmo`
+
+- `NEXMO_KEY` - The API key to the Nexmo API
+- `NEXMO_SECRET` - The API secret to the Nexmo API
+- `NEXMO_BLACKLIST` - A comma-sperated list of country codes you do not want to serve
+
+### Contributing
```
jmrossy: Usually the Contributing sections has details of how to contribute to the project, not run the service. 
Consider renaming to `Running Locally` or `Development`

packages/attestation-service/src/attestation.ts
```diff
@@ -87,6 +119,107 @@ function createAttestationTextMessage(attestationCode: string) {
...
+  attestationCode: string
+) {
+  const textMessage = createAttestationTextMessage(attestationCode)
+  let attestationRecord: AttestationModel | null = null
+  const AttestationTable = await getAttestationTable()
+
+  const transaction = await sequelize!.transaction()
+
+  try {
+    await AttestationTable.findOrCreate({
```
jmrossy: Big Function. Can this DB interaction be moved out?

packages/attestation-service/src/db.ts
```diff
@@ -15,34 +15,36 @@ export function initializeDB() {
 
 export let kit: ContractKit
 
-export function initializeKit() {
+export async function initializeKit() {
   if (kit === undefined) {
     kit = newKit(fetchEnv('CELO_PROVIDER'))
+    const blockNumber = await kit.web3.eth.getBlockNumber()
+    if (blockNumber === 0) {
```
jmrossy, line 22: Is it sufficient to know the blockNumber isn't 0 or should we wait for web3 to be synced?

packages/attestation-service/src/sms/nexmo.ts
```diff
@@ -0,0 +1,102 @@
...
+        if (err) {
+          reject(err)
+        } else {
+          resolve(responseData.numbers)
+        }
+      })
+    })
+  }
+
+  private getMatchingNumber = (countryCode: string) => {
```
jmrossy, line 95: Does Nexmo have an intelligent service to choose the best number for you already? Twillio does that

packages/utils/src/address.ts
```diff
@@ -19,5 +25,7 @@ export const publicKeyToAddress = (publicKey: string) => {
 export const privateKeyToPublicKey = (privateKey: string) => {
   return '0x' + privateToPublic(Buffer.from(privateKey.slice(2), 'hex')).toString('hex')
 }
+export const isValidPrivateKey = (privateKey: string) =>
+  isValidPrivate(Buffer.from(privateKey.slice(2), 'hex'))
```
jmrossy: Would be nice if this and the functions above it only did the slice after confirming the leading chars are '0x'
So happens @yorhodes added some util methods in this file to help with just that:
https://github.com/celo-org/celo-monorepo/pull/1570#discussion_r346573047

packages/attestation-service/README.md
nambrot: I can see how the phrasing is confusing. It's not the address of the `Accounts` smart contract, but the account address of the validator account on the `Accounts` smart contract. Not sure what a better name/description would be

packages/attestation-service/src/attestation.ts
nambrot: Great idea!

packages/attestation-service/src/db.ts
nambrot: We should have proper waiting for sync, raised https://github.com/celo-org/celo-monorepo/issues/1742 for it

packages/attestation-service/src/sms/nexmo.ts
nambrot: Nexmo does not seem to have that support https://developer.nexmo.com/messaging/sms/guides/custom-sender-id

packages/utils/src/address.ts
nambrot: I added a check that the privateKey starts with `0x` as `isValidPrivate` will then do the rest

jmrossy (approved):

PR #1739, opened 2019-11-15
Remove end-to-end attestations test from circle

asaj:
### Description

This test is broken and never passes, removing it so that we don't become desensitized to broken builds.

### Tested

Not tested.

### Other changes

None.


nambrot (approved):
I don't think it always fails, but certainly enough to remove it! (looks about 50% right now)

PR #1749, opened 2019-11-16
Add ability to partially re-lock pending withdrawals

### LINKED ISSUE #1373: Users can re-lock pending withdrawals
### Expected Behavior

Users can re-lock gold which has been unlocked. This can probably be done under the hood in the `lock` command.

e.g. If I have pending withdrawals of 40, 20, and a balance of 100, and I ask to lock 50 gold, I re-lock the pending withdrawal of 40 and then 10 gold.  Stretch goal to allow partial re-locks, so I can re-lock my pending withdrawal of 40 and then half of my pending withdrawal of 20.

### Current Behavior

Can't re-lock pending withdrawals

asaj:
### Description

This PR adds the ability to specify how much of pending withdrawal to re-lock, and updates the CLI to lock pending withdrawals before unlocked gold.

### Tested

Unit tests

### Other changes

- Move account:lock to lockedgold:lock

### Related issues

- Fixes #1373


packages/cli/src/commands/lockedgold/lock.ts
```diff
@@ -28,15 +28,20 @@ export default class Lock extends BaseCommand {
...
     await newCheckBuilder(this)
       .addCheck(`Value [${value.toString()}] is >= 0`, () => value.gt(0))
       .isAccount(address)
-      .hasEnoughGold(address, value)
+      .hasEnoughGold(address, lockValue)
       .runChecks()
 
-    const lockedGold = await this.kit.contracts.getLockedGold()
+    const txos = await lockedGold.relock(address, relockValue)
+    await Promise.all(txos.map((txo) => displaySendTx('relock', txo, { from: address })))
```
nambrot: @mcortesi would want you to use `concurrentMap` here to constraint the concurrency

packages/contractkit/src/wrappers/LockedGold.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  let lockedGold: LockedGoldWrapper
+
+  // Arbitrary value.
+  const value = 120938732980
+  let account: string
+  beforeAll(async () => {
+    account = (await web3.eth.getAccounts())[0]
+    lockedGold = await kit.contracts.getLockedGold()
+    accounts = await kit.contracts.getAccounts()
+    await accounts.createAccount().sendAndWaitForReceipt({ from: account })
```
nambrot: You can set the `defaultAccount` on `kit` to avoid having to specify `from: account` on all txs

nambrot (approved):

packages/cli/src/commands/lockedgold/lock.ts
mcortesi: In fact here i would advice to do things sequentually...
Not sure cli ouptut from `displaySendTx()` looks ok when doing multiple at the same time...

And i have my doubts about `next nonce` computation clashed when doing multiple/parallel txs... i suppose it's fine if the node is the signer (nonce setter)..

packages/contractkit/src/wrappers/LockedGold.ts
```diff
@@ -66,13 +66,61 @@ export class LockedGoldWrapper extends BaseWrapper<LockedGold> {
     tupleParser(parseNumber)
   )
 
+  async getPendingWithdrawalsTotalValue(account: Address) {
+    const pendingWithdrawals = await this.getPendingWithdrawals(account)
+    // Ensure there are enough pending withdrawals to relock.
+    const values = pendingWithdrawals.map((pw: PendingWithdrawal) => pw.value)
+    const reducer = (total: BigNumber, pw: BigNumber) => pw.plus(total)
+    return values.reduce(reducer)
```
mcortesi: what if pendindWithradrawals is empty? you need pass `0` as initial value to reduce

packages/contractkit/src/wrappers/LockedGold.ts
```diff
@@ -66,13 +66,61 @@ export class LockedGoldWrapper extends BaseWrapper<LockedGold> {
...
+      i: number
+    ) => {
+      const valueToRelock = BigNumber.minimum(pw.value, remainingToRelock)
+      if (valueToRelock.isZero()) {
+        remainingToRelock = remainingToRelock.minus(valueToRelock)
+        acc.push(this._relock(i, valueToRelock))
+      }
+      return acc
+    }
+    return Promise.all(pendingWithdrawals.reduceRight(relockPw, []) as Array<
```
mcortesi: you don't need promise.all here...  relockPw doesn't return a promise

asaj: Whoops

PR #1751, opened 2019-11-17
Change validator set during end-to-end sync test

### LINKED ISSUE #1697: End-to-end sync tests should change the validator set
### Expected Behavior

We change the validator set in the end-to-end sync tests to get more coverage of ultralight sync + validator set changes.

### Current Behavior

Static validator set

asaj:
### Description

This PR updates the end-to-end sync test to use a changing validator set. This is important as it tests that ultralight sync properly applies the validator set diff when validating headers. 

### Tested

These are tests.

Fixes #1697


packages/celotool/src/e2e-tests/sync_tests.ts
```diff
@@ -28,18 +36,54 @@ describe('sync tests', function(this: any) {
...
+    const validators = await (await validatorKit._web3Contracts.getValidators()).methods
+      .getRegisteredValidators()
+      .call()
+    const web3 = new Web3('ws://localhost:8557')
+    await sleep(10)
     await waitToFinishSyncing(web3)
+    const kit = newKitFromWeb3(web3)
+    // The validator set size at any one time will be 4, and we rotate two validators per epoch.
+    // This means that if a node were to have a static view of the validator set, we guarantee that
+    // the node will reject blocks.
```
jfoutts-celo: Maybe add reference or link to the celo-blockchain code that will reject the block due to differing validator set?

packages/celotool/src/e2e-tests/sync_tests.ts
```diff
@@ -28,18 +36,54 @@ describe('sync tests', function(this: any) {
...
+      rpcport: 8559,
+      privateKey: groupPrivateKey.slice(2),
       peers: [8545],
     }
     await initAndStartGeth(hooks.gethBinaryPath, fullInstance)
-    const web3 = new Web3('http://localhost:8553')
+    const validators = await (await validatorKit._web3Contracts.getValidators()).methods
+      .getRegisteredValidators()
+      .call()
+    const web3 = new Web3('ws://localhost:8557')
```
jfoutts-celo, line 65: Nice!  We should probably favor ws over http regardless of needing `eth.subscribe()`.

jfoutts-celo (approved):

packages/celotool/src/e2e-tests/sync_tests.ts
asaj: Totally agree, I'll create tech debt issue for this

asaj: https://github.com/celo-org/celo-monorepo/issues/2616

packages/celotool/src/e2e-tests/sync_tests.ts
asaj: Explicitly pointed out the function that would return false

PR #1752, opened 2019-11-17
Explicitly wait for nodes to sync in end-to-end tests

asaj:
### Description

This PR addresses a source of flakiness in which nodes would try to send transactions without being synced.


packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -41,6 +41,19 @@ const GENESIS_PATH = `${TEST_DIR}/genesis.json`
 const NetworkId = 1101
 const MonorepoRoot = resolvePath(joinPath(__dirname, '../..', '../..'))
 
+export async function waitToStartAndFinishSyncing(web3: any) {
+  while (!(await web3.eth.isSyncing())) {
+    await sleep(0.1)
+  }
```
nategraf: If this is called to wait on a node which is already finished syncing it would block forever. If it is valid to check that `await web3.eth.getBlockNumber() === 0` to see if the node had not started syncing then this functions could become `while ((await web3.eth.isSyncing()) || (await web3.eth.getBlockNumber() === 0)) { await sleep(0.1) }`

nategraf (approved):
Looks good overall 👍 I had one comment

PR #1753, opened 2019-11-17
Add @celo/dev-utils

nambrot:
### Description

This PR adds `@celo/dev-utils`, a `utils` package that is meant to be used as a `devDependency`. It's primary use case is to reuse the ganache setup currently present in `cli` and `contractkit` (and soon `attestation-service`). Due to the way `jest` uses `globalSetup`, depending packages will still need to define their own setup/teardown files.

### Tested

- CI


jmrossy:
Unassigning because CI tests failing

packages/contractkit/package.json
```diff
@@ -40,6 +40,7 @@
     "web3-utils": "1.0.0-beta.37"
   },
   "devDependencies": {
+    "@celo/dev-utils": "0.0.1",
```
jmrossy: Can you make this version `-dev` to keep with the new scheme @mcortesi just introduced?

packages/contractkit/package.json
```diff
@@ -20,7 +20,7 @@
     "clean:all": "yarn clean && rm -rf src/generated",
     "build:gen": "yarn --cwd ../protocol build",
     "prepublishOnly": "yarn build:gen && yarn build",
-    "test:prepare": "yarn --cwd ../protocol devchain generate .devchain --migration_override src/test-utils/migration-override.json",
+    "test:prepare": "yarn --cwd ../protocol devchain generate .devchain --migration_override ../dev-utils/src/migration-override.json",
```
jmrossy, line 23: @mcortesi Should this be `pretest` instead of `test:prepare`?

jmrossy (approved):
Looks good to me, just wondering about the version number

packages/contractkit/package.json
nambrot: Will do, though I imagine this is not a package we ever have to publish to npm i think?

packages/contractkit/package.json
nambrot: I believe this command is expected to run before every test which is why it is not named `pretest`

PR #1757, opened 2019-11-18
[Wallet] Implement new redeem invite design and add skip option

### LINKED ISSUE #1720: New design for redeem invite screen
Subtask, see #1523 for details

jmrossy:
### Description

https://www.figma.com/file/9lBnIzsABhO8oXE0d0c8Km/Onboarding-and-Verification?node-id=105%3A640

- Implement new design for enter invite screen:
- Generalize the CodeRow component to be shared btwn verification screen and invite screen
- Add option to skip invite redemption
- Fix small bug in KeyboardSpacer toggle callback
- Clean up strings in NuxNamePin.json a bit

### Tested

Did full onboarding flow on iOS and Android

### Related issues

- Fixes #1720 

### Backwards compatibility

Yes

wa-new-invite (1)


packages/mobile/src/components/CodeRow.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+function CodeRow({
+  status,
+  inputValue,
+  inputPlaceholder,
+  onInputChange,
+  shouldShowClipboard,
+  t,
+}: Props) {
+  if (status === CodeRowStatus.DISABLED) {
```
annakaz, line 40: Any reason not to use a switch statement?

packages/mobile/src/verify/VerificationCodeRow.tsx
```diff
@@ -1,82 +1,46 @@
...
-import colors from '@celo/react-components/styles/colors'
-import fontStyles from '@celo/react-components/styles/fonts'
-import { stripHexLeader } from '@celo/utils/src/address'
 import { extractAttestationCodeFromMessage } from '@celo/walletkit'
 import * as React from 'react'
-import { withNamespaces, WithNamespaces } from 'react-i18next'
-import { ActivityIndicator, StyleSheet, Text, View } from 'react-native'
-import i18n, { Namespaces } from 'src/i18n'
-import { ATTESTATION_CODE_PLACEHOLDER } from 'src/identity/reducer'
+import CodeRow, { CodeRowStatus } from 'src/components/CodeRow'
```
annakaz, line 4: 👨‍🎨  beautiful, thanks for standardizing this code row

packages/mobile/locales/en-US/global.json
```diff
@@ -94,5 +94,8 @@
   "calculateFeeFailed": "Could not calculate fee",
   "failedToSwitchSyncModes": "Failed to switch sync modes",
   "gold": "Gold",
-  "localCurrencyTitle": "Select Currency"
+  "localCurrencyTitle": "Select Currency",
+  "or": "or",
+  "accepted": "Accepted",
```
annakaz, line 99: Any reason we removed the 🎉?

packages/mobile/src/invite/saga.ts
```diff
@@ -259,6 +260,24 @@ export function* doRedeemInvite(inviteCode: string) {
...
 
+export function* skipInvite() {
+  yield take(Actions.SKIP_INVITE)
+  Logger.debug(TAG + '@skipInvite', 'Skip invite action taken')
+  try {
+    Logger.debug(TAG + '@skipInvite', 'Creating new account')
+    yield call(getOrCreateAccount)
+    Logger.debug(TAG + '@skipInvite', 'Marking nux flows as complete')
+    // TODO(Rossy): Remove when import screen is removed
+    yield put(denyImportContacts())
```
annakaz, line 271: Why do we have to deny importing the contacts? Couldn't a user who skipped invite still send funds to their contacts after receiving funds?

packages/mobile/locales/en-US/nuxNamePin1.json
```diff
@@ -20,33 +13,17 @@
...
-    "validating": "Great! Validating copied invite code",
-    "inviteAccepted": "🎉 Invite Accepted!",
-    "askForInvite": {
-      "0": "Request an invite from a friend on Celo or sign up for the {{testnet}} network at ",
-      "1": "celo.org/app"
-    }
+    "body":
+      "If you have an invite code, please enter it here. You can copy and paste the entire message.",
+    "codeHeader": "Invite Code",
+    "codePlaceholder": "Celo code: am9hBM3tiA+CuNb...",
```
annakaz, line 19: Thoughts on having this be "invite code" (or update text sent in invite script) to have it match the invite text? Or even "Hello! Thank you for joining the Celo network. ... " to make it clearer that people can paste the whole code? 

Currently, the invite text doesn't use the words "Celo code" at all: "Hello! Thank you for joining the Celo network. Your invite code is: d5MD3OBeocb2NZTVe068EcD7yy2Kpn/+kfqnDa7yvD8= Download the app at ... "

packages/mobile/src/invite/EnterInviteCode.tsx
```diff
@@ -144,90 +97,88 @@ export class EnterInviteCode extends React.Component<Props, State> {
     navigateToURI(CELO_FAUCET_LINK)
   }
 
+  onPressSkip = () => {
+    this.props.skipInvite()
+  }
+
+  onInputChange = (value: string) => {
+    this.setState({ inputValue: value })
+    const inviteCode = extractValidInviteCode(value)
```
annakaz, line 105: If a user pastes over the full invite text, we may want to show just the extracted invite code in the CodeRow

annakaz (approved):
LGTM, left a few comments

Also, note that I get this warning when loading the page: 
Screenshot_20191118-100005

packages/mobile/src/components/CodeRow.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+      const { toJSON } = render(
+        <CodeRow
+          status={status}
+          inputValue={'test'}
+          inputPlaceholder={'placeholder'}
+          onInputChange={jest.fn()}
+          shouldShowClipboard={jest.fn()}
+        />
+      )
+      expect(toJSON()).toMatchSnapshot()
```
jeanregisser, line 24: Nothing to act on right now, I think I've already mentioned that in the past but I'm not exactly convinced about the value of these snapshots tests.
It's good that it renders but it might as well render something completely unusable and still pass the test.

packages/mobile/src/invite/reducer.test.ts
```diff
@@ -1,25 +0,0 @@
...
-        isRedeemingInvite: false,
-        invitees: {
-          [mockAccount]: mockE164Number,
-        },
-        redeemedInviteCode: '',
-        redeemComplete: false,
-      })
-    })
-  })
-})
```
jeanregisser, line 25: I see the `storeInviteeData` action still exists, shouldn't we keep this test?

packages/mobile/src/invite/saga.ts
```diff
@@ -259,6 +260,24 @@ export function* doRedeemInvite(inviteCode: string) {
...
+    yield put(setHasSeenVerificationNux(true))
```
jeanregisser, line 272: Given we have a meaningful `SKIP_INVITE_SUCCESS` action, we could also act on it in the corresponding reducers and atomically set the right state we want for that. 
What do you think?

packages/mobile/src/verify/VerificationInputScreen.tsx
```diff
@@ -261,6 +249,7 @@ class VerificationInputScreen extends React.Component<Props, State> {
             </View>
           </View>
         </Modal>
+        <KeyboardSpacer onToggle={this.onKeyboardToggle} />
```
jeanregisser, line 252: Nice 👍

packages/mobile/locales/en-US/global.json
jmrossy: You'd have to ask @nityas and @coreycelo :)

jeanregisser (approved):
Very nice work! :+1:

packages/mobile/src/invite/saga.ts
jmrossy: Yes. the action name is a bit misleading here. It's the action that get's dispatched when a user skips the import contact screen. The action updates the state so we don't go back there on app restart. It doesn't prevent us from importing later (i.e. send screen)

packages/mobile/locales/en-US/nuxNamePin1.json
jmrossy: @nityas @coreycelo

packages/mobile/src/invite/EnterInviteCode.tsx
jmrossy: Good idea!

packages/mobile/src/invite/reducer.test.ts
jmrossy: This test was failing and I didn't consider it work fixing because reducers are so trivially simple.

packages/mobile/src/invite/saga.ts
jmrossy: You mean handle that action in the verification reducer? I'd rather not have reducers handle actions from other modules

packages/mobile/src/verify/VerificationInputScreen.tsx
jmrossy: Thanks for the suggestion!

packages/mobile/src/components/CodeRow.test.tsx
jmrossy: I somewhat agree but often find myself short of time and the more in depth tests take longer to write.

packages/mobile/src/components/CodeRow.tsx
jmrossy: None, I just don't like switch statements, particularly for cases which are multiple lines :)
They have more nesting than ifs

packages/mobile/src/invite/saga.ts
jeanregisser: Yes, I know this hasn't been the styled used sofar in the app but having meaningful actions and not just glorified getter/setters actions allows decoupling things as originally intended by Redux.

https://twitter.com/dan_abramov/status/682923564006248448

packages/mobile/locales/en-US/global.json
coreycelo: Largely a tone of voice call. The Invite Code flow has a reskin in the next pass which more simply uses a checkmark. A little odd to "celebrate" at this stage of onboarding.

nityas: we don't use emoji anywhere else in the app and it was strange to just have it here

packages/mobile/locales/en-US/nuxNamePin1.json
coreycelo: Agree with point around matching SMS and app copy. I know there was some intent to also change the SMS copy. @nityas, are we still thinking of doing that?

packages/mobile/src/invite/saga.ts
jmrossy: Open to doing this going forward then

PR #1763, opened 2019-11-18
Utilize new gateway fee structure

nategraf:
### Description

In https://github.com/celo-org/celo-blockchain/pull/602 the network will use a new gateway fee as described in https://github.com/celo-org/celo-blockchain/issues/541

This PR adds a new `gatewayFee` field to transaction objects, renames `gasFeeRecipient` to `gatewayFeeRecipient` and renames `gasCurrency` to `feeCurrency` and makes the appropriate logic changes to handle this new scheme.

Note: `walletkit` is untouched by this PR to maintain compatibility with deployed networks.

### Tested

The e2e transfer test has been updated to test the new behavior. All other e2e and unit tests are run as is.

### Related issues

- Related https://github.com/celo-org/celo-blockchain/issues/541

### Backwards compatibility

Oh most certainly not.


packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -464,34 +469,37 @@ describe('Transfer tests', function(this: any) {
...
+                describe('when specifying the gateway fee', () => {
+                  describe("when using a peer's etherbase as recipient", () =>
                     testTransferToken({
                       expectedGas: GOLD_TRANSACTION_GAS_COST,
                       transferToken: CeloContract.GoldToken,
                       feeToken: CeloContract.GoldToken,
                       txOptions: {
-                        gasFeeRecipient: FeeRecipientAddress,
+                        gatewayFeeRecipient: FeeRecipientAddress,
+                        gatewayFee: '0x10000',
```
asaj: We should probably have tests for most combinations of {fee recipient unset, fee recipient peer etherbase, fee recipient random address } x {gateway fee set, gateway fee set too low, gateway fee unset}

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -561,25 +581,19 @@ describe('Transfer tests', function(this: any) {
...
 
         describe('Transfer CeloDollars', () => {
-          describe('gasCurrency = CeloDollars >', () => {
+          describe('feeCurrency = CeloDollars >', () => {
             testTransferToken({
               expectedGas: 175303,
               transferToken: CeloContract.StableToken,
               feeToken: CeloContract.StableToken,
-              txOptions: {
-                gasFeeRecipient: FeeRecipientAddress,
```
asaj, line 568: Was this there for a reason? Or was this a mistake

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -696,8 +704,7 @@ describe('Transfer tests', function(this: any) {
               const expectedGasUsed = 164005
               txRes = await runTestTransaction(
                 await transferCeloGold(FromAddress, ToAddress, TransferAmount, {
-                  gasCurrency: stableTokenAddress,
-                  gasFeeRecipient: FeeRecipientAddress,
```
asaj, line 688: ditto, needed or mistake?

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -774,8 +780,7 @@ describe('Transfer tests', function(this: any) {
               txRes = await runTestTransaction(
                 await transferCeloGold(FromAddress, ToAddress, TransferAmount, {
                   gas,
-                  gasCurrency: await kit.registry.addressFor(CeloContract.StableToken),
-                  gasFeeRecipient: FeeRecipientAddress,
```
asaj: ditto

packages/contractkit/README.md
```diff
@@ -146,10 +146,11 @@ const goldTokenAddress = await kit.registry.addressFor(CeloContract.GoldToken)
 
 ### Sending Custom Transactions
 
-Celo transaction object is not the same as Ethereum's. There are two new fields present:
+Celo transaction object is not the same as Ethereum's. There are three new fields present:
 
-- gasCurrency (address of the ERC20 contract to use to pay for gas)
-- gasFeeRecipient (address of the beneficiary for the gas, the full node)
+- feeCurrency (address of the ERC20 contract to use to pay for gas)
+- gatewayFeeRecipient (address of the beneficiary for the gas, the full node)
```
asaj: Lines 151 and 152 need changes

packages/contractkit/src/providers/celo-private-keys-subprovider.ts
```diff
@@ -114,20 +119,29 @@ export class CeloPrivateKeysWalletProvider extends PrivateKeyWalletSubprovider {
...
 
-    if (isEmpty(txParams.gasFeeRecipient)) {
-      txParams.gasFeeRecipient = await this.getCoinbase()
-      if (isEmpty(txParams.gasFeeRecipient)) {
+    if (isEmpty(txParams.gatewayFeeRecipient)) {
+      txParams.gatewayFeeRecipient = await this.getCoinbase()
+      if (isEmpty(txParams.gatewayFeeRecipient)) {
         // Fail early. The validator nodes will reject a transaction missing
-        // gas fee recipient anyways.
+        // gateway fee recipient anyways.
```
asaj: Is this true? Won't they just not pay the gateway fee?

packages/docs/developer-resources/contractkit/examples.md
```diff
@@ -66,10 +66,11 @@ const goldTokenAddress = await kit.registry.addressFor(CeloContract.GoldToken)
 
 ## Sending Custom Transactions
 
-Celo transaction object is not the same as Ethereum's. There are two new fields present:
+Celo transaction object is not the same as Ethereum's. There are three new fields present:
 
-- gasCurrency (address of the ERC20 contract to use to pay for gas)
-- gasFeeRecipient (address of the beneficiary for the gas, the full node)
+- feeCurrency (address of the ERC20 contract to use to pay for gas)
+- gatewayFeeRecipient (address of the beneficiary for the gas, the full node)
```
asaj: lines 71 and 72 need updating

packages/protocol/migrationsConfig.js
```diff
@@ -13,7 +13,7 @@ const DefaultConfig = {
     selectIssuersWaitBlocks: 4,
   },
   blockchainParameters: {
-    gasForNonGoldCurrencies: 134000,
+    gasForNonGoldCurrencies: 166000,
```
asaj, line 16: Why 32000 more?

packages/protocol/contracts/common/FeeCurrencyWhitelist.sol
```diff
@@ -2,14 +2,14 @@ pragma solidity ^0.5.3;
 
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 
-import "./interfaces/IGasCurrencyWhitelist.sol";
+import "./interfaces/IFeeCurrencyWhitelist.sol";
```
asaj, line 5: nit: In places where it isn't obvious (e.g. tx params) should we call this `TxFeeCurrencyWhitelist`?

Don't feel strongly

asaj (commented):
Mostly lgtm, just some small nits

packages/contractkit/src/providers/celo-private-keys-subprovider.ts
nategraf: No, and I'm not sure if that was ever true. I've opted to remove the check, but let me know if you think I should keep it and just rewrite the comment instead.

packages/protocol/migrationsConfig.js
nategraf: To match the default set as a part of https://github.com/celo-org/celo-blockchain/pull/602. The payment of the gateway fee adds an additional credit operation, which we have been estimating at 32000 gas.

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: It's intentional. My main motivation for this is that when we are in the `full` sync mode we want the gateway fee fields to be empty. I decided to delete that field because it allows the test to cover that the defaults should be assigned appropriately by the `light` and `ultralight` nodes.

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: See above.

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: See above.

packages/docs/developer-resources/contractkit/examples.md
nategraf: I've updated this.

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: I've added cases to cover all of those. Thanks for the suggestion!

.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ e2e-defaults: &e2e-defaults
   docker:
     - image: celohq/circleci
   environment:
-    CELO_BLOCKCHAIN_BRANCH_TO_TEST: master 
+    CELO_BLOCKCHAIN_BRANCH_TO_TEST: victor/gateway-fee
```
jmrossy, line 39: TODO revert?

jmrossy (dismissed):
I'm unfamiliar with the Gateway concept. Is it just a rebranding of gas?
RE wallet changes, if walletkit isn't updated too, this change will break transactions in the wallet. Let's make sure we have a rollout plan for this :)

nategraf:
Marking `do not merge` until https://github.com/celo-org/celo-blockchain/pull/602 is in

.circleci/config.yml
nategraf: My plan is to merge https://github.com/celo-org/celo-blockchain/pull/602, then revert this, then merge this PR. Thanks for making sure I didn't miss this.

nategraf:
@jmrossy I have some context in https://github.com/celo-org/celo-blockchain/issues/541. In short, the gateway fee replaces `GasFeeRecipient` as the mechanism to pay transaction serving full-nodes, decoupling full-node incentives from the gas system.

On the `walletkit` changes, it is my understanding that it can't be updated until we deploy a network with these changes. Is that accurate? Either way I should have reached out earlier, but would definitely appreciate some guidance on how to roll out that change.

packages/protocol/migrationsConfig.js
asaj: I don't think this is exactly correct, but only because I don't believe what we've been doing is exactly correct. So not in the scope of this PR

packages/celotool/src/e2e-tests/validator_order_tests.ts
```diff
@@ -42,6 +42,9 @@ describe('governance tests', () => {
 
     it('properly orders validators randomly', async function(this: any) {
       this.timeout(160000 /* 160 seconds */)
+      // If a concensus round fails during this test, the results are inconclusive.
```
asaj: nit: spelling

asaj (approved):
✅

packages/protocol/migrationsConfig.js
nategraf: Ya I was wondering why we don't actually meter the ERC20 calls and either charge only what is used, or keep the intrinsic gas, then issue a refund if its under. Is there a reason for that?

packages/protocol/contracts/common/FeeCurrencyWhitelist.sol
nategraf: I can see it both ways. I think my slight preference is toward `FeeCurrencyWhitelist`

asaj:
> I'm unfamiliar with the Gateway concept. Is it just a rebranding of gas?

Not exactly, more details here: https://github.com/celo-org/celo-blockchain/issues/541

asaj:
> RE wallet changes, if walletkit isn't updated too, this change will break transactions in the wallet. Let's make sure we have a rollout plan for this :)

Unfortunately, walletkit depends on deployed environments to build (currently it looks like integration). So we will need to deploy to integration to get this to work.

Any idea how much is left to migrate from walletkit to contractkit?

PR #1764, opened 2019-11-19
[Wallet] Fix Celo Lite crash for second toggle off in production builds

### LINKED ISSUE #1746: Toggling Celo Lite mode 'on' crashes app
### Expected Behavior

app shouldn't crash

### Current Behavior

Steps Taken:
1) start app up, skip verification.
2) go to settings > Celo Lite
3) leave device, do other things
4) re-unlock device, current screen is still Celo Lite
5) toggle Celo Lite mode on - App Crashes and prompts restart (connection error message)

Screenshot_20191115-134821

annakaz:
### Description

The production apk (unlike the local dev build) crashed when turning Celo Lite mode off a second time in one app session. This fixes the toggle by restarting the app before waiting for the web3 sync.

### Tested

Built an alfajores apk and a local build on a pixel 2. Confirmed in both that Celo Lite could be toggled on and off, and when a restart was necessary, that a modal pops up and restarts the app in the desired sync mode. 

### Related issues

- Fixes #1746 

### Backwards compatibility

Yes


packages/mobile/src/web3/saga.ts
```diff
@@ -434,14 +451,9 @@ export function* switchToGethFromZeroSync() {
...
-
-    if (gethAlreadyStartedThisSession) {
-      // If geth is started twice within the same session,
-      // there is an issue where it cannot find deployed contracts.
-      // Restarting the app fixes this issue.
-      restartApp()
-    }
   } catch (e) {
+    const zeroSync3 = yield select(zeroSyncSelector)
+    Logger.info(TAG, `switchToGethFromZeroSync, zeroSync3 mode: ${zeroSync3}`)
```
jeanregisser: Looks like log line should be 
```js
Logger.info(TAG + '@switchToGethFromZeroSync', `zeroSync3 mode: ${zeroSync3}`)
```
to be consistent with the other statements.

packages/mobile/src/web3/saga.ts
```diff
@@ -83,6 +83,11 @@ export function* checkWeb3SyncProgress() {
 
       yield delay(100) // wait 100ms while web3 syncs
     } catch (error) {
+      // Check if error caused by switch to zeroSyncMode
+      const switchedToZeroSyncMode = yield select(zeroSyncSelector)
+      if (switchedToZeroSyncMode) {
+        return true
+      }
```
jeanregisser: If we cancelled this specific saga when zeroSync is turned on, I feel that it might be a bit more robust to unintended side effects.
What do you think?

jeanregisser (approved):
Looks good! 👍  Left one question.

packages/mobile/src/web3/saga.ts
```diff
@@ -83,6 +83,11 @@ export function* checkWeb3SyncProgress() {
 
       yield delay(100) // wait 100ms while web3 syncs
     } catch (error) {
+      // Check if error caused by switch to zeroSyncMode
+      const switchedToZeroSyncMode = yield select(zeroSyncSelector)
```
jmrossy, line 88: Seems like this doesn't really check if the zero sync *caused* the error, just checks if it's turned on.

packages/mobile/src/web3/saga.ts
jmrossy: +1

packages/mobile/src/web3/saga.ts
```diff
@@ -424,8 +429,20 @@ export function* switchToGethFromZeroSync() {
...
 
+    if (gethAlreadyStartedThisSession) {
+      // Call any method on web3 to avoid a persist state issue
+      // This is a temporary workaround as this restart will be
+      // removed when the geth issue is resolved
+      yield call(web3.eth.isSyncing)
+      // If geth is started twice within the same session,
+      // there is an issue where it cannot find deployed contracts.
+      // Restarting the app fixes this issue.
+      restartApp()
```
jmrossy, line 441: Should we communicate this to the user somehow?

packages/mobile/src/web3/saga.ts
jmrossy: I feel like I see mistakes with our logging format often from lots of people. Maybe we should do something more intuitive?
@martinvol suggested this a while back:
`Logger.info(TAG, 'functionName', 'Log message')`

annakaz: Good catch Jean, thanks! I'll fix to follow the `Logger.info(TAG, 'functionName', 'Log message')` format

packages/mobile/src/web3/saga.ts
annakaz: Agreed. Unfortunately this will take some more refactoring, since `checkWeb3SyncProgress()` is called a few different places (not through a watcher where we can easily cancel). I'll address this in a separate PR since it will involve some significant changes to the web3 sync sagas

packages/mobile/src/web3/saga.ts
annakaz: It's communicated in a modal in CeloLite that pops up if a restart will be needed

packages/mobile/src/web3/saga.ts
annakaz: Added a comment to clarify. This will be changed when the saga is cancelled after some web3 refactoring in a future PR

annakaz:
Updated based on review comments, rebuilt alfajores apk and tested on device. Will merge with current fixes as the production app crashes without them, will address cancelling the saga and some web3 saga refactoring in a future PR.

PR #1766, opened 2019-11-19
Sign whole metadata instead of individual claims

nambrot:
### Description

This PR removes the notion of signed claims and instead signs the whole metadata file. Originally I tried to make a blockchain-y construct, but it turns out to be more hassle than worth it.

### Tested

- Unit tests

### Backwards compatible

This PR is not backwards compatible as it will make all prior metadata invalid


packages/contractkit/src/identity/claims/account.test.ts
```diff
@@ -68,7 +68,8 @@ testWithGanache('Account claims', (web3) => {
       IdentityMetadataWrapper.fetchFromURL = () => Promise.resolve(otherMetadata)
 
       const metadata = IdentityMetadataWrapper.fromEmpty(address)
-      signedClaim = await metadata.addClaim(
+      claim = createAccountClaim(otherAddress)
+      await metadata.addClaim(
```
timmoreton: Not sure what’s happening here? Did you mean to pass in claim?

packages/contractkit/src/identity/claims/claim.ts
```diff
@@ -103,6 +82,11 @@ export function hashOfClaim(claim: Claim) {
   return hashMessage(serializeClaim(claim))
 }
 
+export function hashOfClaims(claims: Claim[]) {
+  const hashes = claims.map(hashOfClaim)
+  return hashMessage(hashes.join(''))
```
timmoreton, line 92: Do you get a predictable ordering here?

packages/contractkit/src/identity/claims/account.test.ts
nambrot: I did!

packages/contractkit/src/identity/claims/claim.ts
nambrot: If you are asking whether the `hashOfClaims` is yielding the same hash given an ordered `claims` parameter, then I believe so?

timmoreton: oh yeah of course SG

packages/cli/src/commands/account/claim-keybase.ts
```diff
@@ -96,21 +92,26 @@ export default class ClaimKeybase extends ClaimCommand {
...
-  printManualInstruction(claim: SignedClaim, username: string, address: string) {
+  printManualInstruction(
+    claim: KeybaseClaim,
+    signature: string,
+    username: string,
+    address: string
+  ) {
     const fileName = proofFileName(address)
-    writeFileSync(fileName, JSON.stringify(claim))
+    writeFileSync(fileName, JSON.stringify({ claim, signature }))
```
timmoreton, line 114: what happens if this file already exists?

timmoreton (approved):

packages/cli/src/commands/account/claim-keybase.ts
nambrot: We'll overwrite it

PR #1767, opened 2019-11-19
Add structured logging and Prometheus events

### LINKED ISSUE #1382: Validators should be able to operate the attestation service with proper logging/monitoring
### Expected Behavior

### Current Behavior

What is the current behavior?

nambrot:
### Description

This PR adds structured logging and Prometheus events to the attestation service.

### Tested

- Locally against `namoffchainreveal`

### Related issues

- Fixes #1382


packages/attestation-service/src/index.ts
```diff
@@ -1,38 +1,46 @@
 import * as dotenv from 'dotenv'
+
+// We need to load the config before some of the imports for e.g. the logger
+if (process.env.CONFIG) {
```
mcortesi: this doesn't sound right... imports by ES6 definition must be the first thing on a file, you can't add imports after code

packages/attestation-service/src/index.ts
```diff
@@ -1,38 +1,46 @@
...
+  dotenv.config({ path: process.env.CONFIG })
+}
+
 import express from 'express'
 import RateLimiter from 'express-rate-limit'
+import requestIdMiddleware from 'express-request-id'
+import * as PromClient from 'prom-client'
 import { initializeDB, initializeKit } from './db'
-import { createValidatedHandler } from './request'
+import { logger } from './logger'
```
mcortesi: maybe rename to "rootLogger"

packages/attestation-service/src/logger.ts
```diff
@@ -0,0 +1,19 @@
...
+    : { stream: process.stdout, level: logLevel }
+
+export const logger = createLogger({
+  name: 'attestation-service',
+  serializers: stdSerializers,
+  streams: [stream],
+})
+
+// Somehow it wouldn't let me export logger without export Logger
+export { Logger }
```
mcortesi: try setting the type for const logger explicitely

`const logger: Logger`

packages/attestation-service/src/request.ts
```diff
@@ -1,29 +1,33 @@
...
 import express from 'express'
 import { isLeft } from 'fp-ts/lib/Either'
 import * as t from 'io-ts'
+import { logger } from './logger'
 
 export enum ErrorMessages {
   UNKNOWN_ERROR = 'Something went wrong',
 }
 
 export function catchAsyncErrorHandler<T>(
```
mcortesi: i think `asyncHandler()` is good enough name :P

packages/attestation-service/src/request.ts
```diff
@@ -64,3 +68,22 @@ function serializeErrors(errors: t.Errors) {
 export function respondWithError(res: express.Response, statusCode: number, error: string) {
   res.status(statusCode).json({ success: false, error })
 }
+
+export type Response = Omit<express.Response, 'locals'> & {
+  locals: { logger: Logger } & any
```
mcortesi: why the `& any`?

packages/attestation-service/src/request.ts
```diff
@@ -1,29 +1,33 @@
...
 ) {
-  return async (req: express.Request, res: express.Response) => {
-    try {
-      return handler(req, res)
-    } catch (error) {
-      console.error(error)
-      respondWithError(res, 500, ErrorMessages.UNKNOWN_ERROR)
-    }
+  return (req: express.Request, res: Response) => {
+    handler(req, res)
```
mcortesi: word of caution.... here you don't catch an edge case, suppose the handler() do some things and then return a promise... for example:

```ts
function evilHandler() {
  if (true) {
    throw new Error()
  }
  return callSomethingThatRetunsAPromise()
}
```

In this case... there's a throw before, you are not actually catching the error.

mcortesi (commented):
left some comments

packages/attestation-service/src/request.ts
nambrot: I wasn't entirely sure if the compiler would be capable to intersect both `locals`, but let me try

packages/attestation-service/package.json
```diff
@@ -31,6 +31,8 @@
     "@celo/utils": "^0.1.0",
     "bignumber.js": "^7.2.0",
     "body-parser": "1.19.0",
+    "bunyan": "1.8.12",
+    "bunyan-gke-stackdriver": "0.1.2",
```
jmrossy, line 35: Doesn't this couple the service tightly with GCloud? Validators may want to host this on AWS/Azure

packages/attestation-service/src/request.ts
nambrot: Turns out 

```
export type Response = Omit<express.Response, 'locals'> & {
  locals: { logger: Logger } & Omit<any, 'logger'>
}
```

was the way to go

packages/attestation-service/src/request.ts
nambrot: Great catch, definitely wasn't aware

packages/attestation-service/package.json
nambrot: `bunyan-gke-stackdriver` by @mcortesi is a plugin to `bunyan` to create an alternatively formatted output stream, it does not require any GCP specific functionality to build/run, and can be configured with `LOG_FORMAT=stackdriver`, so is opt-in

mcortesi (approved):

nambrot:
Tests pass, but enabled SSH

PR #1769, opened 2019-11-19
[Wallet] Fix carousel and animations on iOS

jmrossy:
### Description

Fix carousel initial rendering on iOS
Fix lottie animatios on iOS by upgrading package

### Tested

Checked these screens


jeanregisser (approved):
Awesome! 🎉

PR #1549, opened 2019-10-31
Deduplicate yarn packages

jeanregisser:
### Description

yarn deduplication is currently not optimal. After a while of adding/updating/removing packages, some duplicated versions are installed even if they have compatible semver ranges!

See https://github.com/yarnpkg/yarn/issues/3778 and https://medium.com/@bnaya/yarn-deduplicate-the-hero-we-need-f4497a362128

Until yarn fixes this (hopefully in yarn v2) we can use this tool https://github.com/atlassian/yarn-deduplicate to limit this duplication.

For this PR, I ran the following: `npx yarn-deduplicate yarn.lock && yarn`

Here are the results on my machine (late 2013 MacBook Pro):
- Total unique packages in `yarn.lock` went from `4307` to `3868` (**~10%** reduction).
- Size of the repo after install went from `3.8G` to `3.4G` (saved ~400MB, also **~10%** reduction)
- `yarn install` time with a clean yarn cache (`yarn reset` has been run) went from `3m46s` to `3m20s` (**~8%** reduction)
This includes download time of the packages which may vary depending on network conditions.
- `yarn install` time with a populated yarn cache (`yarn reset-modules` has been run, so node_modules are empty) went from `2m29s` to `2m0s` (**~20%** reduction). 
This doesn't require downloading packages since the local yarn cache is populated.

**Important:** this changes our dependency tree. There are certain code paths that now will run with a different set of dependencies.

#### Before the changes

Unique packages in `yarn.lock`:
```
❯ grep "\sversion \"" yarn.lock | wc -l
    4307
```

1st run `yarn install` with a clean yarn cache (`yarn reset` has been run before starting this):
[asciicast](https://asciinema.org/a/xN4jxNCuH1NvjS9FUfBzZHmWK)
2nd run `yarn install` with a populated yarn cache (`yarn reset-modules` has been run before starting this):
[asciicast](https://asciinema.org/a/FdYpivRTlTdmOvfLTFJA8Z56S)

#### After the changes

Unique packages in `yarn.lock`:
```
❯ grep "\sversion \"" yarn.lock | wc -l
    3868
```

1st run `yarn install` with a clean yarn cache (`yarn reset` has been run before starting this):
[asciicast](https://asciinema.org/a/blb35j8CJuxtjdoqGMU4fzdZt)
2nd run `yarn install` with a populated yarn cache (`yarn reset-modules` has been run before starting this):
[...]

PR #1770, opened 2019-11-19
Try CircleCI memory+ class to fix build step (6GB instead of 4GB of memory)

jeanregisser:
### Description

Though https://github.com/celo-org/celo-monorepo/pull/1549 built fine, merging it caused the initial install step to run out of memory:

https://app.circleci.com/jobs/github/celo-org/celo-monorepo/66762
Screenshot 2019-11-19 at 15 11 03

This PR makes CircleCI run the failing step with a VM with more memory (6GB instead of the default 4GB).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,8 @@ general:
 jobs:
   install_dependencies:
     <<: *defaults
+    # Source: https://circleci.com/docs/2.0/configuration-reference/#resource_class
+    resource_class: medium+
     steps:
       - restore_cache:
           keys:


jmrossy (commented):
Any theories why a smaller yarn.lock is now requiring more resources?

jeanregisser:
Given more packages are now deduplicated, there are more packages in the root `node_modules`.
In this specific case the mobile build step consumed more memory apparently.
So my guess is TypeScript parsing more typings from the root `node_modules/@types` folder (by default it includes all typings there).

jmrossy (approved):

PR #1771, opened 2019-11-19
Add first version of public terraform module.

### LINKED ISSUE #1693: Pre-canned GCP Terraform template
A terraform template for deploying on GCP a single validator, a single proxy, and an attestation service pre-configured to connect to Baklava

jcortejoso:
### Description

Added a new package with the terraform module for GCP prepared to be used by any external user.

### Tested

Tested with other GCP account.

### Other changes
-
### Related issues

- Fixes #1693

### Backwards compatibility

Do not affect existing terraform module.

Remaining points:
- [x] Add the Attestation service
- [x] Documentation to use this module
- [x] Test with working parameters (docker images, bootnode parameters, etc.)
- [x] Test the attestation service
- [x] Final changes (update the example tfs and run using documentation)


packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,278 @@
...
...DAwMDAwMDAwMDAwMDAwMDAwMDAwMDgxNTI1MDYwMjAwMTkxNTA1MDYwNDA1MTgwOTEwMzkwZmQ1YjYxMDY2YjgxNjEwNmIxNTY1YjUwNTY1YjYwMDA4MDYwNDA1MTgwODA3ZjZmNzI2NzJlNjM2NTZjNmYyZTZmNzc2ZTY1NzIwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA4MTUyNTA2MDBlMDE5MDUwNjA0MDUxODA5MTAzOTAyMDkwNTA4MDU0OTE1MDUwOTA1NjViNjAwMDYwNDA1MTgwODA3ZjZmNzI2NzJlNjM2NTZjNmYyZTZmNzc2ZTY1NzIwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA4MTUyNTA2MDBlMDE5MDUwNjA0MDUxODA5MTAzOTAyMDkwNTA4MTgxNTU4MTczZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZmZjE2N2Y1MDE0NmQwZTNjNjBhYTFkMTdhNzA2MzViMDU0OTRmODY0ZTg2MTQ0YTIyMDEyNzUwMjEwMTRmYmYwOGJhZmUyNjA0MDUxNjA0MDUxODA5MTAzOTBhMjUwNTA1NmZlYTE2NTYyN2E3YTcyMzA1ODIwZDAwMTUxNzVmZTQ3YzM2YWI1ZWRiNTcwMjAyNTg4MTdhZTQwZTlmNWUxOWVkY2YwYWM4NzA2M2ZmOGNmNTlhMzAwMjkiLCJzdG9yYWdlIjp7IjB4MzRkYzVhMjU1NmIyMDMwOTg4NDgxOTY5Njk2ZjI5ZmVkMzhkNDU4MTNkODAwM2Y2YzcwZTVjMTZhYzkyYWUwZiI6IjQ3ZTE3MkY2Q2ZCNmM3RDAxQzE1NzRmYTNFMkJlN0NDNzMyNjlEOTUifSwiYmFsYW5jZSI6IjAifX0sIm51bWJlciI6IjB4MCIsImdhc1VzZWQiOiIweDAiLCJtaXhIYXNoIjoiMHg2Mzc0Njk2MzYxNmMyMDYyNzk3YTYxNmU3NDY5NmU2NTIwNjY2MTc1NmM3NDIwNzQ2ZjZjNjU3MjYxNmU2MzY1IiwicGFyZW50SGFzaCI6IjB4MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMCJ9"
```
nambrot: I assume we'll plan on fetching this automatically at some point?

packages/terraform-modules-public/testnet/modules/attestation-service/startup.sh
```diff
@@ -0,0 +1,166 @@
...
+        echo -n '$ACCOUNT_ADDRESS' > $DATA_DIR/accountAddress && \
+        echo -n '$CELO_PROVIDER' > $DATA_DIR/celoProvider && \
+        echo -n '$SMS_PROVIDERS' > $DATA_DIR/smsProviders && \
+        echo -n '$NEXMO_KEY' > $DATA_DIR/nexmoKey && \
+        echo -n '$NEXMO_SECRET' > $DATA_DIR/nexmoSecret && \
+        echo -n '$NEXMO_BLACKLIST' > $DATA_DIR/nexmoBlacklist \
+      ) && ( \
+        yarn run db:create:dev && \
+        yarn run db:migrate:dev && \
+        yarn run dev \
```
nambrot: This should just be yarn start, no?

nambrot (approved):
I don't know much about terraform so looks good to me, but you'll probably wanna run it by @tkporter

packages/terraform-modules-public/testnet/variables.tf
```diff
@@ -0,0 +1,271 @@
...
+
+variable attestation_service_docker_image_tag {
+  type        = string
+  description = "The docker image tag for the attestation service"
+  default     = ""
+}
+
+variable attestation_service_account_address {
+  type        = string
+  description = "The etherbase account address for signing the attestations"
```
nambrot: Misleading, this should be the account address of the validator

jcortejoso: Got it! thanks

packages/terraform-modules-public/example/variables.tf
jcortejoso: Yeap, I can add the [http](https://www.terraform.io/docs/providers/http/data_source.html) provider to get it from the bucket. Great idea!

packages/terraform-modules-public/testnet/modules/attestation-service/startup.sh
jcortejoso: 👍
The `yarn run db:create:dev` and `yarn run db:migrate:dev` are still needed with `yarn start`?

jcortejoso:
Still working on this. I am having some issues with the ethereum accounts and the celocli command to register the validator and linking to the attestation service. Keep working on this.

packages/terraform-modules-public/README.md
```diff
@@ -0,0 +1,20 @@
...
+infrastructure.
+Infrastructure and all kind of cloud resources are defined in modules, and Terraform creates/changes/destroys when changes are applied.
+
+Inside the [testnet](./testnet) folder you will find a module (and submodules) to create the setup for running a Celo Validator on Google Cloud Platform. The next logic resources can be created:
+
+- Proxy module for creating a Geth Proxy connected to a validator
+- Validator module for deploying a Validator
+- TX Node for deploying a transmission node, thought to expose the rpc interface and allows interaction with the network easily
+- TX Node Load balancer for deploying a load balancer exposing the tx nodes created
+- Attestation service for deploying the Attestation Service (TODO:link doc)
```
tkporter: Todo here

packages/terraform-modules-public/example/README.md
```diff
@@ -0,0 +1,59 @@
+# Configure your setup
+
+You can take this terraform code as base for your setup. You need to configure the next parameters.
+
+## Project parameters
+
+The file [variables.tf](./variables.tf) contains most of the parameters used by the module. The first parameters to configure are the Google Cloud parameters. Please configure your GCP project, region and zone. Additionally you can coonfigure, directly on [main.tf](./main.tf) the configuration for remote tfstate.
```
tkporter: coonfigure -> configure

packages/terraform-modules-public/example/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+mnemonic: history lemon flight ask umbrella emerge lawsuit bar tortoise demand oak brave together kiss dance filter yellow scheme check victory also daring reward uphold
+privateKey: d497b2c97f5cd276c09e53b80ee5300ff37bbf6c6e9b814d908d2ab654e56137
+publicKey: 041e9f487477b7d9f5c5818a1337601f05b790267ffc052aa98b49bea88a920bb2667aea5c99b47718da9198645669d6fa3643e547b9e2e1d386c4d9ee300db0cd
+accountAddress: 0x2A809BeE654AAe41794838291390BC75BEd100BB
+```
+
+In the example, `0x2A809BeE654AAe41794838291390BC75BEd100BB` would be an account address, `d497b2c97f5cd276c09e53b80ee5300ff37bbf6c6e9b814d908d2ab654e56137` a private key.
+
+You have to generate an account for each component you want to deploy. Please save this credentials securely so you can recover or access your account when you need.
```
tkporter: this -> these

packages/terraform-modules-public/example/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+For the proxy_accounts["private_node_keys"] and proxy_accounts["enodes"] you can use the next bash snippet:
+
+```bash
+DOCKER_IMAGE="gcr.io/celo-testnet/geth-all:5f34ad666ef506de914285b5cd9f7ffe88eaed68"
```
tkporter: I'm worried that by hardcoding a specific commit if we were to potentially change the way pkeys/enodes work in the future things would break (as unlikely as we are to do that). I don't think the bootnode is included in `us.gcr.io/celo-testnet/celo-node:baklava` but maybe we should instead have a docker image `us.gcr.io/celo-testnet/celo-bootnode:baklava`, or maybe we could go off of master `us.gcr.io/celo-testnet/geth-all:master`

packages/terraform-modules-public/example/README.md
```diff
@@ -0,0 +1,59 @@
...
+privateKey: d497b2c97f5cd276c09e53b80ee5300ff37bbf6c6e9b814d908d2ab654e56137
+publicKey: 041e9f487477b7d9f5c5818a1337601f05b790267ffc052aa98b49bea88a920bb2667aea5c99b47718da9198645669d6fa3643e547b9e2e1d386c4d9ee300db0cd
+accountAddress: 0x2A809BeE654AAe41794838291390BC75BEd100BB
+```
+
+In the example, `0x2A809BeE654AAe41794838291390BC75BEd100BB` would be an account address, `d497b2c97f5cd276c09e53b80ee5300ff37bbf6c6e9b814d908d2ab654e56137` a private key.
+
+You have to generate an account for each component you want to deploy. Please save this credentials securely so you can recover or access your account when you need.
+
+For the proxy_accounts["private_node_keys"] and proxy_accounts["enodes"] you can use the next bash snippet:
```
tkporter: It wasn't immediately clear why `bootnode` is being used, could you add a little comment here explaining that it's just used for generating the pkey/enode easily?

packages/terraform-modules-public/example/README.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+Which should return an output of the shape:
+
+```bash
+Enode Address: bcd2cabf722abe26f2d3bb7dd02bd2bc0e94f0d2884a3628088974f96a7636abe189771fa3d40ae929dfe3e08ccaf1eca2c1063ce57411f430c96ec3faacbca6
+Associated private node key: eb99dc2de55a5253dd2834411c2795050e8cb67ac3b6482352aa26b99778eabc
+```
+
+The passwords referred in the variables will be used to import the accounts in the geth deployed (i.e.: The passwords have not to exist previously). They will keep your account safe if somebody access to the keystore file or if you want to unlock the account.
```
tkporter: This first sentence confuses me

packages/terraform-modules-public/example/main.tf
```diff
@@ -0,0 +1,120 @@
...
+  source_subnetwork_ip_ranges_to_nat = "ALL_SUBNETWORKS_ALL_IP_RANGES"
+
+  log_config {
+    enable = false
+    filter = "ERRORS_ONLY"
+  }
+}
+
+module "celo_cluster" {
+  source             = "/Users/jcortejoso/Projects/Celo/celo-monorepo/packages/terraform-modules-public/testnet"
```
tkporter: Can you put something in the readme for the example that this line will also need to be changed?

packages/terraform-modules-public/example/main.tf
```diff
@@ -0,0 +1,120 @@
...
+  attestation_service_private_keys                 = var.attestation_service_accounts["private_keys"]
+  attestation_service_sms_providers                = var.attestation_service_credentials["sms_providers"]
+  attestation_service_nexmo_key                    = var.attestation_service_credentials["nexmo_key"]
+  attestation_service_nexmo_secret                 = var.attestation_service_credentials["nexmo_secret"]
+  attestation_service_nexmo_blacklist              = var.attestation_service_credentials["nexmo_blacklist"]
+  attestation_service_twilio_account_sid           = var.attestation_service_credentials["twilio_account_sid"]
+  attestation_service_twilio_messaging_service_sid = var.attestation_service_credentials["twilio_messaging_service_sid"]
+  attestation_service_twilio_auth_token            = var.attestation_service_credentials["twilio_auth_token"]
+  attestation_service_twilio_blacklist             = var.attestation_service_credentials["twilio_blacklist"]
+  attestation_service_celo_provider                = "https://baklavastaging-forno.celo-testnet.org/"
```
tkporter: Change to be baklava?
Also not sure if we want this to always be their own node or not, @nambrot what do you think?

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,234 @@
...
+  default = {
+    repository = "us.gcr.io/celo-testnet/geth"
+    tag        = "c39d61c8847387727d0b01f85719240957bb585c"
+  }
+}
+
+variable network_id {
+  description = "The ethereum network ID"
+  type        = number
+  default     = 12219
```
tkporter: This just changed to `121119`

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,234 @@
...
+  default     = false
+}
+
+variable geth_node_docker_image {
+  description = "The Celo Blockchain docker image"
+  type        = map(string)
+
+  default = {
+    repository = "us.gcr.io/celo-testnet/geth"
+    tag        = "c39d61c8847387727d0b01f85719240957bb585c"
```
tkporter: Just changed to `5d4224eb0e13683107f3a8ea8d7cb668e676b4fc`-- however, I think we may want to tag it as `baklava` as well

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,234 @@
...
+  default     = 5
+}
+
+variable istanbul_request_timeout_ms {
+  description = "The ethereum request timeout"
+  type        = number
+  default     = 10000
+}
+
+variable geth_exporter_docker_image {
```
tkporter, line 203: Could you add a comment about how this is just for exposing metrics from geth in the prometheus format? & likely will want to change the repo / tag combo to what's in `.env.baklava`

packages/terraform-modules-public/testnet/main.tf
```diff
@@ -0,0 +1,251 @@
...
+    protocol = "tcp"
+    ports    = ["9200"]
+  }
+}
+
+resource "google_compute_firewall" "rpc_firewall" {
+  name    = "${var.celo_env}-rpc-firewall"
+  network = var.network_name
+
+  target_tags = local.firewall_target_tags_proxy
```
tkporter: Probably want to expose tx-nodes as well

And now that I think about it, do we even want to have proxies have this port open to the outside world? Maybe only internally for attestation-service? Could be something to think about later

packages/terraform-modules-public/testnet/main.tf
```diff
@@ -0,0 +1,251 @@
...
+resource "google_compute_firewall" "attestation-service" {
+  name    = "${var.celo_env}-attestation-service-firewall"
+  network = var.network_name
+
+  target_tags   = local.firewall_target_tags_attestation_service
+  source_ranges = [data.google_compute_subnetwork.celo.ip_cidr_range]
+
+  allow {
+    protocol = "tcp"
+    ports    = ["3000"]
```
tkporter: @nambrot shouldn't matter that this is port 3000 and not 80 right?

packages/terraform-modules-public/testnet/main.tf
```diff
@@ -0,0 +1,251 @@
...
+  url = "https://storage.googleapis.com/genesis_blocks/${var.celo_env}"
+
+  request_headers = {
+    "Accept" = "application/json"
+  }
+}
+
+data "http" "static-nodes" {
+  url = "https://storage.googleapis.com/static_nodes/${var.celo_env}"
+  // url = "https://storage.googleapis.com/javier_cortejoso/baklavastaging"
```
tkporter: Can remove
Also this is really cool, I didn't know terraform allows you to do this

packages/terraform-modules-public/testnet/modules/tx-node/main.tf
```diff
@@ -0,0 +1,80 @@
...
+      ethstats_host : var.ethstats_host,
+      genesis_content_base64 : var.genesis_content_base64,
+      geth_exporter_docker_image_repository : var.geth_exporter_docker_image_repository,
+      geth_exporter_docker_image_tag : var.geth_exporter_docker_image_tag,
+      geth_node_docker_image_repository : var.geth_node_docker_image_repository,
+      geth_node_docker_image_tag : var.geth_node_docker_image_tag,
+      geth_verbosity : var.geth_verbosity,
+      in_memory_discovery_table : var.in_memory_discovery_table,
+      ip_address : google_compute_address.tx_node[count.index].address,
+      max_peers : var.tx_node_count * 2,
```
tkporter: We should probably increase this

packages/terraform-modules-public/testnet/modules/attestation-service/startup.sh
```diff
@@ -0,0 +1,136 @@
...
+function save_variable {
+  local var=$1
+  local file=$2
+
+  [ -n "$var" ] && echo -n "$var" > "$file"
+}
+
+# ---- Set Up Logging ----
+
+curl -sSO https://dl.google.com/cloudagents/install-logging-agent.sh
```
tkporter: this is just if the users are using stackdriver-- unless if we want to assume users are using stackdriver (I think we're fine not assuming they're using stackdriver), we can delete these from all the startup scripts. If we want to assume that they do use stackdriver, we'll have to add a logging scope to the instances for this to work

packages/terraform-modules-public/testnet/modules/attestation-service/variables.tf
```diff
@@ -0,0 +1,100 @@
...
+}
+
+variable twilio_auth_token {
+  type        = string
+  description = "Twilio account Auth Token (check twilio documentation)"
+}
+
+variable twilio_blacklist {
+  type        = string
+  description = "Twilio blacklisted country codes, separated by coma  (check twilio documentation)"
```
tkporter: coma -> comma

packages/terraform-modules-public/testnet/modules/attestation-service/variables.tf
```diff
@@ -0,0 +1,100 @@
...
+}
+
+variable nexmo_secret {
+  type        = string
+  description = "Nexmo api secret (check nexmo documentation)"
+}
+
+variable nexmo_blacklist {
+  type        = string
+  description = "Nexmo blacklisted country codes, separated by coma (check nexmo documentation)"
```
tkporter: coma -> comma

packages/terraform-modules-public/testnet/modules/attestation-service/variables.tf
```diff
@@ -0,0 +1,100 @@
...
+}
+
+variable account_address {
+  type        = list(string)
+  description = "The account address for signing the attestations. Must be the address of the associated validator"
+}
+
+variable attestation_key {
+  type        = list(string)
+  description = "The account private key for signing the attestations. Must be the private key of the associated validator"
```
tkporter: @nambrot do we allow you to use a different key than the validator's private key?

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,195 @@
...
+EOF
+systemctl restart docker
+
+# ---- Set Up and Run Geth ----
+
+GETH_NODE_DOCKER_IMAGE=${geth_node_docker_image_repository}:${geth_node_docker_image_tag}
+
+ACCOUNT_ADDRESS=${validator_account_address}
+echo "Address: $ACCOUNT_ADDRESS"
+echo "Private Key: ${validator_private_key}"
```
tkporter: Nit: maybe we shouldn't print this just to be safe

packages/terraform-modules-public/testnet/outputs.tf
```diff
@@ -0,0 +1,19 @@
...
+
+output tx_node_ip_addresses {
+  value = module.tx_node.ip_addresses
+}
+
+output tx_node_lb_ip_address {
+  value = module.tx_node_lb.ip_address
+}
+
+// output validator_internal_ip_addresses {
```
tkporter: Comments to delete?

packages/terraform-modules-public/testnet/variables.tf
```diff
@@ -0,0 +1,250 @@
...
+  type        = string
+  description = "Ethstats url or IP address"
+}
+
+variable gcloud_project {
+  type        = string
+  description = "Name of the Google Cloud project to use"
+}
+
+// variable genesis_content_base64 {
```
tkporter: ditto

packages/terraform-modules-public/testnet/main.tf
```diff
@@ -0,0 +1,255 @@
...
+  nexmo_key                                   = var.attestation_service_nexmo_key
+  nexmo_secret                                = var.attestation_service_nexmo_secret
+  nexmo_blacklist                             = var.attestation_service_nexmo_blacklist
+  twilio_account_sid                          = var.attestation_service_twilio_account_sid
+  twilio_messaging_service_sid                = var.attestation_service_twilio_messaging_service_sid
+  twilio_auth_token                           = var.attestation_service_twilio_auth_token
+  twilio_blacklist                            = var.attestation_service_twilio_blacklist
+}
+
+// module "accounts" {
```
tkporter: comment

packages/terraform-modules-public/testnet/modules/validator/variables.tf
```diff
@@ -0,0 +1,132 @@
...
+}
+
+variable proxy_external_ips {
+  type        = list(string)
+  description = "Array list with the proxy external addresses"
+}
+
+variable static_nodes_base64 {
+  type        = string
+  description = "Content of the genesis file encoded in base64"
```
tkporter: genesis file -> static nodes file

tkporter (commented):
Awesome stuff!

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,240 @@
...
+    account_addresses = [
+      "0xF2...",
+    ]
+    private_keys = [
+      "1b...",
+    ]
+    account_passwords = [
+      "secret1",
+    ]
+    private_node_keys = [
```
tkporter: It looks like `private_keys` and `enodes` are the only values that are used. The README also incorrectly specifies `private_node_keys` and `enodes`

packages/terraform-modules-public/example/main.tf
```diff
@@ -0,0 +1,119 @@
...
+  attestation_service_private_keys                 = var.attestation_service_accounts["private_keys"]
+  attestation_service_sms_providers                = var.attestation_service_credentials["sms_providers"]
+  attestation_service_nexmo_key                    = var.attestation_service_credentials["nexmo_key"]
+  attestation_service_nexmo_secret                 = var.attestation_service_credentials["nexmo_secret"]
+  attestation_service_nexmo_blacklist              = var.attestation_service_credentials["nexmo_blacklist"]
+  attestation_service_twilio_account_sid           = var.attestation_service_credentials["twilio_account_sid"]
+  attestation_service_twilio_messaging_service_sid = var.attestation_service_credentials["twilio_messaging_service_sid"]
+  attestation_service_twilio_auth_token            = var.attestation_service_credentials["twilio_auth_token"]
+  attestation_service_twilio_blacklist             = var.attestation_service_credentials["twilio_blacklist"]
+  // attestation_service_celo_provider                = "https://baklava-forno.celo-testnet.org/"
```
tkporter: @nambrot do we care if it's through forno?

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,240 @@
...
+      "f8...",
+    ]
+  }
+}
+
+variable validator_accounts {
+  description = "The account data for the validator nodes"
+  type        = map
+
+  default = {
```
tkporter, line 76: Going off some questions from Discord from "reem" in #developerchat, we should make this clearer that it should be for `VALIDATOR_SIGNER_ADDRESS`, and also give instructions on how to get the private key for an account. The only way I know of that's semi-easy is what I suggested to him:

```
const Web3 = require('web3')
const w3 = new Web3()
const keystoreFileContents = <object found in the file corresponding to your address in your `<datadir>/keystore` directory>
console.log(w3.eth.accounts.decrypt(keystoreFileContents, yourAccountPassword)) // prints out an object with privateKey
```

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,240 @@
...
+  default     = false
+}
+
+variable geth_node_docker_image {
+  description = "The Celo Blockchain docker image"
+  type        = map(string)
+
+  default = {
+    repository = "us.gcr.io/celo-testnet/celo-node"
+    tag        = "geth"
```
tkporter: Is this correct? `baklava` instead?

packages/terraform-modules-public/example/variables.tf
reem: 👋

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,195 @@
...
+
+
+echo "Proxy enode address: ${proxy_enode}"
+echo "Proxy internal ip address: ${proxy_internal_ip}"
+echo "Proxy external ip address: ${proxy_external_ip}"
+PROXY_INTERNAL_ENODE="enode://${proxy_enode}@${proxy_internal_ip}:30503"
+PROXY_EXTERNAL_ENODE="enode://${proxy_enode}@${proxy_external_ip}:30503"
```
tkporter: This should use port 30303

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,195 @@
...
+
+echo "Pulling geth..."
+docker pull $GETH_NODE_DOCKER_IMAGE
+
+IN_MEMORY_DISCOVERY_TABLE_FLAG=""
+[[ ${in_memory_discovery_table} == "true" ]] && IN_MEMORY_DISCOVERY_TABLE_FLAG="--use-in-memory-discovery-table"
+
+# Load configuration to files
+echo -n '${genesis_content_base64}' | base64 -d > $DATA_DIR/genesis.json
+echo -n '${static_nodes_base64}' | base64 -d > $DATA_DIR/static-nodes.json
```
tkporter: Should delete this

Separately, maybe letting a validator use a router isn't the right option.. @reem ran this without an external IP address and found himself still peering with the static nodes when that shouldn't have been possible

packages/terraform-modules-public/example/README.md
jcortejoso: done thanks

packages/terraform-modules-public/README.md
jcortejoso: added thanks

packages/terraform-modules-public/example/README.md
jcortejoso: It was the tool I found to generate the pair pkey/enode, but I guess we can use celocli for that.

packages/terraform-modules-public/example/README.md
jcortejoso: I updated to use celocli instead.

packages/terraform-modules-public/testnet/modules/validator/startup.sh
jcortejoso: Done :+1:

packages/terraform-modules-public/testnet/variables.tf
jcortejoso: :+1:

packages/terraform-modules-public/example/variables.tf
jcortejoso: updated :+1:

packages/terraform-modules-public/testnet/modules/validator/startup.sh
jcortejoso: Correct thanks

packages/terraform-modules-public/example/README.md
jcortejoso: I think the publicKey returned by `celocli:account new` can be used as the `enode` address if you remove the two leading characters. Am I right? I made some tests and seems right, and I updated the doc according to this.

packages/terraform-modules-public/example/variables.tf
jcortejoso: Updated. Most of it related I had a separate folder for my development/deployment. Thanks!

packages/terraform-modules-public/example/variables.tf
jcortejoso: I updated some that part of the `README.md`. As I see this, if we/the user can provide all the needed parameters using only the `celocli account:new` command would be the ideal solution.
Maybe I am missing something but I think the user can create all the needed accounts with that command. So for example the user should create the accounts for the validators (these ones not included in the terraform variables), proxy (only for the enode address), validator signers and attestation service, and the last two has to be authorized using the celocli.

If so the used should not care about decrypting a keystore json file to a private key.

Also I find a bit difficult if an user want to deploy 10, 20 or 30 validators stacks (validator+proxy+signer+attestation service), and managing 4*replicas mnemonics maybe is not the better solution. 
I would say it would be nice include a section about managing accounts (with celotool or with some tool like `ethereum_hdwallet`), but I am not sure if that section would fit better in the general documentation rather than the terraform module. I have some scripts

packages/terraform-modules-public/testnet/modules/validator/startup.sh
jcortejoso: Deleted.
I guess that submitting the `static-nodes.json` the proxy would connect to those nodes but it would not accept any incoming connection from any other node, but also did not start peering new connections as `--no-discover` is configured. 
Now only connection the connection to the proxy should occur.

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,211 @@
...
+}
+
+variable proxy_accounts {
+  description = "The account data for the proxy nodes"
+  type        = map
+
+  default = {
+    account_addresses = [
+      "0xF2...",
+    ]
```
jeanregisser: This isn't used nor needed for the proxy.

packages/terraform-modules-public/testnet/modules/proxy/main.tf
```diff
@@ -0,0 +1,86 @@
...
+      ip_address : google_compute_address.proxy[count.index].address,
+      istanbul_request_timeout_ms : var.istanbul_request_timeout_ms,
+      max_peers : var.proxy_max_peers,
+      network_id : var.network_id,
+      rid : count.index,
+      proxy_name : var.proxy_name,
+      proxy_private_key : var.proxy_private_keys[count.index],
+      validator_account_address : var.validator_account_addresses[count.index],
+      validator_account_password : var.validator_account_passwords[count.index],
+      validator_private_key : var.validator_private_keys[count.index],
```
jeanregisser: AFAIK, the proxy doesn't need to have the validator password and private key.

packages/terraform-modules-public/testnet/modules/proxy/startup.sh
```diff
@@ -0,0 +1,174 @@
...
+# if this instance has been recreated with the same disk), we skip formatting
+if [[ $CURRENT_DISK_FORMAT == $DISK_FORMAT ]]; then
+  echo "Disk $DISK_PATH is correctly formatted as $DISK_FORMAT"
+else
+  echo "Disk $DISK_PATH is not formatted correctly, formatting as $DISK_FORMAT..."
+  mkfs.ext4 -m 0 -F -E lazy_itable_init=0,lazy_journal_init=0,discard $DISK_PATH
+fi
+
+echo "Mounting $DISK_PATH onto $DATA_DIR"
+mount -o discard,defaults $DISK_PATH $DATA_DIR
```
jeanregisser: I got an error on Debian 10 with mount.
I had to `mkdir -p $DATA_DIR` before `mount` otherwise it would complain that the mount point doesn't exist.

packages/terraform-modules-public/testnet/modules/proxy/startup.sh
```diff
@@ -0,0 +1,174 @@
...
+# We need to override the entrypoint in the geth image (which is originally `geth`).
+# `geth account import` fails when the account has already been imported. In
+# this case, we do not want to pipefail
+docker run \
+  --rm \
+  --net=host \
+  -v $DATA_DIR:$DATA_DIR \
+  --entrypoint /bin/sh \
+  -i $GETH_NODE_DOCKER_IMAGE \
+  -c "geth init $DATA_DIR/genesis.json && geth account import --password $DATA_DIR/account/accountSecret $DATA_DIR/validatorPKey | true"
```
jeanregisser: Related to the comment above, the `geth account import` is not necessary for the proxy.

packages/terraform-modules-public/testnet/modules/proxy/startup.sh
```diff
@@ -0,0 +1,174 @@
...
+[[ ${in_memory_discovery_table} == "true" ]] && IN_MEMORY_DISCOVERY_TABLE_FLAG="--use-in-memory-discovery-table"
+
+# Load configuration to files
+echo -n '${genesis_content_base64}' | base64 -d > $DATA_DIR/genesis.json
+echo -n '${static_nodes_base64}' | base64 -d > $DATA_DIR/static-nodes.json
+echo -n '${rid}' > $DATA_DIR/replica_id
+echo -n '${ip_address}' > $DATA_DIR/ipAddress
+echo -n '${proxy_private_key}' > $DATA_DIR/pkey
+echo -n '${validator_private_key}' > $DATA_DIR/validatorPKey
+echo -n '${validator_account_password}' > $DATA_DIR/account/accountSecret
```
jeanregisser: These can be removed, the proxy doesn't need these.

packages/terraform-modules-public/testnet/modules/proxy/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+
+variable validator_account_passwords {
+  type        = list(string)
+  description = "Array with the Validator etherbase account passwords"
+}
+
+variable validator_private_keys {
+  type        = list(string)
+  description = "Array with the Validator private keys"
+}
```
jeanregisser: We can remove these too for the proxy.

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,192 @@
...
+
+GETH_NODE_DOCKER_IMAGE=${geth_node_docker_image_repository}:${geth_node_docker_image_tag}
+
+ACCOUNT_ADDRESS=${validator_account_address}
+echo "Address: $ACCOUNT_ADDRESS"
+
+echo "Proxy enode address: ${proxy_enode}"
+echo "Proxy internal ip address: ${proxy_internal_ip}"
+echo "Proxy external ip address: ${proxy_external_ip}"
+PROXY_INTERNAL_ENODE="enode://${proxy_enode}@${proxy_internal_ip}:30303"
```
jeanregisser: port should be `30503` for connecting to the internal proxy ip

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,192 @@
...
+# if this instance has been recreated with the same disk), we skip formatting
+if [[ $CURRENT_DISK_FORMAT == $DISK_FORMAT ]]; then
+  echo "Disk $DISK_PATH is correctly formatted as $DISK_FORMAT"
+else
+  echo "Disk $DISK_PATH is not formatted correctly, formatting as $DISK_FORMAT..."
+  mkfs.ext4 -m 0 -F -E lazy_itable_init=0,lazy_journal_init=0,discard $DISK_PATH
+fi
+
+echo "Mounting $DISK_PATH onto $DATA_DIR"
+mount -o discard,defaults $DISK_PATH $DATA_DIR
```
jeanregisser: We should `mkdir -p $DATA_DIR` before `mount` here too (see initial comment above).

packages/terraform-modules-public/testnet/main.tf
```diff
@@ -0,0 +1,240 @@
...
+  reset_geth_data                       = var.reset_geth_data
+
+  validator_name              = var.validator_name
+  validator_account_addresses = var.validator_account_addresses
+  validator_private_keys      = var.validator_private_keys
+  validator_account_passwords = var.validator_account_passwords
+  proxy_enodes                = var.proxy_enodes
+  proxy_internal_ips          = module.proxy.internal_ip_addresses
+  proxy_external_ips          = module.proxy.external_ip_addresses
+  static_nodes_base64         = base64encode(data.http.static-nodes.body)
```
jeanregisser: We can remove `static_nodes_base64` for the validator, right?

jeanregisser (changes requested):
Awesome work @jcortejoso 🎉 

I adapted your work for a French based cloud provider (https://www.scaleway.com/en/) over the holidays and found a few issues that I'm reporting here.
I only tested proxy and validator, not yet the attestation or tx-node.

packages/terraform-modules-public/testnet/main.tf
jcortejoso: Right thanks 👍

packages/terraform-modules-public/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+[Terraform](https://www.terraform.io) is a tool by Hashicorp that allows developers to treat _"infrastructure as code"_, easying the management and repeatibility of the
+infrastructure.
+Infrastructure and all kind of cloud resources are defined in modules, and Terraform creates/changes/destroys when changes are applied.
+
+Inside the [testnet](./testnet) folder you will find a module (and submodules) to create the setup for running a Celo Validator on Google Cloud Platform. The next logic resources can be created:
+
+- Proxy module for creating a Geth Proxy connected to a validator
+- Validator module for deploying a Validator
+- TX Node for deploying a transmission node, thought to expose the rpc interface and allows interaction with the network easily
```
tkporter: Nit: transmission -> transaction
People might be more familiar with "full node" though

packages/terraform-modules-public/example/README.md
tkporter: What do you mean by "The passwords have not to exist previously"?

packages/terraform-modules-public/example/main.tf
```diff
@@ -0,0 +1,117 @@
+provider "google" {
+  project = var.google["project"]
+  region  = var.google["region"]
+  zone    = var.google["zone"]
+}
+
+# For managing terraform state remotely
+terraform {
+  backend "gcs" {
+    bucket = "my_bucket"
```
tkporter, line 10: It might not be clear to people who aren't super familiar with Terraform that they need to create a storage bucket manually & change these lines, could we add a little note to the readme?

Also to mention that terraform will include their account secrets inside the state, so they should be very cautious about who can read the state and make sure it's properly encrypted

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,211 @@
...
+    attestation_service = 1
+  }
+}
+
+variable instance_types {
+  description = "The instance type for each component"
+  type        = map(string)
+
+  default = {
+    validator           = "n1-standard-1"
```
tkporter: I think we should change the validator & proxy to `n1-standard-2` to reflect the hardware we recommend in the docs & what we are running ourselves

packages/terraform-modules-public/example/variables.tf
tkporter: +1 on the `celocli account:new` usage

I agree that we need to think more about key management, it's definitely been an afterthought and puts people who aren't super familiar with key management at a disadvantage

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,211 @@
...
+}	
+
+# Attestation variables
+variable attestation_service_docker_image {
+  description = "The attestation_service docker image"
+  type        = map(string)
+
+  default = {
+    repository = "us.gcr.io/celo-testnet/celo-monorepo"
+    tag        = "attestation-service-c8e3392aa2ca44ff83b4035700ece5fd12ed2b84"
```
tkporter: We should change this to be what's in the docs: `us.gcr.io/celo-testnet/celo-monorepo:attestation-service-baklava`

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,211 @@
...
+  }
+}
+
+# SMS provider configuration
+variable attestation_service_credentials {
+  description = "Provider with the credentials for the SMS provider. Provider must be nexmo or twilio"
+  type        = map(string)
+
+  default = {
+    sms_providers                = "nexmo"
```
tkporter: Since we recommend twilio & the docs show twilio it might be a good idea to default to twilio here

packages/terraform-modules-public/example/variables.tf
```diff
@@ -0,0 +1,211 @@
...
+  type        = number
+  default     = 10000
+}
+
+variable geth_exporter_docker_image {
+  description = "The geth exporter docker image"
+  type        = map(string)
+
+  default = {
+    repository = "jcortejoso/ge"
```
tkporter: intentional?

packages/terraform-modules-public/testnet/main.tf
tkporter: We recently changed the proxy & tx nodes to not use static nodes when joining the network, and to instead use the bootnode. The bootnode enode can be found at `https://storage.googleapis.com/env_bootnodes/${var.celo_env}` and the value of what's found there is passed as the value for the flag `--bootnodes`

packages/terraform-modules-public/testnet/main.tf
tkporter: This should just be for tx-nodes, right? If we want to allow these ports to be used on the proxy like we do when setting `celo_provider` for the attestation service maybe we should just expose these ports on the proxy for internal source IPs in the block right below

packages/terraform-modules-public/testnet/main.tf
```diff
@@ -0,0 +1,239 @@
...
+  source_ranges = [data.google_compute_subnetwork.celo.ip_cidr_range]
+
+  allow {
+    protocol = "tcp"
+    ports    = ["30503"]
+  }
+
+  allow {
+    protocol = "udp"
+    ports    = ["30503"]
```
tkporter: Not a big deal, but we don't need to expose UDP for 30503 because that's only used for discovery

packages/terraform-modules-public/testnet/main.tf
```diff
@@ -0,0 +1,239 @@
...
+
+  target_tags = concat(local.firewall_target_tags_txnode, local.firewall_target_tags_validator, local.firewall_target_tags_proxy)
+
+  allow {
+    protocol = "tcp"
+    ports    = ["30303"]
+  }
+
+  allow {
+    protocol = "udp"
```
tkporter, line 73: Same thing, doesn't really matter but udp 30303 isn't needed for the validator. The only benefit would be to prevent issues where we accidentally have discovery on for the validator but only change this if you feel like it

packages/terraform-modules-public/testnet/main.tf
tkporter: I think we should change port 3000 to port 80 to keep with the docs

packages/terraform-modules-public/testnet/modules/attestation-service/variables.tf
tkporter: Answering my question, this can be the "attestation signer key" according to here https://docs.celo.org/getting-started/baklava-testnet/running-a-validator#account-and-signer-keys

packages/terraform-modules-public/testnet/modules/proxy/main.tf
```diff
@@ -0,0 +1,86 @@
...
+  )
+}
+
+resource "google_compute_disk" "proxy" {
+  name  = "${local.name_prefix}-disk-${count.index}"
+  count = var.validator_count
+
+  type = "pd-ssd"
+  # in GB
+  size                      = 10
```
tkporter: Imo we should consider upping these. The validator & proxy I'm currently running on my personal validator are each 40 gb and that kept me under the quota limit set by GCP

packages/terraform-modules-public/testnet/modules/tx-node/main.tf
tkporter: bump ^

packages/terraform-modules-public/testnet/variables.tf
```diff
@@ -0,0 +1,234 @@
...
+}
+
+variable instance_types {
+  description = "The instance type for each component"
+  type        = map(string)
+
+  default = {
+    validator           = "n1-standard-1"
+    proxy               = "n1-standard-1"
+    txnode              = "n1-standard-1"
```
tkporter: These should be `n1-standard-2` to keep with what we use and recommend too

tkporter (commented):
Looking really good! If you haven't already we should try running it from scratch & going through all the important "running a validator" steps in the docs to make sure that everything works as intended before making it public

packages/terraform-modules-public/example/variables.tf
jcortejoso: Right. I think I left there because it helps (at least me) for identifying the account I was using, although you can identify with any other parameter. I will update it with a comment. Thanks!

packages/terraform-modules-public/example/variables.tf
jcortejoso: The geth-exporter was not public. I have pushed manually the latest version as public. We can also do not run geth-exporter as part of this module.

jcortejoso:
@jeanregisser @tkporter thanks for your great feedback. I have updated with most of the feedback you provided. 
Only missing thing is testing the attestation service with Twilio variables.

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,193 @@
...
+echo "Mounting $DISK_PATH onto $DATA_DIR"
+mkdir -p $DATA_DIR
+mount -o discard,defaults $DISK_PATH $DATA_DIR
+[[ ${reset_geth_data} == "true" ]] && rm -rf $DATA_DIR/geth
+mkdir -p $DATA_DIR/account
+
+# ---- Install Docker ----
+
+echo "Installing Docker..."
+apt update && apt upgrade
```
silasbw: worth tossing in `-y`s here?  `apt` _can_ prompt, but maybe it does something smart depending on stdin and avoids prompting. regardless, it might be useful for development (_e.g._, avoid having the script prompt when you're running it in an ssh session)

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,193 @@
...
+apt install -y apt-transport-https ca-certificates curl software-properties-common gnupg2
+curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add -
+add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/debian $(lsb_release -cs) stable"
+apt update && apt upgrade
+apt install -y docker-ce
+systemctl start docker
+
+echo "Configuring Docker..."
+cat <<'EOF' > '/etc/docker/daemon.json'
+{
```
silasbw, line 49: is it worth splitting out these configuration files and hosting them in separate self-contained files?

If you do that, I'd re-use them for configuring Debian with Azure Templates.

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,193 @@
...
+if [[ $CURRENT_DISK_FORMAT == $DISK_FORMAT ]]; then
+  echo "Disk $DISK_PATH is correctly formatted as $DISK_FORMAT"
+else
+  echo "Disk $DISK_PATH is not formatted correctly, formatting as $DISK_FORMAT..."
+  mkfs.ext4 -m 0 -F -E lazy_itable_init=0,lazy_journal_init=0,discard $DISK_PATH
+fi
+
+echo "Mounting $DISK_PATH onto $DATA_DIR"
+mkdir -p $DATA_DIR
+mount -o discard,defaults $DISK_PATH $DATA_DIR
```
silasbw: do you need to update the fstab (so the fs mounts on boot)?

packages/terraform-modules-public/testnet/modules/validator/startup.sh
```diff
@@ -0,0 +1,193 @@
...
+
+DISK_FORMAT=ext4
+CURRENT_DISK_FORMAT=`lsblk -i -n -o fstype $DISK_PATH`
+
+echo "Checking if disk $DISK_PATH format $CURRENT_DISK_FORMAT matches desired $DISK_FORMAT..."
+
+# If the disk has already been formatted previously (this will happen
+# if this instance has been recreated with the same disk), we skip formatting
+if [[ $CURRENT_DISK_FORMAT == $DISK_FORMAT ]]; then
+  echo "Disk $DISK_PATH is correctly formatted as $DISK_FORMAT"
```
silasbw, line 19: Naive question: under what conditions is the disk already formatted?  will it have content?  if so, do you need to remove that content?

packages/terraform-modules-public/testnet/modules/validator/startup.sh
silasbw: e.g.,

```
echo "Mounting $DISK_PATH onto $DATA_DIR"
mkdir -p $DATA_DIR
DISK_UUID=`blkid $DISK_PATH | cut -d \" -f 2`
echo "UUID=${DISK_UUID}     $DATA_DIR   auto    discard,defaults    0    0" >> /etc/fstab
mount $DATA_DIR
```

packages/terraform-modules-public/testnet/modules/validator/startup.sh
silasbw: ...actually, most of this file is reusable for the Azure debian setup, with some differences on how sensitive data (e.g., `validator_private_key`) is handled.  Worth splitting the whole `startup.sh` file out into a separate PR and location (e.g., `packages/debian` or something)?

packages/terraform-modules-public/testnet/modules/validator/startup.sh
jcortejoso: The disk would be formatted (and with data) if you are updating some parameter of the validator or proxy (like the account secrets, or geth image). There is another variable `reset_geth_data` to delete the existing geth data.

packages/terraform-modules-public/testnet/modules/validator/startup.sh
jcortejoso: Yeah, that makes a lot of sense. I was thinking about the AWS module and most of the script will be reusable also for it, so we can think about organizing the code so it can be easily imported for each component.

packages/terraform-modules-public/testnet/modules/validator/startup.sh
jcortejoso: Yes, also makes sense. Really the module is not tested/thought to restart the instance but we can add support to it, so if someone manages the instances manually this will help.

packages/terraform-modules-public/testnet/modules/validator/startup.sh
silasbw: cool -- do you have cycles to do that?  do you want me to help with moving the `startup.sh`s elsewhere?

packages/terraform-modules-public/testnet/modules/validator/startup.sh
jcortejoso: Probably it's worth. I can imagine for now it has not caused any issue without because the update does not need it AFAIK and for now neither the upgrade command (I think it ask for confirmation when it needs to upgrade/install new packages or remove some installed package because the upgrade). Anyway it does not hurt so I will add it.
Thanks

packages/terraform-modules-public/testnet/modules/validator/startup.sh
jcortejoso: Hi Silas,
I was thinking about merging this PR and then address that in a new PR, so we can close this PR asap and do not postpone too far. What do you think?

silasbw: :+1:

jeanregisser (commented):
One last question about the wording, the existing docs for setting up a validator uses `validator_signer_*` and `attestation_signer_*`, I think it might be confusing that the terraform module doesn't use the same.

I saw you explained this in the README, but shouldn't we still rename the variables to clear the confusion?

packages/terraform-modules-public/example/variables.tf
jeanregisser: bump ☝️

packages/terraform-modules-public/example/variables.tf
jeanregisser: bump ☝️

packages/terraform-modules-public/testnet/modules/attestation-service/variables.tf
jeanregisser: bump ☝️

packages/terraform-modules-public/testnet/modules/attestation-service/variables.tf
jeanregisser: bump ☝️

jeanregisser (commented):
Very nice! just some existing comments from @trevor left that haven't been addressed.

packages/terraform-modules-public/example/variables.tf
jcortejoso: Updated

packages/terraform-modules-public/testnet/main.tf
jcortejoso: Updated

packages/terraform-modules-public/testnet/modules/attestation-service/variables.tf
jcortejoso: Fixed

jeanregisser (approved):

PR #1776, opened 2019-11-19
[Wallet] Handle `/v/<code` deep links for phone verifications

### LINKED ISSUE #896: Users SBAT click deeplinks in verification texts
### Expected Behavior

You can click a deeplink instead copying and pasting

### Current Behavior

Copy and pasting texts if they aren't read

i1skn:
### Description

Add handling of deep links for phone number verification.

### Tested
Ran on Android device and iPhone

### Other changes
* When user tries to enter the same code after app was restarted - it would tell that code has been already used (previously it would say "Invalid code").
* Change verification SMS template
* Add support for deep links for iOS

### Related issues

- Fixes #896 

### Backwards compatibility

Not sure, attestation service should be redeployed AFTER we rolled out new app versions.


packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -53,7 +53,7 @@ function toBase64(str: string) {
 }
 
 function createAttestationTextMessage(attestationCode: string) {
-  return `<#> ${toBase64(attestationCode)} ${process.env.APP_SIGNATURE}`
+  return `<#> celo://wallet/v/${toBase64(attestationCode)} ${process.env.APP_SIGNATURE}`
```
i1skn, line 56: lmk if `celo://wallet/v/` is better to set as an environment variable

packages/mobile/src/app/saga.test.ts
```diff
@@ -90,6 +96,18 @@ describe('App saga', () => {
       .run()
     expect(navigate).toHaveBeenCalledWith(testRoute.routeName, testRoute.params)
   })
+
+  it('Handle Dappkit deep link', async () => {
```
jeanregisser: Nit note to ourself: looks like we're inconsistent with naming these in this file compared to the others.
I've seen a lot of "verb xxx" elsewhere: should be "handles Dappkit deep line".

jeanregisser (commented):
Looking good! :+1: 
Just mentioned orally it could be good to update the attestation parsing tests with the new format.

i1skn:
@jeanregisser I have addressed the comments - PTAL.

jeanregisser (approved):
Great job! 👍

packages/mobile/src/identity/verification.ts
```diff
@@ -365,20 +366,20 @@ function attestationCodeReceiver(
...
-      if (!issuer) {
-        throw new Error('No issuer found for attestion code')
-      }
-
-      const existingCode = yield call(getCodeForIssuer, issuer)
+      const existingCode = yield call(isCodeAlreadyAccepted, code)
       if (existingCode) {
         Logger.warn(TAG + '@attestationCodeReceiver', 'Code already exists in store, skipping.')
-        if (action.inputType === CodeInputType.MANUAL) {
+        if ([CodeInputType.MANUAL, CodeInputType.DEEP_LINK].indexOf(action.inputType) !== -1) {
```
jmrossy: This seems like in inefficient way to do an `or` check ;)

jmrossy (approved):
Looks good, just one small comment

packages/mobile/src/identity/verification.ts
i1skn: @jmrossy  I usually do this way, so when you add more stuff the `if` statement isn't growing to multiple `... === CodeInputType.MANUAL` over and over again. But this got me curious, how slow is it compare to `or` - will update here later with the results.

i1skn: It is indeed way slower (around 25-40 times on a high number of iterations). Will update it :) Thanks for pointing this out!

PR #1779, opened 2019-11-19
Fix yargs import

nambrot:
### Description

Fix yargs import as per https://github.com/yargs/yargs/issues/1439 to remove the annoying warning


nategraf (approved):
Thanks for taking care of this

PR #1786, opened 2019-11-19
Update .env to point to master Geth tag and remove deprecated VERIFICATION_POOL_URL

nategraf:
### Description

`verification_pool_url` and `verification_rewards_url` were removed in `celo-blockchain`. This PR removes them here in `monorepo`

### Backwards compatibility

Updated monorepo cannot be used to deploy older versions of `celo-blockchain`


nambrot (approved):
BEAUTIFUL 🙏 

Should note the backwards compatibility here

PR #1797, opened 2019-11-20
Attestation service deploy updates + support attestation signers

### LINKED ISSUE #1796: Clabs validators should use a separate attestation key
### Expected Behavior

### Current Behavior

What is the current behavior?

### LINKED ISSUE #1813: Users should be able to complete attestations with an attestation key
### Expected Behavior

Should not do that

### Current Behavior

We assume the account key is the attestation signere

nambrot:
### Description

This PR makes some minor updates to the deploy of our attestation service and corresponding fixes:

- It tries to fetch the account address at startup to highlight if it wasn't configured
- It by default makes CLabs validators register a separate attestation key
- It fixes a contractkit bug where we would assume that the account address is the same as the address signing the attestation
- It updates the helm chart to our latest configuration

### Tested

- [x] on `namoffchainreveal`

### Related issues

- Fixes #1796
- Fixes #1813


packages/helm-charts/attestation-service/templates/attestation.statefulset.yaml
```diff
@@ -78,7 +80,11 @@ spec:
         - name: CELO_PROVIDER
           value: https://{{ .Release.Namespace }}-forno.{{ .Values.domain.name }}.org
         - name: APP_SIGNATURE
-          value: {{ .Values.attestation_service.sms_retriever_hash_code }}}
+          value: {{ .Values.attestation_service.sms_retriever_hash_code }}
+        - name: SMS_PROVIDERS
```
timmoreton, line 84: presuming this arrived in another PR

timmoreton (approved):

packages/helm-charts/attestation-service/templates/attestation.statefulset.yaml
nambrot: Yup

PR #1798, opened 2019-11-20
Test that events that are emitted during block finalization are included in the blocks

mrsmkl:
### Description

Just a simple test to check if these events are actually included in the blocks.

### Other changes

Also added some new events that are emitted when paying epoch rewards.

### Related issues

- Test for https://github.com/celo-org/celo-blockchain/pull/584


packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -687,6 +687,15 @@ describe('governance tests', () => {
         }
       }
     })
+
+    it('should have emitted transfer events when paying epoch rewards', async () => {
```
asaj: Can we test that the correct events are emitted, and also that the events are only emitted during epoch blocks?

mrsmkl: Ok, adding more checks.

asaj: Ah, didn't realize that we don't actually emit the main events that we'd want to record.

More specifically, can we check that the following events are emitted in finalizing the last block of the epoch, and that no events are emitted in finalizing any other block?

`TargetVotingYieldUpdated` (needs to be added to `EpochRewards.updateTargetVotingYield`)
`ValidatorEpochPaymentDistributed` (needs to be added to `Validators.distributeEpochPayementsFromSigner`)
`EpochRewardsDistributedToVoters` (needs to be added to `Election.distributeEpochRewards`)

And can you please add unit tests that check for these events as well?

Thanks!

asaj (approved):
Love it!

diminator (approved):
LGTM

PR #1806, opened 2019-11-20
Load test clients

### LINKED ISSUE #982: CLabs SBAT operate a fleet of light clients generating transactions
- [x] resurrect load test client 
- [x] fund accounts for these clients in migrations
- [x] deploy in its own helm chart

tkporter:
### Description

Load test clients!

This introduces two .env vars:

`LOAD_TEST_CLIENTS`:

The number of load test client accounts. These accounts (generated using the mnemonic) are given cGLD in the genesis & cUSD during contract migrations. This is the default number of load test clients that are deployed, but that can be overridden with a cmd flag

`LOAD_TEST_TX_DELAY_MS`:

The number of milliseconds a load test client will sleep between transaction sends.

#### Overview

Load test clients are ultralight clients that are deployed as a statefulset. Each pod has two containers that run simultaneously: an ultralight geth client, and a celotool container running the command `celotool geth simulate-client`. The client simulation uses contractkit to send transactions every `LOAD_TEST_TX_DELAY_MS` milliseconds (unless overridden). The currency and gas currency (cGLD/cUSD) for a transaction are chosen randomly. Transactions from a load test client at index `i` will be sent to the load test client at index `(i + 1) % (total num of clients)`, so the funds will be circulated only amongst load-test accounts.

Here's the help for deploying load-test

```
celotooljs deploy initial load-test

deploy load-test

Options:
  --version                     Show version number                                                                                                    [boolean]
  --verbose                     Whether to show a bunch of debugging output like stdout and stderr of shell commands                  [boolean] [default: false]
  --yesreally                   Reply "yes" to prompts about changing staging/production (be careful!)                                [boolean] [default: false]
  --help                        Show help                                                                                                              [boolean]
  --celo-env, -e                the environment in which you want to execute this command                                                             [required]
  --blockscout-measure-percent  Percent of transactions to measure the time it takes for blockscout to process a transaction. Should be in the range of [0, 100]
                                                                                                                                          [number] [default: 30]
  --delay                       Number of ms a client waits between each transaction, defaults to LOAD_TEST_TX_DELAY_MS in the .env file  [number] [default: -1]
  --replicas                    Number of load test clients to create, defaults to LOAD_TEST_CLIENTS in the .env file                     [number] [default: -1]
```

### Tested

Ran my own testnet, ensured that the genesis & contract migration set the correct account balances. Spun up load-clients, upgraded them, destroyed them, etc. Made sure transactions were succeeding by looking at the logs & by looking at blockscout. You can see it in action right now at https://trevor3-blockscout.celo-networks-dev.org/ (if it doesn't seem like it's happening I probably took down the light clients).

### Other changes

* I wanted to use some helm templates from the `testnet` chart in the `load-test` chart, so I created a common helm chart for shared templates called `common`.

* I also finally moved the bootnode address to be generated with the bootnode account type, and not the load test account type

* Removed a duplicate of the sleep function

* I also saw that contractkit was only aware of the public variable `gasPriceMinimum` (which corresponds to the cGLD gas price, [link](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/GasPriceMinimum.sol#L21)) in `GasPriceMinimum.sol`, and not `getGasPriceMinimum(address tokenAddress)` (which will provide the gas price for any token, [link](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/GasPriceMinimum.sol#L72)), so I added that. I ended up not needing to use that function in this PR after all, but thought it'd be good to keep in.

### Related issues

- Fixes #982 

### Backwards compatibility

Requires new genesis & contract deployments so that load test accounts are fauceted


packages/helm-charts/common/README.md
```diff
@@ -0,0 +1,3 @@
+# common
+
+This contains common helper templates to be used across different helm charts.
```
nambrot, line 3: 🙏

packages/helm-charts/common/templates/_helpers.tpl
```diff
@@ -0,0 +1,30 @@
+{{- define "common.init-genesis-container" -}}
```
nambrot, line 1: Since we have the static nodes on GCP, should we just pull them into their own init container as well to share?

packages/helm-charts/load-test/templates/load-test.statefulset.yaml
```diff
@@ -0,0 +1,167 @@
...
+      component: load-test
+  template:
+    metadata:
+      labels:
+        app: load-test
+        release: {{ .Release.Name }}
+        component: load-test
+    spec:
+      initContainers:
+      - name: generate-keys
```
nambrot, line 42: While we are at it, i feel like we could pull this into `common` as well?

packages/celotool/src/lib/geth.ts
```diff
@@ -468,126 +467,168 @@ const measureBlockscout = async (
   }
 }
 
+export const transferCeloGold = async (
```
nambrot: I was never quite a fan of putting all the code for the light client here in this massive file, maybe consider separating it

packages/celotool/src/lib/geth.ts
```diff
@@ -209,12 +231,12 @@ const validateGethRPC = async (
...
   const transaction = await web3.eth.getTransaction(txHash)
-  if (!transaction.from || transaction.from.toLowerCase() !== from.toLowerCase()) {
-    handleError(!transaction.from || transaction.from.toLowerCase() !== from.toLowerCase(), {
-      location: '[GethRPC]',
-      error: `Expected "from" to equal ${from}, but found ${transaction.from}`,
-    })
-  }
+  const txFrom = transaction.from.toLowerCase()
+  const expectedFrom = from.toLowerCase()
+  handleError(!transaction.from || expectedFrom !== txFrom, {
```
nambrot: We have an `import { eqAddress } from '@celo/utils/lib/address'` that should avoid people having to lowerCase things

packages/celotool/src/lib/geth.ts
```diff
@@ -227,9 +249,11 @@ const checkBlockscoutResponse = (
 
   handleError(json.status !== '1', { location, error: `Invalid status: expected '1', received` })
   handleError(!json.result, { location, error: `No result found: receive ${json.status.result}` })
-  handleError(json.result.from !== from, {
+  const resultFrom = json.result.from.toLowerCase()
+  const expectedFrom = from.toLowerCase()
+  handleError(resultFrom !== expectedFrom, {
```
nambrot: use `eqAddress`

packages/celotool/src/lib/geth.ts
```diff
@@ -468,126 +467,168 @@ const measureBlockscout = async (
...
+  } = {}
+) => {
+  const kitGoldToken = await kit.contracts.getGoldToken()
+  return kitGoldToken.transfer(toAddress, amount.toString()).send({
+    from: fromAddress,
+    ...txOptions,
+  })
+}
+
+export const transferCeloDollars = async (
```
nambrot: I know you just copied it, but you could probably just have a single `transferToken` function instead of having to duplicate the methods here

nambrot (approved):
Just minor arrangement questions that could also be solved in a different PR, looks great overall, thanks for reviving it!

nambrot:
I know we talked about how these clients should be able to connect to full nodes that are not the specified static nodes, but if we could somehow verify that, that would be great

packages/celotool/src/cmds/geth/simulate_client.ts
```diff
@@ -1,141 +1,80 @@
...
-  const { web3, goldToken, stableToken } = web3AndContracts!
-
-  // Needs to be called only once per private key
-  // to create a file in keystore containing all the needed to geth info
-  try {
-    await web3.eth.personal.importRawKey(privateKey, '')
-  } catch (ignored) {
-    // ignore
-  }
+  // So we can transactions to another load testing account
```
nambrot: *send transactions

nambrot (approved):

tkporter:
Update: hoping to merge this today, e2e sync test is failing so I will have to figure that out

tkporter:
Closing in favor of #2073, which included this PR

PR #1809, opened 2019-11-20
Update examples.md

matiasbn:
Changed methods from contract to contracts. Without it it would not work.

### Description

### Tested

### Other changes

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


annakaz (approved):
Thank you for the contribution, @matiasbn!

PR #1814, opened 2019-11-20
Deploy an oracle cronjob

### LINKED ISSUE #981: CLabs SBAT operate a single oracle with Brownian motion

yerdua:
### Description

Helm chart and celotool command to deploy an oracle cronjob to a network. 

Note: In order to get this working for `baklavastaging`, I merged in @tkporter's branch trevor/baklavastaging-nov-21. So, that branch should be merged into master first. Or this one should be merged into that one instead.

### Tested

Tested by running the celotool command to deploy an oracle to `baklavastaging`, and confirming that the rates are reported to the network correctly.

### Other Changes

Update contractkit's version to accurately reflect what's been published on npm

### Related issues

- Fixes #981


.env
```diff
@@ -32,6 +32,12 @@ GETH_BOOTNODE_DOCKER_IMAGE_TAG="master"
 CELOTOOL_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/celo-monorepo"
 CELOTOOL_DOCKER_IMAGE_TAG="celotool-dfdc3e8b26e98aa294b27e2b5621c184488a10db"
 
+CELOCLI_STANDALONE_IMAGE_REPOSITORY="gcr.io/celo-testnet/celocli-standalone"
+# TODO (yerdua): Change this to something actually reasonable
+CELOCLI_STANDALONE_IMAGE_TAG="yerduaoracletest"
```
yerdua: This really doesn't seem ideal. I needed a docker image that can run the cli. The existing dockerfile has deeply hardcoded references to alfajores and runs a full node. In order to get *something* here, I built an image and pushed it manually.

yerdua:
There's a few incomplete things in this PR. The primary one is the handling of the docker image for `celocli`. I've merged the needed cli changes into master, but they're not yet published on npm. How best to handle tagging, versions, and the docker image building are all kind of opaque to me, so I'd appreciate some advice on how to do those right.

Also, there is no default oracle process. There's a docker image in a private repo, for my testnet, and for baklava. In an ideal world, there should probably be some default image, maybe just spitting out the exchange rate value specified in the config. Given the tight deadline, I think it's ok if the command fails when there's no oracle image for now.

packages/celotool/src/lib/oracle.ts
```diff
@@ -0,0 +1,30 @@
...
+    celoEnv,
+    releaseName(celoEnv),
+    helmChartPath,
+    helmParameters(celoEnv)
+  )
+}
+
+function helmParameters(celoEnv: string) {
+  return [
+    `--set celotool.image.repository=${fetchEnv('CELOTOOL_DOCKER_IMAGE_REPOSITORY')}`,
```
tkporter: nit: some of these string literals are already found in `envVar`, if they aren't we should add them and use them to avoid hardcoding strings

packages/celotool/src/lib/oracle.ts
```diff
@@ -0,0 +1,30 @@
...
+    `--set celotool.image.tag=${fetchEnv('CELOTOOL_DOCKER_IMAGE_TAG')}`,
+    `--set mnemonic="${fetchEnv(envVar.MNEMONIC)}"`,
+    `--set oracle.image.repository=${fetchEnv('ORACLE_DOCKER_IMAGE_REPOSITORY')}`,
```
tkporter: is this used? same goes for any use of `ORACLE_DOCKER_IMAGE_REPOSITORY`

.env
tkporter: Could you add both env vars to all .env files?

packages/helm-charts/oracle/templates/oracle.cronjob.yaml
```diff
@@ -0,0 +1,76 @@
...
+  schedule: "{{ .Values.oracleSchedule }}"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      backoffLimit: 1
+      template:
+        spec:
+          initContainers:
+          - name: get-current-price
+            image: gcr.io/celo-testnet/oracle:{{ .Release.Namespace }}
```
tkporter: Ah I think I get it now- should we set the repository and tag here similar to what we do for geth or celotool by having it depend on .env variables?

packages/helm-charts/oracle/values.yaml
```diff
@@ -0,0 +1,4 @@
+nodeSelector: {}
+
+# Expects cron schedule syntax
+oracleSchedule: "*/1 * * * *"
```
tkporter: might be helpful to add a comment what this schedule means in readable terms for people like myself that never remember the cron schedule structure

tkporter (commented):
Couple small changes pending the new cli version

packages/helm-charts/oracle/values.yaml
yerdua: Excellent idea. I had to look it up myself.
Do you think this makes sense here, or should it be an envVar like all the other ones?

packages/helm-charts/oracle/templates/oracle.cronjob.yaml
yerdua: Oops! Thank you! This is where those unused envVars were supposed to go.

yerdua:
Note: This branch has `trevor/baklavastaging-nov-21` merged into it so deployment could work.

jmrossy (commented):
Suggest adding a note at the top of the PR description that Trevor's branch changes must merge first before this :)

.env.baklavastaging
```diff
@@ -0,0 +1,107 @@
...
+
+CELOCLI_STANDALONE_IMAGE_REPOSITORY="gcr.io/celo-testnet/celocli-standalone"
+CELOCLI_STANDALONE_IMAGE_TAG="0.0.30-beta2"
+
+# Schedule for an oracle deployed via celotool, expressed in crontab syntax
+# This schedule is "every 5th minute"
+ORACLE_CRON_SCHEDULE="*/5 * * * *"
+
+ORACLE_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/oracle"
+ORACLE_DOCKER_IMAGE_TAG="baklavastaging"
```
yerdua: The origin of this file is @tkporter's branch for baklavastaging deployment. It's possible that some vars here may have changed since I copied this over. The only changes I made that are required for the oracle deployment are these highlighted variables.

.env.baklavastaging
```diff
@@ -0,0 +1,107 @@
...
```
yerdua, line 49: The origin of this file is @tkporter's branch for baklavastaging deployment, as it is today (Dec 2). These vars are the only ones I added:

```
CELOCLI_STANDALONE_IMAGE_REPOSITORY
CELOCLI_STANDALONE_IMAGE_TAG
ORACLE_CRON_SCHEDULE
ORACLE_DOCKER_IMAGE_REPOSITORY
ORACLE_DOCKER_IMAGE_TAG
```

So, if there are merge conflicts in this file that do not involve these variables, the resolution is to pick whatever the other branch has.

packages/protocol/migrations/08_stabletoken.ts
```diff
@@ -46,7 +47,7 @@ module.exports = deploymentForCoreContract<StableTokenInstance>(
 
     for (const oracle of config.stableToken.oracles) {
       console.info(`Adding ${oracle} as an Oracle for StableToken`)
-      await sortedOracles.addOracle(stableToken.address, oracle)
+      await sortedOracles.addOracle(stableToken.address, ensureHexLeader(oracle))
```
yerdua, line 50: Without the `0x`, the oracle silently doesn't get whitelisted correctly. This needs to be merged before `baklava` deployment.

.env.baklavastaging
yerdua: The only one of these that should be different for `baklava` is:
`ORACLE_DOCKER_IMAGE_TAG="baklava"`

That docker image already exists and is ready to go.

packages/celotool/src/cmds/deploy/destroy/oracle.ts
```diff
@@ -0,0 +1,16 @@
...
+import { DestroyArgv } from '../../deploy/destroy'
+
+export const command = 'oracle'
+
+export const describe = 'destroy the oracle package'
+
+export const builder = {}
+
+export const handler = async (argv: DestroyArgv) => {
+  await createClusterIfNotExists()
```
tkporter: You should be able to remove this line to decrease the run time-- if we're destroying we should assume that the cluster exists & if it doesn't there's no need in creating it

packages/helm-charts/oracle/templates/oracle.cronjob.yaml
```diff
@@ -0,0 +1,75 @@
...
+        spec:
+          initContainers:
+          - name: get-current-price
+            image: {{ .Values.oracle.image.repository }}:{{ .Values.oracle.image.tag }}
+            imagePullPolicy: IfNotPresent
+            command:
+              - sh
+              - "-c"
+              - |
+                echo `./current_rate.sh` > /celo/.celo/current_price
```
tkporter: nit: should be able to just run `./current_rate.sh > /celo/.celo/current_price` right?

tkporter (approved):
Super minor comments, nice!

PR #1817, opened 2019-11-20
Support re-requesting when we are outside of the randomness availability window

### LINKED ISSUE #1815: Users should be able to re-request attestations if they are out of the randomness window
### Expected Behavior

- You should be able to re-request

### Current Behavior

You can't re-request until the attestation has expired

nambrot:
### Description

We currently forced users to wait for an attestation to expire if they could not select for an unselected request due to them being outside the randomness availability window. This PR allows for a user to overwrite the unselected request since it is effectively useless at that point.

I also increased the randomness availability window to match the attestation expiry in any case

### Tested

- Unit tests

### Related issues

- Fixes #1815 

### Backwards compatibility

- Yes


asaj (commented):
Why don't we just couple these, so that we keep randomness around until the attestation request is expired?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -601,4 +602,12 @@ contract Attestations is
     return (attestation.status == AttestationStatus.Incomplete &&
       !isAttestationExpired(attestation.blockNumber));
   }
+
+  function isAttestationRequestSelectable(uint256 attestationRequestBlock)
+    internal
+    view
+    returns (bool)
+  {
+    return block.number < attestationRequestBlock.add(getRandom().randomnessBlockRetentionWindow());
```
timmoreton, line 611: strictly also `block.number >= attestationRequestBlock + 4 && ` ? right?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -166,7 +166,8 @@ contract Attestations is
 
     require(
       state.unselectedRequests[msg.sender].blockNumber == 0 ||
-        isAttestationExpired(state.unselectedRequests[msg.sender].blockNumber),
+        isAttestationExpired(state.unselectedRequests[msg.sender].blockNumber) ||
+        !isAttestationRequestSelectable(state.unselectedRequests[msg.sender].blockNumber),
       "There exists an unexpired, unselected attestation request"
```
timmoreton, line 171: update msg: `that can be selected` ?

timmoreton (approved):

packages/protocol/contracts/identity/Attestations.sol
nambrot: Strictly yes, but `4` should almost always be smaller

PR #1820, opened 2019-11-21
Update with new istanbul lookback window size flag

asaj:
### Description

Needed for https://github.com/celo-org/celo-blockchain/pull/618/


asaj:
LGTM!

PR #1822, opened 2019-11-21
Update migration config to correct protocol parameters

asaj:


timmoreton (approved):
LGTM. Haven’t checked it against the doc to see if there’s anything not changed here that we want to change.

PR #1825, opened 2019-11-21
Fix return

H34D:
### Description

return at least true, when the function is returning bool


m-chrzan (approved):

H34D:
@m-chrzan i am not authorised to merge

PR #1827, opened 2019-11-21
Bump versions for recently published packages

yerdua:
### Description

New versions of `@celo/celocli`, `@celo/contractkit` and `@celo/utils` were published today. This commit bumps the patch version up, and specifies `-dev` for those packages. All other packages depending on these had their dependency versions bumped up.

### Tested

Ran these commands, which all succeeded:
```
yarn
yarn build
yarn lint
```


cmcewen (approved):

jmrossy (changes requested):
Please coordinate with Nam for his PR here: https://github.com/celo-org/celo-monorepo/pull/2114
I think some of these version nums are now out of date

PR #1467, opened 2019-10-24
Allow a specified address to disable/enable the Exchange 

m-chrzan:
### Description

For purposes of Baklava, we might need to periodically turn the Exchange off. This PR adds `freeze` and `unfreeze` functions to the Exchange, allowing an address specified at migration time (and modifiable via `setFreezer`) to disable the `exchange` function.

### Tested

Unit tests.

### Other changes

This PR adds a new helper contract, Freezable, that specifies a `freezer` address and provides modifiers to disable certain contract functions if the contract becomes "frozen".

This is in anticipation of a similar change needed in LockedGold.

### Related issues

- Begins to address #1431 

### Backwards compatibility

Not backwards compatible, changes Exchange storage layout.

PR #1828, opened 2019-11-21
Allow a specified address to disable/enable rewards distribution

### LINKED ISSUE #1431: Stake off admin SBAT enable/disable rewards and exchange
### Expected Behavior

* Specify an address at `LockedGold` initialization time that is permitted to toggle a boolean to enable or disable rewards (if necessary, modify blockchain to observe this setting).
* Do the same in `Exchange` to permit exchanges

m-chrzan:
### Description

For Baklava, we want the organizer to be able to turn off rewards distribution. This PR uses the previously introduced `Freezable` interface to make `updateTargetVotingYield` freezable. Celo-blockchain will, upon reverting on calling the frozen function, produce an epoch block without distributed rewards.

### Tested

Unit and e2e tests.

### Other changes

Reorganized helper functions in e2e governance test.

### Related issues

- Fixes #1431 

### Backwards compatibility

Not backwards compatible, smart contract has new storage layout.


packages/protocol/migrations/14_epoch_rewards.ts
```diff
@@ -3,10 +3,13 @@ import { deploymentForCoreContract } from '@celo/protocol/lib/web3-utils'
...
 import { toFixed } from '@celo/utils/lib/fixidity'
 import { EpochRewardsInstance } from 'types'
+const truffle = require('@celo/protocol/truffle-config.js')
 
-const initializeArgs = async (): Promise<any[]> => {
+const initializeArgs = async (networkName: string): Promise<any[]> => {
+  const network: any = truffle.networks[networkName]
   return [
     config.registry.predeployedProxyAddress,
+    network.from,
```
nategraf, line 12: `network.from` is validator 0. I think this should be a different key because it's validator key is quite sensitive.

packages/protocol/test/governance/epochrewards.ts
```diff
@@ -91,6 +91,7 @@ contract('EpochRewards', (accounts: string[]) => {
```
nategraf, line 91: Freezing should be tested in this file

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -694,6 +718,37 @@ describe('governance tests', () => {
...
   })
 
+  describe('when rewards distribution is frozen', () => {
+    before(restart)
+
+    let epoch: number
+    let blockFrozen: number
+    let latestBlock: number
+
+    beforeEach(async function(this: any) {
```
nategraf: I would think this should be a `before` instead of `beforeEach` because you only need to do it once. Or am I mistaken?

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -362,7 +369,7 @@ contract EpochRewards is Ownable, Initializable, UsingPrecompiles, UsingRegistry
    *   voting Gold fraction.
    * @dev Only called directly by the protocol.
    */
-  function updateTargetVotingYield() external {
+  function updateTargetVotingYield() external onlyWhenNotFrozen {
```
nategraf, line 372: For robustness, `calculateTargetEpochPaymentAndRewards` should return `0, 0` when frozen as no rewards should be distributed.

nategraf (commented):
#1431 mentions being able to freeze the exchange, so I've edited the description to make #1431 as related, but not fixed.

m-chrzan:
@nategraf, thanks for the comments!

> #1431 mentions being able to freeze the exchange, so I've edited the description to make #1431 as related, but not fixed.

#1467 already did the Exchange portion, so this PR will close out the issue.

packages/protocol/test/governance/epochrewards.ts
m-chrzan: The function whose behavior is modified (`updateTargetVotingYield`) can only be called from within geth (`require(msg.sender == address(0))`), so from these ganache-based tests we won't be able to observe a change in behavior. This is tested in the e2e tests instead.

packages/protocol/migrations/14_epoch_rewards.ts
m-chrzan: Any ideas what else this could be? validator-0 is fairly convenient as that has been the address we use in our tooling to manage test networks (throughout celotool, I believe that's the address that has faucet balances), but agree that we might need to think differently about baklava.

nategraf: It could be any key, including one generated on the fly. You could generate it from the same mnemonic and might create a new `AccountType.ADMIN` https://github.com/celo-org/celo-monorepo/blob/master/packages/celotool/src/lib/generate_utils.ts#L23

packages/protocol/test/governance/epochrewards.ts
nategraf: Ah ok.

nategraf (approved):
It would be good to make the key management change I mentioned, but overall LGTM. Thanks for addressing my comments!

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -217,6 +217,65 @@ describe('governance tests', () => {
     assertAlmostEqual(currentBalance.minus(previousBalance), expected)
   }
 
+  const waitForBlock = async (blockNumber: number) => {
```
asaj, line 219: Thanks for cleaning this up!

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -372,6 +379,10 @@ contract EpochRewards is Ownable, Initializable, UsingPrecompiles, UsingRegistry
    * @return The per validator epoch payment and the total rewards to voters.
    */
   function calculateTargetEpochPaymentAndRewards() external view returns (uint256, uint256) {
+    if (frozen) {
```
asaj, line 382: nit: This *shouldn't* strictly be necessary, since if one epoch rewards action fails we won't execute any of them.

packages/protocol/migrations/14_epoch_rewards.ts
asaj: What do we currently set other things to? I.e. the approver, reserve spender, etc.? I agree validator 0 isn't the most robust choice, but I believe it's currently the standard choice

asaj (approved):

packages/protocol/contracts/governance/EpochRewards.sol
m-chrzan: Yep. @nategraf recommended this change for robustness.

One nice thing about how this is implemented is that after Baklava we can just delete the `baklava/` directory and then remove any code that the compiler complains about.

PR #1835, opened 2019-11-22
Fix protocol tests

jfoutts-celo:
### Description

Accidentally committed it.only() is causing the protocol tests to consist of only:
https://github.com/celo-org/celo-monorepo/commit/4570fde48cd2e283b91ffc1578190127ca93019d#diff-ea8b99cb616e766d769dc8b98bda7decR362

### Tested

packages/protocol `yarn test` succeeds with 1065 passing test instead of 1.

### Other changes

Passing async functions to describe() is a bug: https://github.com/mochajs/mocha/issues/2975

The only occurrences of async describe() arguments in celo-monorepo were in the protocol tests, and they were all innocuous due to never await-ing.


m-chrzan (approved):

packages/protocol/test/governance/election.ts
```diff
@@ -901,23 +900,6 @@ contract('Election', (accounts: string[]) => {
       })
     })
 
-    describe('when different random values are provided', () => {
```
asaj, line 904: Now that we're not using this, can we remove all references to `random` in this file?

asaj (approved):

packages/protocol/test/governance/election.ts
jfoutts-celo: Even `registry.setAddressFor(CeloContractName.Random, random.address)` ? I'll look into in separate PR.

Issue #1696, opened 2019-11-13
Error - celocli transfer throws Invalid Number Value when amountInWei is larger than 1e21 (or 1000000000...00000)


diminator:
## Expected Behavior

Devs/Users can transfer more than 1e21 cGLD or cUSD through the cli. Exponential notation has good UX.

## Current Behavior
```
$ ../cli/bin/run account:transfergold --from 0x47e172f6cfb6c7d01c1574fa3e2be7cc73269d95 --to 0x47e172f6cfb6c7d01c1574fa3e2be7cc73269d95 --amountInWei 1000000000000000000000                      
```
```
Sending Transaction: gold.Transfer... !                                                                                                                                                                                                                        
Error: invalid number value (arg="value", coderType="uint256", value="1e+21")

PR #1840, opened 2019-11-23
Various improvements to the CLI, allow voters to revote for a group

### LINKED ISSUE #1829: CLI prints validator score as BigNumber
### Expected Behavior

BigNumber is printed as a human readable number

### Current Behavior

Screen Shot 2019-11-21 at 9 34 11 PM

Also, I think `Name` is set during contract migrations, so we should look into what the problem is there.

asaj:
### Description

Miscellaneous CLI improvements:
- Use `toFixed` rather than `toString`. The latter can format large numbers in exponential notation, causing transactions to fail to send
- Print BigNumbers as string rather than BigNumber objects
- Make transfer commands consistent with the rest of the CLI
- Checks in election:vote, lockedgold:lock
- Fix validation of key lengths
- Fix validatorgroup:member examples
- Infer ECDSA key when registering validator
 
Fix bug in Election.sol that prevented voting twice for the same group, add unit test.

### Tested

Ran a ganache chain with `yarn test:livechain` and ran CLI commands against it
 
### Other changes

- Make devchain command names consistent

### Related issues

- Fixes #1829, #1696


packages/cli/src/utils/helpers.ts
```diff
@@ -3,6 +3,7 @@ import { Block } from 'web3/eth/types'
 import { failWith } from './cli'
 
 export async function nodeIsSynced(web3: Web3): Promise<boolean> {
+  return true
```
lucasege, line 6: Is this intentional? Maybe should be handled by `process.env.NO_SYNCCHECK` unless there's a good reason otherwise

packages/cli/src/commands/lockedgold/lock.ts
```diff
@@ -28,14 +28,18 @@ export default class Lock extends BaseCommand {
 
     this.kit.defaultAccount = address
     const value = new BigNumber(res.flags.value)
+
+    await newCheckBuilder(this)
+      .addCheck(`Value [${value.toFixed()}] is >= 0`, () => value.gt(0))
```
lucasege: Nitpicky, but text says >= 0 while code is gt, not gte (assuming gte is an available function).
Unfamiliar with the greater context here so I'm not sure if this is at all significant. I'm assuming not

lucasege (approved):
Looks good, only small nitpicks

packages/cli/src/utils/helpers.ts
asaj: Not intentional, sent a small cleanup commit

packages/cli/src/commands/lockedgold/lock.ts
asaj: Whoops, nice catch, fixed.

PR #1847, opened 2019-11-24
Fix elect validators migration, deploy integration

asaj:
### Description

This PR fixes the elect validators migration by updating the validators.

### Tested

- Ran modified end-to-end tests with 10 validators

### Other changes

- Remove unused shuffle in elect validators
- Make `Governance` owner of `Accounts`, `EpochRewards`
- Don't override reserve balance in end-to-end tests


.env.integration
```diff
@@ -1,6 +1,6 @@
 ENV_TYPE="integration"
 
-GETH_VERBOSITY=3
+GETH_VERBOSITY=4
```
nambrot: Can we move back to 3?

packages/celotool/src/lib/generate_utils.ts
```diff
@@ -91,7 +91,7 @@ export const privateKeyToStrippedAddress = (privateKey: string) =>
 
 const validatorZeroBalance = fetchEnvOrFallback(
   envVar.VALIDATOR_ZERO_GENESIS_BALANCE,
-  '100010011000000000000000000'
+  '103010030000000000000000000'
 ) // 100,010,011 CG
```
nambrot: Is this comment still valid then?

nambrot (approved):
🙏

PR #1848, opened 2019-11-25
Serialize to hex string

nambrot:
### Description

The default gateway fee was not properly encoded

```
> lib.bytes.fromNat(new BN(10000).toString())
'0x0000'
```

### Tested

- Deployed attestation service to integration

### DIFF
--- a/packages/contractkit/src/providers/celo-private-keys-subprovider.ts
+++ b/packages/contractkit/src/providers/celo-private-keys-subprovider.ts
@@ -124,7 +124,7 @@ export class CeloPrivateKeysWalletProvider extends PrivateKeyWalletSubprovider {
       txParams.gatewayFeeRecipient = await this.getCoinbase()
     }
     if (!isEmpty(txParams.gatewayFeeRecipient) && isEmpty(txParams.gatewayFee)) {
-      txParams.gatewayFee = DefaultGatewayFee.toString()
+      txParams.gatewayFee = DefaultGatewayFee.toString(16)
     }
     debug(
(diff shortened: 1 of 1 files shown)


packages/contractkit/src/providers/celo-private-keys-subprovider.ts
```diff
@@ -124,7 +124,7 @@ export class CeloPrivateKeysWalletProvider extends PrivateKeyWalletSubprovider {
       txParams.gatewayFeeRecipient = await this.getCoinbase()
     }
     if (!isEmpty(txParams.gatewayFeeRecipient) && isEmpty(txParams.gatewayFee)) {
-      txParams.gatewayFee = DefaultGatewayFee.toString()
+      txParams.gatewayFee = DefaultGatewayFee.toString(16)
```
nategraf, line 127: It seems like this should be prefixed with '0x'. If it works without it though, I guess I am mistaken.
```suggestion
      txParams.gatewayFee = '0x' + DefaultGatewayFee.toString(16)
```

nategraf (approved):

packages/contractkit/src/providers/celo-private-keys-subprovider.ts
nambrot: So it works because of this very confusing implmentation

https://github.com/MaiaVictor/eth-lib/blob/master/src/bytes.js#L77

Basically, `fromNat`/`rlp.encode` accepts any value that is a string and has an even number of chars. I figure we want to change this behavior and make it safer, but did not want to dig deeper right now

PR #1810, opened 2019-11-20
Script for reading leaderboard data into database

mrsmkl:
### Description

A script for reading data needed to calculate leaderboard score
Reads the following things:
* Multiplier from the sheet (currently points to test sheet)
* Claimed accounts from account metadata
* Exchange rate from `SortedOracles` contract (probably should be moved into blockscout)

### Tested

Only manual testing so far. Needs credentials to access Google sheets.

### Related issues

- Fixes #1612

PR #1864, opened 2019-11-26
Leaderboard app with query

### LINKED ISSUE #1630: Users SBAT to see TGSTO leaderboard in the web app
Implement the leaderboard according to https://www.figma.com/file/HzDfZq4jvFNgUvPuPmeC80Ss/Website-3.0?node-id=1919%3A0 in the celo web page

diminator:
### Description

This PR requires branch `aaronmgdr/build4`

Implements the dynamic leaderboard app using a graphql query and leaderboard script:
https://github.com/celo-org/celo-monorepo/pull/1810

### Tested

Tested by running locally.

### Related issues

- Fixes #1630 

### To check

I'm not sure how `env-config.js.template` works


packages/web/src/dev/LeaderBoard.tsx
```diff
@@ -97,10 +94,11 @@ function sorter(alpha: Competitor, bravo: Competitor) {
   return -1
 }
 
-const PORTIONS = 8
+const Portions = 8
```
aaronmgdr, line 97: why the change? for consts all caps is really more common, with Pascalcase I tend to think interfaces or classes

packages/web/env-config.js.template
```diff
@@ -9,7 +9,7 @@ const environments = {
       projectId: 'celo-faucet',
     },
     LEADERBOARD: {
-      uri: "http://localhost:8080/v1/graphql",
+      uri: "http://localhost:4000/graphiql",
```
aaronmgdr, line 12: I assume eventually this will be a non localhost URL?

aaronmgdr (approved):

packages/web/env-config.js.template
diminator: yes, it should be the one of `blockscout`

packages/web/src/dev/LeaderBoard.tsx
diminator: true, the linter was complaining, but we can revert

diminator:
Seems circleci is having some issues. probably `build4` branch needs to sync in master

PR #1854, opened 2019-11-25
Baklava docs improvements

aaitor:
### Description

This PR applies some improvements and address comments 

### Related issues

v2 of #1688

PR #1901, opened 2019-11-26
make run docker validator network also run on mac

H34D:
### Description

Made the script run on mac using iterm2


H34D:
this needs double checking by @aaitor

H34D:
handled here: https://github.com/celo-org/celo-monorepo/pull/1854 and here https://github.com/aaitor/celo-monorepo/pull/11

PR #1906, opened 2019-11-26
Add validator:status command to check if a validator signer is online and producing blocks

### LINKED ISSUE #1782: Validators SBAT reference getting started documentation on how to see if they're validating
### Expected Behavior

"Getting started" documentation to help validators understand that they're validating 

### Current Behavior

Validator has no way of knowing if they're validating.

nategraf:
### Description

Create a `validator:status` command to check the registration, election, and signing status of a validator.

### Tested

Manually tested against local testnet

### Backwards compatibility

Will not break anything. New command will work with networks that have parent seals in the block header.

### Related Issues

Fixes: https://github.com/celo-org/celo-monorepo/issues/1782


nategraf:
I've added a note to the `running-a-validator` doc to suggest using this command.

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+import { range } from 'lodash'
+import * as rlp from 'rlp'
+import { Block } from 'web3/eth/types'
+import { BaseCommand } from '../../base'
+import { newCheckBuilder } from '../../utils/checks'
+import { Flags } from '../../utils/command'
+
+export default class ValidatorOnline extends BaseCommand {
+  static description =
+    'Show information about whether the validator signer is elected and validating. This command will check that the validator meets the registration requirements, and its signer is currently elected and actively signing blocks.'
```
asaj, line 13: Wdyt of including this information in `validator:show`?

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  static flags = {
+    ...BaseCommand.flags,
+    signer: Flags.address({
```
asaj, line 17: It might be simpler to only accept a single argument, which is the validator account address.

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+      .runChecks()
+
+    // Determine if the signer is elected, and get their index in the validator set.
+    const election = await this.kit.contracts.getElection()
+    const signers = await election.getCurrentValidatorSigners()
+    const signerIndex = signers.map((a) => a.toLowerCase()).indexOf(signer.toLowerCase())
+    if (signerIndex < 0) {
+      // Determine whether the signer will be elected at the next epoch to provide a helpful error.
+      const frontrunners = await election.electValidatorSigners()
+      if (frontrunners.map((a) => a.toLowerCase()).indexOf(signer.toLowerCase()) >= 0) {
```
mcortesi: you can use [Array.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) with `eqAddress()` function

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+    await newCheckBuilder(this, signer)
+      .canSignValidatorTxs()
+      .signerMeetsValidatorBalanceRequirements()
+      .signerAccountIsValidator()
+      .runChecks()
+
+    // Determine if the signer is elected, and get their index in the validator set.
+    const election = await this.kit.contracts.getElection()
+    const signers = await election.getCurrentValidatorSigners()
+    const signerIndex = signers.map((a) => a.toLowerCase()).indexOf(signer.toLowerCase())
```
mcortesi: you can use [Array.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) with `eqAddress()` function

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
+import { Address } from '@celo/contractkit'
+import { flags } from '@oclif/command'
+import { cli } from 'cli-ux'
+import { range } from 'lodash'
```
mcortesi: Not sure if `lodash` is part of cli `package.json`....
And personally, i would avoid it :P

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+    console.info('Signer has been elected for this epoch')
+
+    if (flags.lookback <= 0) {
+      return
+    }
+
+    // Retrieve blocks to examine for the singers signature.
+    cli.action.start(`Retreiving the last ${flags.lookback} blocks`)
+    const latest = await this.web3.eth.getBlock('latest')
+    const blocks = await Promise.all(
```
mcortesi: Probably not a problem here, but in general it's better to use "bounded" concurrency. That is, have control on how many things you are doing at the same time, since resources are always limited (be it processor, bandwith, whatever). In this case, probably doesn't make sense. 

If not, there's a `concurrentMap` in @celo/utils

mcortesi (approved):
Check lodash, i don't think it's included in package.json...

The rest are minor details!

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+        this.error(`${flags.signer} is not the authorized signer for ${flags.validator}`)
+      }
+    } else {
+      signer = flags.signer!
+    }
+
+    // Use redundant checks to help the user diagnose issues.
+    await newCheckBuilder(this, signer)
+      .canSignValidatorTxs()
+      .signerMeetsValidatorBalanceRequirements()
```
asaj: I would expect us to fail this check whenever the validator has an authorized validator signer... 

Where else is this used?

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+      .signerAccountIsValidator()
+      .runChecks()
+
+    // Determine if the signer is elected, and get their index in the validator set.
+    const election = await this.kit.contracts.getElection()
+    const signers = await election.getCurrentValidatorSigners()
+    const signerIndex = signers.map((a) => a.toLowerCase()).indexOf(signer.toLowerCase())
+    if (signerIndex < 0) {
+      // Determine whether the signer will be elected at the next epoch to provide a helpful error.
```
asaj, line 73: This is great! Can we put this functionality in the `election` module instead?

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+        )
+      }
+    }
+    console.info('Signer has been elected for this epoch')
+
+    if (flags.lookback <= 0) {
+      return
+    }
+
+    // Retrieve blocks to examine for the singers signature.
```
asaj, line 91: Can we also check for block production? i.e. whether or not I'm the miner of the block?

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    console.info(`Signer has signed ${signedCount} of the last ${flags.lookback} blocks`)
+  }
+
+  // Check the parent seal in the given block header for a bit indicating at the given index.
+  private validatorIndexSignedParentHeader(block: Block, index: number) {
+    // See https://github.com/celo-org/celo-blockchain/blob/master/core/types/istanbul.go
+    // tslint:disable:no-bitwise
+    // @ts-ignore
+    const parentSealBitmap = rlp.decode('0x' + block.extraData.slice(66))[5][0]
```
asaj: Can you add a utility function in `@celo/utils` to decode this into an object?

packages/cli/src/commands/validator/status.ts
nategraf: I thought about it, but show has the nice advantage of being very simple, making only a single view call, and displaying more information about the attributes of the validator. At least in my mind this command is distinct enough to be it's own.

packages/cli/src/commands/validator/status.ts
nategraf: It's used here. https://github.com/celo-org/celo-monorepo/blob/master/packages/cli/src/commands/validator/register.ts#L32
I believe it should work. It calls `signerToAccount` on the provided signer then checks the balance requirement on the account that was retrieved. It's not strictly necessary though.

packages/cli/src/commands/validator/status.ts
nategraf: I want this command to work for signers that are no longer authorized in the smart contract, but are still elected for the epoch. Such as when someone has changed their keys and is waiting for the next epoch to come while running two validators, one with the old signing key and one with the new. I figure it will be helpful to know when your old validator is officially safe to take offline.

packages/cli/src/commands/validator/status.ts
nategraf: Added another check for that

packages/cli/src/commands/validator/status.ts
nategraf: Added a few functions to utils to over that

packages/cli/src/commands/validator/status.ts
nategraf: We could add a command like `election:validator-is-elected`, but I personally don't see a ton of value in that because `election:current` and `election:run` exist and combined with grep you can tell whether a specific validator is in the set.

packages/utils/src/istanbul.ts
```diff
@@ -0,0 +1,62 @@
+import BigNumber from 'bignumber.js'
+import { toChecksumAddress } from 'ethereumjs-util'
+import * as rlp from 'rlp'
+import { Address } from './address'
+
+// See https://github.com/celo-org/celo-blockchain/blob/master/core/types/istanbul.go
+export const ISTANBUL_EXTRA_VANITY_BYTES = 32
```
jmrossy: To avoid having the utils exports overcrowded, the general strategy has been to encapsulate each file's utils in a wrapper object and export that instead. See https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/phoneNumbers.ts#L234
and
https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/index.ts#L8

Admittedly we haven't always been great about this but we should be better as this package grows

nategraf: Thanks for  letting me know! I changed it

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,113 @@
...
+import { Flags } from '../../utils/command'
+
+export default class ValidatorStatus extends BaseCommand {
+  static description =
+    'Show information about whether the validator signer is elected and validating. This command will check that the validator meets the registration requirements, and its signer is currently elected and actively signing blocks.'
+
+  static flags = {
+    ...BaseCommand.flags,
+    signer: Flags.address({
+      description: 'address of the validator to check if elected and validating',
```
asaj: Are these descriptions swapped?

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,113 @@
...
+import { newCheckBuilder } from '../../utils/checks'
+import { Flags } from '../../utils/command'
+
+export default class ValidatorStatus extends BaseCommand {
+  static description =
+    'Show information about whether the validator signer is elected and validating. This command will check that the validator meets the registration requirements, and its signer is currently elected and actively signing blocks.'
+
+  static flags = {
+    ...BaseCommand.flags,
+    signer: Flags.address({
```
asaj, line 17: Can you make these flags exclusive? Example in validatorgroup:member

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,113 @@
...
+  async run() {
+    const res = this.parse(ValidatorStatus)
+
+    // Check that the specified validator or signer meets the validator requirements.
+    const checker = newCheckBuilder(this, res.flags.signer)
+    if (res.flags.validator) {
+      const account = res.flags.validator
+      checker.isValidator(account).meetsValidatorBalanceRequirements(account)
+    } else if (res.flags.signer) {
+      checker.signerMeetsValidatorBalanceRequirements().signerAccountIsValidator()
```
asaj: Rather than add additional checks, why don't we just pull the `account` from the signer and run the same checks on the account?

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,113 @@
...
+    // Determine if the signer is elected, and get their index in the validator set.
+    const election = await this.kit.contracts.getElection()
+    const signers = await election.getCurrentValidatorSigners()
+    const signerIndex = signers.map((a) => eqAddress(a, signer)).indexOf(true)
+    if (signerIndex < 0) {
+      // Determine whether the signer will be elected at the next epoch to provide a helpful error.
+      const frontrunners = await election.electValidatorSigners()
+      if (frontrunners.some((a) => eqAddress(a, signer))) {
+        this.error(
+          `Signer ${signer} is not elected for this epoch, but is currently winning in the upcoming election. Wait for the next epoch.`
```
asaj: `but would be elected to the validator set for the next epoch if an election were to be held now. Please wait until the next epoch.`

packages/cli/src/commands/validator/status.ts
```diff
@@ -0,0 +1,113 @@
...
+        )
+      } else {
+        this.error(
+          `Signer ${signer} is not elected for this epoch, and is not currently winning the upcoming election.`
```
asaj: `and would not be elected to the validator set for the next epoch if an election were to be held now.`

asaj (approved):

jmrossy (approved):

packages/cli/src/commands/validator/status.ts
nategraf: 👍

packages/cli/src/commands/validator/status.ts
nategraf: At this point in the code it is unclear whether the accounts given are valid and I wanted to avoid an opaque call reversion. I realized that as coded it still has that issue, so I've added an `isAccount` check.

packages/cli/src/commands/validator/status.ts
nategraf: Done.

packages/cli/src/commands/validator/status.ts
nategraf: Yes. Fixed.

PR #1917, opened 2019-11-26
Minor edits to full-node and other docs

nategraf:
### Description

Here are some edits I made while running through the process of starting a full-node


packages/docs/getting-started/faucet.md
```diff
@@ -30,19 +30,25 @@ You can also use the Celo Blockchain client to create and manage account keypair
...
+```
 
 It will prompt you for a passphrase, ask you to confirm it, and then will output your account address:
 
 `Address: <YOUR-ACCOUNT-ADDRESS>`
 
 This creates a keypair and stores it. Save this address to an environment variable, so that you can reference it later:
 
-`$ export CELO_ACCOUNT_ADDRESS=<YOUR-ACCOUNT-ADDRESS>`
+```bash
```
nambrot, line 49: Not in this PR, but I wonder if we should properly namespace this since this should only apply to Alfajores?

nambrot (approved):

Issue #1880, opened 2019-11-26
CLI command "election:current" prints score as a bignumber object


asaj:
Should convert to string via toFIxed(), see `validator:list` as an example

Issue #1874, opened 2019-11-26
CLI validatorgroup:list, validator:list show "name" as undefined


asaj:
### Expected Behavior

These commands pull the name from the `accounts` wrapper or don't show it.

### Current Behavior

"undefined"

PR #1918, opened 2019-11-26
cli: Fixes for #1880 and #1874

jfoutts-celo:
### Description

Fetches name for validator:list and adds "--no-truncate" option.

### Tested

Tested that validator names are correctly fetched.

### Related issues

#1880 and #1874


packages/cli/src/commands/validator/list.ts
```diff
@@ -7,25 +9,43 @@ export default class ValidatorList extends BaseCommand {
...
   async run() {
-    this.parse(ValidatorList)
+    const res = this.parse(ValidatorList)
 
     cli.action.start('Fetching Validators')
+    const accounts = await this.kit.contracts.getAccounts()
     const validators = await this.kit.contracts.getValidators()
     const validatorList = await validators.getRegisteredValidators()
 
+    var names = []
```
asaj: nit: consider:

```
const names = await Promise.all(validatorList.map((v) => accounts.getName(v.address)))

File: packages/cli/src/commands/validator/list.ts [hunk 2]
```diff
@@ -7,25 +9,43 @@ export default class ValidatorList extends BaseCommand {
...
+    for (let i = 0; i < validatorList.length; i++) {
+      names.push(accounts.getName(validatorList[i].address))
+    }
+    for (let i = 0; i < validatorList.length; i++) {
+      const name = await names[i]
+      validatorList[i].name = name ? name : ''
```
asaj: Would it be easier to modify the `Validators` contractkit wrapper to pull the name? If we're going to want to the name everywhere we pull validator info, that might save us from having to duplicate this code elsewhere

packages/cli/src/commands/validator/list.ts
jfoutts-celo: Nice!

packages/cli/src/commands/validator/list.ts
jfoutts-celo: Sounds good

packages/contractkit/src/wrappers/Validators.ts
```diff
@@ -244,7 +245,14 @@ export class ValidatorsWrapper extends BaseWrapper<Validators> {
   /** Get list of registered validators */
   async getRegisteredValidators(): Promise<Validator[]> {
     const vgAddresses = await this.getRegisteredValidatorsAddresses()
-    return Promise.all(vgAddresses.map((addr) => this.getValidator(addr)))
+    let validators = await Promise.all(vgAddresses.map((addr) => this.getValidator(addr)))
+    const accounts = await this.kit.contracts.getAccounts()
+    return Promise.all(
+      validators.map(async function(v) {
+        v.name = (await accounts.getName(v.address)) || ''
```
asaj: How about we set this in `getValidator` instead? And please update any other CLI commands that show validators to print the name if they don't already

jfoutts-celo: Looks like validator:show and election:current already show name

packages/contractkit/src/wrappers/Validators.ts
```diff
@@ -18,6 +18,7 @@ import {
 } from './BaseWrapper'
 
 export interface Validator {
+  name: string
```
mcortesi, line 21: Not sure if this is a "Validator" thing... instead what about "fetching the account" on the cli when you want to show the Validator details?

I don't know if this sounds like something any contractKit user would want, maybe just the cli.

mcortesi: hehehe.. I see that me and @asaj are thinking completely opposite things!.. 
It's not a big issue so either way work

asaj (approved):

packages/contractkit/src/wrappers/Validators.ts
asaj: Awesome! Will be nice to have those populated as well "for free" 😄

PR #1922, opened 2019-11-26
Have validators use ephemeral node keys by default

### LINKED ISSUE #1268: Validators SBAT use ephemeral node keys
### Expected Behavior

Validators do not need to explicitly set their nodekey, and a random one will be set automatically, and validator direct connections will still work.

### Current Behavior

Validators explicitly set their nodekeys to their consensus private key

nategraf:
### Description

Validators do not need to set their node keys to their validating key. This PR makes it the default not to in Helm deployments and e2e test. Additionally it removes references to setting the node key form documentation.

### Tested

Ran e2e tests and created a testnet then launched a validator and connected to it, confirming that the instructions work and nodes can bootstrap and sync. 

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/1268
- Related https://github.com/celo-org/celo-blockchain/pull/616

### Backwards compatibility

Changes are backwards compatible


asaj:
Can we try launching a small network on GCP and confirming we begin mining blocks?

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -326,7 +324,7 @@ export async function startGeth(gethBinaryPath: string, instance: GethInstanceCo
   }
 
   if (validating) {
-    gethArgs.push('--mine', '--minerthreads=10', `--nodekeyhex=${privateKey}`)
+    gethArgs.push('--mine', '--minerthreads=10')
```
asaj: Can you remove `--minerthreads=10` as well?

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -438,6 +436,29 @@ export async function initAndStartGeth(gethBinaryPath: string, instance: GethIns
   return startGeth(gethBinaryPath, instance)
 }
 
+// Add each validator as a peer of each other validator via the admin interface.
+async function connectValidatorPeers(gethConfig: GethTestConfig) {
```
asaj, line 441: Instead of connecting each validator to the other, can we connect them all to validator 0? This allows us to test that the validators still are able to peer with each other.

packages/helm-charts/testnet/templates/validators.statefulset.yaml
```diff
@@ -141,7 +141,6 @@ spec:
             --wsaddr 0.0.0.0 \
             --wsorigins=* \
             --wsapi=eth,net,web3,debug \
-            --nodekey=/root/.celo/pkey \
```
asaj, line 144: It looks like in this setup we would still have the private key at /root/.celo/pkey. Where is this file created?

packages/helm-charts/testnet/templates/_helpers.tpl
```diff
@@ -199,7 +199,6 @@ spec:
           --metrics \
           --mine \
           --networkid=${NETWORK_ID} \
-          --nodekey=/root/.celo/account/{{ .Node.name}}PrivateKey \
```
asaj, line 202: It looks like in this setup we would still have the private key at /root/.celo/account/*PrivateKey. Where is this file created?

packages/celotool/src/e2e-tests/utils.ts
nategraf: Done

tkporter: might need to push a new commit?

packages/helm-charts/testnet/templates/_helpers.tpl
tkporter: At least here the code is dead & is removed in the proxy infra PR that will be merged soon

packages/helm-charts/testnet/templates/_helpers.tpl
```diff
@@ -1,390 +0,0 @@
...
-heritage: {{ .Release.Service }}
-{{- end -}}
-
-{{- define "standard.short_labels" -}}
-app: {{ template "ethereum.name" . }}
-release: {{ .Release.Name }}
-{{- end -}}
-
-
-{{- define "celo.init-genesis-container" -}}
```
tkporter: Although a lot of this is dead code, we want to keep some of these (like this is still used by the validator statefulset). I deleted the dead parts of this file in the proxy infrastructure PR that's hopefully going to be merged soon, so if you want you can leave this file untouched and I can clean it up with that PR.

tkporter (commented):
could you also remove `--nodekey` from the `startup.sh` scripts in `packages/terraform-modules/testnet`?

packages/helm-charts/testnet/templates/validators.statefulset.yaml
```diff
@@ -56,11 +56,9 @@ spec:
           - |
             [[ $REPLICA_NAME =~ -([0-9]+)$ ]] || exit 1
             RID=${BASH_REMATCH[1]}
-            echo -n $RID >/root/.celo/replica_id
-            echo "Generating private key for rid=$RID"
-            celotooljs.sh generate bip32 --mnemonic "$MNEMONIC" --accountType validator --index $RID > /root/.celo/pkey
+            echo -n $RID > /root/.celo/replica_id
             echo 'Generating address'
-            celotooljs.sh generate account-address --private-key `cat /root/.celo/pkey` > /root/.celo/address
+            celotooljs.sh generate account-address --mnemonic "$MNEMONIC" --accountType validator --index $RID > /root/.celo/address
```
tkporter: Does this command work? I think `account-address` doesn't accept the `--accountType` and `--index` flags, maybe we should just save the generated bip32 provate key as an env variable and still use this command the same way

packages/terraform-modules/testnet/modules/validator/startup.sh
```diff
@@ -83,14 +83,13 @@ docker run -v $DATA_DIR:$DATA_DIR --name geth --net=host --entrypoint /bin/sh -d
...
     echo -n '${ip_address}' > $DATA_DIR/ipAddress && \
-    echo -n '$PRIVATE_KEY' > $DATA_DIR/pkey && \
     echo -n '$ACCOUNT_ADDRESS' > $DATA_DIR/address && \
     echo -n '$BOOTNODE_ENODE_ADDRESS' > $DATA_DIR/bootnodeEnodeAddress && \
     echo -n '$BOOTNODE_ENODE' > $DATA_DIR/bootnodeEnode && \
     echo -n '$GETH_ACCOUNT_SECRET' > $DATA_DIR/account/accountSecret && \
     geth init /var/geth/genesis.json
   ) && ( \
-    geth account import --password $DATA_DIR/account/accountSecret $DATA_DIR/pkey ; \
+    geth account import --password $DATA_DIR/account/accountSecret <(echo -n $PRIVATE_KEY) ; \
```
tkporter: `<` -> `$`

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -438,6 +436,22 @@ export async function initAndStartGeth(gethBinaryPath: string, instance: GethIns
...
+  const admins = gethConfig.instances
+    .filter(({ wsport, rpcport }) => wsport || rpcport)
+    .map(
+      ({ wsport, rpcport }) =>
+        new Admin(`${wsport ? 'ws' : 'http'}://localhost:${wsport || rpcport}`)
+    )
+  const enode = (await admins[0].getNodeInfo()).enode
+  await Promise.all(
+    admins.slice(1).map((admin) => {
+      admin.addPeer(enode)
```
tkporter: I think you might need to return here or not wrap this function call with `{}` in order for Promise.all to work as intended

packages/terraform-modules/testnet/modules/validator/startup.sh
nategraf: Actually this was intentional. It's a super cool feature called process substitution http://tldp.org/LDP/abs/html/process-sub.html
Unfortunately, `sh` doesn't support it. :(

tkporter: ah that's cool, didn't know that!

packages/celotool/src/e2e-tests/utils.ts
nategraf: Thanks!

packages/helm-charts/testnet/templates/validators.statefulset.yaml
nategraf: Thanks, I mistook that for the public key function

packages/helm-charts/testnet/templates/_helpers.tpl
nategraf: I've reinstated this file minus setting the nodekey for the miner.

packages/celotool/src/e2e-tests/utils.ts
nategraf: Done

packages/helm-charts/testnet/templates/validators.statefulset.yaml
nategraf: So I couldn't find a good way of never writing it to disk. But I made sure it gets deleted.

nategraf:
Now I've removed `nodekey` references from validator configs in both `helm` and `terraform` setups and tested the `helm` setup with a GKE deploy. I didn't know how I might test the terraform code.

packages/terraform-modules/testnet/modules/validator/startup.sh
```diff
@@ -83,14 +83,16 @@ docker run -v $DATA_DIR:$DATA_DIR --name geth --net=host --entrypoint /bin/sh -d
...
     echo -n '${ip_address}' > $DATA_DIR/ipAddress && \
-    echo -n '$PRIVATE_KEY' > $DATA_DIR/pkey && \
     echo -n '$ACCOUNT_ADDRESS' > $DATA_DIR/address && \
     echo -n '$BOOTNODE_ENODE_ADDRESS' > $DATA_DIR/bootnodeEnodeAddress && \
     echo -n '$BOOTNODE_ENODE' > $DATA_DIR/bootnodeEnode && \
     echo -n '$GETH_ACCOUNT_SECRET' > $DATA_DIR/account/accountSecret && \
     geth init /var/geth/genesis.json
   ) && ( \
-    geth account import --password $DATA_DIR/account/accountSecret $DATA_DIR/pkey ; \
+    TMP_PRIVATE_KEY_FILE=$(mktemp) \
```
tkporter: Could you put a `;` or `&&` after this? otherwise I think the following line might cause problems

tkporter (approved):

packages/terraform-modules/testnet/modules/validator/startup.sh
nategraf: Done.

PR #1925, opened 2019-11-26
cli: Fix #1897

### LINKED ISSUE #1897: Users SBAT read balance units when "celocli account:balance" is run
### Expected Behavior

Add units when displaying balances.

### Current Behavior

The `celocli account:balance` command should print out the units used when displaying the balances.

lucasege:
### Description

Added some logs to denote the current units used internally (everything is expressed in units of 10^-18).
Didn't add any name for this unit (i.e. 'wei') since it sounds like we are not decided on what we want to call these. 

### Tested

Ran `yarn test` to verify balance logging doesn't change test execution

### Related issues

- Fixes #1897


packages/cli/src/commands/account/balance.ts
```diff
@@ -22,6 +22,7 @@ export default class Balance extends BaseCommand {
       goldBalance: await goldToken.balanceOf(args.address),
       dollarBalance: await stableToken.balanceOf(args.address),
     }
+    console.log('All balances expressed in units of 10^-18.')
```
mcortesi, line 25: Not sure if this is "human readable" but it's better to what we have before.

Maybe check with @cla-bel

mcortesi (approved):
Would like to see what @cla-bel thinks.

Maybe we should divide by 10^18 and show in cGold and cDollar units

lucasege:
I agree that I'd like to get Claire's opinion, so I'll wait on merging until then.
I wanted to divide by 10^-18 but I'm worried about creating inconsistencies if there are other places these values are exposed and those are not reduced.

mcortesi:
i don't there are other places.. but yeah, not sure

packages/cli/src/commands/account/balance.ts
clrblmt: Works for now.

PR #1933, opened 2019-11-27
Removing unused parts of LockedGold

shellygr:
### Description
Removing dead code from LockedGold


packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -41,19 +30,9 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
-    bool exists;
-    // Each account may authorize additional keys to use for voting or valdiating.
-    // These keys may not be keys of other accounts, and may not be authorized by any other
-    // account for any purpose.
-    Authorizations authorizations;
-    Balances balances;
-  }
-
-  mapping(address => Account) private accounts;
+  mapping(address => Balances) private accounts;
```
nambrot: This is probably better named as `balances` to avoid confusion with `Accounts.sol`?

nambrot (commented):
A minor naming suggestion, unless @asaj believes otherwise

packages/protocol/contracts/governance/LockedGold.sol
shellygr: Updated!

asaj (approved):

PR #1951, opened 2019-11-28
Minor fixes to proxy docs

kevjue:
### Description

Minor fixes to the proxy docs

### Tested

N/A


asaj:
I think some of this may be out of date or addressed at this point

PR #1952, opened 2019-11-28
Micro-stake-off feedback

asaj:
### Description

This PR updates the validator getting started docs and celocli validator:register command and adds the celocli account:show command, as a result of a run-through of the current guide.

In particular, it:
- Updates the directory structure used in the guide
- Explicitly exports environment variables where needed
- Does not run the node in the background, so that logs can be seen
- Fixes the validator node to run with the signer address, rather than the account address.
- Removes duplicate instructions for running the accounts node
- Unlocks the accounts before running celocli
- Uses the validator signer ECDSA public key for validator registration, rather than (under the hood) using the validator account ECDSA public key

### Tested

- Used the modified validator:register command to register on baklavastaging


packages/cli/src/commands/account/authorize.ts
```diff
@@ -17,7 +17,7 @@ export default class Authorize extends BaseCommand {
       description: 'Role to delegate',
       required: true,
     }),
-    pop: flags.string({
+    signature: flags.string({
       description: 'Proof-of-possession of the signer key',
```
nategraf:
```suggestion
      description: 'Signature (a.k.a proof-of-possession) of the signer key',
```

packages/cli/src/commands/validator/register.ts
```diff
@@ -10,12 +10,13 @@ export default class ValidatorRegister extends BaseCommand {
...
     from: Flags.address({ required: true, description: 'Address for the Validator' }),
+    ecdsaKey: Flags.ecdsaPublicKey({ required: true }),
     blsKey: Flags.blsPublicKey({ required: true }),
-    blsPop: Flags.blsProofOfPossession({ required: true }),
+    blsSignature: Flags.blsProofOfPossession({ required: true }),
   }
 
   static examples = [
-    'register --from 0x47e172F6CfB6c7D01C1574fa3E2Be7CC73269D95 --blsKey 0x4fa3f67fc913878b068d1fa1cdddc54913d3bf988dbe5a36a20fa888f20d4894c408a6773f3d7bde11154f2a3076b700d345a42fd25a0e5e83f4db5586ac7979ac2053cd95d8f2efd3e959571ceccaa743e02cf4be3f5d7aaddb0b06fc9aff00 --blsPop 0xcdb77255037eb68897cd487fdd85388cbda448f617f874449d4b11588b0b7ad8ddc20d9bb450b513bb35664ea3923900',
+    'register --from 0x47e172F6CfB6c7D01C1574fa3E2Be7CC73269D95 --blsKey 0x4fa3f67fc913878b068d1fa1cdddc54913d3bf988dbe5a36a20fa888f20d4894c408a6773f3d7bde11154f2a3076b700d345a42fd25a0e5e83f4db5586ac7979ac2053cd95d8f2efd3e959571ceccaa743e02cf4be3f5d7aaddb0b06fc9aff00 --blsSignature 0xcdb77255037eb68897cd487fdd85388cbda448f617f874449d4b11588b0b7ad8ddc20d9bb450b513bb35664ea3923900',
```
nategraf: I don't see the new required `ecdsaKey` arg here

asaj:
Note that there is an accompanying celo-blockchain PR that I'll be able to push in ~30 minutes when I land

nategraf (approved):
👍

asaj:
Note that for expediency, the `ecdsaKey` in the example CLI command does not actually match the validator address

PR #1953, opened 2019-11-28
Small pre-stake-off CLI cleanup

asaj:
- Add error message to election:show
- Fix bug in Election wrapper that caused election:activate to not run
- Update migrations config reward values to be per-epoch, not annual, as that's how they're interpreted in the contracts. Corresponding updates to the governance e2e tests now that we hit the max target yield.
- Small display bug in account:show
- Update docs to activate votes at the end


timmoreton (approved):
LGTM!

PR #1960, opened 2019-11-28
Migrations: different CLabs groups gets different votes

timmoreton:
### Description

Implements a sliding scale of voting LockedGold for CLabs groups in migrations.

A new parameter, `votesRatioOfLastVsFirstGroup`, `>=1` determines the ratio of the votes in the last group vs the first. The first will be set as the minimum LockedGold required to stake (plus a small margin), and then each group will receive successively more from there, pro rata based on number of validators.

### Tested

Observed correct behavior in e2e tests.

## Backwards compatibility

Yes


mcortesi (approved):
lgtm

packages/protocol/migrations/20_elect_validators.ts
```diff
@@ -242,15 +257,23 @@ module.exports = async (_deployer: any, networkName: string) => {
       groupName += ` (${idx + 1})`
     }
 
-    console.info(`  Registering validator group: ${groupName} ...`)
-    const [groupAccount, groupLockedGoldValue] = await registerValidatorGroup(
+    // Calculate the LockedGold at this group
+    const groupLockedGoldValue = new BigNumber(lockedGoldPerValAtFirstGroup)
+      .plus(lockedGoldPerValEachGroup.times(idx))
+      .times(groupKeys.length)
```
mcortesi, line 264: I guess you would like to round this number...

PR #1966, opened 2019-11-29
[WIP] Set initial oracle value in migrations with oracle account

yerdua:
### Description

This PR is an attempt to fix a janky thing, where the initial c$/cGLD exchange rate is set by the deployer (`validator-0`). Ideally, only those accounts chosen to be oracles should be doing the reporting. 

This one-time report was fine in a time without oracles reporting on an ongoing basis. With the ongoing reporting, and no automatic way of removing old reports, that initial value hangs around forever. This may have unintended effects on the exchange rate.

However, this code doesn't work, and I haven't been able to figure out why. The problem is noted inline with the changes.

The feedback I'd like on this WIP PR is:
- What's wrong technically? Is there some subtle difference I'm missing between my code and how the same thing's being done in `20_elect_validators`?
- Is this problem worth solving?
- Assuming that this code worked, is this a reasonable way of solving this problem? For example, a possible alternative that could be useful in mainnet is a process that periodically removes expired reports.


packages/protocol/migrations/08_stabletoken.ts
```diff
@@ -37,31 +39,56 @@ module.exports = deploymentForCoreContract<StableTokenInstance>(
...
+      const reportTx = sortedOracles.contract.methods.report(
+        stableToken.address,
+        config.stableToken.goldPrice,
+        1,
+        NULL_ADDRESS,
+        NULL_ADDRESS
+      )
+      await sendTransactionWithPrivateKey(web3, reportTx, config.stableToken.oracleKeys[0], {
+        to: sortedOracles.address,
+      })
```
yerdua, line 90: This is the part that fails. This is the error, thrown inside of `sendTransactionWithPrivateKey`:
```
eth_sendRawTransaction
Error: Error: Error: Returned error: wrong number of fields in data
    at Object.run (/Users/audreypenven/celo/celo-monorepo/node_modules/truffle/build/webpack:/packages/truffle-migrate/index.js:92:1)
```

The same exact transaction sent differently above, from the default `from` on lines 66-72 above works without error.

Additionally, this code here looks identical in format to what's happening many times the later validator election migration. For example: https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/migrations/20_elect_validators.ts#L133-L137

So, I'm baffled as to why this is failing.

nambrot: Weird, looks like it's defined at https://github.com/ethereumjs/ethereumjs-util/blob/master/src/object.ts#L100

Could you maybe print out what `data` and `self._fields` is for you? Looks like some error in serialization maybe?

asaj:
Is this still a WIP?

asaj (commented):
This approach looks reasonable to me, suggest you follow @nambrot 's suggestion and start logging things too see what's wrong.

asaj:
> Is this problem worth solving?

Good question. For now I wouldn't spend too much time on this unless it is likely to be useful for mainnet, during which we'll probably just want to pass a list of addresses and let the "clients" make the actual reports

asaj:
> Assuming that this code worked, is this a reasonable way of solving this problem? For example, a possible alternative that could be useful in mainnet is a process that periodically removes expired reports.

Not exactly sure what you mean here, why is that the same problem? +1 to having such a process though

yerdua:
> > Assuming that this code worked, is this a reasonable way of solving this problem? For example, a possible alternative that could be useful in mainnet is a process that periodically removes expired reports.
> 
> Not exactly sure what you mean here, why is that the same problem? +1 to having such a process though

The main problem that's created by `validator-0` reporting the first value and the oracle account reporting all future ones is that `validator-0`'s initial report hangs out forever and messes up the median. If there's an even number of reports, the median that's chosen is the higher of the two in the middle of the list.

This problem would be solved by:
1. not involving `validator-0` in the first place
2. removing expired reports, getting rid of `validator-0`'s stale initial report

1 is cleaner and seems like an ideal path for this, assuming it can work.
2 is useful more broadly, and indirectly solves the problem. This would also deal with issues like, "`oracle-n` stopped reporting a while ago. Its last report is no longer current, but is the median of the list and defines the exchange rate. `oracle-x` is currently reporting, with up-to-date info, but its reports have not displaced `oracle-n`'s report that is the median.

PR #1967, opened 2019-11-29
[Wallet] Redesigning notification lists

### LINKED ISSUE #1808: Implement latest design for in-app notifications
### Expected Behavior

Notifications in the app should look like here: https://www.figma.com/file/mdNgkyVi2impjf5HRBLq3H/Notifications?node-id=0%3A1

### Current Behavior

They look differently

i1skn:
### Description
- [x] Redesign notification boxes on main mage
- [x] Redesign Incoming transaction list
- [x] ~Create a separate task to change currency in notifications to local~ - https://github.com/celo-org/celo-monorepo/issues/2153
- [x] Unify implementation for Incoming, Outgoing and Escrow notifications.
- [x] Add Screenshots
- [ ] Add Spanish translation

### Tested

iOS & Android

### Other changes

Some refactoring and little design tweaks

### Related issues

- Fixes #1808

### Backwards compatibility

Yes


packages/mobile/src/escrow/EscrowedPaymentLineItem.tsx
```diff
@@ -2,18 +2,19 @@ import fontStyles from '@celo/react-components/styles/fonts'
...
 }
 
 export default function EscrowedPaymentLineItem(props: Props) {
-  const { amount, message, recipientPhone } = props.payment
+  const { amount, recipientPhone } = props.payment
   return (
     <Text numberOfLines={1} ellipsizeMode="middle" style={styles.oneLine}>
       <Text style={[fontStyles.subSmall]}>
-        {recipientPhone} - {message}
+        {recipientPhone ? recipientPhone : i18n.t('global:unknown')} {i18n.t('global:for')}
```
jmrossy: To use the translation function in a component, we prefer wrapping the component in the `withNamespaces` HOC and then accessing t via the props instead of accessing the global i18n object.

packages/mobile/src/home/NotificationBox.test.tsx
```diff
@@ -96,7 +112,39 @@ describe('NotificationBox', () => {
...
+    })
+    const { getByText } = render(
+      <Provider store={store}>
+        <NotificationBox />
+      </Provider>
+    )
+    expect(getByText('outgoingPaymentRequests')).toBeTruthy()
+  })
+
+  it('renders outgoing payment request when they exist', () => {
```
jmrossy, line 134: How does this differ from the test above?

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -278,6 +278,10 @@ const AppStack = createStackNavigator(
     // Note, WalletHome isn't in this stack because it's part of the tab navigator
     [Screens.TabNavigator]: { screen: TabNavigator },
     [Stacks.SendStack]: { screen: SendStack },
+    [Screens.SendConfirmation]: { screen: SendConfirmation },
```
jmrossy, line 282: I assume you've added this here so that clicking back from payment request goes back to home. If so, can you add a comment mentioning that?

packages/mobile/src/notifications/NotificationList.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <View style={[styles.scrollArea]}>{props.items.map(props.listItemRenderer)}</View>
+        </ScrollView>
+      ) : (
+        <Text style={[fontStyles.bodySecondary, styles.empty]}>{i18n.t('global:emptyList')}</Text>
+      )}
+    </SafeAreaView>
+  )
+}
+
+export function titleWithBalanceNavigationOptions(title: string) {
```
jmrossy, line 44: this seems like an HOC / wrapper. Consider renaming to `withTitleBalanceNavigationOptions`

packages/mobile/src/notifications/NotificationList.tsx
```diff
@@ -0,0 +1,95 @@
...
+export function NotificationList<T>(props: Props<T>) {
+  return (
+    <SafeAreaView style={styles.container}>
+      <DisconnectBanner />
+      {props.items.length > 0 ? (
+        <ScrollView>
+          <View style={[styles.scrollArea]}>{props.items.map(props.listItemRenderer)}</View>
+        </ScrollView>
+      ) : (
+        <Text style={[fontStyles.bodySecondary, styles.empty]}>{i18n.t('global:emptyList')}</Text>
```
jmrossy, line 38: Ditto feedback as above for use of i18n here

jmrossy: Also, is 'Empty List' all the design calls for here?

packages/mobile/src/notifications/NotificationList.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+type Props<T> = OwnProps<T>
+
+export function NotificationList<T>(props: Props<T>) {
+  return (
+    <SafeAreaView style={styles.container}>
+      <DisconnectBanner />
+      {props.items.length > 0 ? (
+        <ScrollView>
+          <View style={[styles.scrollArea]}>{props.items.map(props.listItemRenderer)}</View>
```
jmrossy: No need for a style array here, there's just one style.
In general, it's not a big deal but I try to avoid style arrays because they cause a component to always be re-rendered when the parent is.

packages/mobile/src/notifications/NotificationList.tsx
```diff
@@ -0,0 +1,95 @@
...
+      )}
+    </SafeAreaView>
+  )
+}
+
+export function titleWithBalanceNavigationOptions(title: string) {
+  return ({ navigation }: { navigation: NavigationProp<NavigationState> }) => ({
+    ...headerWithBackButton,
+    headerTitle: (
+      <View style={{ alignItems: 'center', flex: 1 }}>
```
jmrossy: Please don't use inline styles. Put all styles in the `styles` object below. Similar reasons to avoiding style arrays, they force a re-render and also make the code a bit harder to read.

packages/mobile/src/notifications/NotificationList.tsx
```diff
@@ -0,0 +1,95 @@
...
+        {title && <Text style={{ ...fontStyles.bodyBold }}>{title}</Text>}
```
jmrossy: `style=fontStyles.bodyBold`

packages/mobile/src/notifications/SummaryNotification.test.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react'
+import { Text, View } from 'react-native'
+import * as renderer from 'react-test-renderer'
+import SummaryNotification from 'src/notifications/SummaryNotification'
+
+const props = () => ({
+  image: require('src/images/placeholder.png'),
```
jmrossy: Define the image path in /src/images and then just import the variable for use elsewhere

packages/mobile/src/notifications/SummaryNotification.tsx
```diff
@@ -0,0 +1,71 @@
...
+const styles = StyleSheet.create({
+  body: {
+    marginTop: 5,
+    flexDirection: 'row',
+  },
+  items: {
+    flex: 1,
+  },
+  moreWithCountText: {
+    ...fontStyles.subSmall,
```
jmrossy: Nit: may as well just use fontStyles.subSmall above

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -51,77 +52,68 @@ export class IncomingPaymentRequestListItem extends React.Component<Props> {
...
         >
-          <View>
-            {this.isDisplayingNumber() && (
-              <Text style={[fontStyles.subSmall, styles.phoneNumber]}>
-                {this.props.requester.displayId}
-              </Text>
-            )}
-            <Text style={[fontStyles.subSmall, styles.comment]}>{this.props.comment}</Text>
-          </View>
+          <Text style={[fontStyles.bodySmall]}>{this.props.comment || t('defaultComment')}</Text>
```
jmrossy: array

packages/mobile/src/paymentRequest/OutgoingPaymentRequestListScreen.tsx
```diff
@@ -44,66 +40,43 @@ const mapStateToProps = (state: RootState): StateProps => ({
...
-          updatePaymentRequestStatus={this.props.updatePaymentRequestStatus}
-          updatePaymentRequestNotified={this.props.updatePaymentRequestNotified}
-          requester={requester}
-          comment={request.comment}
-        />
-        {key < allRequests.length - 1 && <View style={styles.separator} />}
-      </View>
-    )
-  }
+export const listItemRenderer = (params: {
```
jmrossy, line 45: I this `props` is a more conventional name for this var. Or you can use desctructuring and avoid the name entirely

packages/mobile/src/set-clock/SetClock.tsx
```diff
@@ -5,18 +5,24 @@ import colors from '@celo/react-components/styles/colors'
...
 
 export class SetClock extends React.Component<WithNamespaces> {
   static navigationOptions = { header: null }
 
   goToSettings = () => {
-    return AndroidOpenSettings.dateSettings()
+    if (Platform.OS === 'android') {
+      return AndroidOpenSettings.dateSettings()
+    } else {
+      navigate(Screens.WalletHome)
```
jmrossy, line 25: Can you leave a TODO here.
@jeanregisser do we already have a task tracking implementation for this on iOS?

packages/mobile/src/web3/gas.ts
```diff
@@ -11,7 +11,7 @@ let gasPrice: BigNumber | null = null
 let gasPriceLastUpdated: number | null = null
 
 export async function getGasPrice(currency: CURRENCY_ENUM = CURRENCY_ENUM.DOLLAR) {
-  Logger.debug(`${TAG}}/getGasPrice`, 'Getting gas price')
+  Logger.debug(`${TAG}/getGasPrice`, 'Getting gas price')
```
jmrossy, line 14: 👍

packages/mobile/test/values.ts
```diff
@@ -160,4 +160,17 @@ export const mockPaymentRequests: PaymentRequest[] = [
     notified: true,
     type: NotificationTypes.PAYMENT_REQUESTED,
   },
+  {
```
jmrossy, line 163: Should this second be diff from the one above?

packages/react-components/components/BackButton.tsx
```diff
@@ -27,7 +27,7 @@ const SIZE = 16
 
 const style = StyleSheet.create({
   container: {
-    padding: 5,
+    marginLeft: -2,
```
jmrossy: I'm weary of changing this. Have you confirmed it looks right in all the places it's used? On iOS and android?

packages/react-components/components/BaseNotification.tsx
```diff
@@ -1,52 +1,95 @@
...
   onPress: () => unknown
   text: string
 }
 
-// For use in Notification Center and Payment Request Screen
-export default function BaseNotification(props: Props) {
+function Wrapper({ onPress, children }: { children: React.ReactNode; onPress?: () => unknown }) {
+  return onPress ? (
+    Platform.OS === 'android' ? (
+      <TouchableNativeFeedback onPress={onPress}>{children}</TouchableNativeFeedback>
```
jmrossy: Can you just use our custom `Touchable` component and make that work well for both iOS and Android?
packages/react-components/components/Touchable.tsx

packages/react-components/components/BaseNotification.tsx
```diff
@@ -1,52 +1,95 @@
...
+  container: {
+    padding: contentPadding,
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    width: '100%',
+    backgroundColor: colors.background,
+  },
+  iconArea: {
+    paddingRight: contentPadding,
+    flexDirection: 'column',
```
jmrossy: This is the default direction

packages/react-components/styles/styles.ts
```diff
@@ -4,6 +4,16 @@ import { StyleSheet } from 'react-native'
 
 export const TOP_BAR_HEIGHT = 56
 
+export function elevationShadowStyle(elevation: number) {
```
jmrossy, line 7: 👍

yarn.lock
```diff
@@ -17779,9 +17779,9 @@ i18next-xhr-backend@^2.0.0:
   integrity sha512-CP0XPjJsTE4hY1rM1KXFYo63Ib61EBLEcTvMDyJwr0vs9p/UTuA3ENCmzSs9+ghZgWSjdOigc0oUERHaxctbsQ==
 
 i18next@^11.9.1:
-  version "11.9.1"
-  resolved "https://registry.yarnpkg.com/i18next/-/i18next-11.9.1.tgz#f2718b11da2460e29e2dbd3dd8dea05dfc88e2dc"
-  integrity sha512-sUAfCdKYi8EufO6B8aY5RF/hqweCpRKvbxUhlzlmlAr+LgJQ2kXe/5y02kaPu6ApFIiW5WOz8+3ehoy97LUtMw==
+  version "11.10.2"
```
jmrossy: Why has this changed? Consider reverting this file

jmrossy (commented):
Overall really nice work! Please see comments below

packages/mobile/src/home/NotificationBox.test.tsx
i1skn: Single payment request vs multiple

packages/mobile/src/navigator/Navigator.tsx
i1skn: Yes, you are right. Will do!

packages/mobile/src/notifications/NotificationList.tsx
i1skn: It's actually not, just a function. 
Usage: ```Screen.navigationOptions = titleWithBalanceNavigationOptions(title)```

packages/mobile/src/notifications/NotificationList.tsx
i1skn: There is only one call to `t`, is that advisable to still use `withNamespace`?
Can you rephrase "is 'Empty List' all the design calls for here" - I did not get that :)

packages/mobile/src/paymentRequest/OutgoingPaymentRequestListScreen.tsx
i1skn: I wanted to avoid confusion with props in React component and desctructuring would trigger `no-shadow` eslint rules. So, I've came with `params`.

packages/mobile/test/values.ts
i1skn: not really, I just needed two of them. ID is diff, though.

packages/react-components/components/BaseNotification.tsx
i1skn: Oh, I sure can. I thought to make it more authentic for a user, so I've used `TouchableNativeFeedback`

packages/mobile/src/escrow/EscrowedPaymentLineItem.tsx
```diff
@@ -2,18 +2,19 @@ import fontStyles from '@celo/react-components/styles/fonts'
...
       </Text>
       <Text style={[fontStyles.subSmall, fontStyles.semiBold]}>
         {' '}
```
jeanregisser: Concatenating strings like this makes it impossible to translate correctly. Some locale don't use the same ordering in words.
We could instead use the [`<Trans />`](https://react.i18next.com/latest/trans-component) component and [context](https://www.i18next.com/translation-function/context).

Something along those lines:
```ts
<Trans
  i18nKey="escrowPaymentLine"
  tOptions={{ context: !recipientPhone ? 'missingRecipientPhone' : null }}
>
  <Text numberOfLines={1} ellipsizeMode="middle" style={styles.oneLine}>
    <Text style={[fontStyles.subSmall]}>{{ recipientPhone }} for </Text>
    <Text style={[fontStyles.subSmall, fontStyles.semiBold]}>{{ moneyAmount }}</Text>
  </Text>
</Trans>
```

With the following translation strings (not sure about the correctness of the indices I used though):
```
"escrowPaymentLine": "<1><2>{{recipientPhone}} for</2><3>{{moneyAmount}}</3></1>",
"escrowPaymentLine_missingRecipientPhone": "<1><2>Unknown for</2><3>{{moneyAmount}}</3></1>",
```

packages/mobile/src/escrow/EscrowedPaymentListScreen.tsx
```diff
@@ -43,57 +38,31 @@ const mapStateToProps = (state: RootState): StateProps => ({
...
-          <PaymentRequestListEmpty />
-        )}
-      </SafeAreaView>
-    )
-  }
+const EscrowedPaymentListScreen = (props: Props) => {
+  const { dollarBalance, navigation } = props
+  useBalanceInNavigationParam(dollarBalance, navigation)
+  return (
+    <NotificationList<EscrowedPayment>
```
jeanregisser: Can't TypeScript infer that type that from items?

packages/mobile/src/escrow/EscrowedPaymentReminderSummaryNotification.tsx
```diff
@@ -1,64 +1,46 @@
...
-      {
-        text: this.props.t('review'),
-        onPress: () => {
-          CeloAnalytics.track(CustomEventNames.escrowed_payment_review)
-          navigate(Stacks.EscrowStack)
-        },
-      },
-    ]
+  itemRenderer = (item: EscrowedPayment, key: number) => {
+    return <EscrowedPaymentLineItem payment={item} key={key} />
```
jeanregisser: Wondering, do we have some kind of ID we could use instead of the key?

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -30,6 +29,8 @@ interface OwnProps {
 
 type Props = OwnProps & WithNamespaces
 
+const avatarSize = 40
```
jeanregisser: nit: We might be inconsistent on this throughout the codebase but I think it should be `AVATAR_SIZE`.

packages/mobile/src/paymentRequest/OutgoingPaymentRequestListItem.tsx
```diff
@@ -26,14 +25,16 @@ interface OwnProps {
   updatePaymentRequestNotified: typeof updatePaymentRequestNotified
 }
 
+const avatarSize = 40
```
jeanregisser: nit: should be `AVATAR_SIZE`

packages/mobile/src/paymentRequest/OutgoingPaymentRequestListItem.tsx
```diff
@@ -26,14 +25,16 @@ interface OwnProps {
...
 type Props = OwnProps & WithNamespaces
 
 export class OutgoingPaymentRequestListItem extends React.Component<Props> {
   onRemind = () => {
-    const { id } = this.props
+    const { id, t } = this.props
     this.props.updatePaymentRequestNotified(id.toString(), false)
     CeloAnalytics.track(CustomEventNames.outgoing_request_payment_remind)
-    this.onFinalized()
+    Logger.showMessage(t('sendFlow7:requestSent'))
```
jeanregisser, line 37: Note: just so you know, these are not displayed on iOS for now since there's not equivalent to `ToastAndroid`.

packages/mobile/src/paymentRequest/PaymentRequestNotificationInner.tsx
```diff
@@ -2,22 +2,23 @@ import fontStyles from '@celo/react-components/styles/fonts'
...
 
 export default function PaymentRequestNotificationInner(props: Props) {
-  const { requesterE164Number, comment: message, amount, requesterRecipient } = props
+  const { requesterE164Number, amount, requesterRecipient } = props
   return (
     <Text numberOfLines={1} ellipsizeMode="middle" style={styles.oneLine}>
       <Text style={[fontStyles.subSmall]}>
-        {(requesterRecipient && requesterRecipient.displayName) || requesterE164Number} - {message}
+        {(requesterRecipient && requesterRecipient.displayName) || requesterE164Number}{' '}
+        {i18n.t('global:for')}
```
jeanregisser: We should use the injected `t` props instead via `withNamespaces` to be consistent with the other components we have.

packages/mobile/src/send/saga.ts
```diff
@@ -42,7 +42,7 @@ export async function getSendTxGas(
   const tokenContract = await contractGetter(web3)
   const txParams = { from: account, feeCurrency: tokenContract._address }
   const gas = new BigNumber(await tx.estimateGas(txParams))
-  Logger.debug(`${TAG}/getSendTxGas`, `Estimated gas of ${gas.toString()}}`)
+  Logger.debug(`${TAG}/getSendTxGas`, `Estimated gas of ${gas.toString()}`)
```
jeanregisser, line 45: 👍

packages/mobile/src/web3/gas.ts
jeanregisser, line 14: 👍

packages/mobile/src/web3/gas.ts
```diff
@@ -24,7 +24,7 @@ export async function getGasPrice(currency: CURRENCY_ENUM = CURRENCY_ENUM.DOLLAR
     }
     return gasPrice
   } catch (error) {
-    Logger.error(`${TAG}}/getGasPrice`, 'Could not fetch and update gas price.', error)
+    Logger.error(`${TAG}/getGasPrice`, 'Could not fetch and update gas price.', error)
```
jeanregisser, line 27: 👍

packages/react-components/styles/styles.ts
jeanregisser, line 7: Nice addition 👍

jeanregisser (approved):
Awesome work on this 💯 🎉

packages/mobile/src/escrow/EscrowedPaymentReminderSummaryNotification.tsx
i1skn: What would be the benefit to use ID vs key?

packages/mobile/src/escrow/EscrowedPaymentListScreen.tsx
i1skn: It does indeed! I've underestimated TS :)

packages/mobile/src/escrow/EscrowedPaymentLineItem.tsx
i1skn: thanks a lot! It does make a lot of sense! I remember you've mention something like this on one of our meetings.

packages/mobile/src/set-clock/SetClock.tsx
jeanregisser: Good catch, I've just pushed the iOS implementation :+1:

packages/mobile/src/escrow/EscrowedPaymentReminderSummaryNotification.tsx
jeanregisser: I assumed the key here was an index in an array which should be avoided if possible:
https://reactjs.org/docs/lists-and-keys.html#keys

packages/mobile/src/home/NotificationBox.test.tsx
jmrossy: Can we clarify the test description? I think the only diff is one character ;)

packages/mobile/src/notifications/NotificationList.tsx
jmrossy: Ah then maybe a `get` prefix would be helpful here

packages/mobile/src/notifications/NotificationList.tsx
jmrossy: Yes please, still use withNamespace

Nevermind the Empty List bit for now. Was just a comment on the intended design. I think showing the words "Empty List" is suboptimal but fine for now

packages/mobile/src/paymentRequest/OutgoingPaymentRequestListScreen.tsx
jmrossy: Ok

packages/mobile/src/escrow/EscrowedPaymentLineItem.tsx
```diff
@@ -28,3 +37,5 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
   },
 })
+
+export default withNamespaces('inviteFlow11')(EscrowedPaymentLineItem)
```
jmrossy: Nit, use the typed namespace IDs: `Namespaces.inviteFlow11`

packages/mobile/src/escrow/EscrowedPaymentLineItem.tsx
```diff
@@ -1,24 +1,33 @@
...
+        tOptions={{ context: !recipientPhone ? 'missingRecipientPhone' : null }}
+        values={{
+          amount:
+            CURRENCIES[CURRENCY_ENUM.DOLLAR].symbol +
+            getCentAwareMoneyDisplay(divideByWei(amount.toString())),
+          recipientPhone,
+        }}
+      >
+        <Text style={fontStyles.subSmall}>{{ recipientPhone }} for </Text>
+        <Text style={[fontStyles.subSmall, fontStyles.semiBold]}>{{ amount }}</Text>
```
jmrossy: You don't have to fix this but just an FYI, we would generally prefer to have a style defined below which does something like this:
```
{
...fontStyles.subSmall,
...fontStyles.semiBold
}
```

But it's really nbd. Just an FYI on the convention we're _roughly_ trying to keep to elsewhere

jmrossy (approved):
A few more small things but overall LGTM 👍

PR #1974, opened 2019-11-30
README backers sponsors section

KoenBal:
### Description
Added a hidden section with widgets for OpenCollective backers and sponsors

Could an OpenCollective account be created?
Is this legally a bad idea? 
Should an ETH and BTC address be added? 
How should this be done?

more about sponsors on GitHub
https://help.github.com/en/github/building-a-strong-community/displaying-a-sponsor-button-in-your-repository

README educational content I used.
[A beginners guide to writing kickass readmes](https://medium.com/@meakaakka/a-beginners-guide-to-writing-a-kickass-readme-7ac01da88ab3)
[AWESOME READMEs](https://github.com/matiassingers/awesome-readme)

PR #1973, opened 2019-11-30
README project team/code owners

KoenBal:
### Description
Part of a larger README update

Added a section which could be a start of the list of code owners for certain parts of the monorepo.
For an outsider its not really clear who is the core team and who works on celo cli, Android, smart contracts, etc

What do you think?

Rationale: For some users the celo monorepo could be the first impression they have of celo. A 
A nice README is a good way to help people engage in the project as well. A project with nice README and screenshots will get the attention of users better since it’s a direct way to explain why this project matters, and why people should use and contribute to the project. Good README should also include enough details to help a new user get started, e.g. how to compile, how to install, and how to start integrating.

README educational content I used.
[A beginners guide to writing kickass readmes](https://medium.com/@meakaakka/a-beginners-guide-to-writing-a-kickass-readme-7ac01da88ab3)
[AWESOME READMEs](https://github.com/matiassingers/awesome-readme)

PR #1972, opened 2019-11-30
README added mission

KoenBal:
### Descriptions
Added the mission statement
Added an embed image + link to youtube video

Rationale: For some users the celo monorepo could be the first impression they have of celo. A 
A nice README is a good way to help people engage in the project as well. A project with nice README and screenshots will get the attention of users better since it’s a direct way to explain why this project matters, and why people should use and contribute to the project. Good README should also include enough details to help a new user get started, e.g. how to compile, how to install, and how to start integrating.

README educational content I used.
[A beginners guide to writing kickass readmes](https://medium.com/@meakaakka/a-beginners-guide-to-writing-a-kickass-readme-7ac01da88ab3)
[AWESOME READMEs](https://github.com/matiassingers/awesome-readme)

PR #1971, opened 2019-11-30
README contributors

KoenBal:
### Descriptions

Added a Hall of Fame widget
Added a hidden section for opencollective contributors widget

Rationale: For some users the celo monorepo could be the first impression they have of celo. A 
A nice README is a good way to help people engage in the project as well. A project with nice README and screenshots will get the attention of users better since it’s a direct way to explain why this project matters, and why people should use and contribute to the project. Good README should also include enough details to help a new user get started, e.g. how to compile, how to install, and how to start integrating.

README educational content I used.
[A beginners guide to writing kickass readmes](https://medium.com/@meakaakka/a-beginners-guide-to-writing-a-kickass-readme-7ac01da88ab3)
[AWESOME READMEs](https://github.com/matiassingers/awesome-readme)

PR #1975, opened 2019-11-30
README toc

KoenBal:
### Description
Added a Table of Content as part of multiple README updates

As part of a larger README update
#1974
#1973 
#1972 
#1971 
#1970 
#1969 

[README GrowthHacking](https://www.freecodecamp.org/news/what-i-learned-from-an-old-github-project-that-won-3-000-stars-in-a-week-628349a5ee14/)


KoenBal:
@timmoreton Should I do something here? Was it correct to split up each part of the file in a new branch? Or am I taking a bad approach here?

PR #1983, opened 2019-12-02
[wallet] Invite links redirect to Apple Store on iOS

### LINKED ISSUE #1512: Wallet users SBAT invite another user with a link that works on both iOS and android
### Expected Behavior

when I invite a user, the invite link should work regardless of whether they are on an iOS or android device

### Current Behavior

assumes android

martinvol:
### Description

Firebase Dynamic link resolves to the right url depending on platform

### Tested

- Addes tests (more to come if it passes review)

### Other changes

- 

### Related issues

- Fixes #1512 

### Backwards compatibility


packages/mobile/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "private": true,
   "scripts": {
-    "start": "react-native start",
+    "start": "adb reverse tcp:8081 tcp:8081 && react-native start",
```
martinvol: wont merge this

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -135,7 +136,8 @@ export class WalletHome extends React.Component<Props> {
     this.props.refreshAllBalances()
   }
 
-  componentDidMount() {
+  async componentDidMount() {
+    await generateLink('asdf', 'asdf')
```
martinvol: just to make testing easier, wont merge

packages/mobile/src/utils/dynamicLink.ts
```diff
@@ -1,38 +1,32 @@
 import firebase from 'react-native-firebase'
-import { FIREBASE_WEB_KEY } from 'src/config'
 import Logger from 'src/utils/Logger'
 
 const TAG = 'utils/dynamicLink'
 
-export async function dynamicLink(url: string): Promise<string> {
+export async function generateDynamicShortLink(
```
martinvol: if it passes review I'll add tests for this function alone.

packages/mobile/src/invite/saga.ts
```diff
@@ -82,18 +83,36 @@ export function getInvitationVerificationFeeInWei() {
 }
 
 export async function generateLink(inviteCode: string, recipientName: string) {
-  const packageName = VersionCheck.getPackageName().replace(/\.debug$/g, '.integration')
-  const playStoreLink = await VersionCheck.getPlayStoreUrl({ packageName })
+  const AndroidPackageName = VersionCheck.getPackageName().replace(/\.debug$/g, '.integration')
+  const playStoreLink = await VersionCheck.getPlayStoreUrl({ AndroidPackageName })
   const referrerData = encodeURIComponent(`invite-code=${inviteCode}`)
   const referrerLink = `${playStoreLink}&referrer=${referrerData}`
```
annakaz: Rename to `playStoreUrl` to be consistent with `appStoreUrl` below?

packages/mobile/src/invite/saga.ts
```diff
@@ -82,18 +83,36 @@ export function getInvitationVerificationFeeInWei() {
...
-  const shortUrl = await dynamicLink(referrerLink)
-  const msg = i18n.t('sendFlow7:inviteSMS', {
+
+  let appStoreUrl
+  try {
+    appStoreUrl = await VersionCheck.getAppStoreUrl()
+  } catch (error) {
+    // can't get App Store url on Android
+    appStoreUrl = 'http://play.google.com/store/apps/details?id=org.celo.mobile.debug'.replace(
```
annakaz: Could be based on `AndroidPackageName` instead of hardcoding to integration?

packages/mobile/src/invite/saga.ts
```diff
@@ -82,18 +83,36 @@ export function getInvitationVerificationFeeInWei() {
 }
 
 export async function generateLink(inviteCode: string, recipientName: string) {
-  const packageName = VersionCheck.getPackageName().replace(/\.debug$/g, '.integration')
-  const playStoreLink = await VersionCheck.getPlayStoreUrl({ packageName })
+  const AndroidPackageName = VersionCheck.getPackageName().replace(/\.debug$/g, '.integration')
```
annakaz: nit: androidPackageName to follow capitalization convention

packages/mobile/src/invite/saga.ts
```diff
@@ -82,18 +83,36 @@ export function getInvitationVerificationFeeInWei() {
...
+  const playStoreLink = await VersionCheck.getPlayStoreUrl({ AndroidPackageName })
```
annakaz: Does this work on iOS?

packages/mobile/src/utils/dynamicLink.ts
```diff
@@ -1,38 +1,32 @@
...
+  url: string,
```
annakaz: playStoreUrl?

packages/mobile/src/utils/dynamicLink.ts
```diff
@@ -1,38 +1,32 @@
...
-      if (response.status !== 200) {
-        return url
-      }
-      const json = await response.json()
-      return json.shortLink
-    }
+    link.ios.setFallbackUrl(appStoreUrl)
+    link.ios.setBundleId(iOSBundleId)
+
+    // Please not other parameter than UNGUESSABLE is unsafe
```
annakaz: Typo: note

annakaz: Also why are parameters other than UNGUESSABLE unsafe?

packages/mobile/src/invite/saga.ts
```diff
@@ -82,18 +83,36 @@ export function getInvitationVerificationFeeInWei() {
...
+    appStoreUrl = 'http://play.google.com/store/apps/details?id=org.celo.mobile.debug'.replace(
+      /\.debug$/g,
+      '.integration'
+    )
+  }
+
+  const iOSBundleId = await DeviceInfo.getBundleId()
+
+  Logger.info(TAG, `referrerLink before short ${referrerLink}`)
+  const shortUrl = await generateDynamicShortLink(
```
annakaz: I had a bit of trouble following what was for android vs iOS, and that there was a link and id provided for both. Would it be possible to use `playStoreUrl`, `appStoreUrl`, `playStoreId`, and `appStoreId` instead? Or other terms that are the same for both platforms (and stick to playStore/appstore or android/ios for both terms, rather than playStore/android and appstore/ios)

packages/mobile/src/utils/dynamicLink.ts
```diff
@@ -1,38 +1,32 @@
...
+    const shortUrl = await firebase.links().createShortDynamicLink(link, 'UNGUESSABLE')
+    Logger.error(TAG, `dynamic link ${shortUrl}`)
+
+    // It is NOT recommended to shorten this link because it
```
annakaz: How long is the link? Can it still fit in a text

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -135,7 +135,7 @@ export class WalletHome extends React.Component<Props> {
     this.props.refreshAllBalances()
   }
 
-  componentDidMount() {
+  async componentDidMount() {
```
jmrossy: No need for this to be async, please restore

packages/mobile/src/invite/saga.test.ts
```diff
@@ -196,3 +199,25 @@ describe(watchRedeemInvite, () => {
       .run()
   })
 })
+
+describe(generateLink, () => {
+  beforeEach(() => {
```
jmrossy, line 219: Nit: why do you need this since you only have one test?

packages/mobile/src/invite/saga.ts
jmrossy: Can you help me understand why we'd hit this error?

packages/mobile/src/invite/saga.ts
jmrossy: why are we defaulting to `integration` here?

packages/mobile/src/invite/saga.ts
```diff
@@ -82,18 +83,36 @@ export function getInvitationVerificationFeeInWei() {
...
-  const msg = i18n.t('sendFlow7:inviteSMS', {
+
+  let appStoreUrl
+  try {
+    appStoreUrl = await VersionCheck.getAppStoreUrl()
+  } catch (error) {
+    // can't get App Store url on Android
+    appStoreUrl = 'http://play.google.com/store/apps/details?id=org.celo.mobile.debug'.replace(
+      /\.debug$/g,
+      '.integration'
```
jmrossy: ditto, why integration

packages/mobile/src/invite/saga.ts
```diff
@@ -82,18 +83,36 @@ export function getInvitationVerificationFeeInWei() {
...
+    )
+  }
+
+  const iOSBundleId = await DeviceInfo.getBundleId()
+
+  Logger.info(TAG, `referrerLink before short ${referrerLink}`)
```
jmrossy: The link contains an invite code right? We shouldn't log it.

packages/mobile/src/utils/dynamicLink.ts
```diff
@@ -1,38 +1,32 @@
...
+  appStoreUrl: string,
+  androidPackageName: string,
+  iOSBundleId: string
+): Promise<string> {
   try {
-    // TODO: Move 'celo.page.link' to celo-org-mobile firebase project and then flip this bool once
-    // that's done.
-    const useRNFirebase = false
+    // legacy users are using `celo.page.link`, but if we move that domain to the current firebase
+    // those links are gonna break, so better do before a network reset
```
jmrossy: We have a network reset coming up with the next app update so let's do it now

packages/mobile/src/utils/dynamicLink.ts
```diff
@@ -1,38 +1,32 @@
...
+    const link = new firebase.links.DynamicLink(appStoreUrl, 'https://celolink.page.link')
```
jmrossy: Can we use a nicer / shorter link domain for this?

packages/mobile/src/invite/saga.test.ts
jeanregisser: How about we enable clearing mocks between tests using the config? 
https://jestjs.io/docs/en/configuration#clearmocks-boolean

martinvol: It has to be there because otherwise the mock calls number is set on the other tests that run before.  I like the `beforeEach` because if you add more tests you don't have to copy it over.

packages/mobile/src/invite/saga.ts
martinvol: The functions throws returns an error Android.

packages/mobile/src/invite/saga.ts
martinvol: I guess the reason is that the build we use for internal testing is `integration`, but not sure.

packages/mobile/.env.test
```diff
@@ -1,5 +1,5 @@
 ENVIRONMENT=local
-DEFAULT_TESTNET=integration
+DEFAULT_TESTNET=pilot
```
jmrossy: why change this? Pilot is quite out of date

packages/mobile/android/app/src/main/java/org/celo/mobile/MainApplication.java
```diff
@@ -12,9 +12,15 @@
 import com.facebook.soloader.SoLoader;
 import com.reactnativegeth.RNGethPackage;
 import io.invertase.firebase.auth.RNFirebaseAuthPackage;
+import io.invertase.firebase.auth.RNFirebaseAuthPackage;
```
jmrossy: Looks like you have some duplicates in these imports

jmrossy (changes requested):

packages/mobile/src/utils/dynamicLink.ts
martinvol: I created an example link and got this:
`Hi martin! I would like to invite you to join the Celo payments network. Your invite code is: RDfQlvthTUOf1VtZIJawRBU+dWCKJyndP/M0emKwSjc= You can install the Celo app from this link: https://celolink.page.link/SZ1EeUaCHdrDsd9x9`

This link is 228 chars long (dependant on the name of the user) and an (old) SMS only supports 160. What do you think we should do @nityas?

packages/mobile/locales/en-US/sendFlow7.json
```diff
@@ -45,7 +45,7 @@
   "mobileNumber": "Mobile #",
   "walletAddress": "Wallet Address",
   "inviteSMS":
-    "Hi{{name}}! I would like to invite you to join the Celo payments network. Your invite code is: {{code}} You can install the Celo application from the following link: {{link}}",
+    "Hi{{name}}! I'd like to invite you the Celo payments network. Your invite code is: {{code}} You can install the Celo app from this link: {{link}}",
```
martinvol: Try to short the invite a bit more @nityas

packages/mobile/src/invite/saga.ts
```diff
@@ -81,18 +81,26 @@ export function getInvitationVerificationFeeInWei() {
...
-  const playStoreLink = await VersionCheck.getPlayStoreUrl({ packageName })
+  const androidPackageName = VersionCheck.getPackageName().replace(/\.debug$/g, '.integration')
+  const playStoreLink = await VersionCheck.getPlayStoreUrl({ androidPackageName })
   const referrerData = encodeURIComponent(`invite-code=${inviteCode}`)
-  const referrerLink = `${playStoreLink}&referrer=${referrerData}`
-  const shortUrl = await dynamicLink(referrerLink)
-  const msg = i18n.t('sendFlow7:inviteSMS', {
+  const playStoreLinkReferrerLink = `${playStoreLink}&referrer=${referrerData}`
+
+  const iOSBundleId = '1482389446' // This is the id in the Apple Store URL
```
martinvol: Couldn't figure out a way to get it programatically, so up to suggestions. Moving to an `.env` file or at least `conf.ts` feels a happier place if there's no function that yields this number.

PR #1987, opened 2019-12-02
[Wallet] Verification Off-Chain Reveal

### LINKED ISSUE #1040: Wallet Users SBAT to reveal to an attestation service
### Expected Behavior

Depending on https://github.com/celo-org/celo-monorepo/issues/1039 and https://github.com/celo-org/celo-monorepo/issues/1038, the wallet should reveal the phone number to the attestation service.

### Current Behavior

It reveals the phone number to the validator via a reveal transaction.

### LINKED ISSUE #1944: Make the app work with Baklava networks

jmrossy:
### Description

Convert app to use the new off-chain reveal + attestation service based verification.
Remove walletkit's attestation-related code and use contract kit instead.
Related code cleanup and reorganization.

### Tested

Partly tested on baklavastaging but currently blocked

### Related issues

- Fixes #1040 
- Fixes #1944

### Backwards compatibility

No, the new verification scheme is not backwards compatible


packages/mobile/package.json
```diff
@@ -43,7 +43,7 @@
     ]
   },
   "dependencies": {
-    "@celo/client": "9575a01",
+    "@celo/client": "0.0.211",
```
jmrossy: TODO: final version TBD

packages/utils/src/signatureUtils.ts
```diff
@@ -1,10 +1,8 @@
-import assert = require('assert')
```
jmrossy, line 1: This was causing import errors in some places and it's not really necessary here, cutting.

packages/mobile/src/account/Account.tsx
```diff
@@ -173,11 +173,11 @@ export class Account extends React.Component<Props, State> {
     } else {
       return (
         <View style={style.devSettings}>
-          <View style={style.devSettingsItem}>
+          {/* <View style={style.devSettingsItem}>
             <TouchableOpacity onPress={this.revokeNumberVerification}>
               <Text>Revoke Number Verification</Text>
             </TouchableOpacity>
-          </View>
+          </View> */}
```
jeanregisser, line 180: Will this be still possible in the future?

packages/mobile/src/networkInfo/saga.ts
```diff
@@ -15,8 +15,8 @@ export function* waitForRehydrate() {
 }
 
 export function* waitWeb3LastBlock() {
-  yield waitForGethConnectivity()
-  yield waitForWeb3Sync()
+  yield call(waitForGethConnectivity)
+  yield call(waitForWeb3Sync)
```
jeanregisser, line 19: 👍

packages/walletkit/src/contract-utils.ts
```diff
@@ -13,9 +13,9 @@ import { getGasPriceMinimumContract } from './contracts'
 import { getGoldTokenAddress } from './erc20-utils'
 import { Logger } from './logger'
 
-const gasInflateFactor = 1.3
+const gasInflateFactor = 1.5
```
jeanregisser, line 16: Why changing this?

jeanregisser (commented):
I'm not familiar with this part of the code so won't give approval, but what I saw looked good 👍

packages/walletkit/src/contract-utils.ts
jmrossy: because Baklava

packages/mobile/src/account/Account.tsx
jmrossy: Maybe. It's currently broken and we'll need to see how it can be done with the new Attestation contract which collects attestations instead of replacing them when re-verifying.

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -254,7 +239,7 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
             name: nameClaim ? nameClaim.name : undefined,
           }
         } catch (error) {
-          console.error(error)
+          console.info('Error getting attestation service URLs', error)
```
jeanregisser: Why not logging using the error level?

packages/mobile/.env
```diff
@@ -1,11 +1,11 @@
...
 # If ZERO_SYNC_ENABLED_INITIALLY, local geth will not run initially. 
 # If toggled on, it will use DEFAULT_SYNC_MODE. See src/geth/consts.ts for more info
-ZERO_SYNC_ENABLED_INITIALLY=true
+ZERO_SYNC_ENABLED_INITIALLY=false
 DEFAULT_SYNC_MODE=5
 DEV_SETTINGS_ACTIVE_INITIALLY=true
 FIREBASE_ENABLED=true
 SECRETS_KEY=debug
-SHOW_TESTNET_BANNER=true
+SHOW_TESTNET_BANNER=false
```
jeanregisser: Sure you want to merge these too?

packages/mobile/src/identity/verification.ts
```diff
@@ -492,82 +489,68 @@ function* waitForAttestationCode(issuer: string) {
...
+    // if (index < 0) {
+    //   Logger.warn(TAG + '@revokeVerification', 'Account not found')
+    //   return
+    // }
+
+    // const revokeTx = attestationsWrapper.methods.revoke(e164NumberHash, index)
+    // yield call(sendTransaction, revokeTx, account, TAG, 'Revoke attestation')
+
+    // Logger.debug(TAG + '@revokeVerification', 'Done revoking previous verification')
+    // Logger.showMessage('Done revoking phone verification')
```
jeanregisser, line 563: Should we delete?

packages/walletkit/src/contract-utils.ts
jeanregisser: What changed in Baklava?

jeanregisser (commented):
Left minor comments. Giving approval once you've confirmed the .env changes.

packages/mobile/src/identity/verification.ts
jmrossy: We can. i expect the fix code would look similar to this so kept it for a reference. What would you prefer?

packages/mobile/.env
jmrossy: The zero sync was intentional but the banner was not. will revert.

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: Because this error happens quite often and it causes red banners

packages/walletkit/src/contract-utils.ts
jmrossy: No idea but I was seeing failures due to underestimated gas amounts

packages/mobile/src/identity/verification.ts
jeanregisser: Ok, I'm fine leaving this.

packages/contractkit/src/wrappers/Attestations.ts
jeanregisser: ok but this is normal in this process, right?
Maybe just add a comment.

jmrossy: Will do!

jeanregisser (approved):
🎉

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -212,6 +196,7 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
    * @param attestationsRequested The number of attestations to request
    */
   async approveAttestationFee(attestationsRequested: number) {
+    // TODO should this use the registry like we do below in the request() function?
```
nambrot: This does under the hood use the registry, but in this case we actually also need the contract instance to approve

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -239,8 +224,9 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
           const metadata = await IdentityMetadataWrapper.fetchFromURL(metadataURL)
           const attestationServiceURLClaim = metadata.findClaim(ClaimTypes.ATTESTATION_SERVICE_URL)
 
-          if (attestationServiceURLClaim === undefined) {
-            throw new Error(`No attestation service URL registered for ${issuer}`)
+          if (!attestationServiceURLClaim) {
+            console.info(`No attestation service URL registered for ${issuer}`)
```
nambrot, line 227: FYI in https://github.com/celo-org/celo-monorepo/pull/2105/files I'm removing the logging of the errors to the console. IMO it should be the responsibility of the consumer of contractkit to determine appropriate logging when running the code

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -400,12 +378,12 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
    * Selects the issuers for previously requested attestations for a phone number
    * @param phoneNumber The phone number for which to request attestations for
    */
-  async selectIssuers(phoneNumber: string) {
+  selectIssuers(phoneNumber: string) {
```
nambrot, line 380: 🙏

packages/utils/src/attestations.test.ts
```diff
@@ -71,17 +63,9 @@ describe('Attestation Utils', () => {
...
-      expect(sanitizeBase64(NORMAL_CODE)).toBe(NORMAL_CODE)
-    })
-  })
-
-  describe('decodeAttestationCode', () => {
-    it('decodes the base64 message correctly', () => {
-      expect(decodeAttestationCode(MESSAGE_1_CODE)).toBe(MESSAGE_1_DECODED)
-      expect(decodeAttestationCode(MESSAGE_2_CODE)).toBe(MESSAGE_2_DECODED)
-      expect(decodeAttestationCode(MESSAGE_3_CODE)).toBe(MESSAGE_3_DECODED)
+      expect(sanitizeMessageBase64(CODE_1)).toBe(SANITIZED_CODE_1)
```
nambrot, line 67: It would be great to update these test cases with the deeplink prefixes

nambrot (approved):
🎉 🎉 🎉 🎉 🎉 🎉

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: So would this throw? just return null?

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: Will remove comment

packages/utils/src/attestations.test.ts
jmrossy: left a comment to this effect

packages/contractkit/src/wrappers/Attestations.ts
nambrot: We catch it in the surrounding try, and thus just return null. In effect, the method still returns the actionable attestations. If a consumer wanted to log that some of the attestations "should" be completable, but there was not attestation service registered (or any other issue), consumer should use the new method of `getNonCompliantIssuers`

PR #1991, opened 2019-12-02
Error on method to set fee currency

matiasbn:
Changed line 43 from 'setGasCurrency' to 'setFeeCurrency' because the first method doesn't exist

### Description

### Tested

### Other changes

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/packages/docs/developer-resources/contractkit/setup.md
+++ b/packages/docs/developer-resources/contractkit/setup.md
@@ -40,7 +40,7 @@ import { CeloContract } from '@celo/contractkit'
 // default from
 kit.defaultAccount = myAddress
 // paid gas in celo dollars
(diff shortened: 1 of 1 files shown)


nategraf (approved):
Nice! Thanks for submitting a PR for this!

nategraf (changes requested):
Oh wait, this is correct for the branch

nategraf:
So to clarify, the change of `gasCurrency` to `feeCurrency` is checked into `master` but not `alfajores`. If you are using the `alfajores` branch, then `kit.setGasCurrency` should be valid, unless I am mistaken.

nategraf:
Let me know if you want to reopen this

PR #1992, opened 2019-12-02
Don't build docs or web in CI

cmcewen:
### Description

We don't need to build docs or web in CI as a postinstall since those only need to be built for deployment


H34D (approved):
lgtm

PR #1997, opened 2019-12-02
expose signer through getValidator

diminator:
`getValidator(validator)` now also returns `signer:address`. required for ethstats valset management


packages/protocol/test/governance/validators.ts
```diff
@@ -1091,6 +1098,7 @@ contract('Validators', (accounts: string[]) => {
           })
 
           it('should set the validator ecdsa public key', async () => {
+            await registry.setAddressFor(CeloContractName.Accounts, accountsInstance.address)
```
timmoreton, line 1101: why do you need this?

timmoreton (approved):

timmoreton:
Have you tested CLI and other users of `getValidator` works with these changes ? should update `election/current.ts` etc.

packages/protocol/test/governance/validators.ts
diminator: `getValidator` calls `getAccount().getValidatorSigner` account.address needs to be contract, not wallet

diminator: https://github.com/celo-org/celo-monorepo/pull/1997/files/f308cff7a9aa1a716337f18747382358a7984b2f#diff-cee92562946fb3c6fc71ec4d162da770R1087

diminator:
> Have you tested CLI and other users of `getValidator` works with these changes ? should update `election/current.ts` etc.

haven't exposed this to the entire API. only for contract_comm due to time reasons

PR #1998, opened 2019-12-02
Fix validator election migration bug with sorted list insertion

nategraf:
### Description

Some of the validator transactions require correctly ordering the inserted group with existing groups. Recently the migration changed and introduced a bug where the order would be incorrect. This PR fixes that.

### Tested

Ran migrations on a local testnet of 20 validators

### Backwards compatibility

100%


timmoreton (approved):
nice cleanup!

PR #2005, opened 2019-12-03
[BlockchainApi] Add ability to get exchange rates from/to cGLD or cUSD

jeanregisser:
### Description

Add ability to get exchange rates from/to cGLD or cUSD.
- It's now possible to get rates from cGLD to any other currency (cUSD, USD, MXN, etc) and vice versa.
- Supports getting cGLD rate at a specific date/time (with a granularity of 30 mins as that's the current interval at which the rate is being stored)

For now the cUSD/USD and USD/cUSD rates are assumed to be 1.
Also going from cGLD to local currency (or vice versa) is currently assumed to be the same as cGLD -> cUSD -> USD -> local currency. 
And similar to cUSD to local currency, but with one less step.

### Tested

Added new tests.
Successfully deployed on integration: https://integration-dot-celo-testnet.appspot.com

### Other changes

- Use service account key for all envs. This is so `blockchain-api` in `celo-testnet` project can access the firebase db in `celo-org-mobile` project. This is temporary until we migrate `blockchain-api` to `celo-org-mobile` too.
- Speed up jest tests by mocking contractkit.

WITHOUT mocking contractkit:
- yarn test (clean cache): ~27secs
- yarn test (populated cache): 23 secs

WITH mocking contractkit:
- yarn test (clean cache): ~8secs
- yarn test (populated cache): ~4secs

### Related issues

- #834

### Backwards compatibility

Yes


packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private getConversionSteps(fromCode: string, toCode: string) {
+    if (fromCode === toCode) {
+      // Same code, nothing to do
+      return []
+    } else if (fromCode === CGLD || toCode === CGLD) {
+      // cGLD -> X (where X !== cUSD)
+      if (fromCode === CGLD && toCode !== CUSD) {
+        return [CGLD, CUSD, ...insertIf(toCode !== USD, USD), toCode]
```
annakaz, line 54: Do we assume that the cUSD -> local currency rate is the same as the cUSD -> USD -> local currency rate? This may not be the case if the price of cUSD differs across markets. For example, there used to be a consistent spread between BTC -> dollars and BTC -> Korean yen

One workaround, @nityas suggested in the cash in/out spec is to use cUSD -> local currency rate as the stored exchange rate. Not sure whether it's easier to address now or during some of the cash in/out work

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  private getSupportedExchangeRate(
+    fromCode: string,
+    toCode: string,
+    timestamp?: number
+  ): Promise<BigNumber> {
+    let pair = `${fromCode}/${toCode}`
+
+    if (pair === 'cUSD/USD' || pair === 'USD/cUSD') {
+      return Promise.resolve(new BigNumber(1))
```
annakaz, line 83: More of a product question, but we should figure out how we handle the case when cUSD -> USD rate is not 1 (as I think we expect it to float within 1% of 1). 

Depending on what we decide, we may want to store the cUSD/USD exchange rate in firebase (and just have it be 1 or some other simulated rate)

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
jeanregisser: Yes that's the assumption so far given the data we have.
I imagined this would have to change as soon as we have data with more direct markets.

This function is the generic way to get exchange rates but the idea of taking the rate computed from exchanges seems good to me and is something I thought about when designing the new blockchain-api schema changes. I had however only thought about this in the context of exchange transactions for a specific address.
From what you wrote, it looks like @nityas suggested we could store exchanges rates for all exchange transactions as they happen? this seems very appealing, would love to chat directly about this!

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
jeanregisser: Yes exactly!

packages/blockchain-api/src/config.ts
```diff
@@ -14,10 +14,33 @@ function getSecrets(deployEnv: string) {
   return envSecrets
 }
 
+export function getFirebaseAdminCreds(admin: any) {
+  // TODO: move project to celo-org-mobile
+  // until then, using serviceAccountKey for all envs
+  // tslint:disable-next-line: no-constant-condition
+  if (true /* DEPLOY_ENV === 'local' */) {
```
annakaz, line 21: Remove if (true) ? Or uncomment /* DEPLOY_ENV === 'local' */?

packages/blockchain-api/src/currencyConversion/GoldExchangeRateAPI.ts
```diff
@@ -0,0 +1,75 @@
...
+  const hiRate = rates[hi]
+  if (!loRate || !hiRate) {
+    return loRate || hiRate
+  }
+
+  return loRate.timestamp - timestamp < timestamp - hiRate.timestamp ? loRate : hiRate
+}
+
+export default class GoldExchangeRateAPI<TContext = any> extends DataSource {
+  // TODO(jeanregisser): memoize results
```
annakaz, line 43: TODOs here and below for a future PR? Or left in for now?

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    const result = await currencyConversionAPI.getExchangeRate({
+      sourceCurrencyCode: 'cGLD',
+      currencyCode: 'MXN',
+    })
+    expect(result).toEqual(new BigNumber(200))
+    expect(mockDefaultGetExchangeRate).toHaveBeenCalledTimes(1)
+    expect(mockGoldGetExchangeRate).toHaveBeenCalledTimes(1)
+  })
+
+  it('should retrieve rate for MXN/cGLD', async () => {
```
annakaz, line 75: Does this test cover any additional logic from the `MXN/cGLD` one given the mocks?

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    const result = await currencyConversionAPI.getExchangeRate({
+      sourceCurrencyCode: 'USD',
+      currencyCode: 'MXN',
+    })
+    expect(result).toEqual(new BigNumber(20))
+    expect(mockDefaultGetExchangeRate).toHaveBeenCalledTimes(1)
+    expect(mockGoldGetExchangeRate).toHaveBeenCalledTimes(0)
+  })
+
+  it('should retrieve rate for MXN/USD', async () => {
```
annakaz, line 95: Same with this one- is there anything this test would catch that wouldn't be covered by the `USD/MXN` one? 
Could also consider adding some tests for `getConversionSteps`

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  private getSupportedExchangeRate(
+    fromCode: string,
+    toCode: string,
+    timestamp?: number
+  ): Promise<BigNumber> {
+    const pair = `${fromCode}/${toCode}`
+
+    if (pair === 'cUSD/USD' || pair === 'USD/cUSD') {
+      // TODO: use real rates once we have the data
```
annakaz, line 82: 👍

annakaz (approved):
LGTM! Left some minor comments

packages/blockchain-api/src/config.ts
jeanregisser: I wanted to keep it this way so we can uncomment once we deploy it to `celo-org-mobile`.
Makes sense?

packages/blockchain-api/src/currencyConversion/GoldExchangeRateAPI.ts
jeanregisser: I left them for a future PR.

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.test.ts
jeanregisser: They look really similar indeed but the idea was to cover both local to cGLD and cGLD to local.

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.test.ts
jeanregisser: Yes similar to my comment above.
I decided not to test `getConversionSteps` directly to avoid testing the internals too much.

packages/blockchain-api/package.json
```diff
@@ -23,6 +23,7 @@
     "bignumber.js": "^7.2.0",
     "dotenv": "^6.1.0",
     "express": "^4.16.4",
+    "firebase-admin": "^8.8.0",
```
jmrossy: Please use the same version we're using in other monorepo packages (8.6.1)

packages/blockchain-api/src/config.ts
```diff
@@ -14,10 +14,33 @@ function getSecrets(deployEnv: string) {
...
+    try {
+      const serviceAccount = require('../serviceAccountKey.json')
+      return admin.credential.cert(serviceAccount)
+    } catch {
```
jmrossy: Should we catch and log the error object here?

packages/blockchain-api/src/currencyConversion/index.ts
```diff
@@ -0,0 +1 @@
+export { default as CurrencyConversionAPI } from './CurrencyConversionAPI'
```
jmrossy: Why create this index file instead of having other files import from CurrencyConversionAPI directly?

jmrossy (commented):
A few small comments but mostly looks good!

packages/blockchain-api/src/config.ts
jeanregisser: Yes probably, I blindly cut and pasted this from `notification-service`

packages/blockchain-api/package.json
jeanregisser: :+1:

packages/blockchain-api/src/currencyConversion/index.ts
jeanregisser: Mostly to hide the internal file organisation of currency conversion with its somewhat private internal modules (`ExchangeRateAPI` and `GoldExchangeRateAPI`).
But doesn't really matter.

jmrossy (approved):

PR #2008, opened 2019-12-03
[wip] Skip builds if only docs changed

timmoreton:
### Description

### Tested

### Other changes

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


evertonfraga:
You can append `[ci skip]` or `[skip ci]` to the commit message, so CircleCI skips building the whole thing ([reference](https://circleci.com/docs/2.0/skip-build/)). 

It is more efficient, as it does not even start the jobs, instead of handling it in every job. You could even have a `husky` hook that appends that for you, given the file patterns changed. CI config files tends to be bloated in monorepos, and it's good to keep them simple.

PR #2014, opened 2019-12-03
Baklava ethstats workaround

jcortejoso:
Temporal workaround to address the ethstats usage increase over time.
The server process inside the container will restart every 10 minutes.


packages/helm-charts/ethstats/templates/ethstats.deployment.yaml
```diff
@@ -52,14 +52,19 @@ spec:
...
+          done
         ports:
         - name: http
           containerPort: 3000
         resources:
           requests:
-            cpu: 100m
-            memory: 250Mi
+            cpu: 2
+            memory: 1Gi
```
jcortejoso, line 67: Probably this does not have too much effect

H34D:
should be obsolete with https://github.com/celo-org/ethstats-server/pull/13

cmcewen (approved):

jcortejoso:
Close bc master already has most of it.

PR #2018, opened 2019-12-04
Attestation Bot operational improvements

nambrot:
### Description

This PR is a follow on to https://github.com/celo-org/celo-monorepo/pull/1851

It:
- [x] Pass through Twilio address config
- [x] Correctly read the forno provider
- [x] suppress invalid signature error message
- [x] Select phone numbers from various countries

### Tested

Ran it locally against `integration`


packages/celotool/src/cmds/bots/auto-verify.ts
```diff
@@ -185,33 +196,18 @@ async function pollForMessagesAndCompleteAttestations(
       await sleep(POLLING_WAIT)
       continue
     }
-    console.log('')
+    console.info('')
```
jmrossy, line 199: Seems like we should use either logger or console but not both right?

packages/celotool/src/lib/attestation.ts
```diff
@@ -15,19 +16,23 @@ export async function requestMoreAttestations(
   account: Address,
   txParams: CeloTransactionParams = {}
 ) {
-  await attestations
-    .approveAttestationFee(attestationsRequested)
-    .then((txo) => txo.sendAndWaitForReceipt(txParams))
-  await attestations
-    .request(phoneNumber, attestationsRequested)
-    .then((txo) => txo.sendAndWaitForReceipt(txParams))
+  const unselectedRequest = await attestations.getUnselectedRequest(phoneNumber, account)
```
jmrossy, line 19: what does this do? Should I use it in the app?

packages/celotool/src/cmds/bots/auto-verify.ts
nambrot: So yeah I was considering taking it out. Basically, in the line above I print out `.` every `POLLING_WAIT` to indicate to a user running this that it is indeed still polling. the `console.info('')` is there to start a new line so that the next log line is a valid JSON line. I'd vote to keep it for that reason

packages/celotool/src/lib/attestation.ts
nambrot: If the process quits for whatever reason after we requested, but not yet selected, requesting again will fail. So it would make sense for you to adopt it as well in the mobile app.

jmrossy (approved):

PR #2020, opened 2019-12-04
Add node flag to celocli to set the target node for a single command

nategraf:
### Description

I found it added a lot of friction to have to use the `config:set` command everytime I wanted to target a different node with a `celocli` command. We expect validators to run multiple nodes, so they will likely feel the same. I added a quick flag to the base command to set it.

### Tested

I've been using this branch for a few days while working on other tasks and seen no issues.


timmoreton (approved):
Make sure you regenerate cli docs!

PR #2027, opened 2019-12-04
Fix broken links and spruce up CLI docs for accounts command

nategraf:
### Description

* I found a CLI took to identify broken links and fixed them 👍 
* I added some more description to the `claim-*` commands because I was initially confused as to what they do.


yerdua (approved):
Looks good to me! I really appreciate the extra clarity and context.

nambrot (approved):
Thanks so much for this, so much better!

Issue #1807, opened 2019-11-20
Devs SBAT easily deploy local testnets and control individual nodes


diminator:
This is about testing locally from source, not using GCP

- spin up network with X nodes
- node types:
  - [ ] transaction node
  - [x] validator
  - [ ] validator group
  - [x] proxy
  - [x] proxied validator
- [ ] contract migrations

PR #1907, opened 2019-11-26
Add local-testnet celotool command

nategraf:
### Description

This PR creates a command to spin up a number of geth validators, tx-nodes and light clients on your local computer. It's been quite useful to develop other changes.

### Tested

Not thoroughly tested

### Other changes

* Add `repository` option to e2e test configs

### Related issues

- Related https://github.com/celo-org/celo-monorepo/issues/1691

### Backwards compatibility

No effect on  backwards compatibility

PR #2037, opened 2019-12-04
Running local testnet from celotool

H34D:
### Description

- [x] spin up network with validator nodes

### Related issues

- Solves one aspect of https://github.com/celo-org/celo-monorepo/issues/1807


nategraf:
FYI: I've also been working on https://github.com/celo-org/celo-monorepo/pull/1907, mostly for my own benefit, but maybe some of the code would be useful here.

H34D:
> FYI: I've also been working on #1907, mostly for my own benefit, but maybe some of the code would be useful here.

amazing, thanks

H34D:
ci/circleci: contractkit-test is fixed in https://github.com/celo-org/celo-monorepo/pull/2150

H34D:
ci/circleci: protocol-test is fixed in https://github.com/celo-org/celo-monorepo/pull/2155

packages/celotool/package.json
```diff
@@ -35,20 +35,24 @@
   },
   "devDependencies": {
     "@types/bunyan": "1.8.4",
+    "@types/chai": "^4.1.3",
     "@types/dotenv": "^4.0.3",
+    "@types/mocha": "^5.2.1",
     "@types/node-fetch": "^2.1.2",
+    "@types/portscanner": "^2.1.0",
```
nategraf, line 42: Why was this version downgraded?

packages/celotool/package-lock.json
```diff
@@ -0,0 +1,979 @@
+{
```
nategraf: Why was this file added?

packages/celotool/package.json
```diff
@@ -35,20 +35,24 @@
...
     "@types/prompts": "^1.1.1",
     "@types/web3": "^1.0.18",
     "@types/yargs": "^13.0.2",
-    "@types/mocha": "^5.2.1",
-    "@types/chai": "^4.1.3",
     "chai": "^4.1.2",
-    "mocha": "^5.1.1"
+    "commander": "^2.20.3",
```
nategraf, line 47: Why is `commander` being added? Is that use case not covered by `yargs` and `prompts`?

packages/celotool/register.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
nategraf, line 1: Are you intending to check this in? If so, it'd be much better to have this part of a command in Typescript rather than as a shell script.

packages/celotool/register_validatorgroup.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
nategraf, line 1: Same comment as `packages/celotool/register.sh`

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -1,45 +1,36 @@
+import { consoleLogger } from '@celo/utils/lib/logger'
```
nategraf: I don't see a good reason to import this. It's just an alias for `console.log`
https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/logger.ts#L15

nategraf (commented):
Sorry I haven't had time to do a complete review yet. Here are some initial comments.

packages/celotool/package-lock.json
H34D: adding it was unintended, removed.

packages/celotool/package.json
H34D: This is the latest version of the portscanner types. 

we are using 
    `"portscanner": "^2.2.0"`
and 
    `"@types/portscanner": "^2.1.0",`

packages/celotool/src/e2e-tests/utils.ts
H34D: ok, removed

H34D:
given up for #2264

PR #2059, opened 2019-12-05
Some Docker script improvements

aaitor:
### Description

It removes the unused download function and moves the BLS and POS creation to the accounts function.

Moves also the proxy ip & enode to the validator function, allowing to stop and start the validator independently using the run-validator function

### Tested

In baklava

### Other changes

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


jeanregisser (approved):
LGTM 👍

PR #2061, opened 2019-12-05
Use correct phone placeholder depending on the country on joining Celo view

### LINKED ISSUE #865: Users SBAT see phone number input placeholder that matches the country they select
### Expected Behavior

The number of digits in the placeholder matches the country code's default

### Current Behavior

Hardcoded value

Screenshot_1567742041

Pedro-vk:
### Description

_Use correct phone placeholder depending on the country on joining Celo view_

### Tested

_Tested on Android phone_

### Related issues

- Fixes #865


packages/utils/src/phoneNumbers.ts
```diff
@@ -231,6 +231,34 @@ export function anonymizedPhone(phoneNumber: string) {
...
 }
 
+export function getExampleNumber(
+  regionCode: string,
+  showZero: boolean = true,
+  isInternational: boolean = false
+) {
+  let examplePhone =
+    phoneUtil.getExampleNumber(regionCode) ||
+    phoneUtil.getExampleNumber(getRegionCodeFromCountryCode(regionCode) as string)
```
jmrossy: Why do we need this line?

packages/utils/src/phoneNumbers.ts
```diff
@@ -231,6 +231,34 @@ export function anonymizedPhone(phoneNumber: string) {
...
+  }
+
+  const formatedExample = phoneUtil.format(
+    examplePhone,
+    isInternational ? PhoneNumberFormat.INTERNATIONAL : PhoneNumberFormat.NATIONAL
+  )
+
+  if (showZero) {
+    if (isInternational) {
+      return formatedExample.replace(/(^\+[0-9]{1,3} |[0-9])/g, (value, _, i) => (i ? '0' : value))
```
jmrossy, line 254: Can you explain what this is for?

packages/react-components/components/PhoneNumberInput.tsx
```diff
@@ -49,6 +50,7 @@ export default class PhoneNumberInput extends React.Component<Props, State> {
     phoneNumber: '',
     // country data should be fetched before mounting to prevent a second render
     countries: new Countries(this.props.lng),
+    inputPhonePlaceholder: this.props.inputPhonePlaceholder,
```
jmrossy: We should rename the prop, how about `initialInputPhonePlaceholder`?

packages/utils/src/phoneNumbers.ts
Pedro-vk: To be able to accept the country code and the region code (`+54` or `DE`).

packages/utils/src/phoneNumbers.ts
Pedro-vk: It returns the international example phone number with the next format: `+34 000 00 00 00`, so it keeps the prefix as numbers and the rest as `0`.

Is more clear in the test:
https://github.com/celo-org/celo-monorepo/pull/2061/files#diff-69002191f183e7738d1dcea3c12dd480R264-R282

packages/react-components/components/PhoneNumberInput.tsx
Pedro-vk: Sure, I'm going to change it

packages/utils/src/phoneNumbers.ts
jmrossy: Right but why do we need to accept either one? Can't we consistently pass in just one type?

packages/utils/src/phoneNumbers.ts
```diff
@@ -231,6 +231,34 @@ export function anonymizedPhone(phoneNumber: string) {
   return phoneNumber.slice(0, -4) + 'XXXX'
 }
 
+export function getExampleNumber(
+  regionCode: string,
+  showZero: boolean = true,
```
jmrossy: Suggest renaming to useOnlyZeroes

packages/utils/src/countries.ts
```diff
@@ -140,6 +146,10 @@ export class Countries {
         const localizedCountry = {
           names,
           displayName: names[this.language],
+          countryPhonePlaceholder: {
+            national: getExampleNumber(country.countryCallingCodes[0]),
+            internation: getExampleNumber(country.countryCallingCodes[0], true, true),
```
jmrossy: Don't think we need the international version right? Let's not calculate one for each country. Keep in mind this re-runs each time the app opens. 
Also, typo: `internatio` -> `international`

packages/utils/src/phoneNumbers.ts
jmrossy: I see. Why the lambda function instead of just a '0'?

packages/utils/src/phoneNumbers.ts
Pedro-vk: It's ok for me :+1:

packages/utils/src/phoneNumbers.ts
Pedro-vk: Because the first value is the prefix, that needs to be skipped in the transformation.

jmrossy (approved):
Looks good! Please address the `useZeros` comment before merging.

PR #2095, opened 2019-12-06
Slashing: Adding forceDeaffiliateIfValidator fn to Validators.sol

### LINKED ISSUE #2086: Allow validators to be removed from their current group by Slasher contracts
### Expected Behavior

```
Contract Validators {
  function removeSlashedMember(address account) external;
}
```
 Removes Only callable by the registered `DowntimeSlasher` and `DoubleSigningSlasher` contracts. Removes the account from its group if it is a member of one.

### Current Behavior

No way to remove slashed accounts from their groups.

lucasege:
### Description

Added a fn `forceDeaffiliateIfValidator` that wraps `_deaffiliate` on `Validators.sol` to ensure that the provided address is deaffiliated. This is to allow for easy interface for Slasher contracts to remove an offending validator from their current group. This fn can only be called by approved parties (slashing contracts, governance contract, and _for the stakeoff_ one preapproved node).

### Tested

Adding tests to ensure this fn is only callable by these addresses and that it properly removes addresses from their group.

### Related issues

- Fixes #2086


packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1076,4 +1076,14 @@ contract Validators is
     emit ValidatorDeaffiliated(validatorAccount, affiliation);
     return true;
   }
+
+  /**
+   * @notice Removes a validator from the group for which it is a member.
+   * @param validatorAccount The validator to remove and deaffiliate from their affiliated validator group.
+   */
+  function removeSlashedMember(address validatorAccount) external {
```
asaj: Please add a modifier restricting who can call this

packages/protocol/test/governance/validators.ts
```diff
@@ -2059,4 +2059,95 @@ contract('Validators', (accounts: string[]) => {
       })
     })
   })
+
+  describe.only('#removeSlashedMember', () => {
```
asaj: Remove the `.only`

packages/protocol/contracts/governance/Validators.sol
asaj: nit: rename to `deaffilliateSlashedValidator`

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1076,4 +1076,14 @@ contract Validators is
...
+    Validator storage validator = validators[validatorAccount];
```
asaj: What happens if the account is not a validator?

packages/protocol/test/governance/validators.ts
```diff
@@ -2059,4 +2059,95 @@ contract('Validators', (accounts: string[]) => {
...
+          assert.isTrue(await mockElection.isIneligible(group))
+        })
+      })
+    })
+
+    it('should revert when the account is not a registered validator', async () => {
+      await assertRevert(validators.removeSlashedMember(validator))
+    })
+
+    it('should revert when the validator is not affiliated with a validator group', async () => {
```
asaj: Is this the behavior we want?

packages/protocol/contracts/governance/Validators.sol
lucasege: If I understand the flow here, I'm assuming we want to create an `onlySlasher` modifier or something similar so that these fns can only be called from the Slasher contract, is that correct?

packages/protocol/contracts/common/UsingRegistry.sol
```diff
@@ -49,6 +53,15 @@ contract UsingRegistry is Ownable {
     _;
   }
 
+  modifier onlyRegisteredContracts(bytes32[] memory identifierHashes) {
+    bool registered = false;
+    for (uint256 i = 0; i < identifierHashes.length; i++) {
+      if (registry.getAddressForOrDie(identifierHashes[i]) == msg.sender) registered = true;
+    }
+    require(registered);
```
lucasege: If you want to do this a different way lmk, wasn't sure how much to try and "optimize" vs favoring readability

packages/protocol/contracts/governance/Validators.sol
lucasege: Following the precedent of `deaffiliate` I `require` here, but I'm still unsure on the exact assumptions on callers of this fn, so maybe this should be a simple if

packages/protocol/contracts/common/UsingRegistry.sol
asaj: This looks nice to me!

packages/protocol/contracts/common/UsingRegistry.sol
```diff
@@ -49,6 +53,15 @@ contract UsingRegistry is Ownable {
     _;
   }
 
+  modifier onlyRegisteredContracts(bytes32[] memory identifierHashes) {
+    bool registered = false;
+    for (uint256 i = 0; i < identifierHashes.length; i++) {
+      if (registry.getAddressForOrDie(identifierHashes[i]) == msg.sender) registered = true;
```
asaj: Consider breaking out of the loop when we get a match

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1083,4 +1083,25 @@ contract Validators is
     emit ValidatorDeaffiliated(validatorAccount, affiliation);
     return true;
   }
+
+  bytes32[] canForceDeaffiliation = [
```
asaj, line 1087: Make this a constant

packages/protocol/contracts/governance/Validators.sol
asaj: Yeah I would actually consider making this a no-op if the account is not a validator, and calling it `forceDeaffiliationIfValidator`.

If we require that the account is a validator, the Downtime and Double signing slashers will need to check that the account they're slashing is a validator before calling into this function, which is not as efficient as calling into `Validators` once and letting it decide whether or not the deaffiliation needs to take place

packages/protocol/test/governance/validators.ts
```diff
@@ -2059,4 +2059,21 @@ contract('Validators', (accounts: string[]) => {
       })
     })
   })
+
+  describe('#forceDeaffiliate', () => {
+    // TODO(lucas): instantiate governace/slasher contracts to test modifier
```
asaj: You don't actually need to do this, instead, you can register the address that you're making the `forceDeaffiliate` call with as one of the slashers.

There should be examples elsewhere in the tests of how to do this, but it's pretty straightforward

asaj:
Please update the PR title

asaj (commented):
For the stake-off, it would be useful to be able to force de-affiliate validators that go down and are unresponsive, until we have slashing for liveness. Can we update this PR to take an additional address that it will allow force-deaffiliations from, and update the migrations config to set that to CLabs validator-0?

packages/protocol/contracts/governance/Validators.sol
lucasege: Tried, got
`TypeError: Constants of non-value type not yet implemented.` 
Let me know if you know a way around this

packages/cli/src/commands/validator/forceDeaffiliate.ts
```diff
@@ -0,0 +1,36 @@
...
+    'Force deaffiliate a Validator from a Validator Group, and remove it from the Group if it is also a member.'
+
+  static flags = {
+    ...BaseCommand.flags,
+    from: Flags.address({ required: true, description: 'Initiator' }),
+    validator: Flags.address({ required: true, description: "Validator's address" }),
+  }
+
+  static examples = [
+    'forcedeaffiliate --from 0x47e172f6cfb6c7d01c1574fa3e2be7cc73269d95 --validator 0x...',
```
asaj: nit: Per our conventions use an example address here for the validator

packages/cli/src/commands/validator/forceDeaffiliate.ts
```diff
@@ -0,0 +1,36 @@
+import { BaseCommand } from '../../base'
+import { newCheckBuilder } from '../../utils/checks'
+import { displaySendTx } from '../../utils/cli'
+import { Flags } from '../../utils/command'
+
+export default class ValidatorForceDeaffiliate extends BaseCommand {
+  static description =
+    'Force deaffiliate a Validator from a Validator Group, and remove it from the Group if it is also a member.'
```
asaj: add: Used by stake-off admins in order to remove validators from the next epoch's validator set if they are down and consistently unresponsive, in order to preserve the health of the network. This feature will be removed once slashing for downtime is implemented.

packages/protocol/contracts/common/UsingRegistry.sol
```diff
@@ -49,6 +53,19 @@ contract UsingRegistry is Ownable {
...
 
+  modifier onlyRegisteredContracts(bytes32[] memory identifierHashes) {
+    bool registered = false;
+    for (uint256 i = 0; i < identifierHashes.length; i++) {
+      if (registry.getAddressForOrDie(identifierHashes[i]) == msg.sender) {
+        registered = true;
+        break;
+      }
+    }
+    bool isCLabsValZero = (msg.sender == address(0x0f1aa6BB9A2B72cdd8169B2A3Cb8eaF310Eaf756));
```
asaj: Add a TODO above to remove once the DowntimeSlasher is implemented

asaj: Double checking that this is clabs validator 0 on baklava

packages/protocol/test/governance/validators.ts
```diff
@@ -2059,4 +2059,35 @@ contract('Validators', (accounts: string[]) => {
...
+    })
+
+    describe('when the caller is the DowntimeSlasher Address', async () => {
+      beforeEach(async () => {
+        await registry.setAddressFor(CeloContractName.DowntimeSlasher, validator)
+        await registry.setAddressFor(CeloContractName.DoubleSigningSlasher, accounts[3])
+        await registry.setAddressFor(CeloContractName.Governance, accounts[4])
+      })
+
+      it('should succeed when sender is the downtime slasher contract', async () => {
```
asaj: Add tests that it should succeed when called by governance, doublesigningslasher, e.g. `await validators.forceDeafiliateIfValidator(validator, { from: accounts[3] })`

asaj (approved):

packages/protocol/contracts/common/UsingRegistry.sol
lucasege: It is now

packages/cli/src/commands/validator/forceDeaffiliate.ts
```diff
@@ -0,0 +1,36 @@
...
+    "Force deaffiliate a Validator from a Validator Group, and remove it from the Group if it is also a member.  Used by stake-off admins in order to remove validators from the next epoch's validator set if they are down and consistently unresponsive, in order to preserve the health of the network. This feature will be removed once slashing for downtime is implemented."
+
+  static flags = {
+    ...BaseCommand.flags,
+    from: Flags.address({ required: true, description: 'Initiator' }),
+    validator: Flags.address({ required: true, description: "Validator's address" }),
+  }
+
+  static examples = [
+    'forcedeaffiliate --from 0x47e172f6cfb6c7d01c1574fa3e2be7cc73269d95 --validator 0xb7ef0985bdb4f19460A29d9829aA1514B181C4CD',
```
asaj: is it `forcedeaffiliate` or `forceDeaffiliate`? I thought it was based on the filename. If the latter, consider renaming to `force-deaffiliate.ts` for consistency and updating the example here

asaj (approved):
🔥

PR #2096, opened 2019-12-06
Stop running broken end-to-end mobile test

asaj:
### Description

As far as I can tell, this test is failing every time. Remove it so we can (sometimes) get green builds.

### Tested

Not tested.


martinvol (changes requested):
e2e mobile test should be fixed by #1987

asaj:
This is still broken, can we remove?

martinvol (approved):
Test broke again at yarn level. I'm in favor of disabling them while we come up with a robust solution.

PR #2102, opened 2019-12-06
docs: removes sleep from commands

kobigurk:
### Description

The `sleep 1` prevents the infamous liner error. https://github.com/celo-org/celo-blockchain/pull/703 adds it as part of the docker image.


asaj (approved):

asaj:
Looks so much cleaner using the default entrypoint!

PR #2104, opened 2019-12-06
Calculate epoch rewards adjusted for validator group uptime

### LINKED ISSUE #1836: Base rewards to voters on the uptime of the group they voted for
### Expected Behavior

Epoch rewards to voters within a group are multiplied by the average epoch score of elected members in that group.

### Current Behavior

Not based on uptime of group members

nategraf:
### Description

Voters should be awarded only a portion of the possible rewards when the group they vote for has less that perfect uptime. https://github.com/celo-org/celo-monorepo/issues/1836

This PR implements, in the core contracts, epoch score calculation for groups and uses that score to adjust the epoch rewards for each group.

### Tested

Unit tests and e2e tests

### Other changes

* Add precision value to test in epoch rewards to avoid flakes.
* Copy `waitForPort` from e2e tests to wait for ganache to be online.

### Related issues

- Fixes #1836

### Backwards compatibility

Hard fork requiring node upgrade


packages/protocol/contracts/governance/Election.sol
```diff
@@ -412,21 +412,30 @@ contract Election is
...
+    uint256[] calldata uptimes
+  ) external view returns (uint256) {
+    // The group must meet the balance requirements for their voters to receive epoch rewards.
+    if (!getValidators().meetsAccountLockedGoldRequirements(group) || votes.active.total <= 0) {
       return 0;
     }
+
+    uint256 scoreNumerator;
+    uint256 scoreDenominator;
+    (scoreNumerator, scoreDenominator) = getValidators().calculateGroupEpochScore(uptimes);
```
asaj: Can we just return the fixidity representation of the numerator?

packages/protocol/contracts/governance/Election.sol
```diff
@@ -412,21 +412,30 @@ contract Election is
...
+
+    return
+      totalEpochRewards
```
asaj: We'd get more precision by doing both multiplications before both divisions. If we can convince ourselves that this will never overflow, let's do so

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -324,6 +325,51 @@ contract Validators is
...
+    uint256 numerator;
+    uint256 denominator;
+    FixidityLib.Fraction memory sum;
+    for (uint256 i = 0; i < uptimes.length; i = i.add(1)) {
+      (numerator, denominator) = calculateEpochScore(uptimes[i]);
+      sum = sum.add(FixidityLib.newFixedFraction(numerator, denominator));
+    }
+    return (
+      sum.divide(FixidityLib.newFixed(uptimes.length)).unwrap(),
+      FixidityLib.fixed1().unwrap()
```
asaj: No need to return this

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -324,6 +325,51 @@ contract Validators is
     return (epochs, membershipGroups, history.lastRemovedFromGroupTimestamp);
   }
 
+  /**
+   * @notice Calculates the validator score for an epoch from the uptime value for the epoch.
+   * @param uptime The Fixidity representation of the validator's uptime, between 0 and 1.
+   * @dev epoxh_score = uptime ** exponent
+   * @return Numerator and denominator of the epoch score.
+   */
+  function calculateEpochScore(uint256 uptime) public view returns (uint256, uint256) {
```
asaj: Consider having this return the fixidity representation of the uptime score

packages/protocol/runTests.js
```diff
@@ -51,9 +51,8 @@ async function test() {
   try {
     const closeGanache = await startGanache()
     if (isCI) {
-      // if we are running on circle ci we need to wait for ganache to be up
-      // TODO(mcortesi): improvement: check for open port instead of a fixed wait time.
-      await sleep(60)
+      // If we are running on circle ci we need to wait for ganache to be up.
+      await waitForPortOpen('localhost', 8545, 60)
```
asaj, line 55: 🔥

asaj (commented):
Looks great!

And end-to-end test with a down member would give us more assurance that everything WAI, fine for that to be in a follow-up PR

asaj (approved):
Whoops, meant to approve 😄

packages/protocol/contracts/governance/Validators.sol
nategraf: Done

packages/protocol/contracts/governance/Validators.sol
nategraf: Fixed

packages/protocol/contracts/governance/Election.sol
nategraf: In theory, this could just barely overflow, or come close to it, so I refactored it to use fixidity to maintain precision a little better.

packages/protocol/contracts/governance/Election.sol
nategraf: Yes, and I've done so now.

asaj:
There seems to be an issue with the end-to-end tests. Namely, it looks like rewards distribution is failing. What's scary is that the only test that seems to catch this is the events test, which sees that no logs were emitted.

Really curious why the other tests didn't catch that rewards were failing. We *should* revert all PoS related state changes when any action fails, so all of the various tests should have caught this error.

Attaching my logs to save you a bit of time.
[logs.txt](https://github.com/celo-org/celo-monorepo/files/3943505/logs.txt)

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -716,25 +718,36 @@ describe('governance tests', () => {
...
-        'EpochRewardsDistributedToVoters',
-        { fromBlock: currentBlock - 10, currentBlock }
-      )
-      assert(epochRewardsEvents.every((a: any) => a.blockNumber % 10 === 0))
-      assert(validatorRewardsEvents.every((a: any) => a.blockNumber % 10 === 0))
-      assert(electionRewardsEvents.every((a: any) => a.blockNumber % 10 === 0))
-      assert(epochRewardsEvents.length > 0)
-      assert(validatorRewardsEvents.length > 0)
-      assert(electionRewardsEvents.length > 0)
+      const currentBlock = (await web3.eth.getBlock('latest')).number
```
asaj, line 721: This is much nicer 👌

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -659,11 +659,13 @@ describe('governance tests', () => {
           const targetVotingYield = new BigNumber(
             (await epochRewards.methods.getTargetVotingYieldParameters().call({}, blockNumber))[0]
           )
+          assert.isFalse(targetVotingYield.isZero())
```
asaj, line 662: Great call

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -55,7 +55,10 @@ export function assertAlmostEqual(
   if (expected.isZero()) {
     assert.equal(actual.toFixed(), expected.toFixed())
   } else {
-    const isCloseTo = actual.plus(delta).gte(expected) || actual.minus(delta).lte(expected)
```
asaj, line 58: 🤦‍♂

asaj (approved):
🚢

PR #2105, opened 2019-12-06
Add more logging to the attestation bot

nambrot:
### Description

This PR adds more logging to the attestation-bot to identify issuers who are not running the attestation service appropriately

### Tested

- Not tested


cmcewen (approved):

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -215,37 +219,63 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
-        }
-      }
+      this.isIssuerRunningAttestationService
     )
 
-    return withAttestationServiceURLs.filter(notEmpty)
+    return withAttestationServiceURLs.map((_) => (_.isValid ? null : _.issuer)).filter(notEmpty)
+  }
+
+  private async isIssuerRunningAttestationService(arg: {
```
jmrossy, line 250: Nit: I generally expect functions that start with `is` to return a boolean

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -215,37 +219,63 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
 
     const result = await this.contract.methods.getCompletableAttestations(phoneHash, account).call()
 
+    const results = await concurrentMap(
```
jmrossy, line 222: A more descriptive name than `results` would help here, esp. since theres a result var just above

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -215,37 +219,63 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
+
+    return results.map((_) => (_.isValid ? _.result : null)).filter(notEmpty)
+  }
+
+  /**
+   * Returns an array of issuer addresses that were found to not run the attestation service
+   * @param phoneNumber
+   * @param account
+   */
+  async getNonCompliantIssuers(phoneNumber: string, account: Address): Promise<Address[]> {
```
jmrossy, line 236: Seems like this is exactly the same as the getActionableAttestations method but with the ternary in the last line reversed. Can we DRY this up?

jmrossy (commented):
Mostly looks good just a few small things

PR #2106, opened 2019-12-06
Provide an endpoint that can accept POST requests

nambrot:
### Description

Twilio would complain that the specified endpoint does not actually accept POST requests. Since we just poll for messages anyways, I'm specifying an empty requestbin.com endpoint

### Tested

- Not tested


nategraf:
What kind of post requests are these? Is there any risk of leaking security information?

nambrot:
They are the text messages requested by the attestation bot. With knowledge of the phone number and access to the requestbin, you could verify those phone numbers, but I'm not sure that that's any risk.

nategraf (approved):

PR #2113, opened 2019-12-06
Document the need to counter-claim accounts

nambrot:
### Description

We never told validators that they need to claim their validator account address from the validator group account's metadata, so this PR makes that addition

### Tested

- Ran it on my validator


packages/docs/getting-started/running-a-validator.md
```diff
@@ -610,7 +610,17 @@ If everything goes well users should be able to see your claims by running:
 celocli account:get-metadata $CELO_VALIDATOR_ADDRESS
 ```
 
-By now, you should have setup your Validator account appropriately.
+You should see that your claim for `$CELO_VALIDATOR_GROUP_ADDRESS` could not be verified! We need to create the corresponding claim from `$CELO_VALIDATOR_GROUP_ADDRESS` otherwise anyone could claim it!
```
asaj, line 613: We should note, very explicitly, that this is needed to have both your accounts appear on the leaderboard.

Furthermore, if people create additional accounts (e.g. when they add an additional validator), they will need to claim those accounts as well.

Please make sure this information is hard to miss

asaj (approved):

PR #2114, opened 2019-12-06
Deploy CLI 0.0.30-beta9 and ContractKit 0.2.7

nambrot:
### Description

As the title suggests, deployed new versions of the CLI and ContractKit. Noticed a bug in ContractKit and immediately fixed it (from the circular dependency)

### Tested

- Installed the CLI and it ran as expected


jmrossy (approved):
Looks good thanks! Did my PR which refactored some attestations code cause the break? Should have thought to redeploy, sorry! The dangers of merging on a Friday  night. Do we need to also deploy utils since it has some code I moved from contract kit?

nambrot:
Uhm, whatever I deployed with this PR does work right now, so I don't think you broke it. Though I'm not entirely sure whether your changes were in this deploy or not. I might have branched just before you. Certainly something to watch out for on the next deploy though

Issue #2115, opened 2019-12-07
Stake Off participants SBAT to see BTUs in detailed leaderboard 


clrblmt:
### Expected Behavior

Add column with BTUs + button with link from celo.org/build to file below

### Current Behavior

Detailed leaderboard does't have BTUs listed https://docs.google.com/spreadsheets/d/1Me56YkCHYmsN23gSMgDb1hZ_ezN0sTjNW4kyGbAO9vc/edit#gid=0

PR #2118, opened 2019-12-08
[cli] Showing summed up balances for claimed accounts

mrsmkl:
### Description

In the balance command, now shows also the sum of cGLD, cUSD and locked gold balance.
Added command `account:show-claimed-accounts`, it will show all the claimed accounts and their balances.

### Tested

Just tested on few account on Baklava.

### Related issues

I'll use this code for the script needed to implement #2115


packages/cli/src/commands/account/show-claimed-accounts.ts
```diff
@@ -0,0 +1,88 @@
...
+  kit: ContractKit,
+  address: string,
+  data: IdentityMetadataWrapper
+): Promise<string[]> {
+  if (address.substr(0, 2) === '0x') {
+    address = address.substr(2)
+  }
+  const res = [address]
+  const accounts = await kit.contracts.getAccounts()
+  for (const claim of data.claims) {
```
nambrot: Consider using this:

```typescript
data.filterClaims(ClaimTypes.Account).map(async (claim) => await verifyAccountClaim(claim, ensure0x(address), accounts.getMetadataUrl) ? claim.address : null).filter(noEmpty) 
```

packages/contractkit/src/kit.ts
```diff
@@ -75,6 +76,18 @@ export class ContractKit {
     this.contracts = new WrapperCache(this)
   }
 
+  async getTotalBalance(address: string): Promise<BigNumber> {
```
mcortesi: What about returning an object with all the different type of balances?

Something like:
```ts
interface AccountBalance {
   gold: BigNumber,
   usd: BigNumber.
   lockedGold: BigNumber
}
```
?

Also, converting "stable -> gold" sounds like a good add to the `Exchange` wrapper.. probably already has one, since you can ask a quote of how many gold will i received when exchanging XXX stable. In fact, it's call `Exchange.quoteUsdSell()`

mcortesi (approved):
Thanks!!

PR #2126, opened 2019-12-09
Fixing the cli for exchange

### LINKED ISSUE #2116: celocli exchange:gold problem
### Expected Behavior

`celocli exchange:gold --for=90 --from=myaddr --value=100` should exchange my 100 cUSD for new 90 cGold.

### Current Behavior
If I run the command, celocli complains that 100 and 90 are not valid addresses. 
I think the problem is that following part

https://github.com/celo-org/celo-monorepo/blob/1bee3f4c6d28d862cbaff6d3a04c67c70dc11090/packages/cli/src/commands/exchange/dollars.ts#L16

assumes that "from" and "value" are addresses. And there is a discrepancy between the help message and the actual API call. The value for `--commission` is not used in actual API call.

mrsmkl:
### Description

Changed the `--value` and `--for` parameters into numeric values.

### Tested

Didn't test properly yet, network down.

### Related issues

- Fixes #2116


timmoreton (approved):
👍

asaj:
Can you remove the unused `commission` flag as well?

PR #2132, opened 2019-12-09
[NotificationService] Change exchange rate stored format (group by pair)

jeanregisser:
### Description

Follow up to #1945

Firebase RTDB querying capabilities are quite limited, once we query a specific time range using `orderByChild`, we can't then filter on a specific pair.
So here we workaround that problem by storing exchange rates under their pair.

### Tested

Deployed on integration.

**Before**

Screenshot 2019-11-27 at 16 27 40

**After**

Screenshot 2019-12-09 at 17 04 15

### Other changes

Speed up jest tests, similar to what's been done in #2005

### Related issues

- #834

### Backwards compatibility

No. I'm deleting existing stored exchange rates in firebase when deploying this.
This is safe to do at this point since there's no consumer of this data yet.


jmrossy (approved):

annakaz (approved):
LGTM!

PR #2138, opened 2019-12-09
Leaderboard Improvements

### LINKED ISSUE #2135: Leaderboard on mobile has squashed x axis labels
Looks Like 

Screen Shot 2019-12-09 at 10 55 36 AM

should be able to see separate numbers

aaronmgdr:
### Description

Three Main objectives

1. Fix x-axis point labels being squashed together on mobile
2. add `#leaderbaord` for direct linking
3. fix issue of line going thru coin rather than ending at it on some devices

### Tested

on firefox a chrome on mac at various sizes 

### Other changes

- Move Lazyloading from outside of entire leaderboad into each individual Racer (for `#1`)
- Rather than a spinner for loading just load with zero points inituall (for `#2`)
- Animate coin and line containing div rather than individually.

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/2135


packages/web/src/dev/Racer.tsx
```diff
@@ -20,49 +20,29 @@ export default React.memo(function Racer({
   const lineAnimationKeyframes = [
     {
       from: {
-        transform: [{ scaleX: 1 }],
+        transform: [{ translateX: -relativePoints }],
```
aaronmgdr, line 23: Easiest thing to do here would be to animate width from `0 `to `X`, however animating width isn't as smooth as animating transforms. next best would be to set width to x and scale from 0 to 1 however if we just scale the line the coin is just floating all awkwardly, and if we scale the div they are both in the coin goes from skinny af to the oval we know and love. 

prior solution was to animate the line width and coin position as separate animations. which worked but meant they could get out of sync or just mess up as connor demoed. 

therefore set width as desired and then just translateX the entire thing to the left to start. (hiding anything that overflows) 

🎉

cmcewen (approved):

PR #2139, opened 2019-12-10
Governance ContractKit + CLI changes

nambrot:
### Description

This is all @yorhodes 's work to make governance proposals and hotfixes work in contractkit + CLI, just updated to master


jmrossy (commented):
The /mobile and /utils changes lgtm
Looks like tests are failing though.

packages/cli/src/commands/governance/propose.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  async run() {
+    const res = this.parse(Propose)
+
+    const governance = await this.kit.contracts.getGovernance()
+
+    const proposal = await buildProposalFromJsonFile(this.kit, res.flags.jsonTransactions)
+    const tx = governance.propose(proposal)
+    await displaySendTx('proposeTx', tx, { from: res.flags.from, value: res.flags.deposit })
+    const proposalID = await tx.txo.call()
```
mcortesi: does this work?
you first "send the tx" and after that make the same `call`. But will it result on the same proposalID?

mcortesi (approved):

packages/cli/src/commands/governance/propose.ts
yorhodes: definitely not
opened #2335

PR #2142, opened 2019-12-10
Celotool command for deploying hotfixes

nambrot:
### Description

This is a PR on top of #2139 for a celotool command that actually deploys a hotfix (assuming we have sufficient validating power). It also adds the ability to easily define proxy repointings.

### Tested

- Executed a proposal on `namgprop`

### DIFF
--- a/packages/celotool/src/cmds/deploy/upgrade/hotfix.ts
+++ b/packages/celotool/src/cmds/deploy/upgrade/hotfix.ts
@@ -51,11 +51,10 @@ export const handler = async (argv: EthstatsArgv) => {
       kit.web3,
       validatorsProxyAddress
     )
-    const desiredImplementationAddress = '0x1D8d4472DC1bE4B277b4c4825A9f1c021cd27014'
-
+    const desiredImplementationAddress = '0xd18620a5eBE0235023602bB4d490E1e96703EddD'
     console.info('Current Implementation Address: ', currentValidatorsImplementationAddress)
 
-    console.info('Build Proposal')
+    console.info('\nBuild Proposal')
 
     const proposalBuilder = new ProposalBuilder(kit)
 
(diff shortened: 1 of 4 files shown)


packages/celotool/src/cmds/deploy/upgrade/hotfix.ts
```diff
@@ -0,0 +1,151 @@
...
+    //   value: '0',
+    // })
+
+    // Example B
+    proposalBuilder.addProxyRepointingTx(validatorsProxyAddress, desiredImplementationAddress)
+
+    const proposal = await proposalBuilder.build()
+    const proposalHash = proposalToHash(kit, proposal)
+
+    // If your proposal is just made of ContractKit methods, you can print it out
```
yorhodes: I think this works for non contractkit methods
Requirement is contract in celo registry for now

packages/contractkit/src/governance/proposals.ts
```diff
@@ -80,7 +88,7 @@ export class ProposalBuilder {
       if (!method) {
         throw new Error(`Method ${methodName} not found on ${tx.contract}`)
       }
-      const txo = method(tx.args)
+      const txo = method(...tx.args)
```
yorhodes, line 91: +1

packages/contractkit/src/governance/proposals.ts
```diff
@@ -60,6 +61,13 @@ export class ProposalBuilder {
     input: tx.encodeABI(),
   })
 
+  addProxyRepointingTx = (proxyAddress: string, newImplementationAddress: string) => {
```
yorhodes, line 64: I like this pattern!

packages/contractkit/src/governance/proxy.ts
```diff
@@ -0,0 +1,46 @@
+import Web3 from 'web3'
+
+const PROXY_ABI = [
```
yorhodes, line 3: Is stubbing the ABIs a standard practice? Any reason why we do this here/in e2e tests instead of using the generated ABI? I used the generated governance ABI in the proposal builder

mcortesi: i think it's ok here. Since the Proxy is something that won't change.

mcortesi (approved):
Not sure if celotool upgrade is the right place to put examples... but works for now

PR #2143, opened 2019-12-10
Allow the validator account to whitelist hotfixes

### LINKED ISSUE #1832: Hotfix whitelists should use validator account address instead of consensus key address
### Expected Behavior

Validators don't need to use their consensus key (on secure hardware) for interacting with Hotfixes

### Current Behavior

Hotifx uses keys from current validator set precompile without lookup of registered ephemeral key

https://github.com/celo-org/celo-monorepo/blob/6b0f1de881cd87168cd6a6d3f8600cd8d3a12648/packages/protocol/contracts/governance/Governance.sol#L859

nambrot:
### Description

This PR is on top of #2142 and allows validators to whitelist hotfixes with their account key, instead of requiring to do it with their validation signer keys. It fixes #1832. However, given the current urgency, it does not include the following:

- Unit tests
- CLI checks

### Tested

- Not tested

### Related issues

- Fixes #1832

### Backwards compatibility

Should be backwards compatible

### DIFF
--- a/packages/protocol/contracts/governance/Governance.sol
+++ b/packages/protocol/contracts/governance/Governance.sol
@@ -631,7 +631,7 @@ contract Governance is
   /**
    * @notice Returns whether given hotfix hash has been whitelisted by given address.
    * @param hash The abi encoded keccak256 hash of the hotfix transaction(s) to be whitelisted.
-   * @param whitelister Address to check whitelist status of. 
+   * @param whitelister Address to check whitelist status of.
    */
   function isHotfixWhitelistedBy(bytes32 hash, address whitelister) public view returns (bool) {
     return hotfixes[hash].whitelisted[whitelister];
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/governance/Governance.sol
```diff
@@ -877,8 +877,12 @@ contract Governance is
     uint256 tally = 0;
     uint256 n = numberValidatorsInCurrentSet();
     for (uint256 idx = 0; idx < n; idx++) {
-      address validator = validatorAddressFromCurrentSet(idx);
-      if (isHotfixWhitelistedBy(hash, validator)) {
+      address validationSigner = validatorAddressFromCurrentSet(idx);
```
asaj, line 880: nit: `validatorSigner`

asaj, line 880: Please pull in `asaj/hotfix-signer`, which renames this to `validatorSignerAddressFromCurrentSet`

nambrot:
Closing since #2166 made it in

Issue #1773, opened 2019-11-19
Flaky contract kit tests


asaj:
### Expected Behavior

ContractKit tests pass always.

### Current Behavior

Occasionally, all tests fail. Likely something to do with the setup
```
    console.error ../../node_modules/jest-jasmine2/build/jasmine/Env.js:289
      Unhandled error
    console.error ../../node_modules/jest-jasmine2/build/jasmine/Env.js:290
      TypeError: Cannot read property 'pop' of undefined
          at c.x._updateNode (/home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:145275)
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:143284
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:146795
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:392
          at c (/home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:5407)
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:317
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:392
          at c (/home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:5407)
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:317
          at e (/home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:146267)
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:146726
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:143836
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:2:94440
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:392
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:5626
          at /home/circleci/app/node_modules/@celo/ganache-cli/build/ganache-core.node.cli.js:25:317
[...]

PR #2150, opened 2019-12-10
Fix/contractkit test flakyness

### LINKED ISSUE #1996: Make ContractKit tests not flaky
### Expected Behavior

No flaky tests

### Current Behavior

Flaky contractkit tests

H34D:
### Description

- Sequencing test runs alphabetic
- Add more timeout to long running tests
- Properly log ganache startup
- Give ganache some time to start before running first tests

### Tested

ran 5 times:
- yarn test:reset
- yarn test

all tests were green

Closes: #1996 #1773


packages/contractkit/src/wrappers/Accounts.test.ts
```diff
@@ -6,6 +6,8 @@ import { AccountsWrapper } from './Accounts'
 import { LockedGoldWrapper } from './LockedGold'
 import { ValidatorsWrapper } from './Validators'
 
+jest.setTimeout(10 * 10000)
```
jmrossy: 100 seconds? seems high

jmrossy (commented):
Mostly looks good! one comment below. Recommend waiting for @mcortesi approval since he knows this package best :)

packages/contractkit/src/wrappers/Accounts.test.ts
H34D: whoops, wanted 10 seconds

packages/contractkit/src/test-utils/AlphabeticSequencer.ts
```diff
@@ -0,0 +1,14 @@
+const Sequencer = require('@jest/test-sequencer').default
+
+class AlphabeticSequencer extends Sequencer {
+  // @ts-ignore
+  sort(tests) {
+    const copyTests = Array.from(tests)
+    return copyTests.sort(
+      // @ts-ignore
```
mcortesi: why the `ts-ignore`?

packages/contractkit/src/test-utils/AlphabeticSequencer.ts
```diff
@@ -0,0 +1,14 @@
+const Sequencer = require('@jest/test-sequencer').default
+
+class AlphabeticSequencer extends Sequencer {
```
mcortesi: Why do we need a sequencer? I would put some documentation about why & how it works... or what it solves

packages/contractkit/src/test-utils/ganache.setup.ts
```diff
@@ -3,6 +3,15 @@ import baseSetup from '@celo/dev-utils/lib/ganache-setup'
...
+async function sleep(seconds: number) {
+  return new Promise((resolve) => {
+    setTimeout(resolve, seconds * 1000)
+  })
+}
+
+export default async function setup() {
+  console.log('\nstarting ganache...')
+  await baseSetup(path.resolve(path.join(__dirname, '../../.devchain')))
+  await sleep(10)
```
mcortesi: 10 seconds seems a bit excessive, but not sure what was the problem here, again maybe document why we do this...

mcortesi (commented):
There's a `dev-utils` package that @nambrot created, i though contractkit was already using that, since it contains ganache setup and other test utilities.

Maybe we should migrate contractkit to use it, and do the changes there. For now, `celocli` use the same setup, and the attestantion service i believe.

packages/contractkit/src/test-utils/ganache.setup.ts
H34D: Ganache needs some time to start up. I can put this to 5 seconds maybe but then it might fail again from time to time. With 10 seconds it did not fail a single time in my tests. 

Where exactly shall i document this? Inline or in a separate document?

packages/contractkit/src/test-utils/AlphabeticSequencer.ts
H34D: By default jest uses its internal sequencer like this:

https://github.com/facebook/jest/blob/master/packages/jest-test-sequencer/src/index.ts#L83-L110

this sequencer has some logic built in to re-run and re-prioritise test runs. This leads to a different order on every single test run. With an alphabetic order we get a more deterministic testrun than with the built in sequencer. It is more reproducible this way. 

The general problem is that we are using jest for blockchain tests, this is not the best technology fit because it was made for heavy parallelization. This is already being mitigated with the `--runInBand` flag sequencing the tests over all.

Where exactly shall i document this? Inline or in a separate document?

packages/contractkit/src/test-utils/AlphabeticSequencer.ts
H34D: > There's a `dev-utils` package that @nambrot created, i though contractkit was already using that, since it contains ganache setup and other test utilities.
> 
> Maybe we should migrate contractkit to use it, and do the changes there. For now, `celocli` use the same setup, and the attestantion service i believe.

Amazing idea, i will open an issue for it.

H34D: > why the `ts-ignore`?

migrated sequencer back to js, removed the ts-ignore

H34D:
> There's a `dev-utils` package that @nambrot created, i though contractkit was already using that, since it contains ganache setup and other test utilities.
> 
> Maybe we should migrate contractkit to use it, and do the changes there. For now, `celocli` use the same setup, and the attestantion service i believe.

~Amazing idea, i will open an issue for it.~

After looking at the code again, it is using `dev-utils` already this is just the adapter code. All fine.

`import baseSetup from '@celo/dev-utils/lib/ganache-setup'`

packages/contractkit/src/test-utils/ganache.setup.ts
asaj: PTAL @nategraf 's recent PR that added some tooling to spin until ganache comes up. Seems like it should be re-used here

https://github.com/celo-org/celo-monorepo/pull/2104/files#diff-7c886cc81e742d2a57cd84d6a6e42e58R137

H34D: - carved out waitForPortOpen from celotool and moved it to dev-utils
- migrated ganache setup from sleep to waitForPortOpen from dev-utils
in https://github.com/celo-org/celo-monorepo/pull/2150/commits/16e049ff180630cbaac7744869c2842f254c4fe5

packages/contractkit/src/test-utils/AlphabeticSequencer.ts
mcortesi: Thanks for the explanation...
I would document as a comment what you just said...  for future maintainers to know why we are using a diff sequencer.

packages/contractkit/src/test-utils/ganache.setup.ts
mcortesi: Great!
Quick note about it, `waitForPortOpen()` requrires `nc` to be installed. But i guess it's already there for every environment where we run the tests.

mcortesi (approved):
Looks Great!

I've answered some of the comments from the previous review

H34D:
> Looks Great!
> 
> I've answered some of the comments from the previous review

ill put the pending changes on my list

PR #2154, opened 2019-12-10
[Wallet] Fix bug causing banner to stay visible once app is connected

jmrossy:
### Description

- Fix bug causing app banner to stay visible and app to navigate to error screen after 2 minutes
- Add missing i18n string for 'at'

### Tested

Synced the app
Ran through onboarding

### Backwards compatibility

Yes


packages/mobile/src/tokens/saga.ts
```diff
@@ -46,7 +46,6 @@ export async function convertToContractDecimals(value: BigNumber, token: CURRENC
 
 export async function getTokenContract(token: CURRENCY_ENUM) {
   Logger.debug(TAG + '@getTokenContract', `Fetching contract for ${token}`)
-  await waitForWeb3Sync()
```
jmrossy, line 49: @annakaz this was calling the waitForWeb3Sync saga from outside 'saga land'. It doesn't work and navigates the app to error after the sync timeout. Did we definitely need this here for something?

packages/mobile/src/web3/saga.ts
```diff
@@ -353,7 +357,7 @@ export function* unlockAccount(account: string) {
 // Wait for geth to be connected and account ready
 export function* getConnectedAccount() {
   yield call(waitWeb3LastBlock)
-  const account: string = yield getAccount()
+  const account: string = yield call(getAccount)
```
jeanregisser, line 360: 👍

packages/mobile/src/redux/sagas.ts
```diff
@@ -77,6 +77,6 @@ export function* rootSaga() {
   yield spawn(dappKitSaga)
   yield spawn(feesSaga)
   yield spawn(localCurrencySaga)
-  yield spawn(web3Saga)
   yield spawn(sentrySaga)
+  yield spawn(web3Saga)
```
jeanregisser, line 81: Is this important for the fix?

jeanregisser (approved):
Thanks for digging this 👍 
I'm a bit confused by which part of the diff fixed it though.

packages/mobile/src/redux/sagas.ts
jmrossy: It conceptually makes sense to wait til the other sagas are set up before we run web3 b.c. that saga will check the sync status. I moved it while debugging. Not 100% sure if it's *required* for the fix though.

jmrossy: The required parts of a combination of adding a spawn(waitForWeb3) line to the web3 saga and removing the await waitforweb3 line from the getToken utility

packages/mobile/src/tokens/saga.ts
annakaz: Thanks for catching this. We have to ensure contractkit is connected to get the TokenContracts, but I just confirmed that we call `getConnectedAccount` every time before calling `getTokenContract`, so removing this is fine. Could be helpful to throw an error explaining this if it's called when the account isn't connected, but I can add this in another PR

PR #2159, opened 2019-12-10
Do not open RPC port to public

### LINKED ISSUE #2158: Document should not instruct to expose RPC port to public when setting up a validator node
### Expected Behavior

No RPC port of Celo node is opened for public if there is no need.

### Current Behavior

RPC port is opened for public when setting up `celo-accounts` node by following https://docs.celo.org/getting-started/baklava-testnet/running-a-validator.

hqueue:
### Description

Update document to avoid opening RPC port to public when running a celo-accounts node.

### Tested

Tested in Baklava.

### Related issues

- Fixes #2158 

### Backwards compatibility
- No code change

### DIFF
--- a/packages/docs/getting-started/running-a-validator.md
+++ b/packages/docs/getting-started/running-a-validator.md
@@ -157,7 +157,7 @@ To run the node:
 
 ```bash
 # On your local machine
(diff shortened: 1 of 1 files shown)


timmoreton (approved):
Great spot, and thanks for the contribution!

Issue #786, opened 2019-08-29
HOGLers SBAT see more locked gold info including rewards via CLI


timmoreton:
Voters in validator elections should be able to understand the history of their epoch rewards. Currently, there is no way to do so, other than examining their locked gold info before and after the last block of an epoch. This can get muddled if the user made a transaction in that block that would affect this information, e.g. activating additional votes.

We should have a command, e.g. `rewards:show` that will print out the history of rewards for a particular address. This information can be gleaned from the events emitted in block finalization.

Validator/group payments should be relatively straightforward to figure out from the events, rewards to voters may require a bit more work.

diminator:
I can take this if there's a bit more explanation

PR #2160, opened 2019-12-10
cli: Add rewards:show

### LINKED ISSUE #787: Validators and Groups SBAT see payments received
Validators and groups currently have no way of viewing the payments they've received for validating, other than checking their balances before and after the last block of an epoch. This is an unreliable method, as they may have sent/received Celo Dollars during that block.

We should instead read this information from the block finalization receipts

jfoutts-celo:
### Description

Add rewards:show command that shows rewards for the last N epochs.
If --address is supplied: voter, validator, and validator group rewards applying to address are filtered.

### Tested

Tested that rewards are correctly displayed by the CLI on my testnet.

### Related issues

- Fixes #787 and #786


packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,173 @@
...
+    'Show rewards information about a voter, registered Validator, or Validator Group'
+
+  static flags = {
+    ...BaseCommand.flags,
+    address: Flags.address({ required: false, description: 'Address to filter' }),
+    epochs: flags.integer({ required: false, description: 'Number of epochs' }),
+    'no-truncate': flags.boolean({
+      required: false,
+      description: "Don't truncate fields to fit line",
+    }),
```
nategraf: If you just name this flag `truncate` then `--no-truncate` will be added automatically. It would also be really nice if you could make this a global option so it can be used anywhere we have a table.
https://oclif.io/docs/flags

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,173 @@
...
+import { Flags } from '../../utils/command'
+
+export default class Show extends BaseCommand {
+  static description =
+    'Show rewards information about a voter, registered Validator, or Validator Group'
+
+  static flags = {
+    ...BaseCommand.flags,
+    address: Flags.address({ required: false, description: 'Address to filter' }),
+    epochs: flags.integer({ required: false, description: 'Number of epochs' }),
```
nategraf: Please elaborate here. Is this the number of epochs back past the current epoch to look? What is the default?

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,173 @@
...
+        validators.getUniqueValidators(
+          validatorRewardsEvents,
+          (x: any) => x.returnValues.validator.toLowerCase(),
+          blockNumber
+        ),
+      epochSize,
+      res.flags.epochs
+    )
+
+    // For correctness use the Validator Group name at each epoch?
```
nategraf: Is this still a question?

nategraf (dismissed):
General note before I did into this again. We really want to keep type safety, so `any` should be used sparingly. `contractkit` especially has gone to great lengths for type safety. A lot of the helper functions you introduce here break type safety by using `any` and you'll need to refactor it to to avoid that. Honestly I find too many helpers makes code hard to read, and I think many of these could be cut because they are only needed in one place anyway.

jfoutts-celo:
Ok. I've removed the helpers and any "any:".

packages/cli/src/base.ts
```diff
@@ -12,6 +12,13 @@ export abstract class LocalCommand extends Command {
   static flags = {
     logLevel: flags.string({ char: 'l', hidden: true }),
     help: flags.help({ char: 'h', hidden: true }),
+    truncate: flags.boolean({
+      default: true,
+      required: false,
```
nategraf: No need to explicitly set false
```suggestion
```

packages/cli/src/commands/election/list.ts
```diff
@@ -12,15 +12,20 @@ export default class List extends BaseCommand {
...
-    })
+    cli.table(
+      groupVotes,
+      {
+        address: {},
+        votes: { get: (g) => g.votes.toFixed() },
+        capacity: { get: (g) => g.capacity.toFixed() },
+        eligible: {},
+      },
+      { 'no-truncate': !res.flags.truncate }
```
nategraf, line 29: Nice 🙌

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,259 @@
...
+  static args = []
+
+  static examples = ['show --address 0x5409ed021d9299bf6814279a6a1411a7e866a631']
+
+  async run() {
+    const res = this.parse(Show)
+    const election = await this.kit.contracts.getElection()
+    const validators = await this.kit.contracts.getValidators()
+    const electionContract = await this.kit._web3Contracts.getElection()
+    const validatorsContract = await this.kit._web3Contracts.getValidators()
```
nategraf: Avoid calling `_web3Contracts` when possible (It's supposed to be internal and no part of the API)
Instead I'd recommend adding `events` and `getPastEvents` to the `BaseWrapper` following the typings of `web3` https://github.com/ethereum/web3.js/blob/c0e727eac7c16826c39fd43a03a4930b5ff6fff8/packages/web3/types.d.ts#L328-L345

However, this may need the blessing of @jmrossy or someone else more involved with `contractkit`

packages/contractkit/src/wrappers/Validators.ts
```diff
@@ -190,10 +200,24 @@ export class ValidatorsWrapper extends BaseWrapper<Validators> {
...
+    const res = blockNumber
+      ? await this.contract.methods
+          .getValidator(address)
+          // @ts-ignore: Expected 0-1 arguments, but got 2
+          .call({}, blockNumber)
+      : await this.contract.methods.getValidator(address).call()
+
+    let name: string
+    if (blockNumber) {
+      const accounts = await this.kit._web3Contracts.getAccounts()
```
nategraf: I would instead recommend adding `blockNumber` as a parameter of the `getName` function on the Accounts wrapper.

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,259 @@
...
+export default class Show extends BaseCommand {
+  static description =
+    'Show rewards information about a voter, registered Validator, or Validator Group'
+
+  static flags = {
+    ...BaseCommand.flags,
+    address: Flags.address({ required: false, description: 'Address to filter' }),
+    epochs: flags.integer({
+      default: 1,
+      required: false,
```
nategraf: No need for `required: false`
```suggestion
```

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,259 @@
...
+import { newCheckBuilder } from '../../utils/checks'
+import { Flags } from '../../utils/command'
+
+export default class Show extends BaseCommand {
+  static description =
+    'Show rewards information about a voter, registered Validator, or Validator Group'
+
+  static flags = {
+    ...BaseCommand.flags,
+    address: Flags.address({ required: false, description: 'Address to filter' }),
```
nategraf: No need for `required: false`
```suggestion
    address: Flags.address({ description: 'Address to filter' }),
```

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,259 @@
...
+    const electionContract = await this.kit._web3Contracts.getElection()
+    const validatorsContract = await this.kit._web3Contracts.getValidators()
+    const epochSize = await validators.getEpochSize()
+    const currentBlock = await this.web3.eth.getBlockNumber()
+    const lastEpochBlock = Math.floor(currentBlock / epochSize) * epochSize
+    const fromBlock: number = lastEpochBlock - epochSize * ((res.flags.epochs || 1) - 1)
+    const addressVotes: { [key: number]: { [key: string]: BigNumber } } = {}
+    const validatorDetails: { [key: number]: { [key: string]: Validator } } = {}
+    const validatorGroupDetails: { [key: number]: { [key: string]: ValidatorGroup } } = {}
+    const lowerAddress = res.flags.address ? res.flags.address.toLowerCase() : res.flags.address
```
nategraf: I see this is for comparisons. Instead use `eqAddress` from `@celo/utils`
https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/address.ts#L11

nategraf (commented):
Some more comments, thanks for making that big refactor!

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,261 @@
...
+          const group: string = vote.group.toLowerCase()
+          addressVotes[group] = (addressVotes[group] || new BigNumber(0)).plus(vote.active)
+          totalGroupVotes[group] = (totalGroupVotes[group] || new BigNumber(0)).plus(
+            await election.getTotalVotesForGroup(group, blockNumber)
+          )
+        }
+      }
+
+      // voterRewardsEvent applies to address when voterRewardsEvent.group in addressVotes.
+      const epochVoterRewardsEvents = await election.getPastEvents(
```
mcortesi: i would hide `getPastEvents()` here, and have a methods like `getPastVotesRewards()` or something.

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,261 @@
...
+      )
+
+      voterRewards = voterRewards.concat(
+        address
+          ? epochVoterRewards.filter((e: VoterReward) => e.group.toLowerCase() in addressVotes)
+          : epochVoterRewards
+      )
+
+      // validatorRewardEvent applies to address when validatorRewardEvent.validator is address.
+      const epochValidatorRewardsEvents = await validators.getPastEvents(
```
mcortesi: same

mcortesi: The reason is that the end user doesn't really know a lot fo types of events, and event parameters. We need to digest that for the end user

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,261 @@
...
+            validatorPayment: e.returnValues.validatorPayment,
+            validatorScore: epochValidatorDetails[index].score,
+            group: e.returnValues.group,
+            blockNumber: e.blockNumber,
+          })
+        )
+      )
+
+      // validatorRewardEvent applies to address when validatorRewardEvent.group is address.
+      const epochValidatorGroupDetails: ValidatorGroup[] = await Promise.all(
```
mcortesi: use `concurrentMap()` since you might be making hundreds of concurrente request at the same time here

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,261 @@
...
+      // validatorRewardEvent applies to address when validatorRewardEvent.validator is address.
+      const epochValidatorRewardsEvents = await validators.getPastEvents(
+        'ValidatorEpochPaymentDistributed',
+        {
+          fromBlock: blockNumber,
+          toBlock: blockNumber,
+        }
+      )
+
+      const epochValidatorDetails: Validator[] = await Promise.all(
```
mcortesi: use on `concurrentMap`

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,261 @@
...
+      // voterRewardsEvent applies to address when voterRewardsEvent.group in addressVotes.
+      const epochVoterRewardsEvents = await election.getPastEvents(
+        'EpochRewardsDistributedToVoters',
+        {
+          fromBlock: blockNumber,
+          toBlock: blockNumber,
+        }
+      )
+
+      const voterRewardsValidatorGroupDetails: ValidatorGroup[] = await Promise.all(
```
mcortesi: concurrentMap

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,261 @@
...
+    const currentBlock = await this.web3.eth.getBlockNumber()
+    const lastEpochBlock = Math.floor(currentBlock / epochSize) * epochSize
+    const fromBlock: number = lastEpochBlock - epochSize * ((res.flags.epochs || 1) - 1)
+    const address = res.flags.address || ''
+    let voterRewards: VoterReward[] = []
+    let validatorRewards: ValidatorReward[] = []
+    let validatorGroupRewards: ValidatorGroupReward[] = []
+
+    if (address) {
+      await newCheckBuilder(this)
```
mcortesi: is this the only check we can do?

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,261 @@
...
+          value: e.returnValues.value,
+          blockNumber: e.blockNumber,
+          addressVotes: addressVotes[e.returnValues.group],
+          totalGroupVotes: totalGroupVotes[e.returnValues.group],
+        })
+      )
+
+      voterRewards = voterRewards.concat(
+        address
+          ? epochVoterRewards.filter((e: VoterReward) => e.group.toLowerCase() in addressVotes)
```
mcortesi: there's an eqAddress somewhere

mcortesi: no need to lowercase with that

packages/contractkit/src/wrappers/Accounts.ts
```diff
@@ -237,8 +237,16 @@ export class AccountsWrapper extends BaseWrapper<Accounts> {
   /**
    * Returns the set name for the account
    * @param account Account
+   * @param blockNumber Height of result, defaults to tip.
    */
-  getName = proxyCall(this.contract.methods.getName)
+  async getName(account: Address, blockNumber?: number): Promise<string> {
+    return blockNumber
```
mcortesi: you probably don't need the conditional here.. since you would be sending `undefined` when not set...
and code would be cleaner

packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -20,6 +20,22 @@ export abstract class BaseWrapper<T extends Contract> {
...
+  /** Contract getPastEvents */
+  getPastEvents(
+    event: string,
+    options?: {
+      filter?: object
+      fromBlock?: BlockType
+      toBlock?: BlockType
+      topics?: string[]
+    },
+    cb?: Callback<EventLog[]>
```
mcortesi: don't expose the callback variant.. only return promises

packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -20,6 +20,22 @@ export abstract class BaseWrapper<T extends Contract> {
     // TODO fix typings
     return (this.contract as any)._address
   }
+
+  /** Contract getPastEvents */
+  getPastEvents(
```
mcortesi: not sure about making it private.. maybe protected.

mcortesi: Also.... you might want to warn people about doing a query too far in the past or too big. Geth won't like...

cc @asaj @kevjue

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -110,23 +113,43 @@ export class ElectionWrapper extends BaseWrapper<Election> {
...
-      .call()
-    const active = await this.contract.methods
-      .getActiveVotesForGroupByAccount(group, account)
-      .call()
+  async getVotesForGroupByAccount(
+    account: Address,
+    group: Address,
+    blockNumber?: number
+  ): Promise<GroupVote> {
+    const pending = blockNumber
```
mcortesi: same as before... you rpbably don't need the conditonal.. 
for this and all the others

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -110,23 +113,43 @@ export class ElectionWrapper extends BaseWrapper<Election> {
...
-    const votes = await Promise.all(groups.map((g) => this.getVotesForGroupByAccount(account, g)))
+  async getVoter(account: Address, blockNumber?: number): Promise<Voter> {
+    const groups: Address[] = blockNumber
+      ? await this.contract.methods
+          .getGroupsVotedForByAccount(account)
+          // @ts-ignore: Expected 0-1 arguments, but got 2
+          .call({}, blockNumber)
+      : await this.contract.methods.getGroupsVotedForByAccount(account).call()
+
+    const votes = await Promise.all(
```
mcortesi: concurrent map

mcortesi (changes requested):

packages/cli/src/commands/rewards/show.ts
jfoutts-celo: Ok, I changed the comparisons to use eqAddress().  This one is checking membership in a dictionary keyed by group, though.

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    if (res.flags.validator) {
+      if (address) throw Error('Must select --voter, --validator, or --group')
+      address = res.flags.validator
+      await newCheckBuilder(this)
+        .isValidator(address)
+        .runChecks()
+    }
+
+    if (res.flags.group) {
```
mcortesi, line 55: the code seems to imply that you could enter on more than one `if`, thus run more than one `runChecks()`. Or maybe they are exclusive, idk.

But if that's the case, i think it's better to runCheck() only once, thus, you can create the checksBuilder and use the `if`s to add checks conditionally, and then do only one `runChecks()`

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,293 @@
...
+            vote.active
+          )
+          activeGroupVotes[group] = (activeGroupVotes[group] || new BigNumber(0)).plus(
+            await election.getTotalVotesForGroup(group, blockNumber)
+          )
+        }
+      }
+
+      // voterRewardsEvent applies to address when voterRewardsEvent.group in activeAddressVotes.
+      const epochVoterRewardsEvents = await election.getPastVoterRewards({
```
mcortesi: now that you have a method for it, it makes more sense for the signature to be `getPastVoterRewards(from, to)` instead of an object

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,293 @@
...
+      voterRewards = voterRewards.concat(
+        address
+          ? epochVoterRewards.filter(
+              (e: VoterReward) => e.group.toLowerCase() in activeAddressVotes
+            )
+          : epochVoterRewards
+      )
+
+      // validatorRewardEvent applies to address when validatorRewardEvent.validator is address.
+      const epochValidatorRewardsEvents = await validators.getPastValidatorRewards({
```
mcortesi: same

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,293 @@
...
+        }
+      }
+
+      // voterRewardsEvent applies to address when voterRewardsEvent.group in activeAddressVotes.
+      const epochVoterRewardsEvents = await election.getPastVoterRewards({
+        fromBlock: blockNumber,
+        toBlock: blockNumber,
+      })
+
+      const voterRewardsValidatorGroupDetails: ValidatorGroup[] = await concurrentMap(
```
mcortesi: I think that instead of returning the generic `EventLog` you can return something called `VoterRewards` which contains the information already digested

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,293 @@
...
+        toBlock: blockNumber,
+      })
+
+      const epochValidatorDetails: Validator[] = await concurrentMap(
+        10,
+        epochValidatorRewardsEvents,
+        (e: EventLog) => validators.getValidator(e.returnValues.validator, blockNumber)
+      )
+
+      const epochValidatorRewards = epochValidatorRewardsEvents.map(
```
mcortesi: same....
the end user shouldn't know about the `returnValues`, nor that it is an event

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,293 @@
...
+      )
+
+      // validatorRewardEvent applies to address when validatorRewardEvent.group is address.
+      const epochValidatorGroupDetails: ValidatorGroup[] = await concurrentMap(
+        10,
+        epochValidatorRewardsEvents,
+        (e: EventLog) => validators.getValidatorGroup(e.returnValues.group)
+      )
+
+      const epochValidatorGroupRewards = epochValidatorRewardsEvents.map(
```
mcortesi: same

packages/cli/src/commands/rewards/show.ts
```diff
@@ -0,0 +1,293 @@
...
+      default: 1,
+      description: 'Show results for the last N epochs',
+    }),
+  }
+
+  static args = []
+
+  static examples = ['show --address 0x5409ed021d9299bf6814279a6a1411a7e866a631']
+
+  async run() {
```
mcortesi, line 39: In general, i believe this command does too much by itself.. and most of the things here should live in contractKit.

1) **Notion about epoch.** It would be nice to ask ContractKit `getCurrentEpoch` `epochToBlockNumber` etc. Instead of doing the computation by ourselves. "Epoch" is a notion that should be included in contractKit

2) **Notion about rewards** 

Rewards are by epoch, not by block. This command has encoded the notion that to get the rewards you need to query the logs of the last block of the epoch. That notion belongs to contractKit. So you could say `Validator.getValidatorRewards(epochNumber)` and the same for all the others.

In fact, we could have a `Accounts.getRewards(epoch)` or send a epoch range.

Also there's a small gain that we can achieve by saying `give the rewards for this epoch range`. You could cache the validator & validator group details, since they would probably be the same, and you are querying for them on every block.

But as a rule of thumb, if we are doing too much computation here, we should ask ourselves, if this is something that should live in contractKit, and thus be available for any developer as a simple api.

mcortesi (changes requested):

packages/cli/src/commands/rewards/show.ts
jfoutts-celo: Ok, I've added `getCurrentEpoch`, `epochToBlockNumber`, and moved most the reward digesting code to contractKit.  

I'm finding the code loses clarity when trying to pipeline multiple epochs, or avoid some redundant queries.  And we actually want the Validator score from each epoch.

So for those reasons I've made the rewards retrieval API single epoch, and haven't optimized redundant validator group queries.

packages/contractkit/src/kit.ts
```diff
@@ -239,4 +239,17 @@ export class ContractKit {
       ...tx,
     }
   }
+
+  async getCurrentEpoch(): Promise<number> {
+    const validators = await this.contracts.getValidators()
+    const epochSize = await validators.getEpochSize()
+    const currentBlock = await this.web3.eth.getBlockNumber()
+    return Math.floor(currentBlock / epochSize)
```
mcortesi: is this correct? I recall the conversation where you stated that it should be:
`epochNumber(blockNumber) => blockNumber / epochSize + (blockNumber % epochSize != 0)`

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -330,4 +372,72 @@ export class ElectionWrapper extends BaseWrapper<Election> {
...
+    const blockNumber = await this.kit.epochToBlockNumber(epochNumber)
+    const events = await this.getPastEvents('EpochRewardsDistributedToVoters', {
+      fromBlock: blockNumber,
+      toBlock: blockNumber,
+    })
+    const validators = await this.kit.contracts.getValidators()
+    const validatorGroup: ValidatorGroup[] = await concurrentMap(10, events, (e: EventLog) =>
+      validators.getValidatorGroup(e.returnValues.group, true, blockNumber)
+    )
+    const signers = await this.getCurrentValidatorSigners()
```
mcortesi: shouldn't be the signers for that epoch?

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -330,4 +372,72 @@ export class ElectionWrapper extends BaseWrapper<Election> {
       greater: newIdx === currentVotes.length - 1 ? NULL_ADDRESS : currentVotes[newIdx + 1].address,
     }
   }
+
+  /**
+   * Retrieves GroupVoterRewards at epochNumber.
+   * @param epochNumber The epoch to retrieve GroupVoterRewards at.
+   */
+  async getGroupVoterRewards(epochNumber: number): Promise<GroupVoterReward[]> {
```
mcortesi, line 395: It seems to me that `validators` is not really something that belongs to `GroupVoterReward`. Seems like we are doing a lot of calls for something that not everybody would want.
I mean, we asked for the rewards for each group. Having the member of each group is not something we asked for

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -330,4 +372,72 @@ export class ElectionWrapper extends BaseWrapper<Election> {
...
+
+  /**
+   * Retrieves VoterRewards for address at epochNumber.
+   * @param address The address to retrieve VoterRewards for.
+   * @param epochNumber The epoch to retrieve VoterRewards at.
+   */
+  async getVoterRewards(address: Address, epochNumber: number): Promise<VoterReward[]> {
+    const blockNumber = await this.kit.epochToBlockNumber(epochNumber)
+    const voter = await this.getVoter(address, blockNumber)
+    const activeVoterVotes: { [key: string]: BigNumber } = {}
```
mcortesi: you can use `Record<string, BigNumber`. It's the same, but more succint

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -330,4 +372,72 @@ export class ElectionWrapper extends BaseWrapper<Election> {
...
+    const activeGroupVotesQuery: { [key: string]: Promise<BigNumber> } = {}
```
mcortesi: same

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -330,4 +372,72 @@ export class ElectionWrapper extends BaseWrapper<Election> {
...
+    for (const vote of voter.votes) {
+      const group: string = vote.group.toLowerCase()
+      activeVoterVotes[group] = vote.active
+      activeGroupVotesQuery[group] = this.getTotalVotesForGroup(group, blockNumber)
+    }
+    const activeGroupVotes: { [key: string]: BigNumber } = await concurrentValuesMap(
```
mcortesi: same

nategraf:
I'm going to be gone on vacation. I have no more concerns. Thanks for all the iterations on this! 👍

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -330,4 +372,72 @@ export class ElectionWrapper extends BaseWrapper<Election> {
...
+    )
+    return voterRewards.map(
+      (e: GroupVoterReward): VoterReward => ({
+        address,
+        addressPayment: e.groupVoterPayment.times(
+          activeVoterVotes[e.group.address].dividedBy(activeGroupVotes[e.group.address])
+        ),
+        group: e.group,
+        epochNumber: e.epochNumber,
+        validators: e.validators,
```
mcortesi: same here. It's not clear why return the validators

packages/contractkit/src/kit.ts
```diff
@@ -239,4 +239,17 @@ export class ContractKit {
...
+  }
+
+  async epochToBlockNumber(epochNumber: number): Promise<number> {
```
mcortesi: what about renaming this to `getLastBlockNumberForEpoch` or something similar... since it's not clear which block number for an epoch this fn will return

packages/contractkit/src/wrappers/Validators.ts
```diff
@@ -190,10 +210,12 @@ export class ValidatorsWrapper extends BaseWrapper<Validators> {
...
 
   /** Get Validator information */
-  async getValidator(address: Address): Promise<Validator> {
-    const res = await this.contract.methods.getValidator(address).call()
+  async getValidator(address: Address, blockNumber?: number): Promise<Validator> {
+    // @ts-ignore: Expected 0-1 arguments, but got 2
+    const res = await this.contract.methods.getValidator(address).call({}, blockNumber)
     const accounts = await this.kit.contracts.getAccounts()
-    const name = (await accounts.getName(address)) || ''
+    const name = await accounts.getName(address, blockNumber)
```
mcortesi: why remove the `|| ''` part?

packages/contractkit/src/wrappers/Validators.ts
```diff
@@ -255,19 +279,20 @@ export class ValidatorsWrapper extends BaseWrapper<Validators> {
...
     this.contract.methods.getRegisteredValidatorGroups
   )
 
   /** Get list of registered validators */
-  async getRegisteredValidators(): Promise<Validator[]> {
-    const vgAddresses = await this.getRegisteredValidatorsAddresses()
-    return Promise.all(vgAddresses.map((addr) => this.getValidator(addr)))
+  async getRegisteredValidators(blockNumber?: number): Promise<Validator[]> {
+    const vgAddresses = await this.getRegisteredValidatorsAddresses(blockNumber)
+    return Promise.all(vgAddresses.map((addr) => this.getValidator(addr, blockNumber)))
```
mcortesi: use concurrentMap

packages/utils/src/async.ts
```diff
@@ -75,3 +75,25 @@ export async function concurrentMap<A, B>(
...
 }
+
+/**
+ * Map an async function over the values in Object x with a given concurrency level
+ *
+ * @param concurrency number of `mapFn` concurrent executions
+ * @param x associative array of values
+ * @param mapFn mapping function
+ */
+export async function concurrentValuesMap(
```
mcortesi: you can probably type this as:
```
function concurrentValuesMap<IN extends any, OUT extends any>(
  concurrency:number, 
  x: Record<string, IN>
   mapFn: (val: IN, key: string) => Promise<OUT>): Promise<Record<string,OUT>
```

also it looks like `mapFn` should receive the `key` not the index in the array

jfoutts-celo: Great suggestions!

packages/contractkit/src/kit.ts
jfoutts-celo: Updated to call smart contract getEpochNumber().  Note that definition ended up being off by one.  I'm going off the definition in master UsingPrecompiles.

mcortesi (approved):
Great Work! Thanks for all the changes!

PR #2164, opened 2019-12-10
Bump @celo/utils to 0.1.5-dev

nategraf:
### Description

Move `@celo/utils` to an unpublished version


packages/notification-service/package.json
```diff
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@celo/contractkit": "0.1.5",
-    "@celo/utils": "^0.1.0",
+    "@celo/utils": "0.1.5-dev",
```
jmrossy: Small gotcha here. The services that get deployed to App Engine can only depend on published versions of @celo packages, otherwise we're blocked from deploying. Can you use the latest published version here (or the one you're going to deploy)?

packages/verification-pool-api/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "@celo/contractkit": "^0.1.5",
-    "@celo/utils": "^0.1.0",
+    "@celo/utils": "0.1.5-dev",
```
jmrossy: Ditto as notification service

nategraf:
@jmrossy @mcortesi I've updated the `README-dev.md` file to include the publishing instructions as I understand them.

packages/transaction-metrics-exporter/package.json
```diff
@@ -8,7 +8,7 @@
   "private": true,
   "dependencies": {
     "@celo/contractkit": "0.2.4",
-    "@celo/utils": "^0.1.0",
+    "@celo/utils": "0.1.5-dev",
```
mcortesi, line 11: This package is built as a docker container... not sure if that docker has a copy of git's version of utils. If not it should use a published version. Probably the best anyway

yarn.lock
```diff
@@ -2300,6 +2300,28 @@
     lodash "^4.17.14"
     web3-utils "1.0.0-beta.37"
 
+"@celo/utils@0.1.4", "@celo/utils@^0.1.0":
```
mcortesi: it seems there's somebody still using 0.1.0. I wonder who he is

mcortesi (commented):
Thanks for doing this, check the comments

packages/transaction-metrics-exporter/package.json
nategraf: It builds from source, so this is safe https://github.com/celo-org/celo-monorepo/blob/cbdd336341a5d63bd06d232e0056f71320b34252/dockerfiles/transaction-metrics-exporter/Dockerfile#L27

yarn.lock
nategraf: I was looking into this, and I seems like old versions of all of our packages are in here. For example `contractkit` still has a reference to its v0.1.5 in `yarn.lock` I can't find any  reference to `@celo/utils@0.1.0` anywhere else.
https://github.com/celo-org/celo-monorepo/blob/cbdd336341a5d63bd06d232e0056f71320b34252/yarn.lock#L2234-L2251

jmrossy:
Assigning back for now because tests are failing

.circleci/config.yml
```diff
@@ -406,15 +406,6 @@ jobs:
           name: Run Tests
           command: yarn --cwd=packages/contractkit test
 
-      - run:
```
jmrossy, line 409: Why remove these?

README-dev.md
```diff
@@ -9,7 +9,7 @@ Many packages depend on other packages within the monorepo. When this happens, f
 3.  To differentiate published vs unpublished version. Master version (in package.json) must end with suffix `-dev` and should not be published.
 4.  If a developer want to publish a version; then after publishing it needs to set master version to next `-dev` version and change all package.json that require on it.
 
-To check which pakages need amending, you can run (in the root pkg):
+To check which packages need amending, you can run (in the root pkg):
```
jmrossy, line 12: Thanks for updating this file :)

jmrossy (commented):
These changes all LGTM! Just curious about the circle config change.

.circleci/config.yml
nategraf: This came out of talking to Mariano. This test fails when `contracktkit` depends on unpublished packages, namely `@celo/utils` and so we removed this check under because it will always fail for any PR with breaking changes to `@celo/utils`. Do you think that's fair?

nategraf: Oh and it would prevent us from moving to the scheme we are suggesting of pointing to an unpublished version number, so if we want to keep this check we will need to abandon that scheme.

README-dev.md
nategraf: Of course! I hope it's helpful

.circleci/config.yml
mcortesi: yes. We can really install the commited version of the package, since their dependencies are also not published... this type of test don't work anymore for us

mcortesi (approved):

Issue #1756, opened 2019-11-18
[Wallet] Users SBAT see local currency in gold exchange


annakaz:
### Expected Behavior

Users see the price of Celo gold displayed in their local currency throughout the exchange flow. See <img width="2017" alt="66881921-5e169700-ef7d-11e9-821c-b06cfcf37084" src="https://user-images.githubusercontent.com/16907027/68888198-573f8900-06cf-11ea-93dc-02f10a03b790.png">

[Figma](https://www.figma.com/file/Iqu4mAgePHVgKhvKTrGrUP/Gold-Approaches?node-id=10%3A0)

Builds on top of https://github.com/celo-org/celo-monorepo/issues/1484

### Current Behavior

Celo Dollars shown regardless of selected local currency.

annakaz:
Split apart from https://github.com/celo-org/celo-monorepo/issues/1728, as this part will build on top of the new local currency approach (https://github.com/celo-org/celo-monorepo/issues/1484)

Issue #1705, opened 2019-11-13
[Wallet] Users SBAT review exchange without scrolling


annakaz:
### Expected Behavior

The review exchange screen can be displayed in a single screen, and matches the formatting with boxes that have all four sides elsewhere in the app.

### Current Behavior

The review exchange screen is in a narrow box, so users have to scroll down to view the whole exchange. When they have fully scrolled down, the formatting is not intuitive as there is no line closing the bottom of the box. 

View on a pixel 2 (relatively large screen):
![Screenshot_20191113-150626](https://user-images.githubusercontent.com/16907027/68812326-95816d80-0627-11ea-938d-f4ba4fa1c01b.png)

Done scrolling: 
![Screenshot_20191113-150619](https://user-images.githubusercontent.com/16907027/68812318-91ede680-0627-11ea-963a-34888868797f.png)

Issue #1484, opened 2019-10-25
Implement more wholistic solution to managing local currencies


jmrossy:
Currently, we are converting currencies in many places. This leads to corners in the app where we don't properly convert values and also code duplication.

Instead, we should have a single money component which all displayed money values, and that handles currency conversions properly.

Related (should be tackled while doing this): #962

jeanregisser:
Done in #3032

PR #2167, opened 2019-12-10
[Wallet] New exchange trade screen and fix ts build in celotool

### LINKED ISSUE #1396: [Android] “0.00” is not shown on “Exchange” screen when user open the Exchange screen by tapping on gold activity from Gold/Wallet tab  if security fee amount 0.
**App version:** zero-sync build: Staging build v1.5.0 (101548577)
**Repro on:** Sony Xperia C4 (5.0), Samsung J7 Prime (6.0.1), Samsung Galaxy A9 Pro (8.0)
**Testing Account:**  680 208 0265/Backup Kay
 
**Repro Steps:** 
1) Launch the app. 
2) Tap on “Gold” tab. 
3) Tap on any gold activity. 
4) Observe. 

**Investigation:** Issue does not reproduce on Review Exchange screen while user proceeding for exchange.  
**Current Behavior:** “0.00” is not shown on “Exchange” screen
**Expected Behavior:** “0.00” should be shown on “Exchange” screen
**Attachment:**Android_SecurityFeeAmountIssue .png 
Android_SecurityFeeAmountIssue

### LINKED ISSUE #1438: [Android] “cUSD” and “cGLD” amount fields are not shown align properly on “Exchange” screen from Gold tab.
**Frequency:** 100%
**App version:** zero-sync build: Staging build v1.5.0 (101548577)
**Repro on:** Samsung Galaxy Grand Prime (5.1.1), Samsung Galaxy s5 (6.0.1) , Samsung Galaxy j7 prime (6.0.1)
**Does not Repro** Moto E (4.4..4), Samsung Galaxy note 9 (9.0)
 
**Repro Steps:** 
1) Launch the app. 
2) Tap on the “Gold” tab. 
3) Tap on the Exchange button. 
4) Observe. 

**Current Behavior:** “cUSD” and “cGLD” amount fields are not shown align properly
**Expected Behavior:** “cUSD” and “cGLD” amount of fields should be aligned properly. 
**Attachment:**Android_ExchangeScreen.png
Android_ExchangeScreen

### LINKED ISSUE #1704: [Wallet] Users SBAT see Security Fee in exchange confirmation card
### Expected Behavior

Users see the fee listed alongside "Security Fee" on a screen that is formatted consistently with other confirmation screens. 

### Current Behavior

There is an empty space listed alongside "Security Fee", and the confirmation screen is missing a line in the bottom

Screenshot_20191113-145710

### LINKED ISSUE #1728: [Wallet] Users SBAT see updated gold exchange screen
### Expected Behavior

Users commit to direction(buy/sell gold) before entering the flow. UI matches mocks below

66881921-5e169700-ef7d-11e9-821c-b06cfcf37084

[Figma](https://www.figma.com/file/Iqu4mAgePHVgKhvKTrGrUP/Gold-Approaches?node-id=10%3A0)

### Current Behavior

Current exchange screen looks like this:

annakaz:
### Description

Add new Exchange trade screen

Previously, users could toggle between Buying/Selling gold or dollars and entering the value in whichever currency they were going to trade in. The new UI trade screen defines a direction (Buy or Sell) before users input the amount they'd like to trade. This allows us to display a more accurate exchange rate, as we know which direction to show the exchange rate in -- there is small buy/sell spread in the Exchange rate (ex: to sell, receive $9.9499, to buy, pay $10.0502). 

This rate also varies (an even smaller amount than with direction) based on the size of the exchange (ex: selling 1 gold gives rate of $0.1005, 1000 gold a rate of $0.1006), so the 'worst case' rate is shown by estimating the rate at the maximum amount of gold/dollars a user could buy or sell. 

### Tested

Sold and bought gold on `alfajoresstaging`
Updated jest tests

### Other changes

Fix empty fee in confirmation screen
Added missingFullName error text

### Related issues

- Fixes #1728 
- Fixes #1396 (empty fee)
- Fixes #1438 , #1705 (bugs in old trade UI)

Note that local currency-related work is a separate issue (#1756 ) as it will build on top of https://github.com/celo-org/celo-monorepo/issues/1484 

### Backwards compatibility

Yes


packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,20 +1,18 @@
...
 import SafeAreaView from 'react-native-safe-area-view'
 import { connect } from 'react-redux'
-import CeloAnalytics from 'src/analytics/CeloAnalytics'
-import { CustomEventNames } from 'src/analytics/constants'
 import componentWithAnalytics from 'src/analytics/wrapper'
 import AccountOverview from 'src/components/AccountOverview'
 import { fetchExchangeRate } from 'src/exchange/actions'
 import Activity from 'src/exchange/Activity'
 import ExchangeRate from 'src/exchange/ExchangeRate'
-import { CURRENCY_ENUM as Token } from 'src/geth/consts'
```
annakaz: We now use `CURRENCY_ENUM` everywhere in the app.. no more ` CURRENCY_ENUM as Token` silliness

annakaz:
Screenshots: 
Screenshot_20191210-154739
Screenshot_20191210-154746
Screenshot_20191210-154752
Screenshot_20191210-154755
Screenshot_20191210-154809
Screenshot_20191210-154830

annakaz:
Note that build is failing on master so tests aren't running. `yarn run test` in mobile passes locally.

packages/mobile/src/exchange/ExchangeHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return {
+    ...headerWithCancelButton,
+    headerTitle: (
+      <View style={styles.headerTextContainer}>
+        <Text style={fontStyles.headerBoldTitle}>{title}</Text>
+        <View>
+          <Text style={fontStyles.subSmall}>
+            {getMoneyDisplayValue(makerTokenBalance, makerToken, true) +
+              ' ' +
+              i18n.t(`${Namespaces.exchangeFlow9}:lowerCaseAvailable`)}
```
jeanregisser: For better translation we should avoid concatenating strings this way and use interpolation:

```ts
t('moneyAvailable', { moneyAmount: getMoneyDisplayValue(makerTokenBalance, makerToken, true)})
```

And have the following in the locale json file:

```json
moneyAvailable: "{{moneyAmount}} available"
```

packages/mobile/src/exchange/ExchangeHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { StyleSheet, Text, View } from 'react-native'
+import i18n, { Namespaces } from 'src/i18n'
+import { headerWithCancelButton } from 'src/navigator/Headers'
+import { getMoneyDisplayValue } from 'src/utils/formatting'
+
+export const ExchangeHeader = (makerToken: CURRENCY_ENUM, makerTokenBalance: string) => {
+  const title =
+    makerToken === CURRENCY_ENUM.DOLLAR
+      ? i18n.t(`${Namespaces.exchangeFlow9}:buyGold`)
+      : i18n.t(`${Namespaces.exchangeFlow9}:sellGold`)
```
jeanregisser: We should inject `t` using `withNamespace` instead of directly accessing i18n, main reason behind is the new translations get re-rendered when switching language.

Note that once we can upgrade to the latest i18next version we'll be able to use a hook instead which is simpler IMHO.

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -136,9 +232,33 @@ class ExchangeReview extends React.Component<Props> {
...
-    backgroundColor: colors.background,
-    paddingTop: 20,
+    flexDirection: 'column',
+    justifyContent: 'space-between',
+  },
+  paddedContainer: {
+    paddingHorizontal: 16,
+  },
+  column: {
+    flexDirection: 'column',
```
jeanregisser: nit: this is the default direction

packages/mobile/src/exchange/ExchangeTradeScreen.test.tsx
```diff
@@ -55,20 +66,48 @@ describe(ExchangeTradeScreen, () => {
...
+      )
+      component.root.instance.onChangeExchangeAmount('10')
+      expect(mockShowError).toBeCalledWith(ErrorMessages.NSF_DOLLARS) // Can't afford 10 gold
+      component.root.instance.switchInputToken()
+      expect(mockhideAlert).toBeCalled() // Can afford 10 USD worth of gold
+      component.root.instance.onChangeExchangeAmount('20')
+      expect(mockShowError).toBeCalledWith(ErrorMessages.NSF_DOLLARS) // Can't afford 20 cUSD worth of gold
     })
 
     it.only('validates amount', () => {
```
jeanregisser: Looks like only this test case will run, do you know why we did this?

packages/mobile/src/icons/InfoIcon.tsx
```diff
@@ -3,17 +3,21 @@ import { Image } from 'react-native'
...
+    isGrey: false,
   }
 
   render() {
     return (
       <Image
-        source={require('src/images/info.png')}
+        source={
+          this.props.isGrey ? require('src/images/info-grey.png') : require('src/images/info.png')
+        }
```
jeanregisser: `<Image />` style supports `tintColor` which would be better here instead of adding a new image.
In that case it would be more flexible to replace `isGrey` by `tintColor` (either enum of string depending on the flexibility we want).

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -1,133 +1,229 @@
...
   fetchExchangeRate: typeof fetchExchangeRate
   exchangeTokens: typeof exchangeTokens
 }
 
+interface NavProps {
+  exchangeInput: {
+    makerToken: CURRENCY_ENUM
+    makerTokenBalance: string
+    inputToken: CURRENCY_ENUM
+    inputTokenCode: string
```
jeanregisser: Why do we need `inputTokenCode` given we already have `inputToken`?

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -1,133 +1,229 @@
...
-            takerAmount={takerAmount}
-            exchangeRate={rate}
-            fee={fee}
+        <View style={styles.paddedContainer}>
+          <DisconnectBanner />
+          <ScrollView>
+            <View style={styles.column}>
+              <View style={[styles.rowContainer, styles.amountRow]}>
+                <Text style={[fontStyles.body, styles.exchangeBodyText]}>
+                  {t('amount') + ` (${this.state.inputTokenCode})`}
```
jeanregisser: As commented above, we should use a single translation key and use interpolation.

Approximate implementation ([using nesting and interpolation](https://www.i18next.com/translation-function/nesting#passing-nesting-to-interpolated), but there are other ways):

```ts
t('exchangeAmount', {tokenName: `$t('global:${ key }')` })
```

Locale json
```json
exchangeAmount: "Amount ({{tokenName}})"
```

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -1,133 +1,229 @@
...
-            makerAmount={makerAmount}
-            takerAmount={takerAmount}
-            exchangeRate={rate}
-            fee={fee}
+        <View style={styles.paddedContainer}>
+          <DisconnectBanner />
+          <ScrollView>
+            <View style={styles.column}>
+              <View style={[styles.rowContainer, styles.amountRow]}>
+                <Text style={[fontStyles.body, styles.exchangeBodyText]}>
```
jeanregisser: Please pass a single style without an array when there's no dynamic part. It prevents re-rendering unnecessarily.

```ts
exchangeBody: {
  ...fontStyles.body,
  xxx
}
```

Same applies to the other styles used in this screen.

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -1,133 +1,229 @@
...
+                <Text style={[fontStyles.body, styles.currencyAmountText]}>
+                  {getMoneyDisplayValue(this.state.inputAmount, this.state.inputToken, true)}
+                </Text>
+              </View>
+              <View style={styles.line} />
+              <View style={[styles.rowContainer, styles.feeRowContainer]}>
+                <Text style={[fontStyles.body, styles.exchangeBodyText]}>
+                  {t('subtotal') +
+                    ' @ ' +
+                    getMoneyDisplayValue(exchangeRate, CURRENCY_ENUM.DOLLAR, true)}
```
jeanregisser: Same comment about using interpolation instead of concatenation which can't be translated correctly.

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -1,133 +1,229 @@
...
+              ' ' +
+              getMoneyDisplayValue(
+                this.getInputAmountInToken(CURRENCY_ENUM.GOLD),
+                CURRENCY_ENUM.GOLD,
+                false,
+                3
+              ) +
+              ' ' +
+              t('global:gold')
+            }
```
jeanregisser: Here as well, please use interpolation.

jeanregisser (changes requested):
Awesome work! 👍 🎉 

Comments are mostly about our coding style and a better way of doing localization.

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: Thanks for catching this

packages/mobile/src/icons/InfoIcon.tsx
annakaz: Good idea, fixed!

packages/mobile/src/exchange/ExchangeTradeScreen.test.tsx
annakaz: Removed, now we run them all

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: To avoid having to figure out the string again. In the future I would see us using CURRENCIES.code instead, but in the mean time I wanted to avoid copying over the `getInputTokenDisplayText` function

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: Fixed

packages/mobile/src/exchange/ExchangeHeader.tsx
annakaz: I moved this exchangeHeader to Headers.tsx, so we can't inject t using withNamespace. I use the global i18n.t following what we do for other header titles (see Analytics.tsx or Invite.tsx). Let me know if you think another approach would be better

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: Fixed!

annakaz:
Thanks Jean! Switched to using interpolation for translation and a single style where applicable. Also moved the exchangeHeader to Headers.tsx, so using the global i18n variable for translations there.

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -296,6 +296,7 @@ const AppStack = createStackNavigator(
     [Screens.PhotosEducation]: { screen: PhotosEducation },
     [Screens.GoldEducation]: { screen: GoldEducation },
     [Screens.FeeEducation]: { screen: FeeEducation },
+    [Screens.FeeExchangeEducation]: { screen: FeeExchangeEducation },
```
annakaz: Added so we navigate back to exchange confirmation card, rather than always going to exchange. 
Ex flow this fixes: Click exchange notification from home screen, then click FeeExchangeEducation, expect to return back to exchange confirmation card but instead go to the exchange

jeanregisser (approved):
Looking great! 💯

packages/mobile/locales/en-US/nuxVerification2.json
```diff
@@ -1,6 +1,7 @@
 {
   "getVerified": "Verify Phone Number",
   "notReadyForCode": "Not yet ready for code entry",
+  "missingFullName": "Missing Full Name",
```
jmrossy: This was moved to global.json, please remove :)

packages/mobile/locales/es-419/nuxVerification2.json
```diff
@@ -1,6 +1,7 @@
 {
   "getVerified": "Verificar el Teléfono",
   "notReadyForCode": "Aún no está listo para ingresar el código",
+  "missingFullName": "Nombre completo perdido",
```
jmrossy: Ditto

packages/mobile/src/components/FeeIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+  isExchange?: boolean
+  tintColor?: string
+}
+
+export default class FeeIcon extends React.Component<Props> {
+  navigateToEducate() {
+    navigate(Screens.FeeEducation, {})
+  }
+
+  navigateToExchangeEducate() {
```
jmrossy, line 19: this will work fine in this case but FYI, handlers should generally be lamdbas instead of class methods so that the `this` keyword gets bound properly

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
```diff
@@ -130,15 +129,15 @@ export function ExchangeConfirmationCard(props: Props) {
       <View style={styles.feeContainer}>
         <LineItemRow
           currencySymbol={takerToken}
-          amount={fee}
+          amount={fee || '0.00'}
```
jmrossy, line 132: would undefined be a better choice here to signify that the fee isn't available yet? Nbd either way

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -1,133 +1,224 @@
...
+    const { makerToken, makerTokenBalance } = navigation.getParam('exchangeInput')
+    return {
+      ...exchangeHeader(makerToken, makerTokenBalance),
+    }
+  }
+
+  state: State = {
+    makerToken: CURRENCY_ENUM.GOLD,
+    inputToken: CURRENCY_ENUM.GOLD,
+    inputTokenCode: this.props.t('global:gold'),
```
jmrossy: Is the tokenCode a value used for display purposes? If so, maybe rename to something more clear. `inputTokenDisplayName`?

jmrossy: Also, though it's probably fine in this case, using props in state initialization like this is generally considered an anti-pattern:
https://medium.com/@justintulk/react-anti-patterns-props-in-initial-state-28687846cc2e

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -1,133 +1,224 @@
...
+            <View style={styles.column}>
+              <View style={styles.amountRow}>
+                <Text style={styles.exchangeBodyText}>
+                  {t('exchangeAmount', { tokenName: t(`global:${this.state.inputTokenCode}`) })}
+                </Text>
+                <Text style={styles.currencyAmountText}>
+                  {getMoneyDisplayValue(this.state.inputAmount, this.state.inputToken, true)}
+                </Text>
+              </View>
+              <View style={styles.line} />
```
jmrossy: Can we use `HorizontalLine` from react-components here?

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -136,8 +227,38 @@ class ExchangeReview extends React.Component<Props> {
 const styles = StyleSheet.create({
   container: {
     flex: 1,
-    backgroundColor: colors.background,
-    paddingTop: 20,
+    flexDirection: 'column',
```
jmrossy: column is the default

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -136,8 +227,38 @@ class ExchangeReview extends React.Component<Props> {
...
+    justifyContent: 'space-between',
+  },
+  paddedContainer: {
+    paddingHorizontal: 16,
+  },
+  column: {
```
jmrossy: Nit: strange name for something that just justifies to flex-start

packages/mobile/src/navigator/Navigator.tsx
jmrossy: Maybe leave a comment explaining this?

packages/react-components/styles/fonts.tsx
```diff
@@ -167,6 +167,11 @@ export const fontStyles = StyleSheet.create({
     fontFamily: HindSilguri.Bold,
     color: colors.dark,
   },
+  headerBoldTitle: {
```
jmrossy: This is the same as `headerTitle` above it but with one font-size more. Can we merge these two? If not, consider renaming this.

jmrossy (changes requested):
Overall really nice work! Just some minor things to address

packages/mobile/src/components/FeeIcon.tsx
annakaz: Kk good to know!

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: Thanks for linking this- seems like it's okay here as we don't expect the translation function to change, but good to know for elsewhere

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: Good point. Renamed to flexStart

packages/mobile/src/exchange/ExchangeReview.tsx
annakaz: Removed

packages/react-components/styles/fonts.tsx
annakaz: Good point- will merge these to one

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
annakaz: Thanks for flagging this- the ExchangeConfirmationCard is now only used for reviewing past exchange (so we should not have to wait to fetch the fee), but the current fee and exchangeRate are being fetched instead of using the historic ones. Have created an issue here to track: https://github.com/celo-org/celo-monorepo/issues/2269

jmrossy (approved):
Nice!

PR #2170, opened 2019-12-11
[docs] Remove ref to script, Docker port security fixes and warning, remove office hrs

timmoreton:
* Docker with expose ports on all interfaces by default. Be super clear to only expose ports on localhost. Add back security explanation for this.
* Fix this in script
* Remove all references to script
* Remove office hours mentions.
* Clarify need to deregister then wait til end of epoch


asaj (approved):
Very nice

PR #2171, opened 2019-12-11
Changing validator group voting gold

lucasege:
### Description

Changed the amount of voted gold from all of group's gold to a configurable parameter of less gold for all the groups except the first and final group, which still vote with ~100k gold to maintain a presence in elections for monitoring.

### Tested

CI tests

### Other changes

Added config var `groupLockedGold`

### DIFF
--- a/packages/protocol/migrationsConfig.js
+++ b/packages/protocol/migrationsConfig.js
@@ -120,7 +120,6 @@ const DefaultConfig = {
     votesRatioOfLastVsFirstGroup: 2.0,
     groupLockedGold: {
       value: '22000000000000000000000', // 22k gold per group
-      duration: 60 * 24 * 60 * 60, // 60 days
     },
   },
 }
--- a/packages/protocol/migrationsConfig.js
+++ b/packages/protocol/migrationsConfig.js
@@ -118,6 +118,10 @@ const DefaultConfig = {
     groupName: 'C-Labs',
     commission: 0.1,
     votesRatioOfLastVsFirstGroup: 2.0,
+    groupLockedGold: {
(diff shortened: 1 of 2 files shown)


packages/protocol/migrationsConfig.js
```diff
@@ -118,6 +118,10 @@ const DefaultConfig = {
     groupName: 'C-Labs',
     commission: 0.1,
     votesRatioOfLastVsFirstGroup: 2.0,
+    groupLockedGold: {
+      value: '22000000000000000000000', // 22k gold per group
+      duration: 60 * 24 * 60 * 60, // 60 days
```
asaj: What is `duration`?

asaj (commented):
Can you create an issue to remove the previous logic?

Also please have @timmoreton take a look as he was the one who wrote that logic initially and may have other ideas about how this should be done

lucasege:
> Can you create an issue to remove the previous logic?
> 
> Also please have @timmoreton take a look as he was the one who wrote that logic initially and may have other ideas about how this should be done

I could just remove the logic in this PR - it seemed cleaner to leave it just in case we ever wanted to scale our groups' votes in the future, but if we're going to remove it we might as well just do it now.

packages/protocol/migrationsConfig.js
```diff
@@ -118,6 +118,9 @@ const DefaultConfig = {
     groupName: 'C-Labs',
     commission: 0.1,
     votesRatioOfLastVsFirstGroup: 2.0,
```
timmoreton: can remove this now?

timmoreton (approved):

PR #2174, opened 2019-12-11
Use fluentd natively for structured logging

### LINKED ISSUE #1831: Structured logging for VM testnets
### Expected Behavior

### Current Behavior

What is the current behavior?

nambrot:
### Description

We have used the `gcplogs` docker logging driver, however that one was basically impossible to configure. Instead, we are using the already installed `google-fluentd` agent with the `fluentd` log driver to parse JSON where possible. This should still print out non json lines as well. We could probably make an actual geth filter, but for now this should do

### Tested

on my testnet `namtest`

### Other changes

### Related issues

- Fixes #1831 

### Backwards compatibility


tkporter (approved):
Nice

PR #2178, opened 2019-12-11
Add fn groupMembershipInEpoch to validators.sol for slashing

### LINKED ISSUE #2087: Function in Validators to retrieve a validator's group membership at a particular epoch for slashing
### Expected Behavior
```
Contract Validators {
  // Checks the membership history entries of `account` and requires that `epochNumber` 
  // is within the range [entries[index].Epoch, entries[index + 1].Epoch).
  // Returns entries[index].group.
  function groupMembershipInEpoch(address account, uint256 epochNumber, uint256 index) external view returns (address);
}
```

### Current Behavior

No such function

lucasege:
### Description

Added `groupMembershipInEpoch` fn and related testing. Would like review on failure values (currently require(false) on failures). Also a look over logic would be nice.

### Tested

Added unit tests for general usage and edge cases.

### Related issues

- Fixes #2087

### Backwards compatibility

Yes

### DIFF
--- a/packages/protocol/contracts/governance/Validators.sol
+++ b/packages/protocol/contracts/governance/Validators.sol
@@ -1225,6 +1225,6 @@ contract Validators is
     ) {
       return history.entries[index].group;
     }
-    require(false, "provided index does not match provided epochNumber in history.");
+    require(false, "provided index does not match provided epochNumber at index in history.");
   }
 }
--- a/packages/protocol/contracts/governance/Validators.sol
+++ b/packages/protocol/contracts/governance/Validators.sol
@@ -1213,7 +1213,11 @@ contract Validators is
     view
     returns (address)
   {
-    require(isValidator(account) && epochNumber <= getEpochNumber());
+    require(
+      isValidator(account) &&
+        epochNumber <= getEpochNumber() &&
+        index < history.tail.add(history.numEntries)
+    );
     MembershipHistory storage history = validators[account].membershipHistory;
     require(index >= history.tail && history.numEntries > 0, "index out of bounds");
     bool isExactMatch = history.entries[index].epochNumber == epochNumber;
--- a/packages/protocol/contracts/governance/Validators.sol
+++ b/packages/protocol/contracts/governance/Validators.sol
@@ -1213,12 +1213,9 @@ contract Validators is
     view
     returns (address)
   {
-    require(
-      isValidator(account) &&
-        epochNumber <= getEpochNumber() &&
-        index < history.tail.add(history.numEntries)
-    );
+    require(isValidator(account) && epochNumber <= getEpochNumber());
     MembershipHistory storage history = validators[account].membershipHistory;
+    require(index < history.tail.add(history.numEntries));
     require(index >= history.tail && history.numEntries > 0, "index out of bounds");
     bool isExactMatch = history.entries[index].epochNumber == epochNumber;
     bool isLastEntry = index.sub(history.tail) == history.numEntries.sub(1);
--- a/packages/protocol/contracts/governance/Validators.sol
+++ b/packages/protocol/contracts/governance/Validators.sol
@@ -1215,13 +1215,13 @@ contract Validators is
   {
     require(isValidator(account) && epochNumber <= getEpochNumber());
     MembershipHistory storage history = validators[account].membershipHistory;
+    require(index >= history.tail && history.numEntries > 0, "index out of bounds");
+    bool isExactMatch = history.entries[index].epochNumber == epochNumber;
+    bool isLastEntry = index.sub(history.tail) == history.numEntries.sub(1);
+    bool isWithinRange = history.entries[index].epochNumber < epochNumber &&
+      (history.entries[index.add(1)].epochNumber > epochNumber || isLastEntry);
     require(
-      (history.numEntries > 0 &&
-        index >= history.tail &&
-        history.entries[index].epochNumber == epochNumber) ||
-        (history.entries[index].epochNumber < epochNumber &&
-          (history.entries[index.add(1)].epochNumber > epochNumber ||
-            index.sub(history.tail) == history.numEntries.sub(1))),
+      isExactMatch || isWithinRange,
       "provided index does not match provided epochNumber at index in history."
     );
     return history.entries[index].group;
--- a/packages/protocol/contracts/governance/Validators.sol
+++ b/packages/protocol/contracts/governance/Validators.sol
@@ -775,7 +775,6 @@ contract Validators is
       uint256 multiplier = Math.max(1, groups[account].members.numElements);
       uint256[] storage sizeHistory = groups[account].sizeHistory;
       if (sizeHistory.length > 0) {
-        // TODO(lucas): Why isn't this >=
         for (uint256 i = sizeHistory.length.sub(1); i > 0; i = i.sub(1)) {
           if (sizeHistory[i].add(groupLockedGoldRequirements.duration) >= now) {
             multiplier = Math.max(i, multiplier);
@@ -1217,14 +1216,13 @@ contract Validators is
   {
     require(isValidator(account) && epochNumber <= getEpochNumber());
     MembershipHistory storage history = validators[account].membershipHistory;
-    if (
+    require(
       history.entries[index].epochNumber == epochNumber ||
-      (history.entries[index].epochNumber < epochNumber &&
-        (history.entries[index.add(1)].epochNumber > epochNumber ||
-          history.entries[index.add(1)].epochNumber == 0))
-    ) {
-      return history.entries[index].group;
-    }
-    require(false, "provided index does not match provided epochNumber at index in history.");
+        (history.entries[index].epochNumber < epochNumber &&
+          (history.entries[index.add(1)].epochNumber > epochNumber ||
+            index == history.numEntries.sub(1))),
+      "provided index does not match provided epochNumber at index in history."
+    );
+    return history.entries[index].group;
   }
 }
--- a/packages/protocol/contracts/governance/Validators.sol
+++ b/packages/protocol/contracts/governance/Validators.sol
@@ -1211,16 +1211,17 @@ contract Validators is
   function groupMembershipInEpoch(address account, uint256 epochNumber, uint256 index)
     external
     view
-    onlyRegisteredContracts(canForceDeaffiliation)
     returns (address)
   {
     require(isValidator(account) && epochNumber <= getEpochNumber());
     MembershipHistory storage history = validators[account].membershipHistory;
     require(
-      history.entries[index].epochNumber == epochNumber ||
+      (history.numEntries > 0 &&
+        index >= history.tail &&
+        history.entries[index].epochNumber == epochNumber) ||
         (history.entries[index].epochNumber < epochNumber &&
           (history.entries[index.add(1)].epochNumber > epochNumber ||
-            index == history.numEntries.sub(1))),
+            index.sub(history.tail) == history.numEntries.sub(1))),
       "provided index does not match provided epochNumber at index in history."
     );
     return history.entries[index].group;
--- a/packages/protocol/contracts/governance/Validators.sol
+++ b/packages/protocol/contracts/governance/Validators.sol
@@ -775,6 +775,7 @@ contract Validators is
       uint256 multiplier = Math.max(1, groups[account].members.numElements);
       uint256[] storage sizeHistory = groups[account].sizeHistory;
       if (sizeHistory.length > 0) {
+        // TODO(lucas): Why isn't this >=
         for (uint256 i = sizeHistory.length.sub(1); i > 0; i = i.sub(1)) {
           if (sizeHistory[i].add(groupLockedGoldRequirements.duration) >= now) {
             multiplier = Math.max(i, multiplier);
@@ -1199,5 +1200,31 @@ contract Validators is
     require(isValidatorGroup(account));
     ValidatorGroup storage group = groups[account];
     return group.slashInfo.multiplier.unwrap();
+  }
+
+  /**
+   * @notice Returns the group that `account` was a member of during `epochNumber`.
+   * @param account The account whose group membership should be returned.
+   * @param epochNumber The epoch number we are querying this account's membership at.
+   * @param index The index into the validator's history struct for their history at `epochNumber`.
+   * @return The group that `account` was a member of during `epochNumber`.
+   */
+  function groupMembershipInEpoch(address account, uint256 epochNumber, uint256 index)
+    external
+    view
+    onlyRegisteredContracts(canForceDeaffiliation)
+    returns (address)
+  {
+    require(isValidator(account) && epochNumber <= getEpochNumber());
+    MembershipHistory storage history = validators[account].membershipHistory;
+    if (
+      history.entries[index].epochNumber == epochNumber ||
+      (history.entries[index].epochNumber < epochNumber &&
+        (history.entries[index.add(1)].epochNumber > epochNumber ||
+          history.entries[index.add(1)].epochNumber == 0))
+    ) {
+      return history.entries[index].group;
+    }
+    require(false, "provided index does not match provided epochNumber in history.");
   }
 }
--- a/packages/protocol/test/governance/validators.ts
+++ b/packages/protocol/test/governance/validators.ts
@@ -28,6 +28,7 @@ import {
   ValidatorsTestContract,
   ValidatorsTestInstance,
 } from 'types'
+import Web3 = require('web3')
 
 const Accounts: AccountsContract = artifacts.require('Accounts')
 const Validators: ValidatorsTestContract = artifacts.require('ValidatorsTest')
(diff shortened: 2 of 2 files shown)


packages/protocol/contracts/governance/Validators.sol
```diff
@@ -765,6 +765,7 @@ contract Validators is
       uint256 multiplier = Math.max(1, groups[account].members.numElements);
       uint256[] storage sizeHistory = groups[account].sizeHistory;
       if (sizeHistory.length > 0) {
+        // TODO(lucas): Why isn't this >=
```
asaj: What happens when you set it to >=?

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1132,30 @@ contract Validators is
...
+  * @notice Returns the group that `account` was a member of during `epochNumber`.
+  * @param account The account whose group membership should be returned.
+  * @param epochNumber The epoch number we are querying this account's membership at.
+  * @param index The index into the validator's history struct for their history at `epochNumber`.
+  * @return The group that `account` was a member of during `epochNumber`.
+  */
+  function groupMembershipInEpoch(address account, uint256 epochNumber, uint256 index)
+    external
+    view
+    onlyRegisteredContracts(canForceDeaffiliation)
```
asaj: Oops :)

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1132,30 @@ contract Validators is
...
+    MembershipHistory storage history = validators[account].membershipHistory;
+    if (
+      history.entries[index].epochNumber == epochNumber ||
+      (history.entries[index].epochNumber < epochNumber &&
+        (history.entries[index.add(1)].epochNumber > epochNumber ||
+          history.entries[index.add(1)].epochNumber == 0))
+    ) {
+      return history.entries[index].group;
+    }
+    require(false, "provided index does not match provided epochNumber at index in history.");
```
asaj: The more idiomatic approach would be to require that the condition you have in the `if` holds, and then to return `history.entries[index].group;`

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1132,30 @@ contract Validators is
...
+    onlyRegisteredContracts(canForceDeaffiliation)
+    returns (address)
+  {
+    require(isValidator(account) && epochNumber <= getEpochNumber());
+    MembershipHistory storage history = validators[account].membershipHistory;
+    if (
+      history.entries[index].epochNumber == epochNumber ||
+      (history.entries[index].epochNumber < epochNumber &&
+        (history.entries[index.add(1)].epochNumber > epochNumber ||
+          history.entries[index.add(1)].epochNumber == 0))
```
asaj: I don't think this is strictly correct.

In epoch 3, I join group A (entry index N is A)
In epoch 4, I leave group A (entry index N+1 is 0)
in epoch 5, I join group B (entry index N+2 is B)

groupMembershipInEpoch(me, 5, N) == A 😢

asaj: Please add a test case that would have caught this

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2184,116 @@ contract('Validators', (accounts: string[]) => {
...
+    const validator = accounts[0]
+    const groups = accounts.slice(1, -1)
+
+    beforeEach(async () => {
+      await registry.setAddressFor(CeloContractName.DowntimeSlasher, validator)
+      await registry.setAddressFor(CeloContractName.DoubleSigningSlasher, accounts[3])
+      await registry.setAddressFor(CeloContractName.Governance, accounts[4])
+    })
+
+    describe('when the validator is added to different groups sequantially', async () => {
```
asaj: typo

packages/protocol/contracts/governance/Validators.sol
lucasege: Ah cool, interesting "gotcha"

packages/protocol/contracts/governance/Validators.sol
lucasege: I assume this was an edit?

packages/protocol/contracts/governance/Validators.sol
lucasege: Addressed, please review again

packages/protocol/contracts/governance/Validators.sol
asaj: This is a view call, it should be fine for any address to call, so I don't think we need this modifier

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1131,29 @@ contract Validators is
...
+    onlyRegisteredContracts(canForceDeaffiliation)
+    returns (address)
+  {
+    require(isValidator(account) && epochNumber <= getEpochNumber());
+    MembershipHistory storage history = validators[account].membershipHistory;
+    require(
+      history.entries[index].epochNumber == epochNumber ||
+        (history.entries[index].epochNumber < epochNumber &&
+          (history.entries[index.add(1)].epochNumber > epochNumber ||
+            index == history.numEntries.sub(1))),
```
asaj: I don't think this is how you check that this is the last entry in `history` (though I agree that's what you want to check)

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1131,29 @@ contract Validators is
...
+    external
+    view
+    onlyRegisteredContracts(canForceDeaffiliation)
+    returns (address)
+  {
+    require(isValidator(account) && epochNumber <= getEpochNumber());
+    MembershipHistory storage history = validators[account].membershipHistory;
+    require(
+      history.entries[index].epochNumber == epochNumber ||
+        (history.entries[index].epochNumber < epochNumber &&
```
asaj: Are there any issues when any of these values are uninitialized? I don't think so, but please make sure you've thought about it carefully

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2184,129 @@ contract('Validators', (accounts: string[]) => {
       })
     })
   })
+
+  describe('#groupMembershipInEpoch', async () => {
+    const validator = accounts[0]
+    const groups = accounts.slice(1, -1)
+
+    beforeEach(async () => {
```
asaj: Not needed

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2184,129 @@ contract('Validators', (accounts: string[]) => {
...
+          epochs.push(epochNumber)
+          await mineBlocks(blocksUntilNextEpoch, web3)
+
+          if (i != 4) {
+            await validators.affiliate(groups[i])
+            await validators.addFirstMember(validator, NULL_ADDRESS, NULL_ADDRESS, {
+              from: groups[i],
+            })
+          } else {
+            await validators.forceDeaffiliateIfValidator(validator)
```
asaj: You don't need to force deaffiliate, you can just call deaffiliate from the validator address

packages/protocol/contracts/governance/Validators.sol
lucasege: Right, it should be `index.sub(history.tail) == history.numEntries.sub(1)` since the tail is the smallest index, so `index` - `tail` should give the offset in entries, which if equal to the total entries means we are on the last element. Good catch

packages/protocol/contracts/governance/Validators.sol
lucasege: I addressed one that I missed which is if the epoch number really was 0, which I think _is possible_. 
I believe all uninitialized cases are handled, assuming history is updated correctly.

packages/protocol/test/governance/validators.ts
lucasege: I added `await`s before `setAddressFor` to silence linter complaints, so now it is

packages/protocol/contracts/governance/Validators.sol
asaj: For readability, consider the following:
```
require(index >= history.tail && history.numEntries > 0);
bool isExactMatch = history.entries[index].epochNumber == epochNumber;
bool isLastEntry = index.sub(history.tail) == history.numEntries.sub(1);
bool isWithinRange = history.entries[index].epochNumber < epochNumber && (history.entries[index.add(1)].epochNumber > epochNumber || isLastEntry);
require(isExactMatch || isWithinRange);
```

packages/protocol/test/governance/validators.ts
asaj: No I mean there's no need to call `setAddressFor` for this test since `groupMembershipInEpoch` can be called by anyone

packages/protocol/test/governance/validators.ts
```diff
@@ -2178,10 +2178,145 @@ contract('Validators', (accounts: string[]) => {
...
+        await registerValidator(validator)
+        for (const group of groups) {
+          await registerValidatorGroup(group)
+        }
+        for (let i = 0; i < membershipHistoryLength.plus(1).toNumber(); i++) {
+          const blockNumber = await web3.eth.getBlockNumber()
+          const epochNumber = Math.floor(blockNumber / EPOCH)
+          const blocksUntilNextEpoch = (epochNumber + 1) * EPOCH - blockNumber
+          await mineBlocks(blocksUntilNextEpoch, web3)
+          epochs.push([epochNumber + 1, i !== 4 ? groups[i] : NULL_ADDRESS])
```
asaj: Why `i !== 4`? Would be nice to include more instances in which the validator was removed from the group, maybe `i % 3 === 0`?

packages/protocol/test/governance/validators.ts
```diff
@@ -2178,10 +2178,145 @@ contract('Validators', (accounts: string[]) => {
...
+
+          if (i !== 4) {
+            await validators.affiliate(groups[i])
+            await validators.addFirstMember(validator, NULL_ADDRESS, NULL_ADDRESS, {
+              from: groups[i],
+            })
+          } else {
+            await validators.forceDeaffiliateIfValidator(validator)
```
asaj: You don't need to force deaffiliate, you can just call deaffiliate from the validator address

packages/protocol/test/governance/validators.ts
```diff
@@ -2178,10 +2178,145 @@ contract('Validators', (accounts: string[]) => {
...
+          }
+        }
+      })
+
+      it('should always return the correct membership for the given epoch', async () => {
+        for (let i = 0; i < membershipHistoryLength.toNumber(); i++) {
+          assert.equal(
+            await validators.groupMembershipInEpoch(validator, epochs[i + 1][0], 2 + i),
+            i !== 3 ? epochs[i + 1][1] : NULL_ADDRESS
```
asaj: Please include a comment as to why this is different than above

packages/protocol/test/governance/validators.ts
```diff
@@ -2178,10 +2178,145 @@ contract('Validators', (accounts: string[]) => {
...
+          } else {
+            await validators.forceDeaffiliateIfValidator(validator)
+          }
+        }
+      })
+
+      it('should always return the correct membership for the given epoch', async () => {
+        for (let i = 0; i < membershipHistoryLength.toNumber(); i++) {
+          assert.equal(
+            await validators.groupMembershipInEpoch(validator, epochs[i + 1][0], 2 + i),
```
asaj: Can you add more comments explaining these args?

packages/protocol/test/governance/validators.ts
asaj: Ah, I see you do this below. IMO one test like this is sufficient, what extra coverage does this test give us?

packages/protocol/test/governance/validators.ts
```diff
@@ -2178,10 +2178,145 @@ contract('Validators', (accounts: string[]) => {
...
+        }
+        for (let i = 0; i < membershipHistoryLength.multipliedBy(3).toNumber(); i++) {
+          const blockNumber = await web3.eth.getBlockNumber()
+          const epochNumber = Math.floor(blockNumber / EPOCH)
+          const blocksUntilNextEpoch = (epochNumber + 1) * EPOCH - blockNumber
+          await mineBlocks(blocksUntilNextEpoch, web3)
+
+          if (i % 3 === 0) {
+            epochs.push([epochNumber + 1, i !== 9 ? groups[i / 3] : NULL_ADDRESS])
+            if (i !== 9) {
```
asaj: Why do this?

lucasege: The reason behind `i % 3 == 0` was to only add `membershipHistoryLength` entries - the reason was to avoid the confusion of only tracking the last `membershipHistoryLength` entries while still adding epochNumber gaps between entries. 
The reason behind `i !== 9` is to add a deaffiliation _between_ other entries. 
I'll definitely change this to be a function of `membershipHistoryLength` rather than magic numbers 3 and 4.
If you want me to do this differently, let me know and I can look into it.

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1131,30 @@ contract Validators is
...
+   * @return The group that `account` was a member of during `epochNumber`.
+   */
+  function groupMembershipInEpoch(address account, uint256 epochNumber, uint256 index)
+    external
+    view
+    returns (address)
+  {
+    require(isValidator(account) && epochNumber <= getEpochNumber());
+    MembershipHistory storage history = validators[account].membershipHistory;
+    require(index >= history.tail && history.numEntries > 0, "index out of bounds");
```
asaj, line 1219: Should we check the upper bound as well, i.e. `index < history.tail.add(history.numEntries)`

packages/protocol/test/governance/validators.ts
```diff
@@ -2178,10 +2178,98 @@ contract('Validators', (accounts: string[]) => {
...
+        // epochs stores [epochNumber, group] of the corresponding i + 1 indexed entry on chain
+        // i + 1 because registering groups adds a dummy null address as the first entry
+        epochs = []
+        await registerValidator(validator)
+        for (const group of groups) {
+          await registerValidatorGroup(group)
+        }
+        // Magic gapSize to allow for tests of off by one in both directions
+        const gapSize = 3
+        for (let i = 0; i < membershipHistoryLength.multipliedBy(gapSize).toNumber(); i++) {
```
asaj: Should we go for longer than this, so that we can test that getting the membership for an epoch too far in the past fails?

packages/protocol/test/governance/validators.ts
```diff
@@ -2178,10 +2178,98 @@ contract('Validators', (accounts: string[]) => {
...
+        for (let i = 0; i < membershipHistoryLength.toNumber(); i++) {
+          assert.equal(
+            await validators.groupMembershipInEpoch(validator, epochs[i][0], 1 + i),
+            i !== 3 ? epochs[i][1] : NULL_ADDRESS
+          )
+        }
+      })
+
+      it('should correctly get the group address of epoch numbers between switches', async () => {
+        assert.equal(
```
asaj: Why not test these conditions in a for loop through over all epochs? As is I find it a bit hard to read

packages/protocol/test/governance/validators.ts
```diff
@@ -2178,10 +2178,98 @@ contract('Validators', (accounts: string[]) => {
...
+          await validators.groupMembershipInEpoch(validator, epochs[4][0] + 1, 5),
+          epochs[4][1]
+        )
+        // `epochs`[2][0] join group A, epoch `epochs`[3][0] deaffiliate, `epochs`[4][0] join group B.
+        // This verifies we don't mistake 0 addresses as the end of history.
+        await assertRevert(validators.groupMembershipInEpoch(validator, epochs[4][0], 3))
+        await assertRevert(validators.groupMembershipInEpoch(validator, epochs[3][0], 3))
+      })
+
+      describe('when called with various malformed inputs', () => {
```
asaj, line 2278: Can we include tests for bad indexes? i.e. indexes that are either less than or greater than what's covered by `membershipHistory`?

packages/protocol/test/governance/validators.ts
```diff
@@ -2205,13 +2205,116 @@ contract('Validators', (accounts: string[]) => {
...
+
+        it("should revert when epochNumber is greater than the chain's current epochNumber", async () => {
+          const blockNumber = await web3.eth.getBlockNumber()
+          const epochNumber = Math.floor(blockNumber / EPOCH)
+          await assertRevert(
+            validators.groupMembershipInEpoch(validator, epochNumber + 1, contractIndex)
+          )
+        })
+
+        it('should revert when provided index is greater than greatest index on chain', async () => {
```
asaj, line 2307: Should we check that it reverts when the provided index is less than the tail?

asaj (approved):
Looks great!

packages/protocol/test/governance/validators.ts
lucasege: Line 2273 checks this by going through all of the indices that have been "pushed out" by adding more than `membershipHistoryLength` members, but for clarity I'll add another test here

Issue #1837, opened 2019-11-22
Remove autogen JS from CeloStats deploy


timmoreton:
This feels a little weird to generate a javascript file here, what do you think about either passing in the addresses as env variables here or maybe putting this into a .json file that's read by a .js file in ethstats?

_Originally posted by @tkporter in https://github.com/celo-org/celo-monorepo/pull/1714/files_

PR #2183, opened 2019-12-11
Ethstats variables as env variables

### LINKED ISSUE #1856: Pick up ethstats network name from .env
### Expected Behavior

When we deploy ethstats, we should be able to see the name of the network included in the Title of the page. 

### Current Behavior

We see "Celo", hardcoded via config in the Docker image.

jcortejoso:
### Description

Passing the ethstats configuration as environment variables as possible. Avoid compiling ethstats before running it.

### Tested

Tested on https://berlintestnet001-ethstats.celo-networks-dev.org/

### Related issues

- Fixes #1856, #1837, 

### Backwards compatibility

Needs to merge the ethstats PR https://github.com/celo-org/ethstats-server/pull/15


timmoreton (changes requested):
Why do we need blocks out here? I strongly disagree with having a dependency between ethstats and nlockscout - let’s chat

jcortejoso:
Hey @timmoreton ! There is not a strong dependency (ethstats would work with blockscout not available).
The blockscout url is used for the links under the field _Node Address_ and in _recent block proposers_, not sure if we can disable somehow if it is not defined. @diminator do you see any other dependency?

timmoreton (approved):

timmoreton:
@jcortejoso Ok awesome - this isn't the thing we were talking about :-)

diminator:
No other dependencies so far, just potentially dead links upon redirect

diminator (approved):
LGTM

PR #2186, opened 2019-12-11
[Wallet] Create run_app.sh script to facilitate app development

### LINKED ISSUE #2152: Ensure app is correctly configured before running
Currently the app can be built for the wrong SDK or configured for the wrong firebase DB when it's run.
Also `yarn dev` shows an error when first run forcing us to restart the app.

jmrossy:
### Description

Create a new, simpler way to run the app for development that solves the following pain points:
1. WalletKit out of sync with target network
2. Firebase  URL out of sync with target network
3. Android app must be restarted on first  run
4. Simpler way to enable hot reload
5. Simpler way to skip unnecessary steps 

Still TODO:
- convert E2E tests to use this
- Address iOS pain points
- Update blockchain-api to use net firebase integration url

### Tested

Locally on Mac. 
Testing on Linux still TODO, may do later.

### Related issues

- Fixes #2152 

### Backwards compatibility

N/A


i1skn (approved):
Great work!! LGTM from the command API perspective (I'm no expert in `sh` scripts).

packages/mobile/package.json
```diff
@@ -5,17 +5,15 @@
   "license": "Apache-2.0",
   "private": true,
   "scripts": {
-    "start": "react-native start",
-    "start:bg": "react-native start &",
+    "start": "./scripts/run_app.sh",
```
i1skn: Just to clarify: `yarn start` now would fail with a proposal to specify a platform, correct?

jmrossy: Correct. Maybe it's not worth it to have this still here?

i1skn: I think this is fine. I would assume, this is the first candidate to run when somebody has no experience with the codebase.

packages/mobile/package.json
```diff
@@ -5,17 +5,15 @@
...
+    "start:ios": "./scripts/run_app.sh -p ios",
     "lint": "tslint -c tslint.json --project tsconfig.json",
     "build": "yarn run build:ts && yarn run build:metro",
     "build:sdk": "yarn --cwd ../walletkit build:for-env",
     "build:ts": "tsc --noEmit",
     "build:metro": "echo 'NOT WORKING RIGHT NOW'",
     "build:gen-graphql-types": "gql-gen --schema http://localhost:8080/graphql --template graphql-codegen-typescript-template --out ./typings/ 'src/**/*.tsx'",
-    "predev": "./scripts/pre-dev.sh",
-    "dev": "react-native run-android --appIdSuffix \"debug\" --no-packager && yarn start || echo 'Could not start metro server'",
-    "dev:ios": "react-native run-ios --simulator \"iPhone 11\"",
```
jeanregisser, line 18: I have a preference for keeping `dev:android` and `dev:ios` instead of `start:android` and `start:ios` as RN devs are too used to `start` being for starting only the packager (as in `react-native start`).

packages/mobile/scripts/run_app.sh
```diff
@@ -0,0 +1,123 @@
...
+  else 
+    # Run android without packager because RN cli doesn't work with yarn workspaces
+    yarn react-native run-android --appIdSuffix \"debug\" --no-packager
+    yarn react-native start 
+  fi
+
+elif [ $PLATFORM = "ios" ]; then
+  echo "Using platform ios"
+  # TODO have iOS build and start from command line
+  echo -e "\nFor now ios must be build and run from xcode\nStarting RN bundler\n"
```
jeanregisser, line 116: We can do `yarn react-native run-ios --simulator "iPhone 11"`. Last time I checked it worked.

packages/mobile/scripts/run_app.sh
```diff
@@ -0,0 +1,123 @@
...
+  fi
+
+  if [ $MACHINE = "Mac" ]; then
+    echo "Starting packager in new terminal"
+    RN_START_CMD="cd `pwd`;yarn react-native start"
+    OSASCRIPT_CMD="tell application \"Terminal\" to do script \"$RN_START_CMD\""
+    echo "FULLCMD: $OSASCRIPT_CMD"
+    osascript -e "$OSASCRIPT_CMD"
+    # Run android without packager because RN cli doesn't work with yarn workspaces
+    yarn react-native run-android --appIdSuffix \"debug\" --no-packager
```
jeanregisser: nit: We don't need the backslashes

packages/mobile/scripts/run_app.sh
```diff
@@ -0,0 +1,123 @@
...
+
+# ====================================
+# Configure and run the mobile app
+# ====================================
+
+# Flags:
+# -n: Name of the network to run on
+# -p: Platform (android or ios)
+# -f: Fast (skip steps not required unless network or depedencies changes)
+# -h: Hot Reload (Restore nav state on reload)
```
jeanregisser: `-h` is usually used for help, your suggestion of `-r` is better IMHO.
Also would rename `Hot Reload` to what it does: "Restore nav state on reload"

packages/mobile/scripts/run_app.sh
```diff
@@ -0,0 +1,123 @@
...
+    cd ../../ && yarn run jetify && cd packages/mobile
+    echo "Jetified"
+  fi
+
+  if [ $MACHINE = "Mac" ]; then
+    echo "Starting packager in new terminal"
+    RN_START_CMD="cd `pwd`;yarn react-native start"
+    OSASCRIPT_CMD="tell application \"Terminal\" to do script \"$RN_START_CMD\""
+    echo "FULLCMD: $OSASCRIPT_CMD"
+    osascript -e "$OSASCRIPT_CMD"
```
jeanregisser, line 104: Would be cool to do that only if the packager is not already running.
I usually start the packager separately so I can launch it with custom options like `--reset-cache`.

jeanregisser (approved):
Great you took the time for this! 🎉

packages/mobile/scripts/run_app.sh
jmrossy: For me it didn't work but I'll try again :)

Do you want it to start it though? I thought you preferred Xcode? Maybe there's a way to trigger Xcode actions from command line?

packages/mobile/package.json
jmrossy: Yep considered that too. I'm open to `dev`
I would have preferred `run` but that's a yarn keyword already

packages/mobile/scripts/run_app.sh
jmrossy: Ah yep good catch. Copy paste error

packages/mobile/scripts/run_app.sh
jmrossy: Will do!

packages/mobile/scripts/run_app.sh
jmrossy: What does reset cache do? Would you like  a flag  for the script  that sets it?

packages/mobile/scripts/run_app.sh
jeanregisser: oh yeah, iPhone 11 only works from Xcode 11, 
I just checked, and `yarn react-native run-ios --simulator "iPhone Xʀ"` works.

I think it makes sense to build it from the command line by default and have another switch to open Xcode too. What do you think?

packages/mobile/scripts/run_app.sh
jeanregisser: It removes cached files, makes sure all files are transformed again. Fixes edge cases when changing babel config or transformer config.
Not sure it's worth having a flag for now here.

PR #2188, opened 2019-12-11
Update initial cUSD/cGLD exchange rate

yerdua:
### Description

So far, there's been a default, hard-coded value of 10 as the gold price in cUSD. This is used in the oracle report in migrations (see: 08_stabletoken.ts).
Even if this value is soon overwritten by a running oracle, or expired to let the oracle take over, we want to start this out at a value that is close to the simulated oracle values that are reported in testnets. In `baklava`, the simulated oracle values hover around 1.

### Tested

Had to fix a `contractkit` test. Hopefully nothing else relies on the rate being set to `10`.


tkporter (approved):
Thanks!

asaj:
Why put this in the .env file rather than the migrations config?

yerdua:
> Why put this in the .env file rather than the migrations config?

Is there a way to do that in an env-specific way? I didn't see an obvious way of accomplishing that without using envvars.

asaj:
Why do we care to do this in an env specific way?

PR #2189, opened 2019-12-11
Add /test_attestations endpoint on the attestation service

### LINKED ISSUE #2093: Attestation service users should be able to test if they can send text messages correctly
### Expected Behavior

There should be a command that let's validators quickly test that their setup is correct

### Current Behavior

Hard to tell until someone requests attestations from the validator

nambrot:
### Description

This PR adds a test endpoint to let validators verify that they have set up the attestation service correctly. We authenticate the request by signing with the validator account key. You can invoke the endpoint by running the following CLI command:

`identity:test-attestation-service --from 0x0 --phoneNumber +14152223333 --message test`

### Tested

- Tested locally

### Related issues

- Fixes #2093


packages/cli/src/commands/identity/test-attestation-service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  static examples = ['test-attestation-service --from 0x97f7333c51897469E8D98E7af8653aAb468050a3']
+
+  requireSynced = false
+  async run() {
+    const { flags } = this.parse(TestAttestationService)
+    const address = flags.from
+    const { phoneNumber, message } = flags
+
+    console.log(address)
```
mcortesi: debug msg?

mcortesi (approved):

PR #2191, opened 2019-12-11
Check sync status of attestation service

nambrot:
### Description

This PR copies the CLI sync check.

It also changes the default logging scheme from `json` to `human`.

### Tested

- Ran locally


packages/attestation-service/src/db.ts
```diff
@@ -22,11 +23,31 @@ export let kit: ContractKit
 export async function initializeKit() {
   if (kit === undefined) {
     kit = newKit(fetchEnv('CELO_PROVIDER'))
-    const blockNumber = await kit.web3.eth.getBlockNumber()
-    if (blockNumber === 0) {
-      throw new Error(
-        'Could not fetch latest block from web3 provider ' + fetchEnv('CELO_PROVIDER')
+    // Copied from @celo/cli/src/utils/helpers
```
mcortesi, line 26: would be better to have a function for this

mcortesi (approved):

packages/attestation-service/src/db.ts
nambrot: I thought about it, but we only have `utils` and I did not want to add `web3` as a dependency

packages/attestation-service/src/db.ts
```diff
@@ -22,11 +23,31 @@ export let kit: ContractKit
...
+          // on the latest block
+          const ageOfBlock = Date.now() / 1000 - latestBlock.timestamp
+          if (ageOfBlock > 120) {
+            throw new Error(
+              `Latest block is ${ageOfBlock} seconds old, and syncing is not currently in progress`
+            )
+          }
+        }
+      } else {
+        throw new Error('Node is not syncing')
```
szym, line 43: Is this the right message? If `eth_syncing` returned a non-`false` then the node *is syncing*. Should this message read `'Node is not synced'` instead?

nategraf: Good catch. I've opened https://github.com/celo-org/celo-monorepo/pull/2372

PR #2192, opened 2019-12-11
Support more than 1 attesation bot at a time

nambrot:
### Description

1. Increases the number of accounts we faucet cUSD for the attestation bot
2. Converts the deployment to a stateful set so that we can run more than one

### Tested

- [x] On baklavastaging


packages/celotool/src/lib/migration-utils.ts
```diff
@@ -36,7 +36,7 @@ function getAttestationKeys() {
 export function migrationOverrides() {
   const mnemonic = fetchEnv(envVar.MNEMONIC)
   const faucetedAccountAddresses = getFaucetedAccounts(mnemonic).map((account) => account.address)
-  const attestationBotAddresses = getAddressesFor(AccountType.ATTESTATION_BOT, mnemonic, 1)
+  const attestationBotAddresses = getAddressesFor(AccountType.ATTESTATION_BOT, mnemonic, 10)
```
nategraf, line 39: Could this be configurable?
```suggestion
  const attestationBotAddresses = getAddressesFor(AccountType.ATTESTATION_BOT, mnemonic, parseInt(fetchEnvOrFallback(envVar.ATTESTATION_BOTS, '10'), 10))
```

nategraf (approved):

PR #2195, opened 2019-12-11
Add attestation info for reconnecting

nambrot:
### Description

The reconnecting instructions do not contain info re: Attestations


packages/docs/getting-started/reconnecting-to-network.md
```diff
@@ -40,6 +43,12 @@ ls celo-accounts-node/keystore
 ls celo-validator-node/keystore
 ```
 
+```bash
+# On your Attestation machine
+# You should see the keystore file for your Attestation signer key.
+ls celo-attestations-node/keystore
+```
+
```
timmoreton, line 51: "For each of these, now back them up."

timmoreton (approved):

PR #2196, opened 2019-12-11
Upgrade TS version

cmcewen:
### Description

Was having some issues updating web3 types so just decided to do this as a chore.

Should make better types for redux-saga accessible - not sure the work here: https://github.com/redux-saga/redux-saga/issues/1932

### Tested

Ran the app, saw analytics wrapper appeared as expected


annakaz (approved):
lgtm 👍

PR #2199, opened 2019-12-11
Upgrade web3 version in mobile, celotool, and walletkit

cmcewen:
### Description

Need to upgrade our web3 version due to some Forno issues in mobile

### Tested

Tested in the SL pilot

### Other changes

Changed the hex parsing in web3 utils


packages/mobile/src/web3/contracts.ts
```diff
@@ -13,6 +13,7 @@ import { Provider } from 'web3/providers'
 const tag = 'web3/contracts'
 
 export const web3: Web3 = getWeb3()
+// @ts-ignore
```
annakaz: Add a few words on why this is needed?

annakaz (approved):
👍

packages/mobile/src/escrow/saga.ts
```diff
@@ -144,7 +144,7 @@ function* withdrawFromEscrow() {
     signature = signature.slice(2)
     const r = `0x${signature.slice(0, 64)}`
     const s = `0x${signature.slice(64, 128)}`
-    const v = web3.utils.hexToNumber(signature.slice(128, 130))
+    const v = web3.utils.hexToNumber(`0x${signature.slice(128, 130)}`)
```
jmrossy: Nit: use the `ensureHexLeader` helper here
https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/address.ts#L49

packages/walletkit/package.json
```diff
@@ -30,8 +30,8 @@
     "eth-lib": "^0.2.8",
     "lodash": "^4.17.14",
     "sleep-promise": "^8.0.1",
-    "web3": "1.0.0-beta.37",
-    "web3-utils": "1.0.0-beta.37"
+    "web3": "1.2.4",
+    "web3-utils": "1.2.4"
```
jmrossy, line 35: Why do we import web3-utils in wallet kit? It's included in web3 right? Nbd if you'd prefer to leave it as is though

cmcewen: Not sure exactly why - think it is accessible by web3.utils, but I am just trying to minimize walletkit changes till we can deprecate

packages/celotool/src/lib/geth.ts
```diff
@@ -198,24 +190,24 @@ export const checkGethStarted = (dataDir: string) => {
 }
 
 export const getWeb3AndTokensContracts = async () => {
```
annakaz, line 201: Rename to reflect kit is returned?

packages/celotool/src/cmds/geth/trace.ts
```diff
@@ -45,8 +45,8 @@ export const handler = async (argv: TraceArgv) => {
     try {
       web3AndContracts = await getWeb3AndTokensContracts()
       // tslint:disable-next-line: no-shadowed-variable
```
annakaz: Why disable instead of using a different name?

packages/celotool/src/cmds/geth/trace.ts
```diff
@@ -45,8 +45,8 @@ export const handler = async (argv: TraceArgv) => {
...
-      const { web3 } = web3AndContracts!
-      const latestBlock = await web3.eth.getBlock('latest')
+      const { kit } = web3AndContracts!
```
annakaz: Is the ! necessary with a function signature for `getWeb3AndTokensContracts`?

packages/celotool/src/cmds/geth/trace.ts
```diff
@@ -67,7 +67,7 @@ export const handler = async (argv: TraceArgv) => {
     process.exit(1)
   }
 
-  const { web3, goldToken, stableToken } = web3AndContracts!
+  const { kit, goldToken, stableToken } = web3AndContracts!
```
annakaz, line 69: Same comment about !

annakaz (commented):
Has this been tested with geth mode on on mobile (given that SL used forno)? 
Also nit: can you rename PR to reflect that this upgrades celotool and walletkit too

annakaz (approved):
Approving assuming tested on mobile in forno and geth mode!

jmrossy (approved):

Issue #1741, opened 2019-11-15
[Wallet] Users on 3g network SBAT get tx receipts more reliably


cmcewen:
### Expected Behavior

We always get tx receipts and handle properly

### Current Behavior

Sometimes the app will be waiting for a tx to go through even though it was successful. Example - user gets three texts, app submits the txs, but the last one hangs for multiple minutes. User restarts the app and the verification flow is completed right away without any other txs

User sends money, tx appears hanging in feed for a long time and a duplicated transaction shows up

![WhatsApp Image 2019-11-15 at 2 20 18 PM](https://user-images.githubusercontent.com/1307633/68950328-d4dcc500-07b3-11ea-88db-f87a351e48b4.jpeg)

annakaz:
Adding this to 12/11 as it may be related to https://github.com/celo-org/celo-monorepo/issues/2003

annakaz:
Updated title as this issue occurs in geth mode and forno mode when using a slow connection

Issue #451, opened 2019-08-14
Devs SBAT profile celo-blockchain data usage in the wallet


asaj:
### Expected Behavior

We have a clear understanding of the costs (both in data and in time) of syncing, view calls, etc.

### Current Behavior

Back of the envelope calculations

annakaz:
See also making view calls fast- https://github.com/celo-org/celo-monorepo/issues/928

jeanregisser:
Moving to roadmap.

PR #2204, opened 2019-12-11
[Wallet] Use Charles proxy to see eth JSON rpc calls when using forno

annakaz:
### Description

Update wallet network security config so we can use Charles as a proxy. This allows us to see eth JSON rpc calls while using forno, which can be helpful for debugging. Also added instructions to set up Charles

### Tested

On a pixel 2
Does not yet support iOS

### Related issues

Set this up to debug slow transaction receipts (#1741 ) and profile app data usage using forno (see #451)

### Backwards compatibility

Yes, but may need to run `cd android && ./gradlew clean`


annakaz:
Example of what we can see using the proxy 
Screen Shot 2019-12-11 at 2 46 23 PM

jmrossy (approved):
Cool!

jeanregisser (approved):
Great! This should work already on iOS too :+1:

PR #2206, opened 2019-12-11
Remove walletkit from celotool transactions commands

nambrot:
### Description

Removes the dependency on walletkit to parse txs and logs

### Tested

- Ran locally


cmcewen (approved):
🙏

PR #2207, opened 2019-12-12
Wait for only waitTime - 1 blocks

nambrot:
### Description

I realized that we were waiting for too many blocks. Once we are at waitTime - 1, the next TX has to go in at >= waitTime and thus we are fine to submit the `selectIssuers` tx then.

### Tested

- Ran it locally


timmoreton (approved):
Makes sense!

PR #2216, opened 2019-12-12
Update genesis value, show name in election:list, fix assertRevert

asaj:
### Description

This PR does various unrelated things:
- Update baklava genesis balances from 12 to 12,000
- Show name in election:list and don't fetch full validator group when address is sufficient
- Fix assertRevert, which was previously always returning true
- Fix a subtle bug in Validators.sol that in some cases would allow Validator Groups to register one too many members. Add test to cover this.

### Tested

- Unit tests
### Other changes

None


packages/protocol/test/governance/validators.ts
```diff
@@ -1518,15 +1518,34 @@ contract('Validators', (accounts: string[]) => {
...
+              await mockLockedGold.setAccountTotalLockedGold(
+                group,
+                groupLockedGoldRequirements.value.times(2).minus(1)
+              )
+              await validators.addFirstMember(validator, NULL_ADDRESS, NULL_ADDRESS)
+              await registerValidator(validator2)
+              await validators.affiliate(group, { from: validator2 })
+            })
+
+            it('should succeed', async () => {
```
nategraf: Should this be "should revert"?
```suggestion
            it('should revert', async () => {
```

nategraf (approved):
Looks good. I left one comment

packages/protocol/test/governance/validators.ts
asaj: Sure should

PR #2222, opened 2019-12-12
[Wallet] Fix crash on iOS when segment is enabled

jeanregisser:
### Description

While preparing a TestFlight build I noticed the app was crashing on launch.
This was caused by `Segment-Firebase` which tried to initialize Firebase when it had already been done.
We didn't notice until now as we haven't created a new TestFlight build since Segment and React Native were upgraded. And Segment is not enabled in dev builds.

Fix was available upstream in `Segment-Firebase` 2.5.0 (https://github.com/segment-integrations/analytics-ios-integration-firebase/pull/47) but required `Firebase` >= 6.2.

Given the implications of switching to `Firebase` 6, I simply [relaxed that requirement in the podspec](https://github.com/celo-org/analytics-ios-integration-firebase/commit/e849d040239a7e56e02df96aabc0851e56a4cb19) (the `Segment-Firebase` 2.5.0 update doesn't actually require new functionalities from `Firebase` 6.

Note: we can revert this commit when we upgrade to `Firebase` 6

### Tested

App runs without crashing on launch.

### Other changes

N/A

### Related issues

Discussed on slack

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/ios/Podfile
+++ b/packages/mobile/ios/Podfile
@@ -45,6 +45,11 @@ target "celo" do
   pod "Firebase/Database"
   pod "GoogleUtilities", "~> 5.3.7"
 
+  # Using a custom fork so we can have a specific bug fix (crash when firebase already initialized) 
+  # without upgrading to firebase 6
+  # TODO: remove this once we upgrade to Firebase >= 6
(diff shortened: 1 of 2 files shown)


i1skn (approved):
LGTM

jmrossy (approved):

Issue #1639, opened 2019-11-08
End-to-end test of governable block gas limit


asaj:
### Expected Behavior

We have an end-to-end test of the governable block gas limit.

### Current Behavior

No end-to-end test

asaj:
This could be expanded more generally into an end-to-end test of the `BlockChainParameters` interactions, perhaps repurposing `end-to-end-exit-test` to be `end-to-end-block-chain-parameters-test`

mrsmkl:
Was handled by #1245

PR #2223, opened 2019-12-12
Fix/block gas limit exceeded

H34D:
### Description

From time to time i get `block gas limit exceeded` on the e2e tests right in the migrations. I am pretty sure this is the cause.

It looks a little like this when it happens:

tenor

### Tested

`yarn test`
using e2e tests

### Issues

Related to: #1639


asaj (approved):
Lmao, love that episode 😄

H34D:
given up for #2263

PR #2230, opened 2019-12-12
Make attestation opt-in

nambrot:
### Description

This PR makes attestation opt-in. Only when a validator has authorized an attestation key will they be considered as an issue for `Attestations`.

### Tested

- [x] Unit tests

### Other changes

- Change the attestation-bot deploy to pull the genesis instead of generating. It shall be known that @tkporter warned me of this and I ignored him to my detriment.

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/179


tkporter (approved):
I'm no solidity genius but lgtm

Lovely shoutout, much appreciated!

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -572,6 +572,11 @@ contract Attestations is
       seed = keccak256(abi.encodePacked(seed));
       validator = validatorSignerAddressFromCurrentSet(uint256(seed) % numberValidators);
       issuer = getAccounts().validatorSignerToAccount(validator);
+
+      if (!getAccounts().hasAuthorizedAttestationSigner(issuer)) {
```
asaj: There's an optimization we can make here, which is to, outside of the while loop, set `IAccounts accounts = getAccounts();`, and then inside the loop call `accounts.validatorSignerToAccount()/accounts.hasAuthorizedAttestationSigner()`

packages/protocol/test/identity/attestations.ts
```diff
@@ -609,34 +637,6 @@ contract('Attestations', (accounts: string[]) => {
       })
     })
 
-    describe('when an attestor has been authorized', () => {
```
asaj, line 612: Why remove these tests?

nambrot: In the test setup, we set up an authorized attestation signer for every validator, thus this becomes superflous

PR #2232, opened 2019-12-12
[Wallet] Refer to Celo Lite as "Data Saver" mode

### LINKED ISSUE #1985: [Wallet] Devs SNBAT confused Forno/ZeroSync/CeloLite names
### Expected Behavior

Mobile codebase makes it clear that "Forno" refers to hosted node service, and uses some consistent name ("Trusted Mode" perhaps to avoid Celo Light Client/Celo Lite confusion) displayed to app users. 

### Current Behavior

There has been some confusion between whether "Celo Lite" - what's used in settings to refer to using the hosted Forno node as a provider - refers to the Celo Ultralight client (it does not), as Forno/ZeroSync/CeloLite are used interchangeably throughout codebase. 

The Celo Lite terminology comes from spec here- https://github.com/celo-org/celo-monorepo/issues/1171

cc @nityas

annakaz:
### Description

There is internal confusion as to whether Celo Lite refers to the ultralight client (it does not), or whether it is related to the Light Wallet (it is not).

This replaces the term Celo Lite with "Data Saver" (the user-facing term) == zeroSync (developer-facing term which makes it clear we're not syncing block headers) for using the app with a hosted forno node. Note that forno refers to the hosted node service. 

### Tested

Ran tests and asked a few people if this was less confusing

### Related issues

- Fixes #1985 

### Backwards compatibility

Yes, this just changes the wording on the settings page


jmrossy (approved):
💯

PR #2239, opened 2019-12-13
Adding slashing multiplier to validators.sol and reward calculation

### LINKED ISSUE #2085: Add slashing multiplier to Validators contract, read when distributing rewards
### Expected Behavior

Rewards are multiplied by slashing multiplier. 

More specifically, the reward to the validator is multiplied by `M` for the corresponding account, the group share is multiplied by `M` for the group account, and rewards to voters are multiplied by `M` for the group account. 

```
Contract Validators {
  uint256 public slashingMultiplierResetPeriod;
  function setSlashingMultiplierResetPeriod(uint256 value) external onlyOwner;

  // Resets a group’s slashing multiplier if it has been >=
  // slashingPenaltyResetPeriod since the last time the account was slashed.
  function resetSlashingMultiplier() external;

  // Only callable by the registered DoubleSigningSlasher and DowntimeSlasher
  // contracts.
  function halveSlashingMultiplier(address group) external;
}
```
### Current Behavior

No `M`

lucasege:
### Description

Added `setSlashingMultiplierResetPeriod`, `resetSlashingMultiplier`, `getValidatorGroupSlashingMultiplier` and `halveSlashingMultiplier` to `Validators.sol`, as well as adding record keeping to faciliate this: `GroupSlashInfo` struct which is stored within the `ValidatorGroup` struct and maintains the group's `slashingModifier` (`M` in the docs) and the `lastSlashedTimestamp` which tracks the last time a group was slashed to control when the multiplier can be reset yet.

### Tested

Added unit tests for each fn in Validators contract.

### Other changes

Fixed a bug in `Validators.sol` where we were calling `signerToAccount` when we should have been calling `validatorSignerToAccount`. Updated this in most functions that called `signerToAccount` with a few exceptions: `_updateValidatorScoreFromSigner`, `_distributeEpochPaymentsFromSigner`, `getValidatorBlsPublicKeyFromSigner`, and `getMembershipInLastEpochFromSigner`.

Changed `Validators.sol` `initialize` fn signature, IValidators contract, and the MockValidators contract. I also changed the `gasLimit` within truffle config and the ganache config in `runTests` since it was running out of gas when I changed Validator.

### Related issues

- Fixes #2085

### Backwards compatibility

No, changes to initialize fn in Validator


packages/protocol/contracts/governance/Election.sol
```diff
@@ -433,8 +433,16 @@ contract Election is
     FixidityLib.Fraction memory score = FixidityLib.wrap(
       getValidators().calculateGroupEpochScore(uptimes)
     );
+    FixidityLib.Fraction memory slashingMultiplier = FixidityLib.wrap(
+      getValidators().getValidatorGroupSlashingMultiplier(group)
```
asaj: nit: every call to `getValidators()` makes a call to the registry to get the address for  `Validators`. We do that 3 times here, and would expect the same address every time.

Instead, consider 
```
IValidators validators = getValidators();
if(validators.meetsAccountLocked...
```

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,63 @@ contract Validators is
       }
     }
   }
+
+  /**
+   * @notice Sets the slashingMultiplierRestPeriod property if called by owner
+   * @param value New reset period for slashing multiplier
+   */
+  function setSlashingMultiplierResetPeriod(uint256 value) public nonReentrant onlyOwner {
+    // TODO(reviewer): is this necessary? This is done in `setMaxGroupSize`
```
asaj: It's not strictly necessary, no

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,63 @@ contract Validators is
...
+    require(value != slashingMultiplierResetPeriod);
+    slashingMultiplierResetPeriod = value;
+  }
+
+  /**
+   * @notice Resets a group's slashing multiplier if it has been >= the reset period since 
+   *         the last time the group was slashed.
+   */
+  function resetSlashingMultiplier() external nonReentrant {
+    address account = getAccounts().signerToAccount(msg.sender);
```
asaj: Oh boy. This should be `validatorSignerToAccount`, as it should be in almost every other function, other than `_updateValidatorScoreFromSigner`, `_distributeEpochPaymentsFromSigner`, `getValidatorBlsPublicKeyFromSigner`, and `getMembershipInLastEpochFromSigner`.

Would you might updating?

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,63 @@ contract Validators is
...
+    require(isValidatorGroup(account));
+    ValidatorGroup storage group = groups[account];
+    // TODO: Should this fail silently or revert? Or return bool/multiplier value
```
asaj: Revert!

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,63 @@ contract Validators is
...
+    if (now >= group.slashInfo.lastSlashedTimestamp + slashingMultiplierResetPeriod) {
```
asaj: Safemath please

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,63 @@ contract Validators is
...
+  bytes32[] canHalveSlashingMultiplier = [
+    DOWNTIME_SLASHER_REGISTRY_ID,
+    DOUBLE_SIGNING_SLASHER_REGISTRY_ID
+  ];
+
+  /**
+   * @notice Halves the group's slashing multiplier.
+   * @param groupAddress The group being slashed.
+   */
+  function halveSlashingMultiplier(address groupAddress)
```
asaj: nit: `Address` is redundant, consider `account`

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,63 @@ contract Validators is
...
+    external
+    nonReentrant
+    onlyRegisteredContracts(canHalveSlashingMultiplier)
+  {
+    require(isValidatorGroup(groupAddress));
+    ValidatorGroup storage group = groups[groupAddress];
+    group.slashInfo.slashingMultiplier = group.slashInfo.slashingMultiplier.divide(
```
asaj: We don't need fixidity division here:

```
group.slashInfo.slashingMultiplier = FixidityLib.wrap(group.slashInfo.slashingMultiplier.unwrap().div(2));
```

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,63 @@ contract Validators is
...
+      FixidityLib.newFixed(2)
+    );
+    group.slashInfo.lastSlashedTimestamp = now;
+  }
+
+  /**
+	 * @notice Getter for a group's slashing multiplier
```
asaj: style

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,63 @@ contract Validators is
...
+	 */
+  function getValidatorGroupSlashingMultiplier(address groupAddress)
```
asaj: consider `account`

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -81,6 +82,11 @@ contract Validators is
     uint256 lastRemovedFromGroupTimestamp;
   }
 
+  struct GroupSlashInfo {
+    FixidityLib.Fraction slashingMultiplier;
```
asaj: nit: Consider
```
struct SlashingInfo {
  FixidityLib.Fraction penalty;
  uint256 lastSlashed;
}
```

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -108,6 +114,7 @@ contract Validators is
   ValidatorScoreParameters private validatorScoreParameters;
   uint256 public membershipHistoryLength;
   uint256 public maxGroupSize;
+  uint256 public slashingMultiplierResetPeriod;
```
asaj, line 117: If you take the above suggestion, consider `slashingPenaltyResetPeriod`

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2191,97 @@ contract('Validators', (accounts: string[]) => {
       })
     })
   })
+
+  describe('#halveSlashingMultiplier', async () => {
+    const validator = accounts[0]
```
asaj: Why do you need to register and affiliate a validator here?

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2191,97 @@ contract('Validators', (accounts: string[]) => {
...
+    beforeEach(async () => {
+      await registerValidator(validator)
+      await registerValidatorGroup(group)
+      await validators.affiliate(group)
+    })
+
+    describe('when run from an approved address', async () => {
+      beforeEach(async () => {
+        registry.setAddressFor(CeloContractName.DowntimeSlasher, accounts[2])
+        registry.setAddressFor(CeloContractName.DoubleSigningSlasher, accounts[3])
```
asaj: Do you use this? If not, why register it?

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2191,97 @@ contract('Validators', (accounts: string[]) => {
...
+
+    describe('when called from an unapproved address', async () => {
+      it('should revert', async () => {
+        await assertRevert(validators.halveSlashingMultiplier(group))
+      })
+    })
+  })
+
+  describe('#resetSlashingMultiplier', async () => {
+    const validator = accounts[0]
```
asaj, line 2254: Ditto

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2191,97 @@ contract('Validators', (accounts: string[]) => {
...
+    const group = accounts[1]
+
+    beforeEach(async () => {
+      await registerValidator(validator)
+      await registerValidatorGroup(group)
+      await validators.affiliate(group)
+      registry.setAddressFor(CeloContractName.DowntimeSlasher, accounts[0])
+      registry.setAddressFor(CeloContractName.Governance, accounts[2])
```
asaj: Ditto

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2191,97 @@ contract('Validators', (accounts: string[]) => {
...
+
+  describe('#resetSlashingMultiplier', async () => {
+    const validator = accounts[0]
+    const group = accounts[1]
+
+    beforeEach(async () => {
+      await registerValidator(validator)
+      await registerValidatorGroup(group)
+      await validators.affiliate(group)
+      registry.setAddressFor(CeloContractName.DowntimeSlasher, accounts[0])
```
asaj: Would recommend not using `accounts[0]` for this, since it's the default `from` address

packages/protocol/test/governance/validators.ts
```diff
@@ -2184,4 +2191,97 @@ contract('Validators', (accounts: string[]) => {
...
+      registry.setAddressFor(CeloContractName.Governance, accounts[2])
+    })
+
+    describe('when the slashing multiplier is reset after reset period', async () => {
+      it('should return to default 1.0', async () => {
+        await validators.halveSlashingMultiplier(group)
```
asaj: Looks like there's some repeated code here, move to the `beforeEach`

packages/protocol/test/governance/validators.ts
lucasege: Was getting an error about not initializing the contract, but now it went away, maybe I was using it by accident earlier. Removed

packages/protocol/contracts/governance/Election.sol
```diff
@@ -431,10 +433,18 @@ contract Election is
       votes.active.total
     );
     FixidityLib.Fraction memory score = FixidityLib.wrap(
-      getValidators().calculateGroupEpochScore(uptimes)
+      validators.calculateGroupEpochScore(uptimes)
+    );
+    FixidityLib.Fraction memory slashingMultiplier = FixidityLib.wrap(
```
asaj, line 447: We call it `penalty` in the contract but `slashingMultiplier` here. I have no strong feelings (a slight preference for `slashingMultiplier`), but let's be sure to be consistent.

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -282,7 +291,7 @@ contract Validators is
     bytes calldata blsPublicKey,
     bytes calldata blsPop
   ) external nonReentrant returns (bool) {
-    address account = getAccounts().signerToAccount(msg.sender);
+    address account = getAccounts().validatorSignerToAccount(msg.sender);
```
asaj, line 294: Can you make sure to document this bug fix in the PR description?

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,57 @@ contract Validators is
       }
     }
   }
+
+  /**
+   * @notice Sets the slashingMultiplierRestPeriod property if called by owner.
+   * @param value New reset period for slashing multiplier.
+   */
+  function setSlashingMultiplierResetPeriod(uint256 value) public nonReentrant onlyOwner {
+    slashingPenaltyResetPeriod = value;
```
asaj: Ditto, inconsistency. I know I'm probably responsible for suggesting `penalty` in the first place. I think we had previously talked about `rewardsMultiplier`, which I don't like and would conflict with a term we use in `EpochRewards.sol`. `slashingMultiplier`, however, is fine IMO. Happy with w/e you choose.

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,57 @@ contract Validators is
...
+    require(isValidatorGroup(account));
+    ValidatorGroup storage group = groups[account];
+    require(
+      now >= group.slashInfo.lastSlashed.add(slashingPenaltyResetPeriod),
+      "`resetSlashingMultiplier` called before resetPeriod expired"
+    );
+    group.slashInfo.penalty = FixidityLib.fixed1();
+  }
+
+  bytes32[] canHalveSlashingMultiplier = [
```
asaj, line 1174: @yorhodes suggested reading the `slasher` whitelist from `LockedGold` instead. Probably something to save for a future PR, since `isSlasher` doesn't exist yet

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1147,57 @@ contract Validators is
...
+    DOWNTIME_SLASHER_REGISTRY_ID,
+    DOUBLE_SIGNING_SLASHER_REGISTRY_ID
+  ];
+
+  /**
+   * @notice Halves the group's slashing multiplier.
+   * @param account The group being slashed.
+   */
+  function halveSlashingMultiplier(address account)
```
asaj, line 1183: Been thinking about this more, and wondering if it makes sense for different slashers to set their own multiplier between 0 and 1, let's discuss more offline

packages/protocol/migrationsConfig.js
```diff
@@ -120,6 +120,7 @@ const DefaultConfig = {
     groupLockedGold: {
       value: '22000000000000000000000', // 22k gold per group
     },
+    slashingPenaltyResetPeriod: 60 * 60 * 24 * 30, // 30 Days
```
asaj: nit: Move up above the parameters for migration 20, just below `maxGroupSize`

packages/protocol/test/governance/validators.ts
```diff
@@ -91,6 +93,9 @@ contract('Validators', (accounts: string[]) => {
     exponent: new BigNumber(5),
     adjustmentSpeed: toFixed(0.25),
   }
+  const validatorSlashParameters = {
```
asaj: Why make this an object?

asaj (approved):

PR #2249, opened 2019-12-13
[Wallet] Fix persisted data loss on iOS

### LINKED ISSUE #2248: iOS wallet looses persisted app state
### Expected Behavior

iOS wallet should not loose the persisted app state.

### Current Behavior

iOS wallet looses the persisted app state (from AsyncStorage).

Steps to reproduce:
1. delete the existing app
1. install the app
1. follow through the instructions and choose to restore an account
1. once you reach the wallet home, manually kill the app (swipe up and flick the app)
1. launch the app again
1. observe that the app asks you to choose your language as if it was the first time it ran

jeanregisser:
### Description

This fixes AsyncStorage data loss on iOS.

We had modules importing `AsyncStorage` from React Native Core and some others from React Native Community. The 2 implementations were essentially competing and causing the problem.

Since both implementations currently write to the same file on disk, the last one to write wins and erases the existing data written by the other.

See https://github.com/react-native-community/async-storage/issues/118#issuecomment-500138053

Though I updated the offending import in Apollo, the fix here also makes sure this problem won't sneak back in if we add a dependency which still imports `AsyncStorage` from React Native Core. This is done by monkey patching React Native so it also returns the community implementation of AsyncStorage.

Pretty scary bug, I'm glad we didn't ship that 😅 

### Tested

Followed the steps to reproduced the issue as described in #2248, and observed the problem doesn't happen anymore.

### Other changes

N/A

### Related issues

- Fixes #2248

### Backwards compatibility

Yes


cmcewen (approved):
Ouch this looks like a rough one. Nice job tracking it down, and apologies for missing this in the original upgrade. Also nice job preventing future breaks

jmrossy (approved):

PR #2254, opened 2019-12-13
Add ts-ignore to fix build

annakaz:
### Description

Add ts-ignore to stop failing build caused by #2061


annakaz:
Note that build still fails due to @ts-ignore not working until with Revert the reverted ts upgrade

cmcewen (approved):

PR #2267, opened 2019-12-16
Governance, downtime and double signing slasher contracts

### LINKED ISSUE #2088: Slasher for double signing
### Expected Behavior
```
DoubleSigningSlasher {

  struct SlashingIncentives {
    // Value of LockedGold to slash from the account.
    uint256 penalty;
    // Value of LockedGold to send to the observer.
    uint256 reward;
  }
  
  SlashingIncentives public slashingIncentives;
  mapping(bytes32 => bool) isSlashed;

  // Require that `reward` is less than `penalty`.
  function setSlashingIncentives(uint256 penalty, uint256 reward) external onlyOwner;

  // Given two RLP encoded blocks, calls into a precompile that requires that
  // the two block hashes are different, have the same height, have a
  // quorum of signatures, and that `signer` was part of the quorum.
  // Returns the block number of the provided blocks.
  function eval(address signer, bytes blockA, bytes blockB) internal returns (uint256 blockNumber);

  // Requires that `eval` returns true and that this evidence has not
  // already been used to slash `signer`.
  // If so, fetches the `account` associated with `signer` and the group that
  // `signer` was a member of during the corresponding epoch.
  // Then, calls `LockedGold.slash` on both the validator and group accounts.
  // Calls `Validators.removeSlashedMember` to remove the validator from its 
  // current group if it is a member of one.
  // Finally, stores that hash(signer, blockNumber) has been slashed.
  function slash(address signer, bytes blockA, bytes blockB, uint256 groupMembershipHistoryIndex, address[] validatorElectionLessers, address[] validatorElectionGreaters, uint256[] validatorElectionIndices, address[] groupElectionLessers, address[] groupElectionGreaters, uint256[] groupElectionIndices) external returns (bool);
}
```

### Current Behavior

No such slasher

### LINKED ISSUE #2089: Slasher for downtime
### Expected Behavior

```
DowntimeSlasher {
  struct SlashingIncentives {
    // Value of LockedGold to slash from the account.
    uint256 penalty;
    // Value of LockedGold to send to the observer.
    uint256 reward;
  }

  uint256 public downtimeRequired;
  SlashingIncentives public slashingIncentives;
  mapping (address => mapping (uint256 => bool)) isSlashed;

  // Require that `reward` is less than `penalty`.
  function setSlashingIncentives(uint256 penalty, uint256 reward) external onlyOwner;

  // Requires that `signer` was the `signerIndex0`th validator signer in
  // the epoch corresponding to `startBlock`.
  // If `startBlock + downtimeRequired` is in a different epoch than 
  // `startBlock`, requires that `signer` was the `signerIndex1`th validator
  // signer in the epoch corresponding to `startBlock + downtimeRequired`.
  // Calls a new precompile that iterates over block headers from 
  // `startBlock` to `startBlock + downtimeRequired` and returns the sum of
  // parentAggregatedSeal.Bitmap[signerIndex[0|1]], i.e. the number of
  // signatures present by `signer`.
  // Returns true if the value returned by the precompiled == 0.
  function eval(address signer, uint256 startBlock, uint256 signerIndex0, uint256 signerIndex1) internal returns (bool);

  // Requires that `eval` returns true and that the account corresponding to
  // `signer` has not already been slashed for downtime for the epoch
  // corresponding to `startBlock`.
  // If so, fetches the `account` associated with `signer` and the group that
  // `signer` was a member of during the corresponding epoch.
  // Then, calls `LockedGold.slash` on both the validator and group accounts.
  // Calls `Validators.removeSlashedMember` to remove the validator from its 
  // current group if it is a member of one.
  // Finally, stores that (account, epochNumber) has been slashed.
  function slash(address signer,uint256 startBlock, uint256 signerIndex0, uint256 signerIndex1, uint256 groupMembershipHistoryIndex, address[] validatorElectionLessers, address[] validatorElectionGreaters, uint256[] validatorElectionIndices, address[] groupElectionLessers, address[] groupElectionGreaters, uint256[] groupElectionIndices) external returns (bool);
}
```

### Current Behavior

No such slasher

### LINKED ISSUE #2090: Slasher for Governance
### Expected Behavior

```
GovernanceSlasher {

  // Maps a slashed address to the amount to be slashed.
  // Note that there is no reward paid when slashing via governance.
  mapping (address => uint256) slashed;

  // Require penalty >= reward.
  // Sets `slashed[account] = (penalty, reward);
  // Only callable by governance.
  function approveSlashing(address account, uint256 penalty, uint256 reward) external onlyOwner;

  // Calls `LockedGold.slash` on `account` if `account` has an entry in
  // `slashed`.
  function slash(address account, address[] electionLessers, address[] electionGreaters, uint256[] electionIndices) external returns (bool);
}
```
### Current Behavior

No such slasher

mrsmkl:
### Description

Added contracts handling governance, downtime and double signing slashing.

### Tested

Added unit tests. I guess there can be better tests when the slashing is added into LockedGold etc.

### Other changes

Added an util for getting lesser and greater pointers.

### Related issues

- Fixes #2090 
- Fixes #2089 
- Fixes #2088


packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,163 @@
...
+    slashingIncentives.penalty = penalty;
+    slashingIncentives.reward = reward;
+    emit SlashingIncentivesSet(penalty, reward);
+  }
+
+  // Given two RLP encoded blocks, calls into a precompile that requires that
+  // the two block hashes are different, have the same height, have a
+  // quorum of signatures, and that `signer` was part of the quorum.
+  // Returns the block number of the provided blocks.
+  function eval(
```
yorhodes: I think ideally the `eval` functions would expose `external view`s so observers can more effectively automate their claims

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,163 @@
...
+    _setSlashingIncentives(_penalty, _reward);
+  }
+
+  // Require that `reward` is less than `penalty`.
+  function setSlashingIncentives(uint256 penalty, uint256 reward) external onlyOwner {
+    _setSlashingIncentives(penalty, reward);
+  }
+
+  // Require that `reward` is less than `penalty`.
+  function _setSlashingIncentives(uint256 penalty, uint256 reward) internal {
```
yorhodes: is separating the `external modifier` functions from the `internal` logic defined in a style guide somewhere? I started seeing this recently but not sure what I prefer

asaj: We've been using it in cases where the external function is `onlyVm()`, so that we can have a testing contract that exposes the function without `onlyVm()`. Haven't read the code so not yet sure if that applies here

asaj: @mrsmkl what's the motivation for doing this here?

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,169 @@
...
+    require(penalty > reward, "Penalty has to be larger than reward");
+    slashingIncentives.penalty = penalty;
+    slashingIncentives.reward = reward;
+    emit SlashingIncentivesSet(penalty, reward);
+  }
+
+  function countSetBits(uint256 v) internal pure returns (uint256) {
```
asaj, line 35: meganit: curious if other folks know this as `hammingWeight`, or if `countSetBits` is more readable

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,169 @@
...
+      acc = acc >> 1;
+    }
+    return res;
+  }
+
+  // Given two RLP encoded blocks, calls into a precompile that requires that
+  // the two block hashes are different, have the same height, have a
+  // quorum of signatures, and that `signer` was part of the quorum.
+  // Returns the block number of the provided blocks.
+  function eval(address signer, uint256 index, uint256 blockNumber, bytes memory blockA) internal {
```
asaj: We talked about this a bit yesterday and the conclusion was that passing two RLP encoded block headers would be better. Pseudocode here:


```
def doubleSigned(signerIndex, headerA, headerB: bytes) -> bool
  assert hashHeader(headerA) != hashHeader(headerB)
  blockNumber: int = getBlockNumberFromHeader(headerA)
  assert getBlockNumberFromHeader(headerB) == blockNumber
  assert signerIndex < numberValidators(blockNumber)
  for header in { headerA, headerB }
    bitmap = getVerifiedSealBitmapFromHeader(header)
    assert countSetBits(evidenceBitmap) >= quorumSize(blockNumber)
    assert bitIsSet(bitmap, signerIndex)
  return true
```

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,169 @@
...
+      slashingIncentives.reward,
+      groupElectionLessers,
+      groupElectionGreaters,
+      groupElectionIndices
+    );
+    getValidators().halveSlashingMultiplier(group);
+    isSlashed[keccak256(abi.encodePacked(validator, blockNumber))] = true;
+  }
+
+  function slash(
```
asaj, line 121: nit: Consider combining into one function, e.g.:

```
function slash(
    address signer,
    uint256 index,
    bytes memory headerA,
    bytes memory headerB.
    uint256 groupMembershipHistoryIndex,
    address[] memory validatorElectionLessers,
    address[] memory validatorElectionGreaters,
    uint256[] memory validatorElectionIndices,
    address[] memory groupElectionLessers,
    address[] memory groupElectionGreaters,
    uint256[] memory groupElectionIndices
  ) public returns (bool) {
    require(!isSlashed[keccak256(abi.encodePacked(signer, headerA, headerB))], "Already slashed");
    require(eval(signer, index, headerA, headerB));
    isSlashed[keccak256(abi.encodePacked(signer, headerA, headerB))] = true
    address validator = getAccounts().signerToAccount(signer);
    address group = getValidators().groupMembershipAtBlock(
      validator,
      blockNumber / getEpochSize(),
      groupMembershipHistoryIndex
    );
    getLockedGold().slash(
      validator,
      slashingIncentives.penalty,
      msg.sender,
      slashingIncentives.reward,
      groupElectionLessers,
      groupElectionGreaters,
      groupElectionIndices
    );
    getLockedGold().slash(
      group,
      slashingIncentives.penalty,
      msg.sender,
      slashingIncentives.reward,
      groupElectionLessers,
      groupElectionGreaters,
      groupElectionIndices
    );
    getValidators().forceDeaffiliateIfValidator(validator);
    getValidators().halveSlashingMultiplier(group);
  }
```

asaj: Also consider minimizing the number of calls to the registry:

```
ILockedGold lockedGold = getLockedGold();
IValidators validators = getValidators();
```

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,173 @@
...
+
+  // Require that `reward` is less than `penalty`.
+  function _setSlashingIncentives(uint256 penalty, uint256 reward) internal {
+    require(penalty > reward, "Penalty has to be larger than reward");
+    slashingIncentives.penalty = penalty;
+    slashingIncentives.reward = reward;
+    emit SlashingIncentivesSet(penalty, reward);
+  }
+
+  uint256 public slashableDowntime;
```
asaj: nit: move this with the other storage variabls

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,173 @@
...
+
+  function setSlashableDowntime(uint256 interval) external onlyOwner {
+    require(
+      slashableDowntime < getEpochSize(),
+      "Slashable downtime must be smaller than epoch size"
+    );
+    slashableDowntime = interval;
+  }
+
+  function getEpoch(uint256 blockNumber) internal view returns (uint256) {
```
asaj: Consider moving this to `UsingPrecompiles` and using it in `getEpochNumber` (possibly rename that to `getCurrentEpochNumber`)

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,173 @@
...
+    return blockNumber / getEpochSize();
+  }
+
+  function isDown(
+    address account,
+    uint256 startSignerIndex,
+    uint256 endSignerIndex,
+    uint256 endBlock
+  ) internal view returns (bool) {
+    require(endBlock < block.number, "end block must be smaller than current block");
```
asaj: nit: IMO providing the `startBlock` is more readable, any reason to use the `endBlock` here?

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,173 @@
...
+  // Requires that `eval` returns true and that the account corresponding to
+  // `signer` has not already been slashed for downtime for the epoch
+  // corresponding to `startBlock`.
+  // If so, fetches the `account` associated with `signer` and the group that
+  // `signer` was a member of during the corresponding epoch.
+  // Then, calls `LockedGold.slash` on both the validator and group accounts.
+  // Calls `Validators.removeSlashedMember` to remove the validator from its
+  // current group if it is a member of one.
+  // Finally, stores that (account, epochNumber) has been slashed.
+  function slash(
```
asaj, line 130: Ditto the suggestion to just have one function

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,173 @@
...
+    address validator,
+    uint256 endBlock,
+    uint256 signerIndex0,
+    uint256 signerIndex1,
+    address[] memory validatorElectionLessers,
+    address[] memory validatorElectionGreaters,
+    uint256[] memory validatorElectionIndices
+  ) internal returns (bool) {
+    require(isDown(validator, signerIndex0, signerIndex1, endBlock), "Wasn't down");
+    uint256 epoch = getEpoch(endBlock);
```
asaj: Again, it feels more intuitive (to me) to use the `startBlock` here

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1132,16 @@ contract Validators is
       }
     }
   }
+
+  function groupMembershipAtBlock(address account, uint256 epochNumber, uint256 index)
```
asaj: FYI: https://github.com/celo-org/celo-monorepo/pull/2178

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1131,4 +1132,16 @@ contract Validators is
...
+    external
+    view
+    returns (address)
+  {
+    Validator storage validator = validators[account];
+    return validator.affiliation;
+  }
+
+  function halveSlashingMultiplier(address group) external {}
```
asaj: FYI: https://github.com/celo-org/celo-monorepo/pull/2239

packages/protocol/contracts/governance/test/MockLockedGold.sol
```diff
@@ -43,4 +43,15 @@ contract MockLockedGold is ILockedGold {
...
+  function slash(
+    address account,
+    uint256 penalty,
+    address reporter,
+    uint256 reward,
+    address[] calldata lessers,
+    address[] calldata greaters,
+    uint256[] calldata indices
+  ) external {
+    decrementNonvotingAccountBalance(account, penalty);
```
asaj: nit: should this reduce `accountTotalLockedGold` instead?

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,217 @@
...
+      // @ts-ignore bytes type
+      blsPublicKey,
+      // @ts-ignore bytes type
+      blsPoP,
+      { from: address }
+    )
+  }
+
+  const registerValidatorGroup = async (group: string) => {
+    await mockLockedGold.setAccountTotalLockedGold(group, groupLockedGoldRequirements.value)
```
asaj: FYI there's a bug here, we should be locking `groupLockedGoldRequirements.value * numMembers`, see https://github.com/celo-org/celo-monorepo/pull/2216 for a fix

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,217 @@
...
+    registry = await Registry.new()
+    validators = await Validators.new()
+    slasher = await DoubleSigningSlasher.new()
+    await accountsInstance.initialize(registry.address)
+    await registry.setAddressFor(CeloContractName.Accounts, accountsInstance.address)
+    await registry.setAddressFor(CeloContractName.Election, mockElection.address)
+    await registry.setAddressFor(CeloContractName.LockedGold, mockLockedGold.address)
+    await registry.setAddressFor(CeloContractName.Validators, validators.address)
+    await registry.setAddressFor(CeloContractName.DoubleSigningSlasher, slasher.address)
+    await registry.setAddressFor(CeloContractName.DowntimeSlasher, accounts[5])
```
asaj: Do we need to register all of these contracts? Same question for other tests

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,217 @@
...
+      assert.equal(res[0].toNumber(), 10000)
+      assert.equal(res[1].toNumber(), 100)
+    })
+  })
+
+  describe('#setSlashingIncentives()', () => {
+    it('can only be set by the owner', async () => {
+      await assertRevert(slasher.setSlashingIncentives(123, 67, { from: nonOwner }))
+    })
+    it('should have set slashing incentives', async () => {
```
asaj, line 84: Add test for event emission, and in other tests

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,217 @@
...
+    await mockLockedGold.incrementNonvotingAccountBalance(validator, 50000)
+    await mockLockedGold.incrementNonvotingAccountBalance(group, 60000)
+  })
+
+  describe('#initialize()', () => {
+    it('should have set the owner', async () => {
+      const owner: string = await slasher.owner()
+      assert.equal(owner, accounts[0])
+    })
+    it('should have set slashing incentives', async () => {
```
asaj, line 67: Add test that it can't be called again, and in other tests

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,217 @@
...
+DoubleSigningSlasher.numberFormat = 'BigNumber'
+
+const HOUR = 60 * 60
+const DAY = 24 * HOUR
+// Hard coded in ganache.
+const EPOCH = 100
+
+contract('DoubleSigningSlasher', (accounts: string[]) => {
+  let accountsInstance: AccountsInstance
+  let validators: ValidatorsTestInstance
```
asaj: Would it be easier to use `MockValidators` for this? Same question for other tests

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,217 @@
...
+      }
+    }
+  }
+
+  const nonOwner = accounts[1]
+  const validator = accounts[1]
+
+  beforeEach(async () => {
+    accountsInstance = await Accounts.new()
+    // Do not register an account for the last address so it can be used as an authorized validator signer.
```
asaj: Where is this done? Same question for other tests

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,217 @@
...
+    beforeEach(async () => {
+      blockNumber = await web3.eth.getBlockNumber()
+      const block = await web3.eth.getBlock(blockNumber)
+      const hash = block.parentHash
+      await slasher.setParentHashFromHeader(blockA, hash)
+      await slasher.setParentHashFromHeader(blockB, '0x34')
+      await slasher.setParentHashFromHeader(blockC, hash)
+      await slasher.setEpochSigner(Math.floor(blockNumber / EPOCH), validatorIndex, validator)
+      await slasher.setEpochSigner(Math.floor(blockNumber / EPOCH), validatorIndex + 1, validator)
+      const bitmap = '0x0000000000000000000000000000000000000000000000000000000000800001'
```
asaj: What is this used for? Can you include a comment explaining what this bitmap represents? (i.e. downtime of validator X, double signing by validator Y)

packages/protocol/test/governance/governance_slasher.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  describe('#initialize()', () => {
+    it('should have set the owner', async () => {
+      const owner: string = await slasher.owner()
+      assert.equal(owner, accounts[0])
+    })
+  })
+
+  describe('#approveSlashing()', () => {
+    it('should set slashable amount', async () => {
```
asaj, line 61: Add a test for when this is called twice

packages/protocol/test/governance/governance_slasher.ts
```diff
@@ -0,0 +1,167 @@
...
+  })
+
+  describe('#slash()', () => {
+    it('fails if there is nothing to slash', async () => {
+      await assertRevert(slasher.slash(validator, [], [], []))
+    })
+    it('decrements gold', async () => {
+      await slasher.approveSlashing(validator, 1000)
+      await slasher.slash(validator, [], [], [])
+      const sum = await mockLockedGold.nonvotingAccountBalance(validator)
```
asaj: Why `sum`?

packages/protocol/test/governance/governance_slasher.ts
```diff
@@ -0,0 +1,167 @@
...
+    it('can only be called by owner', async () => {
+      await assertRevert(slasher.approveSlashing(accounts[2], 1000, { from: nonOwner }))
+    })
+  })
+
+  describe('#slash()', () => {
+    it('fails if there is nothing to slash', async () => {
+      await assertRevert(slasher.slash(validator, [], [], []))
+    })
+    it('decrements gold', async () => {
```
asaj, line 81: Add a test that it resets the approved slashing amount to 0

packages/protocol/test/governance/validators.ts
```diff
@@ -2171,7 +2171,7 @@ contract('Validators', (accounts: string[]) => {
         assert.equal(parsedValidator.affiliation, NULL_ADDRESS)
       })
 
-      it('should succeed when the sender is the governance contract', async () => {
+      it('should succeed when the sender is the governance slasher contract', async () => {
```
asaj: I think we were thinking that this would stay `Governance`, that way someone could be slashed via governance without necessarily being removed from their group

packages/utils/src/collections.test.ts
```diff
@@ -0,0 +1,62 @@
+import BigNumber from 'bignumber.js'
+import { Item, linkedListChanges } from './collections'
+
+describe('Collection', () => {
+  describe('linkedListChanges', () => {
```
asaj, line 63: Test multiple inserts?

packages/utils/src/collections.ts
```diff
@@ -39,3 +41,43 @@ export function intersection<T>(arrays: T[][]): T[] {
...
+export interface Item {
+  address: string
+  value: BigNumber
+}
+
+const nilAddress = '0x0000000000000000000000000000000000000000'
+
+function insert(sortedList: Item[], change: Item) {
+  const oldIdx = sortedList.findIndex((a) => a.address === change.address)
+  sortedList.splice(oldIdx, 1)
```
asaj, line 58: Is splice with `-1` as the first argument a no-op?

packages/utils/src/collections.ts
```diff
@@ -39,3 +41,43 @@ export function intersection<T>(arrays: T[][]): T[] {
...
+// List of sorted from greater to smaller
+
+export interface Item {
+  address: string
+  value: BigNumber
+}
+
+const nilAddress = '0x0000000000000000000000000000000000000000'
+
+function insert(sortedList: Item[], change: Item) {
```
asaj: nit: Is this actually an `upsert`?

packages/utils/src/collections.ts
```diff
@@ -39,3 +41,43 @@ export function intersection<T>(arrays: T[][]): T[] {
...
+  sortedList.splice(oldIdx, 1)
+  const newIdx = sortedList.findIndex((a) => a.value.lt(change.value))
+  if (newIdx === -1) {
+    sortedList.push(change)
+  } else {
+    sortedList.splice(newIdx, 0, change)
+  }
+}
+
+export function linkedListChanges(sortedList: Item[], changeList: Item[]) {
```
asaj: This is super useful! 👏 

Love that it doesn't rely on sorting, since we've hit bugs from using unstable sorting algorithms in the past.

Can you create an issue to replace `findLesserAndGreater` in contractkit with this?

asaj (commented):
Consider adding a test for governance slashing to `protocol/test/common/slashing.ts` using the migrated contracts, take a look at `protocol/test/common/integration.ts` for an example (or feel free to add to that file instead if easier)

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
mrsmkl: popcnt

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
mrsmkl: It keeps hitting the stack limit, let's see what I can do.

packages/protocol/test/governance/double_signing_slasher.ts
mrsmkl: Perhaps some of them can be dropped if `MockValidators` was used.

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,170 @@
...
+contract DoubleSigningSlasher is Ownable, Initializable, UsingRegistry, UsingPrecompiles {
+  struct SlashingIncentives {
+    // Value of LockedGold to slash from the account.
+    uint256 penalty;
+    // Value of LockedGold to send to the observer.
+    uint256 reward;
+  }
+
+  SlashingIncentives public slashingIncentives;
+  mapping(bytes32 => bool) isSlashed;
```
m-chrzan: Add comment what this exactly represents. Not obvious to me what's on the left-hand side of this mapping.

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,170 @@
...
+
+  event SlashingIncentivesSet(uint256 penalty, uint256 reward);
+
+  /** @notice Initializer
```
m-chrzan: Style nit: natspec comment text starts on the next line, see any other contract for how this is usually formatted. Here and throughout the file.

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,170 @@
...
+    initializer
+  {
+    _transferOwnership(msg.sender);
+    setRegistry(registryAddress);
+    setSlashingIncentives(_penalty, _reward);
+  }
+
+  /** @notice Sets slashing incentives.
+   * @param penalty Penalty for the slashed signer.
+   * @param reward Reward that the informer gets.
```
m-chrzan: Nit: "informer" and "observer" used incosistently throghout the file, probably better to pick one.

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
m-chrzan: I've heard the name but would need a quick Wikipedia search to remember what it is ;). `countSetBits` is, to me, immediately obvious, could mention that this is Hamming weight in the natspec?

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,170 @@
...
+      if (acc & 1 == 1) res++;
+      acc = acc >> 1;
+    }
+    return res;
+  }
+
+  /**
+   * @notice Given two RLP encoded blocks, calls into a precompile that requires that
+   * the two block hashes are different, have the same height, have a
+   * quorum of signatures, and that `signer` was part of the quorum.
```
m-chrzan, line 55: Not sure if I'm missing something, but why do both blocks have to have a quorum of signatures?

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,170 @@
...
+   * @param signer The signer to be slashed.
+   * @param index Validator index at the block.
+   * @param blockA First double signed block.
+   * @param blockB Second double signed block.
+   * @return Block number where double signing occured.
```
m-chrzan: Also comment on behavior when blocks are not double signed (I'm guessing it throws?).

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,170 @@
...
+   */
+  function eval(address signer, uint256 index, bytes memory blockA, bytes memory blockB)
```
m-chrzan: Is this function name from some generic slashing interface we're trying to conform to? If not, would be great to get a more meaningful name.

m-chrzan (commented):
Haven't looked through all the code but submitting pending comments.

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
asaj: Cool, sounds like `countSetBits` is more readable

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,183 @@
...
+  function initialize(address registryAddress, uint256 _penalty, uint256 _reward)
+    external
+    initializer
+  {
+    _transferOwnership(msg.sender);
+    setRegistry(registryAddress);
+    setSlashingIncentives(_penalty, _reward);
+  }
+
+  function getEpoch(uint256 blockNumber) internal view returns (uint256) {
```
asaj: Re-flagging https://github.com/celo-org/celo-monorepo/pull/2267#discussion_r358962472 (which I assume you intentionally left unresolved so that we wouldn't forget), especially since it seems to be duplicated in the downtime slasher

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
asaj: It's valid behavior in BFT to sign two blocks at the same height:

Preprepare sent, received by all
Prepares sent by all, received only by validator A, who sends a commit
Round change timeout
Preprepare sent with round change certificate excluding round change message from validator A
Prepares sent by all
Commits sent by all

asaj: Strictly, it's sufficient to check that one block has a quorum and the other has F+1, but checking both have a quorum is a fine place to start

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,183 @@
...
+    uint256 acc = v;
+    for (uint256 i = 0; i < 256; i++) {
+      if (acc & 1 == 1) res++;
+      acc = acc >> 1;
+    }
+    return res;
+  }
+
+  /**
+   * @notice Given two RLP encoded blocks, calls into a precompile that requires that
```
asaj: The description is slightly out of date given the precompiles we chose

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,183 @@
...
+   * @param blockA First double signed block.
+   * @param blockB Second double signed block.
+   * @return Block number where double signing occured.
+   */
+  function eval(address signer, uint256 index, bytes memory blockA, bytes memory blockB)
+    public
+    view
+    returns (uint256)
+  {
+    require(keccak256(blockA) != keccak256(blockB), "Block headers have to be different");
```
asaj: I think this actually isn't sufficient. @jfoutts-celo pointed out that the input data is malleable, i.e. prefixing leading zeros leads to the same header being decoded.

Disallowing leading zeros doesn't work, because attackers can just change the extradata vanity. I think we may need a precompile that gives you the block hash. Wdyt @nategraf?

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,183 @@
...
+    require(
+      signer == validatorSignerAddress(index, blockNumber),
+      "Wasn't a signer with given index"
+    );
+    uint256 mapA = uint256(getVerifiedSealBitmap(blockA));
+    uint256 mapB = uint256(getVerifiedSealBitmap(blockB));
+    require(mapA & (1 << index) != 0, "Didn't sign first block");
+    require(mapB & (1 << index) != 0, "Didn't sign second block");
+    require(
+      countSetBits(mapA) >= (2 * numberValidators(blockNumber)) / 3,
```
asaj: Quorum sizes are a bit more complicated than this:
https://github.com/celo-org/celo-blockchain/blob/803c05715bd12a3c148fa870e208cbdcebcc8cb8/consensus/istanbul/core/testbackend_test.go#L435

We talked about making a precompile to fetch this, I don't feel strongly as long as the on chain value matches the value used in consensus

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  function groupMembershipAtBlock(
+    address validator,
+    uint256 blockNumber,
+    uint256 groupMembershipHistoryIndex
+  ) internal returns (address) {
+    uint256 epoch = getEpoch(blockNumber);
+    require(epoch != 0, "Cannot slash on epoch 0");
+    return
```
asaj: nit: can you include a comment clarifying why we check epoch.sub(1) rather than epoch?

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,183 @@
...
+    bytes memory headerB,
+    uint256 groupMembershipHistoryIndex,
+    address[] memory validatorElectionLessers,
+    address[] memory validatorElectionGreaters,
+    uint256[] memory validatorElectionIndices,
+    address[] memory groupElectionLessers,
+    address[] memory groupElectionGreaters,
+    uint256[] memory groupElectionIndices
+  ) public {
+    require(!isSlashed[keccak256(abi.encodePacked(signer, headerA, headerB))], "Already slashed");
```
asaj: See previous comment about malleability, which would allow someone to be slashed over and over for the same violation

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,183 @@
...
+    uint256 blockNumber = eval(signer, index, headerA, headerB);
+    isSlashed[keccak256(abi.encodePacked(signer, headerA, headerB))] = true;
```
asaj: Is the compiler smart enough to only call `keccak256(abi.encodePacked(signer, headerA, headerB)` once? My guess is no, may be better to save in a variable

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,183 @@
...
+      validator,
+      slashingIncentives.penalty,
+      msg.sender,
+      slashingIncentives.reward,
+      validatorElectionLessers,
+      validatorElectionGreaters,
+      validatorElectionIndices
+    );
+    address group = groupMembershipAtBlock(validator, blockNumber, groupMembershipHistoryIndex);
+    if (group == address(0)) return;
```
asaj: This should never be true, maybe leave a comment?

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,219 @@
...
+    slashableDowntime = interval;
+    emit SlashableDowntimeSet(interval);
+  }
+
+  function getEpoch(uint256 blockNumber) internal view returns (uint256) {
+    uint256 sz = getEpochSize();
+    return blockNumber.add(sz).sub(1) / sz;
+  }
+
+  function getEpochBlock(uint256 blockNumber) internal view returns (uint256) {
```
asaj: I'm not clear on what this function is doing

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,219 @@
...
+    uint256 sz = getEpochSize();
+    return blockNumber.add(sz).sub(1) % sz;
+  }
+
+  function groupMembershipAtBlock(
```
asaj: Duplicated code, consider moving to a contract that the slashers inherit from

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,219 @@
...
+    return (
+      startEpochBlock,
+      endEpochBlock,
+      isSlashed[keccak256(abi.encodePacked(validator, startEpoch))],
+      isSlashed[keccak256(abi.encodePacked(validator, endEpoch))]
+    );
+  }
+*/
+
+  function checkIfAlreadySlashed(address validator, uint256 startBlock) internal {
```
asaj, line 98: I don't think I fully understand what's happening here, I was imagining:


```
require(isSlashed[keccak256(abi.encodePacked(validator, startEpoch))] < startBlock);
isSlashed[keccak256(abi.encodePacked(validator, startEpoch))] = startBlock.add(slashableDowntime);
```

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,219 @@
...
+   * @param validatorElectionGreaters Greater pointers for validator slashing.
+   * @param validatorElectionIndices Vote indices for validator slashing.
+   * @param groupElectionLessers Lesser pointers for group slashing.
+   * @param groupElectionGreaters Greater pointers for group slashing.
+   * @param groupElectionIndices Vote indices for group slashing.
+   */
+  function slash(
+    address validator,
+    uint256 startBlock,
+    uint256 signerIndex0,
```
asaj: Do you think it would be cleaner to pass an array of signer indices?

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1148,6 +1149,16 @@ contract Validators is
     }
   }
 
+  // TODO: stub
+  function groupMembershipInEpoch(address account, uint256, uint256)
```
asaj: The real thing has been merged into master

packages/protocol/contracts/governance/test/SlashingTestUtils.sol
```diff
@@ -0,0 +1,73 @@
+pragma solidity ^0.5.3;
+
+import "../DoubleSigningSlasher.sol";
+
+contract SlashingTestUtils {
```
asaj: Consider "MockUsingPrecompiles" (if you think that's an accurate name) for consistency

packages/protocol/test/common/integration.ts
```diff
@@ -25,6 +26,117 @@ enum VoteValue {
   Yes,
 }
 
+contract('Integration: Governance slashing', (accounts: string[]) => {
```
asaj, line 29: 👏

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,160 @@
...
+          reward: new BigNumber(67),
+        },
+      })
+    })
+  })
+
+  describe('#slash()', () => {
+    const blockNumber = 100
+    const validatorIndex = 5
+    const blockA = ['0x12', '0x12', '0x12']
```
asaj: nit: headerA?

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,160 @@
...
+    const blockB = ['0x13', '0x13', '0x13']
+    const blockC = ['0x11', '0x13', '0x14']
+    beforeEach(async () => {
+      await slasher.setBlockNumber(blockA, blockNumber)
+      await slasher.setBlockNumber(blockB, blockNumber + 1)
+      await slasher.setBlockNumber(blockC, blockNumber)
+      await slasher.setEpochSigner(Math.floor(blockNumber / EPOCH), validatorIndex, validator)
+      await slasher.setEpochSigner(Math.floor(blockNumber / EPOCH), validatorIndex + 1, validator)
```
asaj: Why are we setting `validator` to be the signer at two different indices in the same epoch? Should this be for different epochs?

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -0,0 +1,160 @@
...
+      await slasher.setVerifiedSealBitmap(blockC, bitmap)
+    })
+    it('fails if block numbers do not match', async () => {
+      await assertRevert(
+        slasher.slash(validator, validatorIndex, blockA, blockB, 0, [], [], [], [], [], [])
+      )
+    })
+    it('fails if is not signed at index', async () => {
+      await assertRevert(
+        slasher.slash(validator, validatorIndex + 1, blockA, blockC, 0, [], [], [], [], [], [])
```
asaj: I must be missing something, I would have expected this to pass given the previous line:

```
      await slasher.setEpochSigner(Math.floor(blockNumber / EPOCH), validatorIndex + 1, validator)
```

packages/protocol/test/governance/downtime_slasher.ts
```diff
@@ -0,0 +1,278 @@
...
+    registry = await Registry.new()
+    validators = await MockValidators.new()
+    slasher = await DowntimeSlasher.new()
+    await accountsInstance.initialize(registry.address)
+    await registry.setAddressFor(CeloContractName.Accounts, accountsInstance.address)
+    await registry.setAddressFor(CeloContractName.LockedGold, mockLockedGold.address)
+    await registry.setAddressFor(CeloContractName.Validators, validators.address)
+    await validators.affiliate(group, { from: validator })
+    await validators.affiliate(accounts[3], { from: accounts[4] })
+    await slasher.initialize(registry.address, 10000, 100, 3)
```
asaj: nit: Please assign these magic constants to variables for readability

packages/protocol/test/governance/double_signing_slasher.ts
mrsmkl: `setEpochSigner` sets the elected signer for epoch.

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
mrsmkl: Looks like @jfoutts-celo 's PR already has that precompile?

asaj: Indeed it does! Cool, so we'll need to use that here instead of calling `keccak` 256 on the raw bytes value

packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -141,10 +145,18 @@ contract UsingPrecompiles {
     return success;
   }
 
+  function getEpoch(uint256 blockNumber) public view returns (uint256) {
```
asaj: Is this a dupe of `getEpochNumber`?

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -152,9 +149,11 @@ contract DoubleSigningSlasher is Ownable, Initializable, UsingRegistry, UsingPre
     address[] memory groupElectionGreaters,
     uint256[] memory groupElectionIndices
   ) public {
-    require(!isSlashed[keccak256(abi.encodePacked(signer, headerA, headerB))], "Already slashed");
-    uint256 blockNumber = eval(signer, index, headerA, headerB);
-    isSlashed[keccak256(abi.encodePacked(signer, headerA, headerB))] = true;
+    require(!isSlashed[signer][blockHashFromHeader(headerA)], "Already slashed");
```
asaj: This is definitely an edge case, but if a validator signed blocks A, B, *and* C at the same height, how many times should they be slashed?

In other words, should we:
```
require(!isSlashed[signer][blockHashFromHeader(headerA)] || !isSlashed[signer][blockHashFromHeader(headerB)]);
```

or:
```
require(!isSlashed[signer][blockHashFromHeader(headerA)] && !isSlashed[signer][blockHashFromHeader(headerB)]);
```

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -111,51 +92,25 @@ contract DowntimeSlasher is Ownable, Initializable, UsingRegistry, UsingPrecompi
...
-    );
-    require(
-      isSlashed[keccak256(abi.encodePacked(validator, endEpoch))] < endEpochBlock,
-      "Already slashed"
-    );
-    isSlashed[keccak256(abi.encodePacked(validator, startEpoch))] =
-      startEpochBlock +
-      slashableDowntime;
-    isSlashed[keccak256(abi.encodePacked(validator, endEpoch))] = endEpochBlock + 1;
+    require(isSlashed[validator][startEpoch] < startBlock, "Already slashed");
```
asaj: What does two checks (and two sets) of `isSlashed` get us here, as opposed to just checking `isSlashed[validator][startEpoch] < startBlock` and setting `isSlashed[validator][startEpoch] = endBlock`?

packages/protocol/contracts/common/UsingPrecompiles.sol
mrsmkl: Well `getEpochNumber` returns the current epoch, perhaps I'll change it so that they share code.

packages/protocol/contracts/governance/DowntimeSlasher.sol
mrsmkl: It's to handle the case where for example if epoch length is 100 and slashable downtime is 10, if there is a downtime starting at block 400 (epoch 4) and it lasts until block 410 (11 blocks downtime). In the current version that can only be slashed once, but in the simpler version it could be slashed twice. (But perhaps it would be better to slash twice so that both groups get slashed?)

packages/utils/src/collections.ts
```diff
@@ -39,3 +42,51 @@ export function intersection<T>(arrays: T[][]): T[] {
...
+
+// List of address value pairs sorted from greater to smaller
+
+export interface Item {
+  address: string
+  value: BigNumber
+}
+
+function upsert(sortedList: Item[], change: Item) {
+  const oldIdx = sortedList.findIndex((a) => a.address === change.address)
```
mcortesi: to be safe, you should use `eqAddress`

packages/utils/src/collections.ts
```diff
@@ -39,3 +42,51 @@ export function intersection<T>(arrays: T[][]): T[] {
 
   return res
 }
+
+// List of address value pairs sorted from greater to smaller
+
+export interface Item {
```
mcortesi: Maybe rename it to ListItem

packages/utils/src/collections.ts
```diff
@@ -39,3 +42,51 @@ export function intersection<T>(arrays: T[][]): T[] {
...
+    sortedList.push(change)
+  } else {
+    sortedList.splice(newIdx, 0, change)
+  }
+}
+
+// Warning: sortedList is modified
+function _linkedListChange(sortedList: Item[], change: Item) {
+  upsert(sortedList, change)
+  const idx = sortedList.findIndex((a) => a.address === change.address)
```
mcortesi: same

packages/utils/src/collections.ts
```diff
@@ -39,3 +42,51 @@ export function intersection<T>(arrays: T[][]): T[] {
...
+  if (newIdx === -1) {
+    sortedList.push(change)
+  } else {
+    sortedList.splice(newIdx, 0, change)
+  }
+}
+
+// Warning: sortedList is modified
+function _linkedListChange(sortedList: Item[], change: Item) {
+  upsert(sortedList, change)
```
mcortesi: upsert could return the new new idx; so you don't need to search it twice

packages/utils/src/collections.ts
```diff
@@ -39,3 +42,51 @@ export function intersection<T>(arrays: T[][]): T[] {
...
+  const idx = sortedList.findIndex((a) => a.address === change.address)
+  const greater = idx === 0 ? NULL_ADDRESS : sortedList[idx - 1].address
+  const lesser = idx === sortedList.length - 1 ? NULL_ADDRESS : sortedList[idx + 1].address
+  return { lesser, greater }
+}
+
+export function linkedListChange(sortedList: Item[], change: Item) {
```
mcortesi: i would add an explicit return type to make evident what it is that we are returning

packages/utils/src/collections.ts
mcortesi: Or `AddressListItem`

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,178 @@
...
+   * @notice Test if a validator has been down.
+   */
+  function isDown(
+    address account,
+    uint256 startSignerIndex,
+    uint256 endSignerIndex,
+    uint256 startBlock
+  ) internal view returns (bool) {
+    uint256 endBlock = getEndBlock(startBlock);
+    require(endBlock < block.number, "end block must be smaller than current block");
```
mcortesi: imo, given that the "ParentCommitsSeal" is on the next block, endBlock should be `< block.number -1`

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,178 @@
...
+    require(interval < getEpochSize(), "Slashable downtime must be smaller than epoch size");
+    slashableDowntime = interval;
+    emit SlashableDowntimeSet(interval);
+  }
+
+  /**
+   * @notice Test if a validator has been down.
+   */
+  function isDown(
+    address account,
```
mcortesi: account seems unnecessary here. you can derive the account from the start/end signer indexes

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,178 @@
...
+      "Bad validator index at start block"
+    );
+    require(endSignerIndex < numberValidatorsInSet(endBlock), "Bad validator index at end block");
+    address startSigner = validatorSignerAddressFromSet(startSignerIndex, startBlock);
+    address endSigner = validatorSignerAddressFromSet(endSignerIndex, endBlock);
+    require(account == getAccounts().signerToAccount(startSigner), "Wrong start index");
+    require(account == getAccounts().signerToAccount(endSigner), "Wrong end index");
+    uint256 startEpoch = getEpochNumberOfBlock(startBlock);
+    for (uint256 n = startBlock; n <= endBlock; n++) {
+      uint256 signerIndex = getEpochNumberOfBlock(n) == startEpoch
```
mcortesi: small improvement... if you know when the epoch changes, you can do 2 `for` loops. One from `startBlock` to `endOfEpoch`, and a second from `startOfEpoch` to `endBlock. 
Thus avoiding the a `getEpochNumberOfBlock` on each iter step

packages/protocol/contracts/governance/DowntimeSlasher.sol
mcortesi: I like `startSignerIndex` and `endSignerIndex` better than `0`,`1` which don't really express something

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,178 @@
...
+   * the group should be found. (For start block)
+   * @param validatorElectionLessers Lesser pointers for validator slashing.
+   * @param validatorElectionGreaters Greater pointers for validator slashing.
+   * @param validatorElectionIndices Vote indices for validator slashing.
+   * @param groupElectionLessers Lesser pointers for group slashing.
+   * @param groupElectionGreaters Greater pointers for group slashing.
+   * @param groupElectionIndices Vote indices for group slashing.
+   */
+  function slash(
+    address validator,
```
mcortesi: This is redundant, you can obtain the account from the signer idxs

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -0,0 +1,178 @@
...
+    uint256 groupMembershipHistoryIndex,
+    address[] memory validatorElectionLessers,
+    address[] memory validatorElectionGreaters,
+    uint256[] memory validatorElectionIndices,
+    address[] memory groupElectionLessers,
+    address[] memory groupElectionGreaters,
+    uint256[] memory groupElectionIndices
+  ) public {
+    require(isDown(validator, signerIndex0, signerIndex1, startBlock), "Wasn't down");
+    checkIfAlreadySlashed(validator, startBlock);
```
mcortesi, line 145: i would first do this check which is surely cheaper than `isDown`, so it's a cheaper revert

packages/protocol/contracts/governance/DowntimeSlasher.sol
mcortesi: I think you should be slashed twice. Once for each epoch. If not, it will depend on how you make the slashing calls...

For example,

* Epoch Length 100
* down blocks: 50 to 150
* downtime blocks: 10

if i make a call with startBlock 95, then i would only penalize the validator once, and only penalize the group from the first epoch. As opposed of making two calls: one with startBlock 50 and another with startBlock 101.

So, to be consistent, we should only check the first epoch. So a call with startBlock 95 will penalize epoch 1 only and you can still do another with block 101

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,179 @@
...
+   * @param signer The signer to be slashed.
+   * @param index Validator index at the block.
+   * @param blockA First double signed block.
+   * @param blockB Second double signed block.
+   * @return Block number where double signing occured. Throws if no double signing is detected.
+   */
+  function checkForDoubleSigning(
+    address signer,
+    uint256 index,
+    bytes memory blockA,
```
mcortesi: i wonder if we hit some tx size limit when trying to send two blocks. (see [here](https://ethereum.stackexchange.com/questions/1106/is-there-a-limit-for-transaction-size) ) which is around 40kb. 

but we should definitely check our block sizes vs tx.size, and check the gas cost of all of this

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,179 @@
...
+    return (2 * numberValidatorsInSet(blockNumber) + 2) / 3;
+  }
+
+  /**
+   * @notice Given two RLP encoded blocks, calls into precompiles to require that
+   * the two block hashes are different, have the same height, have a
+   * quorum of signatures, and that `signer` was part of the quorum.
+   * @param signer The signer to be slashed.
+   * @param index Validator index at the block.
+   * @param blockA First double signed block.
```
mcortesi: are this the blocks or the block's headers?.. if header i would suggest a rename

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,179 @@
...
+    bytes memory headerB,
+    uint256 groupMembershipHistoryIndex,
+    address[] memory validatorElectionLessers,
+    address[] memory validatorElectionGreaters,
+    uint256[] memory validatorElectionIndices,
+    address[] memory groupElectionLessers,
+    address[] memory groupElectionGreaters,
+    uint256[] memory groupElectionIndices
+  ) public {
+    require(!isSlashed[signer][hashHeader(headerA)], "Already slashed");
```
mcortesi: is `hashHEader` and expensive operation?
we are using it 3 times for each block. 2 here, and one in `checkForDoubleSigning`

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -0,0 +1,179 @@
...
+   * @notice Given two RLP encoded blocks, calls into precompiles to require that
+   * the two block hashes are different, have the same height, have a
+   * quorum of signatures, and that `signer` was part of the quorum.
+   * @param signer The signer to be slashed.
+   * @param index Validator index at the block.
+   * @param blockA First double signed block.
+   * @param blockB Second double signed block.
+   * @return Block number where double signing occured. Throws if no double signing is detected.
+   */
+  function checkForDoubleSigning(
```
mcortesi, line 62: Not sure if relevant.... but given 2 blocks on which I, the signer, had sign. Could they be from 2 different networks? like alfajores & baklave? shouldn't we check that?

mcortesi (commented):
Left some comments

packages/protocol/contracts/governance/DowntimeSlasher.sol
mrsmkl: I think it's clearer to gave just one loop, but I now store the epoch size to save gas.

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
mrsmkl: Yeah, they are just headers so we don't have to worry about gas limit.

packages/protocol/contracts/governance/DowntimeSlasher.sol
mrsmkl: In this design it will depend on the order of calls anyway, so in your example, you'd call it 10 times with start blocks 50, 60, etc.
Changing the confusing name `isSlashed` to `lastSlashedBlock`.

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
mcortesi: As @jfoutts-celo mentioned on Slack, this is already checked on `getVerifiedSealBitmapFromHeader`. Since it verifies that both block have the same parent hash and that the hash belongs to the current blockchain

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
mrsmkl: Changed it so that it now only does it twice...

mcortesi (approved):

PR #2277, opened 2019-12-16
[Wallet] Fix non integer wei issue in exchange

annakaz:
### Description

For certain exchange rates, if a user chooses to enter the exchange makerToken in the takerToken, the makerToken value will be a non integer value in wei. This fixes the error by rounding to the nearest wei

### Tested

Sent an exchange transaction 

### Other changes

Rename standbyTransactionFilter -> transactionFilter as it is used to filter the exchange as well

### Backwards compatibility

Yes


packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -116,14 +116,14 @@ export class ExchangeReview extends React.Component<Props, State> {
...
-    const input = this.state.inputAmount
+    let input = this.state.inputAmount
     if (this.state.makerToken !== this.state.inputToken) {
       const exchangeRate = getRateForMakerToken(
         this.props.exchangeRatePair,
         this.state.makerToken,
         this.state.inputToken
       )
-      getTakerAmount(input, exchangeRate)
+      input = getTakerAmount(input, exchangeRate)
```
annakaz, line 126: Bug in last PR, was not setting value of input

packages/mobile/src/exchange/saga.ts
```diff
@@ -153,8 +152,9 @@ export function* exchangeGoldAndStableTokens(action: ExchangeTokensAction) {
     }
 
     // Ensure the user gets makerAmount at least as good as displayed (rounded to EXCHANGE_DIFFERENCE_TOLERATED)
-    const minimumTakerAmount = getTakerAmount(makerAmount, exchangeRate).minus(
-      EXCHANGE_DIFFERENCE_TOLERATED
+    const minimumTakerAmount = BigNumber.maximum(
+      getTakerAmount(makerAmount, exchangeRate).minus(EXCHANGE_DIFFERENCE_TOLERATED),
+      0
```
annakaz, line 157: Round minimum taker amount to 0 if negative

cmcewen (approved):

PR #2283, opened 2019-12-17
Prevent crash of `yarn run dev:android` when there are iOS files missing

### LINKED ISSUE #2282: [Wallet] Run app without some missing unnecessary files fails
### Expected Behavior

When a dev run `yarn run dev:android` without some iOS files, it should work

### Current Behavior

When a dev run `yarn run dev:android` without some iOS files fails because of missing files

Pedro-vk:
### Description

_Prevent crash of `yarn run dev:android` when there are iOS files missing_

### Tested

_On Linux command line_

### Related issues

- Fixes #2282

### DIFF
--- a/packages/mobile/scripts/run_app.sh
+++ b/packages/mobile/scripts/run_app.sh
@@ -63,8 +63,8 @@ sed -i.bak "s/DEV_RESTORE_NAV_STATE_ON_RELOAD=.*/DEV_RESTORE_NAV_STATE_ON_RELOAD
 # Set Firebase settings in google service config files
 ANDROID_GSERVICES_PATH="./android/app/src/debug/google-services.json"
 IOS_GSERVICES_PATH="./ios/GoogleService-Info.plist"
-sed -i.bak "s/celo-org-mobile-.*firebaseio.com/celo-org-mobile-$NETWORK.firebaseio.com/g" $ANDROID_GSERVICES_PATH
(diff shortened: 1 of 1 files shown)


packages/mobile/scripts/run_app.sh
```diff
@@ -63,8 +63,12 @@ sed -i.bak "s/DEV_RESTORE_NAV_STATE_ON_RELOAD=.*/DEV_RESTORE_NAV_STATE_ON_RELOAD
 # Set Firebase settings in google service config files
 ANDROID_GSERVICES_PATH="./android/app/src/debug/google-services.json"
 IOS_GSERVICES_PATH="./ios/GoogleService-Info.plist"
-sed -i.bak "s/celo-org-mobile-.*firebaseio.com/celo-org-mobile-$NETWORK.firebaseio.com/g" $ANDROID_GSERVICES_PATH
-sed -i.bak "s/celo-org-mobile-.*firebaseio.com/celo-org-mobile-$NETWORK.firebaseio.com/g" $IOS_GSERVICES_PATH
+if [ $PLATFORM = "android" ]; then
```
jmrossy: I would prefer to have them both still be set in case the user is actually running both at the same time, as we sometimes do.
Can you instead check if the files exist and then run sed?

jmrossy (commented):
Thanks for looking into this!

packages/mobile/scripts/run_app.sh
Pedro-vk: Done :+1:

jmrossy (approved):

PR #2285, opened 2019-12-17
Add validator:status --all

asaj:
### Description

Allows you to print out the status of all registered validators, very useful for hunting for folks that are not participating in consensus.

### Tested

By running it

### Other changes

None


packages/cli/src/commands/validator/status.ts
```diff
@@ -8,19 +8,30 @@ import { BaseCommand } from '../../base'
...
 export default class ValidatorStatus extends BaseCommand {
   static description =
     'Show information about whether the validator signer is elected and validating. This command will check that the validator meets the registration requirements, and its signer is currently elected and actively signing blocks.'
 
   static flags = {
     ...BaseCommand.flags,
-    signer: Flags.address({
-      description: 'address of the signer to check if elected and validating',
-      exclusive: ['validator'],
-    }),
```
nategraf: Removing this flag will mean that you can't check the status of a signer that is no longer the authorized signer, but is still part of the elected validator set until the end of the epoch. The intention is to allow people who are running two validator nodes as part of the key rotation process to know when they can safely take the rotated out signer offline.

packages/cli/src/commands/validator/status.ts
```diff
@@ -8,19 +8,30 @@ import { BaseCommand } from '../../base'
...
 
+export const statusTable = {
+  address: {},
+  name: {},
+  signer: {},
+  elected: {},
+  frontRunner: {},
+  proposed: {},
+  signatures: {},
+  signatures: { get: (v: Validator) => parseInt(v.signatures * 100) + '%' },
```
nategraf: Duplicate line?

packages/contractkit/src/wrappers/Validators.ts
```diff
@@ -343,6 +354,35 @@ export class ValidatorsWrapper extends BaseWrapper<Validators> {
     this.contract.methods.affiliate
   )
 
+  async getStatus(
+    validator: Address,
+    blocks: any[],
```
nategraf: I believe the type here can be `Block[]` with `import { Block } from 'web3/eth/types'`
https://github.com/ethereum/web3.js/blob/c0e727eac7c16826c39fd43a03a4930b5ff6fff8/packages/web3/types.d.ts#L418

packages/contractkit/src/wrappers/Validators.ts
```diff
@@ -343,6 +354,35 @@ export class ValidatorsWrapper extends BaseWrapper<Validators> {
...
+    const signer = await accounts.getValidatorSigner(validator)
+    const electedIndex = electedSigners.map((a) => eqAddress(a, signer)).indexOf(true)
+    const frontRunnerIndex = frontRunnerSigners.map((a) => eqAddress(a, signer)).indexOf(true)
+    const proposedCount = blocks.filter((b) => b.miner === signer).length
+    let signedCount = 0
+    if (electedIndex >= 0) {
+      signedCount = blocks.filter((b) =>
+        bitIsSet(parseBlockExtraData(b.extraData).parentAggregatedSeal.bitmap, electedIndex)
+      ).length
+    }
```
nategraf: If the blocks cross an epoch boundary, this method may incorrectly attribute blocks signed to this validator. (On second look, the same bug exists in `master`)

packages/contractkit/src/wrappers/Validators.ts
```diff
@@ -343,6 +354,35 @@ export class ValidatorsWrapper extends BaseWrapper<Validators> {
     this.contract.methods.affiliate
   )
 
+  async getStatus(
```
nategraf: As written, this method seems a bit bespoke to be included in `contractkit`. It may be better to just leave it in the CLI code and extract it to `contractkit` when and if another user appears.

packages/cli/src/commands/validator/status.ts
asaj: Oops

packages/contractkit/src/wrappers/Validators.ts
asaj: Let's address in a follow-up PR

asaj: Created an issue

asaj: https://github.com/celo-org/celo-monorepo/issues/2292

packages/cli/src/commands/validator/status.ts
asaj: Oh I love that! Sorry for not understanding, added this functionality back

packages/contractkit/src/wrappers/Validators.ts
asaj: Done

packages/cli/src/commands/validator/status.ts
```diff
@@ -30,82 +55,72 @@ export default class ValidatorStatus extends BaseCommand {
...
+    const checker = newCheckBuilder(this)
+    if (res.flags.signer) {
+      signers = [res.flags.signer]
+      checker.isAccount(res.flags.validator).isValidator(res.flags.validator)
+      await checker.runChecks()
+    } else if (res.flags.validator) {
+      const validator = await accounts.signerToAccount(res.flags.signer)
+      signers = [validator]
+      checker.isAccount(validator).isValidator(validator)
+      await checker.runChecks()
```
nategraf, line 106: I think this conditional might be switched because `res.flags.validator` is used when only `res.flags.signer` is defined and vice versa

packages/cli/src/commands/validator/status.ts
```diff
@@ -30,82 +55,72 @@ export default class ValidatorStatus extends BaseCommand {
...
+    const validators = await this.kit.contracts.getValidators()
+    let signers: string[] = []
+    const checker = newCheckBuilder(this)
+    if (res.flags.signer) {
+      signers = [res.flags.signer]
+      checker.isAccount(res.flags.validator).isValidator(res.flags.validator)
+      await checker.runChecks()
+    } else if (res.flags.validator) {
+      const validator = await accounts.signerToAccount(res.flags.signer)
+      signers = [validator]
```
nategraf: Seems like this should be `accountToValidatorSigner` here

packages/cli/src/commands/validator/status.ts
```diff
@@ -4,108 +4,128 @@ import { concurrentMap } from '@celo/utils/lib/async'
...
-        .isSignerOrAccount()
-        .signerMeetsValidatorBalanceRequirements()
-        .signerAccountIsValidator()
+    const accounts = await this.kit.contracts.getAccounts()
+    const validators = await this.kit.contracts.getValidators()
+    let signers: string[] = []
+    const checker = newCheckBuilder(this)
+    if (res.flags.signer) {
+      signers = [res.flags.signer]
+      const validator = await accounts.signerToAccount(res.flags.signer)
```
nategraf, line 104: Making this call before checking if it is a valid account means that if the address is not a signer or validator the response it `Error: Returned values aren't valid, did it run Out of Gas?`. I think we should try to avoid that.

nategraf (approved):
Overall looks good. Thanks for addressing my comments.
When running it I did hit a confusing error, and I've commented on it below.

PR #2290, opened 2019-12-17
[Wallet] Fix slow activity feed tx confirmations

### LINKED ISSUE #1741: [Wallet] Users on 3g network SBAT get tx receipts more reliably

annakaz:
### Description

Currently, the activity feed can be slow to update transactions in the activity from "confirming..." to "confirmed", especially when on a 3g connection. Balances update but the "confirming..." is still there, along with a duplicate "confirmed" transaction. 

Now, the activity feed removes the "confirming..." transactions promptly and never renders "confirming..." and "confirmed" at the same time. 

### Tested

Updated jest tests
Tested on pixel 2, throttling to 3g speeds using Charles, also using normal wifi. Confirmed that transaction confirmations for send, exchange and invite come in at the same time that the balance is updated

### Other changes

Created Transaction type to align Standby and confirmed (from graphQL) transactions
Updated some logging

### Related issues

- Fixes #1741 

### Backwards compatibility

Yes


packages/mobile/src/escrow/saga.ts
```diff
@@ -86,7 +86,7 @@ function* registerStandbyTransaction(id: string, value: string, address: string)
       id,
       type: TransactionTypes.ESCROW_SENT,
       status: TransactionStatus.Pending,
-      value,
+      value: +value,
```
annakaz, line 90: Note that storing StandbyTransaction as `number` instead of `string` does not impact precision, as the confirmed transaction value is stored as a `number` anyway. Did not want to modify graphQL to change both to strings so as not to overlap with @jeanregisser 's local currency work

packages/mobile/src/transactions/reducer.ts
```diff
@@ -61,6 +46,24 @@ type TransferTransactionTypes =
   | TransactionTypes.INVITE_RECEIVED
   | TransactionTypes.NETWORK_FEE
 
+export const isTransferType = (txType: TransactionTypes) => {
```
annakaz, line 49: Couldn't find a cleaner way to check multiple types but lmk if you think of one

packages/mobile/src/apollo/__mockData__.ts
```diff
@@ -1,93 +1,81 @@
...
 import { SENTINEL_INVITE_COMMENT } from 'src/invite/actions'
-import { TransactionTypes } from 'src/transactions/reducer'
+import { TransactionTypes, TransferTransactionTypes } from 'src/transactions/reducer'
 
 export const invitedAddress = '0x1b173'
 
 const exchangeDollar: HomeExchangeFragment = {
-  __typename: EventTypeNames.Exchange,
-  type: 'EXCHANGE' as TransactionTypes,
+  type: 'EXCHANGE' as TransactionTypes.EXCHANGE,
```
cmcewen: Why not just use the enum's actual value? (not sure who did this originally)

packages/mobile/src/apollo/types.ts
```diff
@@ -1,41 +1,51 @@
 import { Query } from 'react-apollo'
-import { TransactionTypes } from 'src/transactions/reducer'
+import { CURRENCY_ENUM } from 'src/geth/consts'
```
cmcewen, line 2: I just tried it and it doesn't work right now, but i think this file originally came from running `yarn run build:gen-graphql-types` with the blockchain api running locally

packages/mobile/src/transactions/ExchangeFeedItem.test.tsx
```diff
@@ -28,8 +27,7 @@ describe('ExchangeFeedItem', () => {
       <Provider store={createMockStore({})}>
         <ExchangeFeedItem
           status={TransactionStatus.Complete}
-          __typename={EventTypeNames.Exchange}
-          type={'EXCHANGE' as TransactionTypes}
+          type={'EXCHANGE' as TransactionTypes.EXCHANGE}
```
cmcewen, line 30: same enum comment

packages/mobile/src/transactions/reducer.ts
cmcewen: sometimes i will make an array here and then just check if the string is contained in the array

cmcewen (commented):
woot woot woot

packages/mobile/src/apollo/__mockData__.ts
annakaz: I think this is a more realistic mock of what would be returned?

packages/mobile/src/transactions/ExchangeFeedItem.test.tsx
annakaz: Same as above

packages/mobile/src/apollo/types.ts
annakaz: Kk are you saying we should avoid modifying this generated file?

packages/mobile/src/transactions/reducer.ts
annakaz: Kk I find the function cleaner than the array so will keep as is

packages/mobile/src/apollo/types.ts
jmrossy: This file *was* originally made that way but most of that auto gen code has been changed over time. I've made modifications here too.

jmrossy: I believe we diverged away from the auto gen types because of the differing shapes of our standby tx and graphql txs

packages/mobile/src/apollo/__mockData__.ts
jmrossy: I think Connors point is that `TransactionTypes.Exchange` === 'EXCHANGE' so why hardcode the value here?
And I think the reason it was originally hard coded in the first place is that the auto-gen typing tool did it this way

packages/mobile/src/apollo/__mockData__.ts
```diff
@@ -1,93 +1,81 @@
...
   value: 0.3,
-  symbol: 'Celo Gold',
+  symbol: 'Celo Gold' as CURRENCY_ENUM,
   timestamp: Date.now(),
   address: '0x423043cca38e75d7913504fedfd1dd4539cc55b3',
   comment: 'FAKE FAKE FAKE',
   hash: '01010',
 }
 const verificationReward: HomeTransferFragment = {
-  __typename: EventTypeNames.Transfer,
```
jmrossy, line 87: Why don't we need these graphQL provided __typename properties anymore? Is it cause @jeanregisser split the home feed and exchange feed queries in bc-api?

packages/utils/src/phoneNumbers.ts
```diff
@@ -153,7 +153,9 @@ export function parsePhoneNumber(
         }
       : null
   } catch (error) {
-    console.debug(`phoneNumbers/parsePhoneNumber/Failed to parse phone number, error: ${error}`)
+    console.debug(
+      `phoneNumbers/parsePhoneNumber/Failed to parse phone number ${phoneNumberRaw}, error: ${error}`
```
jmrossy: We intentionally excluded the phone number for privacy reasons, as the logs get uploaded to firebase

jmrossy (commented):
Thanks for tackling this! Can you clarify how this solves the problem exactly?

packages/mobile/src/transactions/reducer.ts
jmrossy: Wondering if we can remove the duplication with TransferTransactionTypes here:
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/transactions/reducer.ts#L52

packages/mobile/src/apollo/__mockData__.ts
annakaz: Yeah we don't need them since we know the EventType from the transaction type

cmcewen (approved):

annakaz:
> Thanks for tackling this! Can you clarify how this solves the problem exactly?

This solves the problem as the issue was that the "Confirming..." notification was not being removed once the confirmation came in. It was actually not a network issue of the transaction taking a long time to be confirmed- it was a UI issue of the PendingTransaction not going away (though it occurred more often when internet was slow, perhaps bc it took longer to query the notification feed?). After cleaning up `src/transactions/TransactionFeed.tsx` to treat PendingTransactions and Transactions as the same type, we know not to display the PendingTransaction if we see the matching Transaction item

PR #2293, opened 2019-12-18
Wait one more block on selectIssuers to deal with light client failures

cmcewen:
### Description

The selectIssuers tx was failing on light clients pretty often - this ensure the state is updated.

### Tested

Ran locally with the app and was able to verify


jmrossy (approved):
Nice find!!!!

PR #2302, opened 2019-12-19
[Slashing] Allow voting gold to be slashable

### LINKED ISSUE #2083: Make Locked gold voting in validator elections slashable
### Expected Behavior

```
  // Iterates in reverse order over all groups voted for by `account`, 
  // reducing the total amount of voting gold by `value` in total.
  // For each group, if there `value` gold has left to be slashed, reduces
  // the number of pending and active votes for that group by as much as
  // possible without exceeding the total value that needs to be slashed.
  function slashVotes(address account, uint256 value, address[] lessers, address[] greaters, uint256[] indices) external onlyRegisteredContract(LOCKED_GOLD_REGISTRY_ID) returns (bool);
```

### Current Behavior

Not slashable

lucasege:
### Description

Adding function `slashVotes` within `Election.sol` that relies on another new function `revokeVotes` which combines `revokePending` and `revokeActive` in order to only sort internal group lists once. 

Points of focus: 
* Cannot make the contract `onlyRegisteredContract(LOCKED_GOLD_REGISTRY_ID)` because if I override `mockLockedGold`'s address the tests revert because that contract cannot be properly called, but I also can't use `mockLockedGold`'s address as the caller since that address is unknown to truffle/ganache. Any guidance on this would be helpful, there don't appear to be any examples of doing this anywhere else.

### Tested

Added unit tests

### Related issues

- Fixes #2083

### Backwards compatibility

yes

### DIFF
--- a/packages/protocol/contracts/governance/Election.sol
+++ b/packages/protocol/contracts/governance/Election.sol
@@ -302,6 +302,44 @@ contract Election is
     return true;
   }
 
+  function _revokeVotes(
+    address account,
+    address group,
+    uint256 value,
+    address lesser,
+    address greater,
+    uint256 index
+  ) public returns (uint256) {
+    require(group != address(0) && 0 < value, "null group");
+    uint256 remainingValue = value;
+    uint256 pendingVotes = getPendingVotesForGroupByAccount(group, account);
+    if (pendingVotes > 0) {
+      uint256 maxValue = remainingValue;
+      if (pendingVotes < maxValue) {
+        maxValue = pendingVotes;
+      }
+      decrementPendingVotes(group, account, maxValue);
+      remainingValue = remainingValue.sub(maxValue);
+    }
+    uint256 activeVotes = getActiveVotesForGroupByAccount(group, account);
+    if (activeVotes > 0) {
+      uint256 maxValue = remainingValue;
+      if (activeVotes < maxValue) {
+        maxValue = activeVotes;
+      }
+      decrementActiveVotes(group, account, maxValue);
+      remainingValue = remainingValue.sub(maxValue);
+    }
+    uint256 difference = value.sub(remainingValue);
+    decrementTotalVotes(group, difference, lesser, greater);
+    getLockedGold().incrementNonvotingAccountBalance(account, difference);
+    if (getTotalVotesForGroupByAccount(group, account) == 0) {
+      deleteElement(votes.groupsVotedFor[account], group, index);
+    }
+    emit ValidatorGroupVoteRevoked(account, group, difference);
+    return difference;
+  }
+
   /**
    * @notice Revokes `value` active votes for `group`
    * @param group The validator group to revoke votes from.
@@ -858,6 +896,11 @@ contract Election is
     return res;
   }
 
+  struct SlashVotesInfo {
+    address[] groups;
+    uint256 remainingValue;
+  }
+
   // Iterates in reverse order over all groups voted for by `account`,
   // reducing the total amount of voting gold by `value` in total.
   // For each group, if there `value` gold has left to be slashed, reduces
@@ -869,50 +912,21 @@ contract Election is
     address[] calldata lessers,
     address[] calldata greaters,
     uint256[] calldata indices
-  ) external onlyRegisteredContract(LOCKED_GOLD_REGISTRY_ID) returns (bool) {
-    address[] storage groups = votes.groupsVotedFor[account];
-    uint256 remainingValue = value;
-    for (uint256 i = groups.length; i > 0; i = i.sub(1)) {
-      // What is eligible
-      if (votes.total.eligible.contains(groups[i.sub(1)])) {
-        address group = groups[i.sub(1)];
-        uint256 maxRemoveableVotes = remainingValue;
-        uint256 pendingVotes = getPendingVotesForGroupByAccount(group, account);
-        uint256 activeVotes = getActiveVotesForGroupByAccount(group, account);
-        if (pendingVotes < maxRemoveableVotes) {
-          maxRemoveableVotes = pendingVotes;
-        }
-        // No if else since a revert means something bad (bad group or bad storage)
-        require(
-          revokePending(
-            group,
-            maxRemoveableVotes,
-            lessers[i.sub(1)],
-            greaters[i.sub(1)],
-            indices[i.sub(1)]
-          )
-        );
-        remainingValue = remainingValue.sub(maxRemoveableVotes);
-        if (remainingValue == 0) {
-          return true;
-        }
-        maxRemoveableVotes = remainingValue;
-        if (activeVotes < remainingValue) {
-          maxRemoveableVotes = activeVotes;
-        }
-        require(
-          revokeActive(
-            group,
-            maxRemoveableVotes,
-            lessers[i.sub(1)],
-            greaters[i.sub(1)],
-            indices[i.sub(1)]
-          )
-        );
-        remainingValue = remainingValue.sub(maxRemoveableVotes);
-        if (remainingValue == 0) {
-          return true;
-        }
+  ) external nonReentrant returns (bool) {
+    SlashVotesInfo memory info = SlashVotesInfo(votes.groupsVotedFor[account], value);
+    for (uint256 i = info.groups.length; i > 0; i = i.sub(1)) {
+      info.remainingValue = info.remainingValue.sub(
+        _revokeVotes(
+          account,
+          info.groups[i.sub(1)],
+          info.remainingValue,
+          lessers[i.sub(1)],
+          greaters[i.sub(1)],
+          indices[i.sub(1)]
+        )
+      );
+      if (info.remainingValue == 0) {
+        return true;
       }
     }
     return false;
--- a/packages/protocol/contracts/governance/Election.sol
+++ b/packages/protocol/contracts/governance/Election.sol
@@ -857,4 +857,64 @@ contract Election is
     }
     return res;
   }
+
+  // Iterates in reverse order over all groups voted for by `account`,
+  // reducing the total amount of voting gold by `value` in total.
+  // For each group, if there `value` gold has left to be slashed, reduces
+  // the number of pending and active votes for that group by as much as
+  // possible without exceeding the total value that needs to be slashed.
+  function slashVotes(
+    address account,
+    uint256 value,
+    address[] calldata lessers,
+    address[] calldata greaters,
+    uint256[] calldata indices
+  ) external onlyRegisteredContract(LOCKED_GOLD_REGISTRY_ID) returns (bool) {
+    address[] storage groups = votes.groupsVotedFor[account];
+    uint256 remainingValue = value;
+    for (uint256 i = groups.length; i > 0; i = i.sub(1)) {
+      // What is eligible
+      if (votes.total.eligible.contains(groups[i.sub(1)])) {
+        address group = groups[i.sub(1)];
+        uint256 maxRemoveableVotes = remainingValue;
+        uint256 pendingVotes = getPendingVotesForGroupByAccount(group, account);
+        uint256 activeVotes = getActiveVotesForGroupByAccount(group, account);
+        if (pendingVotes < maxRemoveableVotes) {
+          maxRemoveableVotes = pendingVotes;
+        }
+        // No if else since a revert means something bad (bad group or bad storage)
+        require(
+          revokePending(
+            group,
+            maxRemoveableVotes,
+            lessers[i.sub(1)],
+            greaters[i.sub(1)],
+            indices[i.sub(1)]
+          )
+        );
+        remainingValue = remainingValue.sub(maxRemoveableVotes);
+        if (remainingValue == 0) {
+          return true;
+        }
+        maxRemoveableVotes = remainingValue;
+        if (activeVotes < remainingValue) {
+          maxRemoveableVotes = activeVotes;
+        }
+        require(
+          revokeActive(
+            group,
+            maxRemoveableVotes,
+            lessers[i.sub(1)],
+            greaters[i.sub(1)],
+            indices[i.sub(1)]
+          )
+        );
+        remainingValue = remainingValue.sub(maxRemoveableVotes);
+        if (remainingValue == 0) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
 }
--- a/packages/protocol/contracts/governance/Election.sol
+++ b/packages/protocol/contracts/governance/Election.sol
@@ -346,7 +346,7 @@ contract Election is
    *         Fundamentally calls `revokePending` and `revokeActive` but only resorts groups once.
    * @param account The account whose votes to `group` should be decremented.
    * @param group The validator group to decrement votes from.
-   * @param value The number of votes to decrement and revoke.
+   * @param maxValue The maxinum number of votes to decrement and revoke.
    * @param lesser The group receiving fewer votes than the group for which the vote was revoked,
    *               or 0 if that group has the fewest votes of any validator group.
    * @param greater The group receiving more votes than the group for which the vote was revoked,
@@ -362,8 +362,7 @@ contract Election is
     address greater,
     uint256 index
   ) internal returns (uint256) {
-    require(group != address(0) && 0 < value);
-    uint256 remainingValue = value;
+    uint256 remainingValue = maxValue;
     uint256 pendingVotes = getPendingVotesForGroupByAccount(group, account);
     if (pendingVotes > 0) {
       uint256 decrementValue = Math.min(remainingValue, pendingVotes);
@@ -376,7 +375,7 @@ contract Election is
       decrementActiveVotes(group, account, decrementValue);
       remainingValue = remainingValue.sub(decrementValue);
     }
-    uint256 decrementedValue = value.sub(remainingValue);
+    uint256 decrementedValue = maxValue.sub(remainingValue);
     if (decrementedValue > 0) {
       decrementTotalVotes(group, decrementedValue, lesser, greater);
       emit ValidatorGroupVoteRevoked(account, group, decrementedValue);
@@ -392,7 +391,7 @@ contract Election is
    * @param account The address of the account.
    * @return The total number of votes cast by an account.
    */
-  function getTotalVotesByAccount(address account) public view returns (uint256) {
+  function getTotalVotesByAccount(address account) external view returns (uint256) {
     uint256 total = 0;
     address[] memory groups = votes.groupsVotedFor[account];
     for (uint256 i = 0; i < groups.length; i = i.add(1)) {
@@ -917,12 +916,13 @@ contract Election is
    * @notice Reduces the total amount of `account`'s voting gold by `value` by
    *         iterating over all groups voted for by account.
    * @param account Address to revoke votes from.
-   * @param value Amount of votes to revoke.
+   * @param value Maximum amount of votes to revoke.
    * @param lessers The groups receiving fewer votes than the i'th `group`, or 0 if
    *                the i'th `group` has the fewest votes of any validator group.
    * @param greaters The groups receivier more votes than the i'th `group`, or 0 if
    *                the i'th `group` has the most votes of any validator group.
    * @param indices The indices of the i'th group in the account's voting list.
+   * @return Number of votes successfully decremented.
    */
   function forceDecrementVotes(
     address account,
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/governance/Election.sol
```diff
@@ -294,6 +294,44 @@ contract Election is
...
   }
 
+  function _revokeVotes(
+    address account,
+    address group,
+    uint256 value,
+    address lesser,
+    address greater,
+    uint256 index
+  ) public returns (uint256) {
```
mrsmkl: Change to internal?

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +883,40 @@ contract Election is
...
+  // For each group, if there `value` gold has left to be slashed, reduces
+  // the number of pending and active votes for that group by as much as
+  // possible without exceeding the total value that needs to be slashed.
+  function slashVotes(
+    address account,
+    uint256 value,
+    address[] calldata lessers,
+    address[] calldata greaters,
+    uint256[] calldata indices
+  ) external nonReentrant returns (bool) {
```
mrsmkl: I guess here should be the check that only whitelisted contracts can call.

packages/protocol/contracts/governance/Election.sol
lucasege: Yup good call, this was a remnant from testing, fixed.

packages/protocol/contracts/governance/Election.sol
lucasege: I think ideally we want it to be only callable from a fn within locked gold contract, where the fn within locked gold will only allow whitelisted contracts. 
I can't get this to work in the tests currently (described in the PR description) so I'm looking for solutions.

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +885,40 @@ contract Election is
...
+  struct SlashVotesInfo {
+    address[] groups;
+    uint256 remainingValue;
+  }
+
+  // Iterates in reverse order over all groups voted for by `account`,
+  // reducing the total amount of voting gold by `value` in total.
+  // For each group, if there `value` gold has left to be slashed, reduces
+  // the number of pending and active votes for that group by as much as
+  // possible without exceeding the total value that needs to be slashed.
```
lucasege: Need to update to fit style

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +885,40 @@ contract Election is
     }
     return res;
   }
+
+  struct SlashVotesInfo {
+    address[] groups;
+    uint256 remainingValue;
+  }
```
lucasege: Struct is used as a solution to avoid `CompilerError: Stack too deep, try removing local variables.` since I need at least 1 local variable (`remainingValue`)

packages/protocol/contracts/governance/Election.sol
```diff
@@ -294,6 +294,46 @@ contract Election is
...
+    uint256 value,
+    address lesser,
+    address greater,
+    uint256 index
+  ) internal returns (uint256) {
+    require(group != address(0) && 0 < value);
+    uint256 remainingValue = value;
+    uint256 pendingVotes = getPendingVotesForGroupByAccount(group, account);
+    if (pendingVotes > 0) {
+      uint256 maxValue = remainingValue;
```
asaj: nit:
```
uint256 decrementValue = Math.min(remainingValue, pendingVotes);
decrementPendingVotes(group, account, decrementValue);
remainingValue = remainingValue.sub(decrementValue);
```


Ditto below

packages/protocol/contracts/governance/Election.sol
```diff
@@ -294,6 +294,46 @@ contract Election is
...
+      if (pendingVotes < maxValue) {
+        maxValue = pendingVotes;
+      }
+      decrementPendingVotes(group, account, maxValue);
+      remainingValue = remainingValue.sub(maxValue);
+    }
+    uint256 activeVotes = getActiveVotesForGroupByAccount(group, account);
+    if (activeVotes > 0) {
```
asaj: `if (activeVotes > 0 && remainingValue > 0)`?

packages/protocol/contracts/governance/Election.sol
```diff
@@ -294,6 +294,46 @@ contract Election is
...
+      uint256 maxValue = remainingValue;
+      if (activeVotes < maxValue) {
+        maxValue = activeVotes;
+      }
+      decrementActiveVotes(group, account, maxValue);
+      remainingValue = remainingValue.sub(maxValue);
+    }
+    uint256 difference = value.sub(remainingValue);
```
asaj: nit: Consider `decrementedValue` instead of `difference`

packages/protocol/contracts/governance/Election.sol
```diff
@@ -294,6 +294,46 @@ contract Election is
...
+    if (difference > 0) {
+      decrementTotalVotes(group, difference, lesser, greater);
+      getLockedGold().incrementNonvotingAccountBalance(account, difference);
+      emit ValidatorGroupVoteRevoked(account, group, difference);
+    }
+    if (getTotalVotesForGroupByAccount(group, account) == 0) {
```
asaj: This can go inside the previous `if` statement

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +885,40 @@ contract Election is
...
+          lessers[i.sub(1)],
+          greaters[i.sub(1)],
+          indices[i.sub(1)]
+        )
+      );
+      if (info.remainingValue == 0) {
+        return true;
+      }
+    }
+    require(false, "Cannot slash all provided votes");
```
asaj: Should we actually fail here?

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +885,40 @@ contract Election is
...
+  // possible without exceeding the total value that needs to be slashed.
+  function slashVotes(
+    address account,
+    uint256 value,
+    address[] calldata lessers,
+    address[] calldata greaters,
+    uint256[] calldata indices
+  ) external nonReentrant returns (bool) {
+    SlashVotesInfo memory info = SlashVotesInfo(votes.groupsVotedFor[account], value);
+    for (uint256 i = info.groups.length; i > 0; i = i.sub(1)) {
```
asaj, line 944: nit: Include a comment with the rationale for iterating backwards

packages/protocol/contracts/governance/Election.sol
asaj: If so, we should maybe just `require(value > getTotalVotesByAccount(account));` at the beginning

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +885,40 @@ contract Election is
...
+      info.remainingValue = info.remainingValue.sub(
+        _revokeVotes(
+          account,
+          info.groups[i.sub(1)],
+          info.remainingValue,
+          lessers[i.sub(1)],
```
asaj, line 950: Do we need to require that lessers.length, greaters.length, and indices.length all equal groups.length?

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +885,40 @@ contract Election is
...
+  }
+
+  // Iterates in reverse order over all groups voted for by `account`,
+  // reducing the total amount of voting gold by `value` in total.
+  // For each group, if there `value` gold has left to be slashed, reduces
+  // the number of pending and active votes for that group by as much as
+  // possible without exceeding the total value that needs to be slashed.
+  function slashVotes(
+    address account,
+    uint256 value,
```
asaj, line 929: Require value > 0?

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +885,40 @@ contract Election is
...
+    address[] calldata lessers,
+    address[] calldata greaters,
+    uint256[] calldata indices
+  ) external nonReentrant returns (bool) {
```
asaj: +1, should be `onlyRegisteredContract(LOCKED_GOLD_REGISTRY_ID)`

Normally, the way you would test this would be to register one of the ganache accounts as the `LockedGold` contract. The reason that doesn't work here is due to re-entrancy, this function is supposed to be called by `LockedGold`, and in turn calls *into* `LockedGold`. While there's nothing inherently wrong with this, some would consider it a dangerous pattern.

I see two options:
1. Don't rely on re-entrancy. Don't call `incrementNonVotingAccountBalance`, and let `LockedGold` do that based on the return value of `slashVotes` (i.e. have it return a uint256).
2. Keep things re-entrant, and define `_slashVotes` which is internal and does not have the modifier. Then, expose it in `ElectionTest` the same way we do for `onlyVm` functions.


@mrsmkl @lucasege which do you prefer?

packages/protocol/test/governance/election.ts
```diff
@@ -1212,4 +1212,319 @@ contract('Election', (accounts: string[]) => {
...
+          })
+
+          it("should remove the group from the voter's voted set", async () => {
+            assert.deepEqual(await election.getGroupsVotedForByAccount(voter), [])
+          })
+        })
+      })
+    })
+
+    describe('when the account has voted for more than one group equally', () => {
```
asaj, line 1310: There seems to be a decent amount of duplicated code in the following 3 test cases. Would it be possible to define something that parses objects like the following, sets up the test case, and then asserts the expected values?


```
const testCases = [
  {
    votesBefore: [
       {
          group: 0x1          
          pending: 2000
          active: 500
       }, {
          group: 0x2         
          pending: 1000
          active: 0
       }
    ],
    slash: 2700
    votesAfter: [
       {
          group: 0x1          
          pending: 0
          active: 0
       }, {
          group: 0x2         
          pending: 800
          active: 0
       }
    ]
  }]
}  
```

asaj, line 1310: Can you include a test case in which slashing re-orders the election results?

packages/protocol/contracts/governance/Election.sol
lucasege: From your docs on slashing V3, `Iterates in reverse order over all groups voted for by account` - I don't have a great reason for the reverse ordering. I assumed it was in order to slash the most recently voted for groups first, since maybe they are more likely to be inactive. Is there a deeper reason? Or should I just switch the order?

packages/protocol/contracts/governance/Election.sol
lucasege: I think I'd prefer 1, but maybe we also change the name from `slashVotes` to just `decrementVotes` or something that indicates we are only decrementing and not actually sending rewards yet (since increment and sending are done in lockedgold).

packages/protocol/test/governance/election.ts
lucasege: Up to you if you want to delay merging for this, I'll get started on it once I finish addressing reviews for #2317 since it seems lower priority, but still worth adding.

packages/protocol/contracts/governance/Election.sol
asaj: I think it makes sense, groups that were voted for first are more likely to be more highly preferred by the voter

packages/protocol/contracts/governance/Election.sol
asaj: I like `decrementVotes`

packages/protocol/test/governance/election.ts
asaj: Not worth blocking the PR but something I suggest we do after since it will clean up the testing code quite a bit

packages/protocol/contracts/governance/Election.sol
```diff
@@ -327,6 +327,52 @@ contract Election is
...
+   * @param group The validator group to revoke votes from.
+   * @param value The number of votes to revoke.
+   * @param lesser The group receiving fewer votes than the group for which the vote was revoked,
+   *               or 0 if that group has the fewest votes of any validator group.
+   * @param greater The group receiving more votes than the group for which the vote was revoked,
+   *                or 0 if that group has the most votes of any validator group.
+   * @param index The index of the group in the account's voting list.
+   * @return uint256 Number of votes successfully revoked, with a max of `value`.
+   */
+  function _revokeVotes(
```
asaj: +1 to your suggestion to rename to `decrementVotes`

packages/protocol/contracts/governance/Election.sol
```diff
@@ -327,6 +327,52 @@ contract Election is
     return true;
   }
 
+  /**
+   * @notice Revokes `value` pending or active votes for `group`. First revokes all pending votes
```
asaj: Natspec needs updating

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +891,52 @@ contract Election is
...
+   *         iterating over all groups voted for by account.
+   * @param account Address to revoke votes from.
+   * @param value Amount of votes to revoke.
+   * @param lessers The groups receiving fewer votes than the i'th `group`, or 0 if
+   *                the i'th `group` has the fewest votes of any validator group.
+   * @param greaters The groups receivier more votes than the i'th `group`, or 0 if
+   *                the i'th `group` has the most votes of any validator group.
+   * @param indices The indices of the i'th group in the account's voting list.
+   */
+  function forceRevokeVotes(
```
asaj: `forceDecrementVotes`?

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +891,52 @@ contract Election is
     }
     return res;
   }
+
+  struct RevokeVotesInfo {
```
asaj: DecrementVotesInfo? Include in a comment why this is defined

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +891,52 @@ contract Election is
...
+          account,
+          info.groups[i.sub(1)],
+          info.remainingValue,
+          lessers[i.sub(1)],
+          greaters[i.sub(1)],
+          indices[i.sub(1)]
+        )
+      );
+      if (info.remainingValue == 0) {
+        return value;
```
asaj: Should we break here instead and return `value.sub(info.remainingValue)`? Right now nothing is returned outside of the loop

packages/protocol/contracts/governance/Election.sol
```diff
@@ -845,4 +891,52 @@ contract Election is
...
+   * @param indices The indices of the i'th group in the account's voting list.
+   */
+  function forceRevokeVotes(
+    address account,
+    uint256 value,
+    address[] calldata lessers,
+    address[] calldata greaters,
+    uint256[] calldata indices
+  ) external nonReentrant onlyRegisteredContract(LOCKED_GOLD_REGISTRY_ID) returns (uint256) {
+    require(value > 0 && value <= this.getTotalVotesByAccount(account));
```
asaj: We don't usually use `this`

packages/protocol/contracts/governance/Election.sol
asaj: Emit an event?

packages/protocol/test/governance/election.ts
```diff
@@ -1212,4 +1212,386 @@ contract('Election', (accounts: string[]) => {
...
+          await mineBlocks(EPOCH, web3)
+          await election.activate(group)
+          await mineBlocks(EPOCH, web3)
+          await election.activate(group2)
+          await election.vote(group2, value2 / 2, group, NULL_ADDRESS)
+          await election.vote(group, value / 2, NULL_ADDRESS, group2)
+          await registry.setAddressFor(CeloContractName.LockedGold, accounts[2])
+        })
+
+        describe("when we slash 1 more vote than group 1's passive vote total", async () => {
```
asaj: passive?

packages/protocol/contracts/governance/Election.sol
lucasege: An event `ValidatorGroupVoteRevoked` is emitted if any votes are successfully decremented. This is the same event as in `revokePending` and `revokeActive`. I don't think we need an additional event, let me know if you think otherwise or you just didn't see the existing event

packages/protocol/contracts/governance/Election.sol
lucasege: Ok, I'll solve this by changing `getTotalVotesByAccount` to `public` instead of `external` which shouldn't make any noticeable difference I believe. 
The alternatives are to not call this fn (it isn't necessary anyway, you suggested it since it is more clear/efficient than just failing later), or to copy the fn's logic into this fn, which I don't like.
Let me know if `external` -> `public` has any issues, otherwise I'll do it that way.

packages/protocol/contracts/governance/Election.sol
```diff
@@ -857,4 +905,56 @@ contract Election is
...
+   *         iterating over all groups voted for by account.
+   * @param account Address to revoke votes from.
+   * @param value Amount of votes to revoke.
+   * @param lessers The groups receiving fewer votes than the i'th `group`, or 0 if
+   *                the i'th `group` has the fewest votes of any validator group.
+   * @param greaters The groups receivier more votes than the i'th `group`, or 0 if
+   *                the i'th `group` has the most votes of any validator group.
+   * @param indices The indices of the i'th group in the account's voting list.
+   */
+  function forceDecrementVotes(
```
mcortesi, line 927: add a comment stating what it returns.. not clear to me

packages/protocol/contracts/governance/Election.sol
```diff
@@ -857,4 +905,56 @@ contract Election is
...
+    address account,
+    uint256 value,
+    address[] calldata lessers,
+    address[] calldata greaters,
+    uint256[] calldata indices
+  ) external nonReentrant onlyRegisteredContract(LOCKED_GOLD_REGISTRY_ID) returns (uint256) {
+    require(value > 0 && value <= getTotalVotesByAccount(account));
```
mcortesi: getTotalVotesByAccount is an expensive check (iterates over all the voted groups). What if instead we go on an decrementVotes, and as a last step we check that remainingValue is 0. If it's not then there's a problem and we fail.

With this strategy we optimize for the valid calls (it's cheaper); and deoptimze for the invalid call (will fail after doing more work); which i think is a good strategy.

packages/protocol/contracts/governance/Election.sol
```diff
@@ -857,4 +905,56 @@ contract Election is
...
+    DecrementVotesInfo memory info = DecrementVotesInfo(votes.groupsVotedFor[account], value);
+    require(
```
mcortesi, line 936: it seems this is not required; since we might not need to modify ALL the groups, just a few.

So a check can be `lesser == greater == indices` and `lesser <= groups`

packages/protocol/contracts/governance/Election.sol
```diff
@@ -857,4 +905,56 @@ contract Election is
...
+        lessers.length == greaters.length &&
+        greaters.length == indices.length
+    );
+    // Iterate in reverse order to hopefully optimize removing pending votes before active votes
+    // And to attempt to preserve `account`'s earliest votes (assuming earliest = prefered)
+    for (uint256 i = info.groups.length; i > 0; i = i.sub(1)) {
+      info.remainingValue = info.remainingValue.sub(
+        _decrementVotes(
+          account,
+          info.groups[i.sub(1)],
```
mcortesi, line 948: it seems you could avoid doing `i.sub(1)` on every step if you start the for with `i = length -1`

packages/protocol/contracts/governance/Election.sol
```diff
@@ -857,4 +905,56 @@ contract Election is
...
+  struct DecrementVotesInfo {
+    address[] groups;
+    uint256 remainingValue;
+  }
+
+  /**
+   * @notice Reduces the total amount of `account`'s voting gold by `value` by
+   *         iterating over all groups voted for by account.
+   * @param account Address to revoke votes from.
+   * @param value Amount of votes to revoke.
```
mcortesi: It seems that `value` is not the votes to revoke, but the `max` votes to revoke.

packages/protocol/contracts/governance/Election.sol
```diff
@@ -339,12 +339,60 @@ contract Election is
...
   }
 
+  /**
+   * @notice Decrements `value` pending or active votes for `group` from `account`.
+   *         First revokes all pending votes and then, if `value` votes haven't
+   *         been revoked yet, revokes additional active votes.
+   *         Fundamentally calls `revokePending` and `revokeActive` but only resorts groups once.
+   * @param account The account whose votes to `group` should be decremented.
+   * @param group The validator group to decrement votes from.
+   * @param value The number of votes to decrement and revoke.
```
mcortesi: it seems like this is a max also. I would make that clear on the name

packages/protocol/contracts/governance/Election.sol
```diff
@@ -339,12 +339,60 @@ contract Election is
...
+   */
+  function _decrementVotes(
+    address account,
+    address group,
+    uint256 value,
+    address lesser,
+    address greater,
+    uint256 index
+  ) internal returns (uint256) {
+    require(group != address(0) && 0 < value);
```
mcortesi: being an internal function, it seems that the check are not needed..
group is surely != than 0.
And value > 0 should probably be checked on the outer loop

packages/protocol/contracts/governance/Election.sol
mcortesi: The way arrar deletes are implement (switches idx to delete with last idx & decrement arr.size) the above is not actually true. Array orders doesn't really mean "voted first".

So, not sure if this is worth the trouble

packages/protocol/contracts/governance/Election.sol
lucasege: If I do that then I'll never check the 0'th element, since when i = 0 the loop will break. I can't do `i >= 0` since when it decrements on `i = 0` it wraps around to the highest value since we're using `uint`s

mcortesi (approved):

PR #2303, opened 2019-12-19
Added daily limit for reserve spending

### LINKED ISSUE #2038: Limit reserve spender withdrawal amounts
### Expected Behavior

Governable parameter that limits the fraction of the reserve that can be withdrawn by reserve spenders per day.
 
### Current Behavior

No such limit

mrsmkl:
### Description

Just added a literal daily limit. It's calculated with the first spending of the day using the new ratio parameter.

### Tested

Added unit tests.

### Other changes

Added balanceOf to MockGoldToken

### Related issues

- Fixes #2038


packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -77,6 +82,22 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
     emit TobinTaxStalenessThresholdSet(value);
   }
 
+  /**
+   * @notice Set the ratio of reserve that is spendable per day.
+   * @param ratio Spending ratio as unwrapped Fraction.
+   */
+  function setDailySpendingRatio(uint256 ratio) public onlyOwner {
+    spendingRatio = FixidityLib.wrap(ratio);
```
asaj, line 92: Emit an event and require less than or equal to 1

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -65,6 +69,7 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
     _transferOwnership(msg.sender);
     setRegistry(registryAddress);
     setTobinTaxStalenessThreshold(_tobinTaxStalenessThreshold);
+    spendingRatio = FixidityLib.fixed1();
```
asaj: Can we pass this as an argument and set in the migrations config?

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -204,6 +225,14 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
    */
   function transferGold(address to, uint256 value) external returns (bool) {
     require(isSpender[msg.sender], "sender not allowed to transfer Reserve funds");
+    uint256 currentDay = now / 1 days;
+    if (currentDay > lastSpendingDay) {
+      uint256 balance = getGoldToken().balanceOf(address(this));
```
asaj: Use `getReserveGoldBalance()` here

packages/protocol/test/stability/reserve.ts
```diff
@@ -67,6 +68,16 @@ contract('Reserve', (accounts: string[]) => {
     })
   })
 
+  describe('#setDailySpendingRatio()', async () => {
```
asaj, line 74: Check for event emission

asaj, line 74: Check that it can't be called with ratio > 1

packages/protocol/test/stability/reserve.ts
```diff
@@ -414,7 +439,7 @@ contract('Reserve', (accounts: string[]) => {
     })
   })
 
-  describe.only('#setAssetAllocations', () => {
```
asaj: Whoops

asaj (commented):
This looks great

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -37,6 +37,10 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
   bytes32[] public assetAllocationSymbols;
   uint256[] public assetAllocationWeights;
 
+  uint256 private lastSpendingDay;
```
mcortesi: `lastSpendingDay` & `spendingLimit` should be public, since a caller might want to check if the transfer will fail beforehand.

`spendingRatio` can be private, since we have a getter, which unwraps it.

mcortesi (approved):
Left a comment.. but approved! Thanks

PR #2304, opened 2019-12-19
Prevent the verification of a verified number and address

### LINKED ISSUE #2217: Verify Phone number flow is trigger even when user is signup with same backup key and phone number.
**Frequency:** 100%
**App version:** Staging build v1.5.2
**Repro on:** Samsung Galaxy Note 5 (7.0), Samsung Galaxy s5 (6.0.1)  

**Pre-condition:** Users should signup with the same phone number backup key. 

**Repro Steps:** 
1) Launch the app.   
2) Enter a phone number. 
3) Enter the security PIN. 
4) Click on Restore backup key. 
5) Paste the same backup key for the phone number. 
6) Tap on the Restore button. 
7) Skip the contact flow. 
8) Observed Verify phone number flow is trigger. 

**Investigation:** 
The user did not get the three verification and after a few seconds “Congratulations, You are verified” message is shown. 
**Impact:** User may get frustrated as user has already verified that phone and used a backup of that phone number. 
**Current Behavior:** Verify Phone number flow is a trigger
**Expected Behavior:** Verify Phone number flow should not be trigger if the user has already verified that phone number. 
**Attachment:** [Android_VerifiedPhoneNumber.mp4 ](https://drive.google.com/file/d/1amZfEnvXTXjC7LF9q3BzC39LjiWg7OXE/view?usp=sharing)(Bug observed at 1: 01 sec)

Pedro-vk:
### Description

_Prevent the verification of a verified number and address_

### Tested

_Tested on Android device_

### Related issues

- Fixes #2217

### DIFF
--- a/packages/mobile/src/identity/verification.ts
+++ b/packages/mobile/src/identity/verification.ts
@@ -27,6 +27,7 @@ import {
   InputAttestationCodeAction,
   ReceiveAttestationMessageAction,
   resetVerification,
+  setHasSeenVerificationNux,
   setVerificationStatus,
 } from 'src/identity/actions'
 import { acceptedAttestationCodesSelector, attestationCodesSelector } from 'src/identity/reducer'
@@ -61,6 +62,30 @@ export enum CodeInputType {
 export interface AttestationCode {
   code: string
   issuer: string
+}
+
+export function* checkVerification() {
+  const attestationsWrapper: AttestationsWrapper = yield call([
+    contractKit.contracts,
+    contractKit.contracts.getAttestations,
+  ])
+
+  const account: string = yield call(getConnectedUnlockedAccount)
+  const e164Number: string = yield select(e164NumberSelector)
+
+  const status: AttestationsStatus = yield call(
+    getAttestationsStatus,
+    attestationsWrapper,
+    account,
+    e164Number
+  )
+
+  if (status.isVerified) {
+    yield put(setNumberVerified(true))
+    yield put(setHasSeenVerificationNux(true))
+    return true
+  }
+  return false
 }
 
 export function* startVerification() {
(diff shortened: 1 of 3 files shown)


packages/mobile/src/identity/verification.ts
```diff
@@ -63,6 +64,30 @@ export interface AttestationCode {
...
+
+  const status: AttestationsStatus = yield call(
+    getAttestationsStatus,
+    attestationsWrapper,
+    account,
+    e164Number
+  )
+
+  if (status.isVerified) {
+    yield put(setVerificationStatus(VerificationStatus.Done))
```
jmrossy: Let's not set this. It's meant to track status updates through the verification process but the user won't be doing that process.

packages/mobile/src/import/saga.ts
```diff
@@ -76,6 +77,9 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
     yield put(refreshAllBalances())
     navigate(Screens.ImportContacts)
     yield put(importBackupPhraseSuccess())
+
+    // Check if the account was verified
+    yield call(checkVerification)
```
jmrossy: This must be done before the navigate call above or users could click skip before the saga is done

jmrossy (commented):
Overall looks good! Just some small things to fix

packages/mobile/src/import/saga.ts
```diff
@@ -74,7 +75,12 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
     // Set redeem invite complete so user isn't brought back into nux flow
     yield put(redeemInviteSuccess())
     yield put(refreshAllBalances())
-    navigate(Screens.VerificationEducationScreen)
+
+    // Check if the account was verified
+    const isVerified = yield call(checkVerification)
+
+    navigate(isVerified ? Screens.ExchangeHomeScreen : Screens.VerificationEducationScreen)
```
jmrossy: You should navigate to WalletHome, not ExchangeHome

Pedro-vk: Done!

jmrossy (approved):

PR #2317, opened 2019-12-19
[Slashing] Slash locked gold

### LINKED ISSUE #2084: Make Locked Gold slashable
### Expected Behavior

```
contract LockedGold {
  mapping (address=>bool) public isSlasher;

  function addSlasher(address slasher) external onlyOwner;

  function removeSlasher(address slasher) external onlyOwner;

  // Slashes the account by reducing its nonvoting locked gold by `penalty`.
  // If there is not enough nonvoting locked gold to slash, calls into
  // `Election.slashVotes` to slash the remaining gold.
  // Finally, sends `reward` Gold to the reporter, and `penalty-reward` Gold
  // to the Community Fund.
  function slash(address account, uint256 penalty, address reporter, uint256 reward, address[] lessers, address[] greaters, uint256[] indices) external onlySlasher;
}
```

### Current Behavior

No slashing

lucasege:
### Description

Adds `slash` function to LockedGold to allow slashing contracts to slash an account's gold. 
Adds `isSlasher` mapping to allow addition of registered slashing addresses for permissioned use of functions. 

### Tested

Added unit tests.

### Related issues 

- Fixes #2084 

### Driveby changes

Changed `MockGovernance` to be payable to allow testing of sending to community fund.
Changed Election interface and mock to allow `LockedGold` to test slashing voting gold.
Adds governance slasher as registered slasher in migrations (this will be expanded upon in PR #2347

### DIFF
--- a/packages/protocol/contracts/governance/LockedGold.sol
+++ b/packages/protocol/contracts/governance/LockedGold.sol
@@ -319,9 +319,8 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
       _incrementNonvotingAccountBalance(reporter, reward);
     }
     address communityFund = registry.getAddressForOrDie(GOVERNANCE_REGISTRY_ID);
-    // address payable addr3 = address(uint160(communityFund));
-    // addr3.transfer(penalty.sub(reward));
-    getGoldToken().transfer(communityFund, penalty.sub(reward));
+    address payable communityFundPayable = address(uint160(communityFund));
+    communityFundPayable.transfer(penalty.sub(reward));
     emit AccountSlashed(account, penalty, reporter, reward);
   }
 }
--- a/packages/protocol/contracts/governance/LockedGold.sol
+++ b/packages/protocol/contracts/governance/LockedGold.sol
@@ -31,6 +31,13 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
   }
 
   mapping(address => Balances) private balances;
+  mapping(address => bool) public isSlasher;
+
+  modifier onlySlasher() {
+    require(isSlasher[msg.sender], "Caller must be registered slasher");
+    _;
+  }
+
   uint256 public totalNonvoting;
   uint256 public unlockingPeriod;
 
@@ -238,6 +245,39 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
     list.length = lastIndex;
   }
 
+  /**
+   * @notice Adds `slasher` to whitelist of approved slashing addresses.
+   * @param slasher Address to whitelist.
+   */
+  function addSlasher(address slasher) external onlyOwner {
+    require(slasher != address(0));
+    isSlasher[slasher] = true;
+  }
+
+  /**
+   * @notice Removes `slasher` from whitelist of approved slashing addresses.
+   * @param slasher Address to remove from whitelist.
+   */
+  function removeSlasher(address slasher) external onlyOwner {
+    require(slasher != address(0));
+    isSlasher[slasher] = false;
+  }
+
+  /**
+   * @notice Slashes `account` by reducing its nonvoting locked gold by `penalty`.
+   *         If there is not enough nonvoting locked gold to slash, calls into
+   *         `Election.slashVotes` to slash the remaining gold. Also sends `reward`
+   *         gold to the reporter, and penalty-reward to the Community Fund.
+   * @param account Address of account being slashed.
+   * @param penalty Amount to slash account.
+   * @param reporter Address of account reporting the slasher.
+   * @param reward Reward to give reporter.
+   * @param lessers The groups receiving fewer votes than i'th group, or 0 if the i'th group has
+   *                the fewest votes of any validator group.
+   * @param greaters The groups receiving more votes than the i'th group, or 0 if the i'th group
+   *                 has the most votes of any validator group.
+   * @param indices The indices of the i'th group in `account`'s voting list.
+   */
   function slash(
     address account,
     uint256 penalty,
@@ -246,5 +286,20 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
     address[] calldata lessers,
     address[] calldata greaters,
     uint256[] calldata indices
-  ) external {}
+  ) external onlySlasher {
+    require(
+      getAccountTotalLockedGold(account) >= penalty,
+      "trying to slash more gold than is locked"
+    );
+    uint256 nonvotingBalance = balances[account].nonvoting;
+    // If not enough nonvoting, revoke the difference
+    if (nonvotingBalance < penalty) {
+      uint256 difference = penalty.sub(nonvotingBalance);
+      getElection().slashVotes(account, difference, lessers, greaters, indices);
+    }
+    _decrementNonvotingAccountBalance(account, penalty);
+    _incrementNonvotingAccountBalance(reporter, reward);
+    // TODO(lucas): How to send to community fund
+    // address(uint160(registry.getAddressForOrDie(GOVERNANCE_REGISTRY_ID))).transfer(penalty - reward);
+  }
 }
--- a/packages/protocol/contracts/governance/LockedGold.sol
+++ b/packages/protocol/contracts/governance/LockedGold.sol
@@ -1,5 +1,6 @@
 pragma solidity ^0.5.3;
 
+import "openzeppelin-solidity/contracts/math/Math.sol";
 import "openzeppelin-solidity/contracts/utils/ReentrancyGuard.sol";
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
@@ -276,8 +277,9 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
   /**
    * @notice Slashes `account` by reducing its nonvoting locked gold by `penalty`.
    *         If there is not enough nonvoting locked gold to slash, calls into
-   *         `Election.slashVotes` to slash the remaining gold. Also sends `reward`
-   *         gold to the reporter, and penalty-reward to the Community Fund.
+   *         `Election.slashVotes` to slash the remaining gold. If `account` does not have
+   *         `penalty` worth of locked gold, slashes `account`'s total locked gold.
+   *         Also sends `reward` gold to the reporter, and penalty-reward to the Community Fund.
    * @param account Address of account being slashed.
    * @param penalty Amount to slash account.
    * @param reporter Address of account reporting the slasher.
@@ -287,6 +289,7 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
    * @param greaters The groups receiving more votes than the i'th group, or 0 if the i'th group
    *                 has the most votes of any validator group.
    * @param indices The indices of the i'th group in `account`'s voting list.
+   * @dev Fails if `reward` is greater than `account`'s total locked gold.
    */
   function slash(
     address account,
@@ -297,17 +300,14 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
     address[] calldata greaters,
     uint256[] calldata indices
   ) external onlySlasher {
-    require(
-      getAccountTotalLockedGold(account) >= penalty,
-      "trying to slash more gold than is locked"
-    );
-    require(penalty >= reward, "reward cannot exceed penalty.");
+    uint256 maxSlash = Math.min(penalty, getAccountTotalLockedGold(account));
+    require(maxSlash >= reward, "reward cannot exceed penalty.");
     {
       uint256 nonvotingBalance = balances[account].nonvoting;
       uint256 difference = 0;
       // If not enough nonvoting, revoke the difference
-      if (nonvotingBalance < penalty) {
-        difference = penalty.sub(nonvotingBalance);
+      if (nonvotingBalance < maxSlash) {
+        difference = maxSlash.sub(nonvotingBalance);
         require(
           getElection().forceDecrementVotes(account, difference, lessers, greaters, indices) ==
             difference,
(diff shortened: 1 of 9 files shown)


packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +243,62 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
     list[index] = list[lastIndex];
     list.length = lastIndex;
   }
+
+  /**
+   * @notice Adds `slasher` to whitelist of approved slashing addresses.
+   * @param slasher Address to whitelist.
+   */
+  function addSlasher(address slasher) external onlyOwner {
```
asaj, line 261: Emit an event here and in `removeSlasher`

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +243,62 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
+    require(slasher != address(0));
+    isSlasher[slasher] = true;
+  }
+
+  /**
+   * @notice Removes `slasher` from whitelist of approved slashing addresses.
+   * @param slasher Address to remove from whitelist.
+   */
+  function removeSlasher(address slasher) external onlyOwner {
+    require(slasher != address(0));
```
asaj: remove this and instead require isSlasher[slasher]

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +243,62 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
+    );
+    uint256 nonvotingBalance = balances[account].nonvoting;
+    // If not enough nonvoting, revoke the difference
+    if (nonvotingBalance < penalty) {
+      uint256 difference = penalty.sub(nonvotingBalance);
+      getElection().slashVotes(account, difference, lessers, greaters, indices);
+    }
+    _decrementNonvotingAccountBalance(account, penalty);
+    _incrementNonvotingAccountBalance(reporter, reward);
+    // TODO(lucas): How to send to community fund
```
asaj:
```
address communityFund = registry.getAddressForOrDie(GOVERNANCE_REGISTRY_ID);
// Option A (recommended)
communityFund.transfer(penalty.sub(reward));

// Option B
getGoldToken.transfer(communityFund, penalty.sub(reward))
```

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +243,62 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
+   */
+  function slash(
+    address account,
+    uint256 penalty,
+    address reporter,
+    uint256 reward,
+    address[] calldata lessers,
+    address[] calldata greaters,
+    uint256[] calldata indices
+  ) external onlySlasher {
```
asaj, line 302: require penalty >= reward

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +243,62 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
+    require(
+      getAccountTotalLockedGold(account) >= penalty,
+      "trying to slash more gold than is locked"
+    );
+    uint256 nonvotingBalance = balances[account].nonvoting;
+    // If not enough nonvoting, revoke the difference
+    if (nonvotingBalance < penalty) {
+      uint256 difference = penalty.sub(nonvotingBalance);
+      getElection().slashVotes(account, difference, lessers, greaters, indices);
```
asaj: If this is going to increment the nonvoting account balance, consider calling the function `forceRevokeVotes` or something along those lines

packages/protocol/test/governance/lockedgold.ts
```diff
@@ -412,4 +413,37 @@ contract('LockedGold', (accounts: string[]) => {
       })
     })
   })
+
+  describe('#slash', () => {
```
asaj: Add tests for add/remove slasher

packages/protocol/test/governance/lockedgold.ts
```diff
@@ -412,4 +413,37 @@ contract('LockedGold', (accounts: string[]) => {
...
+    const value = 1000
+    const reporter = accounts[2]
+
+    beforeEach(async () => {
+      // @ts-ignore: TODO(mcortesi) fix typings for TransactionDetails
+      await lockedGold.lock({ value })
+      await lockedGold.addSlasher(account)
+    })
+
+    describe('when the account is slashed for all of its locked gold', () => {
```
asaj, line 451: Add test that this can only be called by a slasher

packages/protocol/test/governance/lockedgold.ts
```diff
@@ -412,4 +413,37 @@ contract('LockedGold', (accounts: string[]) => {
...
+          [NULL_ADDRESS],
+          [0]
+        )
+      })
+
+      it("should reduce account's locked gold balance to 0", async () => {
+        assertEqualBN(await lockedGold.getAccountNonvotingLockedGold(account), 0)
+      })
+
+      it("should increase the reporter's locked gold", async () => {
```
asaj, line 473: Add test that the community fund receives gold

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +243,62 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
+   * @param penalty Amount to slash account.
+   * @param reporter Address of account reporting the slasher.
+   * @param reward Reward to give reporter.
+   * @param lessers The groups receiving fewer votes than i'th group, or 0 if the i'th group has
+   *                the fewest votes of any validator group.
+   * @param greaters The groups receiving more votes than the i'th group, or 0 if the i'th group
+   *                 has the most votes of any validator group.
+   * @param indices The indices of the i'th group in `account`'s voting list.
+   */
+  function slash(
```
asaj, line 294: Emit an event

packages/protocol/test/governance/lockedgold.ts
asaj, line 451: Add a test for when some of the gold that gets slashed is voting

packages/protocol/contracts/governance/LockedGold.sol
lucasege: should I emit one event `AccountSlashed(account, penalty, reporter, reward)` or two events to split account/penalty and reporter/reward. Or is the reporter/reward not as important

packages/protocol/contracts/governance/LockedGold.sol
lucasege: had to go with option B because of conversion from address to address payable in option A

packages/protocol/test/governance/lockedgold.ts
lucasege: How can I test the governance contract's balance? Traditional methods like `web3.eth.getBalance` gives me `0`, but using our contracts gives me issues with addresses not being registered.

packages/protocol/test/governance/lockedgold.ts
lucasege: In the middle of this now, but my reasoning for not including this originally was because this seems like a better fit for e2e testing.
The slowdown right now is figuring out how to get the necessary functions out of the `election` contract within the `lockedgold` tests.

mrsmkl:
I'll try to add a test into `test/common/integration.ts`

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -31,13 +31,23 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
   uint256 public totalNonvoting;
   uint256 public unlockingPeriod;
 
   event UnlockingPeriodSet(uint256 period);
   event GoldLocked(address indexed account, uint256 value);
   event GoldUnlocked(address indexed account, uint256 value, uint256 available);
   event GoldWithdrawn(address indexed account, uint256 value);
+  event SlasherWhitelistAdded(address slasher);
+  event SlasherWhitelistRemoved(address slasher);
+  event AccountSlashed(address slashed, uint256 penalty, address reporter, uint256 reward);
```
mrsmkl: Probably best to make the addresses topics.

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +246,73 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
+    isSlasher[slasher] = true;
+    emit SlasherWhitelistAdded(slasher);
+  }
+
+  /**
+   * @notice Removes `slasher` from whitelist of approved slashing addresses.
+   * @param slasher Address to remove from whitelist.
+   */
+  function removeSlasher(address slasher) external onlyOwner {
+    require(isSlasher[slasher]);
```
mrsmkl: Let's add error messages for all requires.

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +246,73 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
     list[index] = list[lastIndex];
     list.length = lastIndex;
   }
+
+  /**
+   * @notice Adds `slasher` to whitelist of approved slashing addresses.
+   * @param slasher Address to whitelist.
+   */
+  function addSlasher(address slasher) external onlyOwner {
+    require(slasher != address(0));
```
mrsmkl: Perhaps require that it wasn't already whitelisted.

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -236,4 +246,73 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
+      "trying to slash more gold than is locked"
+    );
+    require(penalty >= reward, "reward cannot exceed penalty.");
+    {
+      uint256 nonvotingBalance = balances[account].nonvoting;
+      uint256 difference = 0;
+      // If not enough nonvoting, revoke the difference
+      if (nonvotingBalance < penalty) {
+        difference = penalty.sub(nonvotingBalance);
+        require(
```
mrsmkl, line 312: What should we do if there's not enough to slash anymore?

mrsmkl:
Did you solve the issue with sending to governance? I guess you just need to make MockGovernance able to receive gold like Governance (fallback function).

packages/protocol/contracts/governance/LockedGold.sol
lucasege: I believe we would want to just slash all of the `account`'s gold, I'm assuming this is handled on the caller side (`Slasher` contract) by passing in for `penalty` `max(normalSlashPenalty, totalGoldOwnedByAccount)` where `normalSlashPenalty` is the default penalty and `totalGoldOwnedByAccount` is the total balance + voting balance for the account.

packages/protocol/test/governance/lockedgold.ts
mrsmkl: The code is `getGoldToken().transfer(communityFund, penalty.sub(reward));`
so I think you can test mock locked gold (just have to change it so that transfer actually does something)

packages/protocol/test/governance/lockedgold.ts
```diff
@@ -412,4 +418,82 @@ contract('LockedGold', (accounts: string[]) => {
...
+        assertEqualBN(await lockedGold.getAccountNonvotingLockedGold(reporter), value / 2)
+      })
+
+      it("should increase the community fund's gold", async () => {
+        assertEqualBN(
+          await lockedGold.getAccountNonvotingLockedGold(MockGovernance.address),
```
mrsmkl: I guess this should be `mockGovernance`

packages/protocol/contracts/governance/LockedGold.sol
jfoutts-celo: Now that the other PRs are in, we can update e.g. https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/SlasherUtil.sol#L81 to do the ` max(normalSlashPenalty, totalGoldOwnedByAccount)`.

mrsmkl: Perhaps it's easier to handle in LockedGold, because the needed balances are stored there.

mrsmkl:
Looks like the integration test broke here. Should be fixed by using the migration from https://github.com/mrsmkl/celo-monorepo/blob/test-slashing2/packages/protocol/migrations/19_governance_slasher.ts

mrsmkl:
Looks like there was some flakiness in the validators test, perhaps because the epoch number was calculated differently.

packages/protocol/contracts/governance/LockedGold.sol
lucasege: Based on this I am changing `LockedGold.slash` to only slash `min(totalLockedGold, penalty)`. A problematic implication is that if `totalLockedGold` is somehow less than the `reward` passed in to slash, the call will revert. However, without doing a bunch of wasteful calculations (adjusting `reward` based on its original proportion or something similar) `slash` doesn't have the right info to change `reward` properly. Just something to keep in mind

packages/protocol/contracts/governance/LockedGold.sol
mcortesi: one event is ok imo

packages/protocol/contracts/governance/LockedGold.sol
mcortesi: I think this is OK for now, but i would follow the discussion on this. 

My concern is that Given the `min()` i will be considered "slashed" and thus can't be slashed again for the same thing, but i wouldn't be really paying the penalty that i'm due... i wonder if i can use that to shield me from penalties somehow.

But to analyze this we need to see the bigger picture...

packages/protocol/test/governance/lockedgold.ts
```diff
@@ -412,4 +421,159 @@ contract('LockedGold', (accounts: string[]) => {
...
+          assertEqualBN(await lockedGold.getAccountNonvotingLockedGold(reporter), reward)
+          assertEqualBN(await lockedGold.getAccountTotalLockedGold(reporter), reward)
+        })
+
+        it("should increase the community fund's gold", async () => {
+          assert.equal(await web3.eth.getBalance(mockGovernance.address), penalty - reward)
+        })
+      })
+
+      describe('when the account is slashed for its whole balance', () => {
```
mcortesi, line 542: do we tests for the case where we are slashed by the `min(balance, penalty)`. we should add those

mcortesi (approved):
left a comment, but approved!

packages/protocol/test/governance/lockedgold.ts
lucasege: done

PR #2324, opened 2019-12-20
1666 precompiles assembly rewrite

### LINKED ISSUE #1666: Reduce assembly usage in UsingPrecompiles
### Expected Behavior

Limited assembly usage, at most to parse output `bytes` into return variables.

## Current Behavior

Lots of assembly

Perseverance:
Fixes #1666


asaj (approved):

packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -25,38 +29,19 @@ contract UsingPrecompiles {
...
+    bytes memory result;
+    (success, result) = FRACTION_MUL_EXP.staticcall(
+      abi.encodePacked(aNumerator, aDenominator, bNumerator, bDenominator, exponent, _decimals)
+    );
+    require(
+      success,
+      "UsingPrecompiles :: fractionMulExp Unsuccessful invocation of fraction exponent"
+    );
+    assembly {
+      returnNumerator := mload(add(result, 32))
```
asaj: Actually, are you sure this is right? Doesn't mload take the start, which in this case would be `result` and not `result + 32`?

asaj (changes requested):

packages/protocol/contracts/common/UsingPrecompiles.sol
Perseverance: @asaj AFAIK, for length >= 32 bytes (which is the case with uint256), bytes use the first 32 bytes as length. This means that we need to "forward" 32 bytes from the beginning, hence the `add(result, 32))`. 
Regardless I double checked it and double tested this via the existing unit tests, and this is the correct setup :)

https://solidity.readthedocs.io/en/v0.5.13/miscellaneous.html#bytes-and-string

mcortesi: @Perseverance is right... 
I've tested in remix, here's the code i've used https://gist.github.com/mcortesi/b35caa87abc0376ce58e7a4df11458a6

My doubt was not about that bytes first 32 bytes are the length, but the difference between making an assembly call or a call through the address... it seems they return things differently.

When assembly, there's no extra length field.. the total length is 0x40 bytes. But when we do the call with address.staticcall the returned bytes has a size 0x60 bytes.

mcortesi (approved):

asaj (approved):

PR #2327, opened 2019-12-20
Voting bot for stake off

### LINKED ISSUE #1477: Voting bot for stake off
### Expected Behavior

We would like a number of instances of a voting bot to run in a testnet. For simplicity, we could extend the load testing client to do this for us (see `celo-monorepo/packages/celotool/src/cmds/geth/simulate_client.ts` )

- [ ] lock up gold initially (maybe in the migrations?)
- [x] wake up at a random interval
- [x] with a given probability, do nothing, otherwise evaluate whether to change vote  
- [x] implement a strategy that picks the group that can receive votes AND has highest recent rewards
- [x] add some degree of randomness to it: with certain probability pick a random group 

@rcroessmann and @asaj will have ideas based on their recent simulation.

### Current Behavior

Rely on external users to vote.

yerdua:
### Description

This implements a voting bot, that runs roughly once an epoch and votes on validator groups. For each voting bot account, it does this process:
1. activates any pending votes (votes cast in the previous epoch)
2. randomly selects which of the voting bots are participating in the current round
3. For each bot, it decides based on the performance of the group they're currently voting for (if any), if they should change their vote, or keep it as is.
4. For each bot that is changing its vote, it chooses which strategy to use to pick a group to vote for: a) weighted-random based on scores of currently-elected validators, b) unweighted-random selection of an unelected group
5. It uses the strategy chosen in step 4 to pick a new group
6. Revokes old vote
7. Casts a vote for that new group

### Tested

I ran the following celotool command, pointed at baklava staging, and confirmed that voting occurred.
```celotool bots auto-vote --celoProvider https://baklavastaging-forno.celo-testnet.org```

I confirmed that the cronjob gets deployed and then voting happens on schedule with this command:
```celotool deploy initial voting-bot -e baklavastaging```

### Other changes

fixed a typo or two

### Related issues

- Fixes #1477

### CODE REVIEW

File: .env.baklavastaging [hunk 1]
```diff
@@ -37,7 +37,7 @@ GETH_BOOTNODE_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/geth-all"
 GETH_BOOTNODE_DOCKER_IMAGE_TAG="a85db20179b503b802430aef7085d35aa761c63d"
 
 CELOTOOL_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/celo-monorepo"
-CELOTOOL_DOCKER_IMAGE_TAG="celotool-53016a800a510cddae73444377bcf366d753367a"
+CELOTOOL_DOCKER_IMAGE_TAG="celotool-99f83b1b5554ae6fb0233b7b0f8a236660cb34cd"
```
yerdua: TODO: This will need to be updated to a tag with a commit that exists in master and contains the new commands.

yerdua:
There will be a followup commit to change the `CELOTOOL_DOCKER_IMAGE_TAG` in `.env`, `.env.baklavastaging`, and `.env.baklava` to use celotool image that is built from the merge commit.

packages/celotool/src/cmds/bots/auto-vote.ts
```diff
@@ -0,0 +1,204 @@
...
+    default: 'http://localhost:8545',
+  })
+}
+
+export const handler = async function simulateVoting(argv: SimulateVotingArgv) {
+  try {
+    const mnemonic = fetchEnv(envVar.MNEMONIC)
+    const numBotAccounts = parseInt(fetchEnv(envVar.VOTING_BOTS), 10)
+
+    const kit: ContractKit = newKit(argv.celoProvider)
```
nambrot: Is this type annotation necessary?

packages/celotool/src/cmds/bots/auto-vote.ts
```diff
@@ -0,0 +1,204 @@
...
+        } catch (error) {
+          console.error(`Failed to vote as ${botAccount}`)
+          console.info(error)
+        }
+      }
+    }
+  } catch (error) {
+    console.error(error)
+  } finally {
+    process.exit(0)
```
nambrot, line 138: Should we really always exit with 0, even if there is an error?

packages/celotool/src/cmds/bots/auto-vote.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    const allBotKeys = getPrivateKeysFor(AccountType.VOTING_BOT, mnemonic, numBotAccounts)
+    await activatePendingVotes(kit, allBotKeys)
+
+    const botKeysVotingThisRound = allBotKeys.filter((_) =>
+      changeVoteProbability.isGreaterThan(Math.random())
+    )
+    console.info(`Voting this time: ${botKeysVotingThisRound.length} of ${numBotAccounts}`)
+
+    if (botKeysVotingThisRound.length > 0) {
```
nambrot: This is a pretty massive method, what do you think about extracting the actual voting from the setup?

packages/celotool/src/cmds/bots/auto-vote.ts
```diff
@@ -0,0 +1,204 @@
...
+  } finally {
+    process.exit(0)
+  }
+}
+
+async function calculateGroupWeights(kit: ContractKit): Promise<Map<string, BigNumber>> {
+  const election = await kit.contracts.getElection()
+  const validators = await kit.contracts.getValidators()
+
+  const groupScores = new Map<string, BigNumber[]>()
```
nambrot: What do you think about using `lodash`'s `groupBy` and `mapValues` for a more concise version of this? 

```typescript
  const validatorSigners = await election.getCurrentValidatorSigners()
  const validatorAccounts = await concurrentMap(
    10,
    validatorSigners,
    validators.getValidatorFromSigner
  )

  const validatorsByGroup = groupBy(validatorAccounts, (validator) => validator.affiliation!)
  const validatorGroupScores = mapValues(validatorsByGroup, (vals) =>
    vals.reduce((a, b) => a.plus(b.score), new BigNumber(0))
  )

  return new Map(Object.entries(validatorGroupScores))

```

packages/celotool/src/cmds/bots/auto-vote.ts
```diff
@@ -0,0 +1,204 @@
...
+  const election = await kit.contracts.getElection()
+
+  for (const key of botKeys) {
+    kit.addAccount(key)
+    const account = ensure0x(getAccountAddressFromPrivateKey(key))
+    if (!(await election.hasActivatablePendingVotes(account))) {
+      try {
+        console.info(`activating votes for ${account}`)
+        const activateTxs = await election.activate(account)
+        for (const tx of activateTxs) {
```
nambrot: Consider that this will execute all txs serially which might not be desired. What do you think about doing them reasonably concurrently?

nambrot (commented):
I'm not too familiar with the requirements, but the incumbency bias seems like a big deal?

packages/celotool/src/cmds/bots/auto-vote.ts
yerdua: nope. removing it.

packages/celotool/src/cmds/bots/auto-vote.ts
yerdua: That looks much better. Thanks!

yerdua: hmm... actually this doesn't totally solve it since what needs to happen in calculating the group score is to find the _average_ score of all the validators in the group. Your suggestion gets the sum of them all. Perhaps there's still a use for `groupBy` and `mapValues` though.

packages/celotool/src/cmds/bots/auto-vote.ts
```diff
@@ -0,0 +1,318 @@
...
+  return groupCapacities
+}
+
+async function calculateGroupScores(kit: ContractKit): Promise<Map<string, BigNumber>> {
+  console.info('Calculating weights of groups based on the scores of elected validators')
+  const election = await kit.contracts.getElection()
+  const validators = await kit.contracts.getValidators()
+
+  const validatorSigners = await election.getCurrentValidatorSigners()
+  const validatorAccounts = (
```
jfoutts-celo, line 221: Could we use `validators.getRegisteredValidators()` here?  Do we want to include previously elected (but currently unelected) validators?

.env.baklava
```diff
@@ -118,3 +118,33 @@ LEADERBOARD_DOCKER_IMAGE_TAG="leaderboard-e84c5086e94193fdef430008bd44129e4151bb
 ATTESTATION_BOT_INITIAL_WAIT_SECONDS=600
 ATTESTATION_BOT_IN_BETWEEN_WAIT_SECONDS=600
 ATTESTATION_BOT_MAX_ATTESTATIONS=90
+
+# Number of gold-holding bots that vote for validator groups
+VOTING_BOTS=100
+# 10,000 CG
+VOTING_BOT_BALANCE=10000000000000000000000
```
jfoutts-celo, line 125: With only 10k Celo Gold, will the bots have enough impact?

packages/celotool/src/cmds/bots/auto-vote.ts
```diff
@@ -0,0 +1,318 @@
...
+      try {
+        // Get current vote for this bot.
+        // Note: though this returns an array, the bot process only ever chooses one group,
+        // so this takes a shortcut and only looks at the first in the response
+        const currentVote = (await election.getVoter(botAccount)).votes[0]
+        const currentGroup = currentVote ? currentVote.group : undefined
+
+        // Handle the case where the group the bot is currently voting for does not have a score
+        if (
+          !currentGroup ||
```
jfoutts-celo, line 80: I'm concerned the bots first votes could decimate the incumbency.  (At least if they're run with enough gold, say 25k.  Otherwise I'm concerned there will be little impact.)

To avoid maybe:
- Run bots once with VOTING_BOT_EXPLORE_PROBABILITY=0, long enough that all of the bots are voting for the incumbency.
- Restart bots with VOTING_BOT_EXPLORE_PROBABILITY=1, so that any displacement of incumbency is based on score.

jfoutts-celo (commented):
It would be nice to observe how the bots would vote on Baklava before turning them loose.

Should we start them with a small amount of gold and analyze their actions before giving them more?

.env.baklava
yerdua: This is what @timmoreton suggested. At this amount, each bot has slightly less than each validator. (Each validator gets 12k, and each group gets 12k).

Tim - what was your rationale for that amount? Is this still in line with the level of impact we want each bot to have?

I think it's fine to have this as a starting point. We can adjust if we want more impact - either by adding more bots, or giving each bot more gold. Given the current setup function, adding more bot accounts is a bit easier than increasing the locked gold for existing ones.

packages/celotool/src/cmds/bots/auto-vote.ts
yerdua: As I understand it, the only validators that have a score are ones that are _currently_ elected. If I am mistaken about that, then sure, this can be changed. This function could probably remain intact otherwise.

packages/celotool/src/cmds/bots/auto-vote.ts
yerdua: That could work. I'm not sure setting `VOTING_BOT_EXPLORE_PROBABILITY=1` is the best idea. I think having _some_ bots voting based on good performance is useful. Maybe like.... 0.8?

.env.baklava
jfoutts-celo: Ah, right.  I think I forgot that essentially all the validator groups would be tied.

packages/celotool/src/cmds/bots/auto-vote.ts
jfoutts-celo: I think that when incumbent validators lose election, their validator score remains whatever it was for the last epoch they were elected.

```
celocli validator:list | grep 0\\. | wc -l
Fetching Validators... done
     120
```

yerdua: yeah, you're right. They seem to preserve the score from when they were last elected.

However, that, in itself, does not lead to a decision on what should happen with this line of code.

jfoutts-celo (approved):
Code LGTM!

PR #2340, opened 2019-12-26
Add unit tests and cli checks for validator hotfix interaction

### LINKED ISSUE #2144: Developers should have tests for whitelisting hotfixes
### Expected Behavior

 * Add Unit tests
 * Cli should `runChecks` to verify we are whitelisting with the right key (account key)

### Current Behavior

With https://github.com/celo-org/celo-monorepo/pull/2143, we allow validator account keys to whitelist hotfixes, but we do not have any unit tests nor CLI checks for it.

yorhodes:
### Description

Adds unit tests and cli checks for validator hotfix interaction

### Tested

- Unit tests in `packages/protocol/test/governance.ts`
- Checks added to `pacakages/cli/src/commands/governance`

### Other changes

- Changes overloaded `authorizeValidatorSigner` to `authorizeValidatorSignerWithPublicKey`

### Related issues

- Fixes #2144


packages/cli/src/commands/governance/approvehotfix.ts
```diff
@@ -17,10 +18,15 @@ export default class ApproveHotfix extends BaseCommand {
 
   async run() {
     const res = this.parse(ApproveHotfix)
-
+    const account = res.flags.from
+    this.kit.defaultAccount = account
     const governance = await this.kit.contracts.getGovernance()
+
+    await newCheckBuilder(this)
```
mcortesi, line 26: you could also add the `hotifixIsNotExecuted()` check

packages/cli/src/commands/governance/executehotfix.ts
```diff
@@ -17,10 +19,24 @@ export default class ExecuteHotfix extends BaseCommand {
 
   async run() {
     const res = this.parse(ExecuteHotfix)
+    const account = res.flags.from
+    const hotfix = await buildProposalFromJsonFile(this.kit, res.flags.jsonTransactions)
+    const hash = proposalToHash(this.kit, hotfix)
```
mcortesi: this will need to be updated after the `salt` addition

packages/cli/src/commands/governance/vote.ts
```diff
@@ -20,10 +21,19 @@ export default class Vote extends BaseCommand {
...
 
+    this.kit.defaultAccount = signer
     const governance = await this.kit.contracts.getGovernance()
 
-    const tx = await governance.vote(res.flags.proposalID, res.flags.vote as keyof typeof VoteValue)
-    await displaySendTx('voteTx', tx, { from: res.flags.from })
+    await newCheckBuilder(this, signer)
+      .isVoteSignerOrAccount()
+      .proposalExists(id)
+      .proposalInStage(id, 'Referendum')
```
mcortesi, line 34: don't we have an enum for proposal stages, probably we should.. and we could use it here and on the other checks

packages/cli/src/commands/validator/list.ts
```diff
@@ -1,4 +1,4 @@
-import { Validator } from '@celo/contractkit/src/wrappers/Validators'
+import { Validator } from '@celo/contractkit/lib/wrappers/Validators'
```
mcortesi, line 1: nice find

packages/cli/src/utils/checks.ts
```diff
@@ -168,6 +206,14 @@ class CheckBuilder {
       })
     )
 
+  isVoteSignerOrAccount = () =>
+    this.addCheck(
+      `${this.signer!} is vote signer or registered account`,
+      this.withAccounts(async (accs) => {
+        return accs.voteSignerToAccount(this.signer!).then(() => true, () => false)
```
mcortesi: are we doing this `this.signer!` on the other checks? Looks like an easy way for a developer error..  since signer is an optional parameter on the `newChecksBuilder`

packages/contractkit/src/wrappers/Accounts.ts
```diff
@@ -198,13 +198,13 @@ export class AccountsWrapper extends BaseWrapper<Accounts> {
...
-        this.contract.methods.authorizeValidatorSigner(
+        this.contract.methods.authorizeValidatorSignerWithPublicKey(
           signer,
-          pubKey,
           proofOfSigningKeyPossession.v,
           proofOfSigningKeyPossession.r,
-          // @ts-ignore Typescript does not support overloading.
-          proofOfSigningKeyPossession.s
+          proofOfSigningKeyPossession.s,
+          // @ts-ignore bytes type
```
mcortesi: we can use something like:
```ts
type SolBytes = Array<string | number[]>
const toBytes = (input: any): SolBytes => input
const fromBytes = (input: SolBytes): any => input as any
```
from BaseWrapper to avoid the ts-ignore here

mcortesi (commented):
Left some comments

packages/cli/src/commands/governance/vote.ts
yorhodes: these strings are restricted to the `keyof typeof ProposalStage` enum but I can import and use the properties as well

packages/cli/src/utils/checks.ts
yorhodes: yes, this pattern is already used in several places

packages/cli/src/commands/governance/vote.ts
yorhodes: the oclif input is also restricted to an enum already

mcortesi (approved):

PR #1982, opened 2019-12-02
Limechain/16xx

BatiGencho:
### Description

Small fixes to  few issues described in the sections below.

### Related issues

- Fixes #1601 
- Fixes #1602 
- Fixes #1605 
- Fixes #1606 
- Fixes #1841 
- Fixes #1379 
- Fixes #456

PR #2342, opened 2019-12-27
Fixing multisig tests

mrsmkl:
### Description

Looks like it was broken in #1982, for some reason this test wasn't running there.

### Tested

Let's see if it runs in CI.


mcortesi (approved):
Thanks for doing this!

PR #2347, opened 2020-01-01
[Slashing] Adding whitelist checks for onlySlasher to Validators.sol

### LINKED ISSUE #2281: [Slashing] Change Validators.sol to use LockedGold's slashing whitelist for permissions
**Expected Behavior**
`Validators.sol` should reference `LockedGold`'s whitelist of slashing addresses for permissioned functions.

**Current Behavior**
Hardcoded permissioned addresses.

_Originally posted by @asaj in https://github.com/celo-org/celo-monorepo/pull/2239_

lucasege:
### Description

Added checks into `lockedGold`'s whitelist of approved slashers instead of relying on "hardcoded" slashing addresses for `forceDeaffiliateIfValidator` and `halveSlashingMultiplier`.

Added `addSlasher` calls to migrations for `GovernanceSlasher`, `DowntimeSlasher`, and `DoubleSigningSlasher`.

Removed some tests that were testing hard coded contract addresses.

**Did not add governance contract to whitelist by default.** Let me know if you think it should be.

### Tested

Updated and ran unit tests.

### Related issues

- Fixes #2281 

### Sister PRs

Relies on https://github.com/celo-org/celo-monorepo/pull/2317 which relies on https://github.com/celo-org/celo-monorepo/pull/2302


packages/protocol/contracts/common/UsingRegistry.sol
```diff
@@ -135,4 +135,22 @@ contract UsingRegistry is Ownable {
...
+
+  /**
+   * @notice Iterates over provided array of identifiers, getting the address for each.
+   *         Returns true if `sender` matches the address of one of the provided identifiers.
+   * @param identifiers Array of approved identifiers.
+   * @param sender Address in question to verify membership.
+   * @return True if `sender` corresponds to the address of any of `identifiers`
+   *         registry entries.
+   */
+  function isOneOf(string[] memory identifiers, address sender) internal view returns (bool) {
```
asaj: How is this different from `onlyRegisteredContracts`?

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -256,22 +264,27 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
+   * @param slasherIdentifier Identifier to whitelist.
    */
-  function addSlasher(address slasher) external onlyOwner {
-    require(slasher != address(0) && !isSlasher[slasher], "Invalid address to `addSlasher`.");
-    isSlasher[slasher] = true;
-    emit SlasherWhitelistAdded(slasher);
+  function addSlasher(string calldata slasherIdentifier) external onlyOwner {
+    require(!slashingMap[slasherIdentifier], "Cannot add slasher ID twice.");
+    slashingWhitelist.push(slasherIdentifier);
+    slashingMap[slasherIdentifier] = true;
```
asaj: Should we key by the ID? i.e. the hash of the string?

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -256,22 +264,29 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
...
   /**
    * @notice Adds `slasher` to whitelist of approved slashing addresses.
-   * @param slasher Address to whitelist.
+   * @param slasherIdentifier Identifier to whitelist.
    */
-  function addSlasher(address slasher) external onlyOwner {
-    require(slasher != address(0) && !isSlasher[slasher], "Invalid address to `addSlasher`.");
-    isSlasher[slasher] = true;
-    emit SlasherWhitelistAdded(slasher);
+  function addSlasher(string calldata slasherIdentifier) external onlyOwner {
```
mcortesi, line 276: we can check that the identifier exists on Registry

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -1152,23 +1152,14 @@ contract Validators is
...
-
   /**
    * @notice Removes a validator from the group for which it is a member.
    * @param validatorAccount The validator to deaffiliate from their affiliated validator group.
    */
   function forceDeaffiliateIfValidator(address validatorAccount)
     external
     nonReentrant
-    onlyRegisteredContracts(getCanForceDeaffiliation())
+    onlyRegisteredContracts(getLockedGold().getSlashingWhitelist())
```
mcortesi: we can have an `onlySlasher` that calls a `isSlasher(address)`  on locked Gold.

`isSlasher would internally call `isOneOf`, but from LockedGold, instead of here. That would more efficient in terms of gas

packages/protocol/contracts/common/UsingRegistry.sol
```diff
@@ -61,17 +61,25 @@ contract UsingRegistry is Ownable {
...
+
+  /**
+   * @notice Iterates over provided array of identifiers, getting the address for each.
+   *         Returns true if `sender` matches the address of one of the provided identifiers.
+   * @param identifierHashes Array of hashes of approved identifiers.
+   * @param sender Address in question to verify membership.
+   * @return True if `sender` corresponds to the address of any of `identifiers`
+   *         registry entries.
+   */
+  function isOneOf(bytes32[] memory identifierHashes, address sender) internal view returns (bool) {
```
mcortesi: I initially though this function could live in the `Registry` thus, it would a be more efficient in terms of gas (N internal functions calls vs N public funtion calls).

That would mean changing the Registry, what do you think @asaj ?

mcortesi (commented):
Left some comments

packages/protocol/contracts/common/UsingRegistry.sol
asaj: Oh yeah that is more efficient, +1

lucasege: Done, added change to `registry` and `IRegistry`. Let me know if there are any further reaching implications I'm missing that aren't covered by tests.

mcortesi (approved):
Love it! Thanks for doing the changes

PR #2352, opened 2020-01-02
Optimize Bundles

### LINKED ISSUE #2351: Reduce bundle sizes on website
Currently bundles are much bigger than we want

aaronmgdr:
### Description

defer loading of some libraries and components, rewrite to remove code, and generally do what can be done to decrease bytes sent to browser

* Defer loading Sentry 
* defer loading the segment/isEU library
* replace React.Animated with a few lines of CSS
* Lazy Load the Carousel on the homepage
* combine SVGs of the celo logo light and celo logo dark into one svg with a light and dark prop

### Tested

using `yarn build` with expo showing sizes of each page bundle and via the webkit-bundle-visualizer
Screen Shot 2020-01-02 at 4 49 03 PM

### Related issues

- Fixes #2351

### Backwards compatibility


packages/web/src/fonts/Fonts.tsx
```diff
@@ -113,7 +113,7 @@ interface TextChildren {
 
 export function Ul(props: ViewProps & ViewChildren) {
   return (
-    <View style={[styles.ul, props.style]} accessibilityRole={'list'}>
+    <View style={StyleSheet.flatten([styles.ul, props.style])} accessibilityRole={'list'}>
```
cmcewen: Why use flatten here?

yarn.lock
```diff
@@ -536,21 +536,21 @@
     source-map "^0.5.0"
```
cmcewen: Unless you changed deps somewhere I wouldn't check this in

packages/web/src/fonts/Fonts.tsx
aaronmgdr: there was a type error about how StyleProp[] wasnt assignable to style

packages/web/src/analytics/analytics.ts
```diff
@@ -1,41 +1,49 @@
...
+  if (process.browser && !analyticsStarted && (await canTrack())) {
+    const Segment = await import('load-segment').then((mod) => mod.default)
     const { publicRuntimeConfig } = getConfig()
     analytics = Segment({ key: publicRuntimeConfig.__SEGMENT_KEY__ })
   } else {
     analytics = {
       track: () => null,
     }
   }
+  analyticsStarted = true
```
cmcewen: Could this possibly be called by multiple functions and do duplicate work? Should you check in progress vs completed?

cmcewen (approved):

packages/web/src/analytics/analytics.ts
aaronmgdr: moving this above the import,

its not likely as this should only be called once per "page" when _app.tsx mounts. or when agreement is obtained (which means it wont have been called before)

PR #2358, opened 2020-01-03
Update prettier to 1.19.1 to support TypeScript 3.7 (optional chaining, nullish coalescing, etc)

jeanregisser:
### Description

Update prettier to 1.19.1 to support TypeScript 3.7 (optional chaining, nullish coalescing, etc)

See https://prettier.io/blog/2019/11/09/1.19.0.html#typescript

Note that it now formats some expressions slightly differently. For instance template literals with lines longer than the print width are now different. See https://github.com/prettier/prettier/issues/7162

### Tested

- Ran prettier again through all our files (`yarn run prettify`)
- Ran tests

### Other changes

N/A

### Related issues

N/A

### Backwards compatibility

Yes.


yarn.lock
```diff
@@ -26577,30 +26621,25 @@ prettier-plugin-solidity@1.0.0-alpha.34:
...
+prettier@1.18.2:
   version "1.18.2"
   resolved "https://registry.yarnpkg.com/prettier/-/prettier-1.18.2.tgz#6823e7c5900017b4bd3acf46fe9ac4b4d7bda9ea"
   integrity sha512-OeHeMc0JhFE9idD4ZdtNibzY0+TPHSpSSb9h8FqtP+YnoZZ1sl8Vc9b1sasjfymH3SonAF4QcA2+mzHPhMvIiw==
 
-prettier@^1.14.3:
-  version "1.17.1"
-  resolved "https://registry.yarnpkg.com/prettier/-/prettier-1.17.1.tgz#ed64b4e93e370cb8a25b9ef7fef3e4fd1c0995db"
-  integrity sha512-TzGRNvuUSmPgwivDqkZ9tM/qTGW9hqDKWOE9YHiyQdixlKbv7kvEqsmDPrcHJTKwthU774TQwZXVtaQ/mMsvjg==
+prettier@1.19.1, prettier@^1.14.2, prettier@^1.14.3, prettier@^1.15.3:
```
mcortesi: i would expect us to have only one version... i wonder where the other come from

mcortesi: maybe we ca use a resolution rule; it will probably eliminate other duplicated deps

mcortesi (approved):
Love it

yarn.lock
jeanregisser: I manually tried to deduplicate but some transitive deps had specific versions pinned.
Good idea for the resolution rule, I'm on it 👍

PR #2361, opened 2020-01-03
Removes EpochData

kobigurk:
### Description

In conjuction with https://github.com/celo-org/celo-blockchain/pull/780.
EpochData is not needed anymore since it's now part of the block.


mrsmkl (approved):
LGTM

PR #2371, opened 2020-01-07
Clarify documentation on validator proxy

bgeihsgt:
### Description

I was setting up my validator and it wasn't syncing. After some troubleshooting, I realized it was a problem with my network config on my proxy. I would have caught this earlier if I knew the proxy should have been syncing after I ran it (probably obvious to most, but wasn't to me). I thought this line would help point people in the right direction.

### Tested

Ran Markdown preview to confirm the change.

### Other changes

N/A

### Related issues

N/A

### Backwards compatibility

N/A


timmoreton (approved):
Thanks for the contribution!

PR #2381, opened 2020-01-08
First pass at Azure templates for validator infrastructure

silasbw:
### Description

This is part of the work to provide a managed Azure validator service/system.  This change focuses on the Azure infrastructure and software setup.

### Tested

Manually by deploying the templates and SSHing into hosts to verify properties (_e.g._, firewall, syncing, etc)


packages/azure/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "version": "0.0.1",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "node combine.js components/ > template.json",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "UNLICENSED",
```
tkporter: I assume we'd like to use "Apache-2.0" like the majority of the project? @timmoreton

packages/azure/components/main.json
```diff
@@ -0,0 +1,146 @@
...
+      "defaultValue": "10.0.0.0/24",
+      "metadata": {
+        "description": "Bastion Virtual Network Address prefix"
+      }
+    },
+    "validatorDiskSize": {
+      "type": "string",
+      "defaultValue": "1024",
+      "metadata": {
+        "description": "Validator data disk size"
```
tkporter: In GB I assume?

packages/azure/components/network.json
```diff
@@ -0,0 +1,266 @@
...
+        "name": "[parameters('attesterPublicIpAddressSku')]"
+      }
+    },
+    {
+      "name": "[variables('validatorNetworkSecurityGroupName')]",
+      "type": "Microsoft.Network/networkSecurityGroups",
+      "apiVersion": "2019-02-01",
+      "location": "[parameters('location')]",
+      "properties": {
+        "securityRules": [
```
tkporter, line 77: Should this also include a rule to allow inbound 30303 so that the proxy can send messages back to the validator from its internal port? Or does it seem to be working just fine?

packages/azure/components/network.json
```diff
@@ -0,0 +1,266 @@
...
+        ]
+      }
+    },
+    {
+      "name": "[variables('attesterNetworkSecurityGroupName')]",
+      "type": "Microsoft.Network/networkSecurityGroups",
+      "apiVersion": "2019-02-01",
+      "location": "[parameters('location')]",
+      "properties": {
+        "securityRules": [
```
tkporter, line 157: Same here, do we need to open up 30303 for the full node to talk to the rest of the network?

packages/azure/README.md
```diff
@@ -0,0 +1,98 @@
...
+PUBLIC_KEY=...
+AZURE_USER_ID=... # You can get this via the Portal or with `az ad user list`
+az group deployment create \
+   --resource-group $GROUP \
+   --name ${GROUP}-deployment \
+   --template-file=template.json \
+   --parameters adminPublicKey=@${PUBLIC_KEY} adminUsername=${ADMIN_USER} userObjectId=${AZURE_USER_ID}
+```
+
+Clean up your resources when you're done:
```
tkporter: To clarify, is this if you want to tear down everything or for when you're done creating the whole deployment?

packages/azure/README.md
```diff
@@ -0,0 +1,98 @@
...
+| `proxyName` | string | The proxy name for ethstats |
+| `validatorName` | string | The validator name for ethstats |
+| `adminUsername` | string | Admin username for SSH access |
+| `adminPublicKey` | secureString | Admin public key for SSH access |
+| `attesterPostgreSQLUsername` | string | Admin username for Attester PostgreSQL server |
+| `attesterPostgreSQLPassword` | secureString | Admin password for Attester PostgreSQL server |
+| `validatorAccountAddress` | string | Validator etherbase account address |
+| `proxyPublicKey` | string | Proxy enode public key |
+| `proxyPrivateKey` | secureString | Proxy etherbase account private key |
+| `validatorPrivateKey` | secureString | Validator etherbase account private key |
```
tkporter: I think we should make this more clear that these are for the authorized validator signer, not necessarily the validator account

packages/azure/README.mustache
```diff
@@ -0,0 +1,83 @@
...
+
+```
+az group delete --name $GROUP
+```
+
+### Required Parameters
+
+| Parameter | Type | Description |
+| --------- | -----| ----------- |
+{{#parameters}}
```
tkporter, line 61: Very cool

packages/azure/build-json.js
```diff
@@ -0,0 +1,70 @@
...
+    attesterScript: fs.readFileSync('./build/attester-startup.sh').toString('base64'),
+    proxyScript: fs.readFileSync('./build/proxy-startup.sh').toString('base64'),
+    validatorScript: fs.readFileSync('./build/validator-startup.sh').toString('base64'),
+    variables,
+    parameters,
+    proxyIps,
+    validatorIps,
+  }
+  const objects = contents.map((content) => JSON.parse(mustache.render(content, context)))
+  const armTemplate = objects.slice(1).reduce((acc, val) => mergeARMTemplates(acc, val), objects[0])
```
tkporter: Why the `slice(1)`?

You might be able to change `(acc, val) => mergeARMTemplates(acc, val)` to just `mergeARMTemplates` right?

Maybe a comment would help me understand better what's going on here

packages/azure/components/main.json
```diff
@@ -0,0 +1,373 @@
...
+    "gethNodeDockerImageTag": {
+      "type": "string",
+      "defaultValue": "[parameters('celoEnv')]",
+      "metadata": {
+        "description": "Celo geth Docker image tag"
+      }
+    },
+    "gethExporterDockerImageRepository": {
+      "type": "string",
+      "defaultValue": "gcr.io/celo-testnet-production/geth-exporter",
```
tkporter, line 158: Were you able to pull this? I thought this image would be internal. Either way, 1.9 has added a built-in prometheus metric endpoint when running geth with `--metrics --pprof`, which we will start using to remove the need for running the geth exporter. 1.9 should be landing within the next couple of days so we can probably remove it

packages/azure/components/main.json
```diff
@@ -0,0 +1,373 @@
...
+    "attestationServiceDockerImageRepository": {
+      "type": "string",
+      "defaultValue": "us.gcr.io/celo-testnet/celo-monorepo",
+      "metadata": {
+        "description": "Celo attestation service Docker image repository"
+      }
+    },
+    "attestationServiceDockerImageTag": {
+      "type": "string",
+      "defaultValue": "attestation-service-8e69cf86010b62e283d5f9285f181fca5483733e",
```
tkporter: Going off the docs I think this should be `attestation-service-<celoEnv>`

packages/azure/components/main.json
```diff
@@ -0,0 +1,373 @@
...
+    },
+    "bootnodes": {
+      "type": "string",
+      "defaultValue": "enode://5aaf10664b12431c250597e980aacd7d5373cae00f128be5b00364344bb96bce7555b50973664bddebd1cb7a6d3fb927bec81527f80e22a26fa373c375fcdefc@34.82.45.71:30301",
+      "metadata": {
+        "description": "geth --bootnodes argument"
+      }
+    },
+    "gethVerbosity": {
+      "type": "string",
```
tkporter: Why make these strings and not `int`s? there are a few others, like networkId, blockTime, proxyMaxPeers, etc

packages/azure/README.mustache
```diff
@@ -0,0 +1,83 @@
...
+az network nic ip-config update \
+   --name ipconfig1 \
+   --nic-name validatorNetworkInterface \
+   --resource-group ${GROUP} \
+   --public-ip-address validatorPublicIpAddress
+```
+
+## Future work
+
+1.  Use suggested defaults for VM sizes and OS disk types.
```
tkporter: I see suggested defaults in the parameters, is it not working?

packages/azure/components/compute.json
```diff
@@ -0,0 +1,328 @@
...
+          "computerName": "[variables('validatorVirtualMachineName')]",
+          "adminUsername": "[parameters('adminUsername')]",
+          "linuxConfiguration": {
+            "disablePasswordAuthentication": true,
+            "ssh": {
+              "publicKeys": [
+                {
+                  "path":
+                    "[concat('/home/', parameters('adminUsername'), '/.ssh/authorized_keys')]",
+                  "keyData": "[parameters('adminPublicKey')]"
```
tkporter, line 101: Formatting needs a tab?

packages/azure/scripts/provider-init.sh
```diff
@@ -0,0 +1,43 @@
...
+DATA_DIR=/root/.celo
+CELO_ENV=${celoEnv}
+BOOTNODES=${bootnodes}
+VALIDATOR_KEY_VAULT=${validatorKeyVaultName}
+PROXY_KEY_VAULT=${proxyKeyVaultName}
+ATTESTER_KEY_VAULT=${attesterKeyVaultName}
+GETH_NODE_DOCKER_IMAGE=${gethNodeDockerImageRepository}:${gethNodeDockerImageTag}
+GETH_EXPORTER_DOCKER_IMAGE=${gethExporterDockerImageRepository}:${gethExporterDockerImageTag}
+ATTESTATION_SERVICE_DOCKER_IMAGE=${attestationServiceDockerImageRepository}:${attestationServiceDockerImageTag}
+IN_MEMORY_DISCOVERY_TABLE_FLAG=""
```
tkporter: Probably can remove this. Persisted discovery table sounds like the safest bet

packages/azure/README.md
```diff
@@ -0,0 +1,98 @@
...
+| --------- | -----| ----------- |
+| `userObjectId` | string | The object ID of a user, service principal or security group in the Azure Active Directory tenant for the vault. |
+| `proxyName` | string | The proxy name for ethstats |
+| `validatorName` | string | The validator name for ethstats |
+| `adminUsername` | string | Admin username for SSH access |
+| `adminPublicKey` | secureString | Admin public key for SSH access |
+| `attesterPostgreSQLUsername` | string | Admin username for Attester PostgreSQL server |
+| `attesterPostgreSQLPassword` | secureString | Admin password for Attester PostgreSQL server |
+| `validatorAccountAddress` | string | Validator etherbase account address |
+| `proxyPublicKey` | string | Proxy enode public key |
```
tkporter, line 84: We'll have to make this clear that this is slightly different from the full public key of the proxy-- I think it's something like the public key without the first two characters of the string

packages/azure/scripts/provider-init.sh
```diff
@@ -0,0 +1,43 @@
...
+PROXY_INTERNAL_IP_ADDRESS=${proxyInternalIpAddress}
+VALIDATOR_INTERNAL_IP_ADDRESS=${validatorInternalIpAddress}
+ATTESTER_ACCOUNT_ADDRESS=${attesterAccountAddress}
+VALIDATOR_ACCOUNT_ADDRESS=${validatorAccountAddress}
+PROXY_URL="enode://${proxyPublicKey}@${proxyInternalIpAddress}:30503;enode://${proxyPublicKey}@${proxyExternalIpAddress}:30303"
+SMS_PROVIDERS=twilio
+ATTESTER_TWILIO_ACCOUNT_SID=${attesterTwilioAccountSID}
+ATTESTER_TWILIO_MESSAGE_SERVICE_SID=${attesterTwilioMessageServiceSID}
+ATTESTER_DB_USERNAME="${attesterPostgreSQLUsername}@${attesterDBName}"
+ATTESTER_DB_HOSTNAME="${attesterDBName}.postgres.database.azure.com"
```
tkporter, line 31: Will this conflict with other deployments? Or is this hostname only resolved internally

packages/azure/validator/validator-run
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+set -e
+
+. /etc/default/celo
+
+SECRET_DIR=$(mktemp --directory --tmpdir=/dev/shm)
```
tkporter: 🤤

packages/azure/validator/validator-run
```diff
@@ -0,0 +1,68 @@
...
+      --rpcaddr 0.0.0.0 \
+      --rpcapi=eth,net,web3 \
+      --rpccorsdomain='*' \
+      --rpcvhosts=* \
+      --ws \
+      --wsaddr 0.0.0.0 \
+      --wsorigins=* \
+      --wsapi=eth,net,web3 \
+      --nodekey=$SECRET_DIR/validatorPrivateKey \
+      --etherbase=$ACCOUNT_ADDRESS \
```
tkporter: Is this set anywhere? Should it be `VALIDATOR_ACCOUNT_ADDRESS`?

packages/azure/components/keyvault.json
```diff
@@ -0,0 +1,298 @@
...
+                "GetIssuers",
+                "ListIssuers",
+                "SetIssuers",
+                "DeleteIssuers"
+              ]
+            }
+          },
+          {
+            "objectId":
+              "[reference(concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('validatorUserAssignedIdentityName')), '2018-11-30', 'Full').properties.principalId]",
```
tkporter, line 53: Nice

packages/azure/components/keyvault.json
```diff
@@ -0,0 +1,298 @@
...
+      "apiVersion": "2018-02-14",
+      "location": "[parameters('location')]",
+      "dependsOn": ["[variables('proxyKeyVaultName')]"],
+      "properties": {
+        "value": "[parameters('proxyPrivateKey')]"
+      }
+    },
+    {
+      "type": "Microsoft.KeyVault/vaults/secrets",
+      "name":
```
tkporter, line 234: I assume it assigns which vault the secret belongs to using the name?

packages/azure/attester/attestation-service-run
```diff
@@ -0,0 +1,41 @@
...
+
+DB_PASSWORD=$(get_secret $ATTESTER_KEY_VAULT 'attesterPostgreSQLPassword')
+DB_USERNAME=$ATTESTER_DB_USERNAME
+DATABASE_URL="postgres://${DB_USERNAME}:${DB_PASSWORD}@${ATTESTER_DB_HOSTNAME}:5432/postgres"
+
+CELO_PROVIDER=http://localhost:8545
+TWILIO_AUTH_TOKEN=$(get_secret $ATTESTER_KEY_VAULT 'attesterTwilioAuthToken')
+
+/usr/bin/docker run \
+  --name attestation-service \
```
tkporter: I see that `%N` is passed in `attestation-service.service` to this-- did you mean to set the name with that value `$1` here?

packages/azure/attester/celo-node.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Docker Container %N
+Requires=docker.service
+After=docker.service
+
+[Service]
+Restart=always
+ExecStart=/usr/local/bin/celo-node-run %N
```
tkporter, line 8: I think it will be a minor inconvenience to follow the docs by calling this docker container something other than what we have in the docs (same goes for the validator/proxy nodes). In the docs we have the attestation node as `celo-attestations`, the validator node as `celo-validator`, the proxy as `celo-proxy`. It's probably easier to follow that pattern in these nodes because people following the docs will be confused

packages/azure/scripts/libcelo.sh
```diff
@@ -0,0 +1,9 @@
+get_secret () {
```
tkporter, line 1: What do you think about maybe bringing downloading docker or mounting disks into here?

tkporter (commented):
Looks amazing! I can't imagine how much time this must've taken. Just some small comments but overall it looks really good

packages/azure/scripts/provider-init.sh
silasbw: It's public.  A postgres proxy/gateway sits in front of the actual DB instance and by default rejects all connections (FYI, if you look at the documentation it sometimes refers to this process as "the firewall", which confused me at first).  I don't think you can disable the public IP, but I have a TODO to confirm that (some evidence https://social.msdn.microsoft.com/Forums/SECURITY/en-US/ece09d75-4eea-458d-9e67-661c92fe26d5/deny-public-access-to-azure-postgresql-service?forum=AzureDatabaseforPostgreSQL)

packages/azure/README.md
silasbw: I'll disambiguate.

packages/azure/README.md
silasbw: yeah...so, does that have a name?  Should we tease out the instructions here https://github.com/celo-org/celo-monorepo/blob/master/packages/terraform-modules-public/example/README.md and put them in a common place?

packages/azure/README.md
silasbw: Oops, yeah, I'll disambiguate.

packages/azure/README.mustache
silasbw: This could probably say "Tune suggested defaults" or something.  I'll update.

packages/azure/README.md
jcortejoso: @silasbw I agree would be nice to have a common place explaining how to generate the keys. Maybe can be a good additional content to the key management doc https://github.com/celo-org/celo-monorepo/tree/master/packages/docs/node-operations

packages/azure/attester/attestation-service-run
silasbw: doh, good catch

packages/azure/build-json.js
silasbw: `mergeARMTemplates` might not be worth the trouble: if will throw an error if you pass an object that is missing `$schema` or `contentVersion` (that's the reason for `objects.slice(1)....slice(0)`; and it's function signature is not compatible with `reduce` (it takes an additional 3 optional parameters and they aren't `index` or `source`).  I'll comment or just replace it with a simple readable reduce function.

packages/azure/components/compute.json
silasbw: I'll moneky with it.  I think this is a pre-commit hook formatting `.json`.

packages/azure/components/keyvault.json
silasbw: correct -- the `/` in names is significant.

packages/azure/components/main.json
silasbw: oh cool, so the exporter goes away entirely?

packages/azure/components/network.json
silasbw: No it doesn't.  I'm trying to codify https://docs.celo.org/v/master/getting-started/baklava-testnet/running-a-validator#networking-requirements.

Does the proxy send UDP messages to the validator or establish a TCP connection?  I thought the validator established a TCP connection with the proxy (so there would be no need for additional validator firewall rules).

packages/azure/components/network.json
silasbw: Oops, yes, for this we def. do.

packages/azure/scripts/libcelo.sh
silasbw: Good idea -- will update.

packages/azure/validator/validator-run
silasbw: yeah -- it's creating a tmpfs filesystem pinned to memory to help ensure VMs never write secret data to disk.

packages/azure/validator/validator-run
silasbw: Oops, fixed.

packages/azure/attester/celo-node.service
silasbw: Updated.

packages/azure/components/main.json
silasbw: Updated description.

packages/azure/scripts/libcelo.sh
silasbw: I ended up bringing it into `provider-init.sh`: disk setup and initial pulling images is only done during initialization, but the "run scripts" use `get_secret` (and potentially other functions in libcelo.sh) every time they execute.

packages/azure/components/network.json
```diff
@@ -0,0 +1,291 @@
...
+              "priority": 100,
+              "direction": "Inbound",
+              "sourcePortRanges": [],
+              "destinationPortRanges": [],
+              "sourceAddressPrefixes": [],
+              "destinationAddressPrefixes": []
+            }
+          },
+          {
+            "name": "DiscoveryInternal",
```
tkporter: Because this is just for the normal protocol TCP communication (and not discovery, which is over UDP), I think this name is inaccurate

packages/azure/components/network.json
```diff
@@ -0,0 +1,291 @@
...
+              "access": "Allow",
+              "direction": "Inbound",
+              "sourceAddressPrefix": "*",
+              "sourcePortRange": "*",
+              "destinationAddressPrefix": "*",
+              "destinationPortRange": "22"
+            }
+          },
+          {
+            "name": "Discovery",
```
tkporter: This includes normal protocol messages & not only discovery, so I sort of liked the old name `ProxyExternal`

packages/azure/components/network.json
```diff
@@ -0,0 +1,291 @@
...
```
tkporter: Ditto, is there a name that would also show this is for non-discovery protocol messages?

packages/azure/components/network.json
```diff
@@ -0,0 +1,291 @@
...
+            "properties": {
+              "protocol": "*",
```
tkporter, line 174: looking at the docs the options are `Tcp, Udp, Icmp, Esp, *, Ah`, do you think there's any downside to allowing those non TCP/UDP protocols?

packages/azure/components/network.json
tkporter: Ah yeah you're totally right

packages/azure/package.json
tkporter: I think it's safe to go with Apache-2.0

packages/azure/components/main.json
tkporter: We haven't done any work to incorporate it into our own infrastructure yet, but because this isn't for our purposes and we've released the geth 1.9 merge, you're fine to just remove the geth exporter entirely from this PR

packages/azure/README.md
tkporter: I don't know if it has a name, but maybe `proxyEnodePublicKey`. Adding instructions to that link @jcortejoso sent would be great

packages/azure/scripts/provider-init.sh
tkporter: If it's public would there then be a problem with conflicting names if another person were to deploy?

I see that `attesterDBName` is created using `deployment().name`, is that value `${GROUP}-deployment` (from the readme)? Or is it something more unique?

tkporter (commented):
Thanks for making the changes! Just some small comments

One other thing that would be helpful (doesn't have to be a part of this PR or anything) would be to give commands in the readme describing how to ssh into a specific VM & interact with the docker container running so any maintenance would be easier

packages/azure/build-json.js
```diff
@@ -0,0 +1,70 @@
...
+  if (!stat.isDirectory()) {
+    return acc.concat([fs.readFileSync(pathname).toString()])
+  }
+  fs.readdirSync(pathname).forEach((entry) => {
+    acc = load(acc, path.join(pathname, entry))
+  })
+  return acc
+}
+
+function main() {
```
codyborn, line 20: This is a clever approach to building the template.json output. I believe the best practice is to use [nested templates](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/linked-templates). This allows you to call other templates like you would call functions in a programming language. In that case you wouldn't need to do any compiling of the templates to create the template.json. 
So you'd have a mainTemplate.json which would take in the user specified arguments and deploy all of the nested template deployments as **Microsoft.Resources/deployments**. Happy to jump on a call to discuss more.

packages/azure/components/main.json
```diff
@@ -0,0 +1,373 @@
...
+        "description": "OS Disk Type for every Virtual Machine"
+      }
+    }
+  },
+  "variables": {
+    "virtualNetworkName": "[concat(deployment().name, '-', 'virtualNetwork')]",
+    "virtualNetworkId":
+      "[resourceId(resourceGroup().name, 'Microsoft.Network/virtualNetworks', variables('virtualNetworkName'))]",
+    "subnetId": "[concat(variables('virtualNetworkId'), '/subnets/default')]",
+    "validatorKeyVaultName": "[concat(deployment().name, '-validatorKeyVault')]",
```
codyborn: I believe the uniqueness requirements for the deployment name and the key vault are not the same. Deployment name must be unique for a user's subscription; however, the key vault name must be globally unique. In this case, I recommend incorporating a unique ID generated from the [UniqueString function](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-string#uniquestring) when generating the KeyVault name.

codyborn: Also, in general, we'll need to keep in mind the naming length restrictions for all resources. For example, AKV name must be <= 24 chars: https://feedback.azure.com/forums/906355-azure-key-vault/suggestions/34763077-allow-more-than-24-characters-for-key-vault-name-l

codyborn: [This site](https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/naming-and-tagging#naming-scope) explains it better. See the nested link for a table of each resource and its naming uniqueness and length constraints. For example, [here's the restrictions for AKV](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftkeyvault).

packages/azure/components/main.json
```diff
@@ -0,0 +1,373 @@
...
+        "description": "Admin username for SSH access"
+      }
+    },
+    "adminPublicKey": {
+      "type": "secureString",
+      "metadata": {
+        "description": "Admin public key for SSH access"
+      }
+    },
+    "attesterPostgreSQLUsername": {
```
codyborn, line 36: We could just use the adminUsername specified above for SQL as well. Any reason they should be different?

packages/azure/README.md
```diff
@@ -0,0 +1,114 @@
+# Azure ARM Templates
+
```
codyborn, line 2: It would be useful to include a pre-built template.json so that users can perform a one-click deploy to Azure from GitHub, similar to this repo: https://github.com/Azure/azure-quickstart-templates/tree/master/101-vm-simple-linux/

packages/azure/README.md
```diff
@@ -0,0 +1,114 @@
...
+| `attesterAccountAddress` | string | Attester account address |
+| `attesterPrivateKey` | secureString | Attester account private key |
+| `attesterGethAccountSecret` | secureString | Password to secure attesterPrivateKey |
+| `attesterTwilioAccountSID` | string | Attester Twilio account SID |
+| `attesterTwilioAuthToken` | secureString | Attester Twilio authentication token |
+| `attesterTwilioMessageServiceSID` | string | Attester Twilio message service SID |
+
+### SSH access
+
+The proxy and attester expose port 22 on their public IP addresses.
```
codyborn, line 97: It's best practice to leave these locked down by default, but to create an explicit "Deny" policy to make it easy for the user to open after the deployment.

packages/azure/components/network.json
silasbw: Hmm, I'm not sure, but I'll split this into two rules (`Tcp` and `Udp`) to be safe.

packages/azure/components/main.json
silasbw: I have a TODO to address this (and the same constraints on postgres).

packages/azure/components/main.json
silasbw: They don't need to be different. The user doesn't even need access to the `attesterPostgreSQLUsername` (or `attesterPostgreSQLPassword`, or `validatorGethAccountSecret`, or `attesterGethAccountSecret`).  I was going to deal with cleaning those all up together on a second pass.

packages/azure/build-json.js
silasbw: Yeah, I agree i should split out self-contained templates and use nesting.

The janky build-json.js process has been useful for iterating on the entire system and for building the base64 encoded startup.sh script (which depends on the `parameters` and `variables`)

Do you think it's OK to push off the template refactor into another self-contained PR?

packages/azure/README.md
silasbw: oh, good idea.

packages/azure/components/network.json
silasbw: ...actually it's not possible to have two rules on the same port with the same direction.  I'm going to go with `*` for now.  maybe @codyborn has a suggestion?

packages/azure/README.md
silasbw: I'm planning on writing a "Running a Validator on Azure" .md.  Are you OK with pushing that work out into a self-contained PR?

tkporter (approved):
LGTM, deferring to Cody for his thoughts

packages/azure/README.md
tkporter: Sounds good to me

packages/azure/build-json.js
codyborn: Yes, it'll take some refactoring to make this change.

packages/azure/components/main.json
codyborn: Sounds good... until this is fixed, the user experience won't be great in the case of naming mismatch. They deployment will happily kick off but fail midway through once it detects the naming conflict.

codyborn (approved):

codyborn (approved):

Issue #1478, opened 2019-10-25
Devs should have a centralized/unilateral ganache config


yorhodes:
### Expected Behavior

One file for editing ganache server config
`packages/protocol/truffle-config.js` is a good candidate for other packages to source from.

### Current Behavior

Multiple locations for different ganache server configurations
To name a few...
`packages/protocol/scripts/devchain.ts`
`packages/protocol/runTests.js`
`packages/contractkit/src/test-utils/ganache.setup.ts`

BatiGencho:
@asaj Reserved for limechain @ Ventsi

nambrot:
@yorhodes Is this still relevant?

PR #2382, opened 2020-01-08
all ganache configurations are centralized in packages/protocol/truff…

Holymir:
Devs should have a centralized/unilateral ganache config #1478

Expected Behavior
One file for editing ganache server config
packages/protocol/truffle-config.js is a good candidate for other packages to source from.

Current Behavior
Multiple locations for different ganache server configurations
To name a few...
packages/protocol/scripts/devchain.ts
packages/protocol/runTests.js
packages/contractkit/src/test-utils/ganache.setup.ts


asaj:
@Holymir looks like a lot of tests are broken

PR #2392, opened 2020-01-08
Fund voting bot accounts in genesis

yerdua:
### Description

As part of implementation for #1477, a number of voting bot accounts specified in .env gets funded in the genesis block.

Since this change should go in before the next network deploy, this small change has been split out of the larger PR #2327 

In addition, this PR fixes a bug in which constants derived from envvars were not getting the values from the `.env` files. The balance constants (`faucetBalance`, `oracleBalance`, etc) are defined at file load, which is before the celoMiddleware is added, and therefore before the appropriate `.env` file is loaded. This fix keeps the file structure as-is, with all the balances grouped together, but changes them to functions that are called _after_ the env is loaded. If there's a preference against doing it this way, I've prepared an alternate PR here: #2398

### Tested

Tested by spinning up a new testnet and confirming that the accounts were in genesis.

### Related issues

- #1477
- #2327


packages/celotool/src/lib/generate_utils.ts
```diff
@@ -112,6 +113,7 @@ const validatorBalance = fetchEnvOrFallback(
 ) // 10,011 CG
 const faucetBalance = fetchEnvOrFallback(envVar.FAUCET_GENESIS_BALANCE, '10011000000000000000000') // 10,011 CG
 const oracleBalance = fetchEnvOrFallback(envVar.ORACLE_GENESIS_BALANCE, '100000000000000000000') // 100 CG
+const votingBotBalance = fetchEnvOrFallback(envVar.VOTING_BOT_BALANCE, '10000000000000000000000') // 10,000 CG
```
yerdua: None of these are reading correctly from the .env file, and I don't know why. When I tested generating a genesis file with a .env file that had non-default values for all of these, they did not carry through.

nambrot: @yerdua What do you mean by they are not reading correctly from the `.env` file? Are you sure you did not have another .env file overriding the one you were changing?

nambrot (approved):

Issue #2390, opened 2020-01-08
Initial reserve cGLD transferability


rcroessmann:
### Expected Behavior

With X cGLD allocated to the reserve at launch, only Y cGLD should be transferable by reserve spenders  immediately after launch (for rebalancing purposes) whereas the rest, X-Y cGLD, should be frozen and become unfrozen linearly over the course of Z days. The parameters X, Y and Z are tbd. 

Additionally, similar to `getReserveGoldBalance()`, there should be a function `getUnfrozenReserveGoldBalance()` that returns the sum of the currently unfrozen reserve balance and the balances of the `otherReserveAddresses`.

### Current Behavior

All X cGLD allocated to the reserve are, ignoring spending limits, immediately transferable after launch.

PR #2395, opened 2020-01-09
Add isReserveLow function

trianglesphere:
### Description

Adds a function to determine if the reserve ratio is above or below the target reserve ratio (a linear function of time).

Also add `getReserveRatio` which does account for non cGLD assets in the reserve.

### Tested

Unit tests

### Other changes

Switches `assetAllocationWeights` to a mapping under the hood, but keeps the external array based interface.

### Related issues

- Part of issue 529 (epoch rewards)

### Backwards compatibility

Yes


packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -380,6 +380,22 @@ contract EpochRewards is Ownable, Initializable, UsingPrecompiles, UsingRegistry
     _updateTargetVotingYield();
   }
 
+  /**
+   * @notice Determines if the reserve is low enough to demand a diversion from
+   *    the community reward. Targets initial ratio of 2 with a linear decline
+   *    until 25 years have passed where the target ratio will be 1.
+   */
+  function isReserveLow() external view returns (bool) {
+    // TODO: Does this need to be onlyWhenNotFrozen ?
```
asaj: Nope!

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -281,6 +281,32 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
   /*
    * Internal functions
    */
+
+  /**
+   * @notice Computes the ratio of current reserve balance to total stable token valuation.
+   * @return Reserve ratio in a fixed point format.
+   */
+  function getReserveRatio() public view returns (uint256) {
```
asaj, line 300: Lots of duplicated between this and `computeTobinTax`, please call `getReserveRatio` in `computeTobinTax`

asaj (approved):

packages/protocol/contracts/stability/Reserve.sol
trianglesphere: Will do. I actually got the code to calc the reserve ratio from the tobin tax bit

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -299,11 +300,21 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
       uint256 aStableTokenValueInGold = stableTokenSupply.mul(goldAmount).div(stableAmount);
       stableTokensValueInGold = stableTokensValueInGold.add(aStableTokenValueInGold);
     }
+    return
+      FixidityLib
+        .newFixed(reserveGoldBalance)
+        .divide(FixidityLib.newFixed(stableTokensValueInGold))
+        .unwrap();
```
rcroessmann: @asaj @trianglesphere When designing the schedules and setting the cutoffs, we assumed that the reserve ratio is the total-reserve-value (including for example the BTC and ETH reserve funds) over stabletokens-value. Could we divide `reserveGoldBalance` by the cGLD weight in `assetAllocationWeights` to approximate the reserve ratio as we need it for these schedules and cutoffs to make sense? 

If we decide to do the above, we should make sure that the `unfrozenReserveGoldBalance` is used instead of `reserveGoldBalance` in the reserve ratio calculation as soon as #2390 is tackled (@timmoreton).

trianglesphere: That's actually what the tobin tax did in the first place (it did assume a 50% allocation into gold).
If it's hard enough to quickly compute ETH/BTC values in cGLD & the reserve stays close enough to it's target, using the target cGLD weight makes senses.

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -281,15 +290,20 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
...
-  function computeTobinTax() private view returns (FixidityLib.Fraction memory) {
+  function getReserveRatio() public view returns (uint256) {
     address sortedOraclesAddress = registry.getAddressForOrDie(SORTED_ORACLES_REGISTRY_ID);
     ISortedOracles sortedOracles = ISortedOracles(sortedOraclesAddress);
     uint256 reserveGoldBalance = getReserveGoldBalance();
     uint256 stableTokensValueInGold = 0;
+    FixidityLib.Fraction memory cgldWeight = FixidityLib.wrap(assetAllocationWeights["cGLD"]);
+    if (cgldWeight.equals(FixidityLib.wrap(0))) {
+      cgldWeight = FixidityLib.fixed1();
+    }
```
rcroessmann: This seems a bit dangerous.  How about requiring here that assetAllocationWeights["cGLD"]  is positive and making sure that it is set in migrations?

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -299,11 +313,22 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
...
+   * @notice Computes a tobin tax based on the reserve ratio.
+   * @return The numerator of the tobin tax amount, where the denominator is 1000.
+   */
+  function computeTobinTax() private view returns (FixidityLib.Fraction memory) {
     // The protocol calls for a 0.5% transfer tax on Celo Gold when the reserve ratio < 2.
-    // The protocol aims to keep half of the reserve value in gold, thus the reserve ratio
-    // is two when the value of gold in the reserve is equal to the total supply of stable tokens.
-    if (reserveGoldBalance >= stableTokensValueInGold) {
+    FixidityLib.Fraction memory ratio = FixidityLib.wrap(getReserveRatio());
+    if (ratio.gte(FixidityLib.newFixed(2))) {
```
rcroessmann, line 330: Could we make the cutoff (currently set to two) a governable parameter that one can query? Could be that we would even want to set this to zero initially and turn the tobin tax on later.

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -380,6 +380,27 @@ contract EpochRewards is Ownable, Initializable, UsingPrecompiles, UsingRegistry
     _updateTargetVotingYield();
   }
 
+  /**
+   * @notice Determines if the reserve is low enough to demand a diversion from
+   *    the community reward. Targets initial ratio of 2 with a linear decline
+   *    until 25 years have passed where the target ratio will be 1.
+   */
+  function isReserveLow() external view returns (bool) {
+    // Target reserve ratio = 2 - time in second / 25 years
```
rcroessmann: Nitty: Rethinking this, I find "target" a bit misleading as a term here - it is more a critical value. How about renaming this to criticalRatio?

packages/protocol/contracts/stability/Reserve.sol
trianglesphere: Good idea, ticket is here: https://github.com/celo-org/celo-monorepo/issues/2431

packages/protocol/contracts/stability/Reserve.sol
trianglesphere: This is tricky, but I think that b/c we're using unsigned ints/fixed point numbers, it can't be negative. The only case where it is 0 is when it does not have an associated weight (either not set yet or intentionally zero).

WRT setting it in migrations, if it's required to be set early on, I think that it'd make more sense to then move it to the initialize stage. I feel like it'd just be then set to 1 by default by most people. You're right that it feels dangerous, and I'm not quite sure the best approach here.

rcroessmann: Not sure I understand what you mean by "I feel like it'd just be then set to 1 by default by most people." The reserve cGLD weight should be the same for everybody and only modifiable via governance. 
It's hard for me to judge whether the initial weights should be set in migrations or in initialization stage - any thoughts @asaj?
If we make sure that the cGLD weight must be set and cannot be set to zero (or something negative), we should be able to get rid of the above if statement.

trianglesphere: RE cGLD weight being set, it's mainly for testing / start up. Long term the weights will exist and be updated more infrequently.

Making the cGLD weight privileged by having it be required does make sense from a stability sense, but will require more changes in how weights are handled (no thought is currently given to what the symbols for the weights are, just that they add up to 1, in addition, (@asaj), we probably need to ensure that the symbols in the input array are unique now that we have the mapping.

asaj (approved):

PR #2396, opened 2020-01-09
Reverting to old lockedgold amount and new voting amount

lucasege:
### Description

Changing back to Tim's previous method of scaling groups' locked gold linearly, so group 0 gets 50k locked gold and the value scales linearly to group 14 with 100k gold. (+~760 each group)

Changed the groups' voting to only vote with all of the groups' locked gold on the first and last group (one proxied, one not), while the intermediary groups only vote the minimum 10k. (Not sure if we want to increase this or not).

### Tested

CI Tests

### Other changes

Changed migrationsConfig vars back


asaj:
Why scale up the locked gold to 100k? If our goal is to have every validator easily displaced except for one proxied and one non-proxied validator this seems unnecessary

asaj (approved):

asaj:
Flaky tests are flaky, merging so we can deploy staging

PR #2399, opened 2020-01-09
Reading data to leaderboard sheet

### LINKED ISSUE #2385: LeaderBoard: Automate scripts
Make sure all the scripts that write to the DB and spreadsheet are automated or scheduled/alerted

mrsmkl:
### Description

Added script that loads participant names and BTUs to a spreadsheet with addresses.

### Tested

Only manual testing so far.

### Other changes

Incorporating some fixes to leaderboard script that were related to connections etc.

### Related issues

- Fixes #2385


mrsmkl:
The credentials and token have to be first generated locally (by running `leaderboard/src/upload.js`). For the token to work, we need to have an user that is able to write to the leaderboard sheet https://docs.google.com/spreadsheets/d/1Me56YkCHYmsN23gSMgDb1hZ_ezN0sTjNW4kyGbAO9vc/edit#gid=0

packages/leaderboard/src/board.ts
```diff
@@ -147,18 +144,28 @@ async function readAssoc(lst: string[]) {
...
+            metadata = IdentityMetadataWrapper.fromEmpty(a)
+          }
+        }
+        await processClaims(kit, a, metadata)
+      } catch (err) {
+        console.error('Bad address', a, err.toString())
+      }
+    })
+  )
+  client.end()
```
cmcewen, line 169: Do you think it would be more clear if client.connect and client.end were in the same function?

packages/leaderboard/src/upload.ts
```diff
@@ -0,0 +1,295 @@
...
+        console.log('Attestations requested', req, 'fulfilled', full, 'by', address)
+        if (req == 0) data.push(0)
+        else data.push(full / req)
+      } catch (err) {
+        console.error('Cannot resolve attestations for', address, err)
+        data.push('')
+      }
+    }
+  }
+  let req = {
```
cmcewen: seems like you could dedupe this with the above req and copy + spread only the `values: [data]` part

cmcewen (changes requested):
I also noticed that there is a separate secret just for the leaderboard, which seems to use the same values as the blockscout secret. Why not just use 1?

packages/leaderboard/src/board.ts
mrsmkl: This won't be very clear without using promises.

jcortejoso (approved):
lgtm

mrsmkl:
@jcortejoso can you check Connor's comment about two secrets?

cmcewen (approved):

PR #2403, opened 2020-01-09
Geo-Blacklist Announcements

aaronmgdr:
### Description

For legal reason it will be necessary to block our announcements from showing in certain countries, therefore this adds ip address based location lookups using an austrian based company. Also cache the announcement lookup. 

IP address will only be looked up when there is an announcement which has countries in the `block` field

### Tested

one unit test, 
ngrok to do a manual integration test

### Related issues

- fixes https://github.com/celo-org/celo-labs/issues/202

### Backwards compatibility

if no country is specified in airtable it works the same

Uses changes to `cache` interface made by https://github.com/celo-org/celo-monorepo/pull/2394


aaronmgdr:
we will need to add `ipstack.com` to our privacy policy with language indicating that your ip *might* be sent to this third party and that it cannot be opted out of since we need to block certain info based on jurisdiction

cmcewen (approved):

aaronmgdr:
dont merge yet

PR #2407, opened 2020-01-09
Remove references to OG infrastructure

### LINKED ISSUE #505: Devs SNBAT reference og infrastructure
### Expected Behavior

We should wholly remove "og", now that we have no actual infrastructure referencing it

### Current Behavior

We had some legacy setup with hard codede validator/tx node identities that we still kinda support, but mostly not really.

tkporter:
### Description

Removes the remaining code references to the old OG infrastructure.

### Tested

Ran some celotool commands to use the code that was changed. `yarn build` and `yarn lint` as well

### Other changes

The husky pre-commit hook to prettify files started complaining after I changed the validators.statefulset.yaml file because it doesn't like the helm templates. I changed the pre-commit prettify to only apply to the same files as `yarn prettify`, which doesn't include yaml files

### Related issues

- Fixes #505

### Backwards compatibility

OG infra is completely no longer supported


timmoreton (approved):
Beautiful!

PR #2416, opened 2020-01-10
Update genesis block accounts config to be correct for phase 2

nategraf:
### Description

Updates the `baklava_genesis_balances.json` to reflect the correct accounts for Baklava Phase 2 participants.


clrblmt (approved):
LGTM

PR #2418, opened 2020-01-10
Slashing params for stake off phase 2

timmoreton:
### Description

Set appropriate params for slashing for TGCSO

### Tested

No


packages/protocol/migrationsConfig.js
```diff
@@ -23,13 +23,13 @@ const DefaultConfig = {
...
-    penalty: '1000000000000000000000', // 1000 cGLD
+    reward: '20000000000000000000', // 20 cGLD
+    penalty: '100000000000000000000', // 100 cGLD
   },
   downtimeSlasher: {
     reward: '10000000000000000000', // 10 cGLD
-    penalty: '1000000000000000000000', // 1000 cGLD
-    slashableDowntime: 5,
+    penalty: '20000000000000000000', // 20 cGLD
+    slashableDowntime: 480,
```
asaj, line 32: Just a note that this cannot exceed 1 epoch (which it hasn't)

asaj (approved):

packages/protocol/migrationsConfig.js
timmoreton: yep! figured

nategraf (approved):
✅

PR #2420, opened 2020-01-10
Bump @celo/celocli version to 0.0.34

nategraf:
### Description

Following the publishing instructions in https://github.com/celo-org/celo-monorepo/tree/master/README-dev.md I am bumping the versions for `@celo/utils`, `@celo/contractkit` and `@celo/celocli`

I will publish:
* `@celo/celocli` version `0.0.33`
* `@celo/utils` version `0.1.5`
* `@celo/contractkit` version `0.2.9`


packages/blockchain-api/package.json
```diff
@@ -17,7 +17,7 @@
     "deploy": "./deploy.sh"
   },
   "dependencies": {
-    "@celo/contractkit": "0.2.0",
+    "@celo/contractkit": "0.2.10-dev",
```
jmrossy: blockchain api, notification service, and verification pool can only rely on published versions. A limitation of the AppEngine / Cloud Functions deployment process. Recommend just leaving their CK versions as is unless they need this upgrade. Or upgrade them to the latest published version (not -dev) but would be  good to test them before merging :)

cmcewen (approved):

PR #2433, opened 2020-01-14
Add unfreeze-contracts command to celotool

nategraf:
### Description

Add an `unfreeze-contracts` command to celotool to unfreeze (or freeze) the Exchange and EpochRewards contracts on demand.

### Tested

Manually against a local testnet and `baklavastaging`

### DIFF
--- a/packages/celotool/src/cmds/unfreeze_contracts.ts
+++ b/packages/celotool/src/cmds/unfreeze_contracts.ts
@@ -62,14 +62,14 @@ export const handler = async (argv: UnfreezeContractsArgv) => {
     ])
 
     for (const [name, contract] of Object.entries({ exchange, epochRewards })) {
-      if (contract == null) {
+      if (contract === null) {
         continue
       }
 
       if (argv.precheck) {
         const frozen = await contract.methods.frozen().call()
         // console.debug(`${name}.frozen = ${frozen}`)
-        if (argv.freeze == frozen) {
+        if (argv.freeze === frozen) {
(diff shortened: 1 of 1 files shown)


packages/celotool/src/cmds/unfreeze_contracts.ts
```diff
@@ -0,0 +1,114 @@
...
+      type: 'boolean',
+      description: 'Affect epoch rewards',
+      default: true,
+    })
+    .option('freeze', {
+      type: 'boolean',
+      description: 'Freeze contracts instead of unfreezing',
+      default: false,
+    })
+    .option('precheck', {
```
nambrot, line 38: Shouldn't we always do this?

packages/celotool/src/cmds/unfreeze_contracts.ts
```diff
@@ -0,0 +1,114 @@
...
+      type: 'boolean',
+      description: 'Check the contract freeze status before continuing',
+      default: true,
+    })
+    .option('verify', {
```
nambrot, line 43: Same, shouldn't we always do this?

nambrot (approved):

packages/celotool/src/cmds/unfreeze_contracts.ts
nategraf: I want this to be an option in case it causes some kind of issue.

packages/celotool/src/cmds/unfreeze_contracts.ts
nategraf: I want this to be an option in case it causes some kind of issue.

PR #2434, opened 2020-01-14
[Wallet] Rollback zeroSync toggle in case it was not successful

### LINKED ISSUE #2307: “Data saver” toggle is enabled without entering Security PIN
**Frequency:** 100%
**App version:** IOS test flight build v1.5.2 (17)
**Repro on:** iPhone 7 (13.3), iPhone XS Max (13.2), iPhone 7+ (12.4)

Pre-condition: Data saver toggle is disabled in initial state.  

**Repro Steps:** 
1) Launch the app. 
2) Tap on Cog icon. 
3) Tap on Data Saver option. 
4) Turn ON data saver option. 
5) Make sure Enter Security PIN screen is shown. 
6) Tap on the Back button on enter PIN screen. 
7) Observed. 

**Current Behavior:** User is able to enable the toggle without entering the PIN. 
**Expected Behavior:** User should not be able to enable the toggle without entering the PIN.
**Attachment:** [IOS_DataSaver_ToggleEnablewithoutPIN.mov](https://drive.google.com/file/d/1gjUngPWgf6A10soCyB2OQNS4qxAiZIVn/view?usp=sharing) (Bug observed at 10 sec)

i1skn:
### Description

If Data Saver mode was not changed successfully - it needs to be rolled back. In case of rollback we are initiating geth a second time - so we need restart the app. I've added a warning of potential restart, so it won't be a surprise for user.

### Tested

Android Device

### Related issues

- Fixes #2307

### Backwards compatibility

Yes


annakaz (approved):
🚢

packages/mobile/locales/es-419/accountScreen10.json
```diff
@@ -7,17 +7,21 @@
...
-    "El modo Data Saver te permite comunicarte con la Red Celo  a través de un nodo confiable. Puedes cambiar este modo en la configuración de la aplicación.",
-  "restartModal": {
+  "dataSaverDetail": "El modo Data Saver te permite comunicarte con la Red Celo  a través de un nodo confiable. Puedes cambiar este modo en la configuración de la aplicación.",
+  "restartModalSwitchOff": {
     "header": "Reiniciar para apagar Data Saver",
     "body": "Para encender y apagar Data Saver repetidamente, deberá reiniciar la aplicación.",
     "restart": "Reiniciar para alternar"
   },
+  "restartModalSwitchOn": {
+    "header": "~~Beware Of Restart",
```
jmrossy: translations

packages/mobile/src/account/DataSaver.tsx
```diff
@@ -82,16 +99,36 @@ export class DataSaver extends React.Component<Props, State> {
         >
           <Text style={fontStyles.body}>{t('enableDataSaver')}</Text>
         </SettingsSwitchItem>
-        <Modal isVisible={this.state.modalVisible}>
+        <Modal isVisible={this.state.switchOffModalVisible}>
```
jmrossy: Could this be one modal with varying content to avoid this duplication?

i1skn: It's not only the content, also callbacks. I think bunch of `()?:` would make it ugly.

packages/mobile/locales/es-419/accountScreen10.json
i1skn: Yeah, were waiting on approve of the approach and now will do the translations :)

jmrossy (approved):

Issue #299, opened 2019-08-04
Developers SBAT know that the oldest oracle report used by CP-DOTO hasn't expired


clrblmt:
### TODO

Check the oldest report isn't expired. If the oldest report *is* expired, `shouldUpdateBuckets()` should return false. 

Note (from sprint planning Dec 18): This does not need to remove expired reports from `SortedOracles`. Just avoid updating buckets if some reports are too stale.

**References:**
[Exchange.sol](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/Exchange.sol)
` function shouldUpdateBuckets() private view returns (bool) `

yerdua:
What exactly should happen here? I'm assuming that the ideal case is that an expired report is not affecting the median, and therefore isn't used to set the buckets.

Currently, every function in the path for getting the updated buckets is a `view` function. In order to remove expired reports and stop CP-DOTO from using them, `removeExpiredReports` in `SortedOracles.sol` needs to be called. This is a function that requires gas.

BatiGencho:
@asaj Reserved for Ventsi @Limechain

PR #2435, opened 2020-01-14
Developers SBAT know that the oldest oracle report used by CP-DOTO hasn't expired #299

Holymir:
### Description

Developers SBAT know that the oldest oracle report used by CP-DOTO hasn't expired #299

### Tested

unit test


packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -118,6 +118,21 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
     }
   }
 
+  /**
+   * @notice Check if last report is expired.
+   * @param token The address of the token for which the Celo Gold exchange rate is being reported.
+   */
+  function isReportActive(address token) external view returns (bool) {
```
yerdua: The purpose of this function would be more clear with a name that includes the word "expired", like: `isOldestReportExpired`, `expiredReportsExist`, `oldestReportExpired`, etc

Unless we're planning to introduce other qualities that could make a report "active" or "inactive", I think we should stick with "expired" as the adjective here.

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -118,6 +118,21 @@ contract SortedOracles is ISortedOracles, Ownable, Initializable {
...
+    require(token != address(0));
+    address oldest = timestamps[token].getTail();
+    uint256 timestamp = timestamps[token].getValue(oldest);
+    // solhint-disable-next-line not-rely-on-time
+    if (now.sub(timestamp) >= reportExpirySeconds) {
+      return false;
+    }
+    return true;
```
yerdua: Since this is the same as the check that's used in `removeExpiredReports`, could this reasonably be used inside that function?

Are there any complexities with `external` vs `public` that make code duplication preferable?

yerdua (changes requested):

packages/protocol/contracts/stability/SortedOracles.sol
asaj: +1 to `isOldestReportExpired`

packages/protocol/contracts/stability/SortedOracles.sol
asaj: +1, `removeExpiredReports` can remove the oldest reports until `isOldestReportExpired` returns false

packages/protocol/test/stability/sortedoracles.ts
```diff
@@ -158,6 +159,41 @@ contract('SortedOracles', (accounts: string[]) => {
     })
   })
 
+  describe('#isOldestReportExpired', () => {
+    beforeEach(async () => {
+      await sortedOracles.addOracle(aToken, anOracle)
+    })
+
+    it('should return true if there are no reports', async () => {
+      let isReportExpired = await sortedOracles.isOldestReportExpired(aToken)
```
asaj, line 169: `const`

asaj (approved):

yerdua (approved):

PR #2436, opened 2020-01-14
Downtime slashing when epoch changes

mrsmkl:
### Description

Looks like there is some error when checking downtime when validator index changes. This PR should fix it.

### Tested

Added a unit test.


packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -83,7 +83,7 @@ contract DowntimeSlasher is SlasherUtil {
       uint256 signerIndex = epochNumberOfBlock(n, sz) == startEpoch
         ? startSignerIndex
         : endSignerIndex;
-      if (uint256(getParentSealBitmap(n)) & (1 << signerIndex) != 0) return false;
+      if (uint256(getParentSealBitmap(n + 1)) & (1 << signerIndex) != 0) return false;
```
mcortesi, line 88: So, not sure of the semantics of `getParentSealBitmap`. 
From your change it seems that `getParentSeal(x)` will get the parentSeal that's on block X header. which is a seal of all signers for block X-1.

In that case, this is correct.. and great catch by the way!
Can you confirm? The other case is that getParentSealBitmap(x) get's the parent seal that have signer for block X and that lives on block X+1 header..

In any case, a comment would be nice :D

mrsmkl: I think I'll check from geth, but I tried to check from baklava, the current version didn't seem to work.

mrsmkl: It gets the header from the block given as argument https://github.com/celo-org/celo-blockchain/blob/master/core/vm/contracts.go#L797

So it should be correct now.

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -83,7 +83,9 @@ contract DowntimeSlasher is SlasherUtil {
       uint256 signerIndex = epochNumberOfBlock(n, sz) == startEpoch
         ? startSignerIndex
         : endSignerIndex;
-      if (uint256(getParentSealBitmap(n)) & (1 << signerIndex) != 0) return false;
+      // We want to check signers for block n,
+      // so we get the parent seal bitmap for the next block
+      if (uint256(getParentSealBitmap(n + 1)) & (1 << signerIndex) != 0) return false;
```
asaj, line 88: Do we need to change `    require(endBlock < block.number - 1, "end block must be smaller than current block");`?

asaj (approved):

packages/protocol/contracts/governance/DowntimeSlasher.sol
mrsmkl: No, I think it's correct now.

Issue #2438, opened 2020-01-14
Lookup contact details in escrow payment reclaim components


jmrossy:
### Expected Behavior

EscrowedPaymentListItem and ReclaimPaymentConfirmationScreen should look up the recipient number in the recipientCache and show a name instead of a number if possible.

### Current Behavior

No lookup, we just show the phone number

tarikbellamine:
Duplicate of #4957

PR #2439, opened 2020-01-14
[Wallet] Cleanup unused StateProps references

jmrossy:
### Description

Cleanup unused stateprops references to the e164<->address mappings and remove unused contact field in escrow payments to avoid confusion

No functional changes, just cleanup

### Tested

Ran tests

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
+++ b/packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
@@ -69,7 +69,8 @@ export class EscrowedPaymentListItem extends React.PureComponent<Props> {
 
   getDisplayName() {
     const { payment } = this.props
-    return payment.recipientContact ? payment.recipientContact.displayName : payment.recipientPhone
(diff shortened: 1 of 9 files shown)


packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
```diff
@@ -69,7 +69,8 @@ export class EscrowedPaymentListItem extends React.PureComponent<Props> {
 
   getDisplayName() {
     const { payment } = this.props
-    return payment.recipientContact ? payment.recipientContact.displayName : payment.recipientPhone
```
jmrossy, line 72: the EscrowPayment recipientContact field was always undefined. And we shouldn't use it, we should be looking up the contact info in the recipient cache each time to get updates. Removing that field and its references to avoid confusion and opened https://github.com/celo-org/celo-monorepo/issues/2438

annakaz (approved):
👍

PR #2440, opened 2020-01-14
Add feeCurrency to base command flags

yorhodes:
### Description

### Tested

### Other changes

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/packages/cli/src/base.ts
+++ b/packages/cli/src/base.ts
@@ -1,4 +1,4 @@
-import { ContractKit, newKitFromWeb3 } from '@celo/contractkit'
+import { CeloContract, ContractKit, newKitFromWeb3 } from '@celo/contractkit'
 import { CeloProvider } from '@celo/contractkit/lib/providers/celo-provider'
 import { Command, flags } from '@oclif/command'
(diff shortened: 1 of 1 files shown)


packages/cli/src/base.ts
```diff
@@ -73,6 +74,13 @@ export abstract class BaseCommand extends LocalCommand {
     if (res.flags && res.flags.privateKey) {
       this._kit.addAccount(res.flags.privateKey)
     }
+    if (res.flags && res.flags.feeCurrency) {
+      const token =
+        res.flags.feeCurrency === 'cUSD' ? CeloContract.StableToken : CeloContract.GoldToken
+      return this._kit.setFeeCurrency(token).then(() => {
```
nambrot, line 80: Does this work? I don't think consumers are expecting a promise. Unfortunately it looks like we'd have to update all call sites?

yorhodes: hmm I thought this was a hack to get around returning a non-promise but I guess not

yorhodes:
closing while this is not possible

PR #2443, opened 2020-01-15
[Wallet] Add 10.0.2.2 to domain-config -> cleartextTrafficPermitted

i1skn:
### Description

When app starts in Emulator - I would need to manually set bundle location to `localhost:8081`, otherwise it would not load the bundle. This PR adds `10.0.2.2`(Special alias to your host loopback interface (i.e., 127.0.0.1 on your development machine) - https://developer.android.com/studio/run/emulator-networking) to the allowed addresses.

### Tested
Android emulator

### Backwards compatibility

Yes


packages/mobile/android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -1,11 +1,12 @@
 <network-security-config> 
   <domain-config cleartextTrafficPermitted="true">
     <domain includeSubdomains="true">localhost</domain>
+    <domain includeSubdomains="true">10.0.2.2</domain>
```
jmrossy, line 4: Good catch! I also just discovered this solution yesterday to unblock the e2e tests. Sorry I didn't share earlier.
But I believe we should remove this file and instead have this network_security_config be only in the `app/src/DEBUG/res/xml` folder so it only affects debug builds

jeanregisser:
Cool! 
Do you know why we didn't need that previously? or what did we change that made this necessary?

packages/mobile/android/app/src/main/res/xml/network_security_config.xml
i1skn: I will test that!

i1skn:
I'm not sure, but I think it's either RN changed `localhost` to `10.0.2.2` in the bundle URL identification or this IP was somehow allowed by default in gradle build before.

i1skn:
Not needed anymore

Issue #1261, opened 2019-10-09
Users SBAT see switched currency order on home screen


nityas:
### Expected Behavior

[Figma](https://www.figma.com/file/Pp1N48FwfpBzb4b2ggpsjR/Local-Currency?node-id=0%3A1) (includes examples)

**Wallet Balance**
<img width="2172" alt="balance" src="https://user-images.githubusercontent.com/51388535/66526502-c3feac80-eaad-11e9-927c-be321c8b9b95.png">
Changes:
* "Celo Dollar" text removed
* Local currency and CUSD swapped in priority

Note: The local currency mockup shows both the currency symbol (€) and acronym (EUR). This is a bit redundant for Euro, but not so with $/MXN. 
Some options: 
* Only show acronym
* Be dynamic based on country (sometimes show symbol, sometimes show both)
* Always show both

**Transaction Feed** 
<img width="2451" alt="feed" src="https://user-images.githubusercontent.com/51388535/66444816-8b050000-e9f9-11e9-91bd-1db7acda5d30.png">
Changes:
* CUSD value is removed
* Local currency elevated
* Currency symbol added

**Transaction Drilldown**
<img width="1600" alt="drilldown" src="https://user-images.githubusercontent.com/51388535/66444811-85a7b580-e9f9-11e9-8165-3dda3bdbf1ab.png">
Changes:
* CUSD balance added for Non-USD local currency
* Text weight for text displaying comment went from Light -> Regular

### Current Behavior
- cUSD on top
- both currencies in feed
- no cUSD in drilldown

coreycelo:
Added designs.

@nityas Check out note under Wallet Balance above.

Pedro-vk:
@nityas @jmrossy 
The scope of this user story is "_Users SBAT see switched currency order on home screen_" or all the changes that are in the description?

nityas:
@Pedro-vk the scope is the changes in the description- thanks for clarifying!

nityas:
Noticed on the latest build that the order at the top of the home screen is switched. However, the transaction feed and drilldown changes have not been implemented, reopening

PR #2446, opened 2020-01-15
[Wallet] Historical currency conversions in the transaction feed

### LINKED ISSUE #834: Currency conversion in activity feed should use historical rate

### LINKED ISSUE #1386: [Android] [Random]“Payment sent” transaction is shown twice on “Activity” Section from “Wallet” tab when user proceed for payment while device network connectivity is low.
**Frequency:** 40%
**App version:** zero-sync build: Staging build v1.5.0 (101548577)
**Repro on:**  Galaxy Grand Prime (5.1.1), Samsung Galaxy A8+ (8.0.0)
**Testing Account:**  680 208 0265/Backup Kay

**Pre-condition:** Make sure device network connectivity should be low i.e 2G/3G
 
**Repro Steps:** 
1) Launch the app. 
2) Tap on + tab. 
3) Select any celo contact user. 
4) Proceed for sending cUSD. 
5) Make sure “payment Sent ” Confirming...”  message is shown on the Activity section from the wallet tab. 
6) Wait for a moment. 
7) Observe.
 
**Investigation:** 
1) Issue does not reproduce if device network connectivity is good. 
2) One of Payment Sent section is shown with Confirming… message and another is shown as payment sent with date and time. 
3) After cold launch the app issue Payment sent with confirming message is get removed from the Activity section.
 
**Impact:** User may get confused and assumed that Payment is not sent. 
**Current Behavior:** “Payment sent” section is shown twice 
**Expected Behavior:** Payment sent section is shown only once if user proceed it for only one time. 
**Attachment:** Android_PaymentSentShownTwice.png 
Android_PaymentSentShownTwice

jeanregisser:
### Description

This PR enables historical currency conversions in the transaction feed.

Main highlights:
- Implemented the updated schema in `blockchain-api` which allows querying transactions only for a specific token (cUSD on the home screen, cGLD on the exchange screen), retrieving
amounts in local currency, and later on will make it easy to paginate the data as proposed in https://docs.google.com/document/d/1ryfz7Gd0bk9QQHDBCWyaBy2AeOui63hgp-HdFtyBTyc/edit
- Fixed the broken `gql-gen` tool and made GraphQL related code type safe again by using `graphql-codegen` for the client. (we should use it for `blockchain-api` too)
- Started moving displayed money amounts into the `CurrencyDisplay` component. This component takes care of doing the actual currency conversions when needed. It will evolve as we move more of the existing code to it. It already simplified the feed items components.
- Refactored code to use the same `TransactionsList` component for both the home and exchange screens and added tests for it.

We should probably run `graphql-codegen` as part of our React Native dev startup script to spot any type issues with the `blockchain-api` we're connecting to. I'll address that in a separate PR.

In the `ExchangeFeedItem` fragment, the `amount` field might feel redundant as it's either the `makerAmount` or the `takerAmount` with the addition of the sign (+/-), which we could derive client side. My main goal was to make the interface right first and optimize later when needed.

The `TransactionsList.text.tsx` test outputs console logs about invalid/missing fragments, this is caused by the Apollo `MockProvider` context missing details about the `interface` type we're using. I tried to pass the right `fragmentMatcher` like we do in the normal client (see https://www.apollographql.com/docs/react/data/fragments/#fragments-on-unions-and-interfaces)
but it then failed to return any data to the querying component.
In short, the test works as expected but outputs a lot of misleading noise.

Finally, this PR looks a bit big, but there are a lot of simple renames and snapshots changes, so hopefully shouldn't be too hard to review 🙏 

### Tested

- Confirmed that converted local amounts in the transaction feed now use historical rates.
- Added new tests.

### Other changes

- Currency conversion API now outputs a `string` instead of a `number` to prevent rounding issues.
- The amounts in the transactions API are now `string` as well for the same reason stated above.
- Memoized gold exchange rates queries (per requests).
- Fixed a redux state update that was done in a non immutable way.
- I had a problem with the test suite when running with `--runInBand` like in the CI and it was caused by the apollo query component not being unmounted after the test.
I upgraded `react-native-testing-library` to be able to cleanup mounted components after testing. See https://github.com/callstack/react-native-testing-library/blob/master/docs/API.md#cleanup
I decided to make that behavior global for all tests as I feel it's the right thing to do for us.
See https://github.com/celo-org/celo-monorepo/pull/2446/commits/b99484a90b6465ffcbbdddf5c1fb8585902322d8
I just had to change one test which relied on a component being mounted between test.
See https://github.com/celo-org/celo-monorepo/pull/2446/commits/6a9031067ad2f89c428c6f917dce57b7f36b4e6d
Let me know if you have any concern with this.

### Related issues

- Fixes #834
- Fixes #1386
- #1261
- Starts addressing #1484

### Backwards compatibility

Yes.


packages/blockchain-api/src/blockscout.ts
```diff
@@ -214,6 +218,107 @@ export class BlockscoutAPI extends RESTDataSource {
...
+          outEvent = transactions[1]
+        } else {
+          inEvent = transactions[1]
+          outEvent = transactions[0]
+        }
+
+        // Find the event related to the queried token
+        const tokenEvent = [inEvent, outEvent].find((event) => event.tokenSymbol === args.token)
+        if (tokenEvent) {
+          const timestamp = new BigNumber(inEvent.timeStamp).toNumber()
```
i1skn: would `parseInt` do the job here?

packages/blockchain-api/src/currencyConversion/GoldExchangeRateAPI.ts
```diff
@@ -56,7 +57,19 @@ export default class GoldExchangeRateAPI<TContext = any> extends DataSource {
     const date = timestamp ? new Date(timestamp) : new Date()
 
     const pair = `${sourceCurrencyCode || CUSD}/${currencyCode}`
+    const cacheKey = `${pair}-${date.getTime()}`
 
+    let promise = this.memoizedResults.get(cacheKey)
```
i1skn, line 62: What do you think about:
```
    let promise = this.memoizedResults.get(cacheKey)
    if (!promise) {
       promise = this.performRequest(pair, date)
       this.memoizedResults.set(cacheKey, promise)
    }
    return promise
```
?

packages/mobile/src/send/TransferConfirmationCard.tsx
```diff
@@ -67,56 +60,47 @@ const renderTopSection = (props: Props) => {
...
+    case TransactionType.NetworkFee:
       return (
-        <MoneyAmount
-          symbol={CURRENCIES[currency].symbol}
-          amount={getNetworkFeeDisplayValue(value, true)}
+        <CurrencyDisplay
+          type={DisplayType.Big}
+          amount={amount}
+          // tslint:disable-next-line: jsx-no-lambda
+          formatAmount={(value) => getNetworkFeeDisplayValue(value, true)}
```
i1skn: Is there particular reason to use lambda function here?

packages/mobile/src/transactions/TransactionFeed.tsx
```diff
@@ -38,137 +28,95 @@ interface StateProps {
...
-        if (tx.comment === SENTINEL_INVITE_COMMENT) {
-          if (tx.type === TransactionTypes.SENT) {
-            tx.type = TransactionTypes.INVITE_SENT
-          } else if (tx.type === TransactionTypes.RECEIVED) {
-            tx.type = TransactionTypes.INVITE_RECEIVED
-          }
+    const { addressToE164Number, invitees, recipientCache } = this.props
+
+    if (tx.hasOwnProperty('comment')) {
+      // @ts-ignore
```
i1skn: Is it possible to use`if (tx.__typename === 'TransactionTransfer') {` to avoid `// @ts-ignore` here?

packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -0,0 +1,291 @@
...
+`
+
+class UserTransactionsComponent extends Query<
+  UserTransactionsQuery,
+  UserTransactionsQueryVariables
+> {}
+
+const mapStateToProps = (state: RootState): StateProps => ({
+  address: currentAccountSelector(state),
+  standbyTransactions: state.transactions.standbyTransactions,
```
i1skn, line 70: what is the convention about selectors, do we create them only when some logic is involved to encapsulate it?

packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -0,0 +1,291 @@
...
+      standbyTransactions,
+    } = this.props
+
+    const queryAddress = address || ''
+    const token = currency === CURRENCY_ENUM.GOLD ? Token.CGld : Token.CUsd
+
+    return (
+      <UserTransactionsComponent
+        query={TRANSACTIONS_QUERY}
+        pollInterval={10000}
```
i1skn: Can we move it to `const`?

i1skn:
Great work!!! See a couple of non-blocker comments I've left :)

packages/blockchain-api/src/blockscout.ts
jmrossy: We generally opt for BigNum over parseInt

packages/blockchain-api/src/currencyConversion/GoldExchangeRateAPI.ts
jeanregisser: Yep better thanks!

packages/mobile/src/send/TransferConfirmationCard.tsx
jeanregisser: No, let me change this 👍

packages/mobile/src/transactions/TransactionFeed.tsx
jeanregisser: Yes, will change it 👍

packages/mobile/src/transactions/TransactionsList.tsx
jeanregisser: 👍

packages/mobile/src/transactions/TransactionsList.tsx
jeanregisser: I usually introduce them when it's needed in more than 2 places. Or If I need reselect.

packages/blockchain-api/src/blockscout.ts
i1skn: Did not know that. Is the reason because `BigNumber` is more universal and can handle different input formats compare to `parseInt`?

i1skn (approved):
Thanks for the tremendous improvement 🚢 !

packages/blockchain-api/src/blockscout.ts
```diff
@@ -214,6 +218,107 @@ export class BlockscoutAPI extends RESTDataSource {
     )
     return rewards.sort((a, b) => b.timestamp - a.timestamp)
   }
+
+  async getTransactions(args: TransactionArgs) {
```
jmrossy: Can you leave a comment or rename this to clarify how it differs from the existing `getTokenTransactions` and  `getFeedEvents` methods?

packages/blockchain-api/src/blockscout.ts
```diff
@@ -214,6 +218,107 @@ export class BlockscoutAPI extends RESTDataSource {
...
+
+    // Mapping to figure out what event each raw transaction belongs to
+    const txHashToEventTransactions = new Map<string, any>()
+    for (const tx of rawTransactions) {
+      const currentTX = txHashToEventTransactions.get(tx.hash) || []
+      currentTX.push(tx)
+      txHashToEventTransactions.set(tx.hash, currentTX)
+    }
+
+    await this.ensureTokenAddresses()
```
jmrossy, line 251: Not a new change in this PR but we should consider reworking this bit or at least leave a comment. This design of forcing users of token addresses to first call this method to ensure they're set is problematic. Easy to mess up. Would be better if the token addresses had getters that fetched them when null. cc @annakaz

packages/blockchain-api/src/blockscout.ts
```diff
@@ -214,6 +218,107 @@ export class BlockscoutAPI extends RESTDataSource {
...
+    // Generate final events
+    txHashToEventTransactions.forEach((transactions: BlockscoutTransaction[], txhash: string) => {
```
jmrossy, line 253: Can we DRY this up with the `getFeedEvents` method above? Or is one being deprecated?

packages/blockchain-api/src/schema.ts
```diff
@@ -106,26 +217,57 @@ export const typeDefs = gql`
...
   Query: {
-    events: async (_source: any, args: EventArgs, { dataSources }: Context) => {
+    events: async (_source: any, args: EventArgs, context: Context) => {
+      const { dataSources } = context
       return dataSources.blockscoutAPI.getFeedEvents(args)
     },
     rewards: async (_source: any, args: EventArgs, { dataSources }: Context) => {
       return dataSources.blockscoutAPI.getFeedRewards(args)
     },
+    transactions: async (_source: any, args: TransactionArgs, context: Context) => {
```
jmrossy: My problem with `transaction` is that it's too vague. It could be anything from an ERC20 token transfer to a verification request to a smart contract deployment. Is there a more descriptive name we could use?

packages/mobile/src/apollo/types.ts
```diff
@@ -1,54 +1,328 @@
...
-  id?: string
-  hash?: string
-  value: number
-  symbol: CURRENCY_ENUM
-  timestamp: number
-  address: string
-  comment: string
-  status?: TransactionStatus
-  // TODO: fee needs to be added here
+export type Event = Exchange | Transfer
```
jmrossy, line 28: I guess you kept the old Event interface for backwards compatibility with older app versions? But the next time we ship the app, we'll be resetting the network too and forcing users to update. We could actually remove all that.

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -2,61 +2,128 @@ import colors from '@celo/react-components/styles/colors'
...
-  amount: BigNumber
-  size: number
-  type: CURRENCY_ENUM
-  currencySymbol?: LocalCurrencySymbol | null
+  type: DisplayType
+  amount: MoneyAmount
+  size: number // only used for DisplayType.Big
+  useColors: boolean
+  hideSymbol: boolean
+  formatAmount: (amount: BigNumber.Value) => string
```
jmrossy, line 23: Can the formatAmount be determined from the DisplayType? Or maybe made optional? Just wondering if we can keep as much currency formatting related logic encapsulated in this file as possible.

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -2,61 +2,128 @@ import colors from '@celo/react-components/styles/colors'
...
+}
+
 interface Props {
-  amount: BigNumber
-  size: number
-  type: CURRENCY_ENUM
-  currencySymbol?: LocalCurrencySymbol | null
+  type: DisplayType
+  amount: MoneyAmount
+  size: number // only used for DisplayType.Big
```
jmrossy, line 20: what is size?

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -2,61 +2,128 @@ import colors from '@celo/react-components/styles/colors'
...
-  symbolStyle(fontSize: number) {
-    const size = Math.floor(fontSize * symbolRatio)
-    return {
-      fontSize: size,
-      color: this.color(),
-      lineHeight: Math.round(size * 1.4),
-      transform: [{ translateY: Math.round(size * 0.1) }],
-    }
-  }
+// TODO(Rossy) This is mostly duped by MoneyAmount, converge the two
```
jmrossy, line 39: Does this PR solve this TODO?

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -2,61 +2,128 @@ import colors from '@celo/react-components/styles/colors'
...
+  type,
+  size,
+  useColors,
+  hideSymbol,
+  amount,
+  formatAmount,
+  style,
+}: Props) {
+  const localCurrencyCode = useLocalCurrencyCode()
+  // tslint:disable-next-line: react-hooks-nesting
```
jmrossy, line 50: why is this triggering?

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -2,61 +2,128 @@ import colors from '@celo/react-components/styles/colors'
...
+  return (
+    <Text numberOfLines={1} style={[styles.container, { color }, style]}>
+      {sign}
+      {!hideSymbol && currencySymbol}
+      {formattedValue}
+    </Text>
+  )
+}
+
+CurrencyDisplay.defaultProps = {
```
jmrossy, line 110: If we're setting these default props, should these values be made optional with `?` in the interface above?

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -2,61 +2,128 @@ import colors from '@celo/react-components/styles/colors'
...
+  type: DisplayType.Default,
+  size: 48,
+  useColors: true,
+  hideSymbol: false,
+  formatAmount: (amount: BigNumber.Value) => getMoneyDisplayValue(amount),
 }
 
 const styles = StyleSheet.create({
-  container: {
+  container: {},
```
jmrossy: nit : remove

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
```diff
@@ -42,7 +33,7 @@ const getExchangeRate = (props: Props) => {
   }
 
   // For feed drilldown, the exchange rate has not been provided
-  return makerAmount.dividedBy(takerAmount)
+  return new BigNumber(makerAmount.amount).dividedBy(takerAmount.amount)
```
jmrossy: Per the thread you started on Slack, would be cool if we could encapsulate more logic in MoneyAmount and avoid this pattern :)

packages/mobile/src/firebase/notifications.ts
```diff
@@ -59,17 +59,20 @@ function* handlePaymentReceived(
...
 
     navigateToPaymentTransferReview(
-      TransactionTypes.RECEIVED,
+      TransactionType.Received,
       new BigNumber(transferNotification.timestamp).toNumber(),
       {
-        value: divideByWei(transferNotification.value),
-        currency: resolveCurrency(transferNotification.currency),
+        amount: {
+          amount: divideByWei(transferNotification.value).toString(),
```
jmrossy: Another common source of confusion is whether or not a value is in wei or not. You don't need to fix in this PR but would be cool if we had a wei to type that and avoid potential mistakes.

packages/mobile/src/redux/migrations.test.ts
```diff
@@ -68,4 +68,16 @@ describe('Redux persist migrations', () => {
     expect(migratedSchema.identity.startedVerification).toBeUndefined()
     expect(migratedSchema.identity.hasSeenVerificationNux).toBeTruthy()
   })
+
+  it('work for v4 to v5', () => {
```
jmrossy, line 72: I don't believe v4 has shipped to users yet so no need for a v5. But it's fine either way.
In fact, we should blow these away and start fresh before the next deployment because we can only do that when the update coincides with a network reset.

packages/mobile/src/transactions/ExchangeFeedItem.tsx
```diff
@@ -4,161 +4,44 @@ import colors from '@celo/react-components/styles/colors'
...
-  const amountProps = showGoldAmount
-    ? getGoldAmountProps(exchangeProps)
-    : getDollarAmountProps(exchangeProps)
-
-  const { inValue, outValue } = exchangeProps
-  const { displayAmount, amountDirection, amountColor } = amountProps
-
+  const icon =
+    makerAmount.currencyCode === CURRENCIES[CURRENCY_ENUM.GOLD].code
+      ? require('src/transactions/ExchangeGoldGreen.png')
```
jmrossy: please move these `require` calls to src/images/Images.ts and then use the exported values here

packages/mobile/src/transactions/TransactionsList.test.tsx
```diff
@@ -0,0 +1,195 @@
+import * as React from 'react'
```
jmrossy, line 1: Nice testing here 🥇

packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -0,0 +1,310 @@
...
+}
+
+interface DispatchProps {
+  removeStandbyTransaction: typeof removeStandbyTransaction
+}
+
+type Props = OwnProps & StateProps & DispatchProps
+
+// See https://github.com/microsoft/TypeScript/issues/16069#issuecomment-565658443
+function isPresent<T>(t: T | undefined | null | void): t is T {
```
jmrossy: Consider moving this to a shared place. Like utils/typescript.ts

jmrossy (commented):
Overall really nice work! Some comments below but nothing major

packages/blockchain-api/src/blockscout.ts
jeanregisser: Yes! `getFeedEvents` was kept for backward compatibility, it is intended to be removed at some point.

packages/blockchain-api/src/blockscout.ts
jeanregisser: yes `getFeedEvents` is being deprecated.

packages/blockchain-api/src/schema.ts
jeanregisser: Makes sense, `tokenTransactions` feels like the right name then, what do you think?

packages/mobile/src/apollo/types.ts
jeanregisser: Indeed! Let's remove it then.

packages/mobile/src/components/CurrencyDisplay.tsx
jeanregisser: Though it's not visible here, it's optional (see `defaultProps`) and used to format some network fee amounts. 
DisplayType is currently used for determining how we display the symbol (superscript vs normal, maybe I should rename that too).
I thought about providing another enum for the format, but decided to leave it as is with a default implementation until I refactor more of the code related to local currencies.
I agree it should encapsulate all formatting logic too.

packages/mobile/src/components/CurrencyDisplay.tsx
jeanregisser: Not entirely, I'll move existing `MoneyAmount` usage in a follow up PR related to a  more holistic solution for displaying currencies.

packages/mobile/src/components/CurrencyDisplay.tsx
jeanregisser: It's the fontSize, somewhat redundant with the style property. I want to address that in a follow up PR.

packages/mobile/src/components/CurrencyDisplay.tsx
jeanregisser: TypeScript does this automatically when consuming this component from JSX.

packages/mobile/src/transactions/TransactionsList.tsx
jeanregisser: Yep!

packages/mobile/src/components/CurrencyDisplay.tsx
jeanregisser: It's because of the conditional at the end `|| xxx`.

packages/mobile/src/redux/migrations.test.ts
jeanregisser: Ok yes I'll leave them and we can reset all of these in another PR.

packages/mobile/src/firebase/notifications.ts
jeanregisser: Yep, I will definitely continue thinking about this as we improve currencies handling.

packages/blockchain-api/src/blockscout.ts
jeanregisser: Also I suspect this call is slowing down queries given a new data source is instantiated for each request.

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
jeanregisser: Yes we can definitely explore that.
I like the simplicity of having plain objects and functions operating on them. But money amount is an essential block in our app and we might benefit encapsulating it in its own class 👍

packages/blockchain-api/src/blockscout.ts
jmrossy: exactly

packages/blockchain-api/src/blockscout.ts
jmrossy: Ok. Up to you then if you want to remove it now or later. Please leave TODOs explaining this

packages/blockchain-api/src/schema.ts
jmrossy: Sounds good

packages/mobile/src/apollo/types.ts
jeanregisser: I'll remove the deprecated code after this PR is merged to prevent breaking things unnecessarily given `blockchain-api` has to be redeployed.

packages/blockchain-api/src/schema.ts
jeanregisser: done 👍

jmrossy (approved):

PR #2450, opened 2020-01-15
Configure the time to wait for text messages

nambrot:
### Description

We hard coded to 10 minutes which is a little longer than we need to. Most of the time, either people complete attestations quite promptly, or it never arrives. So this PR makes that configurable.

### Tested

- Not tested


packages/celotool/src/cmds/bots/auto-verify.ts
```diff
@@ -59,6 +60,11 @@ export const builder = (yargs: Argv) => {
       description: 'The node to use',
       default: 'http://localhost:8545',
     })
+    .option('timeToPollForTextMessages', {
+      type: 'number',
+      description: 'How long to poll for text messages',
```
nategraf: Units
```suggestion
      description: 'How long to poll for text messages in minutes',
```

nategraf (approved):

PR #2456, opened 2020-01-16
Add validator:signed-blocks command and fix validator:status bug

nategraf:
### Description

The `validator:status` command currently makes an incorrect assumption that the index of a validator in the election list is constant across epochs. This PR fixes that assumption and adds a `validator:signed-blocks` command to cover a need in seeing which blocks are or are not signed.

### Tested

Manual testing


timmoreton (approved):
Nice! Can I suggest `validator:signed` or similar?  I can see a use for something like `validator:heartbeat` as a command which just returns an exit status if it's not elected or not signed the last block -- for folks to integrate for monitoring purposes. 

In both commands, having flags to fill these use cases would be helpful:
* all blocks in current epoch
* all blocks in previous complete epoch 
* all blocks in a given epoch, e
* last b blocks

I can also foresee a request for calculating uptime for an epoch based on this... but it is accessible via the end of epoch block (somehow!)

Also you might need to add some documentation and rebuild the docs too.

PR #2457, opened 2020-01-16
[Wallet] Fix type check regression for components wrapped by our custom `withTranslation`

jeanregisser:
### Description

Fixes TypeScript failing to type check components wrapped by our custom `withTranslation` at call site (i.e. in JSX).
Their props were considered to be `any`.
Here we reuse the original implementation types to fix it.

### Tested

Tested that errors are raised again when passing props that are of the wrong types or missing when using a component wrapped by our custom `withTranslation` helper

### Other changes

None

### Related issues

Discussed on Slack.

### Backwards compatibility

Yes.


jmrossy (approved):
Nice!!

PR #2468, opened 2020-01-17
Update copyright year + inline button 

aaronmgdr:
### Description

copyright year is now set automatically to current year

inline button no longer will have its own size or font, instead getting from the Text it is inline with. 

### Tested

used snapshots to see which pages were changed and then visually checked they looked right

some examples...

Screen Shot 2020-01-17 at 10 50 24 AM
Screen Shot 2020-01-17 at 10 50 48 AM
Screen Shot 2020-01-17 at 10 51 04 AM
Screen Shot 2020-01-17 at 10 51 35 AM
Screen Shot 2020-01-17 at 10 52 10 AM
Screen Shot 2020-01-17 at 11 01 10 AM

### Related issues

Fixes https://github.com/celo-org/celo-labs/issues/222
fixes https://github.com/celo-org/celo-labs/issues/221


packages/web/src/shared/Footer.3.tsx
```diff
@@ -127,6 +127,8 @@ const Navigation = React.memo(function _Navigation({
   )
 })
 
+const YEAR = new Date().getFullYear()
```
jmrossy, line 130: why the UPPERCASE here?

jmrossy (approved):
One nit but otherwise lgtm

packages/web/src/shared/Footer.3.tsx
aaronmgdr: often when i declare a constant outside of a function I do so in upper case, a habit leftover from the days of ruby, but i think it's common in JS land too

PR #2478, opened 2020-01-20
Add Validators list page to website

Pedro-vk:
### Description

_Add Validators list page to website_

### Tested

_Tested on chrome_

**Note:** I used `/build/validators` path.

image


packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+    return celoValidatorGroups
+      .map(({ account, affiliates, votes }) => {
+        const group = account
+        const gName = group.name
+        return {
+          name: (gName || '').length <= 20 ? gName : `${gName.substr(0, 17)}...`,
+          address: group.address,
+          usd: group.usd / 10 ** 18,
+          gold: group.ockedGold / 10 ** 18,
```
diminator: `ocekedGold` > `lockedGold`

packages/web/src/dev/ValidatorsListApp.tsx
```diff
@@ -0,0 +1,87 @@
...
+        usd
+      }
+      affiliates(first: 10) {
+        edges {
+          node {
+            address
+            elected
+            groupAddressHash
+            member
+            online
```
diminator: how is `online` measured? @mrsmkl

packages/web/src/dev/ValidatorsListApp.tsx
```diff
@@ -0,0 +1,87 @@
...
+import getConfig from 'next/config'
+import * as React from 'react'
+import { ApolloProvider, Query } from 'react-apollo'
+import ShowApolloError from 'src/dev/ShowApolloError'
+import ValidatorsList from 'src/dev/ValidatorsList'
+import { I18nProps, withNamespaces } from 'src/i18n'
+
+function createApolloClient() {
+  return new ApolloClient({
+    uri: getConfig().publicRuntimeConfig.LEADERBOARD.uri,
```
diminator: we can replace `LEADERBOARD` to `BLOCKSCOUT` (also for the leaderboard app)

diminator (changes requested):

diminator:
Looks very good already, some minor edits.
What was interesting is that the number of elected validators only accounts for 51 of them... where are the other ones?

diminator:
Maybe we can add links to the celo account info pages on blockscout (on the validator names)?

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,279 @@
...
+      </View>
+    )
+  }
+}
+
+export default withNamespaces('dev')(ValidatorsListApp)
+
+const styles = StyleSheet.create({
+  container: {
+    // marginTop: HEADER_HEIGHT,
```
aaronmgdr: remove?

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,279 @@
...
+              ))}
+            </tbody>
+          </table>
+        </View>
+      </View>
+    )
+  }
+}
+
+export default withNamespaces('dev')(ValidatorsListApp)
```
aaronmgdr, line 396: preferable to use the `Namespaces.dev` constant

packages/web/src/dev/ValidatorsList.scss
```diff
@@ -0,0 +1,233 @@
+// TODO: Move to files
+// START // Colors
+$palette: (
```
aaronmgdr: i'm not a fan of duplicating these color variables here.

packages/web/src/dev/ValidatorsList.scss
```diff
@@ -0,0 +1,233 @@
...
+        display: inline-block;
+        margin-left: 16px;
+        margin-right: 24px;
+        width: 20px;
+        text-align: center;
+      }
+    }
+
+    &--title-padding {
+      padding-left: 16px + 20px + 24px !important;
```
aaronmgdr: can you do this without important

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,279 @@
...
+import { HEADER_HEIGHT } from 'src/shared/Styles'
+import { colors, standardStyles, textStyles } from 'src/styles'
+
+interface ValidatorsListProps {
+  data: any
+  isLoading: boolean
+}
+
+export interface State {
+  expanded?: number
```
aaronmgdr: typically i think of `?` as meaning the key could be missing. where as `expanded: number | undefined` says the key is there but value could be undefined

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,279 @@
...
+import { Clipboard, StyleSheet, View } from 'react-native'
+import css from 'src/dev/ValidatorsList.scss'
+import { H1 } from 'src/fonts/Fonts'
+import { I18nProps, withNamespaces } from 'src/i18n'
+import Chevron, { Direction } from 'src/icons/chevron'
+import { HEADER_HEIGHT } from 'src/shared/Styles'
+import { colors, standardStyles, textStyles } from 'src/styles'
+
+interface ValidatorsListProps {
+  data: any
```
aaronmgdr, line 48: can this be more precise?

packages/web/src/shared/menu-items.ts
```diff
@@ -15,6 +15,10 @@ export const menuItems = {
     name: 'Build',
     link: '/build',
   },
+  VALIDATORS: {
+    name: 'Validators List',
+    link: '/build/validators',
```
aaronmgdr, line 20: I think we may want to check with @tjlahey and @vslavich  about the page path.

aaronmgdr (changes requested):
We don't use .scss anywhere else on site (except one small legacy area) so i'm not keen to add in for this as it creates a maintenance issue with duplicated colors and font definitions. If it is going to be there, in my experience using more than 1 level of indentation leads to way to high style specificity.

aaronmgdr:
Also would love to see a snapshot test added for the page in `src/_page-tests`

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: Done, thanks!

packages/web/src/dev/ValidatorsListApp.tsx
Pedro-vk: Done, thanks!

packages/web/src/dev/ValidatorsList.scss
Pedro-vk: We could investigate a way to use the colors in SCSS and TypeScript, but in other PR.

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: Done, thanks!

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: Done, thanks!

packages/web/src/dev/ValidatorsListApp.tsx
mrsmkl: It returns the latest block that has the validator's signature in the parent seal.

packages/web/src/dev/ValidatorsList.scss
```diff
@@ -0,0 +1,233 @@
+// TODO: Move to files
```
cmcewen: Agree with @aaronmgdr - just for consistency's sake here, I would prefer not to introduce another method of using CSS. We use react-native-web's stylesheet everywhere else, and try to stick to flexbox (https://flexboxfroggy.com/) as much as possible.

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,276 @@
...
+export interface State {
+  expanded?: number
+}
+
+class ValidatorsListApp extends React.PureComponent<ValidatorsListProps & I18nProps, State> {
+  state = {
+    expanded: undefined,
+  }
+
+  cutAddress(address: string) {
```
cmcewen: I typically just keep helper functions that don't operate on state/props outside of the component

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+  expand(expanded: number) {
+    if (this.state.expanded === expanded) {
+      this.setState({ expanded: undefined })
+    } else {
+      this.setState({ expanded })
+    }
+  }
+
+  cleanData({ celoValidatorGroups, latestBlock }: any) {
```
cmcewen, line 94: also would prefer not using any here

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,276 @@
...
+    const totalVotes: BigNumber = celoValidatorGroups
+      .map(({ votes }) => new BigNumber(votes))
+      .reduce((acc: BigNumber, _) => acc.plus(_), new BigNumber(0))
+
+    return celoValidatorGroups
+      .map(({ account, affiliates, votes }) => {
+        const group = account
+        const gName = group.name
+        return {
+          name: (gName || '').length <= 20 ? gName : `${gName.substr(0, 17)}...`,
```
cmcewen: I believe you should be able to do this in CSS https://www.w3schools.com/cssref/css3_pr_text-overflow.asp

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,276 @@
...
+          address: group.address,
+          usd: group.usd / 10 ** 18,
```
cmcewen: would be nice as a helper - weiToDecimal or whatever this is

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,276 @@
...
+              elected,
+              online,
+              account: { name, usd, lockedGold },
+              score,
+            } = validator.node
+            return {
+              name: (name || '').length <= 20 ? name : `${name.substr(0, 17)}...`,
+              address,
+              usd: usd / 10 ** 18,
+              gold: lockedGold / 10 ** 18,
```
cmcewen: there's a little duplicate logic here from above

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,276 @@
...
+                    <td className={css['table__cell--center']}>
+                      {this.formatNumber(group.gold, 2)}
+                    </td>
+                    <td className={css['table__cell--center']}>
+                      {this.formatNumber(group.uptime, 1)}%
+                    </td>
+                  </tr>
+                  {i === expanded && (
+                    <tr>
+                      <td colSpan={8}>
```
cmcewen: should this be a separate, smaller component?

packages/web/src/dev/ValidatorsListApp.tsx
```diff
@@ -0,0 +1,87 @@
...
+    fetch: async (...args) => {
+      const response = await fetch(...args)
+      const { data } = await response.json()
+      return new (Response as any)(JSON.stringify({ data }))
+    },
+  })
+}
+
+const query = gql`
+  query ValidatorGroups {
```
cmcewen, line 26: We were using this thing before (https://github.com/dotansimha/graphql-code-generator) - not sure if it's been updated but could be useful here to get TS types

cmcewen, line 26: Not sure if it is working now but we were using this thing (https://github.com/dotansimha/graphql-code-generator) a while ago in the mobile app to get TS types

packages/web/env-config.js.template
```diff
@@ -8,7 +8,7 @@ const environments = {
       databaseURL: 'https://celo-faucet.firebaseio.com',
       projectId: 'celo-faucet',
     },
-    LEADERBOARD: {
+    BLOCKSCOUT: {
```
aaronmgdr, line 11: this is only the template, if the encyrypted file isnt changed this wont matter

packages/web/src/dev/ValidatorsList.scss
aaronmgdr: yes and why not keep with the norms of the project and use Stylesheet?  I'm definitely  not one to rally for consistency just for its own sake, and to me that a change in how styling is done/ new way of doing it should have a discussion around it as it could lead to an increasingly large bundle size if down the road more ways of styling are added and could create confusion for future developers. 🤷🏻‍♂️

packages/web/env-config.js.template
Pedro-vk: I don't know how or who can change the file. Will ask tomorrow

aaronmgdr: ok maybe we do the leaderboard to blockscout change in a separate PR ?

packages/web/src/dev/ValidatorsList.scss
Pedro-vk: The problem is that is not possible to align the content as designed. From my side, I'm ok changing it to the table component, but it doesn't look as expected.
@cmcewen do you think that is ok?

aaronmgdr: how does it look? 
what the alignment that doesn't seem to work?

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: Only the components that are block-like can apply `text-overflow` property.
There are some ways to fix that:
- Adding to the table the property `table-layout: fixed`, that used the same width for each column. It does the columns that have short content wide.
- Use `table-layout: fixed` and configure the width of some of the params.
- Adding a `div` inside the `td` with a `max-width`, probably better.

I prefer the last one for this case, but anyway I'll see after applying the rest of the changes.

packages/web/src/dev/ValidatorsList.scss
Pedro-vk: image
I mean that. The solution is hardcode the width of each column. On this case, I suppose that is not a problem, but the table won't be fluid anymore.

diminator (approved):

packages/web/src/dev/ValidatorsList.scss
aaronmgdr: ok.  @cmcewen whats your thought?

aaronmgdr: Since this is obviously a table, one option would be to use react-native-web's `createElement` method like we do for video tag on the home/about page. this way you get native table/ td/tr/ th elements which you can style with `Stylesheet.create`

packages/web/env-config.js.template
aaronmgdr: looks like connor handled it

aaronmgdr:
regarding the path... I just had a convo with @timmoreton @vslavich and @tjlahey, given that we will be changing website to have a `/developers` page and a `/validators` page it would make more sense for the url of this list to be something like  `/validators/current` or `/validators/list` or `/validators/explore`

packages/web/src/dev/CopyToClipboard.tsx
```diff
@@ -0,0 +1,83 @@
...
+  // }
+
+  render() {
+    const { copied, hover } = this.state
+    const copy = this.copy.bind(this)
+    return (
+      <View
+        style={[styles.container]}
+        onClick={copy}
+        /*onMouseOver={this.hover.bind(this, true)}*/
```
aaronmgdr: is this in progress?

packages/web/src/dev/CopyToClipboard.tsx
```diff
@@ -0,0 +1,83 @@
...
+    event.stopPropagation()
+    if (!this.state.copied) {
+      copyToClipboad(this.props.content)
+      this.setState({ copied: true })
+      setTimeout(() => this.setState({ copied: false }), 2000)
+    }
+  }
+
+  // Simulating css hover
+  // hover(hover: boolean) {
```
aaronmgdr: there is a Hoverable Component which may be useful

packages/web/src/dev/ProgressCutBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+        </View>
+      ))}
+    </View>
+  )
+})
+export default ProgressCutBar
+
+const styles = StyleSheet.create({
+  barContainer: {
+    // width: 60,
```
aaronmgdr: nit remove?

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,623 @@
...
+}
+
+const HeaderCell = React.memo(function HeaderCellFn({
+  style,
+  name,
+  order,
+  onClick,
+}: HeaderCellProps) {
+  return (
+    <View onClick={onClick} style={[styles.tableHeaderCell, ...((style || []) as any)]}>
```
aaronmgdr, line 33: can you change this to `onPress` for the prop name. (variable im not so concerned with).

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,623 @@
...
+})
+
+interface ValidatorsListProps {
+  data: any
+  isLoading: boolean
+}
+
+export interface State {
+  expanded: number | undefined
+  orderBy: string
```
aaronmgdr: if there are only a limited set of strings that it can be ordered by its nice to list them like 

`orderedBy: 'name' | 'votes'`

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,623 @@
...
+
+  expand(expanded: number) {
+    if (this.state.expanded === expanded) {
+      this.setState({ expanded: undefined })
+    } else {
+      this.setState({ expanded })
+    }
+  }
+
+  orderBy(orderBy: string) {
```
aaronmgdr: one small style nit.

there was a decision at one point for methods on react components classes to be defined like `expand = () => {}` this way they are always scoped to class.

packages/web/src/dev/ValidatorsList.tsx
aaronmgdr: its just more idiomatic for React native web. 

and i think RNW even doesn't guarantee it wont break onClick (but i might be wrong)

aaronmgdr (commented):
Thank you for the work getting to using the react native stylesheet for styling. 🤗

Pedro-vk:
@aaronmgdr @cmcewen can you take a look again?

packages/web/src/dev/CopyToClipboard.tsx
Pedro-vk: Will reuse with Hoverable component, thanks!

packages/web/src/dev/CopyToClipboard.tsx
Pedro-vk: Thanks, will use it

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: `onPress` shows a TypeScript error used on `View` and `onClick` is working fine. What could produce that?

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: Can you give me an example of that?

packages/web/src/dev/CopyToClipboard.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const { copied, hover } = this.state
+    const copy = this.copy.bind(this)
+    return (
+      <View style={[styles.container]} onClick={copy}>
+        {copied ? (
+          <Text style={[styles.copied]}>Copied</Text>
+        ) : (
+          <Text style={[styles.icon]}>
+            <Hoverable
+              onHoverIn={this.hover.bind(this, true)}
```
aaronmgdr: rather than binding here would you create a class arrow function

```
hover = (hover: boolean) => {
    this.setState({ hover })
  }
```

binding within render creates a new function i believe which messes with prop  check heuristics, therefore we try to use on the above style 👍🏻

packages/web/src/shared/menu-items.ts
```diff
@@ -15,6 +15,10 @@ export const menuItems = {
     name: 'Build',
     link: '/build',
   },
+  VALIDATORS: {
```
aaronmgdr: would you mind calling this one `VALIDATORS_LIST` there are plans for a top level validators page

packages/web/src/dev/CopyToClipboard.tsx
Pedro-vk: @aaronmgdr What can I do with the functions that needs some params?

packages/web/src/dev/ValidatorsListApp.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { ApolloProvider, Query } from 'react-apollo'
+import ShowApolloError from 'src/dev/ShowApolloError'
+import ValidatorsList from 'src/dev/ValidatorsList'
+import { I18nProps, withNamespaces } from 'src/i18n'
+
+function createApolloClient() {
+  return new ApolloClient({
+    uri: getConfig().publicRuntimeConfig.BLOCKSCOUT.uri,
+    cache: new InMemoryCache(),
+    // TODO: Remove this workaround when the backend service fixes not needed errors
```
aaronmgdr, line 16: just checking if this TODO still applies

Pedro-vk: I think so, the API is having some changes right now, and sometimes is needed. From my point of view, we should keep it there until we stabilize the API.

packages/web/src/dev/ValidatorsList.tsx
aaronmgdr: hmm

oh right I forgot `onPress` only applies to `Text` and `Touchable*`s  

nevermind

packages/web/src/dev/CopyToClipboard.tsx
```diff
@@ -0,0 +1,78 @@
...
+    if (!this.state.copied) {
+      copyToClipboad(this.props.content)
+      this.setState({ copied: true })
+      setTimeout(() => this.setState({ copied: false }), 2000)
+    }
+  }
+
+  render() {
+    const { copied, hover } = this.state
+    const copy = this.copy.bind(this)
```
aaronmgdr, line 38: same here with  arrow function replacing bind

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -0,0 +1,632 @@
...
+class ValidatorsListApp extends React.PureComponent<ValidatorsListProps & I18nProps, State> {
+  state = {
+    expanded: undefined,
+    orderBy: 'name' as orderByTypes,
+    orderAsc: true,
+  }
+  private orderAccessors = {
+    name: (_) => _.name.toLowerCase(),
+    total: (_) => _.numMembers * 1000 + _.elected,
+    votes: (_) => +_.votesAbsolute,
```
aaronmgdr, line 69: curious about the `+` here 

whats it do?

aaronmgdr (approved):
I notice a few comments by connor that havent been responded to

If you can change the methods on classes to be arrow functions to avoid the need to bind inline that would be awesome

packages/web/src/dev/CopyToClipboard.tsx
Pedro-vk: In this case, is obvious that it can be done easily. But there is a good way to do it if I need to send a param?
(I'll change it)

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: The sorting of this column required to sort by number of members and then by elected, so:

- 3 total & 1 elected = 3001
- 3 total & 2 elected = 3002
- 2 total & 2 elected = 2002

A easy way to sort it without 2 sorting functions or checking 2 conditions and the same time

packages/web/src/dev/CopyToClipboard.tsx
aaronmgdr: forgive me if im not following but 

if copy is defined on a Component like
```
class Component {

copy = (param) => {
 // stuff in funciton as normal
}
}

```

you can access `this.copy()` like normal

packages/web/src/dev/ValidatorsList.tsx
aaronmgdr: ah i meant the `+_.votesAbsolute`

aaronmgdr: ive seen the `+` prefix a few places in this PR but not really before that so im curious

tjlahey:
image
Please center the gold dot between the two numbers

tjlahey:
image
Vertically align the numbers under the Elected/Total label - I know this is different from the mock.

PR #2489, opened 2020-01-21
Require Consent for tracking globally

### LINKED ISSUE #2467: Show cookie banner to all vistors on website
Given California's new Privacy law, celo.org should now show the privacy notice it currently displays to EU residents to all visitors

aaronmgdr:
### Description

Whereas before only likely EU citizens would be asked if they consented to light analytics tracking, now all users will have chance to opt in or out, in response to California's new privacy law.  

### Other changes
removes `in-eu` package. 

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/2467


packages/web/src/analytics/analytics.ts
```diff
@@ -65,3 +56,7 @@ export default {
     return segment.track(key, properties, options)
   },
 }
+
+function noTrack() {
```
jmrossy, line 60: Nit: seems like this tiny function could be an inline lambda

jmrossy (approved):

PR #2501, opened 2020-01-22
Fix celotool testnet deploy

jfoutts-celo:
### Description

Fixes issues preventing celotool testnet deploy from successfully electing validators:
- Migration overrides unfreezing rewards and setting minElectableValidators=1.
- Use imagePullPolicy: Always in helm-charts/testnet to support the `GETH_NODE_DOCKER_IMAGE_TAG=master` in .env.
- Adds `--proxy.allowprivateip` to geth invocation in helm-charts/testnet to support private network deployments.

### Other changes

- Commented out `rm /root/.celo/pkey` in helm-charts/testnet so if the container is restarted on error, it doesn't obscure the first error.

### Tested

Tested that `celotooljs deploy initial testnet && celotooljs deploy initial contracts` brings up a testnet with elected validators.


packages/helm-charts/testnet/templates/validators.statefulset.yaml
```diff
@@ -136,22 +136,22 @@ spec:
...
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
         command: ["/bin/sh"]
         args:
         - "-c"
         - |-
           set -euo pipefail
-          # Delete the private key as it is no longer needed
-          rm /root/.celo/pkey
+          # Don't delete the private key as it is still needed
```
tkporter: What was the issue here? I don't see where it's used

tkporter (approved):

packages/helm-charts/testnet/templates/validators.statefulset.yaml
jfoutts-celo: Ah, you're right. It isn't used.  I was seeing Validator0 failing to start with `rm: can't remove '/root/.celo/pkey': No such file or directory`.  But now I see that was because of another error.  When it was automatically restarted, the same code ran twice, and it produced that error (obscuring the root cause). I updated so it still leaves pkey, but revised explanation in comment.

Issue #1691, opened 2019-11-13
Local developer testnet setup


timmoreton:
Allow a developer to run from local checkouts: 
* several instances of celo-blockchain
* celostats
* blockscout

based on e2e scripts.

PR #2512, opened 2020-01-23
Running local testnet from celotool

### LINKED ISSUE #2220: Governance e2e tests fail sometimes on epoch score and reward
### Expected Behavior

`end-to-end-geth-governance-test` goes through

### Current Behavior

`end-to-end-geth-governance-test` fails with an error (see below). In circle ci it does not fail always. Locally it always fails for me now.

https://circleci.com/gh/celo-org/celo-monorepo/94321?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
  9 passing (10m)
  2 failing

  1) governance tests
       when the validator set is changing
         should update the validator scores at the end of each epoch:
     AssertionError: expected 0 to almost equal 100000000000000000000000 +/- 5000000000000
      at Object.assertAlmostEqual (src/e2e-tests/utils.ts:44:16)
      at Context.<anonymous> (src/e2e-tests/governance_tests.ts:481:29)
      at Generator.next (<anonymous>)
      at fulfilled (src/e2e-tests/governance_tests.ts:5:58)
      at process._tickCallback (internal/process/next_tick.js:68:7)

  2) governance tests
       when the validator set is changing
         should distribute epoch rewards at the end of each epoch:
     AssertionError: expected 0 to almost equal 89165304570705000000 +/- 5000000000000
      at Object.assertAlmostEqual (src/e2e-tests/utils.ts:44:16)
      at Context.<anonymous> (src/e2e-tests/governance_tests.ts:557:29)
      at Generator.next (<anonymous>)
      at fulfilled (src/e2e-tests/governance_tests.ts:5:58)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

H34D:
### Description

- [x] spin up network with validator nodes

### Related issues

- Solves one aspect of #1807 #1691
- Fixes #2220


packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,180 @@
...
+      default: 1,
+    })
+    .option('with-proxy', {
+      type: 'boolean',
+      description: 'Start with proxy in front',
+      default: false,
+    })
+    .option('verbosity', {
+      type: 'number',
+      description: 'Verbosity level',
```
nategraf: To distinguish this from the `verbose` flag below.
```suggestion
      description: 'Geth verbosity level',
```

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,180 @@
...
+      default: 5,
+    })
+    .option('verbose', {
+      type: 'boolean',
+      description: 'Verbose',
```
nategraf: To distinguish it from verbosity above
```suggestion
      description: 'Command verbosity flag',
```

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +625,476 @@ export const transferERC20Token = async (
     }
   }
 }
+
```
nategraf, line 642: A comment on this method would be appreciated

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +625,476 @@ export const transferERC20Token = async (
...
+export const runGethNodes = async ({
+  gethConfig,
+  validators,
+  validatorPrivateKeys,
+  verbose,
+}: {
+  gethConfig: GethRunConfig
+  validators: any[]
+  validatorPrivateKeys: any
```
nategraf: These should not be `any`.

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,180 @@
...
+      default: false,
+    })
+    .option('purge', {
+      type: 'boolean',
+      description: 'purge',
```
nategraf: This needs a description

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,180 @@
...
+export const builder = (argv: yargs.Argv) => {
+  return addCeloGethMiddleware(argv)
+    .option('network-id', {
+      type: 'string',
+      description: 'network id',
+      default: '1101',
+    })
+    .option('sync-mode', {
+      choices: ['full', 'fast', 'light', 'ultralight'],
+      demandOption: true,
```
nategraf: This could reasonably default to "full", especially if the caller specifies the `mining` flag.

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,180 @@
...
+    .option('wsport', {
+      type: 'number',
+      description: 'WS-RPC server listening port',
+      default: 8546,
+    })
+    .option('amount', {
+      type: 'number',
+      description: 'Amount of nodes to run',
+      default: 1,
+    })
```
nategraf: I found this flag a bit odd at first glance. `amount` isn't exactly clear from the name alone and how it working with the other flags is a bit unintuitive (e.g. you also specify a port as a flag and the user would have to open the code to know that if `amount > 1` the subsequent ports are chosen to be increments of 2). As a name change `replicas` or `instances` may be more clear. Not sure how to clarify the other issue.

packages/celotool/src/lib/interfaces/geth-instance-config.ts
```diff
@@ -0,0 +1,25 @@
+import { GethRunConfig } from './geth-run-config'
+
+export interface GethInstanceConfig {
+  gethRunConfig: GethRunConfig
```
nategraf: Please flatten this structure with respect to `GethRunConfig`. Recursive data structure are a headache, and it make initialization messy.

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +625,476 @@ export const transferERC20Token = async (
...
+  verbose: boolean
+}) => {
+  const validatorsFilePath = `${gethConfig.runPath}/nodes.json`
```
nategraf: This will not behave as expected when `gethConfig.runPath` is `''`. Please use Node's `path.join` https://nodejs.org/api/path.html#path_path_join_paths

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +625,476 @@ export const transferERC20Token = async (
...
+  const validatorInstances = gethConfig.instances.filter((x: any) => x.validating)
+  const validatorEnodes =
+    validatorInstances.length > 0
+      ? validatorPrivateKeys.map((x: any, i: number) => {
+          return getEnodeAddress(privateKeyToPublicKey(x), '127.0.0.1', validatorInstances[i].port)
```
nategraf: Validator node key should not be related to the private key, so this is not (should not) be a correct way to establish static nodes. We try not to use a static nodes file in testnets, which is why this code exists https://github.com/celo-org/celo-monorepo/blob/1f8f5e1147096f733ec42659e0042e446fe106d2/packages/celotool/src/e2e-tests/utils.ts#L439-L460

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +625,476 @@ export const transferERC20Token = async (
...
+}: {
+  gethConfig: GethRunConfig
+  validators: any[]
+  validatorPrivateKeys: any
+  verbose: boolean
+}) => {
+  const validatorsFilePath = `${gethConfig.runPath}/nodes.json`
+  const validatorInstances = gethConfig.instances.filter((x: any) => x.validating)
+  const validatorEnodes =
+    validatorInstances.length > 0
```
nategraf: No need for a ternary statement here. `.map` on an empty list yields empty list.

nategraf (commented):
Here are some more comments. Sorry, but I haven't had time to do a complete review yet. I was hoping it would be alight pass, but there are some code health issues I am concerned about.
As a general note, we need to avoid using `any` except when absolutely necessary.

packages/celotool/src/cmds/geth/start.ts
H34D: done

packages/celotool/src/cmds/geth/start.ts
H34D: done

packages/celotool/src/lib/geth.ts
H34D: done

packages/celotool/src/lib/geth.ts
H34D: done

packages/celotool/src/lib/geth.ts
H34D: done

packages/celotool/src/cmds/geth/start.ts
H34D: done

packages/celotool/src/cmds/geth/start.ts
H34D: renamed to instances

packages/celotool/src/lib/interfaces/geth-instance-config.ts
H34D: done

packages/celotool/src/cmds/geth/start.ts
H34D: done

packages/celotool/src/lib/geth.ts
H34D: done

packages/celotool/src/lib/geth.ts
H34D: done

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,208 @@
...
+  const rpcport = argv.rpcport
+  const wsport = argv.wsport
+
+  const mining = argv.mining
+  const minerGasPrice = argv.minerGasPrice
+  const network = 'local'
+  const instances = argv.instances
+
+  const mnemonic =
+    'jazz ripple brown cloth door bridge pen danger deer thumb cable prepare negative library vast'
```
yerdua: Does it make sense to have this hard-coded, rather than using the value defined in `.env.mnemonic`/`.env.mnemonic.$TESTNET_NAME`?

yerdua: Does it make sense to hard-code this here rather than use the value from `.env.mnemonic`/`env.mnemonic.$TESTNET_NAME`?

yerdua (commented):
Other than my question about the mnemonic, this looks good to me. However, I'd prefer to leave approval to @nategraf since he has more context and already reviewed.

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  const purge = argv.purge
+  const withProxy = argv.withProxy
+
+  const ethstats = argv.ethstats
+
+  console.info(`sync mode is ${syncMode}`)
```
nategraf: Why log this to the console? It seems unnecessary.
```suggestion
```

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,208 @@
...
+    .option('mining', {
+      type: 'boolean',
+      description: 'Is mining enabled',
+      default: false,
+    })
+    .option('miner-gas-price', {
+      type: 'number',
+      description: 'Mining gas price',
+      default: 0,
+    })
```
nategraf: FYI: Gas pricing in Celo applies EIP-1559 and has a network enforced gas price minimum, and block proposers only get the gas price in excess of this minimum. As a result, this geth arg doesn't quite have the same effect as it might for an Ethereum node, and I can't really think of a scenario in which you'd actually want to set this as it wouldn't actually help the validator earn more rewards. (In fact it may be counter-productive to earning more rewards as not including transactions drives the gas price minimum up) https://docs.celo.org/celo-codebase/protocol/transactions/gas-pricing
We can leave this flag if you want. I just don't see a reason to ever use it.

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,208 @@
...
+
+export const command = 'start'
+export const describe = 'command for running geth'
+
+interface StartArgv extends GethArgv {
+  networkId: string
+  syncMode: string
+  mining: boolean
+  blockTime: number
+  nodekeyhex: string
```
nategraf: Dead flag
```suggestion
```

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -0,0 +1,208 @@
...
+
+export const builder = (argv: yargs.Argv) => {
+  return addCeloGethMiddleware(argv)
+    .option('network-id', {
+      type: 'string',
+      description: 'network id',
+      default: '1101',
+    })
+    .option('sync-mode', {
+      choices: ['full', 'fast', 'light', 'ultralight'],
```
nategraf: With Geth 1.9 `lightest` will be the new name for `ultralight` as `ultralight` is already a Geth 1.9 sync mode. Might as well add it here now.
```suggestion
      choices: ['full', 'fast', 'light', 'ultralight', 'lightest'],
```

packages/celotool/src/e2e-tests/blockchain_parameters_tests.ts
```diff
@@ -1,23 +1,40 @@
...
 describe('Blockchain parameters tests', function(this: any) {
   this.timeout(0)
 
   let kit: ContractKit
   let parameters: BlockchainParametersWrapper
 
-  const gethConfig: GethTestConfig = {
+  const gethConfig: GethRunConfig = {
     migrateTo: 19,
+    migrate: true,
```
nategraf: No need to specify both `migrate` and `migrateTo`. https://github.com/H34D/celo-monorepo/blob/a7e958900d46dc3396879cb91bf358959cd8f1cc/packages/celotool/src/e2e-tests/utils.ts#L277
```suggestion
```

packages/celotool/src/e2e-tests/blockchain_parameters_tests.ts
```diff
@@ -30,7 +47,11 @@ describe('Blockchain parameters tests', function(this: any) {
 
     // TODO(mcortesi): magic sleep. without it unlockAccount sometimes fails
     await sleep(2)
-    kit = newKit('http://localhost:8545')
+
+    console.log('rpcURL', rpcURL)
```
nategraf: Please avoid adding log statement for things that are already known, like constants.
```suggestion
```

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -355,11 +396,21 @@ describe('governance tests', () => {
       // Wait for an extra epoch transition to ensure everyone is connected to one another.
       await waitForEpochTransition(epoch)
 
+      const groupWeb3Url = 'ws://localhost:8555'
+
+      if (verbose) {
+        console.log('groupWeb3Url', groupWeb3Url)
+      }
+
```
nategraf: Please don't log constants.

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -369,10 +420,21 @@ describe('governance tests', () => {
       const membersToSwap = [validatorAccounts[0], validatorAccounts[1]]
       const memberSwapper = await newMemberSwapper(groupKit, membersToSwap)
 
+      const validatorRpc = 'http://localhost:8549'
+      if (verbose) {
+        console.log('validatorRpc', validatorRpc)
+      }
+
```
nategraf: Please don't log constants.

packages/celotool/src/e2e-tests/slashing_tests.ts
```diff
@@ -37,21 +40,58 @@ function headerFromBlock(web3: Web3, block: any) {
...
+        syncmode: 'full',
+        port: 30309,
+        rpcport: 8551,
+      },
     ],
   }
 
-  const gethConfig: GethTestConfig = {
+  const gethConfig: GethRunConfig = {
+    gethRepoPath: '../../../celo-blockchain',
```
nategraf: Do not include relative paths outside of the repository. I am a little frustrated that I've made this same comment three separate times now.

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -385,109 +125,155 @@ export async function migrateContracts(
...
+export async function startBootnode(
+  bootnodeBinaryPath: string,
+  mnemonic: string,
+  gethConfig: GethRunConfig,
+  verbose: boolean
+) {
+  const bootnodePrivateKey = getPrivateKeysFor(AccountType.BOOTNODE, mnemonic, 1)[0]
+  const bootnodeLog = joinPath(gethConfig.runPath, 'bootnode.log')
+  const bootnodeArgs = [
+    '--verbosity=5',
```
nategraf: Verbosity 5 eats disk space at a furious pace. We shouldn't make this the default.
```suggestion
    '--verbosity=4',
```

packages/celotool/src/cmds/geth/start.ts
nategraf: Will this break if not the same as https://github.com/H34D/celo-monorepo/blob/a7e958900d46dc3396879cb91bf358959cd8f1cc/packages/celotool/src/e2e-tests/utils.ts#L168-L169?
If so, they should share the same constant.

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -385,109 +125,155 @@ export async function migrateContracts(
...
-  const gethBinaryPath = `${gethRepoPath}/build/bin/geth`
+
+  gethConfig.gethRepoPath = argv.localgeth || '/tmp/geth'
+  const gethBinaryPath = `${gethConfig.gethRepoPath}/build/bin/geth`
+  const bootnodeBinaryPath = `${gethConfig.gethRepoPath}/build/bin/bootnode`
 
   const before = async () => {
-    if (!argv.localgeth) {
-      await checkoutGethRepo(branch, gethRepoPath)
+    if (!argv.localgeth && gethConfig.gethRepoPath) {
```
nategraf: `gethConfig.gethRepoPath` should always be truthy based on the code above.
```suggestion
    if (!argv.localgeth) {
```

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -385,109 +125,155 @@ export async function migrateContracts(
...
+      await checkoutGethRepo(branch, gethConfig.gethRepoPath)
+    }
+
+    if (gethConfig.gethRepoPath) {
```
nategraf: See comment above. No need for a conditional here since it will always be true.

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -385,109 +125,155 @@ export async function migrateContracts(
...
   const validatorPrivateKeys = getPrivateKeysFor(AccountType.VALIDATOR, mnemonic, numValidators)
   const attestationKeys = getPrivateKeysFor(AccountType.ATTESTATION, mnemonic, numValidators)
-  const validators = getValidators(mnemonic, numValidators)
+  const validators = getValidatorsInformation(mnemonic, numValidators)
+
+  const proxyInstances = gethConfig.instances.filter((x: any) => x.isProxy)
+  const numProxies = proxyInstances.length
+
+  const proxyPrivateKeys = getPrivateKeysFor(AccountType.PROXY, mnemonic, numProxies)
+  const proxyEnodes = proxyPrivateKeys.map((x: any, i: number) => [
```
nategraf: Can't this just be `string` instead of `any`?
```suggestion
  const proxyEnodes = proxyPrivateKeys.map((x: string, i: number) => [
```

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -385,109 +125,155 @@ export async function migrateContracts(
...
+    if (gethConfig.useBootnode) {
+      bootnodeEnode = await startBootnode(bootnodeBinaryPath, mnemonic, gethConfig, verbose)
+    }
+
     let validatorIndex = 0
+    let proxyIndex = 0
+
     for (const instance of gethConfig.instances) {
+      // Non proxied validators and proxies should connect to the bootnode
+      if ((instance.validating && !instance.isProxied) || instance.isProxy) {
```
nategraf: Just `!instance.isProxied` should work here, unless I am missing something, and is more direct.
```suggestion
      if (!instance.isProxied) {
```

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -385,109 +125,155 @@ export async function migrateContracts(
...
+        if (gethConfig.useBootnode) {
+          instance.bootnodeEnode = bootnodeEnode
+        }
+      } else if (instance.validating && instance.isProxied) {
```
nategraf: If you apply the above suggestion, then this is just `else`
```suggestion
      } else {
```

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -385,109 +125,155 @@ export async function migrateContracts(
...
+
+    // Start all the instances
+    for (const instance of gethConfig.instances) {
+      await initAndStartGeth(gethConfig, gethBinaryPath, instance, verbose)
+    }
+
+    await connectValidatorPeers(gethConfig.instances)
+
+    // Give validators time to connect to each other
+    await sleep(60, true)
```
nategraf: A full minute is probably too much and will likely become annoying with time. I'd either recommend just lowering this value (migration can afford to start a little early because the transactions will propagate when the nodes connect) or, for extra credit, finding some way to poll the nodes to see when they are connected and wait for that to be true.

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -497,35 +283,48 @@ export function getContext(gethConfig: GethTestConfig) {
...
-      await snapshotDatadir(instance)
+      await snapshotDatadir(gethConfig.runPath, instance, verbose)
     }
   }
 
   const restart = async () => {
     await killGeth()
+
+    if (gethConfig.useBootnode) {
+      await killBootnode()
```
nategraf, line 306: `killGeth` kills all running processes with `geth` in the name, so I don't think you should have to separately kill the bootnode. https://github.com/H34D/celo-monorepo/blob/a7e958900d46dc3396879cb91bf358959cd8f1cc/packages/celotool/src/e2e-tests/utils.ts#L64-L67

packages/celotool/src/lib/geth.ts
```diff
@@ -582,6 +593,20 @@ export const onLoadTestTxResult = async (
...
+ * @param token the token contract to use
+ * @param from sender to send the token from
+ * @param to receiver that gets the tokens
+ * @param amount the amount of tokens to be sent
+ * @param password the password of the account to use
+ * @param txParams additional transaction parameters
+ * @param onTransactionHash callback, fired when the transaction has is generated
+ * @param onReceipt callback, fired when the receipt is returned
+ * @param onError callback, fired in case of an error, containing the error
+ */
```
nategraf, line 609: Yay for comments! Thanks!  🙌

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +639,501 @@ export const transferERC20Token = async (
...
+    nodeInfo = await admin.getNodeInfo()
+  } catch {
+    console.error(`Unable to get node info from ${enodeRpcUrl}`)
+  }
+
+  return nodeInfo.enode
+}
+
+export async function addStaticPeers(datadir: string, peers: string[], verbose: boolean) {
+  const staticPeersPath = `${datadir}/static-nodes.json`
```
nategraf: Like I've mentioned before, it's better to use `path.join` by default to avoid edge cases.
```suggestion
  const staticPeersPath = path.join(datadir, 'static-nodes.json')
```

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +639,501 @@ export const transferERC20Token = async (
...
+    )
+  }
+
+  if (wsport) {
+    gethArgs.push(
+      '--wsorigins=*',
+      '--ws',
+      '--wsport',
+      wsport.toString(),
+      '--wsapi=eth,net,web3,debug,admin,personal'
```
nategraf: Might as well expose the same APIs over WS
```suggestion
      '--wsapi=eth,net,web3,debug,admin,personal,txpool,istanbul'
```

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +639,501 @@ export const transferERC20Token = async (
...
+  if (ethstats) {
+    gethArgs.push(`--ethstats=${instance.name}@${ethstats}`, '--etherbase=0')
+  }
+
+  const gethProcess = spawnWithLog(gethBinaryPath, gethArgs, `${datadir}/logs.txt`, verbose)
+  instance.pid = gethProcess.pid
+
+  gethProcess.on('error', (err) => {
+    throw new Error(`Geth crashed! Error: ${err}`)
+  })
```
nategraf, line 966: Nice addition!

packages/celotool/src/lib/geth.ts
```diff
@@ -614,3 +639,501 @@ export const transferERC20Token = async (
...
+      await Promise.all(
+        enodes.map(async (enode, j) => {
+          if (i === j) {
+            return
+          }
+          /*
+          console.log(
+            `connecting ${instances[i].name} with ${instances[j].name} using enode ${enode}`
+          )
+          */
```
nategraf: Please uncomment this as `verbose` logging or delete it. We just don't want to leave commented out code lying around.

nategraf (commented):
I think we should be good after this round. Thanks for addressing all my comments so far.

packages/celotool/src/lib/geth.ts
H34D: thanks

packages/celotool/src/lib/geth.ts
H34D: wrapped in if verbose

packages/celotool/src/e2e-tests/slashing_tests.ts
H34D: removed

packages/celotool/src/e2e-tests/governance_tests.ts
H34D: removed

packages/celotool/src/e2e-tests/governance_tests.ts
H34D: removed

packages/celotool/src/cmds/geth/start.ts
H34D: removed

packages/celotool/src/cmds/geth/start.ts
H34D: It will not, they can be different. in `start` it is just a preset. I will make it an option and use this as preset.

H34D: done

packages/celotool/src/e2e-tests/utils.ts
H34D: removed

packages/celotool/src/e2e-tests/utils.ts
H34D: bootnodes are started with the `bootnode` command not `geth` so killGeth will not kill bootnodes.

packages/celotool/src/e2e-tests/utils.ts
H34D: replaced with `waitForSync` methods

H34D:
all green, here we go

packages/celotool/src/cmds/geth/start.ts
H34D: there is no mnemonic in there for the network `local`

nategraf (approved):
Looks good to me! Thanks for writing all this up and addressing my comments

PR #2526, opened 2020-01-23
[Wallet] Prompt users with connectivity issues to switch to forno

### LINKED ISSUE #1559: Users SBAT be prompted to switch to forno when experiencing connectivity issues
### Expected Behavior

(Copied over from #1171, splitting into two issues- cc @nityas )

We propose a modal that gives users with connectivity issues the option to enable Forno mode.

NOTES:

If geth has restarted more than twice, or syncing has taken longer than 15s, but there is still some internet connection(eg type:unknown) show the modal below.
same modal template as the 'Are you sure?' Backup Key Modal, with enable being the primary cta:
a user should only see this modal after they have completed the 'Welcome to Celo' screen
once ‘Dismiss’ is pressed, we should not show the modal again, to avoid an overly intrusive or confusing experience.
(TODO: are there other signals we can use to fine tune this?)

_____________
**Switch Connection Mode?**

We’ve noticed you’re having some trouble connecting. We recommend enabling **Celo-Lite** mode to allow you to keep using the Celo Wallet with intermittent connection.

Celo-Lite allows you to communicate with the Celo Network through a trusted, hosted node, which can give faster performance than running a node on your device. You can always change this mode in Settings.

[Dismiss]  [Enable Celo-Lite]

### Current Behavior

Working as toggle from settings screen

annakaz:
### Description

When geth is taking too long to sync (>15 seconds for web3 to sync or seeing connectivity issues with geth), prompt them to turn on forno. Do not prompt if users have already declined to avoid an instrusive experience, and wait to prompt until after they have gone through the welcome screen

### Tested

On a pixel 2 by changing sync to take too long and confirming that users are prompted to switch to forno

### Other changes

Cancel geth sagas when switching to forno rather than checking within sagas whether forno is enabled

### Related issues

- Fixes #1559 

### Backwards compatibility

Yes


packages/mobile/locales/es-419/accountScreen10.json
```diff
@@ -21,7 +21,11 @@
     "body": "En caso de ingresar un PIN incorrecto, la aplicación se reiniciará.",
     "understand": "Entendido"
   },
-
+  "promptZeroSyncModal": {
+    "header": "Cambiar modo de conexión?",
+    "body": "Hemos notado que tienes problemas para conectarte. Recomendamos habilitar el modo de ahorro de datos para que pueda seguir usando Celo Wallet con una conexión intermitente.",
+    "switchToDataSaver": "Habilitar Data Saver"
```
jmrossy: Once https://github.com/celo-org/celo-monorepo/pull/2539 merges please use Ahorro de Datos here

packages/mobile/src/geth/actions.ts
```diff
@@ -15,6 +17,20 @@ export const setInitState = (state: InitializationState): SetInitState => ({
   state,
 })
 
+export const cancelGethSaga = () => ({
+  type: Actions.CANCEL_GETH_SAGA,
+})
+
+interface SetPromptZeroSync {
```
jmrossy: Nit, I think the name convention we use elsewhere is `SetPromptZeroSyncAction`, though I guess not in this file for some reason

packages/mobile/src/geth/saga.ts
```diff
@@ -120,7 +128,13 @@ export function* initGethSaga() {
     Logger.error(TAG, 'Geth initialization failed, restarting the app.')
     deleteChainDataAndRestartApp()
   } else {
-    navigateToError('networkConnectionFailed')
+    // Suggest switch to forno for network-related errors
+    if (yield select(promptZeroSyncIfNeededSelector)) {
+      yield put(setPromptZeroSync(false))
+      navigate(Screens.DataSaver, { promptModalVisible: true })
```
jmrossy, line 134: I assumed we would just enable forno silently in this case. @nityas can you confirm that navigating the user to the Data Saver screen is the desired behavior?

packages/mobile/src/invite/JoinCelo.tsx
```diff
@@ -136,6 +139,7 @@ export class JoinCelo extends React.Component<Props, State> {
       return
     }
 
+    this.props.setPromptZeroSync(true) // Allow zero sync prompt after Welcome screen
```
jmrossy, line 142: Why set it manually instead of making true the default value?

packages/mobile/src/geth/selectors.ts
```diff
@@ -0,0 +1,7 @@
+import { InitializationState } from 'src/geth/reducer'
+import { RootState } from 'src/redux/reducers'
+
+export const isGethConnectedSelector = (state: RootState) =>
```
jmrossy, line 4: Dupe of the existing selector in geth/reduer.ts

packages/mobile/src/invite/JoinCelo.tsx
annakaz: We don't want to prompt the switch until after the user has gone through the Welcome screen. By having it default false we prevent the pop up from showing up at the beginning of the new user UX

packages/mobile/src/geth/saga.ts
annakaz: The user gets navigated to the DataSaver screen with a modal, then once they confirm/decline the switch they go back to wherever they were

packages/mobile/src/geth/selectors.ts
annakaz: Removed from reducer to be consistent with keeping selectors in /selectors

packages/mobile/locales/es-419/accountScreen10.json
annakaz: Updated translation

packages/mobile/src/geth/saga.ts
annakaz: Confirmed with @nityas that this prompt is the expected behavior!

packages/mobile/src/geth/actions.ts
annakaz: Updated all actions in this file to use the normal naming convention

packages/mobile/src/geth/saga.ts
```diff
@@ -205,6 +213,8 @@ function* monitorAppState() {
 
 export function* gethSaga() {
   yield call(initGethSaga)
-  yield fork(monitorAppState)
-  yield fork(monitorGeth)
+  const gethRelatedSagas = yield all([fork(monitorAppState), fork(monitorGeth)])
+  yield take(Actions.CANCEL_GETH_SAGA)
+  yield cancel(gethRelatedSagas)
```
jmrossy, line 218: Can you confirm this cancels the individual forks and not just the synthesized one created by `all()`? If monitorGeth's logs no longer show up after cancel that would confirm it :)

packages/mobile/test/schemas.ts
```diff
@@ -63,6 +63,7 @@ export const vNeg1Schema = {
   geth: {
     initialized: 'INITIALIZED',
     connected: true,
+    promptZeroSyncIfNeeded: false,
```
jmrossy: state schema changes, including new additions, need to go in the latest schema (v5 below) since it didn't exist in v-1

packages/mobile/src/geth/saga.ts
annakaz: Yes confirmed that monitorGeth logs go away! Agree that the redux saga cancel all documentation is unclear though (https://redux-saga.js.org/docs/advanced/TaskCancellation.html / outdated post https://github.com/redux-saga/redux-saga/issues/988#issuecomment-301064683)

jmrossy (approved):

annakaz:
Note that build is failing on master due to attestation service (unrelated to this PR)

packages/mobile/src/geth/saga.ts
nityas: 1. confirmed-- it adds a layer of centralization and we do not want to do this without user consent

PR #2527, opened 2020-01-24
 Micro Improvement to web tests

aaronmgdr:
### Description

adds test for leaderboard and and abortable fetch 
removes old missiontext component
update coverage targets 

### Other changes

I changed the target on mobile as 10% seemed way to low.


.codecov.yml
```diff
@@ -5,13 +5,13 @@ coverage:
...
-        threshold: 40%
-        target: 10%
+        threshold: 5%
+        target: 70%
       web:
         paths: "packages/web/"
         flags: web
-        threshold: 5%
-        target: 70%
+        threshold: 2%
```
aaronmgdr, line 13: from my understanding threshold is how much it can decrease from master before a flag is raised, and target is the absolute amount of lines that need to be tested overall

jeanregisser:
> I changed the target on mobile as 10% seemed way to low.

Yes 😅 👍

jmrossy (approved):

PR #2536, opened 2020-01-24
Two CLI commands for checking the state of the network

### LINKED ISSUE #2531: Add a command that validators can use to see how much votes they need to get elected
### Expected Behavior

Command that outputs the number of votes and shows validators in order.

### Current Behavior

They can use `election:run`, but this doesn't show the number of votes, and it just shows the groups in order, not validators.

mrsmkl:
### Description

Adds two commands that I've been using for monitoring the state of the network, `election:compare` which shows how many votes a validator needs to get elected, and `network:downtime` to show information about downtime.

### Tested

Only manual testing in baklava.

### Related issues

- Fixes #2531


barbaraliau:
@medhak1 is this something that would be useful to add to the CLI?

packages/cli/src/commands/election/compare.ts
```diff
@@ -0,0 +1,63 @@
...
+  votes: {},
+  score: {},
+  name: {},
+  address: {},
+  groupName: {},
+  affiliation: {},
+}
+
+export default class ElectionCompare extends BaseCommand {
+  static description =
```
nambrot, line 16: Does this need to be updated or is it the same as `election:run`?

packages/cli/src/commands/network/downtime.ts
```diff
@@ -0,0 +1,119 @@
...
+  while (str.length < 100) str = '0' + str
+  let res = ''
+  for (let i = 0; i < 100; i++) {
+    res += str.charAt(i) === '1' ? '.' : 'X'
+  }
+  return res
+}
+
+export default class Analyze extends BaseCommand {
+  static description = 'Analyze downtime'
```
nambrot, line 47: We probably want to have a more verbose description for this?

AlexBHarley:
Closing due to age and no willing engineer to shepherd this over the line. 

Slack conversation [here](https://celo-org.slack.com/archives/CP5V8KF51/p1624447187357800).

PR #2537, opened 2020-01-24
vesting smart contracts and related ts configs only

### LINKED ISSUE #800: Genesis accounts SBAT lock Gold and receive funds only on a vesting schedule
### Expected Behavior

We want to be able to include in the genesis block a contract (call it `VestingFactory`)

While vesting, funds should be able to be locked using the `LockedGold` contract so that they can be used for voting for validators, in on-chain governance, and earn compound rewards. To do this, `VestingFactory` might have a function `createVesting()` that creates a new contract, owned by a specified `beneficiary`, that exposes wrappers for the public functions of `LockedGold` and `Accounts`.

The beneficiary will be able to call `LockedGold.lock` to lock up gold, and `Accounts.authorize*()` to specify keys that can be used for other governance actions (e.g. voting, running a validator).  At any time, the `beneficiary` may `unlock` locked gold to return it to the vesting contract. Gold in the vesting contract may be transferred to the `beneficiary` pursuant to the vesting contract's `Schedule`.

Note that we may not need to support wrappers around `Validators`, `Governance`, and `Election`, since the beneficiary will be able to authorize other keys that can perform those actions.

The `Schedule` will be composed of the following parameters:
* parameters `vestingCliff `, `vestAmountPerPeriod`, `vestingStartTime`, `vestPeriodSec`, `totalAmount`, as per below

The user can withdraw `VestingInstance.balance * (vestAmountPerPeriod * ((now - vestingStartTime)/vestPeriodSec)) / totalAmount`, provided `now > vestingCliff + vestingStartTime`. This allows users to withdraw their rewards pro rata according to the schedule, and supports continuous (e.g. monthly) vesting.

The data in `VestingFactory` will need to comprise a map: `address -> Schedule` such that for a given `address` we have:
* the schedule has a boolean `revokable`

As we go we'll also need to track:
* `withdrawnAmount`, initially zero.
* `vestingRevokedFrom`, initially zero but can be set to revocation time (see below)
 
Once the original vesting schedule would be complete, i.e `vestAmountPerPeriod*(now-vestingStartTime)/vestPeriodSec>=totalAmount`, all funds may be withdrawn.

**Further requirement for revocation:** the Celo Foundation may need to revoke vesting schedule instances. Each `Schedule` should be either `revocable` or not.
 
VestingFactory should also take two parameters: 
* `revoker` the only address that can call `revoke(address, vestingEndTime)`
* `refundDestination` an address that then receives refunds  

To revoke a schedule that hasn't started, the `VestingFactory` should support a  `revoke(vestingEndTime)` function. Can only be called once. This should have no effect if address has a `Schedule` that is `revokable=false`. Otherwise, if they have not called 'vesting:start', then set `vestingRevokedFrom ` and calculate the portion to return to transfer to `refundDestination`. All rewards  are retained by the user.  The vesting instance contracts should support a similar function.  

We want support for this contract in ContractKit, and in the CLI, as a new set of methods,  `vesting:authorize`, `vesting:lock`, `vesting:unlock`, etc.

**Requirement for pause:**

If a grant is `revokable`, then the `revoker` can also pause withdrawals for a certain period, up to 365 days. This does not affect the underlying timing of vesting, it only affects whether withdrawals can be made.  We would need something like a `pause(seconds)` method only callable by `revoker`, then on `withdraw` check that a pause is not currently in place.

### Current Behavior

No support for this.

BatiGencho:
### Description

Implementation of Vesting Contract & Vesting Contract Factory - smart contracts and related infrastructure only!

### Tested

Unit tests for factory and vesting instance are included in the commit.

### Related issues

- Fixes #800


asaj:
Please address merge conflicts

Issue #2511, opened 2020-01-23
Revoke for other groups always left 1


wjdfx:
### Expected Behavior

User can’t revoke all token from a group. It’s always left 1 token on the election:show xxx —voter list. Debug log see attach.

[message.txt](https://github.com/celo-org/celo-monorepo/files/4103588/message.txt)

### Current Behavior

celocli election:show 116a647ce0e9a012099508f285a7735f86e333a4 --voter
Running Checks:
   ✔  116a647ce0e9a012099508f285a7735f86e333a4 is a registered Account
address: 116a647ce0e9a012099508f285a7735f86e333a4
votes:
  0:
    group: 0x93c110Dc9812dc075C1541b2c5D1030B2Ad6C293
    pending: 0
    active: 1
  1:
    group: 0xCd061dF755F558A5ea809EA1589b0f8a5a5Bfa27
    pending: 0
    active: 1
  2:
    group: 0x2947EA31ab65B1FEE886DC6c113F79B222c804B7
    pending: 0
    active: 1

celocli election:revoke --for 0x2947EA31ab65B1FEE886DC6c113F79B222c804B7 --from 0x116a647Ce0E9A012099508f285a7735F86e333A4 --value 1
Running Checks:
   ✔  0x116a647Ce0E9A012099508f285a7735F86e333A4 is Signer or Account
   ✔  0x2947EA31ab65B1FEE886DC6c113F79B222c804B7 is ValidatorGroup
SendTransaction: revoke
txHash: 0xbe179973023c57efc847d7a957f7bbdd818e8eb9a41745cedac90e2e290b4bab
Sending Transaction: revoke... done

celocli election:show 116a647ce0e9a012099508f285a7735f86e333a4 --voter
Running Checks:
   ✔  116a647ce0e9a012099508f285a7735f86e333a4 is a registered Account
address: 116a647ce0e9a012099508f285a7735f86e333a4
votes:
  0:
    group: 0x93c110Dc9812dc075C1541b2c5D1030B2Ad6C293
    pending: 0
    active: 1
  1:
    group: 0xCd061dF755F558A5ea809EA1589b0f8a5a5Bfa27
    pending: 0
    active: 1
  2:
    group: 0x2947EA31ab65B1FEE886DC6c113F79B222c804B7
    pending: 0
    active: 1

mrsmkl:
Duplicate of #2507

PR #2540, opened 2020-01-24
Fix rounding error in Election.sol

### LINKED ISSUE #2507: Rounding error prevents revoking active votes for group
They have hit the max # of validator groups to vote for. They seem to have voted for each of the validator groups with 1 wei.

When running transactions to revoke, eg celocli election:revoke --for 0x2947EA31ab65B1FEE886DC6c113F79B222c804B7 --from 0x116a647Ce0E9A012099508f285a7735F86e333A4 --value 1, the transaction succeeds. 

However, celocli election:show 116a647ce0e9a012099508f285a7735f86e333a4 --voter still shows them voting for the validator group.

Their locked gold balance (both total & non-voting) celocli lockedgold:show 0x116a647Ce0E9A012099508f285a7735F86e333A4 increases by 1 wei. Current theory is it's a rounding error

asaj:
### Description

This PR fixes a rounding error that prevents users from potentially revoking all of their active votes. In some cases, the user may try to revoke all of their active votes, but due to rounding errors `getActiveVotesUnitsDelta` returns one fewer than their total active vote units. To avoid this, a special case is added for when the value being revoked is exactly equal to the number of active votes for the user.

Note that this does not address *all* forms of rounding errors with respect to active votes. There still may be cases where, for example, the total number of active votes does not equal the sum of active votes over all (voter, group) pairs.
  
### Tested

Reproduced the issue in unit tests by distributing rewards after making active votes. Confirmed that this fix solved it.

### Other changes

None

### Related issues

- Fixes #2507, #2511 

### Backwards compatibility

Yes


packages/protocol/test/governance/election.ts
```diff
@@ -713,55 +713,70 @@ contract('Election', (accounts: string[]) => {
     })
   })
 
-  describe('#revokeActive', () => {
-    const voter = accounts[0]
-    const group = accounts[1]
-    const value = 1000
+  describe.only('#revokeActive', () => {
```
nategraf: Revert `only`
```suggestion
  describe('#revokeActive', () => {
```

nategraf (approved):
One important change. Otherwise, looks good! 👍

Issue #1659, opened 2019-11-11
Devs SBAT rely on mobile End to end test 


martinvol:
### Expected Behavior

Test should fail if they don't meet what it's expected.

### Current Behavior

Currently failing due https://github.com/wix/Detox/issues/1723 and set to green manually to avoid having the whole repo red.

PR #2542, opened 2020-01-24
[Wallet] E2E test improvements

jmrossy:
### Description

- Simplify E2E related scripts into one single run_e2e.sh
- run_e2e.sh handles properly building and configuring the app
- Update docs accordingly
- Fix bundling issue with android VMs
- Update detox
- Update RN fast crypto and RN bip 39 with @cmcewen fixes for crashes

### Tested

Ran e2e tests

### Related issues

- #1659

### Backwards compatibility

Yes


packages/mobile/.env
```diff
@@ -1,5 +1,5 @@
 ENVIRONMENT=local
-DEFAULT_TESTNET=alfajoresstaging
+DEFAULT_TESTNET=integration
```
jeanregisser, line 2: Is it intentional to switch the testnet in this PR?

packages/mobile/README.md
```diff
@@ -130,21 +123,9 @@ See [`src/identity/verification.test.ts`] for an example.
...
-Please set `123456` as the pin code in the emulator, since the e2e tests rely on
-that.
-
-Next, the VM snapshot settings should be modified:
-
-1.  Close all apps and lock the emulator (go to lock screen).
-2.  Power off the emulator
-3.  Power it back on and go to emulator settings (... button) -> Snapshots -> Settings
-4.  Set Auto-Save to No
+must have the proper emulator set up. Follow the instrutions in `e2e/README.md`.
```
jeanregisser: We should make it a link to the e2e readme.

packages/mobile/e2e/init.js
```diff
@@ -2,14 +2,23 @@ const detox = require('detox')
...
+// This is strictly optional.
+jasmine.getEnv().addReporter(assignReporter)
+
+// Increase default jest timeout
+jest.setTimeout(120000)
+
 beforeAll(async () => {
   await detox.init(config)
-})
+}, 300000)
```
jeanregisser, line 21: Nice addition! We should add a comment as to why this is needed.
From what I saw, jest swallows errors or timeouts here and happily continues with executing the test suite though this crucial init failed.

packages/mobile/scripts/run_e2e.sh
```diff
@@ -1,78 +1,105 @@
...
-
-echo "Closing emulator"
-kill -s 9 `ps -a | grep "Nexus_5X_API_28_x86" | grep -v "grep"  | awk '{print $1}'`
+echo "Done test, cleaning up"
+yarn react-native-kill-packager
 
-echo "Closing pidcat"
-kill -s 9 `ps -a | grep "pidcat" | grep -v "grep"  | awk '{print $1}'`
+echo "Closing emulator (if active)"
+adb devices | grep emulator | cut -f1 | while read line; do adb -s $line emu kill; done
```
jeanregisser: Should we move this to the android only block?

jeanregisser (commented):
Great improvements!! :+1:
Couple of small questions before merging.

packages/mobile/scripts/run_e2e.sh
```diff
@@ -1,78 +1,105 @@
...
-   yarn test:detox
-   STATUS=$?
-fi
+yarn react-native-kill-packager || echo "Failed to kill package manager, proceeding anyway"
+
+# Build the app and run it
+if [ $PLATFORM = "android" ]; then
+  echo "Using platform android"
+
+  if [ -z $ANDROID_SDK_ROOT ]; then
```
jeanregisser: I think it should be `$ANDROID_HOME` as recommended in our [eng SETUP](https://github.com/celo-org/celo-monorepo/blob/cd7be93a3abd1333d7972765f23cfa70d17cc492/SETUP.md#install-android-dev-tools) and [React Native setup guide](https://github.com/facebook/react-native-website/blob/067141dca4e4568074bcafb65213aaa95a4eef67/docs/getting-started.md#3-configure-the-android_home-environment-variable)

packages/mobile/scripts/run_app.sh
```diff
@@ -1,26 +1,29 @@
...
 
 # Flags:
-# -n: Name of the network to run on
 # -p: Platform (android or ios)
-# -f: Fast (skip steps not required unless network or depedencies changes)
-# -r: Hot Reload (Restore nav state on reload)
+# -n (Optional): Name of the network to run on 
+# -f (Optional): Fast (skip steps not required unless network or depedencies changes)
+# -r (Optional): Hot Reload (Restore nav state on reload, useful for UI dev-ing)
+# -b (Optional): Just configure and build, skip running
```
jeanregisser: nit: to avoid the confusion, how about we use this: 'Just configure and build **sdk**, skip running' ?
given it doesn't build the app itself.

packages/mobile/.env
jmrossy: Yes

packages/mobile/e2e/init.js
jmrossy: I'm not 100% sure, I'm just aligning this file with the example on Detox's repo

packages/mobile/scripts/run_e2e.sh
jmrossy: Good idea! Thought it's idempotent so nbd either way

packages/mobile/scripts/run_e2e.sh
jmrossy: Good catch! Looks like I set both in my bash profile, not sure why

packages/mobile/scripts/run_app.sh
jmrossy: Done

jeanregisser (approved):

PR #2544, opened 2020-01-24
Add tests for wrong index being passed in Governance

### LINKED ISSUE #298: Developers SBAT know that when indexes are used in Governance.sol they're always less than the array length
### TODO

Everywhere we use an index, require it's less than the array length.

Reference:
[Governance.sol](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/Governance.sol)

asaj:
### Description

Indices can be passed to external functions for dequeued proposals, this PR adds tests for out of range indices being passed.

### Tested

Added unit tests

### Other changes

None

### Related issues

- Fixes #298 
### Backwards compatibility

Yes


packages/protocol/contracts/governance/Governance.sol
```diff
@@ -1056,6 +1056,7 @@ contract Governance is
     uint256 proposalId,
     uint256 index
   ) private view returns (bool) {
+    require(index < dequeued.length);
```
lucasege: Could you add an error message for `require`

packages/protocol/test/governance/governance.ts
```diff
@@ -1382,6 +1382,10 @@ contract('Governance', (accounts: string[]) => {
       })
     })
 
+    it.only('should revert when the index is out of bounds', async () => {
+      await assertRevert(governance.approve(proposalId, index + 1))
+    })
+
```
lucasege: Remove `.only` for all 3 tests

lucasege:
Also saw you didn't add a `require` for `deleteDequeuedProposal`, I assume because it will always be preceded by a call to `isDequeuedProposal` with the same index. If that was your reasoning, maybe adding
"* @dev `deleteDequeuedProposal` must always be preceded by `isDequeuedProposal`" to `deleteDequeuedProposal` would make sense. 

Also, the same index is used within `vote` into `referendumVotes` without a bounds check, not sure if `referendumVotes` and `dequeued`'s lengths are linked or if we need another test

packages/protocol/test/governance/governance.ts
asaj: 🤦‍♂

asaj:
> Also, the same index is used within vote into referendumVotes without a bounds check, not sure if referendumVotes and dequeued's lengths are linked or if we need another test

This is okay because referendumVotes is a mapping and not an array

lucasege (approved):
Looks good, thanks for making changes. Pending tests passing or being flakes, looks good to merge

PR #2547, opened 2020-01-26
Opsec audit script

timmoreton:
### Description

A script for testing aspects of a deployment's security. 
  
### Tested

On centos, debian.

### Backwards compatibility

Yes

### DIFF
--- a/packages/opsec-audit/opsec-audit.sh
+++ b/packages/opsec-audit/opsec-audit.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+tput clear
+trap ctrl_c INT
+
+function ctrl_c() {
+        echo "**Exiting as encountered Ctrl+C"
+	exit 0;
+}
+#
+echo "###############################################"
+echo "Welcome to the Celo Security audit of your linux machine:"
+echo "###############################################"
+echo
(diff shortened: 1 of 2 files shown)


packages/opsec-audit/opsec-audit.sh
```diff
@@ -0,0 +1,124 @@
...
+echo
+sleep 3
+echo "###############################################"
+echo "You Are $HOSTNAME..:"
+echo
+sleep 3
+echo "Starting"
+START=$(date +%s)
+echo -e "\e[0;33m 1. Downloading Lynis////// \e[0m"
+curl https://downloads.cisofy.com/lynis/lynis-2.7.5.tar.gz -o lynis-2.7.5.tar.gz
```
tkporter, line 25: What if this a proxied validator that's entirely disconnected from the outside world?

timmoreton: Great question. Think we are going to change this so they download the script ahead of time and we just check its md5

jcortejoso: Maybe packing the script as a container would be a solution, so we do not need to download anything

barbaraliau:
Is this still relevant or should this be closed?

Issue #1339, opened 2019-10-15
[Android] Push notification is not received for user B when user A proceed for payment 


Lss-Ankit:
**App version:** Zero - sync  build v1.5.0, 
**Does not repro: ** Play store build v1.5.0
**Repro on:**  Samsung Galaxy S6 (7.0), Samsung Galaxy S7 (8.0.0)
**Testing Account:** 680 208 0265/[Backup Kay](https://docs.google.com/spreadsheets/d/1Iy41GGA6e2pES2XHjBn7nb8X3n8zZsu7EW1jvwD2e70/edit?usp=sharing)
 
**Repro Steps:** 
1) Launch the app and sign up as user A. 
2) Launch the app and sign up as user B. 
3) Put the app in the background for user B. 
4) User A sends a dollar to User B. 
5) Observe for user B Push notification is not received. 

**Investigation:** Payment request push notification is received for User B when user A Proceed for Payment request. 
**Impact:** User did not get the idea about received payment if an app is a background. 
**Current Behavior:** Push notification is not received
**Expected Behavior:** Push notification should be received
**Attachment:**[ Android_PushNotificationIssue.MOV](https://drive.google.com/file/d/1jOSKVL_J_R__yTFqpLibacR-nAW9Up86/view?usp=sharing) (Bug observe after 44 sec)

Lss-Ankit:
Also, observe that "Payment received" in-app notification is not received for user B when user A proceed for payment.
**Pre-condition:** App should foreground.  
**Note:** "Payment request"  in-app notification is received for user B when user A sends a payment request. 
[Attachment: ](https://drive.google.com/file/d/1WGCGlA7Iw5AjKeK-Hzk9MR4A_hTE4_oe/view?usp=sharing)

Lss-Ankit:
@jmrossy and @cmcewen Verified with Android build v1.5.1 (104694405) (Internal Early Access) and observe that the issue is still reproduced. 

Issue: 
1.  Push notification is not received for user B when user A proceed for payment
2.  "Payment received" in-app notification is not received for user B when user A proceed for payment.

[Attachment](https://drive.google.com/file/d/1X8X06SJeDA9-xGjh7ux5vkSXRJf2HIwa/view?usp=sharing)

jmrossy:
Ivan and I just tested a payment and it works now. I believe it was just a misconfigured firebase db and/or the address normalization issue Ivan fixed.

Lss-Ankit:
[...]

PR #2548, opened 2020-01-27
[Wallet] Update balance on new transaction

### LINKED ISSUE #1335: [Android] ‘cUSD’ balance is not updated automatically when user receives an amount from other users.
**App version:** Zero - sync  build v1.5.0, Play store build v1.5.0
**Repro on:** Samsung Galaxy S6 (7.0), Samsung Galaxy S7 (8.0.0)
 
**Repro Steps:** 
1) Launch the app and register into the app as User A from device1. 
2) Register into the app as User B from device2.
3) User A sends a ‘cUSD’ to user B. 
4) Make sure the transaction process successful. 
5) Observe on user B from device 2 ‘cUSD’ balance amount is not get updated automatically

**Investigation:** 
1) ‘cUSD’ balance is shown updated after user sending ‘cUSD’ to other users. 
2) ‘cUSD’ and ‘cGLD’ balance is shown updated when user exchange ‘cUSD’ to ‘cGLD’ or ‘cGLD’ to ‘cUSD’.
3) Balance amount is shown updated after refresh the balance.

**Impact:** User is not updated with the current balance without refreshing the page.  
**Current Behavior:**  ‘cUSD’ balance is not updated automatically 
**Expected Behavior:** “cUSD” balance should update automatically. 
**Attachment:** [Android_CeloDollarsUpdateIssue.mov](https://drive.google.com/file/d/1hpn8LVpSHaS8eb9lOAxLLpjHloQQGWbX/view?usp=sharing) (Bug observe after 44  sec)

i1skn:
### Description
If transaction list has changed we need to update the balance.
For good UX we want new transactions appear at the same time we update balance. To achieve this we optimistically re-calculate balance, based on the new transactions and show it to a user until we have new balance from the contract

### Tested

Ran in iOS simulator.

### Other changes

Remove `currency` parameter from `mapTransferStandbyToFeedItem` method because it was not used there.

### Related issues

- Fixes #1335

### Backwards compatibility

Yes


packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -234,7 +239,22 @@ export class TransactionsList extends React.PureComponent<Props> {
...
+    // new balance from the contract
+    if (this.props.dollarBalanceLastFetch) {
+      const newBalance = transactions.reduce((acc, tx) => {
+        if (tx.timestamp * 1000 > this.props.dollarBalanceLastFetch!) {
+          return acc.plus(tx.amount.value)
+        }
+        return acc
+      }, new BigNumber(this.props.dollarBalance || 0))
+      this.props.setBalance(newBalance.toString())
+    }
```
jeanregisser: `TransactionsList` is used for both the cUSD and cGLD feeds (home and gold tabs).
`fetchDollarBalance`/`dollarBalanceLastFetch`/`setBalance` should be made generic and dependant on the `currency` prop passed to `TransactionsList`, as is it will set an incorrect cUSD balance when the component is used from the gold tab.

I see the call here to refresh the balance is going to duplicate the work done in https://github.com/celo-org/celo-monorepo/blob/25feb786595f56c1f88c9ec99be0f3f5dcbf91c6/packages/mobile/src/home/saga.ts#L68 since the GraphQL query is also doing polling. Looks like we could skip the manual fetch here given this info.

I think there's an edge case leading to an incorrect balance if this callback is called more than once and the `fetchDollarBalance` from the contract hasn't had time to run.
Can we make the optimistic balance update be resistant to this?

jeanregisser (commented):
Nice to the see the optimistic update 👍

packages/mobile/src/transactions/TransactionsList.tsx
jeanregisser: Ah my bad, I see `setBalance` also updates `lastFetch` so please ignore my comment about the edge case.

jeanregisser: `tx.timestamp` is already in milliseconds, we shouldn't multiply it.

i1skn: Ah, ok. I've just looked in tests: https://github.com/celo-org/celo-monorepo/blob/cd7be93a3abd1333d7972765f23cfa70d17cc492/packages/mobile/src/transactions/TransactionsList.test.tsx#L71 Should we update the mocks then?

packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -234,7 +239,22 @@ export class TransactionsList extends React.PureComponent<Props> {
...
+    // Transaction list has changed and we need to update the balance.
+    this.props.fetchDollarBalance()
+
+    // For good UX we want new transactions appear at the same time we update
+    // balance. To achieve this we optimistically re-calculate balance,
+    // based on the new transactions and show it to a user until we have
+    // new balance from the contract
+    if (this.props.dollarBalanceLastFetch) {
+      const newBalance = transactions.reduce((acc, tx) => {
+        if (tx.timestamp * 1000 > this.props.dollarBalanceLastFetch!) {
```
annakaz: Is ! non null redundant given the ` if (this.props.dollarBalanceLastFetch)` above?

annakaz (commented):
Thanks for fixing this! Just confirming this will also update for Gold/Dollar exchange transactions since the gold balance will only change if the stabletoken balance did too?

packages/mobile/src/transactions/TransactionsList.tsx
i1skn: Yeah, I though so as well, but tslint gives me an error without `!`. Probably because it's in lamdba inside reduce function. Not sure...

i1skn:
@jeanregisser @annakaz I've removed the part with optimistic update because it was getting complicated and essentially becoming a duplicate of code in the contract. I've tested it a couple of times and usually the balance gets refreshed in 1-2 sec, which is a reasonable delay to my opinion. Please take a look and let me know your thoughts! Thanks!

jeanregisser:
😢  I really think we should keep the optimistic update. 
To me it's very important for the user to have a coherent balance. Having a difference between what is happening in the feed and the balance itself doesn't inspire confidence.

annakaz:
Curious which cases this updates faster for than the outgoing transaction refresh here: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/transactions/saga.ts#L57 and incoming transactions here: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/firebase/notifications.ts#L56 ?

i1skn:
@annakaz as far as I understand https://github.com/celo-org/celo-monorepo/blob/70aedb77b8224a8ba36d6d6fa8ecaba7356dd166/packages/mobile/src/transactions/saga.ts#L57 is only called after user initiated send transaction is confirmed, so it's not our case here.

As for https://github.com/celo-org/celo-monorepo/blob/70aedb77b8224a8ba36d6d6fa8ecaba7356dd166/packages/mobile/src/firebase/notifications.ts#L56
it should be called, when user receives notification about received transaction, which is currently broken https://github.com/celo-org/celo-monorepo/issues/1339. I think this is the true reason we observe this bug now, cause before user would receive a notification and balances would be refreshed. 

#### Let me describe the current state:
1. The only source of truth for balance is smart contract
2. We receive new transactions for the feed from GraphQL
3. We trigger balance re-fresh when we receive a push notification
4. ~Because of https://github.com/celo-org/celo-monorepo/issues/1339 user does not receive notifications anymore, hence balance is not updated when new transactions arrive~ (#1339 is fixed)

#### Issues I see:
1. Push notification is an unreliable source to trigger the balance update, which is essential for user, when transaction arrives.
2. Because we fetch balance and transactions from two different sources, for some small amount of time (1-2 sec usually) discrepancy between them could be observed.

#### Proposal:
1. We move balance refresh trigger to the TransactionList. This way we do not rely on push notifications, but rather on GraphQL.
2. Create a separate issue and include product to discussion on how to mitigate it. Cause we can go different ways about removing this time delay (showing loader instead of balance, moving balance calculation to GraphQL, doing optimistic update on the wallet side, etc.).

Big thanks to @annakaz and @jeanregisser, your comments/questions allowed me to go deeper into this issue and understand the root cause of it. Let me know what do you guys think.

jeanregisser:
Very good summary of the different moving parts 👍 
Voting for proposal 2.

jmrossy:
@i1skn I'd recommend we trigger the balance refresh whenever we see a change in the tx list, and add some visual indicator to show the balance is stale for the few seconds while we refresh.

But please do feel free to start a thread about it with Corey and Nitya as well, would be good to get there thoughts.

coreycelo:
Another option could be to slow down the transaction list update. The ideal I think is for the balance and transaction list to update simultaneously and wonder if we could wait to update the transaction from "Confirming..." to done until after the balance is ready. The downside is It might make us seem slower than we are. How reliably is the delay 1-2s (and not much longer)?

Honestly don't know if that's ^ a good idea. If not, I'd lean towards Rossy's suggestion. I have some desire to avoid displaying the balance in any kind of "deteriorated" state that might imply unreliability.

i1skn:
Ok, I've created a separate issue to implement the smooth update: https://github.com/celo-org/celo-monorepo/issues/2647.

This PR scope is only fix https://github.com/celo-org/celo-monorepo/issues/1335. 
The issue appeared only because push notifications were not working properly atm and this showed us, that we should not rely on them (because generally there is no guarantee, that they would be delivered). Hence, we move balance refresh trigger to the TransactionList. This way we do not rely on push notifications, but rather on GraphQL.

jeanregisser (approved):
LGTM 👍

Issue #2541, opened 2020-01-24
Exchange commands sometimes run out of gas


nategraf:
### Expected Behavior

As observed by Chainflow on Discord using the `exchange:dollars` command will sometimes fail. An example of this are transactions `0xfdfe2f8ed85a04e63a549bc4d27655220e8d079e788aff8e555a1a34690bec37` and `0x96e0fe5d75984e8113492f529013f2e07d130add37b76c4b5b3cb01708e7b213` on the Baklava phase 2 network. The first appears to have run out of gas, while the seconds succeeded. Both were the result of running the command with the same parameters. 

It seems likely that this is a result of the `exchange` smart contract method having a state-dependent step to update the uniswap buckets. If the transaction gas estimation is run against a state that would not result in an update, that gas will not be added to the gas limit. If the transaction is included in a block were it would update the buckets, the transaction will subsequently fail.

https://github.com/celo-org/celo-monorepo/blob/ee4ffdd9cc8941cd930915e4772c64bd2f5b4c43/packages/protocol/contracts/stability/Exchange.sol#L101

One potential solution to this is to measure the upper bound of this transaction and set it in the CLI instead of using gas estimation.

mrsmkl:
Looks like most gas used by successful transaction is 211518. These probably succeed because the transferred value is 0. Probably setting to 300000 or something should work.

PR #2552, opened 2020-01-27
Add explicit gas to exchange transactions to prevent errors

nategraf:
### Description

Add explicit `gas` parameter to exchange transactions, with suggested value from @mrsmkl,  in the CLI to prevent an `out of gas` error that randomly effects CLI users as detailed by https://github.com/celo-org/celo-monorepo/issues/2541

### Tested

Manually running the command against a testnet about 10 times.

### Related issues

- Related https://github.com/celo-org/celo-monorepo/issues/2541

### Backwards compatibility

100%


packages/cli/src/commands/exchange/dollars.ts
```diff
@@ -41,6 +41,7 @@ export default class ExchangeDollars extends BaseCommand {
     await displaySendTx('approve', stableToken.approve(exchange.address, sellAmount.toFixed()))
 
     const exchangeTx = exchange.exchange(sellAmount.toFixed(), minBuyAmount!.toFixed(), false)
-    await displaySendTx('exchange', exchangeTx)
+    // Set explciit gas based on github.com/celo-org/celo-monorepo/issues/2541
```
mrsmkl, line 44: Spelling.

mrsmkl (approved):
LGTM. I think we should keep the issue open so that we remember to check that we now have enough gas.

PR #2569, opened 2020-01-28
Voting bot can exclude groups from its consideration

### LINKED ISSUE #2538: Voting Bot SNBAT vote for C-Labs validators
### Expected Behavior

The purpose of voting bot is to foster competition among TGCSO participants. C Labs validators should not be eligible for votes from the bot.

### Current Behavior

The bot considers all registered validators.

yerdua:
### Description

In order to help TGCSO be more dynamic and competitive, we want to take C Labs validators and validator groups out of it. They should not be eligible for votes from the voting bot. This PR adds the ability to exclude groups by address, as determined when deploying.

### Tested

I ran `celotool bots auto-vote --celoProvider https://baklavastaging-forno.celo-testnet.org --excludedGroups $SOME_CLABS_GROUPS`, with some extra logging and verified that they were removed from the considered set, and that voting still worked.

I ran `celotool deploy initial voting-bot -e baklavastaging --excludedGroups $SOME_CLABS_GROUPS` with `echo {{ .Values.votingBot.excludedGroups }}` in the helm command to verify that the values made it though.

### Related issues

- Fixes #2538


packages/celotool/src/cmds/bots/auto-vote.ts
```diff
@@ -19,21 +19,35 @@ export const describe = 'for each of the voting bot accounts, vote for the best
...
+      default: 'http://localhost:8545',
+    })
+    .option('excludedGroups', {
+      type: 'string',
+      description: 'A comma separated list of groups to exclude from voting eligibility',
+      coerce: (addresses: string) => {
+        return addresses
+          .split(',')
+          .filter((a) => a.length > 0)
+          .map(ensure0x)
```
jfoutts-celo: Instead of `ensure0x`, consider something like '0x' + `normalizeAddress` from @celo/utils/lib/address, here and elsewhere.  Any mapping from an address, e.g.`excludedGroups`, should normalize the keys.  This normalizes for '0x' prefix but I don't think for capitalization.

jfoutts-celo (approved):

packages/celotool/src/cmds/bots/auto-vote.ts
yerdua: oh, good catch! Thank you!

PR #2577, opened 2020-01-28
Freeze initial reserve gold

### LINKED ISSUE #2390: Initial reserve cGLD transferability

jfoutts-celo:
### Description

Freezes the initial reserve gold according to a linear parameterization.

### Tested

Added unit tests.

### Related issues

- Fixes #2390


jfoutts-celo:
Will the reserve already have it's initial balance at the time of `initialize()`?

This relies on block.timestamp (as does the spending limits which I copied).  Would it be better to use block.number?

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -278,6 +306,24 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
     return reserveGoldBalance;
   }
 
+  function getUnfrozenReserveGoldBalance() public view returns (uint256) {
+    uint256 reserveGoldBalance = getUnfrozenBalance();
+    for (uint256 i = 0; i < otherReserveAddresses.length; i++) {
+      reserveGoldBalance = reserveGoldBalance.add(otherReserveAddresses[i].balance);
+    }
```
rcroessmann: Nitty: This is also found in getReserveGoldBalance() and for readability it might be worth to create a separate function getOtherReserveAddressesGoldBalance() and to then use this function in getUnfrozenReserveGoldBalance() and getReserveGoldBalance().

rcroessmann:
This looks good to me but would love to have a review from you @asaj as well as I am still pretty new to this stuff. 

Also, I left a comment in https://github.com/celo-org/celo-monorepo/pull/2395 a while back that I think applies now:  "If we decide to do the above, we should make sure that the unfrozenReserveGoldBalance is used instead of reserveGoldBalance in the reserve ratio calculation as soon as #2390 is tackled." Should we create an issues for that?

packages/protocol/contracts/stability/Reserve.sol
jfoutts-celo: Good call. Updated.

jfoutts-celo:
Should create an issue for more discussion?

If we're sure that `computeTobinTax()` should use `getUnfrozenReserveGoldBalance()` instead of `getReserveGoldBalance()` then we can put into this PR.

But we'll need to be careful with the parameters, as they could trigger Tobin tax right after launch.

asaj:
> If we're sure that computeTobinTax() should use getUnfrozenReserveGoldBalance() instead of getReserveGoldBalance() then we can put into this PR.

This seems like the right approach to me

rcroessmann:
> > If we're sure that computeTobinTax() should use getUnfrozenReserveGoldBalance() instead of getReserveGoldBalance() then we can put into this PR.
> 
> This seems like the right approach to me

+1 on adding it to this PR

rcroessmann:
Maybe also worth putting into this PR are two other things that should be updated to use unfrozenReserveGoldBalance. 
1. The exchange cGLD tank size: Currently, tanks are set as a fraction of the reserveGoldBalance. This could lead to a situation where the exchange contract tries to transfer more cGLD to a user as is currently unfrozen. So it seems like we would need to use the unfrozenReserveGoldBalance in this line https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/Exchange.sol#L303
2. The spending limit: Reserve spenders can currently still (try to) spend x% of the reserveGoldBalance each day but the pct. should be applied to the unfrozenReserveGoldBalance. https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/Reserve.sol#L241

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -281,14 +303,38 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
     return weights;
   }
 
+  function getUnfrozenBalance() public view returns (uint256) {
+    uint256 balance = address(this).balance;
+    uint256 frozenReserveGold = getFrozenReserveGoldBalance();
+    return balance > frozenReserveGold ? balance - frozenReserveGold : 0;
```
asaj: nit: use safemath

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -281,14 +303,38 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
...
+  }
+
   function getReserveGoldBalance() public view returns (uint256) {
-    uint256 reserveGoldBalance = address(this).balance;
+    return address(this).balance + getOtherReserveAddressesGoldBalance();
```
asaj: ditto

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -281,14 +303,38 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
...
+  function getOtherReserveAddressesGoldBalance() public view returns (uint256) {
+    uint256 reserveGoldBalance = 0;
     for (uint256 i = 0; i < otherReserveAddresses.length; i++) {
       reserveGoldBalance = reserveGoldBalance.add(otherReserveAddresses[i].balance);
     }
     return reserveGoldBalance;
   }
 
+  function getUnfrozenReserveGoldBalance() public view returns (uint256) {
+    return getUnfrozenBalance() + getOtherReserveAddressesGoldBalance();
```
asaj: ditto

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -281,14 +303,38 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
...
+  }
+
+  function getFrozenReserveGoldBalance() public view returns (uint256) {
+    uint256 currentDay = now / 1 days;
+    uint256 frozenDays = currentDay - frozenReserveGoldStartDay;
+    if (frozenDays >= frozenReserveGoldDays) return 0;
+    return
+      frozenReserveGoldStartBalance -
```
asaj: ditto

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -281,14 +303,38 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
...
     return reserveGoldBalance;
   }
 
+  function getUnfrozenReserveGoldBalance() public view returns (uint256) {
+    return getUnfrozenBalance() + getOtherReserveAddressesGoldBalance();
+  }
+
+  function getFrozenReserveGoldBalance() public view returns (uint256) {
+    uint256 currentDay = now / 1 days;
+    uint256 frozenDays = currentDay - frozenReserveGoldStartDay;
```
asaj: ditto

packages/protocol/migrationsConfig.js
```diff
@@ -96,6 +96,8 @@ const DefaultConfig = {
     goldBalance: 100000000,
     tobinTaxStalenessThreshold: 60 * 60, // 1 hour
     dailySpendingRatio: '1000000000000000000000000', // 100%
+    frozenGold: 0,
```
asaj, line 99: @rcroessmann what do you suggest for these values?

asaj (approved):
LGTM pending changes

packages/protocol/contracts/stability/Reserve.sol
jfoutts-celo: Thanks. I forgot there's no operator overloads in Solidity.  Created an issue to add a SafeMath always lint: https://github.com/celo-org/celo-monorepo/issues/2810

rcroessmann (approved):
LGTM, pending failed tests.

PR #2581, opened 2020-01-28
Deploy alfajoresstaging and alfajores

cmcewen:
### Description

Changes for deploying alfajoresstaging

### Tested

Running on the network now


.env.alfajoresstaging
```diff
@@ -77,11 +79,10 @@ FAUCET_CUSD_WEI=60000000000000000000000
 VALIDATORS=10
 PROXIED_VALIDATORS=5
 TX_NODES=5
-STATIC_IPS_FOR_GETH_NODES=true
+STATIC_IPS_FOR_GETH_NODES=false
```
jmrossy, line 82: Why did we toggle this?

jmrossy (commented):
LGTM, just one question below

.env.alfajoresstaging
cmcewen: Since there aren't any external validators, we don't need to use static ips so other nodes can peer

jmrossy (approved):

PR #2584, opened 2020-01-28
[Wallet] Clean up forno mode

### LINKED ISSUE #2583: [Wallet] Geth still sycning if forno initially enabled
### Expected Behavior

Geth related sagas turned off when using forno

### Current Behavior

Geth related sagas run if forno is initially enabled. They do not run if forno is toggled from the DataSaver screen.

### LINKED ISSUE #2585: [Wallet] Users who turn off forno should always see forno turned off
### Expected Behavior

When users turn off forno, forno is turned off regardless of whether a restart is required (required if geth was already stared in the given app session)

### Current Behavior

When users turn off forno, if a restart is not required, then forno is turned on instead of off

annakaz:
### Description

Fixes a couple bugs when the app that only appear when the app is build with `FORNO_ENABLED_INITIALLY=true`. The geth saga starts unnecessarily, and the web3 sync can fail if geth isn't yet connected. This fixes it by only starting gethSaga when not in forno mode, and by waiting for geth to be connected before checking web3 sync. 

Note that the bulk of this PR is the forno/zeroSync rename so I have annotated the non rename changes to make it easier to review. This rename is necessary because while I had initially used zeroSync mode to make it clear the app is not syncing, we have been using the term forno internally (not zeroSync) so wanted to make the codebase consistent. 

### Tested

Tested a default forno build turning on and off forno. Tested a default geth build turning on and off forno. 

### Related issues

- Fixes #2583
- Fixes #2585 (forno does not turn off when no app restart needed because of restart modal)


packages/mobile/src/account/DataSaver.tsx
```diff
@@ -122,48 +122,52 @@ export class DataSaver extends React.Component<Props, State> {
...
+    this.props.toggleFornoMode(false)
     navigateBack()
   }
 
-  handleZeroSyncToggle = (zeroSyncMode: boolean) => {
-    if (!zeroSyncMode && this.props.gethStartedThisSession) {
-      // Starting geth a second time this app session which will
-      // require an app restart, so show restart modal
-      this.showSwitchOffModal()
+  handleFornoToggle = (fornoMode: boolean) => {
```
annakaz, line 139: Fixes if else- if switching off forno mode, show the modal if need be, else just turn it off. If switching on forno mode, show switch on modal. 

Previously, if switching off forno mode and modal not necessary, this would switch forno on (opposite of intended behavior)

packages/mobile/src/geth/saga.ts
```diff
@@ -218,3 +219,9 @@ export function* gethSaga() {
   yield cancel(gethRelatedSagas)
   yield put(setGethConnected(true))
 }
+
+export function* gethSagaIfNecessary() {
```
annakaz, line 225: Only start geth saga if using geth

packages/mobile/src/redux/selectors.ts
```diff
@@ -27,9 +28,11 @@ export const isBackupTooLate = (state: RootState) => {
 export const getNetworkConnected = (state: RootState) => state.networkInfo.connected
 
 export const isAppConnected = createSelector(
+  fornoSelector,
   isGethConnectedSelector,
   getNetworkConnected,
-  (gethConnected, networkConnected) => gethConnected && networkConnected
+  (fornoEnabled, gethConnected, networkConnected) =>
```
annakaz, line 34: Only care if gethConnected if geth is being using

packages/mobile/src/shared/DisconnectBanner.tsx
```diff
@@ -39,15 +41,16 @@ class DisconnectBanner extends React.PureComponent<Props> {
   }
 
   render() {
-    const { t, appConnected, appSynced } = this.props
+    const { t, appConnected, appSynced, fornoEnabled } = this.props
+    const appSyncedIfNecessary = appSynced || fornoEnabled
```
annakaz, line 45: Only care if app is synced if geth is being used

packages/mobile/src/web3/saga.ts
```diff
@@ -455,71 +454,71 @@ export function* switchToGethFromZeroSync() {
     }
 
     yield spawn(gethSaga)
-
     switchWeb3ProviderForSyncMode(false)
-    // Ensure web3 is fully synced using new provider
+    yield take(GethActions.SET_GETH_CONNECTED)
```
annakaz: Wait for geth to be connected before starting web3 sync

annakaz:
Note that #2583 was introduced by #2526 as I did not test with a default forno build

jeanregisser (approved):
Looks great 👍 Thanks for making the naming consistent!

packages/mobile/src/web3/saga.ts
jmrossy: Should we use `waitForGethConnectivity` here like this line: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/web3/saga.ts#L133

jmrossy (commented):
Looks great! Just one question about `waitForGethConnectivity`

packages/mobile/src/web3/saga.ts
annakaz: Yes, updated!

annakaz:
Rearranged some forno-related state:
- moved `gethStartedThisSession` to `/geth` as it should not be persisted across app restarts like other `/geth` state variables
- moved `promptFornoIfNeeded` to `/account`
https://github.com/celo-org/celo-monorepo/pull/2584/commits/a34b52662157a0319aa5cb3df823c8d3e6a10baf

jmrossy (approved):

Issue #2006, opened 2019-12-03
--verbose not working in celotooljs.sh deploy initial for helm packages


jcortejoso:
If using `--verbose` with command `./bin/celotooljs.sh deploy initial mypackage -e myenv`, if `mypackage` deployment consists on a helm package, the deployment fails with the next error:

```
Building any chart dependencies...
$ helm dep build ../helm-charts/leaderboard --dry-run --debug
Error: unknown flag: --dry-run

Error: unknown flag: --dry-run

celotooljs deploy initial leaderboard

deploy the leaderboard for the specified network

Options:
  --version       Show version number                                                                                                                                                                                                [boolean]
  --verbose       Whether to show a bunch of debugging output like stdout and stderr of shell commands                                                                                                              [boolean] [default: false]
  --yesreally     Reply "yes" to prompts about changing staging/production (be careful!)                                                                                                                            [boolean] [default: false]
  --help          Show help                                                                                                                                                                                                          [boolean]
  --celo-env, -e  the environment in which you want to execute this command                                                                                                                                                         [required]

{ Error: Command failed: helm dep build ../helm-charts/leaderboard --dry-run --debug
Error: unknown flag: --dry-run

    at ChildProcess.exithandler (child_process.js:294:12)
    at ChildProcess.emit (events.js:198:13)
    at ChildProcess.EventEmitter.emit (domain.js:448:20)
    at maybeClose (internal/child_process.js:982:16)
    at Socket.stream.socket.on (internal/child_process.js:389:11)
    at Socket.emit (events.js:198:13)
    at Socket.EventEmitter.emit (domain.js:448:20)
    at Pipe._handle.close (net.js:606:12)
  killed: false,
  code: 1,
  signal: null,
  cmd:
   'helm dep build ../helm-charts/leaderboard --dry-run --debug' }
```

The problem seem to be in the next code:
```
async function helmCommand(command: string) {
  if (isCelotoolVerbose()) {
    await execCmdWithExitOnFailure(command + ' --dry-run --debug')
  }

  await execCmdWithExitOnFailure(command)
}
```


Without `--verbose` option it works as expected.

PR #2589, opened 2020-01-29
Multiple blockscout instances

### LINKED ISSUE #2532: As a blockscout admin I SBAT re-index/deploy in production environment without affecting the service
### Expected Behavior

Is expected for Blockscout to support a blue-green deployment strategy.
 
When new versions of `blockscout` are released requiring re-indexing, this process can be done in production deploying in a "no visible" environment for the final user.
When the re-indexing is finished and is confirmed that all the functionalities are working as expected, the admin can redirect the traffic in the ingress to the new blockscout deployment. 

### Current Behavior

The deployment of new releases of blockscout requiring re-indexing result in a non reliable execution of blockscout during the time the re-indexing is happening (currently between 1 and 2 hours). In future mainnet deployments, with a network of several millions of blocks/transactions a complete re-index could take even more.

jcortejoso:
### Description

Allow running multiple instances of blockscout, so we can switch between them for avoiding fetching issues when a new schema is deployed.
There is a common ingress resource (managed outside helm) for $env-blockscout.$domain that is supposed to be pointing to the stable/non-syncing blockscout instance.

### Tested

Tested in a local testnet.

### Other changes

- Updated the warning: `WARNING: Flag --gce-zone is deprecated and will be removed by release 255.0.0. Use --zone instead.`
- Fixed the `--verbose` when `helm dep build` is invoked.
- Refactor to avoid running `add-iam-policy-binding` when policy already binded to that role.
- Removed `--export` in `kubectl get` as it is deprecated and not included in latest `kubectl` v1.17.

### Related issues

- Fixes #2532 and #2006 

### Backwards compatibility

Can be some problems with current deployments of blockscout that may require to sync the helm package outside celotool.


packages/helm-charts/blockscout/templates/_helpers.tpl
```diff
@@ -54,8 +54,6 @@ volumes:
   value: prod
 - name: LOGO
   value: /images/celo_logo.svg
-- name: CHAIN_SPEC_PATH
```
tkporter, line 57: Just want to make sure-- this is for sure able to be removed?

.env.baklava
```diff
@@ -19,8 +19,6 @@ BLOCKSCOUT_WEB_REPLICAS=3
 # existing database.
 BLOCKSCOUT_DB_SUFFIX="1"
 BLOCKSCOUT_SUBNETWORK_NAME="Baklava"
-# Needed by blockscout to get the balances from the baklava genesis block
```
nambrot, line 22: Is this supposed to be removed here?

nambrot (approved):
Maybe a bit more documentation about the upgrade path for existing deployments would be nice

packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -577,8 +587,10 @@ async function helmParameters(celoEnv: string) {
 }
 
 async function helmCommand(command: string) {
-  if (isCelotoolVerbose()) {
+  if (isCelotoolVerbose() && !command.includes(' dep build ')) {
```
tkporter, line 590: nice, this has been annoying

I take it the surrounding spaces aren't an issue?

tkporter (approved):
So is the flow switching entirely to a new deploy like this?

1. Change the DB suffix to something new
2. `celotooljs deploy initial blockscout ....`
3. `celotooljs deploy switch blockscout ....`
4. Change the DB suffix back to the old one
5. `celotooljs deploy destroy blockscout ....`
6. Change the DB suffix back to the new one

.env.baklava
jcortejoso: Now `CHAIN_SPEC_PATH` is optional (if it is not present it will try to download from gpc default bucket using `BLOCKSCOUT_SUBNETWORK_NAME`). If present you can setup the url for the chainfile or only a json with the rewards as before (https://github.com/celo-org/blockscout/pull/70/files)

packages/helm-charts/blockscout/templates/_helpers.tpl
jcortejoso: Yeap, it is optional now (https://github.com/celo-org/blockscout/pull/70/files)

packages/celotool/src/lib/helm_deploy.ts
jcortejoso: Should not be. The command is `helm dep build ...`. I tried to be as restrictive as possible but I guess there is no real case difference between with/out spaces.

jcortejoso:
@nambrot so migrating from this approach from previous one have problems related to some k8s resources that changed the name (which is not possible), so help upgrade cannot apply. 
The easiest way to proceed is destroying the package with celotool, and deploying initially again, but this would recreate the database.

If you want to keep the DB the way to proceed is manually deleting the existing blockscout helm release (`helm delete --purge $env-blockscout`), manually backup&destroy the secret `env-blockscout` and create it back with name `$env-blockscout$suffix`, and deploy manually the blockscout helm package passing the values as celotool would do.

I have not tested it but also could be possible renaming (create&destroy) the secret and use `celotool deploy migrate blockscout...`.

PR #2591, opened 2020-01-29
[Wallet] Fix gold charts staying in loading state when USD is the selected currency

### LINKED ISSUE #2494: Continuous loading indicator and “Loading Exchange Rate...” is shown when a user navigates to the “Gold” tab.
**Frequency:** 100%
**App version:** Integration build v1.5.2 (101548777) 
**Repro on:** Google Pixel XL 2 (9.0), Samsung Galaxy A8+ (8.0.0) 

**Repro Steps:** 
1) Launch the app.  
2) Tap on “Gold” tab. 
3) Observed  

**Investigation:** 
1. Issue also repro even if a user has Gold activity. 
2. The gold exchange rate is not shown 

**Impact:** Users will not able to see the Gold Activity graph and exchange rate. 
**Current Behavior:** Continuous loading indicator and “Loading Exchange Rate...” is shown
**Expected Behavior:** Gold activity graph should be shown and the exchange rate should be calculated 
**Attachment:**[Android_LoadingIndicatorIsShown.mp4](https://drive.google.com/file/d/13nNK56xfRYcxE17ffowj-7w5OO4QyUZZ/view?usp=sharing) (Bug Observed after 3 sec)

jeanregisser:
### Description

Fix gold charts staying in loading state when USD is the selected currency.

### Tested

image

### Other changes

None

### Related issues

- Fixes #2494

### Backwards compatibility

Yes.

### DIFF
--- a/packages/mobile/src/exchange/CeloGoldHistoryChart.tsx
+++ b/packages/mobile/src/exchange/CeloGoldHistoryChart.tsx
@@ -158,7 +158,7 @@ function CeloGoldHistoryChart({ t, testID }: Props) {
   ])
   const localExchangeRate = useSelector(getLocalCurrencyExchangeRate)
   const dollarsToLocal = useCallback(
-    (amount) => convertDollarsToLocalAmount(amount, localExchangeRate),
+    (amount) => convertDollarsToLocalAmount(amount, localCurrencyCode ? localExchangeRate : 1),
     [localExchangeRate]
   )
   const [range] = useState(30 * 24 * 60 * 60 * 1000) // 30 days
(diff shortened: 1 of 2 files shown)


packages/mobile/src/exchange/CeloGoldHistoryChart.tsx
```diff
@@ -158,7 +158,7 @@ function CeloGoldHistoryChart({ t, testID }: Props) {
   ])
   const localExchangeRate = useSelector(getLocalCurrencyExchangeRate)
   const dollarsToLocal = useCallback(
-    (amount) => convertDollarsToLocalAmount(amount, localExchangeRate),
+    (amount) => convertDollarsToLocalAmount(amount, localCurrencyCode ? localExchangeRate : 1),
```
jmrossy, line 161: Nit: Could we keep things simpler by just having `useSelector(getLocalCurrencyExchangeRate)` return `1` when there is no localCurrencyCode?

jeanregisser: Yes that would probably work as well. I don't want to make this change in this PR to avoid breaking something else unexpectedly (though unlikely) but as we continue to work on improving local currencies, this will be simplified.

jmrossy (approved):

Issue #2601, opened 2020-01-29
Validators SNBAT use the attestation signer key or vote signer key for managing validators/groups


yorhodes:
### Expected Behavior

`getAccounts().validatorSignerToAccount(msg.sender)` for all Validators.sol account related actions

### Current Behavior

What is the current behavior?

`getAccounts().signerToAccount(msg.sender)` is used for Validator registration, affiliation, and key rotation

This could lead to a scenario where the ephemeral attestation or vote signing key is compromised and someone's validator is manipulated as a result

Changed on #1643

yorhodes:
the `DowntimeSlasher.sol` and `DoubleSigningSlasher.sol` also use `signerToAccount` despite explicitly a Validator signer key

mrsmkl:
In `DowntimeSlasher.sol` and `DoubleSigningSlasher.sol` we have to use `signerToAccount` because `validatorSignerToAccount` requires the current validator signer.

mrsmkl:
Looks like these have already been changed.

PR #2604, opened 2020-01-30
Add docs for Accounts key/signer management

### LINKED ISSUE #794: celotool terraform commands should always init before
### Expected Behavior

`terraform init` always run before another celotool command that is not a part of the deployment process (like refresh, output, but not apply)

### Current Behavior

Some commands do not run `terraform init` before. This is bad for cases where a backend switch is required

### LINKED ISSUE #1762: Internal audit of signer/account usage
### Expected Behavior

We should:
- have an internal audit that we are using signers correctly, 
- that they are in the right spots
- determine that we have appropriate tooling
- documenting their existence as well as how a key rotation, especially for validating, looks like

### Current Behavior

In `Accounts.sol`, we have so called "signers" to which accounts can delegate specific actions to. As of right now, they are for:
- voting
- validating
- attestation signing

yorhodes:
### Description

Adds documentation around `Accounts` key management and rotation
Scripting Validator key rotation reliably is challenging and has the added benefit of acting as a differentiator for Validator BTUs on baklava

### Tested

Zero downtime key rotation on `baklava-staging`

### Related issues

- Fixes #1762 
- Fixes #794
- Opens #2601


hqueue:
Thanks for the PR 👍  As a participant of the stake-off, we are looking for this guide :)

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,67 @@
+# Key management
+
+There is a pretty funny [tweet](https://twitter.com/LeaKissner/status/1198595109756887040) which goes something like this:
+
+> Crypto is a tool for turning a whole swathe of problems into key management problems. Key management problems are way harder than (virtually all) cryptographers think.
```
nategraf: I like this a lot. We can certainly include it, but we should rewrite the opening line to avoid the first-person voice. Maybe cite the tweet below, instead of presenting it like a friend telling a joke.

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,67 @@
...
+
+To help with this, we have various mechanisms to reduce the impact of the loss or compromise of keys, mainly through our `Accounts` smart contract. On it, accounts can authorize other keys to perform certain actions on behalf of the account. Keys that need to be accessed frequently (e.g. for signing blocks) are at greater risk of being compromised, and thus have more limited permissions, while keys that need to be accessed infrequently (e.g. for locking Celo Gold) are less onerous to store securely, and thus have more expansive permissions. Below is a summary of the various keys that are used in the Celo network, and a description of their permissions.
```
nategraf: Avoid the first-person voice. "We" are C Labs so it's important to distinguish the company/us from Celo tyhe platform.
```suggestion
Celo has various mechanisms to reduce the impact of the loss or compromise of keys, mainly through the `Accounts` smart contract. On it, accounts can authorize other keys to perform certain actions on behalf of the account. Keys that need to be accessed frequently (e.g. for signing blocks) are at greater risk of being compromised, and thus have more limited permissions, while keys that need to be accessed infrequently (e.g. for locking Celo Gold) are less onerous to store securely, and thus have more expansive permissions. Below is a summary of the various keys that are used in the Celo network, and a description of their permissions.
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+
+| Name of the key        | Purpose                                                                                                                                                                                                                                                               |
+| ---------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Account key            | This is the key with the highest level of permissions, and is thus the most sensitive. It can be used to lock and unlock Celo Gold, and authorize vote, validator, and attestation keys. Note that the account key also has all of the permissions of the other keys. |
```
nategraf: The last sentence isn't entirely true. Once a signer is authorized, the account key can no longer serve that function. (i.e. if a validator key is authorized, the account key can no longer sign consensus messages)
```suggestion
| Account key            | This key has the highest level of permissions, and is thus the most sensitive. It can be used to lock and unlock Celo Gold, and authorize vote, validator, and attestation keys. |
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+| Validator signer key   | This is the key that has permission to register and manage a Validator or Validator Group, and participate in BFT consensus.                                                                                                                                          |
```
nategraf: The BFT consensus part is the one I'd want to the reader to think about most in relation to this key, so I'd put it first.
```suggestion
| Validator signer key   | This key has permission to participate in BFT consensus, and register and manage a Validator or Validator Group.                                                                                                                                          |
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+| Vote signer key        | This key can be used to vote in Validator elections and on-chain governance.                                                                                                                                                                                          |
```
nategraf:
```suggestion
| Vote signer key        | This key can be used to vote in Validator elections and for on-chain governance proposals.                                                                                                                                                                                          |
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+| Attestation signer key | This key is used to sign attestations in Celo's lightweight identity protocol.                                                                                                                                                                                        |
+
+Note that account and signer keys must be unique and may not be reused.
```
nategraf:
```suggestion
{% hint style="info" %}
Each signer key must be unique and may not be reused. Once a signer key is authorized, the only way to deauthorize that key is to authorize a new key that has never been used before.
{% endhint %}
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Key Rotation
+
+Loss of an authorized key no longer has catastrophic implications, as the Account key can just authorize another key in that case. In general, it is good hygiene to rotate your keys, just like you should rotate your passwords.
```
nategraf: I think we should strengthen the second sentence. (It's also good hygiene to floss, but most reader aren't doing that)
```suggestion
Loss of an authorized key doesn't have to be catastrophic, as the Account key can just authorize another key in that case. It is strongly recommended to regularly rotate keys to limit the impact of an undiscovered compromise.
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Key Rotation for Consensus
+
+Key rotation for Consensus is a bit trickier. Let's assume that a validator is currently elected, actively signing consensus messages and wants to rotate their validator signer key. What they should do is the following:
```
nategraf: First-person voice.
```suggestion
Key rotation for Consensus is a bit trickier. Assuming that a validator is currently elected, actively signing consensus messages and wants to rotate their validator signer key. They should do the following:
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Bring up a new validator node and create a new Validator signer key (VSK) on it.
```
nategraf: We shouldn't use acronyms in a short context like this. The reader my be skipping around the doc and will not have read this line when the come across the acronym later.
```suggestion
- Bring up a new validator node and create a new Validator signer key on it.
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+- Authorize the new Validator signer key with the Account key to overwrite the old Validator signer key.
+
+```bash
+celocli account:authorize --from <ACCOUNT_KEY_ADDRESS> --role validator --signer <NEW_VSK_ADDRESS> --signature <PROOF_OF_NEW_VSK_POSSESSION>
+```
+
+**This should also update your ECDSA public key automatically.**
```
nategraf:
```suggestion
**This will also update your ECDSA public key automatically.**
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Update your BLS key used for consensus. Rotating both the ECDSA and BLS keys is recommended.
+
+```bash
+celocli validator:update-bls-public-key --from <NEW_VSK_ADDRESS> --blsKey <NEW_BLS_KEY> --blsPop <PROOF_OF_NEW_BLS_KEY_POSSESSION>
+```
+
+- Your old node will continue to sign consensus messages for the current epoch, but upon the next epoch will find itself no longer authorized.
+- Your new node will find itself unauthorized, but will be authorized to sign as the next epoch begins.
```
nategraf: The introduction to this list refers to "their validator", so using "your" in the next few lines is inconsistent. One or the other should change. We have used "you" for instructions elsewhere, so I think it's ok to use it here.

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+
+**Waiting for the epoch to flip before bringing down the retired validator node is essential to zero-downtime rotation when elected.**
```
nategraf: I think the terms "flip" and "zero-downtime" may not be immediately obvious, especially with readers who are not native English speakers, so I would reword this. (And add the warning callout)
```suggestion
{% hint style="warning" %}
The newly authorized keys will only take effect in the next epoch, so the node operating with the old key must remain running until the end of the current epoch to avoid downtime.
{% endhint %}
```

packages/docs/celo-codebase/protocol/identity/accounts.md
```diff
@@ -0,0 +1,75 @@
...
+
+`Validators` Expects user to be registered account signing with Validator key (or account key)
+
+`Attestations` Expects issuers to be registered accounts; expects completers to be signing with Attestation key (or account key)
+
+### Celo Blockchain Client
+
+`Consensus` Expects participants to be registered accounts and signing with Validator key
+
+`Precompiles` Keeps track of current validator set and proof of possession using authorized Validator key addresses
```
nategraf: This is a useful list a reference to developers, but I wouldn't include it here because this location is slightly higher level. Also, this list will surely become out-of-date unless it is kept somewhere it will be read frequently.

nategraf (commented):
In general, I've made some comments about using the first-person voice. We want to make sure to avoid saying "we" to avoid conflating C Labs and Celo. Also the warmth in the voice is nice, but this is technical documentation, so sometimes it's not the best for conveying complex information in a clear way.
Also, the protocol docs are generally for describing the protocol at a high level. The instructions should really go somewhere else, although I don't think we have an operating guide yet, but it may be time to start one. I am happy to commit this here though and we can move it into the operating guide when it is created.

packages/docs/celo-codebase/protocol/identity/accounts.md
yorhodes: haha okay @nambrot

PR #2606, opened 2020-01-30
[Wallet] Fix Pay & Decline buttons when open app from notification

### LINKED ISSUE #2567: Impossible to Pay or Cancel incoming payment request
### Expected Behavior

When app is in background, user receives notification about incoming payment request. Then user SBAT click on it and on the screen, which appears in the app user SBAT to Pay or Cancel the request.

### Current Behavior

Both buttons does not work

### Potential reason

Here:
https://github.com/celo-org/celo-monorepo/blob/fe529dff7651717445dd13d281612ade70f66e67/packages/mobile/src/transactions/actions.ts#L143
we do not pass `onConfirm` and `onCancel` as we do here:
https://github.com/celo-org/celo-monorepo/blob/fe529dff7651717445dd13d281612ade70f66e67/packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx#L45-L46

i1skn:
### Description

Fixes buttons `Pay` and `Decline` on the Send Confirmation screen, when the app was in background and opened by pressing on notification.

### Tested

Ran between Android device and emulator. Also deployed notification service to integration.

### Other changes

* Add firebase pendingRequest `uid` to notification payload, so we can identify the request, when app is opened from notification.
* Small boy scout refactoring and some little text changes. 

### Related issues

- Fixes #2567

### Backwards compatibility

Yes


packages/mobile/src/firebase/saga.ts
```diff
@@ -178,8 +182,25 @@ function* updatePaymentRequestStatus({ id, status }: UpdatePaymentRequestStatusA
...
+}
+
+function* cancelPaymentRequest({ id }: CancelPaymentRequestAction) {
+  CeloAnalytics.track(CustomEventNames.outgoing_request_payment_cancel)
+  yield call(updatePaymentRequestStatus, id, PaymentRequestStatus.CANCELLED)
+}
+
 export function* watchPaymentRequestStatusUpdates() {
-  yield takeLeading(Actions.PAYMENT_REQUEST_UPDATE_STATUS, updatePaymentRequestStatus)
+  yield takeLeading(Actions.PAYMENT_REQUEST_COMPLETE, completePaymentRequest)
```
jmrossy: The downside of this approach is that we are spawning 3x as many saga forks. I don't know if there's a perf impact to spawning saga forks, maybe it's negligible.

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -117,4 +109,6 @@ const styles = StyleSheet.create({
   },
 })
 
-export default withTranslation(Namespaces.paymentRequestFlow)(IncomingPaymentRequestListItem)
+export default connect<{}, DispatchProps, {}, RootState>(null, {
```
jmrossy: In general, it's better not to `connect` items in a list but instead just connect the parent container and pass in props as necessary. Not a big deal though

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -112,24 +119,35 @@ class SendConfirmation extends React.Component<Props, State> {
...
-    const { onConfirm } = this.getNavParams()
 
     this.props.sendPaymentOrInvite(
       amount,
       reason,
       recipient,
       recipientAddress,
       inviteMethod,
-      onConfirm
+      () => {
```
jmrossy: Can we add this logic to the saga instead of a callback here. We should try to keep our component classes as free of business logic as possible

jmrossy (commented):
Mostly looks good! Just a few comments

i1skn:
@jmrossy thanks a lot for the review! I've addressed the comments, PTAL!

jmrossy (approved):
🚢 it!

PR #2614, opened 2020-01-31
Adjust governance proposal CLI and refactor contract based on baklava feedback

yorhodes:
### Description

- Fix `governance.getDequeue` to not filter zeroes by default for index calculation (used by `approve`, `vote`, `execute`)
- Move `ProposalBuilder` usage to `propose` command to make proposal formation more obvious
- Refactor `isExpired` and `removeExpired` in Governance.sol to reduce confusion in how timings are calculated

### Tested

`yarn --cwd packages/protocol test governance`
[WIP]

### Related 

Fixes problems encountered while running proposal on baklava


packages/protocol/contracts/governance/Governance.sol
```diff
@@ -457,6 +457,38 @@ contract Governance is
     return proposalCount;
   }
 
+  /**
+   * @notice Expires a proposal if it is queued and expired.
+   * @param proposalId The ID of the proposal to remove.
+   * @return Whether the proposal was expired.
+   */
+  function removeIfQueuedAndExpired(uint256 proposalId) private returns (bool) {
+    bool result = queue.contains(proposalId) && isQueuedProposalExpired(proposalId);
```
asaj: nit: s/result/expired

packages/protocol/contracts/governance/Governance.sol
```diff
@@ -457,6 +457,38 @@ contract Governance is
...
+    }
+    return result;
+  }
+
+  /**
+   * @notice Requires a proposal is dequeued and removes it if expired.
+   * @param proposalId The ID of the proposal.
+   * @return The proposal storage struct and stage corresponding to `proposalId`. 
+   */
+  function requireDequeuedAndRemoveExpired(uint256 proposalId, uint256 index)
```
asaj: nit: "AndRemoveIfExpired"

asaj: Maybe "AndDeleteIfExpired"?

packages/protocol/contracts/governance/Governance.sol
```diff
@@ -471,31 +503,21 @@ contract Governance is
...
-    ) {
-      queue.remove(voter.upvote.proposalId);
-      emit ProposalExpired(voter.upvote.proposalId);
-    }
+    removeIfQueuedAndExpired(voter.upvote.proposalId);
+
     // We can upvote a proposal in the queue if we're not already upvoting a proposal in the queue.
     uint256 weight = getLockedGold().getAccountTotalLockedGold(account);
     require(weight > 0, "cannot upvote without locking gold");
-    require(isQueued(proposalId), "cannot upvote a proposal not in the queue");
```
asaj, line 500: Are you sure we can remove this?

packages/cli/src/commands/governance/propose.ts
```diff
@@ -32,6 +36,23 @@ export default class Propose extends BaseCommand {
...
       .runChecks()
 
+    const builder = new ProposalBuilder(this.kit)
+
+    // BUILD FROM JSON
+    const jsonString = readFileSync(res.flags.jsonTransactions).toString()
+    const jsonTransactions: ProposalTransactionJSON[] = JSON.parse(jsonString)
+    jsonTransactions.forEach((tx) => builder.addJsonTx(tx))
+
+    // BUILD FROM CONTRACTKIT FUNCTIONS
```
asaj, line 50: Remove? Or is this intended to be left as an example?

asaj (commented):
Nice work, this is much more readable

packages/protocol/contracts/governance/Governance.sol
yorhodes: nope, good catch

packages/cli/src/commands/governance/propose.ts
yorhodes: was leaving as an example

packages/protocol/contracts/governance/Governance.sol
```diff
@@ -457,6 +457,38 @@ contract Governance is
     return proposalCount;
```
nategraf, line 459: These changes are a useful refactor, but should be commit this right now given that the code is being audited in preparation to launch mainnet with audited contracts? We may need to start a new branch for any smart contract changes, but that is a heap of process outside the scope of this PR.

packages/cli/src/commands/governance/executehotfix.ts
```diff
@@ -23,7 +27,13 @@ export default class ExecuteHotfix extends BaseCommand {
   async run() {
     const res = this.parse(ExecuteHotfix)
     const account = res.flags.from
-    const hotfix = await buildProposalFromJsonFile(this.kit, res.flags.jsonTransactions)
+
+    const jsonString = readFileSync(res.flags.jsonTransactions).toString()
+    const jsonTransactions: ProposalTransactionJSON[] = JSON.parse(jsonString)
+
+    const builder = new ProposalBuilder(this.kit)
+    jsonTransactions.forEach((tx) => builder.addJsonTx(tx))
```
nategraf, line 35: Nit: Prefer `for` loops instead of `.forEach`
```suggestion
    for (const tx of jsonTransactions) {
      builder.addJsonTx(tx)
    }
```

packages/cli/src/commands/governance/list.ts
```diff
@@ -19,22 +19,35 @@ export default class List extends BaseCommand {
...
       ID: { get: (p) => valueToString(p.proposalID) },
       stage: {},
     })
+
+    console.log(chalk.red.bold('Expired Proposals:'))
+    const expiredQueue = expiredQueueMap.map((_, idx) => queue[idx].proposalID)
+    const expiredDequeue = expiredDequeueMap.map((_, idx) => dequeue[idx])
+    cli.table(expiredQueue.concat(expiredDequeue), {
+      ID: { get: (id) => valueToString(id) },
+    })
```
nategraf, line 51: If this grows too large we may want to add ` --all` flag. May be worth adding now

nategraf (commented):
Only serious question is whether the changes to the smart contract interfere with the audit.

packages/protocol/contracts/governance/Governance.sol
nategraf: Actually because we already have a followup audit planned before mainnet, we can include this. 👍

jfoutts-celo (approved):

PR #2617, opened 2020-01-31
Vesting contract revamp

### LINKED ISSUE #2521: Option to set withholding % in vesting contracts
Each vesting entry should be able to specify a `witholdingPercentage` which effectively splits the vesting funds for purposes of withdrawal. A separate method `withdrawVestedWithheldFunds` would retrieve these, only executable by the foundation address.

### LINKED ISSUE #2522: Flag to allow vesting contract to be stake For Validators/groups
Some vesting contracts will be able to stake with unvested funds. Add a flag to configure this, and a method to get a validatorSigner that first checking this flag.

lucasege:
### Description

Modifying the "vesting" (now `releasing`) contract to enable all different types of gold grants.

Key changes:
* Adding `liquidityProvision` flag
* Adding `maxDistribution` value to allow finer control over % of gold released.
* Adding `canVote` and `canValidate` to control if contracts can perform operations with un-released gold. `canValidate` also covers attestation.

### Tested

`release_gold.ts` unit test suite.

### Related Issues
Fixes #2521 
Fixes #2522

### Other changes

Migration changes.


lucasege:
Questions for reviewer:
* Can/should I make `ReleaseGoldInstance` proxied? It seems like a good idea in theory, but the logistics of it (creating a proxy in the `releasegoldfactory` code itself) sounds dubious.
* Is `~uint256(0)` a good way to get `MAX_UINT`?
* Should I keep `FIXED1_UINT` in `releasegoldinstance` or go through the changes to make that public from `FixidityLib`?

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
```diff
@@ -0,0 +1,86 @@
...
+    setRegistry(registryAddress);
+  }
+
+  event NewReleaseGoldInstanceCreated(address indexed beneficiary, address atAddress);
+
+  /**
+   * @notice Factory function for creating a new release gold contract instance.
+   * @param releaseCliffTime Duration (in seconds) after `releaseStartTime` of the golds' cliff.
+   * @param numReleasePeriods Number of releasing periods.
+   * @param releaseStartTime The time (in Unix time) at which point releasing starts.
```
asaj: meganit: Consider making this the first argument

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
```diff
@@ -0,0 +1,86 @@
...
+   * @param releasePeriod Duration (in seconds) of each release period.
```
asaj, line 30: remove space

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
```diff
@@ -0,0 +1,86 @@
...
+   * @param amountReleasedPerPeriod The released gold amound per period.
+   * @param revocable Whether the release schedule is revocable or not.
+   * @param beneficiary Address of the beneficiary to whom released tokens are transferred.
+   * @param owner Address capable of revoking, setting the liquidity provision
+   *              and setting the withdrawal amount.
+   * @param subjectToLiquidityProvision If this schedule is subject to a liquidity provision.
+   * @param _canValidate If this schedule's gold can be used for validating.
```
asaj, line 42: nit: Is the _ prefix necessary here and below?

packages/protocol/contracts/common/UsingRegistry.sol
```diff
@@ -52,7 +52,9 @@ contract UsingRegistry is Ownable {
   bytes32 constant SORTED_ORACLES_REGISTRY_ID = keccak256(abi.encodePacked("SortedOracles"));
   bytes32 constant STABLE_TOKEN_REGISTRY_ID = keccak256(abi.encodePacked("StableToken"));
   bytes32 constant VALIDATORS_REGISTRY_ID = keccak256(abi.encodePacked("Validators"));
-  bytes32 constant VESTING_FACTORY_REGISTRY_ID = keccak256(abi.encodePacked("VestingFactory"));
+  bytes32 constant RELEASE_GOLD_FACTORY_REGISTRY_ID = keccak256(
```
asaj: Suggest we remove this from the registry, as multiple factories are likely to be deployed and they're not "core" to the protocol

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
```diff
@@ -0,0 +1,86 @@
...
+    bool _canVote
+  ) external onlyOwner returns (address) {
+    uint256 releaseGoldAmount = numReleasePeriods.mul(amountReleasedPerPeriod);
+    require(
+      getGoldToken().balanceOf(address(this)) >= releaseGoldAmount,
+      "Factory balance is insufficient to create requested release gold contract"
+    );
+
+    require(
+      releases[beneficiary] == address(0),
```
asaj: I'm wondering if we should remove this requirement and instead map to a list of addresses, as some folks will have multiple of the same type of grant.

On the other hand, this check protects us against accidentally deploying too many instances for the same beneficiary. 

Wdyt?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+
+import "../common/UsingRegistry.sol";
+
+contract ReleaseGoldInstance is UsingRegistry, ReentrancyGuard, IReleaseGoldInstance {
+  using SafeMath for uint256;
+  using FixidityLib for FixidityLib.Fraction;
+
+  struct ReleaseSchedule {
+    // Number of release periods.
+    uint256 numReleasePeriods;
```
asaj: meganit: Consider making the order here match the order of the arguments to `createReleaseGoldInstance` and the constructor

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+    releaseSchedule.releaseStartTime = releaseStartTime;
+
+    beneficiary = _beneficiary;
+    revocationInfo.revocable = revocable;
+    releaseOwner = _releaseOwner;
+
+    // Set maxDistribution default to maxUInt, i.e. no set max.
+    maxDistribution = ~uint256(0);
+    liquidityProvisionMet = (subjectToLiquidityProvision) ? false : true;
+    canValidate = _canValidate;
```
asaj, line 217: Are there any instances where a revokable instance can also validate? If not, consider adding a sanity check that revocable != canValidate

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+   */
+  function isRevoked() public view returns (bool) {
+    return revocationInfo.revokeTime > 0;
+  }
+
+  /**
+   * @notice Controls if the liquidity provision has been met.
+   *         If true, allows gold to be withdrawn.
+   */
+  function setLiquidityProvision(bool met) external onlyReleaseOwner {
```
asaj: Do we ever need to set this from met to unmet? If not, we may not need to take an argument here, though I suppose the current method is strictly more general

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+    liquidityProvisionMet = met;
+    emit LiquidityProvisionSet(beneficiary, met);
+  }
+
+  uint256 private constant FIXED1_UINT = 1000000000000000000000000;
```
asaj: I think we can get this from fixidityLib

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+
+  /**
+   * @notice Controls the maximum distribution percentage.
+   *         Calculates `distributionPercentage` of current `totalBalance()`
+   *         and sets this value as the maximum allowed gold to be currently withdrawn.
+   */
+  function setMaxDistribution(uint256 distributionPercentage) external onlyReleaseOwner {
+    if (distributionPercentage == FIXED1_UINT) {
```
asaj: nit: Please add a small comment explaining the reasoning behind this

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+  }
+
+  uint256 private constant FIXED1_UINT = 1000000000000000000000000;
+
+  /**
+   * @notice Controls the maximum distribution percentage.
+   *         Calculates `distributionPercentage` of current `totalBalance()`
+   *         and sets this value as the maximum allowed gold to be currently withdrawn.
+   */
+  function setMaxDistribution(uint256 distributionPercentage) external onlyReleaseOwner {
```
asaj, line 245: Should we require distributionPercentage to be less than or equal to 100%?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+    if (distributionPercentage == FIXED1_UINT) {
+      maxDistribution = ~uint256(0);
+    } else {
+      uint256 totalBalance = getTotalBalance();
+      maxDistribution = FixidityLib.fromFixed(
+        FixidityLib.newFixed(totalBalance).multiply(FixidityLib.wrap(distributionPercentage))
```
asaj: Do we need to worry about overflow or anything here? Worth noting that we probably don't need the precision provided by fixidity if we're worried about overflow

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+  function getRemainingLockedBalance() public view returns (uint256) {
+    return getLockedGold().getAccountTotalLockedGold(address(this));
+  }
+
+  /**
+   * @dev Calculates the total amount that has already released up to now.
+   * @return The already released amount up to the point of call.
+   * @dev The returned amount may vary over time due to locked gold rewards.
+   */
+  function getCurrentReleasedTotalAmount() public view returns (uint256) {
```
asaj, line 359: Should this be the function that checks the liquidity provision, rather than `withdraw`? Or in other words, when an instance is revoked when the liquidity provision has yet to be hit, what should happen?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+   * @param greater The group receiving more votes than `group`, or 0 if `group` has the
+   *                most votes of any validator group.
+   * @dev Fails if `group` is empty or not a validator group.
+   */
+  function vote(address group, uint256 value, address lesser, address greater)
+    external
+    nonReentrant
+    onlyWhenInProperState
+  {
+    require(getElection().vote(group, value, lesser, greater), "Voting for a group failed");
```
asaj: Do the other wrapped functions return booleans? If so, for cleanliness, let's be consistent in requiring them (or not)

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+   * @notice Increments the number of total and pending votes for `group`.
+   * @param group The validator group to vote for.
+   * @param value The amount of gold to use to vote.
+   * @param lesser The group receiving fewer votes than `group`, or 0 if `group` has the
+   *               fewest votes of any validator group.
+   * @param greater The group receiving more votes than `group`, or 0 if `group` has the
+   *                most votes of any validator group.
+   * @dev Fails if `group` is empty or not a validator group.
+   */
+  function vote(address group, uint256 value, address lesser, address greater)
```
asaj: Should we have this wrapper at all? Won't I be able to vote even if `canVote` is false?

packages/protocol/test/governance/release_gold.ts
```diff
@@ -0,0 +1,1708 @@
+import { CeloContractName } from '@celo/protocol/lib/registry-utils'
```
asaj, line 1: Can you rename this to `vesting.ts` just to make the review easier? Then, once it's approved, we can switch to the proper name

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
lucasege: I don't see any space here so i'm not sure what this was meant to refer to. Maybe a formatting error on github?

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
lucasege: Yes if we want to keep the parameter name and the storage variable the same. 
https://ethereum.stackexchange.com/questions/37716/explicitly-refer-to-state-variables-instead-of-local-variables-with-the-same-nam

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
lucasege: I agree with switching to a list of addresses, it adds a bit of operational burden to us but saves users some complexity (managing their beneficiary addresses)

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: I was wondering this as well, which is why I went for the more general option to be safe. I support removing the parameter.

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: Yes but I will have to make it public

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: We can, I'm currently relying on Fixidy's overflow protection to prevent any issues from >100%. Other than that, maxDistribution can only limit how much you can withdraw, it can't add to how much you withdraw.
I'll add <= 100% to be clear

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
```diff
@@ -0,0 +1,86 @@
...
+import "./interfaces/IReleaseGoldFactory.sol";
+
+import "../common/Initializable.sol";
+import "../common/UsingRegistry.sol";
+
+contract ReleaseGoldFactory is Initializable, UsingRegistry, IReleaseGoldFactory {
+  using SafeMath for uint256;
+
+  // Mapping between beneficiary addresses and associated release schedule contracts.
+  mapping(address => address) public releases;
```
m-chrzan: Can/should this be an `address => ReleaseGoldInstance` mapping?

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
m-chrzan: Thought: a simple workaround would be just asking the party receiving multiple grants for multiple addresses.

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
```diff
@@ -0,0 +1,86 @@
...
+      getGoldToken().balanceOf(address(this)) >= releaseGoldAmount,
+      "Factory balance is insufficient to create requested release gold contract"
+    );
+
+    require(
+      releases[beneficiary] == address(0),
+      "Only one release gold contract per beneficiary allowed"
+    );
+
+    address newReleaseGoldInstance = address(
```
m-chrzan: Nit, as per comment above: cast to `address` might be unnecessary.

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+  bool public canVote;
+
+  // Public struct housing params pertaining to releasing gold.
+  ReleaseSchedule public releaseSchedule;
+
+  // Public struct housing params pertaining to revocation.
+  RevocationInfo public revocationInfo;
+
+  event ReleaseScheduleRevoked(
+    address indexed beneficiary,
```
m-chrzan: Do these events need the beneficiary listed, given that a given contract instance corresponds to only one beneficiary?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+    releaseSchedule.releasePeriod = releasePeriod;
+    releaseSchedule.releaseCliff = releaseStartTime.add(releaseCliffTime);
+    releaseSchedule.releaseStartTime = releaseStartTime;
+
+    beneficiary = _beneficiary;
+    revocationInfo.revocable = revocable;
+    releaseOwner = _releaseOwner;
+
+    // Set maxDistribution default to maxUInt, i.e. no set max.
+    maxDistribution = ~uint256(0);
```
m-chrzan: For me personally, this would be more readable as `0xff...ff` explicitly written out.

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
m-chrzan: Can you use `FixidityLib.fixed1`, and wrap whatever you're comparing this value to into a Fixidity fraction before the comparison instead of after?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,579 @@
...
+  uint256 private constant FIXED1_UINT = 1000000000000000000000000;
+
+  /**
+   * @notice Controls the maximum distribution percentage.
+   *         Calculates `distributionPercentage` of current `totalBalance()`
+   *         and sets this value as the maximum allowed gold to be currently withdrawn.
+   */
+  function setMaxDistribution(uint256 distributionPercentage) external onlyReleaseOwner {
+    if (distributionPercentage == FIXED1_UINT) {
+      maxDistribution = ~uint256(0);
```
m-chrzan: Again, personally not a fan of writing MAX_INT this way. Also, since you're using MAX_INT in multiple places, I'd suggest defining a top-level constant for it.

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: This seems like a situation we can't allow (revocable being allowed to validate), so I will add this check

lucasege: What about instances where it isn't revocable and also can't validate? I don't know if these do exist, but I also don't know that they don't.

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: I added the check here, which is correct and something I overlooked.
I'm also keeping the check in `withdraw` since I think it makes sense for withdraw to revert with a message about liquidity provision, rather than reverting for the generic "you have 0 funds available to withdraw".

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: All functions that don't `require` don't return bools. However, all of the bools returned are useless (imo) since they just return true. I'll remove the requires.

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
asaj: Yeah we should allow for that

asaj: Looks like you did :)

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,558 @@
...
+    releaseSchedule.amountReleasedPerPeriod = amountReleasedPerPeriod;
+    releaseSchedule.releasePeriod = releasePeriod;
+    releaseSchedule.releaseCliff = releaseStartTime.add(releaseCliffTime);
+    releaseSchedule.releaseStartTime = releaseStartTime;
+
+    beneficiary = _beneficiary;
+    revocationInfo.revocable = revocable;
+    releaseOwner = _releaseOwner;
+
+    // Set maxDistribution default to maxUInt, i.e. no set max.
```
asaj: I don't think we can do this for the LLC grants

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,558 @@
...
+    require(!liquidityProvisionMet, "Liquidity provision has already been set");
+    liquidityProvisionMet = true;
+    emit LiquidityProvisionSet(beneficiary);
+  }
+
+  /**
+   * @notice Controls the maximum distribution percentage.
+   *         Calculates `distributionPercentage` of current `totalBalance()`
+   *         and sets this value as the maximum allowed gold to be currently withdrawn.
+   * @param distributionPercentage Percentage in range [0, 100] indicating % of total balance
```
asaj: Can we add one more decimal point of precision here?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,558 @@
...
+    uint256 periodsSinceStart = timeSinceStart.div(releaseSchedule.releasePeriod);
+    return totalBalance.mul(periodsSinceStart).div(releaseSchedule.numReleasePeriods);
+  }
+
+  /**
+   * @notice A wrapper function for the lock gold method.
+   * @param value The value of gold to be locked.
+   */
+  function lockGold(uint256 value) external nonReentrant onlyWhenInProperState {
+    require(
```
asaj: nit: this is probably superfluous, since the following line would fail

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,558 @@
...
+  function setAccountMetadataURL(string calldata metadataURL) external onlyWhenInProperState {
+    getAccounts().setMetadataURL(metadataURL);
+  }
+
+  /**
+   * @notice Converts `account`'s pending votes for `group` to active votes.
+   * @param group The validator group to vote for.
+   * @dev Pending votes cannot be activated until an election has been held.
+   */
+  function activate(address group) external nonReentrant onlyWhenInProperState {
```
asaj: we can remove this, since this can be done with the voting signer

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,558 @@
...
+   * @param group The validator group to revoke votes from.
+   * @param value The number of votes to revoke.
+   * @param lesser The group receiving fewer votes than the group for which the vote was revoked,
+   *               or 0 if that group has the fewest votes of any validator group.
+   * @param greater The group receiving more votes than the group for which the vote was revoked,
+   *               or 0 if that group has the most votes of any validator group.
+   * @param index The index of the group in the account's voting list.
+   * @dev Fails if the account has not voted on a validator group.
+   */
+  function revokeActive(
```
asaj, line 543: note that the revoke wrappers need to remain so that the release owner can revoke votes so that gold can be unlocked

asaj: Note that participating in on-chain governance can also block unlocking gold [1]. I *think* things work as-is, since the release owner can authorize a different vote signer, and eventually the governance votes will expire and the gold can be unlocked, but please double check me here.


https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/LockedGold.sol#L150

packages/protocol/test/governance/vesting.ts
```diff
@@ -38,22 +39,32 @@ const authorizationTestDescriptions = {
...
-  vestingPeriodSec: number
-  vestAmountPerPeriod: BigNumber
-  vestingRevocable: boolean
-  vestingRevoker: string
-  vestingMaxPausePeriod: number
+interface IReleaseGoldInstance {
+  releaseGoldStartTime: number
+  releaseGoldCliff: number
+  releaseGoldNumPeriods: number
+  releaseGoldPeriodSec: number
```
asaj: nit: can we rename (and reorder, if necessary) to match the smart contract?

packages/protocol/test/governance/vesting.ts
```diff
@@ -552,298 +548,333 @@ contract('Vesting', (accounts: string[]) => {
...
       assert.equal(fetchedKey, longDataEncryptionKey)
     })
   })
 
-  Object.keys(authorizationTestDescriptions).forEach((key) => {
-    describe('authorization tests:', () => {
-      let authorizationTest: any
-      let vestingInstanceAddress: any
-      let vestingInstance: any
+  describe('#setMaxDistribution', () => {
```
asaj: Can you include a test for the 100% special case?

asaj:
@m-chrzan adding you as a reviewer for another pair of eyes

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: Your logic seems sound, authorizing a new vote signer should overwrite any existing one, allowing you to get control of votes and unlock gold

packages/protocol/test/governance/vesting.ts
lucasege: Done, also added testing of rewards

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
```diff
@@ -0,0 +1,93 @@
...
+import "./interfaces/IReleaseGoldFactory.sol";
+
+import "../common/Initializable.sol";
+import "../common/UsingRegistry.sol";
+
+contract ReleaseGoldFactory is Initializable, UsingRegistry, IReleaseGoldFactory {
+  using SafeMath for uint256;
+
+  // Mapping between beneficiary addresses and associated release schedule contracts.
+  mapping(address => address[]) public releases;
```
m-chrzan: Should/can this be an `address => ReleaseGoldInstance[]` mapping?

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
m-chrzan: Do we have a `canValidate` and `canVote` variable in this contract?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,570 @@
...
+   * @param releaseStartTime The time (in Unix time) at which point releasing starts.
+   * @param releaseCliffTime Duration (in seconds) after `releaseStartTime` of the golds' cliff.
+   * @param numReleasePeriods Number of releasing periods.
+   * @param releasePeriod Duration (in seconds) of each release period.
+   * @param amountReleasedPerPeriod The released gold amount per period.
+   * @param revocable Whether the release schedule is revocable or not.
+   * @param _beneficiary Address of the beneficiary to whom released tokens are transferred.
+   * @param _releaseOwner Address capable of revoking, setting the liquidity provision
+   *                      and setting the withdrawal amount.
+   *                      Null if grant is not subject to these operations.
```
m-chrzan: Nit: maybe s/Null/0x0 to be explicit? Here and below.

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,570 @@
...
+   * @param _refundAddress Address that receives refunded funds if contract is revoked.
+   *                       Null if contract is not revocable.
+   * @param subjectToLiquidityProvision If this schedule is subject to a liquidity provision.
+   * @param initialDistributionPercentage Percentage of total rewards available for distribution.
```
m-chrzan, line 148: Is there a reason this isn't an unwrapped Fixidity fraction?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,570 @@
...
+      _beneficiary != address(0),
+      "The release schedule beneficiary cannot be the zero addresss"
+    );
+    require(registryAddress != address(0), "The registry address cannot be the zero address");
+    require(
+      releaseSchedule.releaseStartTime.add(
+        releaseSchedule.numReleasePeriods.mul(releaseSchedule.releasePeriod)
+      ) >
+        block.timestamp,
+      "Release schedule end time must be in the future"
```
m-chrzan, line 190: Just wondering, why is this necessary?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,570 @@
...
+    releaseOwner = _releaseOwner;
+    refundAddress = _refundAddress;
+
+    if (initialDistributionPercentage < 1000) {
+      // Cannot use `getTotalBalance()` here because the factory has not yet sent the gold.
+      uint256 totalGrant = releaseSchedule.amountReleasedPerPeriod.mul(
+        releaseSchedule.numReleasePeriods
+      );
+      // Initial percentage is expressed to 3 significant figures: [0, 1000].
+      maxDistribution = totalGrant.mul(initialDistributionPercentage).div(1000);
```
m-chrzan, line 211: Will this freeze a portion of the distribution due to rounding errors?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,570 @@
...
+    uint256 timeSinceStart = block.timestamp.sub(releaseSchedule.releaseStartTime);
+    uint256 periodsSinceStart = timeSinceStart.div(releaseSchedule.releasePeriod);
+    return totalBalance.mul(periodsSinceStart).div(releaseSchedule.numReleasePeriods);
+  }
+
+  /**
+   * @notice A wrapper function for the lock gold method.
+   * @param value The value of gold to be locked.
+   */
+  function lockGold(uint256 value) external nonReentrant onlyWhenInProperState {
```
m-chrzan: Do we want to allow the release owner to lock gold in a revoked instance?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -0,0 +1,570 @@
...
+  function unlockGold(uint256 value) external nonReentrant onlyWhenInProperState {
+    getLockedGold().unlock(value);
+  }
+
+  /**
+   * @notice A wrapper function for the relock locked gold method function.
+   * @param index The index of the pending locked gold withdrawal.
+   * @param value The value of gold to be relocked for the release schedule instance.
+   */
+  function relockGold(uint256 index, uint256 value) external nonReentrant onlyWhenInProperState {
```
m-chrzan: Ditto here and other wrappers other than `unlock` and `withdrawLockedGold`.

packages/protocol/test/governance/vesting.ts
```diff
@@ -38,22 +40,34 @@ const authorizationTestDescriptions = {
...
-  vestingBeneficiary: string
-  vestingNumPeriods: number
-  vestingCliff: number
-  vestingStartTime: number
-  vestingPeriodSec: number
-  vestAmountPerPeriod: BigNumber
-  vestingRevocable: boolean
-  vestingRevoker: string
-  vestingMaxPausePeriod: number
+interface IReleaseGoldInstance {
```
m-chrzan: Naming nit: I think I would prefer a different name for this type, to not confuse with the Solidity interface's name. Maybe `ReleaseGoldInstanceConfig`?

packages/protocol/test/governance/vesting.ts
```diff
@@ -71,48 +85,59 @@ const HOUR = 60 * 60
...
-      vestingSchedule.vestAmountPerPeriod,
-      vestingSchedule.vestingRevocable,
-      vestingSchedule.vestingRevoker,
-      vestingSchedule.vestingMaxPausePeriod,
+  const createNewReleaseGoldInstanceTx = async (
+    releaseGoldSchedule: IReleaseGoldInstance,
+    web3: Web3
+  ) => {
+    releaseGoldSchedule.releaseStartTime = (await getCurrentBlockchainTimestamp(web3)) + 5 * MINUTE
+    const ReleaseGoldInstanceTx = await releaseGoldFactoryInstance.createReleaseGoldInstance(
```
m-chrzan: Style nit: value names should start with a lowercase letter. (although here you could even just directly return)

packages/protocol/test/governance/vesting.ts
```diff
@@ -71,48 +85,59 @@ const HOUR = 60 * 60
...
-      vestingSchedule.vestingBeneficiary,
-      vestingSchedule.vestingNumPeriods,
-      vestingSchedule.vestingCliff,
-      vestingSchedule.vestingStartTime,
-      vestingSchedule.vestingPeriodSec,
-      vestingSchedule.vestAmountPerPeriod,
-      vestingSchedule.vestingRevocable,
-      vestingSchedule.vestingRevoker,
-      vestingSchedule.vestingMaxPausePeriod,
+  const createNewReleaseGoldInstanceTx = async (
```
m-chrzan: Does this function create a transaction object? I'm confused by the `Tx` suffix in this name - seems like it just creates a new release instance.

packages/protocol/test/governance/vesting.ts
```diff
@@ -132,418 +157,392 @@ contract('Vesting', (accounts: string[]) => {
...
+        web3
+      )
+      releaseGoldInstanceAddress = await releaseGoldFactoryInstance.releases(beneficiary, 0)
+      releaseGoldInstance = await ReleaseGoldInstance.at(releaseGoldInstanceAddress)
     })
 
-    it('should fail if a non-owner attempts to create a vesting instance', async () => {
-      const vestingSchedule = _.clone(vestingDefaultSchedule)
-      vestingSchedule.vestingStartTime = (await getCurrentBlockchainTimestamp(web3)) + 5 * MINUTE
+    it('should fail if a non-owner attempts to create a releaseGold instance', async () => {
```
m-chrzan: This `it` doesn't depend on the `beforeEach` above. Consider putting the `beforeEach` and `it`s that use the release instance created there in a `describe('when an instance is properly created')` or similar.

packages/protocol/test/governance/vesting.ts
```diff
@@ -132,418 +157,392 @@ contract('Vesting', (accounts: string[]) => {
...
-      assert.equal(newVestingInstanceAddress, vestingInstance.address)
+      assert.equal(newReleaseGoldInstanceAddress, releaseGoldInstance.address)
     })
 
-    it('should revert when vesting factory has insufficient balance to create new instance', async () => {
-      const vestingSchedule = _.clone(vestingDefaultSchedule)
-      vestingSchedule.vestingNumPeriods = 3
-      vestingSchedule.vestAmountPerPeriod = ONE_GOLDTOKEN
-      await assertRevert(createNewVestingInstanceTx(vestingSchedule, web3))
+    it('should revert when releaseGold factory has insufficient balance to create new instance', async () => {
```
m-chrzan: Ditto.

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
```diff
@@ -46,28 +51,35 @@ contract ReleaseGoldFactory is Initializable, UsingRegistry, IReleaseGoldFactory
...
-    address payable owner,
+    address releaseOwner,
+    address payable refundAddress,
     bool subjectToLiquidityProvision,
+    uint256 initialDistributionPercentage,
     bool _canValidate,
     bool _canVote
   ) external onlyOwner returns (address) {
     uint256 releaseGoldAmount = numReleasePeriods.mul(amountReleasedPerPeriod);
+    ReleaseGoldInstance.ReleaseSchedule storage rs = ReleaseGoldInstance.ReleaseSchedule(
```
asaj: Why is this a storage pointer?

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -145,46 +153,63 @@ contract ReleaseGoldInstance is UsingRegistry, ReentrancyGuard, IReleaseGoldInst
...
-      releaseStartTime.add(numReleasePeriods.mul(releasePeriod)) > block.timestamp,
+      releaseSchedule.releaseStartTime.add(
+        releaseSchedule.numReleasePeriods.mul(releaseSchedule.releasePeriod)
+      ) >
+        block.timestamp,
       "Release schedule end time must be in the future"
     );
     require(!(revocable && _canValidate), "Revocable contracts cannot validate");
+    require(
+      initialDistributionPercentage >= 0 && initialDistributionPercentage <= 1000,
```
asaj: nit: This first check (>=0) isn't necessary as it's a uint

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -211,20 +236,20 @@ contract ReleaseGoldInstance is UsingRegistry, ReentrancyGuard, IReleaseGoldInst
...
    *         and sets this value as the maximum allowed gold to be currently withdrawn.
-   * @param distributionPercentage Percentage in range [0, 100] indicating % of total balance
-   *                               available for distribution.
+   * @param distributionPercentage Percentage in range [0, 1000] (3 significant figures)
+   *                               indicating % of total balance available for distribution.
    */
   function setMaxDistribution(uint256 distributionPercentage) external onlyReleaseOwner {
     require(
-      distributionPercentage >= 0 && distributionPercentage <= 100,
+      distributionPercentage >= 0 && distributionPercentage <= 1000,
```
asaj: nit: first condition also not necessary since it's a uint

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -275,10 +300,10 @@ contract ReleaseGoldInstance is UsingRegistry, ReentrancyGuard, IReleaseGoldInst
     );
     uint256 revokerAmount = getRemainingUnlockedBalance();
     require(
-      getGoldToken().transfer(releaseOwner, revokerAmount),
+      getGoldToken().transfer(refundAddress, revokerAmount),
       "Transfer of gold to releaseOwner failed"
```
asaj: nit: Change string here to refundAddress

asaj (approved):
🚢

packages/protocol/contracts/governance/ReleaseGoldFactory.sol
lucasege: Artifact of previous work, removed.

m-chrzan (commented):
Submitting some comments. Nothing big, so feel free to continue with merge and potentially do a small follow-up PR to clean up anything you feel is worth cleaning up.

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: https://github.com/celo-org/celo-monorepo/pull/2617#discussion_r375561722

The precision was determined not to be worth the increased overflow risk working with 24 digit numbers

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: This is left from Limechain's original take on vesting, but I left it because it seems like a decent safeguard against mistakes in provided parameters. Hitting this check would be equivalent to just transferring the granted amount to the beneficiary, with the only downside in transfer being the lack of clear record. 
I don't feel strongly either way, but have a slight bias towards making operator mistakes harder, i.e. making it less probable to provide the wrong start date

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: Not sure what you mean by freeze: if you mean that we could set distribution percentage to a precision that gets rounded away, leaving less gold to be withdrawn than strictly detailed, then yes, but the response would be 1) we don't allow that much precision, so its unlikely and 2) its not a big problem, since we can just take that into account when we set the withdrawn percentages and adjust if necessary to over/under compensate

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
lucasege: Good point, probably not.

PR #2618, opened 2020-02-02
Freeze StableToken at contracts deployment

### LINKED ISSUE #2560: StableToken transfers should be freezable/unfreezable
### Expected Behavior

StableToken transfers are frozen at mainnet launch, and can be unfrozen via governance proposal.

This allows the network to launch before Celo Gold is integrated into exchanges, after which StableToken transfers can be turned on via governance proposal.

### Current Behavior

StableToken transfers on by default

m-chrzan:
### Description

Before Celo assets are established on exchanges, cUSD transfers should be frozen since the stability mechanism depends on an oraclized cGLD price.

The ability to unfreeze is given to Governance during migrations.

Once unfrozen, `freeze` cannot be called again.

Added `stableToken.unfreezeImmediately` option to the migrations configuration to be able to run a dev/testing environment with unchanged behavior.

TODO: 
* Probably need to fix integration transfer test expected gas amounts.

### Tested

Unit tests still pass.

### Related issues

- Fixes #2560 

### Backwards compatibility

This affects fresh network deployments.


packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -109,6 +111,11 @@ contract StableToken is
     require(inflationRate != 0, "Must provide a non-zero inflation rate.");
 
     _transferOwnership(msg.sender);
+
+    // At network launch, stable token transfers should be frozen.
+    frozen = true;
+    setFreezer(msg.sender);
```
mcortesi: mmm... not sure about this.

I think this is `validator0`... but we then transfer ownership to the governance contract. Probably freezer should also be the owner... in fact it's probably not necessary to have a duplicated field for taht

m-chrzan: I'm reusing the `Freezable` contract we already had defined for other functionalities, where we did want the owner and freezer to be different parties. 

At contract deployment/proxy initialization, all control is given to validator0, which is then transferred to Governance in one of the last migrations. This PR adds code to also transfer `freezer` to Governance at that point.

mcortesi: I think it's more "clean" if we keep owner and freezer as the same. So instead of making the migration responsible of doing that, modify `setOwner()` to also call `setFreezer()`.

That way; the semantic rules are internal to the contract, and not dependent of the migration code

what do you think?

mcortesi (approved):
LGTM!

Issue #1154, opened 2019-10-01
Automate iOS build with CircleCI


jeanregisser:
### Expected Behavior

iOS build / tests should be done on CircleCI

### Current Behavior

iOS build / tests are not done on CircleCI

ole-nov:
closed because is the duplicate of #6103

PR #2622, opened 2020-02-03
[Wallet] Run E2E test on iOS

jeanregisser:
### Description

This PR enables the wallet E2E test on iOS and runs it on CircleCI.

Main highlights:
- Spent a lot of time making sure the test runs reliably. Let's hope it will pass the test of time 🤞 🍀 
- Made detox initialization failures easier to spot and troubleshoot. Previously it would still try to run the test suite after an init failure, leading to a lot of confusing errors. Detox init now blocks further progress if it fails, example: Screenshot 2020-02-11 at 14 02 25
- `victory` charts package was preventing the E2E test to run because it's using `d3-timer` under the hood which itself uses `requestAnimationFrame` continuously. Detox itself monitors all timers shorter than 1.5 secs and doesn't mark itself as ready while those are still running. I had to monkey patch it to workaround this during the E2E test
- We're now preloading the app bundle before starting the E2E test. Without this, the app was randomly failing to start fully with the red screen `Could not connect to development server`, though I could see the bundle building reach 100% in the packager logs when this happened.
- Improved `yarn install` time on CircleCI.
  - Stats for the `install_dependencies` step:
    - previously it took ~9mins on CircleCI (with or without an existing cache)
    - now with an empty cache it still takes ~9 mins, but with a populate cached it's now around 5 mins.
- Reported the issue of `victory-native` including `victory-core` twice: https://github.com/FormidableLabs/victory-native/issues/535
- CocoaPods now uses the CDN for faster installs. Especially true on CircleCI as the huge CocoaPods specs git repo doesn't need to be cloned anymore. See http://blog.cocoapods.org/CocoaPods-1.7.2/
- Fail builds if `Podfile.lock` or `Gemfile.lock` have differences after the install step. This is to help developers remember to run the install commands when touching the dependencies. As is already the case for yarn.
- Skip uploading iOS debug symbols to Sentry for debug builds

Future work:
- restore running E2E test on Android on CircleCI
- update skipped or outdated E2E tests (right now it doesn't do much, launches the app and goes through the first few screens only. i.e. no actual transactions)

### Tested

iOS E2E test runs successfully on CircleCI.

### Other changes

- Improved `stringToBoolean` util to trim whitespace.
- Removed now unneeded `npm rebuild scrypt` step in CircleCI

### Related issues

- Epic https://github.com/celo-org/celo-monorepo/issues/2641
- #1659
- #1154

### Backwards compatibility

Yes.


packages/mobile/e2e/Transfer.spec.js
```diff
@@ -56,7 +56,13 @@ describe('Transfer Works', () => {
   it('NUX->Pin', async () => {
     await expect(element(by.id('SystemAuthTitle'))).toBeVisible()
     await expect(element(by.id('SystemAuthContinue'))).toBeVisible()
-    enterPin()
+
+    if (device.getPlatform() === 'android') {
+      enterPin()
```
jmrossy: Android should use the custom keypad too. I removed the scripts bits that unlocked the android emulator to improve the reliability

packages/mobile/src/setupE2eEnv.e2e.ts
```diff
@@ -0,0 +1,26 @@
+import Config from 'react-native-config'
+// @ts-ignore
+import { TimerContext } from 'victory-core'
+// @ts-ignore
+import { TimerContext as TimerContextES } from 'victory-core/es'
+
+// Inject custom e2e config
+Config.IS_E2E = true
```
jmrossy, line 10: Why inject this in here instead of setting `IS_E2E=true` in ` .env.test`? I thought I added it there already, I think that was my mistake

packages/mobile/src/setupE2eEnv.e2e.ts
```diff
@@ -0,0 +1,26 @@
+import Config from 'react-native-config'
```
jmrossy, line 1: What causes this `.e2e.ts` file to be imported instead of the other ts file? Is that a Detox thing?

jmrossy (commented):
Amazing work! Just a few small questions below

i1skn (approved):
Thanks for the great effort on making this work! LGTM!

packages/mobile/src/setupE2eEnv.e2e.ts
jeanregisser: It's activated in `metro.config.js` with the env var `CELO_TEST_CONFIG=e2e`.

packages/mobile/src/setupE2eEnv.e2e.ts
jeanregisser: There was a script adding this variable to the env file, but it was removed recently and IS_E2E wasn't set anywhere anymore.
I thought about adding it back to the env file, but then I found that it was more flexible to have it injected this way as it allows using any existing env file without modifying it.

packages/mobile/e2e/Transfer.spec.js
jeanregisser: Thanks will update this! I wasn't able to run the Android test as mentioned on Slack.

jmrossy (approved):

PR #2623, opened 2020-02-03
Correct epoch definitions in protocol and contractkit

### LINKED ISSUE #2718: Epoch number definition mismatch
### Expected Behavior

Epoch number is defined the same across Celo codebase.  CLabs Validators are slashable.

### Current Behavior

There are conflicting definitions of epoch number that are off-by-one.  This culminates in the CLabs Validators on Baklava having an empty `membershipHistory` despite being members of Validator Group, and this prevents the CLabs Validators from being slashable.

jfoutts-celo:
### Description

Corrects epoch calculations in protocol and contractkit so they match `celo-blockchain/blob/master/consensus/istanbul`.

### Tested

Tested slashing with corrected epoch definitions on testnet.

### Related issues

Fixes #2718

### Backwards compatibility

Anywhere absolute epoch numbers are presented or stored is affected.  I think `MembershipHistoryEntry.epochNumber` in the Validators smart contract is affected most, however I think there's almost no impact.


jfoutts-celo:
Currently the changes fix issue #2718, however they seem to increase the duration of protocol tests from 30 minutes to 60 minutes.

packages/protocol/contracts/governance/DowntimeSlasher.sol
```diff
@@ -49,7 +49,13 @@ contract DowntimeSlasher is SlasherUtil {
     pure
     returns (uint256)
   {
-    return blockNumber.add(epochSize).sub(1) / epochSize;
+    // Follows GetEpochNumber from celo-blockchain/blob/master/consensus/istanbul/utils.go
```
asaj: Use `getEpochNumberOfBlock` from `UsingPrecompiles`

asaj (approved):

PR #2627, opened 2020-02-04
Multiproxy infrastructure [do not merge]

### LINKED ISSUE #2597: Multiproxy infrastructure
### Expected Behavior

A developer should be able to deploy a testnet with multiple proxies per validator on K8s & VMs

### Current Behavior

Only single proxy per validator supported

tkporter:
### Description

Multiproxy infrastructure

## Design

A new variable in .env files called `VALIDATOR_PROXY_COUNTS` has replaced `PROXIED_VALIDATORS`. This variable allows specifying varying numbers of proxies per validator with the format `<# of validators>:<proxy count>,<# of validators>:<proxy count>;…`. For example, to have 2 validator with 5 proxies and 3 validators with 4 proxies, it would be: `VALIDATOR_PROXY_COUNTS=2:5,3:4`

Celotool expands VALIDATOR_PROXY_COUNTS into an array where each index corresponds to a validator, and each element is the number of proxies for the corresponding validator. Continuing the previous example, the array would be: `[5, 5, 4, 4, 4]`

## K8s Setup

If `n` validators are to be proxied, the first `n` validators of the validator statefulset are still proxied. Each proxied validator has a corresponding statefulset for its proxies (rather than the previous setup where there was 1 statefulset and each replica was associated with the corresponding validator).

## Terraform VM Setup

Terraform allows you to create multiple replicas of a resource in two ways: using `count`, which will create the specified number of replicas and allow specific resources to be indexed by their `count.index` (in the range of [0, count - 1]). The other way is using `for_each`, which allows a developer to specify a list or map and will create a resource for each corresponding list element or map key. I changed the full node Terraform module to use `for_each` to allow us to avoid having to map index ranges to validators, which would also make scaling up proxies harder.

Specifying proxies for the `vm-exec` and `ssh-vm-node` celotool commands is done by providing `<validator index>:<proxy index>` for the index parameter

### Tested

Ran public, private k8s testnets. Ran terraform testnets

### Other changes

- Removed .env.integrationtesting, which is no longer used
- Removed unused helm variables
- Removed some unused env vars
- Created a new env var for specifying the disk size in GB for nodes
- Made max_peers configurable for VM tx nodes / proxies
- Passed some arrays to helm as actual arrays rather than joined strings or variable names with numbers in them

### Related issues

- Fixes #2597 

Blocked by the multiproxy PR: https://github.com/celo-org/celo-blockchain/pull/821

### Backwards compatibility

Not backwards compatible. One flag to geth has been changed, and proxy keys are generated differently


tkporter:
Closing in favor of the latest and greatest version #4150

PR #2634, opened 2020-02-04
Some changes to opsec script

jcortejoso:
Added:
- Do not download files if not needed (probably need to check here if there is internet connection and if not do not run speedtest).
- Check distribution and show only the files for your distribution.
- Only run script if root/sudo (if you are not root the script fails)
- Silent&fail in curl


barbaraliau:
Is this PR still relevant? Wants to be merged into #2547.

jcortejoso:
Closing for now. It would need some update & maintenance if we want to keep them.

PR #2648, opened 2020-02-05
[Wallet] Lock the whole wallet with PIN

### LINKED ISSUE #2574: Wallet Security/PIN improvements
#### Prompt on App Open
Default Behavior: off

- on opening the app (after setting a pin), the user should be prompted for their pin before being able to see home screen

- add an item 'security' to the settings page with a toggle (similar to 'Analytics' section)
"Require PIN on App Open"

#### Add to Safeguards
Additionally: Add a prompt for pin before viewing either safeguards or the backup key (this is not managed by the toggle)

### Current Behavior

no prompt for pin on app open or safeguards

i1skn:
### Description

This PR adds ability to lock your wallet, when you close/background the app. I've added item `Security` in the Settings, so user can enable/disable this feature (only when user has used Custom Pin - see **Caveats** for more details). If the feature is enabled - PIN code screen would appear each time when user has closed/backgrounded the app (if PIN is not cached).

#### Caveats
* **[Android, iOS]** I've discovered some difficulties on implementing this feature, when user has chosen `System Android PIN` (The PIN view itself is outside of RN and it's hard to change it's behaviour), hence `Security` item in Settings in this case would be DISABLED for now.
* **[Android, iOS]** I do not know how to check PIN correctness, when we are using Forno, hence `Security` item in Settings in this case would be DISABLED for now.
*  **[Android]** To hide app content when it's backgrounded I've used `WindowManager.LayoutParams.FLAG_SECURE` which does not support changing color of the overlay (it's hardcoded to white) - https://stackoverflow.com/a/33315627. So on Android I've change the implementation slightly so it's smooth with white color, instead of celo green (like in iOS).

### Tested

Ran on Android & iOS device

### Other changes

* PIN code screen would NOT navigate back if wrong PIN was entered. Instead it would just notify user and clear the previous PIN.
* Move App state monitoring from `geth/saga.ts` saga to `app/saga.ts` (I did not see it's doing anything but logging)

### Related issues

- Fixes #2574

### Backwards compatibility

Yes


packages/mobile/src/account/saga.ts
```diff
@@ -65,8 +65,10 @@ export function* getPincode(useCache = true) {
       }
     }
 
-    const pincodeEntered = new Promise((resolve, reject) => {
-      navigate(Screens.PincodeConfirmation, { resolve, reject })
```
i1skn, line 69: Because `PincodeConfirmation` will try again and again to ask for correct PIN there is no need for `reject`

packages/mobile/src/app/Background.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return <View style={styles.container} />
+}
+
+Background.navigationOptions = { header: null }
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    width: '100%',
+    backgroundColor: Platform.OS === 'android' ? colors.white : colors.celoGreen,
```
i1skn, line 63: See **Caveats** in the description to see why is so

packages/mobile/src/account/Security.tsx
```diff
@@ -0,0 +1,45 @@
...
+})
+
+const style = StyleSheet.create({
+  scrollView: {
+    flex: 1,
+    backgroundColor: colors.background,
+  },
+})
+
+export default withTranslation(Namespaces.accountScreen10)(Security)
```
jmrossy: Would be cleaner to just use the `useTranslation` hook instead of the HOC for this functional component

packages/mobile/src/account/saga.ts
jmrossy: How many times will it try? There should be some max

packages/mobile/src/app/saga.ts
```diff
@@ -112,20 +113,31 @@ export function* navigateToProperScreen() {
     return
   }
 
+  const appLockedAwareNavigate =
```
jmrossy: Should we put this in NavigatorService?

packages/mobile/src/geth/saga.ts
```diff
@@ -186,37 +186,9 @@ function* monitorGeth() {
...
-        appStateChannel.close()
-      }
-    }
-  }
-}
-
 export function* gethSaga() {
   yield call(initGethSaga)
-  const gethRelatedSagas = yield all([fork(monitorAppState), fork(monitorGeth)])
+  const gethRelatedSagas = yield all([fork(monitorGeth)])
```
jmrossy: no need for the `all` here now

packages/mobile/src/app/saga.ts
```diff
@@ -144,39 +156,75 @@ export function* handleDeepLink(action: OpenDeepLink) {
...
+function* watchAppState() {
+  Logger.debug(`${TAG}@monitorAppState`, 'Starting monitor app state saga')
+  const appStateChannel = yield createAppStateChannel()
+  while (true) {
+    try {
+      const newState = yield take(appStateChannel)
+      Logger.debug(`${TAG}@monitorAppState`, `App changed state: ${newState}`)
+      yield put(setAppState(newState))
+    } catch (error) {
+      Logger.error(`${TAG}@monitorAppState`, error)
```
jmrossy: the error should be the 3rd param when calling Logger.error (see it's interface)

packages/mobile/src/app/saga.ts
```diff
@@ -144,39 +156,75 @@ export function* handleDeepLink(action: OpenDeepLink) {
...
+}
+
 export function* appSaga() {
   yield spawn(navigateToProperScreen)
   yield spawn(toggleToProperSyncMode)
-  yield spawn(checkAppDeprecation)
   yield spawn(watchNavigatePinProtected)
   yield spawn(watchDeepLinks)
+  yield spawn(watchAppState)
+  yield takeLatest(Actions.SET_APP_STATE, handleSetAppState)
```
jmrossy, line 225: instead of watching for SET_APP_STATE couldn't we just do the handleSetAppState in watchAppState where we're already taking from the event channel?

packages/mobile/src/navigator/NavigatorWrapper.tsx
```diff
@@ -54,15 +60,17 @@ export class NavigatorWrapper extends React.Component<Props> {
...
         <AppContainer
           ref={this.setNavigator}
           onNavigationStateChange={navigationStateChange}
           {...getPersistenceFunctions()}
         />
+
         <View style={styles.floating}>
-          <BackupPrompt />
+          {!appLocked && <BackupPrompt />}
           <AlertBanner />
```
jmrossy, line 80: Should alert banner be conditional too?

packages/mobile/src/navigator/NavigatorWrapper.tsx
```diff
@@ -99,6 +107,12 @@ export const headerArea = {
   },
 }
 
-export default connect<null, DispatchProps>(null, {
+const mapStateToProps = (state: RootState) => {
```
jmrossy: Nit: Move to top like most other files

jmrossy (commented):
Mostly looks good! Just some small comments below

packages/mobile/src/account/saga.ts
i1skn: Yes, you are right. We decided to tackle it in the follow-up issue : https://github.com/celo-org/celo-monorepo/issues/2649

packages/mobile/src/account/Security.tsx
i1skn: Well spotted!

packages/mobile/src/navigator/NavigatorWrapper.tsx
i1skn: We need `<AlertBanner />` to show user `Incorrect PIN` if that's the case.

packages/mobile/src/app/saga.ts
i1skn: I think of it like this: one place to emit `Actions.SET_APP_STATE` and other places are subscribed to this event (reducer, saga(s)). This way other modules, like `geth`, could just subscribe to `Actions.SET_APP_STATE` too if they need it.  In this implementation `watchAppState` is rather a propagator than a handler, if that makes sense.

jmrossy (approved):

PR #2651, opened 2020-02-05
[WalletKit] Fix using incorrect gcs bucket for downloading production artifacts

jeanregisser:
### Description

Walletkit build-sdk script is downloading artifacts from the incorrect bucket when using production environments. This PR fixes it.

### Tested

It now downloads the artifacts from the `contract_artifacts_production` bucket when running `yarn --cwd packages/walletkit build:for-env alfajores`.

### Other changes

None

### Related issues

Discussed on Slack https://celo-org.slack.com/archives/CL7BVQPHB/p1580865465023000?thread_ts=1580864490.021300&cid=CL7BVQPHB

### Backwards compatibility

Yes.

### DIFF
--- a/packages/walletkit/bin/build-sdk.js
+++ b/packages/walletkit/bin/build-sdk.js
@@ -17,7 +17,7 @@ function getEnvFile(celoEnv) {
   if (fs.existsSync(filePath)) {
     return filePath
   } else {
-    throw new Error(`Unable to find env file for ''${celoEnv}`)
+    throw new Error(`Unable to find env file for '${celoEnv}'`)
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/walletkit/bin/build-sdk.js
```diff
@@ -4,31 +4,44 @@ const execSync = require('child_process').execSync
...
+    return filePath
+  } else {
+    throw new Error(`Unable to find env file for '${celoEnv}'`)
+  }
+}
+
 function isProduction(env) {
-  return env.endsWith('production')
+  const envConfig = config({ path: getEnvFile(env) })
+  return envConfig.parsed['TESTNET_PROJECT_NAME'] === 'celo-testnet-production'
```
jmrossy, line 26: Did you mean to add a config line of TESTNET_PROJECT_NAME to all the mobile .env files? Should that be in this PR?

jeanregisser: It uses the env files at the root of the monorepo.

jmrossy: Oh or does this access the env files at the monorepo root?

jmrossy (approved):

Issue #2221, opened 2019-12-12
[Proposal] Move e2e tests out of celotool into its own folder


H34D:
### Expected Behavior

Move e2e tests out of celotool into its own folder

### Current Behavior

e2e tests are in the celotool codebase

- it clutters up the code base of celotool
- it lacks a proper separation of concerns (which code belongs to what?)
- Potential for creating more generic implementations of all the task by making the code available in e.g. `dev-utils`

PR #2678, opened 2020-02-05
Add migrate functionality when using geth start command on celotool

H34D:
### Description

You now can migrate contracts when running a local testnet

### Related issues

- #1691 #1807


packages/celotool/src/lib/geth.ts
```diff
@@ -8,7 +8,7 @@ import { unlockAccount } from '@celo/walletkit'
 import BigNumber from 'bignumber.js'
 import { spawn } from 'child_process'
 import fs from 'fs'
-import { range } from 'lodash'
+import _, { range } from 'lodash'
```
nategraf: Only import merge instead of `_`.
```suggestion
import { range, merge } from 'lodash'
```

packages/celotool/src/lib/geth.ts
```diff
@@ -1141,3 +1141,64 @@ export async function connectValidatorPeers(instances: GethInstanceConfig[]) {
     instances.filter(({ wsport, rpcport, validating }) => validating && (wsport || rpcport))
   )
 }
+
```
nategraf, line 1150: You might consider putting this in [migration-utils.ts](https://github.com/celo-org/celo-monorepo/blob/master/packages/celotool/src/lib/migration-utils.ts), but I'll leave that up top you.

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -103,6 +106,21 @@ export const builder = (argv: yargs.Argv) => {
...
+      implies: 'monorepo-dir',
+    })
+    .option('migrateTo', {
+      type: 'number',
+      description: 'Migrate contracts to level x',
+      implies: 'monorepo-dir',
+    })
+    .option('monorepo-dir', {
+      type: 'string',
+      description: 'Directory of the mono repo',
```
nategraf, line 122: Useful to have as an option. I imagine most users will want to use the current monorepo directory, so making that the default would be good.

packages/celotool/src/lib/geth.ts
```diff
@@ -1147,3 +1147,64 @@ export async function connectValidatorPeers(instances: GethInstanceConfig[]) {
...
+export async function migrateContracts(
+  monorepoRoot: string,
+  validatorPrivateKeys: string[],
+  attestationKeys: string[],
+  validators: string[],
+  to: number = 1000,
+  overrides: any = {},
+  verbose: boolean = true
+) {
+  const migrationOverrides = merge(
```
mcortesi, line 1160: this configuration was devised for e2e test... are we sure it's the overrides we want to launch any geth?

packages/celotool/src/cmds/geth/start.ts
```diff
@@ -103,6 +106,21 @@ export const builder = (argv: yargs.Argv) => {
       description: 'Block Time',
       default: 1,
     })
+    .option('migrate', {
```
mcortesi, line 109: What's the use case here? I mean, launching a celo testnet probably means launching more than one geth instance... but you should run migration only once...
And you should follow the steps:
 - launch all geth validator nodes
 - run migrations

So, your proposition is for us to run all geth minus the last, and then run migrations?

mcortesi (commented):
left some comments

packages/celotool/src/cmds/geth/start.ts
H34D: > And you should follow the steps:
> 
> * launch all geth validator nodes
> * run migrations

sorry maybe i don't understand the question but this command will run a given number of geth nodes. The `migrate` switch will either migrate to the newly created network or not.

packages/celotool/src/lib/geth.ts
H34D: well, running a local network is becoming a functionality of the product celotool with this issue: https://github.com/celo-org/celo-monorepo/issues/1807 therefore it makes sense to enable this functionality even outside the e2e tests.

Also i would like to point out this issue: [Proposal] Move e2e tests out of celotool into its own folder https://github.com/celo-org/celo-monorepo/issues/2221 to have a better separation of concerns between celotool and the e2e tests.

packages/celotool/src/cmds/geth/start.ts
H34D: it is only relevant when using the migrate option so it will be a special case. But i can of course set a default.

packages/celotool/src/lib/geth.ts
```diff
@@ -1147,3 +1147,64 @@ export async function connectValidatorPeers(instances: GethInstanceConfig[]) {
...
+    {
+      downtimeSlasher: {
+        slashableDowntime: 6,
```
m-chrzan, line 1163: For better generality, this could depend on the epoch length - this number needs to be <= epoch length, otherwise migrations will fail.

packages/celotool/src/cmds/geth/start.ts
mcortesi: No worries... i though this command only ran a single geth, but it runs the whole cluster. So makes sense

packages/celotool/src/lib/geth.ts
mcortesi: Ok! My concern was that if somehow we want to have a different config for e2e test, now this is shared with the "local geth", but we can split that later

mcortesi (approved):

PR #2685, opened 2020-02-06
Adding lastEpochBlock to random.sol

### LINKED ISSUE #2625: Random smart contract should store the randomness of the last epoch's block.
### Expected Behavior

With https://github.com/celo-org/celo-blockchain/pull/844, we'll rely on having the last epoch's block randomness to be available in the smart contract.

### Current Behavior

We currently only keep a window of blocks' randomness.

lucasege:
### Description

Adding an index tracker of the last block of a given epoch for the randomness contract. This ensures that when we call `getBlockRandomness` with the last block of the epoch, it always gives us the randomness, protecting us in case the epoch size is greater than the randomness history window.

Effectively adds another slot into the history reserved for `lastEpochBlock`

### Tested

CI test suite.

### Other changes

Added unit tests

### Related issues

- Fixes #2625

### Backwards compatibility
Yes


nambrot (approved):
Looks good to me though maybe let's run this by @asaj if adding this as a boolean parameter is the way to go

packages/protocol/contracts/identity/Random.sol
```diff
@@ -48,23 +49,35 @@ contract Random is IRandom, Ownable, Initializable {
...
    * should be a special transaction to address 0x0 with 64 bytes of data - the concatenated
    * `randomness` and `newCommitment`. Before running regular transactions, this function should be
    * called.
    */
-  function revealAndCommit(bytes32 randomness, bytes32 newCommitment, address proposer) external {
+  function revealAndCommit(
+    bytes32 randomness,
+    bytes32 newCommitment,
+    address proposer,
+    bool isLastEpochBlock
```
asaj: Rather than have this passed as a boolean can we determine this from `block.number`? UsingPrecompiles should have some functions that can help

packages/protocol/contracts/identity/Random.sol
```diff
@@ -88,25 +105,31 @@ contract Random is IRandom, Ownable, Initializable {
...
+      delete history[lastEpochBlock];
+      lastEpochBlock = block.number;
     } else {
-      // historySize < randomnessBlockRetentionWindow
-      historySize++;
+      if (historySize == 0) {
+        historyFirst = block.number;
+        historySize = 1;
+      } else if (historySize > randomnessBlockRetentionWindow) {
+        delete history[historyFirst];
```
asaj: Looks like this would cause us to delete the randomness of the last epoch block

asaj (changes requested):

asaj:
Please add tests!

packages/protocol/contracts/identity/Random.sol
lucasege: Agreed. I was initially confused if this case would ever be reached, but now I'm thinking this case is to handle if we set `randomnessBlockRetentionWindow` to a new value less than `historySize`. I'll add another conditional

packages/protocol/contracts/identity/Random.sol
```diff
@@ -93,20 +95,31 @@ contract Random is IRandom, Ownable, Initializable {
    */
   function addRandomness(uint256 blockNumber, bytes32 randomness) internal {
     history[blockNumber] = randomness;
-    if (historySize == 0) {
-      historyFirst = block.number;
```
lucasege, line 99: Was this correct? I changed this to be `blockNumber` now, after seeing some bad behavior in my tests.

nambrot: Interesting question. My understanding is that this should always equal to `block.number`? What was the bad behavior?

packages/protocol/test/identity/random.ts
```diff
@@ -101,5 +105,43 @@ contract('Random', (accounts: string[]) => {
         assert.equal(randomValues[3], await random.getTestRandomness(3, 6))
       })
     })
+
+    describe("when relying on the last block of each epoch's randomness", async () => {
+      const EPOCH = 100
```
nambrot: I assume this is our default, can we just get this with the precompile?

packages/protocol/contracts/identity/Random.sol
```diff
@@ -93,20 +95,31 @@ contract Random is IRandom, Ownable, Initializable {
...
-    } else if (historySize > randomnessBlockRetentionWindow) {
-      delete history[historyFirst];
-      delete history[historyFirst + 1];
-      historyFirst += 2;
-      historySize--;
-    } else if (historySize == randomnessBlockRetentionWindow) {
-      delete history[historyFirst];
-      historyFirst++;
+    if ((blockNumber + 1) % getEpochSize() == 0) {
+      if (lastEpochBlock != historyFirst) {
```
nambrot: Interesting, haven't thought about it. If the retention window is actually larger than the epoch size, then we are in trouble too? I.e. we should only delete `if lastEpochBlock < historyFirst`?

packages/protocol/test/identity/random.ts
```diff
@@ -101,5 +105,43 @@ contract('Random', (accounts: string[]) => {
...
+        // i = 2 to only consider elements within the retentionWindow + 1: [2, 3, 4, 5, 6]
+        for (let i = 2; i <= retentionWindow + 1; i++) {
+          assert.equal(
+            randomValues[i],
+            await random.getTestRandomness(lastBlockOfEpoch + i, totalBlocks)
+          )
+        }
+      })
+
+      it('should still not retain other blocks not covered by the retention window', async () => {
```
nambrot, line 158: What do you think about a test that tests that we are properly removing the randomness of the previous epoch block?

lucasege: Sounds like a good idea

packages/protocol/contracts/identity/Random.sol
lucasege: Mine was a special case in testing when I was passing arbitrary blocknumbers in rather than using the real current `block.number`.
The part I think was "incorrect" here before was using both, i.e. `history[blockNumber]` and `historyFirst = block.number` I think both should use the same. Like you said, should never be a problem but was confusing to me

packages/protocol/contracts/identity/Random.sol
lucasege: True. The code around deleting seems very fishy to me. Good catch

packages/protocol/test/identity/random.ts
```diff
@@ -107,40 +109,59 @@ contract('Random', (accounts: string[]) => {
...
 
     describe("when relying on the last block of each epoch's randomness", async () => {
-      const EPOCH = 100
       let lastBlockOfEpoch: any
       let totalBlocks: any
       const retentionWindow = 5
       beforeEach(async () => {
         const epochNumber = await currentEpochNumber(web3)
-        lastBlockOfEpoch = (epochNumber + 1) * EPOCH - 1
+        lastBlockOfEpoch = (epochNumber + 1) * EPOCH - 2
```
nambrot: Why `2`?

packages/protocol/test/identity/random.ts
```diff
@@ -107,40 +109,59 @@ contract('Random', (accounts: string[]) => {
...
 
       it('should retain the usual `retentionWindow` worth of blocks', async () => {
-        // i = 2 to only consider elements within the retentionWindow + 1: [2, 3, 4, 5, 6]
-        for (let i = 2; i <= retentionWindow + 1; i++) {
+        for (let i = 0; i < retentionWindow; i++) {
           assert.equal(
-            randomValues[i],
-            await random.getTestRandomness(lastBlockOfEpoch + i, totalBlocks)
+            randomValues[1],
+            await random.getTestRandomness(totalBlocks - i, totalBlocks)
```
nambrot: I'm ESL, but what do you think about `currentBlock` instead of `totalBlocks`?

packages/protocol/test/identity/random.ts
```diff
@@ -107,40 +109,59 @@ contract('Random', (accounts: string[]) => {
...
+          totalBlocks += 1
+          await random.addTestRandomness(totalBlocks, randomValues[1])
         }
       })
 
-      it('should retain the last epoch block randomness', async () => {
-        assert.equal(randomValues[0], await random.getTestRandomness(lastBlockOfEpoch, totalBlocks))
+      it("should retain the last epoch block's randomness", async () => {
+        // Get start of epoch and then subtract one for last block of previous epoch
+        const lastEpochBlock = totalBlocks - (totalBlocks % EPOCH) - 1
```
nambrot: What is the difference between `lastBlockOfEpoch` and `lastEpochBlock`?

packages/protocol/contracts/identity/Random.sol
```diff
@@ -93,20 +95,33 @@ contract Random is IRandom, Ownable, Initializable {
...
-      historySize = 1;
-    } else if (historySize > randomnessBlockRetentionWindow) {
-      delete history[historyFirst];
-      delete history[historyFirst + 1];
-      historyFirst += 2;
-      historySize--;
-    } else if (historySize == randomnessBlockRetentionWindow) {
-      delete history[historyFirst];
-      historyFirst++;
+    if ((blockNumber + 1) % getEpochSize() == 0) {
```
asaj: Why `+1`?

packages/protocol/contracts/identity/Random.sol
```diff
@@ -93,20 +95,33 @@ contract Random is IRandom, Ownable, Initializable {
...
+      }
+      lastEpochBlock = blockNumber;
     } else {
-      // historySize < randomnessBlockRetentionWindow
-      historySize++;
+      if (historySize == 0) {
+        historyFirst = blockNumber;
+        historySize = 1;
+      } else if (historySize > randomnessBlockRetentionWindow) {
+        if (historyFirst != lastEpochBlock) {
```
asaj: nit: Consider, for readability, using the following here and below:

```
function deleteHistoryIfNotLastEpochBlock(uint256 blockNumber) internal {
  if (blockNumber != lastEpochBlock) {
    delete history[blockNumber];
  }
}
```

packages/protocol/contracts/identity/Random.sol
lucasege: Confusion over epoch indexing - to be clear for others: epochs are arranged by blocks as:
block 0 : epoch 0
block 1 - 100 (inclusive): epoch 1
block 101 - 200 (inclusive): epoch 2
Thus the last block of any epoch is a multiple of epoch_size, so an epoch's block number is sort-of 1-indexed. The typescript testing utils seem to be built aware of the 1-indexing, but not of the 0'th epoch, thus it treats block 1 as in epoch 0, and block 0 would break `currentEpochNumber`.
Luckily our tests only seem to use `currentEpochNumber` as a relative measurement, but I'll create an issue to fix this.

asaj: I think @jfoutts-celo has pointed this out as well

asaj (approved):

packages/protocol/contracts/identity/Random.sol
jfoutts-celo: Created Issue #2718 Epoch number definition mismatch, to track this.

lucasege:
Tested this with `nambrot/historical-randomness` on geth and was able to sync nodes without any error messages around failing to fetch randomness, hoping that means its good to go.

asaj:
🚢

PR #2364, opened 2020-01-04
Move most of mobile contract calls to use ContractKit wrappers

cmcewen:
### Description

This begins moving more of mobile to use contractkit

PR #2706, opened 2020-02-07
[contractKit] Refactor Celo Provider

### LINKED ISSUE #810: ContracKit LocallySigned Setup SN fetch every single block
### Expected Behavior

By looking at https://github.com/MetaMask/web3-provider-engine/blob/master/index.js
it seems that the ProviderEngine we use for locally signed transaction is constantly polling for new blocks, and getting the full block, thus totally suboptimal when we are using a ultralight client; since we'll get every block.

We should investigate why this behaviour and if this can be avoided. I don't know why ProviderEngine does this.

### LINKED ISSUE #1858: properly shutdown the local signing provider
### Expected Behavior

No spurious errors.

### Current Behavior

Whenever using the ContractKit local signing provider, we see lots of this:

```
Error: Invalid JSON RPC response: ""
    at Object.InvalidResponse (/Users/tim/Celo/celo-monorepo-new2/node_modules/web3-core-helpers/src/errors.js:42:16)
    at XMLHttpRequest.request.onreadystatechange (/Users/tim/Celo/celo-monorepo-new2/node_modules/web3-providers-http/src/index.js:92:32)
    at XMLHttpRequestEventTarget.dispatchEvent (/Users/tim/Celo/celo-monorepo-new2/node_modules/xhr2-cookies/xml-http-request-event-target.ts:44:13)
```

We don't properly shutdown the provider, so e.g. during metadata upload to GCS it gets rudely interrupted when we stop port-forwarding.

@yerdua said: I believe the problem is related to the account being added by instantiating a new web3 provider here: https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/utils/web3-utils.ts#L13-L17

### LINKED ISSUE #2619: Improve ContractKit Debugging
### Expected Behavior

Improve/Review contractKit debugging support (aka. DEBUG=”kit:*” ). It should be simple to see what messages are being passed to the node.

gastonponti:
### Description

Refactor of how we use the contractKit:
- The contractKit, will have a CeloProvider (not optional anymore)

Refactor in CeloProvider
- The CeloProvider will have a Wallet that contains the local accounts and can sign Tx/Messages with them
- The MissingTxParamsPopulator, fills the required data that was not added.
- Removed the 0x dependencies (change the logic to not use web3-provider-engine and subProviders anymore)
- Add the debug functionality to the CeloProvider, instead of using a wrapper to do it

### Tested

Unit tests added for:
- CeloProvider (packages/contractkit/src/providers/celo-provider.test.ts)
- MissingTxParamsPopulator (packages/contractkit/src/utils/missing-tx-params-populator.test.ts)
- RpcCaller (packages/contractkit/src/utils/rpc-caller.test.ts)
- Wallet (packages/contractkit/src/utils/wallet.test.ts)

### Related issues

- Fixes #2619
- Fixes #1858 
- Fixes #810 

### Backwards compatibility

- The constructor for the kit, needs a web3 instance with a *valid provider*


packages/contractkit/package.json
```diff
@@ -26,13 +26,14 @@
...
   "dependencies": {
-    "@0x/subproviders": "^5.0.0",
     "@celo/utils": "0.1.6-dev",
+    "@types/ethereumjs-util": "^5.2.0",
     "@types/debug": "^4.1.5",
     "bignumber.js": "^7.2.0",
     "cross-fetch": "3.0.4",
     "debug": "^4.1.1",
     "eth-lib": "^0.2.8",
+    "ethers": "4.0.0-beta.3",
```
yorhodes: I'm not sure we want to add this dependency
there is ongoing work to minimize bundle sizes (ie stabilize version of web3 used across contractkit and mobile) and this bloats the package
I also don't think we want to encourage devs to use web3 and ethers

packages/contractkit/src/utils/sign_typed_data_utils.ts
```diff
@@ -0,0 +1,111 @@
...
+      encodedValues.push(hashValue)
+    } else if (field.type.lastIndexOf(']') === field.type.length - 1) {
+      throw new Error('Arrays currently unimplemented in encodeData')
+    } else {
+      encodedTypes.push(field.type)
+      const normalizedValue = _normalizeValue(field.type, value)
+      encodedValues.push(normalizedValue)
+    }
+  }
+  return ethers.utils.defaultAbiCoder.encode(encodedTypes, encodedValues)
```
yorhodes: can we replace this with web3 equivalent?

packages/contractkit/src/kit.ts
```diff
@@ -270,4 +278,13 @@ export class ContractKit {
     // Reverses protocol/contracts getEpochNumber()
     return (epochNumber + 1) * epochSize.toNumber()
   }
+
+  stop() {
+    // This should be true except that the user actively decide to change the web3 provider
```
yorhodes: nit: `sed 's/that/when'`

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,233 @@
...
+
+  private async handleSignTransaction(payload: JsonRPCRequest): Promise<any> {
+    const txParams = payload.params[0]
+    const filledParams = await this.paramsPopulator.populate(txParams)
+    const signedTx = await this.wallet.signTransaction(filledParams)
+    return { raw: signedTx.raw, tx: txParams }
+  }
+
+  private async handleSendTransaction(payload: JsonRPCRequest): Promise<any> {
+    const txParams = payload.params[0]
```
yorhodes: nit: dedup with `handleSignTransaction`

packages/contractkit/src/utils/celo-tx.ts
```diff
@@ -0,0 +1,7 @@
+import { Tx } from 'web3/eth/types'
+
+export interface CeloTx extends Tx {
```
yorhodes: I believe we have this interface in contractkit already

yorhodes: I see you deleted it
maybe try to merge with `types/web3.d.ts`

packages/contractkit/src/utils/wallet.ts
```diff
@@ -0,0 +1,115 @@
...
+  // Original code taken from
+  // https://github.com/0xProject/0x-monorepo/blob/78c704e3d/packages/subproviders/src/subproviders/private_key_wallet.ts
+  /**
+   * Sign an EIP712 Typed Data message. The signing address will be calculated from the private key.
+   * The address must be provided it must match the address calculated from the private key.
+   * @param address Address of the account to sign with
+   * @param data the typed data object
+   * @return Signature hex string (order: rsv)
+   */
+  async signTypedData(address: string, typedData: EIP712TypedData): Promise<string> {
```
yorhodes: nit: dedup with `signPersonalMessage`
this code block also looks familiar, see if there is anything in utils which does the same

packages/contractkit/package.json
gastonponti: The problem is that web3 is already using it. This is from the `yarn.lock`
```
web3-eth-abi@1.0.0-beta.37:
  version "1.0.0-beta.37"
  resolved "https://registry.yarnpkg.com/web3-eth-abi/-/web3-eth-abi-1.0.0-beta.37.tgz#55592fa9cd2427d9f0441d78f3b8d0c1359a2a24"
  integrity sha512-g9DKZGM2OqwKp/tX3W/yihcj7mQCtJ6CXyZXEIZfuDyRBED/iSEIFfieDOd+yo16sokLMig6FG7ADhhu+19hdA==
  dependencies:
    ethers "4.0.0-beta.1"
    underscore "1.8.3"
    web3-utils "1.0.0-beta.37"
```
Of course, I have to change the "4.0.0-beta.3" to "4.0.0-beta.1", but this dependency was already in use.
I've just added it because it was wrong to asume that we had it, just because we have the web3

gastonponti: Ok, I can change this to use web3. We sill have the problem that web3 uses the ethers, but for now it's easier to use it from web3

From the version `2.0.0-alpha` they added its own implementation

mcortesi: My two cents...

1. web3.js uses etherjs for abi coding. But it does some magic, so for now instead of using web3 abi coder, we should use ether.js directly.
2. We should use the exact same version that web3.js uses, so we don't have 2 different ether.js version around.

packages/contractkit/src/utils/wallet.ts
mcortesi: mmm.. for EIP712 i'm pretty sure we don't.. but idk.

packages/contractkit/package.json
yorhodes: hmm I didn't know web3 used ethers 🤷‍♂

packages/contractkit/src/utils/celo-tx.ts
gastonponti: Ok I can use the `Tx` from the `web3/eth/types` that's the same

packages/contractkit/package.json
gastonponti: I checked and the function that we are using should work without any change from the web3 abi coder

packages/contractkit/src/utils/wallet.ts
gastonponti: No, we don't have the version for the EIP712
I tried not to change these function because are being copied from the 0x repo. I know that I could pass a function parameter and unify all, but probably won't be clear for the person following the link to the place that was extracted

gastonponti:
The code `packages/mobile/__mocks__/@celo/contractkit/index.ts` was taken from #2364 to fix the failing tests

packages/contractkit/src/utils/celo-tx.ts
gastonponti: Removed

packages/contractkit/src/providers/celo-provider.ts
gastonponti: changed!

packages/contractkit/package.json
```diff
@@ -26,8 +26,8 @@
     "lint": "tslint -c tslint.json --project ."
   },
   "dependencies": {
-    "@0x/subproviders": "^5.0.0",
     "@celo/utils": "0.1.6-dev",
+    "@types/ethereumjs-util": "^5.2.0",
```
mcortesi: @types dependencies should be devDependencies

packages/contractkit/package.json
```diff
@@ -26,8 +26,8 @@
     "lint": "tslint -c tslint.json --project ."
   },
   "dependencies": {
-    "@0x/subproviders": "^5.0.0",
```
mcortesi, line 30: we also removed provider engine dependency. It doesn't seem to be here...maybe on some other package.json? I would search for it just to be sure

packages/contractkit/src/kit.ts
```diff
@@ -150,7 +157,8 @@ export class ContractKit {
   }
 
   addAccount(privateKey: string) {
-    addLocalAccount(this.web3, privateKey)
+    const provider = this.web3.currentProvider as CeloProvider
```
mcortesi: Just to be sure. It's better to check with `instanceof` and in the case it's not a CeloProvider throw an error.

We don't know if the user changed the provider without us knowing so

packages/contractkit/src/kit.ts
```diff
@@ -270,4 +278,13 @@ export class ContractKit {
     // Reverses protocol/contracts getEpochNumber()
     return (epochNumber + 1) * epochSize.toNumber()
   }
+
+  stop() {
+    // This should be true except when the user actively decide to change the web3 provider
+    // after creating the kit
+    if (this.web3.currentProvider && this.web3.currentProvider.hasOwnProperty('stop')) {
```
mcortesi: maybe check if it's a instanceof of CeloProvider.. if not throw

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,231 @@
...
-import { WrappingSubprovider } from './wrapping-subprovider'
+import { Socket } from 'net'
+import { Callback, JsonRPCRequest, JsonRPCResponse, Provider } from 'web3/providers'
+import { MissingTxParamsPopulator } from '../utils/missing-tx-params-populator'
+import { IRpcCaller, RpcCaller, rpcCallHandler } from '../utils/rpc-caller'
+import { IWallet, Wallet } from '../utils/wallet'
 
-const debug = debugFactory('kit:provider:celo-provider')
+const debug = debugFactory('kit:provider:celo-provider:connection')
+const debugPayloadExistingProvider = debugFactory('kit:existing-provider:rpc:payload')
```
mcortesi: i suggest using a namespace to encapsulate both... so something like:

`kit:provider:public:...` and `kit:provider:internal`

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,231 @@
...
+    const nodeAccountsResp = await this.rpcCaller.call('eth_accounts', [])
+
+    return nodeAccountsResp.result.concat(this.wallet.getAccounts())
   }
 
-  addAccount(privateKey: string): CeloProvider {
-    this.localSigningProvider.addAccount(privateKey)
-    return this
+  isLocalAccount(address?: string): boolean {
+    return address != null && this.wallet.hasAccount(address.toLowerCase())
```
mcortesi: the wallet should do the lowercase, since it's its responsability to verify the address no matter the case

mcortesi: same for null

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,231 @@
...
 
   stop() {
-    this.providerEngine.stop()
-
+    if (this.alreadyStopped) {
+      return
+    }
     try {
-      if (this.underlyingProvider.hasOwnProperty('stop')) {
+      if (this.existingProvider.hasOwnProperty('stop')) {
```
mcortesi: Is this for the WebSocket provider ? if so i would add a comment to mention which provider we are interacting with...

same for the IPC one

mcortesi: nit. a simple improvmente to about the ts-ignore is to have:

```
function hasStopFunction(provider) provider is { stop():void } {
    return this.existingProvider.hasOwnProperty('stop')
}

// and then
if (hasStopFunction(provider) {
  provider.stop()
}
```

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,231 @@
...
+    }
   }
 
   stop() {
-    this.providerEngine.stop()
-
+    if (this.alreadyStopped) {
+      return
+    }
     try {
```
mcortesi, line 132: we can probably extract this logic to some provider-utils file, and then have a `stopProvider()`... it might be useful for others that are not using kit

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,231 @@
...
+    return { raw: signedTx.raw, tx: txParams }
+  }
+
+  private async handleSendTransaction(payload: JsonRPCRequest): Promise<any> {
+    const signedTx = await this.handleSignTransaction(payload)
+    const response = await this.rpcCaller.call('eth_sendRawTransaction', [signedTx.raw])
+    return response.result
+  }
+
+  private _forwardSend(payload: JsonRPCRequest, callback: Callback<JsonRPCResponse>): void {
```
mcortesi: nit: remove the _

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,231 @@
...
+      debugResponseExistingProvider
+    )
+    this.existingProvider.send(payload, decoratedCallback)
+  }
+
+  private celoCallback(
+    payload: JsonRPCRequest,
+    callback: Callback<JsonRPCResponse>
+  ): Callback<JsonRPCResponse> {
+    return this.debugDecoratorFn(
```
mcortesi: not sure if i like this... you're doing the first log too late. Since it might be already failed before getting to this line...

I would suggest logging as one of the first things you do on `send()`...
also maybe there, you can replace the callback... for a "log & call original" callback.

packages/contractkit/src/utils/missing-tx-params-populator.ts
```diff
@@ -0,0 +1,110 @@
...
+
+// Default gateway fee to send the serving full-node on each transaction.
+// TODO(nategraf): Provide a method of fecthing the gateway fee value from the full-node peer.
+const DefaultGatewayFee = new BigNumber(10000)
+
+export class MissingTxParamsPopulator {
+  private chainId: number | null = null
+  private gatewayFeeRecipient: string | null = null
+
+  constructor(readonly rpc: IRpcCaller) {}
```
mcortesi: we are using `RpcCaller` for the interface name, and then `DefaultRpcCaller` for the concrete type

packages/contractkit/src/utils/missing-tx-params-populator.ts
```diff
@@ -0,0 +1,110 @@
...
+    return this.gatewayFeeRecipient
+  }
+
+  private async getGasPrice(feeCurrency: string | undefined): Promise<string | undefined> {
+    // Gold Token
+    if (!feeCurrency) {
+      return this.getGasPriceInCeloGold()
+    }
+    throw new Error(
+      `celo-private-keys-subprovider@getGasPrice: gas price for ` +
```
mcortesi: fix the error message

packages/contractkit/src/utils/missing-tx-params-populator.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  private async getGasPriceInCeloGold(): Promise<string> {
+    // Reference: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gasprice
+    const result = await this.rpc.call('eth_gasPrice', [])
+    const gasPriceInHex = result.result.toString()
+    return gasPriceInHex
+  }
+
+  private isEmpty(value: string | undefined) {
```
mcortesi: this should an utility function, outside the class

packages/contractkit/src/utils/rpc-caller.ts
```diff
@@ -0,0 +1,69 @@
+import { Callback, JsonRPCRequest, JsonRPCResponse, Provider } from 'web3/providers'
+
+export function rpcCallHandler(
+  payload: JsonRPCRequest,
+  handler: (p: JsonRPCRequest) => Promise<any>,
+  callback: Callback<JsonRPCResponse>
+) {
+  handler(payload).then(
```
mcortesi: missing cases...

what if `handler()` fails before creating a promise, outside a promise context... you need a try/catch for it...

2) what if then() execution fails... it will go to the catch, thus calling the callback twice...

maybe something like:

```ts
try {
  const p = handler()
  p.then(....)
  p.catch(..)
} catch(err) {
 ....
}
```

mcortesi: with some love this could be a `promisfy()` function for utils

packages/contractkit/src/utils/wallet.ts
```diff
@@ -0,0 +1,115 @@
...
+  privateKeyToAddress,
+  trimLeading0x,
+} from '@celo/utils/lib/address'
+import * as ethUtil from 'ethereumjs-util'
+import { Tx } from 'web3/eth/types'
+import { EncodedTransaction } from 'web3/types'
+import { EIP712TypedData, generateTypedDataHash } from './sign-typed-data-utils'
+import { signTransaction } from './signing-utils'
+
+export interface IWallet {
```
mcortesi: no `I` prefix

packages/contractkit/src/utils/wallet.ts
```diff
@@ -0,0 +1,115 @@
...
+  signPersonalMessage: (address: string, data: string) => Promise<string>
+}
+
+export function normalizeKey(key: string) {
+  return ensureLeading0x(key).toLowerCase()
+}
+
+export class Wallet implements IWallet {
+  // Account addresses are hex-encoded, lower case alphabets
+  private readonly accountAddressToPrivateKey = new Map<string, string>()
```
mcortesi: we could name this `privateKeys` for simplicity

packages/contractkit/src/utils/wallet.ts
```diff
@@ -0,0 +1,115 @@
...
+export interface IWallet {
+  addAccount: (privateKey: string) => void
+  getAccounts: () => string[]
+  hasAccount: (address: string) => boolean
+  signTransaction: (txParams: Tx) => Promise<EncodedTransaction>
+  signTypedData: (address: string, typedData: EIP712TypedData) => Promise<string>
+  signPersonalMessage: (address: string, data: string) => Promise<string>
+}
+
+export function normalizeKey(key: string) {
```
mcortesi: `normalizeAddress` maybe? and use type `Address`

packages/contractkit/src/utils/wallet.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  private getPrivateKeyFor(account: string): string {
+    if (account) {
+      const maybePk = this.accountAddressToPrivateKey.get(normalizeKey(account))
+      if (maybePk != null) {
+        return maybePk
+      }
+    }
+    throw Error(`tx-signing@getPrivateKey: ForPrivate key not found for ${account}`)
```
mcortesi: fix error message

packages/contractkit/src/utils/wallet.ts
mcortesi: There's a normalizeAddress on utils...

packages/contractkit/package.json
gastonponti: the @0x/subproviders that we were using has this version of the provider-engine:
`"web3-provider-engine": "14.0.6"`
but I'll check the other packages too

packages/contractkit/src/providers/celo-provider.ts
gastonponti: Change it to use as different logs the `rpc call` and the `provider send`

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,170 @@
...
-    // Initializer "on" conditionally.
-    if (existingProvider.hasOwnProperty('on')) {
-      this.on = (event: string, handler: () => void): void => {
-        this.providerEngine.on(event, handler)
-      }
-    }
+  private readonly wallet: Wallet
+  private readonly rpcCaller: RpcCaller
+  private readonly paramsPopulator: MissingTxParamsPopulator
+  private alreadyStopped: boolean = false
```
jmrossy, line 25: Can a provider be stopped and then restarted?

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,170 @@
...
+    if (this.alreadyStopped) {
+      throw Error('CeloProvider already stopped')
+    }
 
-    try {
-      if (this.underlyingProvider.hasOwnProperty('stop')) {
-        // @ts-ignore
-        this.underlyingProvider.stop()
+    switch (payload.method) {
+      case 'eth_accounts': {
```
jmrossy: Can we use an enum for these values

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,80 +1,170 @@
...
+    return this.getAccounts()
+  }
+
+  private async handleSignTypedData(payload: JsonRPCRequest): Promise<any> {
+    const [address, typedData] = payload.params
+    const signature = this.wallet.signTypedData(address, typedData)
+    return signature
+  }
+
+  private async handleSign(payload: JsonRPCRequest): Promise<any> {
```
jmrossy: Suggest rename to `handleSignPersonalMessage` for clarity

packages/contractkit/src/utils/missing-tx-params-populator.ts
```diff
@@ -0,0 +1,109 @@
...
+  return (
+    value === undefined ||
+    value === null ||
+    value === '0' ||
+    value.toLowerCase() === '0x' ||
+    value.toLowerCase() === '0x0'
+  )
+}
+
+export class MissingTxParamsPopulator {
```
jmrossy: Nit: How about `TxParamsNormalizer`?

packages/contractkit/src/utils/missing-tx-params-populator.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (this.gatewayFeeRecipient == null) {
+      throw new Error(
+        'missing-tx-params-populator@getCoinbase: Coinbase is null, we are not connected to a full ' +
+          'node, cannot sign transactions locally'
+      )
+    }
+    return this.gatewayFeeRecipient
+  }
+
+  private async getGasPrice(feeCurrency: string | undefined): Promise<string | undefined> {
```
jmrossy: no need for async here

packages/contractkit/src/utils/sign-typed-data-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+  return ethUtil.sha3(
+    Buffer.concat([
+      Buffer.from('1901', 'hex'),
+      _structHash('EIP712Domain', typedData.domain, typedData.types),
+      _structHash(typedData.primaryType, typedData.message, typedData.types),
+    ])
+  ) as Buffer
+}
+
+function _findDependencies(
```
jmrossy: Nit: no need for leading underscores. If the functions aren't exported then they're private to this module anyway

packages/mobile/__mocks__/@celo/contractkit/index.ts
```diff
@@ -0,0 +1,45 @@
+import BigNumber from 'bignumber.js'
+
+export const newKitFromWeb3 = () => ({
```
jmrossy, line 3: I would suggest keeping a contract kit mock in the contract kit package so we can keep it up to date and then have consumers like mobile import it here

jmrossy (commented):
Overall looks great! just some small comments below

packages/contractkit/src/utils/sign-typed-data-utils.ts
gastonponti: Yes! My bad

packages/contractkit/src/providers/celo-provider.ts
gastonponti: No, we are closing/destroying the underlying connection. We could eventually make a restart, if we get all the connection info before the disconnection, but I'm not sure if there is a use case where the user wants that instead of creating again the provider.
I mean, the provider has some state, so probably we will need to understand what to add or remove in a restart (and have a deeper control of the connection)
In my opinion, the cost-benefit of a restart instead of re-creating the Provider it not seems good.
One thing that we could eventually do, is to make the `wallet` used by the `CeloProvider` settable, this way you can use the same `local addresses data` for different Providers

packages/contractkit/src/utils/rpc-caller.ts
```diff
@@ -0,0 +1,95 @@
...
+      .then(
+        (result) => {
+          callback(null, toRPCResponse(payload, result))
+        },
+        (error) => {
+          callback(error, toRPCResponse(payload, null, error))
+        }
+      )
+      .catch((error) => {
+        debugRpcCallback('Callback for handling the JsonRPCResponse fails')
```
mcortesi, line 26: Add a comment explaining when we would get into this catch

packages/docs/developer-resources/contractkit/reference/classes/_contract_cache_.wrappercache.md
```diff
@@ -44,7 +44,7 @@ Provides access to all contract wrappers for celo core contracts
 
 \+ **new WrapperCache**(`kit`: [ContractKit](_kit_.contractkit.md)): *[WrapperCache](_contract_cache_.wrappercache.md)*
 
-*Defined in [packages/contractkit/src/contract-cache.ts:79](https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/contract-cache.ts#L79)*
+*Defined in [contractkit/src/contract-cache.ts:79](https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/contract-cache.ts#L79)*
```
mcortesi, line 47: @yorhodes is this path change desired??

mcortesi (approved):
First PR Congrats!!!

packages/docs/developer-resources/contractkit/reference/classes/_contract_cache_.wrappercache.md
gastonponti: I verified and at least the links are working fine

PR #2708, opened 2020-02-07
[Wallet] Improve clipboard watching

### LINKED ISSUE #2592: Invite and Verification Code paste interaction does not consistently work
### Expected Behavior

- the paste icon should always display if a code is copied to clipboard
- android and iOS long tap should bring up paste icon (support existing paradigm)

### Current Behavior

Once verification code (or invite code) text box has been highlighted, the paste icon does not display and the code can therefore not be pasted (requires backgrounding and reopening app)

jmrossy:
### Description

Add logic to have paste watcher recheck the clipboard every second

### Tested

Played around with the invite and verification input fields

### Other changes

Fix some minor issues with logging in wallet and faucet
Add mock for RN secure key store to prevent jest hanging handles

### Related issues

- Fixes #2592 

### Backwards compatibility

Yes


annakaz (approved):
Great 🚢 

While this is beyond the scope of this PR, I wonder how we think about flagging features that could slow down the app on low end devices but provide a better experience for most users

jmrossy:
> Great 🚢
> 
> While this is beyond the scope of this PR, I wonder how we think about flagging features that could slow down the app on low end devices but provide a better experience for most users

My intuition is that a 1s setInterval is totally negligible in terms of impact. But in general I'm open to flagging and streamlining for slow phones! I have some ideas on things we could cut, like logging, analytics, or geth itself

PR #2709, opened 2020-02-07
Increment @celo/utils, @celo/contractkit, and @celo/celocli version numbers

jfoutts-celo:
### Description

Increment package version numbers and inter-package dependencies, following https://github.com/celo-org/celo-monorepo/blob/master/README-dev.md#monorepo-inter-package-dependencies


yorhodes (approved):
LGTM

jmrossy (approved):
We should probably update notification service and blockchain api to use the latest published NPM versions but that can be a different PR :)

Issue #2373, opened 2020-01-07
Cash In Integration with Moonpay


nityas:
### Expected Behavior

User SBAT cash in and out with moonpay.io API. [Full Spec](https://docs.google.com/document/d/1xfnIZtPutiJOO6tLiTZqgVblbqGbvoJcdSPtbgCYWj4/edit#)

### Current Behavior

no integration

annakaz:
Update: 
Moonpay widget merged, disabled with a feature flag: https://github.com/celo-org/celo-monorepo/pull/2710
Handling the request signing with a firebase function: https://github.com/celo-org/celo-monorepo/pull/2749

Waiting on Moonpay to integrate with our testnet, expected end of week (2/14) perhaps spillover to the weekend (2/16)

annakaz:
Removed Cash Out from this issue!

annakaz:
Closing this out as other work is captured in Cash In redesign

PR #2710, opened 2020-02-08
[Wallet] Add funds with Moonpay widget

annakaz:
### Description

Add the Moonpay widget to the Account page with the option to buy funds (disabled with a feature flag right now). 

Note that the "signMoonpay" firebase function signing the requests is not included in this PR, as I am not sure whether it should be added to the monorepo. It is a straightforward function for HMAC signatures deployed here: https://console.cloud.google.com/functions/details/us-central1/signMoonpay?organizationId=54829595577&project=celo-org-mobile&tab=general
This signing is required if we want to prefill a user's address and is linked to the C Labs Moonpay account. 

### Tested

Tested that the widget comes up. Since we do not yet have production keys I cannot test that I can buy anything (can only go through auth flow), but we will ~~chat with Moonpay on Monday to get this started~~  be able to test this integration by the end of the week after Moonpay integrates on their side.

### Related issues

- Partially addresses #2373


annakaz:
Screenshot_20200210-133609

annakaz:
Note that build is failing on master (unrelated to this PR)

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -0,0 +1,89 @@
...
+  componentDidMount() {
+    if (this.props.account) {
+      signMoonpayUrl(this.props.account, this.props.localCurrency)
+        .then((signedUrl) => this.setState({ signedUrl }))
+        .catch(() => this.handleError())
+    }
+  }
+
+  handleError = () => {
+    showError(ErrorMessages.FIREBASE_FAILED)
```
cmcewen, line 78: Is this the right error?

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -0,0 +1,89 @@
...
+    headerTitle: i18n.t('accountScreen10:addFunds'),
+  })
+
+  state: State = {
+    signedUrl: '',
+  }
+
+  componentDidMount() {
+    if (this.props.account) {
+      signMoonpayUrl(this.props.account, this.props.localCurrency)
```
cmcewen: Could you make this a separate function called in componentDidMount so that you can use async/await?

cmcewen (approved):

packages/mobile/src/account/FiatExchange.tsx
annakaz: Yeah this is an error thrown when Firebase can't sign the URL

Issue #2673, opened 2020-02-05
Address concerns from smart contract audits audit concerns


nategraf:
Tracking issue for issue that come up related to smart contract audits

Issue #2011, opened 2019-12-03
Certora should be able to run in CI


asaj:
### Expected Behavior

Smart contract rules are checked in Circle

### Current Behavior

No rules, no checking

PR #2721, opened 2020-02-10
Certora Lockedgold and Accounts verification

lucasege:
### Description

Adding Certora scripts, circleCI config changes, and specifications/harnesses for the LockedGold and Accounts contracts.

### Tested

Changes to contracts are minimal (function visibility settings), but those have been tested.

### Other changes

Changing `Accounts.sol.accounts` and `LockedGold.sol.balances`  to internal.

### Related issues

- #2673 and #2011 - Should not close either issue. 

### Backwards compatibility

Yes


.circleci/config.yml
```diff
@@ -845,11 +845,66 @@ jobs:
...
 
+
+  certora-test:
+    working_directory: ~/app
+    docker:
+      - image: circleci/openjdk:11-jdk
+    steps:
+      - attach_workspace:
+          at: ~/app
+      - run:
```
asaj, line 801: Flagging that we'll want to use the smarter CI running tool @AlexBHarley is working on in https://github.com/celo-org/celo-monorepo/pull/3761

.circleci/config.yml
```diff
@@ -845,11 +845,66 @@ jobs:
...
+            mv solc-static-linux solc
+            chmod +x solc
+      - run:
+          name: Build and run - Locked Gold
+          command: |
+            cd packages/protocol
+            python3 scripts/certora/certoraBuild.py specs/harnesses/GoldTokenHarness.sol specs/harnesses/LockedGoldHarness.sol --link LockedGoldHarness:goldToken=GoldTokenHarness --solc ~/app/solc --verify LockedGoldHarness:specs/lockedGold.spec --settings -assumeUnwindCond
+            CERTORA_RUN_NAME=LockedGoldRun`date +%s`           
+            java -jar scripts/certora/prover_cli.jar verify --name=$CERTORA_RUN_NAME .
+            python3 scripts/certora/getMyCertoraOutput.py celo_ci $CERTORA_RUN_NAME
```
asaj: Can we have this take the output file as an argument?

.circleci/config.yml
```diff
@@ -845,11 +845,66 @@ jobs:
...
+            CERTORA_RUN_NAME=AccountsRun`date +%s`           
+            java -jar scripts/certora/prover_cli.jar verify --name=$CERTORA_RUN_NAME .
+            python3 scripts/certora/getMyCertoraOutput.py celo_ci $CERTORA_RUN_NAME
+            mv out.json AccountsActual.json
+            res2=`python3 scripts/certora/compareResults.py AccountsActual.json specs/AccountsExpected.json`||true
+            rc2=$?
+            echo Result $res2
+            echo Success $rc2
+      - run:
+          name: Check all results are good
```
asaj: Is this redundant? It seems like we do something similar in the `Build and run` steps

packages/protocol/accounts.conf
```diff
@@ -0,0 +1,54 @@
+{
+    "contracts": [
```
asaj: This appears to have some formatting issues

packages/protocol/accounts.conf
```diff
@@ -0,0 +1,54 @@
+{
```
asaj: Under what circumstances does this file need to be updated?

packages/protocol/lockedGold.conf
```diff
@@ -0,0 +1,60 @@
+{
```
asaj: Same questions as `account.conf`

packages/protocol/accounts.conf
asaj: Should this file be in a `conf` directory? Seems like we will have one per contract

packages/protocol/lockedGold.conf
```diff
@@ -0,0 +1,60 @@
...
+        "specs/harnesses/LockedGoldHarness.sol:LockedGoldHarness",
+		"specs/harnesses/GoldTokenHarness.sol"
+		
+    ],
+	"link": {
+        "LockedGoldHarness": {
+            "goldToken": "GoldTokenHarness"
+        }
+    },
+    "output": ".certora_build",
```
asaj: A lot of this looks the same as `accounts.conf`, does it make sense to consolidate the shared pieces?

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
...
+import re
+import platform
+from collections import OrderedDict
+from datetime import datetime
+import distutils.dir_util
+import certoraUtils
+from certoraUtils import debug_print, is_windows, safe_create_dir, run_cmd, get_file_basename, get_file_extension, read_from_conf, handle_file_list, get_file_name, current_conf_to_file
+from certoraUtils import OPTION_SOLC, OPTION_SOLC_ARGS, OPTION_SOLC_MAP, OPTION_PATH, OPTION_OUTPUT, OPTION_OUTPUT_FOLDER, OPTION_OUTPUT_VERIFY, OPTION_VERIFY, OPTION_LINK, OPTION_PACKAGES, OPTION_PACKAGES_PATH, OPTION_CACHE, DEFAULT_CONF, OPTION_ADDRESS, OPTION_LINK_CANDIDATES, fatal_error
+
+def print_usage(): # TODO use the macros in print usage as well?
```
asaj: I think using `argparse` might be helpful here?

https://docs.python.org/3/library/argparse.html

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
...
+	if OPTION_PACKAGES_PATH not in parsed_options:
+		parsed_options[OPTION_PACKAGES_PATH] = os.getenv("NODE_PATH","%s/node_modules" % (os.getcwd())).replace("\\","/")
+		
+	# If packages were not specified, try to find them from package.json, if it exists
+	if OPTION_PACKAGES not in parsed_options:
+		try:
+			with open("package.json","r") as package_json_file:
+				package_json = json.load(package_json_file)
+				deps = set(list(package_json["dependencies"].keys()) if "dependencies" in package_json else list() + list(package_json["devDependencies"].keys()) if "devDependencies" in package_json else list()) # May need both
+				# Don't know which ones we need, so we take them all
```
asaj: Remove?

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
...
+
+	# Add default for "path"
+	if OPTION_PATH not in parsed_options:
+		parsed_options[OPTION_PATH] = "%s/contracts/" % (os.getcwd().replace("\\","/"))
+		
+	# Add default packages path
+	if OPTION_PACKAGES_PATH not in parsed_options:
+		parsed_options[OPTION_PACKAGES_PATH] = os.getenv("NODE_PATH","%s/node_modules" % (os.getcwd())).replace("\\","/")
+		
+	# If packages were not specified, try to find them from package.json, if it exists
```
asaj: Any reason we can't just always read them from package.json? Seems redundant to additionally specify them in the `.conf` file

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
...
+		cygwin_config_path = "/%s" % (config_path[0].lower() + config_path[1:].replace(":",""))
+		print("Cygwin config path: %s" % (cygwin_config_path))
+			
+
+	def collect_funcs(contract,data):	
+		funcs = []
+		abi = data["contracts"][contract]["abi"]
+		for f in filter(lambda x: x["type"]=="function",json.loads(abi)):
+			inputs = f["inputs"]
+			#inputTypes = ",".join(map(lambda x: "\"%s\""%(x["type"]), inputs))
```
asaj: Remove?

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
...
+			inputTypes = [x["type"] for x in inputs]
+			if "outputs" in f:
+				outputs = f["outputs"]
+				#outputTypes = ",".join(map(lambda x: "\"%s\""%(x["type"]), outputs))
```
asaj: Remove?

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
...
+	library_addresses = [] # Will become primary contracts too
+		
+	safe_create_dir(certoraUtils.config_path)	
+		
+	if "iscygwin" in parsed_options:
+		cygwin_config_path = "/%s" % (config_path[0].lower() + config_path[1:].replace(":",""))
+		print("Cygwin config path: %s" % (cygwin_config_path))
+			
+
+	def collect_funcs(contract,data):
```
asaj: Would `eth-abi` be useful here?

https://pypi.org/project/eth-abi/

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
+#!/usr/bin/env python3
```
asaj: This file is quite large and hard to review, could it be simplified by using existing libraries? Perhaps made more readable by breaking out some functions here into a different file? I'm afraid I don't fully understand what this is doing.

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
...
+			else:
+				isNotPayable = not f["payable"]	# Only if something is definitely non-payable, we treat it as such
+			
+			# Nice to have hex too
+			base = "%s(%s)" % (f["name"],','.join([x.replace('"','') for x in inputTypes]))
+			hash = keccak.new(digest_bits=256)
+			hash.update(str.encode(base))
+			hex = hash.hexdigest()[0:8]
+						
+			funcs.append({"name":f["name"],"args":inputTypes,"returns":outputTypes,"sighash":hex,"notpayable":isNotPayable,"isABI":True})
```
asaj: IIUC this just transforms the JSON ABI into a different format, is there any reason the original format can't be used?

packages/protocol/scripts/certora/certoraBuild.py
```diff
@@ -0,0 +1,708 @@
...
+				
+			sources_to_update = get_matching_sdc_names_from_SDCs(src_contract)
+			for source_to_update in sources_to_update:
+				if "state" not in SDCs[source_to_update]:
+					debug_print("Initializing state of %s" %(source_to_update))
+					SDCs[source_to_update]["state"] = {}
+				debug_print("Linking %s (%s) to %s in slot %s" % (src_contract,source_to_update,dst_address,resolved_src_slot))
+				SDCs[source_to_update]["state"][resolved_src_slot] = dst_address
+
+	#if OPTION_ADDRESS in parsed_options:
```
asaj: Remove?

packages/protocol/scripts/certora/certoraUtils.py
```diff
@@ -0,0 +1,232 @@
...
+	return ''.join(file.split("/")[-1])
+
+def get_file_basename(file):
+	return ''.join(file.split("/")[-1].split(".")[0:-1])
+
+def get_file_extension(file):
+	return file.split("/")[-1].split(".")[-1]
+
+def safe_create_dir(path):	
+		try:
```
asaj: Inconsistent whitespace

packages/protocol/specs/AccountsExpected.json
```diff
@@ -0,0 +1,59 @@
+{
```
asaj: How is this generated?

packages/protocol/scripts/certora/minitester.py
```diff
@@ -0,0 +1,73 @@
...
+                    gotFuncSet = getFunctionsSet(rulesResults[rule])
+                                        
+                    #warnings = checkMissingFunctions(warnings, testName, rule, expectedFuncSet, gotFuncSet)
+                    
+                    for result, funcList in rulesResults[rule].items():
+                        funcList.sort()
+                        expectedRulesResults[rule][result].sort()
+                        if funcList != expectedRulesResults[rule][result]: # compare functions sets (current results with expected)
+                            for funcName in funcList:
+                                # if funcion appears in current results but does not appear in the expected ones
```
asaj: typo

packages/protocol/scripts/certora/minitester.py
```diff
@@ -0,0 +1,73 @@
...
+                if type(rulesResults[rule]) == str: # flat rule ( ruleName: result )
+                    if rulesResults[rule] != expectedRulesResults[rule]:
+                        test = False
+                        errors = addError(errors, testName, rule, rulesResults[rule], expectedRulesResults[rule])
+                        
+                else: # nested rule ( ruleName: {result1: [funcionts list], result2: [funcionts list] ... } )
+                    expectedFuncSet = getFunctionsSet(expectedRulesResults[rule])
+                    gotFuncSet = getFunctionsSet(rulesResults[rule])
+                                        
+                    #warnings = checkMissingFunctions(warnings, testName, rule, expectedFuncSet, gotFuncSet)
```
asaj: remove?

asaj (commented):
Still need to review the harnesses

This is exciting! So so so looking forward to having this run in CI 😄 

My main concern is around the python scripts and their possible fragility. It seems like a lot of things are hard coded. Any ideas on how to make them more robust?

.circleci/config.yml
nd-certora: Currently, the output.json is a fixed name. Would you like this to be parametric?

.circleci/config.yml
nd-certora: We want the verification process to run on all contracts even if one failed. So once everything finished we have another step that checks that everything passed.

packages/protocol/accounts.conf
nd-certora: we are no longer using this file, will remove

packages/protocol/specs/AccountsExpected.json
nd-certora: The content is generated by the tool and we commit this file. This file allows one to indicate that a rule may fail and it is ok. We will change the compare script to not need this file.  In the new version, we will require that everything passes successfully.

.circleci/config.yml
```diff
@@ -862,12 +862,44 @@ jobs:
...
-
+  
+  certora-test:
+    working_directory: ~/app
+    docker:
+      - image: circleci/openjdk:11-jdk
+    steps:
+      - attach_workspace:
+          at: ~/app
+      - run:
```
asaj, line 801: Per https://github.com/celo-org/celo-monorepo/pull/2721/files#r423128751, can we add the following lines before installing certora dependencies:

```
      - run:
          name: Check if the test should run
          command: |
            ./scripts/ci_check_if_test_should_run_v2.sh @celo/protocol
```

IIUC this should prevent the test from running if nothing has changed in the protocol package or its dependencies

asaj (approved):
Wow, what an improvement! Go `certora-cli`!! 🤩

.circleci/config.yml
asaj: Added when resolving conflicts

PR #2722, opened 2020-02-10
Move the faucet to use ContractKit

### LINKED ISSUE #2707: Escrow withdrawal from faucet-generated invite doesn't work
### Expected Behavior

$10 is withdrawn from escrow after verificaiton

### Current Behavior

App reports `Escrow payment is empty, skpping` even though the tx does go through and is visible on blockscout.
Escrow withdrawal from celotool-generated codes works.

cmcewen:
### Description

We were having faucet problems, so I just moved it to use ContractKit to make it easier to work with (and not require redeploys for address changes).

### Tested

Running right now

fixes #2707


jmrossy (approved):
Amazing 🥇

Issue #1590, opened 2019-11-06
Everyone SBAT reference a Governance constitution


asaj:
### Expected Behavior

Governance constitution that makes it harder to upgrade smart contracts, change registry pointers, etc. at a minimum.

### Current Behavior

Everything requires a simple majority

PR #2748, opened 2020-02-11
On-chain governance constitution

asaj:
### Description

Enumerates all on-chain governable functions and the corresponding threshold needed to pass a proposal that calls that function.

### Tested

Not tested.

### Other changes

None

### Related issues

- #1590, will need to be added to the migrations to close out this issue.

### Backwards compatibility

Yes.


nategraf:
We will likely want to go through a round of tuning these before launch, but this looks god to me

nategraf (approved):

PR #2749, opened 2020-02-11
Add Moonpay auth package for signing Moonpay requests

annakaz:
### Description

This package implements a simple firebase function to sign Moonpay requests with a private API key. 

### Tested

Deployed to celo-org-mobile and sent request. 
Deleted .env file to confirm that it can be decrypted

### Related issues

- Partially addresses #2373 

### Backwards compatibility

Yes


jmrossy (commented):
Can we rename this folder to something more descriptive? Like `moonpay-auth`?

scripts/key_placer.sh
```diff
@@ -22,6 +22,7 @@ files=(
   "packages/mobile/secrets.json"
   "packages/web/env-config.js"
   "packages/web/server-env-config.js"
+  "packages/auth/.env"
```
jmrossy: Should be moonpay-auth right?

jmrossy (approved):
Lgtm, just one small problem in key_placer

PR #2755, opened 2020-02-12
Fixing address comparison when getting affiliates

### LINKED ISSUE #2714: celocli validatorgroup:show doesn't show affiliates when using different cased address
### Expected Behavior

I sometimes use different casing for my validator group addresses.

When I run `celocli validatorgroup:show 0xd4e068fb4467816c768fa3460ce9d5ae5642de37` it should show the same affiliate list as `celocli validatorgroup:show 0xD4e068fB4467816C768fa3460ce9d5Ae5642dE37`

### Current Behavior

What is the current behavior?

When I run `celocli validatorgroup:show 0xd4e068fb4467816c768fa3460ce9d5ae5642de37`, it does not show any affiliates even if that group has affiliates.

It looks like the code isn't using `eqAddress` where it should be using it on this line: https://github.com/celo-org/celo-monorepo/blob/95e50a672485178a03228b38032d9b5833d9c135/packages/contractkit/src/wrappers/Validators.ts#L248

mrsmkl:
### Description

Changing to use `eqAddress` instead of normal comparison in places found so far.

### Tested

Only manual testing.

### Related issues

- Fixes #2714


nategraf (approved):
Nice! 🙌

PR #2766, opened 2020-02-13
Fix flaky e2e governance test

jfoutts-celo:
### Description

Fix flaky governance_test: when the validator set is changing > should update the validator scores at the end of each epoch.  The flakiness was due to the test assuming validator uptime was always one.  Fixed by adding a `calculateUptime` that replicates the celo-blockchain `updateUptime` tallying.

### Tested

Tested no failures in 50 runs (with the test `it.only()`): 
```
for ((i=0; i<50; i++));
  do ./ci_test_governance.sh local ~/celo-blockchain 2>&1 | tee test4/run$i;
done
```

### Related issues

Part of epic https://github.com/celo-org/celo-blockchain/issues/847 Builds reliably green


mcortesi (approved):
Awesome job!!!!

PR #2767, opened 2020-02-13
Adjust docs for Phase 3 launch

nategraf:
### Description

With the introduction of Geth 1.9, a few docs need to be updated. Also I've added a short section on hotfixes and added governance to the command line docs.


clrblmt:
LGTM

PR #2771, opened 2020-02-13
Atomic validator key update

### LINKED ISSUE #2644: Validators should be able to rotate ECDSA and BLS keys together atomically
Currently if a validator is working to rotate their keys and they update one, then the epoch changes before they update the other, they may get into a state where they are unable to participate in consensus. In order to prevent this scenario, the Account smart contract should support an atomic rotation of both.

mrsmkl:
### Description

Added a function to accounts for atomically changing validator signer and keys.

### Tested

Unit tests, contractkit test, CLI test, also changed governance e2e test to use this method.

### Other changes

Changed governance fallback function to fail if data is not empty.

### Related issues

- Fixes #2644


packages/cli/src/commands/account/authorize.ts
```diff
@@ -22,6 +22,14 @@ export default class Authorize extends BaseCommand {
...
+      description:
+        'The BLS public key that the validator is using for consensus, should pass proof of possession. 96 bytes.',
+    }),
+    blsPop: Flags.blsProofOfPossession({
+      description:
+        'The BLS public key proof-of-possession, which consists of a signature on the account address. 48 bytes.',
+    }),
   }
 
   static args = []
```
asaj, line 35: Can you add an example of rotating both keys to `examples`?

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -604,6 +604,36 @@ contract Validators is
...
+   * @param account The address under which the validator is registered.
+   * @param signer The address which the validator is using to sign consensus messages.
+   * @param ecdsaPublicKey The ECDSA public key corresponding to `signer`.
+   * @param blsPublicKey The BLS public key that the validator is using for consensus, should pass
+   *   proof of possession. 96 bytes.
+   * @param blsPop The BLS public key proof-of-possession, which consists of a signature on the
+   *   account address. 48 bytes.
+   * @return True upon success.
+   */
+  function updateKeys(
```
asaj: nit: Consider `updatePublicKeys` for consistency

asaj (approved):

PR #2776, opened 2020-02-13
[dev-utils] Avoid devchain datadir corruption between test runs

### LINKED ISSUE #2761: Following tests runs against a corrupt instance of the ganache devchain
### Expected Behavior

Every batch of tests has to run against a devchain as if it was just migrated

### Current Behavior

Every test runs against the same devchain. If one of the test overrides something, in future runs, will use a corrupt chain (it forces you to run `yarn test:reset`)

gastonponti:
### Description

Copy the devchain for every test run. This way we avoid problems with tests that corrupt the `chain`.

### Tested

Tested manually running the `cli` and `contractKit` tests

### Other changes

Added the possibility to create the chain as a `tar.gz`. This allows in the future to keep in a storage different kinds of chains to be tested.

### Related issues

- Fixes #2761 

### Backwards compatibility

Everything OK


mcortesi (approved):
LGTM!

PR #2782, opened 2020-02-14
[Wallet]: Invite text message link support for iOS

### LINKED ISSUE #1512: Wallet users SBAT invite another user with a link that works on both iOS and android

### LINKED ISSUE #1729: Invite text message deeplinking SBAT work on iOS
### Expected Behavior

deeplinking works with iOS and android devices

### Current Behavior

deeplinking presently works with android only

i1skn:
### Some context 
I've closed #1983 and created this one to start fresh. Most of the work in this PR was already done by @martinvol and my task is just to get this feature shipped.

### TODO
- [x] Test all 4 cases (app installed/not installed on Android/iOS) on real devices

### Description

This PR makes link in the invitation text to navigate to Play **or** App Store depend on which platform it was opened (Android or iOS). 
**Only on Android** this link will store `inviteCode` in the `referrer` at PlayStore, so later on `Enter Invite Code` screen user would see the code pre-filled.

### Tested

Ran on Android & iOS device

### Related issues

- Fixes #1512

### Backwards compatibility

Yes


packages/mobile/src/firebase/dynamicLinks.ts
```diff
@@ -0,0 +1,36 @@
...
+  appStoreUrl,
+  bundleId,
+}: {
+  link: string
+  playStoreUrl: string
+  appStoreUrl: string
+  bundleId: string
+}): Promise<string> {
+  try {
+    const firebaseLink = new firebase.links.DynamicLink(link, 'https://celol.page.link')
```
jeanregisser, line 18: Is celo**l** intentional?

packages/mobile/src/firebase/dynamicLinks.ts
```diff
@@ -0,0 +1,36 @@
...
+    firebaseLink.android.setPackageName(bundleId)
+
+    firebaseLink.ios.setFallbackUrl(appStoreUrl)
+    firebaseLink.ios.setBundleId(bundleId)
+
+    // Please note other parameter than UNGUESSABLE is unsafe
+    const shortUrl = await firebase.links().createShortDynamicLink(firebaseLink, 'UNGUESSABLE')
+
+    // It is NOT recommended to shorten this link because it
+    // can break deep links
```
jeanregisser, line 30: Do you have more info on this?

packages/mobile/__mocks__/react-native-firebase.ts
```diff
@@ -1 +1,30 @@
-export default {}
+let toFail = false
+
+function links() {
+  return {
+    createShortDynamicLink: jest.fn(async () => {
+      if (toFail) {
+        throw new Error('test error')
+      }
```
jeanregisser: nit: we're using private state to test a failure case. A slightly nicer approach would be to expose a `createShortDynamicLink` mock that doesn't change and let the test itself set the kind of return value it wants for it.

jeanregisser (commented):
Nice work! 👍 
If I understand this correctly, we're not yet using the full potential of firebase dynamic links, right?
I mean we're not using a deep link which would contain the invite code and prefill it directly, working for both iOS and Android.

packages/mobile/src/firebase/dynamicLinks.ts
i1skn: I've got the hint from @martinvol that using other shortener (like bit.ly) on on top of already shortened link is a bad idea. I did not investigate it deeply why, but we do not have that many reasons to use (text now fits in a SMS just fine). Comment was left by @martinvol.

packages/mobile/src/firebase/dynamicLinks.ts
i1skn: Yes, because `celo.page.link` is used in other firebase project according to @martinvol. But anyway this is temporary - we would use our own domain for this (I just need to figure out who can add TXT DNS record to our domain).

i1skn:
> If I understand this correctly, we're not yet using the full potential of firebase dynamic links, right?
> I mean we're not using a deep link which would contain the invite code and prefill it directly, working for both iOS and Android.

Yes. But the reason is that Dynamic LInks require universal links in place. Here https://firebase.google.com/docs/dynamic-links/create-manually#parameters we can see, that `link` should start with `http` or `https`. So, we can not use `celo://` here and to use the whole potential we would need to setup universal links and think how we want to handle invitation page in web.

P.S.: I did not know that and had to remove handling of `celo://` deeplinks after implementing it first, see here: https://github.com/celo-org/celo-monorepo/pull/2782/commits/e298831c136a2a8d2f625988fe9d98febefba549

packages/mobile/__mocks__/react-native-firebase.ts
i1skn: This sound like a good idea, will change it

packages/mobile/src/firebase/dynamicLinks.ts
jeanregisser: Ok, I thought it was related to an option in the dynamic link generation where it could output a shorter link.

jeanregisser (approved):
Yes it would be great if both the web invite and and the app used dynamic links with universal links.

Anyway this is looking good for now :+1:

PR #2784, opened 2020-02-14
change capitalization of DApp to dApp

aaronmgdr:
according to vanessa the preferred style is now lowercase `d`  uppercase `A`  like how its `cUSD`


jmrossy (commented):
We should probably be consistent with the rest of the industry though, which does an uppercase D

gastonponti (dismissed):

aaronmgdr:
closing as this points to baklava not master,

PR #2786, opened 2020-02-14
Use fixidity in "FractionMulExp" precompile

### LINKED ISSUE #1366: Use fixidity in "FactionMulExp" precompile
### Expected Behavior

We don't use fractions anymore, but fixidity. Change the precompile to work with fixidity values.

Usage example in `StableToken`
```solidity
    (numerator, denominator) = fractionMulExp(
      inflationState.factor.unwrap(),
      FixidityLib.fixed1().unwrap(),
      inflationState.rate.unwrap(),
      FixidityLib.fixed1().unwrap(),
      timesToApplyInflation,
      decimals_
    );
```

Tasks:
 - [ ] Change ganache precompile
 - [ ] Change celo-blockchain precompile
 - [x] Change use of fractionMulExp in solidity smart contracts

jfoutts-celo:
### Description

Add a mulExp wrapper around the fractionMulExp precompile to accept and return unwrapped FixidityLib.Fraction.

### Tested

Unit tests.

### Related issues

- Fixes #1366

### Backwards compatibility

Yes.


jfoutts-celo:
We may not need to update the precompiles:
- https://github.com/celo-org/ethereumjs-vm/blob/master/lib/evm/precompiles/fc-fractionmulexp.js#L20
- https://github.com/celo-org/celo-blockchain/blob/master/core/vm/contracts.go#L503

I think it looks pretty good.  There's not much efficiency lost to passing the arguments as currently done, and it's a little more extensible.

jfoutts-celo:
We need the current `fractionMulExp` interface (at least the digits argument) to support https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/stability/StableToken.sol#L42

packages/protocol/test/stability/stabletoken.ts
```diff
@@ -328,12 +328,30 @@ contract('StableToken', (accounts: string[]) => {
       const [numerator, denominator] = await stableToken.fractionMulExp.call(1, 2, 3, 4, 5, 6)
       assert.equal(numerator.toNumber(), 118652)
       assert.equal(denominator.toNumber(), 1000000)
+
+      const value = await stableToken.mulExp.call(toFixed(1 / 2), toFixed(3 / 4), 5)
```
mcortesi, line 332: you end up using `floats` here.... maybe this can introduce some flakyness?

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -358,17 +358,7 @@ contract Validators is
...
-    (numerator, denominator) = fractionMulExp(
-      FixidityLib.fixed1().unwrap(),
-      FixidityLib.fixed1().unwrap(),
-      uptime,
-      FixidityLib.fixed1().unwrap(),
-      validatorScoreParameters.exponent,
-      18
-    );
-    return FixidityLib.newFixedFraction(numerator, denominator).unwrap();
+    return mulExp(FixidityLib.fixed1().unwrap(), uptime, validatorScoreParameters.exponent);
```
mcortesi, line 358: is this the only place where we use it?

packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -16,6 +18,28 @@ contract UsingPrecompiles {
...
+   * @param a First unwrapped Fraction.
+   * @param b The unwrapped Fraction to be exponentiated.
+   * @param exponent Exponent to raise b to.
+   * @return The computed quantity as an unwrapped Fraction.
+   */
+  function mulExp(uint256 a, uint256 b, uint256 exponent) public view returns (uint256) {
+    uint256 denominator = FixidityLib.fixed1().unwrap();
+    uint256 returnNumerator;
+    uint256 returnDenominator;
+    (returnNumerator, returnDenominator) = fractionMulExp(
```
mcortesi, line 32: Ideally what we want is for the precompile to actually use fixidity numbers and thus receive only 3 parameters. But this is a good first step.

The problem is that to make that change, we need to:

 1. change geth
 2. change ganache/ethereumvm

But it would be great to do this before mainnet.... after that it will be harder

mcortesi (commented):
Left some comments

packages/protocol/contracts/governance/Validators.sol
jfoutts-celo: There's only one other place we use it, but it's in `StableToken` which has a parameter:
``` 
 /**
   * @return The number of decimal places to which StableToken is divisible.
   */
 uint8 internal decimals_;
```
So unless we want to remove the `decimals_` parameter of StableToken, and thus make all StableToken divisible to the same number of digits as Fixidity fractions, we need to keep the digits argument to `fractionMulExp`.

packages/protocol/test/stability/stabletoken.ts
jfoutts-celo: There's lots of other examples in the tests of passing the return of `toFixed()` (which returns `BigNumber`) to smart contract calls.  Pretty sure this works and `toFixed(x).toFixed()` is not necessary.

packages/protocol/contracts/common/UsingPrecompiles.sol
jfoutts-celo: How important is saving 64 bytes of argument packing and unpacking?  If we decide to keep `StableToken.digits_`, then it might be best to have two precompiles.

packages/protocol/test/stability/stabletoken.ts
jfoutts-celo: Seems `toFixed(x).toFixed()` is needed in the migrations, though.

packages/protocol/contracts/common/UsingPrecompiles.sol
asaj: +1, I was imagining we would change the precompiles here as well, I'm not sure of the value of this PR in a vacuum

PR #2793, opened 2020-02-16
Clarify that `validatorgroup` uses zero based indices 

keefertaylor:
### Description

Clarify that validators are listed inside of the group with a zero based index, instead of a 1 based index. 

I know that generally technically minded folks assume a zero based index but other tools in the Celo ecosystem, such as [BlockScout's group list](https://baklava-blockscout.celo-testnet.org/address/0xd5c181972763b082d4d6ca84662878674c0bd35c/celo) use 1 based indices. This leads me to run commands to test the numbering scheme used every time I have to reorder my validator group. This seems like a sharp edge that we can easily fix with documentation. 

### Tested

Relying on CI; no behavioral changes are introduced. 

### Other changes

N/A

### Related issues

N/A

### Backwards compatibility

Fully backwards compatible, this is only a docs update.


keefertaylor:
Meta: The [Contributing Guidelines](https://docs.celo.org/community/contributing#submitting-prs) says to make sure that a core team member is assigned, but it looks like you folks have adopted CODEOWNERS so reviews should get farmed out automatically.  You may want to update your documentation (or I can if you give me a go ahead :) )

asaj (approved):

PR #2794, opened 2020-02-17
[M05] Missing counter addition

mrsmkl:
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/268


nambrot:
Can we set the number to something higher? Is 100 ok?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -550,6 +575,23 @@ contract Attestations is
     return (matches, new address[](totalAddresses));
   }
 
+  function getPossibleAttestationIssuers() public view returns (address[] memory, uint256) {
```
nambrot: Was this extraction necessary? It also contains an error as it makes it possible to add the same issuer more than once?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -562,35 +604,29 @@ contract Attestations is
...
-        continue;
-      }
-
       currentIndex = currentIndex.add(1);
       attestation.status = AttestationStatus.Incomplete;
       attestation.blockNumber = unselectedRequest.blockNumber;
       attestation.attestationRequestFeeToken = unselectedRequest.attestationRequestFeeToken;
       state.selectedIssuers.push(issuer);
 
+      issuersLength = issuersLength.sub(1);
```
nambrot, line 644: I don't understand this logic, it just pops it out from the end of the list?

nambrot (changes requested):

packages/protocol/contracts/identity/Attestations.sol
mrsmkl: Yeah, it's more efficient to not extract everything at start. Added a test to check that validator cannot be selected twice.

packages/protocol/contracts/identity/Attestations.sol
mrsmkl: Added comment, hopefully that helps.

nambrot (approved):

PR #2796, opened 2020-02-17
[L02] Inconsistent quorum calculation

mrsmkl:
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/302


packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -83,7 +83,8 @@ const INTRINSIC_TX_GAS_COST = 21000
 // Additional intrinsic gas for a transaction with fee currency specified
 const ADDITIONAL_INTRINSIC_TX_GAS_COST = 50000
 
-const stableTokenTransferGasCost = 20653
+// const stableTokenTransferGasCost = 20653
```
asaj: Why is this different?

packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -234,4 +234,21 @@ contract UsingPrecompiles {
     }
     return x;
   }
+
+  /**
+   * @notice Returns the minimum number of required signers for a given block number.
+   * @dev Computed in Geth as int(math.Ceil(float64(2*valSet.Size()) / 3))
```
asaj: nit: celo-blockchain

packages/celotool/src/e2e-tests/transfer_tests.ts
mrsmkl: This sometimes changes, haven't found the reason yet...

packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -234,4 +234,21 @@ contract UsingPrecompiles {
...
+   */
+  function minQuorumSize(uint256 blockNumber) internal view returns (uint256) {
+    return numberValidatorsInSet(blockNumber).mul(2).add(2).div(3);
+  }
+
+  /**
+   * @notice Computes byzantine quorum from current validator set size
+   * @return Byzantine quorum of validators.
+   */
+  function byzantineQuorumValidatorsInCurrentSet() public view returns (uint256) {
```
nategraf: Could you rename one of these two to be match whether `min` or `byzantine` is used at the start?

packages/celotool/src/e2e-tests/transfer_tests.ts
nategraf: +1 to this. We need to make sure we know why gas costs change.

mrsmkl: It was because a public method was added to UsingPrecompiles, so the routing code changed.

asaj (approved):

PR #2797, opened 2020-02-17
[C01] Funds of the Escrow contract can be locked by anyone

mrsmkl:
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/259


mcortesi (approved):
I would just ask @nambrot if payment is not something that can/should be repeated

mrsmkl:
Ok. Looking from some code, the `paymentId` is randomly generated address. But perhaps it will be changed later on.

nategraf (approved):
Same comment as Mariano, but overall it looks good!

nambrot (approved):
Indeed, a payment ID should never be repeated

PR #2798, opened 2020-02-17
Fix yarn install cache miss on CircleCI

jeanregisser:
### Description

Fix `yarn install` cache miss on CircleCI.

In https://app.circleci.com/jobs/github/celo-org/celo-monorepo/125242 I noticed the `yarn install` step was ignoring the cached `node_modules`, though `yarn.lock` and patches had not changed.

Digging deeper I found out it was caused by
`node_modules/.yarn-integrity` generation not being stable for a given `yarn.lock` through different runs of `yarn install`.

I diff'ed the generated `.yarn-integrity` file between the cached node_modules and the one after running `yarn install` (which should be the same given the `yarn.lock` file was the same).

<details>
  <summary>diff details</summary>

```diff
--- ../app/node_modules/.yarn-integrity	2020-02-17 11:12:42.677301222 +0000
+++ node_modules/.yarn-integrity	2020-02-13 21:32:53.000000000 +0000
@@ -35,6 +35,7 @@
     "@0x/sol-trace@^2.0.16",
     "@0x/subproviders@^4.1.0",
     "@0x/subproviders@^5.0.0",
+    "@0x/subproviders@^5.0.0",
     "@babel/core@^7.6.2",
     "@babel/core@^7.6.2",
     "@babel/runtime@^7.6.2",
@@ -170,7 +171,6 @@
     "@types/enzyme@^3.9.1",
     "@types/ethereumjs-util@^5.2.0",
     "@types/ethereumjs-util@^5.2.0",
-    "@types/ethereumjs-util@^5.2.0",
     "@types/express-rate-limit@2.9.3",
     "@types/express@4.17.0",
     "@types/express@^4.16.1",
@@ -226,7 +226,6 @@
     "@types/supertest@^2.0.7",
     "@types/targz@^1.0.0",
     "@types/tmp@^0.1.0",
-    "@types/tmp@^0.1.0",
     "@types/twilio@^0.0.10",
     "@types/underscore@^1.8.8",
     "@types/utf8@^2.1.6",
@@ -671,7 +670,6 @@
     "targz@^1.0.1",
     "tls@^0.0.1",
     "tmp@^0.1.0",
-    "tmp@^0.1.0",
     "truffle-artifactor@^3.0.7",
     "truffle-plugin-blockscout-verify@git+https://github.com/celo-org/truffle-plugin-blockscout-verify",
     "truffle-resolver@^4.0.4",
@@ -7098,7 +7096,11 @@
       "build/scrypt_lib.target.mk",
       "build/scrypt_wrapper.target.mk",
       "scrypt/scrypt-1.2.0",
-      "scrypt/scrypt-1.2.0/config.log"
+      "scrypt/scrypt-1.2.0/Makefile",
+      "scrypt/scrypt-1.2.0/config.h",
+      "scrypt/scrypt-1.2.0/config.log",
+      "scrypt/scrypt-1.2.0/config.status",
+      "scrypt/scrypt-1.2.0/stamp-h1"
     ],
     "usb@1.6.0": [
       "build",
@@ -7140,6 +7142,7 @@
       "builderror.log"
     ],
     "nodent-runtime@3.2.1": [
+      "dist",
       "dist/index.js"
     ],
     "sleep@6.1.0": [
@@ -7311,10 +7314,24 @@
     "@celo/client@0.0.266": [
       "Empty.m",
       "build/bin",
+      "build/bin/Geth.framework",
+      "build/bin/Geth.framework/Geth",
+      "build/bin/Geth.framework/Headers",
+      "build/bin/Geth.framework/Modules",
+      "build/bin/Geth.framework/Resources",
       "build/bin/Geth.framework/Versions",
       "build/bin/Geth.framework/Versions/A",
+      "build/bin/Geth.framework/Versions/A/Geth",
       "build/bin/Geth.framework/Versions/A/Headers",
+      "build/bin/Geth.framework/Versions/A/Headers/Geth.h",
+      "build/bin/Geth.framework/Versions/A/Headers/Geth.objc.h",
+      "build/bin/Geth.framework/Versions/A/Headers/Universe.objc.h",
+      "build/bin/Geth.framework/Versions/A/Headers/ref.h",
+      "build/bin/Geth.framework/Versions/A/Modules",
+      "build/bin/Geth.framework/Versions/A/Modules/module.modulemap",
       "build/bin/Geth.framework/Versions/A/Resources",
+      "build/bin/Geth.framework/Versions/A/Resources/Info.plist",
+      "build/bin/Geth.framework/Versions/Current",
       "libGeth.a"
     ],
     "@sentry/cli@1.48.0": [
```
</details>

Which indeed has slight variations.

The notable thing is that we see duplicated entries for the same package. This has been reported upstream in https://github.com/yarnpkg/yarn/issues/6703

This causes `yarn install` to redo all the work when deps haven't changed.

As a workaround, we now skip the install step when we know the deps haven't changed.

### Tested

It now skips `yarn install` when we have cached node_modules for the given `yarn.lock` (and patches).

### Other changes

None

### Related issues

None

### Backwards compatibility

Yes


.circleci/config.yml
```diff
@@ -61,11 +61,31 @@ commands:
             - node_modules-v3-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "~/.tmp/checksumfiles/patches" }}
       - run:
           name: Extract cached node_modules (if present)
-          command: if [ -e ~/.tmp/node_modules.tgz ]; then tar -xf ~/.tmp/node_modules.tgz; fi
+          command: |
+            if [ -e ~/.tmp/node_modules.tgz ]; then 
+              tar -xf ~/.tmp/node_modules.tgz
+            else
+              # Dependencies have changed, create beacon file used by the install and check licenses steps
+              touch ~/.tmp/yarn_deps_have_changed
```
mcortesi, line 71: does this get the case where `yarn.lock` was not comitted, but package.json changed?

jeanregisser: Ah good point! This change will miss the uncommitted `yarn.lock` changes.

In that case, we do need to take the checksum of `package.json` into account for the cache key, so that case is handled too.

jeanregisser: Issue addressed in https://github.com/celo-org/celo-monorepo/pull/2798/commits/74da4ef52ebc6a3eee1b7b21236c398d7776c9a8

The tradeoff is that the cache will be directly ignored if any of the `package.json` files we have change, even if no dependencies were touched.
But uncommitted changes to `yarn.lock` will be flagged directly without waiting for the next fresh install.

mcortesi: nice

mcortesi (approved):
looks good!

PR #2800, opened 2020-02-17
Use Geth1.9 on e2e tests

mcortesi:
### Description

Changes on e2e configuration to adapt to 1.9

### Tested

e2e tests


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ android-defaults: &android-defaults
 e2e-defaults: &e2e-defaults
   <<: *defaults
   docker:
-    - image: celohq/circleci
+    - image: celohq/circleci:geth1.9
```
tkporter, line 37: Is this needed? If so should we change the default tag to be the 1.9 one?

packages/protocol/contracts/common/Proxy.sol
```diff
@@ -41,7 +41,8 @@ contract Proxy {
 
     // Avoid checking if address is a contract or executing delegated call when
     // implementation address is 0x0
-    if (implementationAddress == address(0)) return;
+    require(implementationAddress != address(0), "No Implementation set");
+    // if (implementationAddress == address(0)) return;
```
tkporter: Probably can remove this

tkporter (approved):

PR #2803, opened 2020-02-17
SDK Docs for 2/21 Release

nityas:
### Description

batched documentation changes

### Tested

n/a

### Other changes

n/a

### Related issues

- Fixes #[issue number here]


packages/docs/developer-resources/contractkit/notes-web3-with-contractkit.md
```diff
@@ -0,0 +1,38 @@
...
+
+web3.eth.getBalance(someAddress)
+```
+
+or
+
+```ts
+const web3 = kit.web3
+
+web3.eth.getBlock("earliest")
```
mcortesi: let's use `latest` not `earliest`

packages/docs/developer-resources/contractkit/notes-web3-with-contractkit.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+will work the same way
+
+## Web3 limitations
+
+As you have read in our guide, Celo uses some extra fields: `feeCurrency`, `gatewayFeeRecipient` and `gatewayFee`, that among other things allows you yo pay gas with ERC20 Tokens. These fields are expected by the node.
+
+To facilitate the life of every developer, we decided to wrap the `Provider` set in the `Web3` instance, and add our way to handle local signing using these new fields. Similar to what *Metamask* does, we intercept every transaction and perform a local signing when required.
```
mcortesi: maybe mention this is called `CeloProvider` and that can be used without `kit`

mcortesi (approved):

PR #2804, opened 2020-02-18
[Pilot] Add Earn screen skeleton to wallet

annakaz:
### Description

Add skeleton of Earn screen to wallet for upcoming Earn pilot. Allows users to enter a Figure Eight ID (which will be precreated for pilot participants) and displays amount earned. 

Note that the Figure Eight iFrame will be added in a future PR so this PR alone will not allow users to complete work, just to login with their ID and view the amount earned. 

### Tested

With celo-org-mobile-pilot firebase


annakaz:
No review needed on this PR just making it easier to keep track of the Earn pilot build

PR #2805, opened 2020-02-18
[M02] Functions in StableToken do not emit Transfer event

mrsmkl:
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/265


mcortesi (dismissed):
Nice!

asaj:
> discussion on the issue about a different approach

+1

packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -527,4 +530,48 @@ contract StableToken is
     return true;
   }
 
+  mapping(address => uint256) reserved;
```
asaj: I'm worried about the gas implications of the extra writes that using this mapping requires. What do you think about the following implementation?

```
  function debitGas(address from, uint256 value)
    external
    onlyVm
    updateInflationFactor
    onlyWhenNotFrozen
  {
    uint256 units = _valueToUnits(inflationState.factor, value);
    totalSupply_ = totalSupply_.sub(units);
    balances[from] = balances[from].sub(units);
  }

function payGas(address from, address feeRecipient, address gatewayFeeRecipient, uint256 gas, uint256 gasPrice, uint256 gasUsed, uint256 gatewayFee) external onlyVm {
  uint256 gasValue = gas.mul(gasPrice);
  uint256 units = _valueToUnits(gasValue.add(gatewayFee));
  uint256 unitsUsed = _valueToUnits(gasValue);
  uint256 unitsGateway = _valueToUnits(gatewayFee);
  totalSupply_ = totalSupply_.add(units);
  balances[feeRecipient] = balances[feeRecipient].add(unitsUsed);
  balances[gatewayFeeRecipient] = balances[gatewayFeeRecipient].add(unitsGateway);
  balances[from] = balances[from].add(units.sub(unitsUsed).sub(unitsGateway));
  emit Transfer(from, feeRecipient, gasValue);
  emit Transfer(from, gatewayFeeRecipient, gatewayFee);
}
```

asaj: I guess I should say, the added cost of the additional writes to the storage trie, costs that of course wouldn't directly be passed down to the user since the gas cost to the sum of these operations is hardcoded in BlockChainParameters.

To be fair, the storage trie changes for `reserved` would never actually be written to disk, since any changes are reverted by the end of tx processing.

asaj: cc @kevjue to get your thoughts here

kevjue: Is there a corresponding PR in the geth code that uses these new functions?  It would help to get more context.

mrsmkl: It's here: https://github.com/celo-org/celo-blockchain/pull/882

kevjue: I like @asaj 's proposal (perhaps rename `debitGas` to `debitGasFee` and `payGas` to `payGasFee`.  That proposal seems to be much more explicit for the purpose of what's intended as opposed to @mrsmkl 's.

One thing I do like about @mrsmkl 's implementation is that it's explicit about the debited gas.  E.g. it was initially confusing to me in @asaj 's when the totalSupply was decreased in `debitGas`.  So if you go w/ @asaj 's then I recommend adding comments about that.

packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -519,4 +522,48 @@ contract StableToken is
...
+  }
+
+  /**
+   * @notice Refund balance after making payments for gas in this StableToken currency.
+   * @param to The account to credit balance to
+   * @param value The amount of balance to refund
+   * @dev We can assume that the inflation factor is up to date as `reserveGas`
+   * will have been called in the same transaction
+   */
+  function refundGas(address to, uint256 value) external onlyVm onlyWhenNotFrozen {
```
asaj: One issue with refunding and crediting in separate functions is that there's no guarantee that the transaction does not change the total supply due to rounding errors.

I don't think this is critical, but would be curious to get your thoughts on the tradeoffs between this and the approach I suggested

packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -489,6 +490,7 @@ contract StableToken is
     uint256 units = _valueToUnits(inflationState.factor, value);
     totalSupply_ = totalSupply_.sub(units);
     balances[from] = balances[from].sub(units);
+    emit Transfer(from, address(0), value);
```
asaj: Can we delete `debitFrom`/`creditTo`?

packages/protocol/contracts/stability/StableToken.sol
mrsmkl: About rounding error, the current version is the same as the previous one (actually there isn't much difference now).
Your approach would move some logic from celo-blockchain to the contracts, I think that is good, but would prefer that same would then work for gold token too.

packages/protocol/contracts/stability/StableToken.sol
mrsmkl: Yeah, going to cleanup the other PR too.

packages/protocol/contracts/stability/StableToken.sol
asaj: Good point. I'm happy with whatever you decide is best

packages/protocol/contracts/stability/StableToken.sol
mcortesi: A few comments...

1. We need to `reserve` token for gas. So changing `totalSupply` doesn't make sense.
2. The protocol is very strict that for every `reserve` there's a `credit`. If we use that information we don't actually need to save anything on the contract. At most, you could save a `reservedValue` but since that information can be given to you at the `credit` call. It makes no sense to write to the trie storage.
3. Even if writes to trie storage won't be persisted, it's surely cheaper (computationally wise) to let the node (and not the contract carry the state)

mcortesi: So... I'd go with @asaj approach...

I'm a bit uncertain about the names... `debit` and `pay` sound like synonims to me. I would change `debit` for something like `reserveForGas` `lockForGas`... but not totally sure. And maybe chage the `pay` to something like `applyGasPayment`

Also, i would change the signature of the second function to

`applyGasPayment(addresses[3], amounts[3])` where addresses[0] does not emit a transfer since it's the `from`
or 
`applyGasPayment(address from, address feeRecipient, address gatewayFee, uint32 amountFrom, amountFeeR, amountGat..)`

basically, i wouldn't make the contract make the gas price computation

asaj: > We need to reserve token for gas. So changing totalSupply doesn't make sense.
This was done originally to account for rounding errors. If we have an approach that is resilient to this by calculating the exact same total of units on both ends of the tx, changing total supply won't be necessary. I think this is another argument for the proposal above.

RE changing the second function signature, +1 to either of @mcortesi 's suggestions.

Finally, on naming, suggest `debitGasFee` and `creditGasFees`, which at the very least is consistent with our current naming scheme.

mrsmkl: Can somebody check that `units.sub(unitsUsed).sub(unitsGateway)` won't fail because of rounding errors?

mcortesi: it won't

mcortesi: Just figured out that we need the `totalSupply` anyway.

If the tx call `setInflactionFactor()` then `valueToUnits()` result will change; since inflation factor might have changed.

packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -461,47 +462,74 @@ contract StableToken is
...
 
   /**
-   * @notice Deduct balance for making payments for gas in this StableToken currency.
-   * @param from The account to debit balance from
-   * @param value The value of balance to debit
+   * @notice Transfers StableToken from one address to another
+   * @param to The address to transfer StableToken to.
+   * @param value The amount of StableToken to be transferred.
+   */
+  function _transfer(address to, uint256 value) internal returns (bool) {
```
mcortesi, line 469: who uses this function? why is it changed on this PR?

packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -461,47 +462,74 @@ contract StableToken is
...
+    address feeRecipient,
+    address gatewayFeeRecipient,
+    address communityFund,
+    uint256 refund,
+    uint256 tipTxFee,
+    uint256 gatewayFee,
+    uint256 baseTxFee
+  ) external onlyVm onlyWhenNotFrozen {
+    uint256 units = _valueToUnits(inflationState.factor, refund);
+    balances[from] = balances[from].add(units);
```
mcortesi, line 517: this should be the last line... since creditGas does not decrement the balance...

packages/protocol/contracts/stability/StableToken.sol
mrsmkl: Just some weirdness with diffs, there are no real changes.

packages/protocol/contracts/stability/StableToken.sol
mrsmkl: The refund is computed in the celo-blockchain now.

mcortesi: nevermind... i was reading the previous impl in my head

mcortesi (approved):

PR #2808, opened 2020-02-18
[C02] Any owner of the Multisig can execute unconfirmed transactions

mcortesi:
### Related issues

- Fixes celo-org/celo-labs#261


asaj:
I don't think this compiles

asaj (approved):

PR #2827, opened 2020-02-19
Validator proof handshake e2e test

tkporter:
### Description

This adds a new e2e test that tests the validator proof handshake: https://github.com/celo-org/celo-blockchain/pull/873

I changed e2e tests to allow random node keys and to set the max peer count.

Both tests do very similar things, the first is for a non-proxied validator and the second is for a proxied validator. They both test that a validator with a new enode URL with an accurate val enode table will become a `ValidatorPurpose` peer of another validator (whose maxpeers is 0) as a result of the handshake.

### Tested

Ran the e2e test locally

### Other changes

n/a

### Related issues

n/a

### Backwards compatibility

Yes


packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -31,6 +32,8 @@ import { ensure0x, spawnCmd, spawnCmdWithExitOnFailure } from '../lib/utils'
 
 const MonorepoRoot = resolvePath(joinPath(__dirname, '../..', '../..'))
```
nategraf: Mega-nit: The capitalization on this variable is bothering me
```suggestion
const monorepoRoot = resolvePath(joinPath(__dirname, '../..', '../..'))
```

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -339,3 +284,111 @@ export function getContext(gethConfig: GethRunConfig, verbose: boolean = verbose
...
+export async function jsonRpc(web3: Web3, method: string, params: any[] = []): Promise<any> {
+  return new Promise((resolve, reject) => {
+    web3.currentProvider.send(
+      {
+        jsonrpc: '2.0',
+        method,
+        params,
+        // salt id generation, milliseconds might not be
+        // enough to generate unique ids
+        id: new Date().getTime() + Math.floor(Math.random() * (1 + 100 - 1)),
```
nategraf: It's important to shift the timestamp left two digits before adding the randomness, or you actually _increase_ the chance for collisions because a value could collide with another value up to 199ms later.
```suggestion
        id: (new Date().getTime() * 100) + Math.floor(Math.random() * 100),
```
Note that this method gets close to maxing out the size of `number`

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -171,31 +174,17 @@ export function getHooks(gethConfig: GethRunConfig) {
 }
 
 export function getContext(gethConfig: GethRunConfig, verbose: boolean = verboseOutput) {
-  const mnemonic =
-    'jazz ripple brown cloth door bridge pen danger deer thumb cable prepare negative library vast'
   const validatorInstances = gethConfig.instances.filter((x: any) => x.validating)
```
nategraf, line 196: Nit: The following would clarify that only the count is needed here.
```suggestion
  const numValidators = gethConfig.instances.filter((x: any) => x.validating).length
```

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -339,3 +284,111 @@ export function getContext(gethConfig: GethRunConfig, verbose: boolean = verbose
     hooks: { before, after, restart, gethBinaryPath },
   }
 }
+
+// Modifies gethConfig to properly configure proxies and proxied instances
+export function configureInstances(gethConfig: GethRunConfig, bootnodeEnode?: string) {
```
nategraf: A less generic name would be helpful.
```suggestion
export function setProxyConfigurations(gethConfig: GethRunConfig, bootnodeEnode?: string) {
```

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -339,3 +284,111 @@ export function getContext(gethConfig: GethRunConfig, verbose: boolean = verbose
...
+}
+
+export async function restartInstance(
+  gethConfig: GethRunConfig,
+  instance: GethInstanceConfig,
+  removeNodeKey: boolean = false,
+  verbose: boolean = false
+) {
+  await killInstance(instance)
+  await restoreDatadir(gethConfig.runPath, instance)
```
nategraf, line 382: Possibly a race condition where the data directory is restored before the instance is dead. Something like:
```ts
while ((await spawnCmd('kill', ['-0', instance.pid.toString()]) === 0) {
  await sleep(0.25)
}
```
Could help.

packages/celotool/src/lib/geth.ts
```diff
@@ -859,6 +859,10 @@ export async function startGeth(
   const privateKey = instance.privateKey || ''
   const lightserv = instance.lightserv || false
   const etherbase = instance.etherbase || ''
+  // Allow maxPeers to be zero
+  const maxPeers = instance.maxPeers !== undefined ? instance.maxPeers : gethConfig.instances.length
```
nategraf: Why default to `gethConfig.instances.length`? That seems low and it would be better, in my opinion to default to not providing the flag.

packages/celotool/src/lib/geth.ts
```diff
@@ -859,6 +859,10 @@ export async function startGeth(
...
+  // Default to true
+  const setNodeKey = instance.setNodeKey !== undefined ? instance.setNodeKey : true
```
nategraf: You should be able to use the [null coalescing operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing) (`??`) here.

nategraf: Setting the node key by default is actually a relic that should have been removed already. We don't recommend it in the official docs, and we shouldn't do it here by default unless there is a reason I am not aware of.

packages/celotool/src/lib/geth.ts
```diff
@@ -1135,9 +1144,12 @@ export async function connectPeers(instances: GethInstanceConfig[], verbose: boo
...
 }
 
-// Add validator 0 as a peer of each other validator.
+// Peer every validator so they are fully connected
 export async function connectValidatorPeers(instances: GethInstanceConfig[]) {
   await connectPeers(
-    instances.filter(({ wsport, rpcport, validating }) => validating && (wsport || rpcport))
+    instances.filter(
+      ({ wsport, rpcport, validating, isProxy, isProxied }) =>
+        isProxy || (validating && !isProxied && (wsport || rpcport))
```
nategraf: Parenthesis are a bit off unless I am mistaken.
```suggestion
        (isProxy || (validating && !isProxied)) && (wsport || rpcport)
```

packages/celotool/src/e2e-tests/validator_handshake_tests.ts
```diff
@@ -0,0 +1,263 @@
+import { eqAddress } from '@celo/utils/lib/address'
+import { assert } from 'chai'
+import _ from 'lodash'
+import sleep from 'sleep-promise'
```
nategraf: We don't always import `sleep` consistently, but in the `e2e` tests, `utils.ts` seems to be consensus
```suggestion
import { sleep } from './utils'
```

packages/celotool/src/e2e-tests/validator_handshake_tests.ts
```diff
@@ -0,0 +1,263 @@
...
+    await context.hooks.before()
+  })
+
+  after(context.hooks.after)
+
+  describe('Validator handshake', () => {
+    before(async function() {
+      // reset the instances for each test
+      gethConfig.instances = getInstances()
+      configureInstances(gethConfig)
```
nategraf: Is this something we should be doing for other tests? If not could you expand on the comment as to why this type of reset is needed?

packages/celotool/src/e2e-tests/validator_handshake_tests.ts
```diff
@@ -0,0 +1,263 @@
...
+    instances.push({
+      name: `proxy${i}`,
+      validating: false,
+      syncmode: 'full',
+      port: 30303 + 3 * i + 1,
+      proxyport: 30303 + 3 * i + 2,
+      rpcport: 8545 + 2 * i + 1,
+      lightserv: false,
+      isProxy: true,
+      setNodeKey: true,
```
nategraf: Why is setting the node key important here?

packages/celotool/src/e2e-tests/validator_handshake_tests.ts
```diff
@@ -0,0 +1,263 @@
...
+      // Retry up to two times to mitigate this issue. Restarting the nodes is not needed.
+      this.retries(2)
+
+      const val0Provider = `http://localhost:${gethConfig.instances[0].rpcport}`
+      const val0Web3 = new Web3(val0Provider)
+      const val0Admin = new Admin(val0Provider)
+      const val0InfoBefore = await val0Admin.getNodeInfo()
+      const val0EnodeBefore = val0InfoBefore.enode
+
+      const val1Provider = `http://localhost:${gethConfig.instances[1].rpcport}`
```
nategraf: The fact that this code works to get the correct node seems to be a consequence of the exact implementation of `getInstances` and the constant chosen above. Please refactor this to be less fragile in case someone comes in and makes changes to those because it would not be clear what the issue is.

packages/celotool/src/e2e-tests/validator_handshake_tests.ts
```diff
@@ -0,0 +1,263 @@
...
+
+      assert(
+        !_.isEqual(val0EnodeBefore, val0EnodeAfter),
+        'Restarting instance 0 with a new nodekey did not result in a new enode'
+      )
+
+      // validator 0 still has its valEnodeTable, so it will try to connect to
+      // validator 1 immediately. It should identify itself as a validator and
+      // immediately be accepted as a ValidatorPurpose peer by validator 1
+      const val1Peers = await val1Admin.getPeers()
```
nategraf, line 104: Could this be the result of val1 connecting to the bootnode, gossiping the announce messages, and having the connection be established after the val0 node received the latest announce?

packages/celotool/src/e2e-tests/utils.ts
```diff
@@ -339,3 +284,111 @@ export function getContext(gethConfig: GethRunConfig, verbose: boolean = verbose
...
+        // salt id generation, milliseconds might not be
+        // enough to generate unique ids
+        id: new Date().getTime() + Math.floor(Math.random() * (1 + 100 - 1)),
+      },
+      // @ts-ignore
+      (err: any, result: any) => {
+        if (err) {
+          return reject(err)
+        }
+        return resolve(result)
```
nategraf: It seems that `result` here may be of type `JsonRPCResponse`, which would need one more indirection to get to the result I believe you want.
https://github.com/ethereum/web3.js/blob/c0e727eac7c16826c39fd43a03a4930b5ff6fff8/packages/web3/types.d.ts#L232-L234

packages/celotool/src/e2e-tests/utils.ts
tkporter: I originally pulled this from the `packages/protocol` test utils, I just removed the copy in this file (forgot to before, I originally made the copy because of some typescript issues with the protocol test utils) and fixed the issue in the original function

packages/celotool/src/e2e-tests/utils.ts
tkporter: Fixed in the function in `packages/protocol` that is now the one used

packages/celotool/src/e2e-tests/validator_handshake_tests.ts
tkporter: This test doesn't have `useBootnode: true` in the `GethRunConfig`, and because validator 0 will have a new enode, none of the other nodes will be able to initiate a connection to validator 0.

As an extra check to make sure we can rely on the val enode table, I added some code to ensure that the val enode table of validator 0 contains the correct enode of validator 1 after validator 0 has been restarted. There's also a check at the end of the test to make sure that the connection between validator 0 and validator 1 was initiated by validator 0, which confirms that it was able to get around validator 1's max peer check.

packages/celotool/src/lib/geth.ts
tkporter: It's needed for proxies so that a proxied validator can specify the internal/external enode urls for its proxy when starting up without having to wait for the proxy to be deployed first.

I just made setNodeKey default to being falsy, and force the node key to be set when the node is a proxy. We can decide later on if it's worth changing up the e2e tests to remove the need entirely

packages/celotool/src/lib/geth.ts
tkporter: Good catch!

packages/celotool/src/e2e-tests/validator_handshake_tests.ts
tkporter: Added a comment, it's because the tests are modifying the geth config (like changing proxy config, or the max peers). Changed this to `beforeEach` as well

packages/celotool/src/e2e-tests/validator_handshake_tests.ts
tkporter: Originally this was to ensure that the proxied validator knew the enode of the proxy, however I just removed this now that when starting up get it will always set the node key if the node is a proxy

patches/@segment/analytics-react-native-firebase+1.1.1-beta.3.patch
```diff
@@ -3,24 +3,24 @@ patch-package
 +++ b/node_modules/@segment/analytics-react-native-firebase/android/build.gradle
```
jmrossy: Was it intentional to include this change in this PR?

tkporter: This is super old so when I pulled master I was battling the patches-- I think it's gone away now that I am using the correct `yarn.lock`, so this will likely be unchanged from what's on master. Still cleaning things up

nambrot:
Closing for now, please feel free to re-open if still valid

PR #2830, opened 2020-02-19
[contractKit] Fix ganache-core failing in unit tests

### LINKED ISSUE #2829: Ungraceful test finish if it wasn't using the devchain
### Expected Behavior
All test must finish without ganache-core problems

### Current Behavior
When a unit test is not run in batch, ganache-core fails with an 'undefined pop' error

gastonponti:
### Description

All the contrackit and cli tests that didn't use ganache and were run separately were throwing a ganache-core error

### Related issues

- Fixes #2829


packages/cli/.npmignore
```diff
@@ -12,4 +12,6 @@
 /src
 
 /lib/**/*.test.*
-/lib/test-utils
\ No newline at end of file
+/lib/test-utils
+
+/tmp/
```
mcortesi: i would suggest using a `.tmp` so it's hidden...

mcortesi (approved):

PR #2838, opened 2020-02-20
Use the last address attested in the wallet

cmcewen:
### Description

On Alfajores, we may add attestations for the same phone number to multiple addresses. This switches to use the last one instead of the first

### Tested

Used the app to send a payment to Marek's address


annakaz (approved):
Nice

jmrossy (approved):

PR #2840, opened 2020-02-20
Unlock: adding a check whether account is voting

### LINKED ISSUE #2835: When unlocking, check that account is not currently voting
### Expected Behavior

When unlocking, test if the account is currently voting.

### Current Behavior

`Error: Returned error: gas required exceeds allowance (20000000) or always failing transaction` etc.

mrsmkl:
### Description

When trying to unlock funds, now tests if the account is voting.

### Tested

Only manual testing, and only tested that it still works while not voting.

### Related issues

- Fixes #2835


packages/cli/src/utils/checks.ts
```diff
@@ -259,6 +259,13 @@ class CheckBuilder {
       `${address} is not registered as an account. Try running account:register`
     )
 
+  isNotVoting = (address: Address) =>
+    this.addCheck(
+      `${address} is not currently voting`,
+      this.withGovernance((gov) => negate(gov.isVoting(address))),
+      `${address} is currently voting. Wait until the referendum ends.`
```
nategraf: Upvotes would cause the same effect, in which case the action is different.
```suggestion
      `${address} is currently voting in governance. Revoke your upvotes or wait for the referendum to end.`
```

packages/cli/src/utils/checks.ts
```diff
@@ -259,6 +259,13 @@ class CheckBuilder {
       `${address} is not registered as an account. Try running account:register`
     )
 
+  isNotVoting = (address: Address) =>
+    this.addCheck(
+      `${address} is not currently voting`,
```
nategraf: It is good to distinguish this case from voting for validator groups
```suggestion
      `${address} is not currently voting on a governance proposal`,
```

nategraf (approved):
Just a comment about the wording of the error message

PR #2841, opened 2020-02-20
Add attestation column on validator explorer

Pedro-vk:
### Description

_Add attestation column on validator explorer_


packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -518,7 +566,7 @@ const styles = StyleSheet.create({
     textAlign: 'center',
   },
   tableCellTitleNumber: {
-    marginLeft: 15 + 20 + 20,
```
aaronmgdr, line 521: nit: im guessing that your adding values together as they signify something, consider using a constant so its clear what 20 means and what other places are 20 for the same reason

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -222,6 +245,12 @@ class ValidatorsListApp extends React.PureComponent<ValidatorsListProps & I18nPr
                 name="Uptime"
                 order={orderBy === 'uptime' ? orderAsc : null}
               />
+              <HeaderCell
+                onClick={this.orderBy.bind(this, 'attestation')}
```
aaronmgdr: btw binding on every render will hurt reacts optimization heuristics as a new function will be created each render

Pedro-vk: Should I create N functions for that? or there is a way more simple?

aaronmgdr: maybe could do something in the constructor like 
```
Object.keys(orderByTypes).forEach(orderType => {
   this[`orderBy${orderType}`] = () => this.orderBy(orderType)
)
```

at least i think it would work.

aaronmgdr:
one more overall comment is that I would suggest breaking into smaller components. This will mean you can create some small pure components which will increase render performance. 

I use a rough rule that a render should be probably less than about 100 lines, and usually more like 15 to 60 lines.

aaronmgdr:
would you add a jest test for the `packages/web/src/dev/ValidatorsList.tsx` component? Id like to keep test coverage up. :D

aaronmgdr:
found an issue on mobile. 

IMG_5461F7652C8A-1


I think the following could help

```js
       <View style={[styles.cover, styles.pStatic]}>
-        <View style={[styles.pStatic]}>
+        <View style={[styles.pStatic, styles.allowSideScroll]}>

```

```js
+  allowSideScroll: { overflow: 'scroll' },
   cover: {
     marginTop: HEADER_HEIGHT,
     backgroundColor: colors.dark,
-    maxWidth: '100vw',
+    minHeight: '100vh',
   },
```


add minHeight so that there isnt a moment of white screen before it loads. 
remove maxWidth so that background can expand out if needed (such as side scroll)
add overflow scroll so that it is added to entire area not just to specific table (this would be the fix for image above

Pedro-vk:
@aaronmgdr it is not possible, because if we set the overflow in a parent of a sticky element (used for the header) it stops working.

Pedro-vk:
@aaronmgdr 
Seems that the solution is not simple.
I think that we should merge this PR, that has bugfixes and additions. And then work in a fix for that in a separate PR.

aaronmgdr:
@Pedro-vk  ok yeah it is a separate issue so thats cool, 

did you see my other comments?

Pedro-vk:
> @Pedro-vk ok yeah it is a separate issue so thats cool,
> 
> did you see my other comments?

@aaronmgdr  Test added :+1:

packages/web/src/dev/ValidatorsList.tsx
aaronmgdr: @Pedro-vk

packages/web/src/dev/ValidatorsList.tsx
aaronmgdr: @Pedro-vk

packages/web/src/dev/ValidatorsListRow.tsx
```diff
@@ -0,0 +1,185 @@
...
+import { cutAddress, formatNumber } from 'src/utils/utils'
+
+class Text extends RNText {
+  render() {
+    return <RNText style={[styles.defaultText, this.props.style]}>{this.props.children}</RNText>
+  }
+}
+
+interface Props {
+  group: any
```
aaronmgdr: highly recommend typing this with am interface Group {address: string, name: string, ...etc}

aaronmgdr (approved):

PR #2850, opened 2020-02-20
Add election view function with set size as parameter

asaj:
### Description

Allows users to pull the validator set that would be elected right now, taking the minimum and maximum validator set sizes as input. 

### Tested

Not tested.

### Other changes

None
### Related issues

- Needed for direct announce changes

### Backwards compatibility

Backwards compatible


packages/protocol/contracts/governance/Election.sol
```diff
@@ -817,9 +817,22 @@ contract Election is
...
+  }
+
+  /**
+   * @notice Returns a list of elected validators with seats allocated to groups via the D'Hondt
+   *   method.
+   * @return The list of elected validators.
+   * @dev See https://en.wikipedia.org/wiki/D%27Hondt_method#Allocation for more information.
+   */
+  function electValidatorSigners(uint256 minElectableValidators, uint256 maxElectableValidators)
+    external
```
m-chrzan: Wouldn't this function need to be `public` now since it's called from within the contract?

m-chrzan (approved):
Just one potential issue, which should (hopefully!) be caught by governance tests.

packages/protocol/contracts/governance/Election.sol
asaj: Definitely

asaj:
```suggestion
    public
```

packages/protocol/contracts/governance/Election.sol
```diff
@@ -817,9 +817,22 @@ contract Election is
    * @notice Returns a list of elected validators with seats allocated to groups via the D'Hondt
    *   method.
    * @return The list of elected validators.
-   * @dev See https://en.wikipedia.org/wiki/D%27Hondt_method#Allocation for more information.
    */
   function electValidatorSigners() external view returns (address[] memory) {
+    return electValidatorSigners(electableValidators.min, electableValidators.max);
```
asaj:
```suggestion
    return electNValidatorSigners(electableValidators.min, electableValidators.max);
```

packages/protocol/contracts/governance/Election.sol
```diff
@@ -817,9 +817,22 @@ contract Election is
...
+  }
+
+  /**
+   * @notice Returns a list of elected validators with seats allocated to groups via the D'Hondt
+   *   method.
+   * @return The list of elected validators.
+   * @dev See https://en.wikipedia.org/wiki/D%27Hondt_method#Allocation for more information.
+   */
+  function electValidatorSigners(uint256 minElectableValidators, uint256 maxElectableValidators)
```
asaj:
```suggestion
  function electNValidatorSigners(uint256 minElectableValidators, uint256 maxElectableValidators)
```

PR #2866, opened 2020-02-21
Private tx nodes for cLabs services

### LINKED ISSUE #2632: Testnet deployment should create a set of private full-nodes to serve infrastructure
### Expected Behavior

Blockscout and the Oracle should be run from a set of private full-nodes sequestered behind a firewall in private network.

tkporter:
### Description

Introduces a new env var, `PRIVATE_TX_NODES` which indicates how many private TX nodes to create. Private TX nodes are the same as tx nodes but with their RPCs only accessible to local IP addresses. For services like blockscout, we want to expose privileged RPC APIs that are dangerous to expose externally (as we saw with `debug_setHead`). For services like the oracle, we want to make sure they rely on nodes whose RPC ports are not exposed to the internet and more vulnerable to attacks.

### Tested

Ran k8s & vm testnets and deployed blockscout/oracle

### Other changes

- Renamed exposed TCP ports on k8s to be `celo` instead of `discovery`
- Added session affinity for the client IP to k8s services (like forno)
- Deleted integrationtesting env as I'm fairly sure it's not used anymore

### Related issues

- Fixes #2632 

### Backwards compatibility

Will require an upgrade of blockscout & oracle deployments once this infrastructure is deployed.


.env.alfajores
```diff
@@ -67,6 +67,8 @@ FAUCET_CUSD_WEI=60000000000000000000000
 
 VALIDATORS=70
 TX_NODES=10
+# Nodes whose RPC ports are only internally exposed
+PRIVATE_TX_NODES=1
```
nategraf: Could we run 2 or 3 in the prod environments? (`baklava` and `alfrajores`)

packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -38,7 +38,7 @@ async function failIfSecretMissing(secretName: string, namespace: string) {
 async function copySecret(secretName: string, srcNamespace: string, destNamespace: string) {
   console.info(`Copying secret ${secretName} from namespace ${srcNamespace} to ${destNamespace}`)
   await execCmdWithExitOnFailure(`kubectl get secret ${secretName} --namespace ${srcNamespace} -o yaml |\
-  grep -v creationTimestamp | grep -v resourceVersion | grep -v selfLink | grep -v uid |\
+  grep -v creationTimestamp | grep -v resourceVersion | grep -v selfLink | grep -v uid | grep -v namespace |\
```
nategraf: Nit: This could be consolidated.
```suggestion
  grep -Ev 'creationTimestamp|resourceVersion|selfLink|uid|namespace' |\
```

nategraf (approved):

packages/celotool/src/lib/helm_deploy.ts
tkporter: Ah cool, thanks

packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -38,7 +38,7 @@ async function failIfSecretMissing(secretName: string, namespace: string) {
 async function copySecret(secretName: string, srcNamespace: string, destNamespace: string) {
   console.info(`Copying secret ${secretName} from namespace ${srcNamespace} to ${destNamespace}`)
   await execCmdWithExitOnFailure(`kubectl get secret ${secretName} --namespace ${srcNamespace} -o yaml |\
-  grep -v creationTimestamp | grep -v resourceVersion | grep -v selfLink | grep -v uid |\
+  grep -Ev 'creationTimestamp|resourceVersion|selfLink|uid|namespace' |\
```
yorhodes: 😈

packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -376,6 +376,10 @@ export async function createStaticIPs(celoEnv: string) {
     await Promise.all(
       range(numValdiators).map((i) => registerIPAddress(`${celoEnv}-validators-${i}`))
     )
+    const numPrivateTxNodes = parseInt(fetchEnv(envVar.PRIVATE_TX_NODES), 10)
+    await Promise.all(
```
yorhodes, line 368: there is a `concurrentMap` in `utils/async` if you weren't aware

PR #2877, opened 2020-02-21
[M10] Not using OpenZeppelin contracts

jfoutts-celo:
### Description

Upgrade to openzeppelin-solidity 2.5.0

### Tested

Unit tests.

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/273


jfoutts-celo:
This PR addresses the issues raised in celo-org/celo-labs#273. However there seems to be no way to use thew new `ReentrancyGuard` with proxies, because it's state is private and only initialized via constructor.  So this PR introduces another occurrence of the "Not using OpenZeppelin contracts" issue that it's also fixing.

Should I open an openzeppelin-contracts issue about ReentrancyGuard and proxy support?

packages/protocol/contracts/common/Signatures.sol
```diff
@@ -29,8 +31,14 @@ library Signatures {
     pure
     returns (address)
   {
-    bytes memory prefix = "\x19Ethereum Signed Message:\n32";
-    bytes32 prefixedHash = keccak256(abi.encodePacked(prefix, messageHash));
-    return ecrecover(prefixedHash, v, r, s);
+    uint256 v32 = v;
+    bytes memory signature = new bytes(96);
+    assembly {
```
jfoutts-celo, line 36: This recombination of arguments is unfortunate.  Should I send a PR to openzeppelin https://github.com/OpenZeppelin/openzeppelin-contracts/compare/master...jfoutts-celo:master ?

packages/protocol/contracts/common/SafeCastTest.sol
```diff
@@ -2,7 +2,7 @@ pragma solidity ^0.5.0;
 // TODO: Remove this and use upstream when
 // https://github.com/OpenZeppelin/openzeppelin-contracts/pull/1926/files gets merged
 
-import "./SafeCast.sol";
+import "openzeppelin-solidity/contracts/utils/SafeCast.sol";
```
asaj: Let's remove this and `safecast.ts`

asaj (approved):

packages/protocol/contracts/common/Signatures.sol
come-maiz: @jfoutts-celo I think our teammates that maintain the contracts would appreciate it :D At least we should discuss if we should support this alternate way of calling it.

come-maiz:
> Should I open an openzeppelin-contracts issue about ReentrancyGuard and proxy support?

Yes please! 

@jfoutts-celo I recommend to explain in the docstring why you are copying an old version of the contract, and link to the issue you reported. So future readers of the code won't have to dig for this pull request.

PR #2881, opened 2020-02-22
Adjust migrations for mainnet, make freezability governable

### LINKED ISSUE #2517: Update migrations config for RC
- [ ] set {min, max}ElectableValidators
- [ ] update slashing parameters
- [ ] reset time based params, eg epochSize
- [ ] make validator 0 the Foundation 
- [ ] only run validators migration during tests or internal testnets

### LINKED ISSUE #2819: Rewards, gold transfers, and dollar transfers can be unfrozen by governance
### Expected Behavior

Epoch rewards, Celo Gold transfers, and Celo dollar transfers can be unfrozen via on-chain governance.

### Current Behavior

non-governance "freezer" has unilateral power

asaj:
### Description

This PR modifies the migrationsConfig and migrations to be in line with the intended mainnet deployment process. Specifically, it modifies the migrationsConfig with the intended mainnet configration, removes reserve funding and oracle reporting from the migrations (the former can be overridden), and adds support for configuration of reserve spenders and other reserve addresses.

Furthermore, it standardizes the way smart contract functionality is frozen via the `Freezer` and `Freezable` smart contracts.

### Tested

- Unit tests

### Other changes

- Some cleanup here and there

### Related issues

- Fixes #2517 
- Fixes #2819 
- Creates #2908

### Backwards compatibility

No, changed storage layout of some smart contracts


packages/protocol/contracts/common/Freezable.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.5.3;
+
+import "./UsingRegistry.sol";
+
+contract Freezable is UsingRegistry {
+  // onlyWhenNotFrozen functions can only be called when `frozen` is false, otherwise they will
+  // revert.
+  modifier onlyWhenNotFrozen() {
```
nambrot, line 8: Off topic question, but do we know the gas impact on requiring this on all transfers forever?

packages/protocol/contracts/common/UsingRegistry.sol
```diff
@@ -40,6 +41,7 @@ contract UsingRegistry is Ownable {
   bytes32 constant FEE_CURRENCY_WHITELIST_REGISTRY_ID = keccak256(
     abi.encodePacked("FeeCurrencyWhitelist")
   );
+  bytes32 constant FREEZER_REGISTRY_ID = keccak256(abi.encodePacked("Freezer"));
```
nambrot, line 44: Is this a general freezer, or the GoldToken freezer specifically? If it's the latter, I would maybe name it as such?

packages/protocol/migrations/07_reserve.ts
```diff
@@ -28,14 +29,22 @@ module.exports = deploymentForCoreContract<ReserveInstance>(
...
+    config.reserve.otherAddresses.forEach(async (otherAddress) => {
+      console.info(`Marking ${otherAddress} as an "otherReserveAddress"`)
+      await reserve.addOtherReserveAddress(otherAddress)
     })
-    console.info(`Marking ${network.from} as a reserve spender`)
-    await reserve.addSpender(network.from)
+    if (config.reserve.initialBalance) {
+      console.info('Sending the reserve an initial gold balance')
+      const network: any = truffle.networks[networkName]
+      await web3.eth.sendTransaction({
```
nambrot: The PR description says that we removed the funding of the reserve in the migrations?

packages/protocol/test/common/goldtoken.ts
```diff
@@ -6,23 +7,38 @@ import {
...
   const TWO_GOLDTOKEN = new BigNumber('2000000000000000000')
   const sender = accounts[0]
   const receiver = accounts[1]
 
   beforeEach(async () => {
+    freezer = await Freezer.new()
     goldToken = await GoldToken.new()
+    registry = await Registry.new()
+    await registry.setAddressFor(CeloContractName.Freezer, freezer.address)
+    await goldToken.initialize(registry.address)
```
nambrot, line 41: Should we test that freezing works?

packages/protocol/migrations/22_governance.ts
```diff
@@ -62,7 +46,9 @@ module.exports = deploymentForCoreContract<GovernanceInstance>(
       'Escrow',
       'Exchange',
       'FeeCurrencyWhitelist',
+      'Freezer',
       'GasPriceMinimum',
+      'GoldToken',
```
nambrot: Should we also remove the elect validators migration?

nambrot (commented):
Ideally in this PR, but probably not strictly necessary, but can we have another method/command that executes all the removed functionality for when developers do not want to run mainnet-like migrations?

packages/protocol/contracts/common/Freezable.sol
asaj: It wouldn't need to be on all transfers, forever, as the modifier could (and should) eventually be removed via governance proposal

packages/protocol/contracts/common/UsingRegistry.sol
asaj: General, as it also freezes stabletoken transfers, the exchange, and epoch rewards

packages/protocol/migrations/07_reserve.ts
asaj: We did by default, it can be overridden for testing purposes, updated the PR description

packages/protocol/migrations/22_governance.ts
asaj: It's useful for testing purposes, and can be easily skipped as it's the very last migration. I think it's fine for now

packages/protocol/test/common/goldtoken.ts
asaj: We have tests elsewhere, I'm pretty comfortable not testing here (famous last words)

nambrot (approved):

PR #2886, opened 2020-02-23
[M03] Constraining slashing penalties

### LINKED ISSUE #2943: Slashing reward recipient receives double rewards from Downtime and DoubleSigning
### Expected Behavior

Slashing reporter should receive `reward` worth of nonvoting lockedgold after a successful slash

### Current Behavior

Slashing reporter will receive `reward * 2` worth of nonvoting lockedgold.

Cause: `SlasherUtil` calls `lockedgold.slash` once on the "bad actor" and once on their group, both calls with the same `recipient` and `reward`, meaning `recipient` receives `reward` twice.

Proposed solution: easiest solution is to just make it clear that `slashingIncentives.reward` should be 1/2 of the proposed reward, or change `setSlashingIncentives` to set `reward` to 1/2 of the provided reward param.

lucasege:
### Description

Constrained the max penalty when calling `performSlashing` to be the minimum of an account's total locked gold and the given penalty.

### Tested

Unit tests

### Related issues

- https://github.com/celo-org/celo-labs/issues/266
- Fixes #2943

### Backwards compatibility

Yes


asaj:
I'm not sure this actually fixes the "problem" (which I do not believe to be a problem in the first place).

The raised concern was that validators can avoid slashing by front-running the call to slashing by withdrawing locked gold, causing the slashing transaction to fail. I don't think this addresses that, the slashing transaction would just fail earlier.

However, I don't think this is a real issue, as the penalties should always be less than the minimum locked gold needed to run a validator, so validators would not be able to withdraw enough gold to avoid slashing unless they had already been not validating for >= 60 days (or whatever the current requirement duration is), which should be considered the statute of limitations on slashing.

lucasege:
Agree about not actually "solving" the problem - as we discussed last week, this PR only seeks to remove the possibility of setting a slashing penalty greater than the locked gold requirement for a validator. This is only a "sanity check" of sorts.
If you don't think we need to add this to move the issue's severity lower then I can close this PR.

packages/protocol/contracts/governance/SlasherUtil.sol
```diff
@@ -65,11 +66,16 @@ contract SlasherUtil is Ownable, Initializable, UsingRegistry, UsingPrecompiles
     uint256[] memory groupElectionIndices
   ) internal {
     ILockedGold lockedGold = getLockedGold();
+    // Cannot set reward as local because of solc stack too deep
+    uint256 penalty = Math.min(
```
asaj: Instead, what if in `lockedGold.slash`, slash `actualPenalty = Math.min(getAccountTotalLockedGold(), penalty)`, and `actualReward = Math.min(reward, actualReward)`?

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -323,28 +323,41 @@ contract LockedGold is ILockedGold, ReentrancyGuard, Initializable, UsingRegistr
     address[] calldata greaters,
     uint256[] calldata indices
   ) external onlySlasher {
-    uint256 maxSlash = Math.min(penalty, getAccountTotalLockedGold(account));
-    require(maxSlash >= reward, "reward cannot exceed penalty.");
+    uint256 actualPenalty = Math.min(
+      penalty,
+      getValidators().getAccountLockedGoldRequirement(account)
```
asaj: I don't think we need to look at the requirement here, just the locked gold balance

packages/protocol/contracts/governance/SlasherUtil.sol
```diff
@@ -69,7 +70,7 @@ contract SlasherUtil is Ownable, Initializable, UsingRegistry, UsingPrecompiles
       validator,
       slashingIncentives.penalty,
       recipient,
-      slashingIncentives.reward,
+      slashingIncentives.reward / 2,
```
asaj: nit: use safemath

packages/protocol/contracts/governance/SlasherUtil.sol
```diff
@@ -80,7 +81,7 @@ contract SlasherUtil is Ownable, Initializable, UsingRegistry, UsingPrecompiles
       group,
       slashingIncentives.penalty,
       recipient,
-      slashingIncentives.reward,
+      slashingIncentives.reward / 2,
```
asaj: ditto

asaj (approved):

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -158,5 +160,17 @@ contract('DoubleSigningSlasher', (accounts: string[]) => {
         slasher.slash(validator, validatorIndex, headerA, headerC, 0, [], [], [], [], [], [])
       )
     })
+
+    describe("when the validator's lockedgold requirement is less than the penalty", () => {
```
asaj: nit: "when the validator's LockedGold balance is less than the penalty"

packages/protocol/test/governance/double_signing_slasher.ts
```diff
@@ -158,5 +160,17 @@ contract('DoubleSigningSlasher', (accounts: string[]) => {
...
+      beforeEach(async () => {
+        await validators.setAccountLockedGoldRequirement(validator, slashingPenalty - 1)
```
asaj: I think you need to call something on `mockLockedGold` here

asaj:
Closing for now as it turns out we've already mostly implemented what we intended to here

PR #2890, opened 2020-02-23
[L13] Fraction reduction is not applied in all methods, [L18] Lack of checks for denominator

asaj:
### Description

This PR removes the almost completely unused `FractionUtil`, which should be deprecated in favor of `Fixidity`.

### Tested

Unit tests

### Other changes

None
### Related issues

- Fixes #https://github.com/celo-org/celo-labs/issues/313
- Fixes #https://github.com/celo-org/celo-labs/issues/318

### Backwards compatibility

Yes


m-chrzan (approved):
I *rate* this PR 5/5

PR #2892, opened 2020-02-23
[L15] Failure of `ecrecover` function does not revert

asaj:
### Description

Adds a check that `ecrecover` is able to recover the signer address, which returns 0 if not.

### Tested

Unit tests
### Other changes

None

### Related issues

- Fixes #https://github.com/celo-org/celo-labs/issues/315

### Backwards compatibility

Yes


m-chrzan (approved):

m-chrzan:
Looks like transfer costs changed/e2e tests broken.

asaj:
> Looks like transfer costs changed/e2e tests broken.

 I think this was just minor flakiness on circleci/github, I don't think this test should be affected.

PR #2901, opened 2020-02-23
[N17] Different address nomenclature for precompiled contracts

asaj:
### Description

Consistency.
### Tested

No
### Other changes

No
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/294

### Backwards compatibility

Yes


yerdua (approved):
Approving since this does make the address style consistent. Why is this the preferred way though?

PR #2905, opened 2020-02-23
[N23] Unused code

asaj:
### Description

This PR removes unused code, except for FractionUtil, which is removed in https://github.com/celo-org/celo-monorepo/pull/2890.

Note that the maximum values were being exposed for testing purposes, and the existing code already has tests showing transactions that would cause overflows properly revert. For posterity, these maximum values are instead preserved as part of the docstring on the relevant functions.

### Tested

Not tested

### Other changes

### Related issues

- Fixes #https://github.com/celo-org/celo-labs/issues/300
### Backwards compatibility

Yes


packages/protocol/contracts/common/FixidityLib.sol
```diff
@@ -80,16 +72,6 @@ library FixidityLib {
     return 115792089237316195423570985008687907853269984665640564;
   }
 
-  /**
-   * @notice Maximum value that can be safely used as an addition operator.
-   * @dev Test maxFixedAdd() equals maxUint256()-1 / 2
-   * Test add(maxFixedAdd(),maxFixedAdd()) equals maxFixedAdd() + maxFixedAdd()
-   * Test add(maxFixedAdd()+1,maxFixedAdd()+1) throws
-   */
-  function maxFixedAdd() internal pure returns (uint256) {
```
m-chrzan, line 89: Hmm, so I think this could still be a potentially useful number to keep in this file, at least in a comment. Can be useful information for consumers of this library.

packages/protocol/contracts/common/FixidityLib.sol
```diff
@@ -214,6 +191,7 @@ library FixidityLib {
    * Test multiply(maxFixedMul()+1,maxFixedMul()+1) fails
    */
   function multiply(Fraction memory x, Fraction memory y) internal pure returns (Fraction memory) {
+    require(x.value <= maxFixedMul() && y.value <= maxFixedMul());
```
m-chrzan: I believe there are unit tests that verify this reverts anyways without this check when the inputs are too large. I understand being conservative, as this library aims to be safe, but it also aims to be efficient, do we need an additional `require` here?

packages/protocol/contracts/common/FixidityLib.sol
```diff
@@ -278,6 +256,7 @@ library FixidityLib {
    */
   function divide(Fraction memory x, Fraction memory y) internal pure returns (Fraction memory) {
     require(y.value != 0, "can't divide by 0");
+    require(x.value <= maxFixedDividend());
```
m-chrzan: Similar question as above.

m-chrzan (approved):
A few comments I don't feel super strongly about but please take a look. Otherwise, looks *fixed* :sunglasses: 
Oh, also looks like this affects transfer costs, so e2e tests are broken.

packages/protocol/contracts/common/FixidityLib.sol
asaj: Good point. Will make the functions that expose these values external, so that clients that consume this code can do input sanitation more easily.

asaj: Actually nevermind, as this increases bytecode size. Going to move everything to comments.

packages/protocol/contracts/common/FixidityLib.sol
asaj: +1 to keeping as a comment

PR #2910, opened 2020-02-24
[protocol] add multisig to governance approver and reserve spender

### LINKED ISSUE #1599: Deploy multisigs for Governance Approver, Reserve Spender
### Expected Behavior

We should deploy two instances of the multisig wallet and set them as the approver for Governance proposals and the reserve spender.

### Current Behavior

Approver and spender set to externally owned account.

aslawson:
### Description

Add multiSig functionality for governance approver and reserve spender.  Approver and spender now point to multi addresses.  Two multiSig contracts+proxies are added.

TODO -- will update cli command line tooling to add/update the signatories and number of confirmations required

### Tested

Updated testing to governance integration tests.  Currently adding integration tests that did not exists for reserve spender as fast follow.

### Other changes

N/A

### Related issues

- Fixes #1599 

### Backwards compatibility

- backward compatible -- just changes address of spender/approver to multisig contracts


packages/protocol/migrations/07_reserve_sender_multisig.ts
```diff
@@ -0,0 +1,18 @@
+import { CeloContractName } from '@celo/protocol/lib/registry-utils'
```
asaj: nit: file name

packages/protocol/migrations/07_reserve_sender_multisig.ts
```diff
@@ -0,0 +1,18 @@
...
+    config.reserveSpenderMultiSig.signatories,
+    config.reserveSpenderMultiSig.numRequiredConfirmations,
+  ]
+}
+
+module.exports = deploymentForProxiedContract<ReserveSpenderMultiSigInstance>(
+  web3,
+  artifacts,
+  CeloContractName.ReserveSpenderMultiSig,
+  initializeArgs
```
asaj: One subtlety here that I neglected to mention is that ownership of the proxy should be turned over to itself the MultSig by setting the proxy owner to the proxy address. That will allow the MultiSig signatories to upgrade the MultiSig.

You should be able to do this in the `then` callback here, where the address of the proxy will be `reserveSpenderMultiSig.address`, where `reserveSpenderMultiSig` is the first argument.

packages/protocol/migrations/23_governance_approver_multisig.ts
```diff
@@ -0,0 +1,18 @@
...
+    config.governanceApproverMultiSig.signatories,
+    config.governanceApproverMultiSig.numRequiredConfirmations,
+  ]
+}
+
+module.exports = deploymentForProxiedContract<GovernanceApproverMultiSigInstance>(
+  web3,
+  artifacts,
+  CeloContractName.GovernanceApproverMultiSig,
+  initializeArgs
```
asaj: Ditto

packages/protocol/migrations/24_governance.ts
```diff
@@ -0,0 +1,79 @@
...
+  getDeployedProxiedContract,
+  transferOwnershipOfProxy,
+  transferOwnershipOfProxyAndImplementation,
+} from '@celo/protocol/lib/web3-utils'
+import { config } from '@celo/protocol/migrationsConfig'
+import { toFixed } from '@celo/utils/lib/fixidity'
+import { GovernanceInstance, GovernanceApproverMultiSigInstance } from 'types'
+
+const initializeArgs = async (): Promise<any[]> => {
+  const governanceApprovalMultiSig: GovernanceApproverMultiSigInstance = await getDeployedProxiedContract<
```
asaj: nit: governanceApproverMultiSig

packages/protocol/migrations/24_governance.ts
```diff
@@ -0,0 +1,79 @@
...
+  initializeArgs,
+  async (governance: GovernanceInstance) => {
+    const proxyOwnedByGovernance = ['GoldToken']
+    await Promise.all(
+      proxyOwnedByGovernance.map((contractName) =>
+        transferOwnershipOfProxy(contractName, governance.address, artifacts)
+      )
+    )
+
+    //TODO @amyslawson do multisig implementations get added here?
```
asaj: Great question!

They do not. The addresses that play the role of reserve spender/approver can be changed via governance, but the MultiSig should "own itself".

This can be accomplished by setting the proxy owner to the proxy address. The MultiSig signatories would then be able to upgrade the contract by submitting and confirming a MultiSig transaction via the proxy that calls `_setImplementation`.

packages/protocol/test/common/integration.ts
```diff
@@ -281,8 +290,12 @@ contract('Integration: Governance', (accounts: string[]) => {
 
   describe('When approving that proposal', () => {
     before(async () => {
+      // @ts-ignore
+      let txData = governance.contract.methods.approve(proposalId, dequeuedIndex).encodeABI()
```
asaj: nit: move this down after the timetravel

packages/protocol/test/common/integration.ts
```diff
@@ -150,8 +153,12 @@ contract('Integration: Governance slashing', (accounts: string[]) => {
...
```
asaj: ditto

packages/protocol/migrationsConfig.js
```diff
@@ -80,6 +80,10 @@ const DefaultConfig = {
     participationBaselineUpdateFactor: 1 / 5,
     participationBaselineQuorumFactor: 1,
   },
+  governanceApproverMultiSig: {
+    signatories: [0x5409ed021d9299bf6814279a6a1411a7e866a631],
```
asaj: nit: If tests would pass with this as an empty list, consider leaving it empty with a comment that says it needs to be overridden.

Ditto for the spender.

asaj (commented):
Nice work! 👏👏 👏

packages/protocol/migrationsConfig.js
asaj: And either way actually, please leave a comment that indicates that a 1/1 multisig is just a placeholder until the correct signatory addresses are known.

aslawson: i won't be able to do this with the current implementation -- there is validation in the multiSig that confirms that the signatory count is greater than 0, the required confirmations is greater than 0, and that the required confirmations is less than the signatory count.  

unless we wanted to remove this validation in the MultiSig contract...

I will add a comment regardless.

packages/protocol/runTests.js
```diff
@@ -16,6 +16,12 @@ const migrationOverrides = {
   downtimeSlasher: {
     slashableDowntime: 60, // epoch length is 100 for unit tests
   },
+  governanceApproverMultiSig: {
+    signatories: [network.from],
```
asaj, line 20: nit: Consider including `numRequiredConfirmations` here as well, so that if the `migrationsConfig` changed to e.g. 2 the tests would still work

asaj (approved):

PR #2912, opened 2020-02-24
Terraform: AWS v1

bgeihsgt:
### Description

Add a set of terraform scripts, modules, and submodules that set up:

1. VPC with private and public subnets in multiple availability zones.
2. Proxies (0 or more)
3. Validators (0 or more)
4. Attestation Services (0 or more)

The structure was modeled after the GCP structure with `testnet` configuring the VPC and nodes and `example` being the launch point. See README for more information.

Note that this change moved the GCP config into its own `gcp` folder.

### Tested

Tested each module apply and destroy individually across two availability zones in `us-west-1`. Verified that the nodes synced via Docker logs.

### Other changes

As mentioned above, the GCP scripts are now in a `gcp` folder under `terraform-modules-public`

Also added a couple more items to the terraform `.gitignore`

### Related issues

Couldn't find any issues for this.

### Backwards compatibility

This does not affect backwards compatibility.


packages/terraform-modules-public/aws/README.md
```diff
@@ -0,0 +1,28 @@
...
+- `validator` module for deploying a Validator
+- `attestation-service` for deploying the Attestation Service (https://docs.celo.org/getting-started/baklava-testnet/running-a-validator#running-the-attestation-service). Each attestation service will connect to the same multi-AZ RDS DB instance. Only Twilio is supported at this time.
+
+## Operating System
+
+All nodes run on the Ubuntu LTS 18.04 AMI. Running `terraform apply` will select the latest available AMI.
+
+## Hardening & Security
+
+Celo nodes will get provisioned with some recommended security settings from Celo's Security Audit team. Most of this hardening is done in `install-base.sh` and `final-hardening.sh`.
```
tkporter: Nit: could you change "Celo's Security Audit team" to be "cLab's Security Audit team"?

packages/terraform-modules-public/aws/example/variables.tf
```diff
@@ -0,0 +1,215 @@
...
+  default = {
+    az1     = {}
+    az2     = {}
+  }
+  # Here is an example configuration. We recommend putting this into a secret.auto.tfvars file.
+  # default = {
+  #   az1 = {
+  #     myvalidator_az1_01 = {
+  #       validator_name            = "myvalidator_az1_01"
+  #       validator_signer_address  = "0000000011111111222222223333333344444444"
```
tkporter, line 83: Does it matter if these are `0x` prefixed?

packages/terraform-modules-public/aws/example/variables.tf
```diff
@@ -0,0 +1,215 @@
...
+      az1 = {}
+      az2 = {}
+    }
+  # Here is an example configuration. We recommend putting this into a secret.auto.tfvars file.
+  # default = {
+  #   az1 = {
+  #     myvalidator_az1_01 = {
+  #       name                          = "myvalidator_az1_01"
+  #       signer_address                = "0000000011111111222222223333333344444444"
+  #       signer_private_key            = "...Place the keystore file contents here Make sure to escape the double quotes. ..."
```
tkporter: I think the variable name could be a little confusing, wdyt of something that makes it clearer the value is supposed to be the JSON contents of the keystore file for the address and not the real private key? `signer_private_key_file_contents` or something

I also think a new user may not be familiar enough with the keystore to understand what this means, being more specific here about which file is needed would help

packages/terraform-modules-public/aws/example/variables.tf
```diff
@@ -0,0 +1,215 @@
...
+    description     = "Configuration for zero or more validators in each availability zone"
+    type            = object({ 
+      az1 = map(object({
+        name                          = string
+        signer_address                = string
+        signer_private_key            = string
+        signer_private_key_password   = string
+        signer_private_key_filename   = string
+        proxy_enode                   = string
+        proxy_private_ip              = string
```
tkporter: How would the private/public IPs be known ahead of time if the proxies & validators are both deployed in the same deployment in the `testnet/main.tf`? What do you think about having the `proxy` module have the private & public IPs be outputs & then pass the IPs to the `validator` modules as variables at the `testnet/main.tf` level?

packages/terraform-modules-public/aws/testnet/modules/attestation-service/variables.tf
```diff
@@ -0,0 +1,64 @@
...
+}
+
+variable key_pair_name {
+  type          = string
+  description   = "Name of the SSH key pair to access this node from the bastion"
+}
+
+variable volume_size {
+  type          = number
+  description   = "Size of the EBS volume for this instance"
```
tkporter: in GB?

packages/terraform-modules-public/aws/testnet/modules/bastion/main.tf
```diff
@@ -0,0 +1,29 @@
...
+    ami                         = module.ami.ami_ids.ubuntu_18_04
+    instance_type               = var.instance_type
+    subnet_id                   = var.subnet_id
+    associate_public_ip_address = true
+    vpc_security_group_ids      = [
+        data.aws_security_group.bastion.id
+    ]
+    key_name                    = var.key_pair_name
+
+    user_data                       = join("\n", [
```
tkporter: Might need to run `terraform fmt` on some of these files

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/install-chrony.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+apt-get update
+apt-get --assume-yes remove ntp*
+apt-get --assume-yes install chrony
+
+mv /etc/chrony/chrony.conf /etc/chrony/chrony.conf.old
+sed '/^pool ntp\.ubuntu\.com*/i server 169.254.169.123 prefer iburst minpoll 4 maxpoll 4' /etc/chrony/chrony.conf.old > /etc/chrony/chrony.conf
```
tkporter, line 10: Could you add a comment about what this is doing?

packages/terraform-modules-public/aws/testnet/modules/vpc/modules/subnet-private/main.tf
```diff
@@ -0,0 +1,82 @@
...
+    ingress {
+        rule_no     = 130
+        protocol    = "tcp"
+        from_port   = 30303
+        to_port     = 30303
+        cidr_block  = var.vpc_cidr_block
+        action      = "allow"
+    }
+
+    ingress {
```
tkporter: What does this rule & the other one in the public subnet do?

packages/terraform-modules-public/aws/testnet/main.tf
```diff
@@ -0,0 +1,149 @@
...
+    length                      = 50
+    special                     = false
+    min_lower                   = 1
+    min_upper                   = 1
+    min_numeric                 = 1
+}
+
+resource "aws_db_subnet_group" "attestation" {
+    name            = "celo-db-subnet-group"
+    subnet_ids      = [module.celo_vpc.subnet_ids.az1.private, module.celo_vpc.subnet_ids.az2.private]
```
tkporter: nice 😎

packages/terraform-modules-public/aws/testnet/main.tf
```diff
@@ -0,0 +1,149 @@
...
+}
+
+module "celo_bastion_az2" {
+    source            = "./modules/bastion"
+
+    subnet_id         = module.celo_vpc.subnet_ids.az2.public
+    key_pair_name     = var.key_pair_name
+    name              = "celo-bastion-az2"
+    instance_type     = var.instance_types.bastion
+}
```
jcortejoso: Is ssh access to the private instances the goal for the bastion instances? If that is the case, we could also consider aws session manager (https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager.html). It has some usability issues (like slow pasting to the session), but from a security point of view I find it very nice.

packages/terraform-modules-public/aws/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+All nodes run on the Ubuntu LTS 18.04 AMI. Running `terraform apply` will select the latest available AMI.
+
+## Hardening & Security
+
+Celo nodes will get provisioned with some recommended security settings from Celo's Security Audit team. Most of this hardening is done in `install-base.sh` and `final-hardening.sh`. 
+
+## Requirements
+
+Inside the [example](./example) folder you can find an example tf to use the module. We recommend you to use that tf as base file for your deployment, modifying the account variables used for your convenience.
```
tkporter: In an effort to emphasize documentation as something that is also decentralized, could you reword the places where "we" is used? eg "We recommend" -> "It's recommended"

tkporter (commented):
Thanks so much for this PR, Brian! This will be super helpful

A few small comments as a first pass, but it's looking really good.

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/final-hardening.sh
```diff
@@ -0,0 +1,60 @@
+# Enable autologout
```
jcortejoso, line 1: Nice. Probably we should consider importing this script to the rest of cloud setups.

packages/terraform-modules-public/aws/testnet/modules/vpc/main.tf
```diff
@@ -0,0 +1,246 @@
...
+    availability_zone_id            = data.aws_availability_zones.available.zone_ids[1]
+    nat_gateway_id                  = module.celo_public_subnet_az2.nat_gateway_id
+    vpc_cidr_block                  = aws_vpc.celo.cidr_block
+}
+
+resource "aws_security_group" "attestation_service" {
+    name            = "celo-attestation-service"
+    vpc_id          = aws_vpc.celo.id
+
+    ingress {
```
jcortejoso: Not a big deal but is there any functionality issue for using inline rules for some security groups and `aws_security_group_rule` for others? I used to use `aws_security_group_rule` because inline rules used to cause some issues with create&destroy the SG which caused some downtime, but I have read it was fixed some time ago in the aws provider.

jcortejoso (approved):
Great job. Very nice :).

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/run-proxy-node.sh
```diff
@@ -0,0 +1,14 @@
...
+CELO_VALIDATOR_SIGNER_ADDRESS=${validator_signer_address}
+
+NODE_DIRECTORY=/home/ubuntu/celo-proxy-node
+
+mkdir $NODE_DIRECTORY
+cd $NODE_DIRECTORY
+docker run -v $PWD:/root/.celo --rm $CELO_IMAGE init /celo/genesis.json
+BOOTNODE_ENODES=`docker run --rm --entrypoint cat $CELO_IMAGE /celo/bootnodes`
+
+docker run -d --name celo-proxy --restart always -p 30303:30303 -p 30303:30303/udp -p 30503:30503 -p 30503:30503/udp -v $PWD:/root/.celo $CELO_IMAGE --verbosity 3 --networkid $NETWORK_ID --syncmode full --maxpeers 100 --nousb --proxy.proxy --proxy.proxiedvalidatoraddress $CELO_VALIDATOR_SIGNER_ADDRESS --proxy.internalendpoint :30503 --etherbase $CELO_VALIDATOR_SIGNER_ADDRESS --bootnodes $BOOTNODE_ENODES --ethstats=${validator_name}-proxy@baklava-ethstats.celo-testnet.org
```
codyborn: It'd be great if this value could be passed in during deployment:
"{validator_name}-proxy@baklava-ethstats.celo-testnet.org"

packages/terraform-modules-public/aws/example/variables.tf
```diff
@@ -0,0 +1,215 @@
...
+  #     }
+  #     myvalidator_az1_02 = {
+  #       validator_name            = "myvalidator_az1_02"
+  #       validator_signer_address  = "5555555566666666777777778888888899999999"
+  #     }
+  #   }
+  #   az2 = {
+  #     myvalidator_az2_01 = {
+  #       validator_name            = "myvalidator_az2_01"
+  #       validator_signer_address  = "4444444433333333222222221111111100000000"
```
codyborn, line 100: Why are there multiple validators deployed with different keys?

packages/terraform-modules-public/aws/README.md
bgeihsgt: Will do!

packages/terraform-modules-public/aws/example/variables.tf
bgeihsgt: I don't think so. I haven't tested it though.

packages/terraform-modules-public/aws/example/variables.tf
bgeihsgt: Good call. I will change this to `signer_private_key_file_contents` and also make the comment/instructions more clear.

packages/terraform-modules-public/aws/example/variables.tf
bgeihsgt: I thought about this and ideally would feed in data from the proxy dynamically in the way you mentioned. Unfortunately, the enode makes this tricky. I'd need a way to get the enode from the proxy which isn't trivial at provisioning time.

So to provision a validator, the user needs to provision the proxy, harvest the values, and then put them in their variables. I think a more automated solution would be a candidate for a future enhancement, so not planning to make a change here.

packages/terraform-modules-public/aws/testnet/modules/attestation-service/variables.tf
bgeihsgt: Will fix.

packages/terraform-modules-public/aws/testnet/modules/bastion/main.tf
bgeihsgt: Will do

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/install-chrony.sh
bgeihsgt: Will do. This is a good way to keep your clocks in sync on EC2: https://aws.amazon.com/blogs/aws/keeping-time-with-amazon-time-sync-service/

packages/terraform-modules-public/aws/testnet/modules/vpc/modules/subnet-private/main.tf
bgeihsgt: Network ACLs are stateless in AWS. So you need to open inbound ports for connections established on outbound requests. This is necessary for outbound connections to work properly.

packages/terraform-modules-public/aws/README.md
bgeihsgt: Will do

packages/terraform-modules-public/aws/testnet/main.tf
bgeihsgt: Yup, the intent here is to facilitate SSH to private instances in a secure manner.

Thanks for the tip on session manager feels like a superior solution security-wise.

I think this is something to consider for a future enhancement. You ok with that?

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/final-hardening.sh
bgeihsgt: :+1: makes sense.

packages/terraform-modules-public/aws/testnet/modules/vpc/main.tf
bgeihsgt: Inline rules have worked well for me when I change them. In my testing, terraform has properly done SG updates instead of create/destroy.

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/run-proxy-node.sh
bgeihsgt: Agreed this should be a variable and not hardcoded. Will change.

packages/terraform-modules-public/aws/example/variables.tf
bgeihsgt: The idea is you can manage a fleet of proxies and validators from this script to manage multiple validators and/or hot standby validators. It's a requirement these would have different signer keys so we don't accidentally double sign. The user would authorize these keys to their validator address when they want to switch which node should sign.

packages/terraform-modules-public/aws/testnet/modules/vpc/modules/subnet-private/main.tf
bgeihsgt: Network ACLs in AWS are stateless, so you have to open inbound ports to allow outbound connections to get return traffic. These are the ports prescribed by AWS' public/private subnet best practice guide.

packages/terraform-modules-public/aws/README.md
bgeihsgt: 9399d1a6f0f14179c97c8ecf6f3f248a080cc158

packages/terraform-modules-public/aws/example/variables.tf
bgeihsgt: ad0ae49ad029ede980be183fe917a0b429c3a0d3

packages/terraform-modules-public/aws/testnet/modules/attestation-service/variables.tf
bgeihsgt: ddb63213d3a3806098fbd5b30f12e05f7dbf2bad

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/install-chrony.sh
bgeihsgt: 0359da61031bb00ada651a4b7bbf360fcbcff95d

packages/terraform-modules-public/aws/README.md
bgeihsgt: 9399d1a6f0f14179c97c8ecf6f3f248a080cc158

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/run-proxy-node.sh
bgeihsgt: b908a853017f09192724da95adec7c11386c18be

packages/terraform-modules-public/aws/testnet/modules/bastion/main.tf
bgeihsgt: fb0007ed24b90da642e7551538c8e2ede84b5a82

bgeihsgt:
@tkporter  @codyborn @jcortejoso I addressed your comments and this is ready for another pass.

packages/terraform-modules-public/aws/example/variables.tf
codyborn: I wonder if it'd be simpler to just run the entire script multiple times for each validator set you want to deploy. Given that they have different keys, I view them as different entities altogether. @tkporter what do you think?

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/run-proxy-node.sh
codyborn: thanks!

packages/terraform-modules-public/aws/example/variables.tf
bgeihsgt: That is an interesting alternative approach. I decided to model the AWS setup after the GCP setup (it allows multiple validators/proxies), so if we moved to another approach I'd think we'd want to unify it across GCP and AWS in a new PR.

bgeihsgt: Although, I'd need to do more thinking on pros/cons of the approach you're suggesting.

packages/terraform-modules-public/aws/testnet/main.tf
jcortejoso: 100% agree.

packages/terraform-modules-public/aws/testnet/modules/vpc/main.tf
jcortejoso: Nice.

packages/terraform-modules-public/aws/example/variables.tf
jcortejoso: Maybe I am missing something, but the enode is the public key. So if the user provides as an input the proxy's nodekey (private part), the enode would be also known in advance.

packages/terraform-modules-public/aws/example/variables.tf
tkporter: Personally I'm fine with the way it is, especially given that it's similar to the GCP setup. I think Terraform is a little painful when it comes to managing multiple deployments and this might encourage hot standby validators. I'm up for bringing others into the loop on this though. @bgeihsgt have you thought any more about this?

packages/terraform-modules-public/aws/example/variables.tf
tkporter: So how do you deploy it the way it is? Do you target only proxies first? I still think this would be good to have, and we could explicitly set the proxy private key. The enode is the public key minus the first two characters, docs are being written up for this too. I think this is how we do the GCP and azure infra right? Wdyt @jcortejoso ?

packages/terraform-modules-public/aws/example/variables.tf
tkporter: @jcortejoso do you have opinions?

packages/terraform-modules-public/aws/example/variables.tf
jcortejoso: @tkporter Yes exactly what I mean. As for [GCP](https://github.com/celo-org/celo-monorepo/blob/master/packages/terraform-modules-public/testnet/variables.tf#L114) and [Azure](https://github.com/celo-org/celo-monorepo/blob/master/packages/azure/template.json#L1038) modules, if the user provides a private key for the proxy nodekey, the enode will be explicitly known and will be the uncompressed public key minus the initial `04` characters. It is explained here https://github.com/celo-org/celo-monorepo/blob/master/packages/terraform-modules-public/example/README.md#ethereum-accounts

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/install-chrony.sh
jcortejoso: Cool!

packages/terraform-modules-public/aws/example/variables.tf
jcortejoso: As @tkporter pointed I would say with terraform is easier to manage all your stuff from a single point, although is achievable with workspaces or duplicating 
For me what would makes sense is having a more static package and offer as an fast&easy deployment way (with cloudformation templates for example), and the terraform module offering a more configurable option if people want to customize the deployment a bit more.

jcortejoso: Anyway I would say that if we think is better to build a "validator stack" module, would be better to address this in a future PR in conjunction with the GCP module.

codyborn: Sounds good, thank all!

codyborn (approved):

packages/terraform-modules-public/aws/example/variables.tf
bgeihsgt: Ah, very interesting. Thanks for this - I didn't know enodes were deterministic in this way. Today, this approach requires you to stand up the proxy and then use params from the proxy in the validator, so you have to run terraform apply multiple times to stand up the cluster (not ideal).

I will update this to allow it all to stand up in one apply call.

bgeihsgt: (The original implementation was modeled after the Running a Validator docs, which made me assume a proxy needed to be running before a validator config could be known)

bgeihsgt:
Hey all! Wanted to say I haven’t forgotten about this :) 

Just haven’t had the time to make the last fix to make it more streamlined. Still on my radar to finish before mainnet though.

bgeihsgt:
Ok @tkporter / @jcortejoso - I am trying to wrap this up.

I'm trying to allow it to work on a single `terraform apply` instead of two. I am able to successfully set the `nodekey` on the proxy so I can predict the enode URL.

The problem I'm having is easily getting the full private & public key hex values when I'm generating that account. Things I've tried:

1. Generate an account using geth `account new`. Use two npm libraries (`keythereum` and `elliptic`) to derive the private and public keys from the address. Use the private key for the nodekey. Use the public key minus the first two numbers to predict the enode address. This works great, but is cumbersome to get the key hex values. Is there an easier way to get these from a keystore file? Perhaps we should add an `account:show-keys` command to the CLI?
2. Generate a private/public keypair using `celocli account:new`. Use the returned private key for the nodekey. The problem here is the public key that's printed out from the CLI. It appears to only have half of what would be on the final enode URL. Is this some kind of shorthand? Would this value work in an enode URL?

Thanks for your help!

bgeihsgt:
@tkporter / @jcortejoso Got the proxy parameter automation done, so only one `terraform apply` necessary now. 

Stuff left that I will submit as separate PRs:

1. Adding a `mainnet` folder to do terraform mainnet automation.
2. Updating attestation setup to match new Baklava instructions. I didn't get to that yet and it's lower priority than getting mainnet working.

That work for you to submit those as separate PRs?

bgeihsgt:
Wanted to ping on this - would like to get this merged soon so I can submit a PR for mainnet. Would that be possible?

bgeihsgt:
@tkporter / @jcortejoso Any chance you guys could take a look at this and get it merged?

tkporter (approved):
Thanks for your patience & for all of this, can't wait for the mainnet PR! I think providing the proxy enode makes it a lot easier to interact with

I'm not sure what the deal is with celocli giving a shorter public key. I'll ask around :)

PR #2913, opened 2020-02-24
[H02] Commissions can deceive members of a validator group

mcortesi:
### Description

Add a time lock so that commission updates can only take effec after an X block delay

### Tested

 * contractkit , cli & solidity tests

### Related issues

- Fixes celo-org/celo-labs#262

### Backwards compatibility

* Changes validator ABI
* Requires new version of cli & contractkit


packages/protocol/migrationsConfig.js
```diff
@@ -125,6 +125,7 @@ const DefaultConfig = {
       adjustmentSpeed: 0.1,
     },
     membershipHistoryLength: 60,
+    commissionUpdateDelay: 30000,
```
asaj: Consider a number that corresponds to a integer number of days. Suggest 3
```suggestion
    commissionUpdateDelay: 51840,  // Approximately 3 days
```

asaj (approved):
Very nice!

PR #2918, opened 2020-02-24
[Wallet] Fix invites because of different dev bundleId on Android and iOS

i1skn:
### Description

Fix broken invites because of different dev bundleId on Android and iOS

### Tested

iOS and Android devices

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/web3/saga.ts
+++ b/packages/mobile/src/web3/saga.ts
@@ -20,7 +20,7 @@ import { UNLOCK_DURATION } from 'src/geth/consts'
 import { deleteChainData, stopGethIfInitialized } from 'src/geth/geth'
 import { gethSaga, waitForGethConnectivity } from 'src/geth/saga'
 import { gethStartedThisSessionSelector } from 'src/geth/selectors'
-import { navigate, navigateToError } from 'src/navigator/NavigationService'
+import { navigate, navigateBack, navigateToError } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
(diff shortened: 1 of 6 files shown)


packages/mobile/src/web3/saga.ts
```diff
@@ -343,6 +343,7 @@ export function* unlockAccount(account: string) {
     }
 
     const pincode = yield call(getPincode)
+    navigateBack()
```
jmrossy: I dont't think this belongs here. We use `unlockAccount` in `getConnectedUnlockedAccount` which is used in many places. I don't think navigating back is always the right behavior.

jmrossy (changes requested):

packages/mobile/src/web3/saga.ts
i1skn: Well, previously it was navigation back all the time it was asking for a PIN(PTAL here https://github.com/celo-org/celo-monorepo/pull/2648/files#diff-fdbb05a987fa942d9270541910265e9c in methods `onPressConfirm` and `onPressCancel`). Then when we introduced "Lock wallet with PIN" and I changed it, and forget to update it here. As far as I got when `getConnectedUnlockedAccount` is called and PIN is not cached - the PIN window appears from nowhere and then it goes to nowhere, when user has entered correct PIN or pressed back.

i1skn: Correct, had addressed it.

jmrossy (approved):

PR #2925, opened 2020-02-24
Release Gold migration script

### LINKED ISSUE #2516: Vesting schedules migration complete
### Expected Behavior

Celo Gold vesting schedules can be written to json file, migration that reads these schedules and deploys vesting instances.

May be easiest to just create a command that takes a json file and deploys vesting instances, idempontently.

lucasege:
### Description

Added deploy script for ReleaseGold contracts. Intended usage is for the script to be run for each grant-ing entity (LLC, Reg D, RAU, Coinlist), consuming a list of grants and creating ReleaseGold contracts for each.

Included is a sample construct of a grant .json file, intended to be removed before merge.

Each `ReleaseGold` contract is proxied and owned by a two-of-two `MultiSig` of the `beneficiary` and the given entity.

### Tested

Tested on local ganache chain.

### Other changes

Removed `ReleaseGoldFactory`, added proxy for `ReleaseGold`.

### Related issues

- Fixes #2516


packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -0,0 +1,574 @@
+pragma solidity ^0.5.3;
```
lucasege: No changes to `ReleaseGold`, just changed the name from `ReleaseGoldInstance` to `ReleaseGold` since `instance` is redundant

packages/protocol/contracts/governance/interfaces/IReleaseGold.sol
```diff
@@ -0,0 +1,27 @@
+pragma solidity ^0.5.3;
```
lucasege, line 1: Also name change

packages/protocol/test/governance/release_gold.ts
```diff
@@ -25,9 +25,8 @@ import {
   MockValidatorsInstance,
```
lucasege, line 27: Summary of changes:
Name changes and removing factory code (tests and utilities).

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,75 @@
...
+    const releases = []
+    const handleJSONFile = async (err, data) => {
+      if (err) {
+        throw err
+      }
+      data = JSON.parse(data)
+      const entity = data.entity
+      const releaseGoldConfigs = data.grants
+      for (const releaseGoldConfig of releaseGoldConfigs) {
+        const releaseStartTime = new Date(releaseGoldConfig.releaseStartTime)
```
lucasege: For reviewer: any suggestions to avoid manually pulling each element out of the struct? TS does not seem to allow passing an interface/dict in for a list of parameters.
I've tried using `.apply()` but that doesn't seem to work.

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,75 @@
...
+          refundAddress,
+          subjectToLiquidityProvision,
+          initialDistributionPercentage,
+          canValidate,
+          canVote,
+          registry.address
+        )
+        const releaseGoldProxyInstance = await ReleaseGoldProxy.new()
+        await releaseGoldProxyInstance._setImplementation(releaseGoldInstance.address)
+        const multiSigInstance = await Multisig.new()
```
lucasege: @Reviewer:
Multisig is traditionally proxied, but I haven't done that here because another level of indirection seemed too complicated here, let me know if I need to add it.

lucasege:
@reviewer:
I'm concerned about the management of the multisig: is simply printing the address at the end of execution sufficient? Obviously everything is on chain, but still questioning. 
Also, the interaction between this and being allowed to change a beneficiary #2927 : I'm not sure how best to implement this currently. It seems to me that `setBeneficiary` will need to have a reference to the multisig to signal that the multisig must change one of its owners. Not sure how this will work, or if it can.

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,75 @@
...
+  ReleaseGoldProxyContract,
+} from 'types'
+
+module.exports = async (callback: (error?: any) => number) => {
+  try {
+    const argv = require('minimist')(process.argv.slice(2), {
+      string: ['network', 'grants'],
+    })
+    const registry = await getDeployedProxiedContract<RegistryInstance>('Registry', artifacts)
+    const Multisig: MultiSigContract = artifacts.require('MultiSig')
```
asaj: Consider defining a `ReleaseGoldMultiSig` that just inherits from `MultiSig`

packages/protocol/scripts/truffle/deploy_release_contracts.ts
asaj: Yeah I was wondering about that myself. I think it could be worth it, with the Proxy owning itself.

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,75 @@
...
+    const ReleaseGold: ReleaseGoldContract = artifacts.require('ReleaseGold')
+    const ReleaseGoldProxy: ReleaseGoldProxyContract = artifacts.require('ReleaseGoldProxy')
+    const releases = []
+    const handleJSONFile = async (err, data) => {
+      if (err) {
+        throw err
+      }
+      data = JSON.parse(data)
+      const entity = data.entity
```
asaj: Is this needed?

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,75 @@
...
+          subjectToLiquidityProvision,
+          initialDistributionPercentage,
+          canValidate,
+          canVote,
+          registry.address
+        )
+        const releaseGoldProxyInstance = await ReleaseGoldProxy.new()
+        await releaseGoldProxyInstance._setImplementation(releaseGoldInstance.address)
+        const multiSigInstance = await Multisig.new()
+        await multiSigInstance.initialize([entity, beneficiary], 2)
```
asaj: Use `releaseOwner` instead of `entity`

packages/protocol/scripts/truffle/releaseGoldContracts.json
```diff
@@ -0,0 +1,20 @@
+{
+  "entity": "0x000000000000000000000000000000000000ce12",
+  "grants": [
+    {
+    "releaseStartTime": "02-24-2020",
+    "releaseCliffTime": 1,
```
asaj, line 4: What does `1` mean in this context?

packages/protocol/scripts/truffle/releaseGoldContracts.json
```diff
@@ -0,0 +1,20 @@
...
+    "numReleasePeriods": 1,
+    "releasePeriod": 100000000,
+    "amountReleasedPerPeriod": 1,
+    "revocable": true,
+    "beneficiary": "0x000000000000000000000000000000000000ce11",
+    "releaseOwner": "0x000000000000000000000000000000000000ce11",
+    "refundAddress": "0x000000000000000000000000000000000000ce11",
+    "subjectToLiquidityProvision": true,
+    "initialDistributionPercentage": 1000,
```
asaj: nit: "Percentage" is confusing here since we add an extra decimal of precision

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,75 @@
...
+        const amountReleasedPerPeriod = releaseGoldConfig.amountReleasedPerPeriod
+        const revocable = releaseGoldConfig.revocable
+        const beneficiary = releaseGoldConfig.beneficiary
+        const releaseOwner = releaseGoldConfig.releaseOwner
+        const refundAddress = releaseGoldConfig.refundAddress
+        const subjectToLiquidityProvision = releaseGoldConfig.subjectToLiquidityProvision
+        const initialDistributionPercentage = releaseGoldConfig.initialDistributionPercentage
+        const canValidate = releaseGoldConfig.canValidate
+        const canVote = releaseGoldConfig.canVote
+        const releaseGoldInstance = await ReleaseGold.new(
```
asaj: How is this contract funded with gold?

packages/protocol/scripts/truffle/releaseGoldContracts.json
lucasege: This was all dummy data, ignore

packages/protocol/scripts/truffle/releaseGoldContracts.json
lucasege: Ratio? factor? Can't think of a good word for this
Going with ratio for now with mentioning ratio over 1000

packages/protocol/scripts/truffle/deploy_release_contracts.ts
nambrot: I assume you mean object destructuring? http://budiirawan.com/typescript-destructuring-object/

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,94 @@
...
+    const ReleaseGoldMultiSig: ReleaseGoldMultiSigContract = artifacts.require(
+      'ReleaseGoldMultiSig'
+    )
+    const ReleaseGoldMultiSigProxy: ReleaseGoldMultiSigProxyContract = artifacts.require(
+      'ReleaseGoldMultiSigProxy'
+    )
+    const ReleaseGold: ReleaseGoldContract = artifacts.require('ReleaseGold')
+    const ReleaseGoldProxy: ReleaseGoldProxyContract = artifacts.require('ReleaseGoldProxy')
+    const releases = []
+    const handleJSONFile = async (err, data) => {
```
nambrot, line 32: Are we doing no validation on the input? Seems kinda scary to me.

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,94 @@
...
+          registry.address
+        )
+        const releaseGoldProxyInstance = await ReleaseGoldProxy.new()
+        const releaseGoldMultiSigProxyInstance = await ReleaseGoldMultiSigProxy.new()
+        await goldToken.transfer(
+          releaseGoldProxyInstance.address,
+          amountReleasedPerPeriod * numReleasePeriods,
+          { from: releaseOwner }
+        )
+        // const initializeAbi = (releaseGoldProxyInstance as any).abi.find(
```
nambrot: Remove?

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,94 @@
...
+          refundAddress,
+          subjectToLiquidityProvision,
+          initialDistributionPercentage,
+          canValidate,
+          canVote,
+          registry.address
+        )
+        const releaseGoldProxyInstance = await ReleaseGoldProxy.new()
+        const releaseGoldMultiSigProxyInstance = await ReleaseGoldMultiSigProxy.new()
+        await goldToken.transfer(
```
nambrot: Should we only transfer after the multisig owner is setup?

packages/protocol/scripts/truffle/deploy_release_contracts.ts
nambrot: Maybe we should add a prompt for each grant to confirm?

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,94 @@
...
+          releaseGoldMultiSigInstance.address
+        )
+        releases.push([
+          releaseGoldConfig.beneficiary,
+          releaseGoldProxyInstance.address,
+          releaseGoldMultiSigProxyInstance.address,
+        ])
+      }
+      // tslint:disable-next-line: no-console
+      console.log(releases)
```
nambrot, line 91: Maybe leave an option to write it into a file?

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,94 @@
...
+        // await releaseGoldProxyInstance._setAndInitializeImplementation(releaseGoldInstance.address, callData)
+        await releaseGoldProxyInstance._setImplementation(releaseGoldInstance.address)
+        const releaseGoldMultiSigInstance = await ReleaseGoldMultiSig.new()
+        await releaseGoldMultiSigInstance.initialize([releaseOwner, beneficiary], 2)
+        await releaseGoldProxyInstance._transferOwnership(releaseGoldMultiSigInstance.address)
+        await releaseGoldMultiSigProxyInstance._setImplementation(
+          releaseGoldMultiSigInstance.address
+        )
+        releases.push([
+          releaseGoldConfig.beneficiary,
```
nambrot, line 69: Having the TX hashes for prosperity also seem useful, also in a dictionary instead of a tuple

packages/protocol/scripts/truffle/deploy_release_contracts.ts
asaj: +1 to that suggestion, though it can wait for a follow-on PR

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,105 @@
...
+          releaseGoldConfig.canValidate,
+          releaseGoldConfig.canVote,
+          registry.address,
+        ]
+        const releaseGoldMultiSigInitializeArgs = [
+          [releaseGoldConfig.releaseOwner, releaseGoldConfig.beneficiary],
+          2,
+          2,
+        ]
+        const initializeProxyContract = async (contractInstance, proxyInstance, args) => {
```
asaj: Can you use `setInitialProxyImplementation`?

packages/protocol/scripts/truffle/deploy_release_contracts.ts
asaj: +1

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,105 @@
...
+          releaseGoldMultiSigInstance,
+          releaseGoldMultiSigProxyInstance,
+          releaseGoldMultiSigInitializeArgs
+        )
+        const multiSigTxHash = await initializeProxyContract(
+          releaseGoldInstance,
+          releaseGoldProxyInstance,
+          releaseGoldInitializeArgs
+        )
+        await releaseGoldInstance.transferOwnership(releaseGoldMultiSigProxyInstance.address)
```
asaj: This isn't doing what it should be doing because it's not being called via the proxy

packages/protocol/scripts/truffle/deploy_release_contracts.ts
lucasege: Is this handled now? Not sure how the wording is meant

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,100 @@
...
+    )
+    const ReleaseGold: ReleaseGoldContract = artifacts.require('ReleaseGold')
+    const ReleaseGoldProxy: ReleaseGoldProxyContract = artifacts.require('ReleaseGoldProxy')
+    const releases = []
+    const handleJSONFile = async (err, data) => {
+      if (err) {
+        throw err
+      }
+      for (const releaseGoldConfig of JSON.parse(data)) {
+        const releaseGoldMultiSigProxyInstance = await ReleaseGoldMultiSigProxy.new()
```
asaj: nit: consider dropping the `Instance` suffix

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,100 @@
...
+          releaseGoldMultiSigInstance,
+          releaseGoldMultiSigProxyInstance,
+          'ReleaseGoldMultiSig',
+          [releaseGoldConfig.releaseOwner, releaseGoldConfig.beneficiary],
+          2,
+          2
+        )
+        const releaseGoldProxyInstance = await ReleaseGoldProxy.new()
+        const releaseGoldInstance = await ReleaseGold.new()
+        const gold = new BigNumber(
```
asaj: nit: For readability, consider:
```
const goldTotal = web3.utils.toWei(releaseGoldConfig.amountReleasedPerPeriod * releaseGoldConfig.numReleasePeriods)
```

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,100 @@
...
+          web3.utils.toWei(releaseGoldConfig.amountReleasedPerPeriod.toString())
+        )
+        await goldToken.transfer(
+          releaseGoldProxyInstance.address,
+          gold.multipliedBy(releaseGoldConfig.numReleasePeriods),
+          { from: releaseGoldConfig.releaseOwner }
```
asaj: So perhaps counterintuitively, the release owner will actually not be the same address, or at least should not *need* to be the same address, as the address that deploys and funds the contracts.

I don't think you need to explicitly set `from` here, since it will be pulled from the truffle config automatically

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,100 @@
...
+        const multiSigTxHash = await _setInitialProxyImplementation(
+          web3,
+          releaseGoldMultiSigInstance,
+          releaseGoldMultiSigProxyInstance,
+          'ReleaseGoldMultiSig',
+          [releaseGoldConfig.releaseOwner, releaseGoldConfig.beneficiary],
+          2,
+          2
+        )
+        const releaseGoldProxyInstance = await ReleaseGoldProxy.new()
```
asaj: nit: ditto on dropping `Instance`

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,100 @@
...
+        const releaseGoldTxHash = await _setInitialProxyImplementation(
+          web3,
+          releaseGoldInstance,
+          releaseGoldProxyInstance,
+          'ReleaseGold',
+          new Date(releaseGoldConfig.releaseStartTime).getTime() / 1000,
+          releaseGoldConfig.releaseCliffTime,
+          releaseGoldConfig.numReleasePeriods,
+          releaseGoldConfig.releasePeriod,
+          web3.utils.toHex(gold),
```
asaj: nit: not needed if you take the suggestion above

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,100 @@
...
+          releaseGoldConfig.beneficiary,
+          releaseGoldConfig.releaseOwner,
+          releaseGoldConfig.refundAddress,
+          releaseGoldConfig.subjectToLiquidityProvision,
+          releaseGoldConfig.initialDistributionRatio,
+          releaseGoldConfig.canValidate,
+          releaseGoldConfig.canVote,
+          registry.address
+        )
+        const releaseGoldAtProxy = await ReleaseGold.at(releaseGoldProxyInstance.address)
```
asaj: nit: Consider `proxiedReleaseGold`

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -0,0 +1,100 @@
...
+        await releaseGoldAtProxy.transferOwnership(releaseGoldMultiSigProxyInstance.address)
+        await releaseGoldProxyInstance._transferOwnership(releaseGoldMultiSigProxyInstance.address)
+        await releaseGoldMultiSigProxyInstance._transferOwnership(
```
asaj: nit: Consider transferring ownership of the multisig proxy after you call `_setInitialProxyImplementation`

packages/protocol/scripts/truffle/deploy_release_contracts.ts
asaj: +1, can also be done in a follow-up

packages/protocol/scripts/truffle/deploy_release_contracts.ts
asaj: Actually maybe this is okay, since the proxy will be owned by the deployer until ownership is transferred to the multisig

asaj (approved):
🚢

PR #2926, opened 2020-02-24
[M06] Prove control of walletAddress

nambrot:
### Description

Right now, any account could have claimed any wallet address. However, this does not actually prove control of the wallet address, meaning that anyone could claim a high-balance wallet and appear to with a higher balance than reality. This PR requires a proof of possession unless its the zero address or msg.sender

### Tested

- Unit tests

### Other changes

- Minor name changes

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/269

### Backwards compatibility

- Not backwards compatible for smart contract users


packages/contractkit/src/wrappers/Accounts.ts
```diff
@@ -317,8 +313,42 @@ export class AccountsWrapper extends BaseWrapper<Accounts> {
...
+          walletAddress,
+          proofOfPossession.v,
+          proofOfPossession.r,
+          proofOfPossession.s
+        )
+      )
+    } else {
+      return toTransactionObject(
+        this.kit,
+        this.contract.methods.setAccount(
```
asaj, line 340: Why can we call `setAccount` here without a proof of possession?

packages/contractkit/src/wrappers/Accounts.ts
```diff
@@ -336,7 +366,27 @@ export class AccountsWrapper extends BaseWrapper<Accounts> {
...
+        this.kit,
+        this.contract.methods.setWalletAddress(
+          walletAddress,
+          proofOfPossession.v,
+          proofOfPossession.r,
+          proofOfPossession.s
+        )
+      )
+    } else {
+      return toTransactionObject(
```
asaj, line 384: Ditto

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -107,12 +117,21 @@ contract Accounts is IAccounts, Ownable, ReentrancyGuard, Initializable, UsingRe
...
    * @dev Wallet address can be zero. This means that the owner of the wallet
    *  does not want to be paid directly without interaction, and instead wants users to
    * contact them, using the data encryption key, and arrange a payment.
+   * @dev v, r, s constitute `signer`'s signature on `msg.sender` (unless the wallet address
+   *      is 0x0 or msg.sender).
    */
-  function setWalletAddress(address walletAddress) public {
+  function setWalletAddress(address walletAddress, uint8 v, bytes32 r, bytes32 s) public {
     require(isAccount(msg.sender), "Unknown account");
+    if (!(walletAddress == msg.sender || walletAddress == address(0x0))) {
```
asaj, line 136: Assuming this is why, but there's nothing in the contractkit code that prevents users from calling `setWalletAddress` with a null signature but a non-zero, non-msg.sender walletAddress

asaj (approved):

packages/protocol/contracts/common/Accounts.sol
nambrot: Indeed, but they would just have the TX reverted on them? It sounds like you are saying we should do a check in CK itself?

nambrot: It's kind of hard to do because we do not have access to `msg.sender` easily, so I would advocate to keep it in the responsibility of the caller?

packages/contractkit/src/wrappers/Accounts.test.ts
```diff
@@ -86,4 +86,15 @@ testWithGanache('Accounts Wrapper', (web3) => {
     const sig = await getParsedSignatureOfAddress(account, signer)
     await accountsInstance.authorizeValidatorSignerAndBls(signer, sig, newBlsPublicKey, newBlsPoP)
   })
+
+  test('SBAT set the wallet address to the caller', async () => {
+    await accountsInstance.createAccount()
+    await accountsInstance.setWalletAddress(accounts[0])
+  })
+
+  test('SBAT set the wallet address to a different wallet address', async () => {
```
nategraf, line 95: A test that it is not possible to set the wallet address without the PoP would be good.

packages/contractkit/src/wrappers/Accounts.ts
```diff
@@ -317,8 +313,42 @@ export class AccountsWrapper extends BaseWrapper<Accounts> {
    * @param name A string to set as the name of the account
    * @param dataEncryptionKey secp256k1 public key for data encryption. Preferably compressed.
    * @param walletAddress The wallet address to set for the account
+   * @param
```
nategraf: I assume this param is supposed to be `proofOfPossession`.
```suggestion
   * @param proofOfPossession Signature from the wallet address key over the sender's address
```

nategraf (approved):
Minor comments

packages/contractkit/src/wrappers/Accounts.ts
nambrot: We cannot, and the tests failed appropriately

nambrot: I rely on the estimateGas failing here as there is no easy access to `msg.sender`

packages/contractkit/src/wrappers/Accounts.test.ts
nambrot: Added

PR #2927, opened 2020-02-24
Allow ReleaseGoldInstance beneficiary to transfer beneficiary

nambrot:
### Description

This PR allows a ReleaseGoldInstance beneficiary to transfer the beneficiary to another address

### Tested

- Unit tests


packages/protocol/test/governance/release_gold.ts
```diff
@@ -377,6 +380,44 @@ contract('ReleaseGold', (accounts: string[]) => {
...
+      releaseGoldInstance = await ReleaseGoldInstance.at(releaseGoldInstanceAddress)
+    })
+
+    it('should set a new beneficiary as the old beneficiary', async () => {
+      await releaseGoldInstance.setBeneficiary(newBeneficiary, { from: beneficiary })
+      const actualBeneficiary = await releaseGoldInstance.beneficiary()
+      assertSameAddress(actualBeneficiary, newBeneficiary)
+    })
+
+    it('should revert when setting a new beneficiary from the releas owner', async () => {
```
asaj: typo

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -198,6 +199,7 @@ contract ReleaseGoldInstance is UsingRegistry, ReentrancyGuard, IReleaseGoldInst
 
     setRegistry(registryAddress);
     beneficiary = _beneficiary;
+    emit BeneficiarySet(beneficiary);
```
asaj: Please define a `_setBeneficiary` that doesn't have the modifier, and then use that here, and then `setBeneficiary` just wraps `_setBeneficiary` with the modifier

asaj (approved):

lucasege:
How will this play in with `ReleaseGold` being deployed under a 2-of-2 multisig (one address is the beneficiary, one is the grant-ing entity)? Seems to me that the beneficiary multisig address will need to be swapped as well, which will probably require `ReleaseGold` to have a reference to the `Multisig`.
Maybe there are other solutions (maybe the multisig can listen for the `setBeneficiary` event some how)?
Any ideas?

nambrot:
I think the only reasonable way will be to have the beneficiary be responsible to rotate themselves with the help of the release manager. Ideally we could have a modified multisig that reads one of the owners from `ReleaseGold`, but it's not worth it the security risk IMO

asaj:
> I think the only reasonable way will be to have the beneficiary be responsible to rotate themselves with the help of the release manager. Ideally we could have a modified multisig that reads one of the owners from ReleaseGold, but it's not worth it the security risk IMO


In that case, to avoid the beneficiary accidentally swapping addresses without updating the multisig, should we require `setBeneficiary` to be `onlyMultisig`?

nambrot:
Great idea, @lucasege would you mind making the change once this goes in?

lucasege:
> Great idea, @lucasege would you mind making the change once this goes in?

Yeah can do, if it gets automerged in a reasonable amount of time

packages/protocol/contracts/governance/ReleaseGoldInstance.sol
```diff
@@ -254,6 +256,16 @@ contract ReleaseGoldInstance is UsingRegistry, ReentrancyGuard, IReleaseGoldInst
     emit DistributionLimitSet(beneficiary, maxDistribution);
   }
 
+  /**
+   * @notice Sets the beneficiary of the instance
+   * @param newBeneficiary The address of the new beneficiary
+   */
+  function setBeneficiary(address payable newBeneficiary) external onlyBeneficiary {
```
codyborn, line 262: Similarly, should we allow updating of the refundAddress?

codyborn, line 262: Do we have tests for this contract?

nambrot: Ideally, probably yes, though the likelihood of key rotation for that refund address seem to be much lower @asaj

nambrot: Yes, does it not show in the diff for you?

codyborn: I see it, thanks!

codyborn (approved):

PR #2931, opened 2020-02-25
[L06] Lack of validation may produce unexpected state in the distribution of epoch rewards

m-chrzan:
### Description

This PR adds a sanity check to make sure a validator for whom `distributeEpochPaymentsFromSigner` is called is registered with a validator group.

### Tested

Tests still pass.

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/306

### Backwards compatibility

Yes.


asaj (approved):

packages/protocol/contracts/governance/Validators.sol
```diff
@@ -451,6 +451,7 @@ contract Validators is
     // The group that should be paid is the group that the validator was a member of at the
     // time it was elected.
     address group = getMembershipInLastEpoch(account);
+    require(group != address(0), "Validator not registered with a group");
```
nategraf, line 454: If this line fails, reward distribution will stop. I've created https://github.com/celo-org/celo-blockchain/pull/881 to help address that and it should be merged along with this.

PR #2932, opened 2020-02-25
[Wallet] Fix PIN entering screen, when account is not created yet

i1skn:
### Description

This fixed the following bug:
1) Choose lang -> enter phone + name -> enter pin two times
2) Close the app or wait until cached PIN is expired
3) Paste invite code
4) **PIN screen would hang** (this PR fixes this)

### Tested

Ran on Android device

### Backwards compatibility

Yes


jeanregisser (approved):
Nice find! 👍

PR #2940, opened 2020-02-25
[Wallet] Invite text message deeplinking for iOS

### LINKED ISSUE #1729: Invite text message deeplinking SBAT work on iOS

i1skn:
### Description

Introduces support for deeplinking invite code for iOS app and changes the way how it was done before in Android (using `referrer-data`). Now both platforms will use Firebase Dynamic Links. For a user it means, that if app was opened/downloaded from a deeplink - there is no need to enter invite code.

### Tested

Tested, when app is installed on Android with `adb shell am start -a "android.intent.action.VIEW" -d "<url>"` and on iOS with `xcrun simctl openurl booted '<url>`. Can not test, when app is not installed, so I hope it's gonna work. Worst case - it won't, but paste invite would be still possible.

### Other changes

* Remove `react-native-version-check` from deps, cause there is no more need for it.
* `clickTimestamp` and `installTimestamp` inside `referrer-data` were never used, so I removed them together with `react-native-install-referrer` altogether. Let me know if there are plans to use it in the near future - will change bring them back.
* Move to `https://l.celo.org/abc123` instead of `https://celol.page.link/abc123`.

### Related issues

- Fixes #1729

### Backwards compatibility

Yes


jeanregisser (approved):
Awesome! 🚀 

> Remove react-native-version-check from deps, cause there is no more need for it.

Because firebase figures out the PlayStore link automatically?

packages/mobile/package.json
```diff
@@ -91,7 +91,6 @@
     "react-native-fs": "^2.14.1",
     "react-native-gesture-handler": "^1.5.2",
     "react-native-geth": "https://github.com/celo-org/react-native-geth#6f993d9",
-    "react-native-install-referrer": "git://github.com/celo-org/react-native-install-referrer#343bf3d",
```
jmrossy, line 94: ❤️ for removing deps

packages/mobile/src/invite/EnterInviteCode.tsx
```diff
@@ -67,19 +67,18 @@ export class EnterInviteCode extends React.Component<Props, State> {
   }
 
   async componentDidMount() {
-    await this.checkIfValidCodeInClipboard()
-    await this.checkForReferrerCode()
+    await this.checkInviteCode()
   }
 
-  checkForReferrerCode = async () => {
+  checkInviteCode = async () => {
```
jmrossy: Nit: this name makes me think it's checking an invite code (like for validity). how about `checkForInviteCode`?

packages/mobile/src/invite/saga.ts
```diff
@@ -85,20 +84,18 @@ export function getInvitationVerificationFeeInWei() {
...
-
-  // iOS part
-  const appStoreId = await getAppStoreId(bundleId)
-  const appStoreUrl = await VersionCheck.getAppStoreUrl({ appID: appStoreId })
+  // trying to fetch appStoreId needed to build a dynamic link
+  let appStoreId
+  try {
+    appStoreId = await getAppStoreId(bundleId)
+  } catch (error) {
+    Logger.error(TAG, 'Failed to load AppStore ID: ' + error.toString())
```
jmrossy, line 93: what happens in this case? Can we still generate a link even without appStoreId?

jmrossy (commented):
Looks great! Just two small things

packages/mobile/src/invite/saga.ts
i1skn: Yes, it would just open `https://celo.org/build/wallet` on iOS instead of AppStore link

packages/mobile/src/invite/EnterInviteCode.tsx
i1skn: Sounds good!

i1skn:
> Because firebase figures out the PlayStore link automatically?

@jeanregisser Yeah, I assume that firebase under the hood is doing the same thing, what we were doing manually with `referrer-data`.

jmrossy (approved):

PR #2947, opened 2020-02-25
Remove codecov for protocol until coverage working again

asaj:
### Description

Remove codecov for protocol until coverage working again
### Tested

No
### Other changes

None
### Related issues

None
### Backwards compatibility

Yes


jeanregisser (approved):
LGTM 👍
Wondering, should we comment the section instead of deleting and leave some info about why we’re disabling for now?

PR #2956, opened 2020-02-25
[N07] Add docs about SIM swaps

nambrot:
### Description

Updates the doc with realistic attacks onto the Attestations smart contract. To reiterate, the contract itself does not protect against any attacks or makes claims over the legitimate owner of a phone number. It solely records attestations, as given by validators.

Fixes https://github.com/celo-org/celo-labs/issues/284


packages/docs/celo-codebase/protocol/identity/README.md
```diff
@@ -16,7 +16,7 @@ When Alice receives the text message, she can take that signed message to the `A
...
+The `Attestations` contract records all attestations of a phone number to any number of addresses. That for example could happen when a user loses their private key and wants to map a new wallet address. However, it could also happen through the collusion of a validator with Alice. Therefore, it is important that clients of the identity protocol highlight possible conflicting attestations. Attestations, albeit unlikely, can be added without the "legitimate" owner of the phone number, whether by a service provider, collusion with a validator or SIM swapping. Notably, completing "malicious" attestations does not lead to a loss of funds, as the private key is still the neccessary and sufficient condition for transactions of an account. In general, the more attestations an address has received, the more likely it is to map to the valid owner of a phone number based upon the existing trust assumptions for validators for consensus.
```
nategraf: We have to be careful not to sounds dismissive of legitimate security concerns, especially because we will be encouraging users to identify each other via phone number and an average user will probably never directly check the number of attestations or the underlying address (although the app should have some checks that it can surface through the design)

I'd suggest we expand on the threat and consequence of each validator collusion and SIM swapping separately. Maybe something like this, but feel free to modify or reject this suggestion.
```suggestion
The `Attestations` contract records all attestations of a phone number to any number of addresses. That for example could happen when a user loses their private key and wants to map a new wallet address. However, it could also happen through the collusion of a validator with Alice. Therefore, it is important that clients of the identity protocol highlight possible conflicting attestations. 
Some risk exists for attestations to be added without the without the permission of the "legitimate" owner of the phone number. One such risk is that the phone service provider or [SIM swap](https://en.wikipedia.org/wiki/SIM_swap_scam) attacker could take control of the phone number and complete a number of attestations. Another risk is that a sufficient number of Attestation Service providers may collude to complete fake attestations. Notably, completing malicious attestations does not lead to a loss of funds, as the private key is still the necessary and sufficient condition for transactions of an account. However, without proper care, future senders may be tricked into sending funds to the newly associated address. In general, then number and age of attestations for an address, should be taken into account to identify the valid owner of a phone number.
```

nambrot:
@nategraf I committed your suggestion. I agree that it is a real risk, however the following paragraph mentions that clients are expected to take on that responsibility of properly communicating that risk to their users. The protocol itself cannot do that

packages/docs/celo-codebase/protocol/identity/README.md
```diff
@@ -16,7 +16,8 @@ When Alice receives the text message, she can take that signed message to the `A
...
+Some risk exists for attestations to be added without the without the permission of the "legitimate" owner of the phone number. One such risk is that the phone service provider or [SIM swap](https://en.wikipedia.org/wiki/SIM_swap_scam) attacker could take control of the phone number and complete a number of attestations. Another risk is that a sufficient number of Attestation Service providers may collude to complete fake attestations. Notably, completing malicious attestations does not lead to a loss of funds, as the private key is still the necessary and sufficient condition for transactions of an account. However, without proper care, future senders may be tricked into sending funds to the newly associated address. In general, then number and age of attestations for an address, should be taken into account to identify the valid owner of a phone number.
```
asaj:
```suggestion
Some risk exists for attestations to be added without the without the permission of the "legitimate" owner of the phone number. One such risk is that the phone service provider or [SIM swap](https://en.wikipedia.org/wiki/SIM_swap_scam) attacker could take control of the phone number and complete a number of attestations. Another risk is that a sufficient number of Attestation Service providers may collude to complete fake attestations. Notably, completing malicious attestations does not lead to a loss of funds, as the private key is still the necessary and sufficient condition for transactions of an account. However, without proper care, future senders may be tricked into sending funds to the newly associated address. In general the number and age of attestations for an address should be taken into account to identify the valid owner of a phone number.
```

asaj (approved):

PR #2964, opened 2020-02-26
Added --proxy.allowprivateip to validators opts for k8s testnet

jcortejoso:
### Description

Added --proxy.allowprivateip to validators opts for k8s testnet deployments

### Tested
On k8s testnet

### DIFF
--- a/packages/helm-charts/testnet/templates/validators.statefulset.yaml
+++ b/packages/helm-charts/testnet/templates/validators.statefulset.yaml
@@ -155,7 +155,7 @@ spec:
           [[ "$GETH_DEBUG" == "true" ]] && PROXY_ALLOW_PRIVATE_IP_FLAG="--proxy.allowprivateip"
 
           PROXIED_FLAGS=""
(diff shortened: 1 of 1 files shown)


packages/helm-charts/testnet/templates/validators.statefulset.yaml
```diff
@@ -155,7 +155,7 @@ spec:
           [[ "$GETH_DEBUG" == "true" ]] && PROXY_ALLOW_PRIVATE_IP_FLAG="--proxy.allowprivateip"
```
tkporter, line 155: We could remove this and all references to `GETH_DEBUG` now

tkporter (approved):

packages/helm-charts/testnet/templates/validators.statefulset.yaml
jfoutts-celo: The other option, which I believe @kevinjue tested yesterday, is to add `GETH_DEBUG=true` to .env.  I'm fine with getting rid of it, though.

jcortejoso: True! Thanks. Changed in https://github.com/celo-org/celo-monorepo/pull/2968

Issue #2744, opened 2020-02-11
Local Currency in Gold Flow


jmrossy:
### Expected Behavior

More wholistic local currency management and extend coverage to gold flow

Related:
- #1756 
- #2594 
- #1484
- #1261
- #2329

### Current Behavior

Local currency is in some parts of the app but not gold flow

PR #2965, opened 2020-02-26
[Wallet] Local currency in the gold flow

### LINKED ISSUE #1756: [Wallet] Users SBAT see local currency in gold exchange

### LINKED ISSUE #2594: Switch currency option in the Gold Buy/Sell flow does not use local currency
### Expected Behavior

should read “Switch to CAD”

### Current Behavior

In the “Buy Gold” flow using CAD, the ‘switch currency’ option reads “switch to USD

jeanregisser:
### Description

This PR enables buy/sell gold using local currency.

### Tested

**with USD local currency**

Buy (entering gold):

Buy (entering USD):

Sell (entering gold):

Sell (entering USD):

**with EUR local currency**

Buy (entering gold):

Buy (entering EUR):

Sell (entering gold):

Sell (entering EUR):

### Other changes

- `CurrencyDisplay` now supports both Dollar and Gold conversions.
- Updated `CeloGoldOverview` and `CeloDollarsOverview` to use `CurrencyDisplay`.
- `getLocalCurrencyCode` selector now returns `USD` instead of `null` when `USD` is selected. This is to make its behavior less surprising and overall simplifies code.

### Related issues

- Epic #2744
- Fixes #1756 
- Fixes #2594 
- #1484

### Backwards compatibility

Yes.


packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -20,13 +26,14 @@ interface Props {
   size: number // only used for DisplayType.Big
   useColors: boolean
   hideSymbol: boolean
-  formatAmount: (amount: BigNumber.Value) => string
+  showLocalAmount?: boolean
+  formatAmount: (amount: BigNumber.Value, currency?: CURRENCY_ENUM) => string
```
jmrossy: I think it may be nicer have formatting logic defined here instead by the consumers of this component

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -172,63 +194,115 @@ export class ExchangeReview extends React.Component<Props, State> {
...
                   })}
                 </Text>
-                <Text style={styles.currencyAmountText}>
-                  {getMoneyDisplayValue(this.state.inputAmount, this.state.inputToken, true)}
-                </Text>
+                <CurrencyDisplay
+                  style={styles.currencyAmountText}
+                  amount={{
+                    value: this.state.inputAmount.toString(),
+                    currencyCode: CURRENCIES[this.state.inputToken].code,
```
jmrossy, line 201: Can this be simplified? E.g. CurrencyDisplay takes the currency and finds the code itself?

packages/mobile/src/exchange/ExchangeTradeScreen.tsx
```diff
@@ -171,40 +183,62 @@ export class ExchangeTradeScreen extends React.Component<Props, State> {
...
+  getInputTokenAmount = () => {
+    const parsedInputAmount = parseInputAmount(this.state.inputAmount)
+    if (this.state.inputToken === CURRENCY_ENUM.GOLD) {
+      return parsedInputAmount
+    }
+
+    const { localCurrencyCode, localCurrencyExchangeRate } = this.props
+
+    let dollarsAmount
+    if (localCurrencyCode && localCurrencyCode !== LocalCurrencyCode.USD) {
```
jmrossy: Rather than have multiple places in the code where we branch based on if local currency is not USD, could we just use the same code path but have the USD - USD exchange rate be 1?

jmrossy (commented):
Overall looks like a nice improvement! Some suggestions below

packages/mobile/src/components/CurrencyDisplay.tsx
jeanregisser: Agreed, `formatAmount` was more of an escape hatch for the case where we display as a fee.

packages/mobile/src/exchange/ExchangeReview.tsx
jeanregisser: Yes, this will be simplified in a future PR where we merge/refactor `CURRENCY_ENUM` and `LocalCurrencyCode`.

packages/mobile/src/exchange/ExchangeTradeScreen.tsx
jeanregisser: Yes this condition here is not really necessary, exchange rate is already set to 1 when USD currency is selected.

jeanregisser:
Thanks @jmrossy for your review, I've addressed your comments.

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -36,44 +49,98 @@ function getBigSymbolStyle(fontSize: number, color: string) {
...
       ? LocalCurrencySymbol[displayAmount.currencyCode as LocalCurrencyCode]
       : CURRENCIES[currency].symbol
-  const value = new BigNumber(displayAmount.value)
-  const sign = value.isNegative() ? '-' : ''
-  const formattedValue = formatAmount(value.absoluteValue())
+    : null
+  const value = displayAmount ? new BigNumber(displayAmount.value) : null
+  const sign = value?.isNegative() ? '-' : ''
+  const formatAmount = getFormatFunction(formatType)
+  const formattedValue =
```
cmcewen, line 136: I don't really have a suggestion, but just noting this a a little gnarly with so many ternaries.

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -172,63 +194,115 @@ export class ExchangeReview extends React.Component<Props, State> {
...
               <HorizontalLine />
               <View style={styles.subtotalRowContainer}>
                 <Text style={styles.exchangeBodyText}>
-                  {t('subtotalAmount', {
-                    rate: getMoneyDisplayValue(exchangeRate, CURRENCY_ENUM.DOLLAR, true),
-                  })}
-                </Text>
-                <Text style={styles.exchangeBodyText}>
-                  {getMoneyDisplayValue(dollarAmount, CURRENCY_ENUM.DOLLAR, true)}
+                  <Trans i18nKey="subtotalAmount" ns={Namespaces.exchangeFlow9}>
```
cmcewen, line 208: Curious on switching to the `Trans` components - when do you prefer this vs just using the `t` function?

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -172,63 +194,115 @@ export class ExchangeReview extends React.Component<Props, State> {
...
+                Buy or sell{' '}
+                <CurrencyDisplay
+                  amount={{
+                    value: this.getInputAmountInToken(CURRENCY_ENUM.GOLD).toString(),
+                    currencyCode: CURRENCIES[CURRENCY_ENUM.GOLD].code,
+                  }}
+                />
+                Gold
+              </Trans>
+            }
```
cmcewen, line 305: Getting to be a bit big as a component - wondering if we can make it smaller

packages/mobile/src/exchange/ExchangeTradeScreen.test.tsx
```diff
@@ -40,89 +35,154 @@ describe(ExchangeTradeScreen, () => {
...
+
+    fireEvent.changeText(getByTestId('ExchangeInput'), '500')
+    expect(getByTestId('ExchangeReviewButton').props.disabled).toBe(true)
+
+    fireEvent.changeText(getByTestId('ExchangeInput'), '0.001')
+    expect(getByTestId('ExchangeReviewButton').props.disabled).toBe(true)
+
+    // This is the minimum amount when exchanging dollars (see DOLLAR_TRANSACTION_MIN_AMOUNT)
+    fireEvent.changeText(getByTestId('ExchangeInput'), '0.01')
+    expect(getByTestId('ExchangeReviewButton').props.disabled).toBe(false)
```
cmcewen, line 186: 🙏 appreciate the test coverage

cmcewen (approved):
Overall looks good to me - mostly abstract feedback. I see you mentioned there is still some work to do on this, so if it's not captured in an issue somewhere, would be great to create one

cmcewen:
Also thanks for the extensive testing + screenshots 🙏

jmrossy (approved):

packages/mobile/src/exchange/ExchangeReview.tsx
jeanregisser: I tried keeping just the `t` function and passing `<CurrencyDisplay ... />` as a variable, but it always ended up being escaped to `[object Object]` not matter what I tried (`useRawValueToEscape: true`, `escapeValue: false` and their permutations).

`<Trans ... />` is the only way I found to make it work with the component approach.

One alternative would be to introduce a function returning a simple string to continue using the `t` function.
Maybe will see the need for this later on.

packages/mobile/src/components/CurrencyDisplay.tsx
jeanregisser: Yes this will probably evolve and get refactored to a nicer way when we proceed with the merge/refactor of `CURRENCY_ENUM` and `LocalCurrencyCode`.

jeanregisser:
> Overall looks good to me - mostly abstract feedback. I see you mentioned there is still some work to do on this, so if it's not captured in an issue somewhere, would be great to create one

Thanks! I've opened https://github.com/celo-org/celo-monorepo/issues/3025

packages/mobile/src/exchange/ExchangeReview.tsx
jeanregisser: Note: we would need to pass some of the redux state for currency conversion for the function approach. Which could add some boilerplate at the call site compared to the component approach.

PR #2966, opened 2020-02-26
Update ethers.js library to web3.js

critesjosh:
### Description

This is an update to [this documentation page](https://docs.celo.org/v/master/developer-guide/start/hellocelo).

I updated the library that creates Celo accounts in the base [repo](https://github.com/critesjosh/helloCelo) from ethers.js to web3.js. This PR is to update the docs to remain consistent with the base repo.


nityas (approved):
Awesome!

PR #2968, opened 2020-02-26
Fix contract migrations due to non-authorized oracle address

jcortejoso:
### Description

Explicitly add deployment address as an oracle authorized address.

### Other changes

Removed variable `GETH_DEBUG`.


tkporter (approved):
😍

PR #2972, opened 2020-02-26
Small adjustments to migrations config

asaj:
### Description

Update the migrations config with latest parameters.
### Tested

Not tested.
### Other changes

None
### Related issues

None
### Backwards compatibility

Yes


packages/protocol/migrationsConfig.js
```diff
@@ -69,8 +70,7 @@ const DefaultConfig = {
     spread: 5 / 1000,
     reserveFraction: 1 / 100,
     updateFrequency: 5 * MINUTE, // 5 minutes
-    // TODO(asa): This is too low
-    minimumReports: 1,
+    minimumReports: 10,
```
rcroessmann: Still discussing this in the oracles bubble but likely needs to be smaller than 10
```suggestion
    minimumReports: 5,
```

packages/protocol/migrationsConfig.js
```diff
@@ -69,8 +70,7 @@ const DefaultConfig = {
     spread: 5 / 1000,
     reserveFraction: 1 / 100,
```
rcroessmann: Must be increased due to frozen reserve cGLD
```suggestion
    reserveFraction: 1 / 20,
```

packages/protocol/migrationsConfig.js
```diff
@@ -51,7 +52,7 @@ const DefaultConfig = {
     targetVotingYieldParameters: {
       initial: 0.00016, // (x + 1) ^ 365 = 1.06
       max: 0.0005, // (x + 1) ^ 365 = 1.20
-      adjustmentFactor: 1 / 365,
+      adjustmentFactor: 1 / 3650,
```
rcroessmann: This feels perfect for the beginning (max. yield decrease per year is now 3.33% and max. yield increase per year is 6.66%). Could be decrease by governance if voting percentage roughly settles at around 66.67%.

packages/protocol/migrationsConfig.js
```diff
@@ -116,16 +116,16 @@ const DefaultConfig = {
...
-    frozenDays: 0,
+    frozenGold: 105000000000000000000000000, // 105M
+    frozenDays: 3 * YEAR,
     spenders: [],
+    // To be populated once addresses are generated.
     otherAddresses: [],
-    assetAllocationSymbols: ['cGLD', 'BTC', 'ETH'], // TODO(roman)
-    assetAllocationWeights: [0.5, 0.25, 0.25], // TODO(roman)
+    assetAllocationSymbols: ['cGLD', 'BTC', 'ETH'],
+    assetAllocationWeights: [0.5, 0.3, 0.2],
```
rcroessmann:
```suggestion
    assetAllocationSymbols: ['cGLD', 'BTC', 'ETH', 'DAI'],
    assetAllocationWeights: [0.5, 0.2, 0.1, 0.2],
```

packages/protocol/migrationsConfig.js
```diff
@@ -82,20 +82,20 @@ const DefaultConfig = {
...
-    concurrentProposals: 10,
-    approvalStageDuration: 3 * DAY,
-    referendumStageDuration: WEEK,
-    executionStageDuration: WEEK,
+    queueExpiry: WEEK, // Change to 4 weeks once mainnet activated
+    dequeueFrequency: DAY, // Change to 1 week once mainnet activated
+    concurrentProposals: 3, // Change to 10 once mainnet activated
+    approvalStageDuration: DAY, // Change to 3 days once mainnet activated
+    referendumStageDuration: DAY, // Change to 1 week once mainnet activated
+    executionStageDuration: DAY, // Change to 1 week once mainnet activated
```
nategraf: Once the execution stage ends, the proposal will expire, even though it was already approved. I think it'd be better to keep this as a week because we (or anyone) could execute the proposal as soon as the it's ready, or if something happens that we don't expect, the proposal would stay valid for up to 7 days.
```suggestion
    executionStageDuration: WEEK,
```

nategraf (commented):
My only comment. Otherwise looks good to me. I'll defer to Roman for approval.

packages/protocol/migrationsConfig.js
aslawson: +1 -- I will update

rcroessmann (approved):

packages/protocol/migrationsConfig.js
```diff
@@ -95,7 +95,7 @@ const DefaultConfig = {
     participationBaselineQuorumFactor: 1,
   },
   governanceApproverMultiSig: {
-    //Placeholder until addresses for 2/5 multsig are generated.
+    // Placeholder until addresses for 2/5 multsig are generated.
```
asaj: We have three of these, which I think is enough for now. Can we include them here, and make both required confirmations 2?

packages/protocol/migrationsConfig.js
```diff
@@ -116,16 +116,16 @@ const DefaultConfig = {
   reserve: {
     tobinTaxStalenessThreshold: HOUR, // 1 hour
     dailySpendingRatio: toFixed(0.05).toFixed(), // 5%
-    // TODO(Roman): Set these appropriately.
-    frozenGold: 0,
-    frozenDays: 0,
+    frozenGold: 105000000000000000000000000, // 105M
+    frozenDays: 3 * YEAR,
```
asaj: As @aslawson pointed out, the value provided is in seconds rather than days

packages/protocol/migrationsConfig.js
```diff
@@ -116,16 +116,16 @@ const DefaultConfig = {
...
     spenders: [],
+    // To be populated once addresses are generated.
```
asaj: We have one of these generated

packages/protocol/migrationsConfig.js
```diff
@@ -116,16 +116,16 @@ const DefaultConfig = {
...
     otherAddresses: [],
-    assetAllocationSymbols: ['cGLD', 'BTC', 'ETH'], // TODO(roman)
-    assetAllocationWeights: [0.5, 0.25, 0.25], // TODO(roman)
+    assetAllocationSymbols: ['cGLD', 'BTC', 'ETH', 'DAI'],
+    assetAllocationWeights: [0.5, 0.2, 0.1, 0.2],
   },
   reserveSpenderMultiSig: {
-    //Placeholder until addresses for 2/2 multsig are generated.
+    // Placeholder until addresses for 2/2 multsig are generated.
```
asaj: These are ready 😄

packages/protocol/migrationsConfig.js
```diff
@@ -145,7 +145,8 @@ const DefaultConfig = {
     frozen: true,
   },
   transferWhitelist: {
-    addresses: [], // TODO(Alec): get whitelist of addresses.
+    // To be populated with addresses once generated.
+    addresses: [],
```
asaj: I think we're ready to populate this as well

packages/protocol/migrationsConfig.js
```diff
@@ -27,8 +28,8 @@ const DefaultConfig = {
     gasForNonGoldCurrencies: 50000,
     minimumClientVersion: {
       major: 1,
-      minor: 8,
-      patch: 23,
+      minor: 9,
+      patch: 8,
```
asaj, line 32: Can we bump the double signing slasher penalty to 9k gold?

packages/protocol/migrations/08_reserve.ts
```diff
@@ -29,8 +30,8 @@ const initializeArgs = async (): Promise<[
     registry.address,
     config.reserve.tobinTaxStalenessThreshold,
     config.reserve.dailySpendingRatio,
-    config.reserve.frozenGold,
-    config.reserve.frozenDays,
+    0, // frozenGold must be set after initial balance
```
asaj: nit: "frozenGold cannot be set until the reserve us funded", ditto below

packages/protocol/migrations/08_reserve.ts
```diff
@@ -59,7 +60,9 @@ module.exports = deploymentForCoreContract<ReserveInstance>(
       await web3.eth.sendTransaction({
         from: network.from,
         to: reserve.address,
-        value: web3.utils.toWei(config.reserve.initialBalance.toString(), 'ether').toString(),
+        value: new BigNumber(
+          web3.utils.toWei(new BigNumber(config.reserve.initialBalance).toString(10), 'ether')
```
asaj: Why change this?

packages/protocol/migrationsConfig.js
```diff
@@ -97,8 +97,12 @@ const DefaultConfig = {
     participationBaselineQuorumFactor: 1,
   },
   governanceApproverMultiSig: {
-    //Placeholder until addresses for 2/5 multsig are generated.
-    signatories: [`0x5409ed021d9299bf6814279a6a1411a7e866a631`],
+    // 1/3 placeholder until addresses for 2/5 multsig are generated.
```
asaj: Can we make this 2/3 for both regular and rquired. And feel free to remove this comment

packages/protocol/migrationsConfig.js
```diff
@@ -118,18 +122,19 @@ const DefaultConfig = {
...
     tobinTaxStalenessThreshold: HOUR, // 1 hour
     dailySpendingRatio: toFixed(0.05).toFixed(), // 5%
-    // TODO(Roman): Set these appropriately.
-    frozenGold: 0,
-    frozenDays: 0,
     spenders: [],
-    otherAddresses: [],
-    assetAllocationSymbols: ['cGLD', 'BTC', 'ETH'], // TODO(roman)
-    assetAllocationWeights: [0.5, 0.25, 0.25], // TODO(roman)
+    // To be populated once addresses are generated.
```
asaj: Remove comment

packages/protocol/migrationsConfig.js
```diff
@@ -118,18 +122,19 @@ const DefaultConfig = {
...
   reserveSpenderMultiSig: {
-    //Placeholder until addresses for 2/2 multsig are generated.
-    signatories: [`0x5409ed021d9299bf6814279a6a1411a7e866a631`],
-    numRequiredConfirmations: 1,
+    // 2/2 multsig
+    signatories: [
+      '0x49eFFA2ceF5FccA5540f421d6b28e76184cc0fDF',
+      '0x4550F1576fAC966Ac8b9F42e1D5D66D3A14dD8D3',
+    ],
+    numRequiredConfirmations: 2,
```
asaj, line 137: Can we make both required confirmations be 2 for this?

packages/protocol/migrationsConfig.js
asaj: Also FYI one more address will be ready in an hour, let's try to get that in this PR

packages/protocol/migrations/08_reserve.ts
aslawson: it doesn't work for large values because toString puts it in scientific notation.  since it appears this is only used for testing its probably unnecessary

packages/protocol/migrationsConfig.js
aslawson: sure

packages/protocol/migrationsConfig.js
aslawson: i'm guessing you mean for regular and internal?

so it will then be 2/4 when we have the new address

asaj: Yeah 2/4 for both regular and internal for now

packages/protocol/migrations/08_reserve.ts
asaj: If it's a bug fix definitely keep it, thanks for the explanation

packages/protocol/migrationsConfig.js
```diff
@@ -84,11 +84,11 @@ const DefaultConfig = {
     frozen: true,
   },
   governance: {
-    queueExpiry: 4 * WEEK,
-    dequeueFrequency: WEEK,
-    concurrentProposals: 10,
-    approvalStageDuration: 3 * DAY,
-    referendumStageDuration: WEEK,
+    queueExpiry: WEEK, // Change to 4 weeks once mainnet activated
+    dequeueFrequency: DAY, // Change to 1 week once mainnet activated
```
asaj: On second thought, given the number of governance proposals that will need to be passed, suggest speeding these up even further for RC0:

```
dequeueFrequency: MINUTE,
approvalStageDuration: 30 * minute,
referendumStageDuration: HOUR,
```

packages/protocol/migrationsConfig.js
asaj: We never really came to a conclusion on what we wanted to do for transitioning from the genesis to elected validator set in RC0. My suggestion would be to lower the `minElectableValidators` for now to a very low number, e.g. 5, with a comment that it should be raised to 50 for mainnet

asaj: In order to start with voter rewards turned off, we'll need to set `initial`, and `adjustmentFactor` here to both be zero. Please preserve the desired values here in a comment that they should be set to those values in order to enable voter rewards.

asaj (commented):
Just a few more suggestions

asaj:
LGTM!

PR #2973, opened 2020-02-26
registers whitelist with migration

alecps:
### Description

Registers transferWhitelist contract with registry so its address can be read from layer 1.

### Tested

Not tested.

### Other changes

None

### Related issues

- Related to #[251](https://github.com/celo-org/celo-labs/issues/251)

### Backwards compatibility

Yes.

### DIFF
--- a/packages/protocol/migrations/03_transferwhitelist.ts
+++ b/packages/protocol/migrations/03_transferwhitelist.ts
@@ -1,6 +1,6 @@
+import { CeloContractName } from '@celo/protocol/lib/registry-utils'
 import { getDeployedProxiedContract } from '@celo/protocol/lib/web3-utils'
 import { config } from '@celo/protocol/migrationsConfig'
-import { CeloContractName } from 'lib/registry-utils'
 import { RegistryInstance, TransferWhitelistInstance } from 'types'
 
 const name = CeloContractName.TransferWhitelist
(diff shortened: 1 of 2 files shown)


packages/protocol/migrations/03_transferwhitelist.ts
```diff
@@ -10,5 +11,7 @@ module.exports = (deployer: any) => {
     const contract: TransferWhitelistInstance = await Contract.deployed()
     await contract.setWhitelist(config.transferWhitelist.addresses)
     await contract.setRegisteredContracts(config.transferWhitelist.registryIds)
+    const registry = await getDeployedProxiedContract<RegistryInstance>('Registry', artifacts)
+    await registry.setAddressFor(name, contract.address)
```
asaj, line 16: Can you add `name` to the registry ID enum?

packages/protocol/migrations/03_transferwhitelist.ts
```diff
@@ -10,5 +12,7 @@ module.exports = (deployer: any) => {
     const contract: TransferWhitelistInstance = await Contract.deployed()
     await contract.setWhitelist(config.transferWhitelist.addresses)
     await contract.setRegisteredContracts(config.transferWhitelist.registryIds)
+    const registry = await getDeployedProxiedContract<RegistryInstance>('Registry', artifacts)
```
asaj, line 15: You can use `CeloContractName` here too

asaj (approved):

PR #2977, opened 2020-02-26
Update bls12377js to fix hash to G1

kobigurk:
Hash to G1 didn't take into account the "greatest" flag correctly.


jmrossy (approved):
LGTM

PR #2978, opened 2020-02-26
[Wallet] Implement retry logic for sent txs

### LINKED ISSUE #2916: Add retry logic to all sent txs

jmrossy:
### Description

- Add retry, timeout, and optional cancel capability for sent transactions
- Reorganize some of the code in web3/saga.ts
- Fix a missing i18n string

### Tested

Sending multiple txs on forno and geth mode.
Dropped tx timeout down to 2s to force timeout.

### Related issues

- Fixes #2916

### Backwards compatibility

Yes


packages/mobile/src/web3/privateKey.ts
```diff
@@ -0,0 +1,73 @@
+import * as Crypto from 'crypto'
```
jmrossy, line 1: Moved here from web3/saga.ts which was getting too big

packages/mobile/src/transactions/send.ts
```diff
@@ -116,11 +122,103 @@ export function* sendTransaction(
...
+      nonce,
+      staticGas
+    )
+    const result = yield confirmation
+    return result
+  }
+  yield call(wrapSendTransactionWithRetry, txId, account, sendTxMethod, cancelAction)
+}
+
+export function* wrapSendTransactionWithRetry(
```
jmrossy, line 144: Working around the 3-promise based interface we're using in `sendTransactionPromises` turned out to be a pain. It has me rethinking if the promises are actually a nicer pattern in general then callbacks. They're great if consumers will be interacting with the results in a 'straight line' but not so nice for adding new middleware btwn the caller and sending utilities

i1skn (approved):
Great work! 🚢 it!

jeanregisser (approved):
Great work! 👍

PR #2981, opened 2020-02-26
ReleaseGold CLI

### LINKED ISSUE #2820: Releasing Gold contractkit + CLI
### Expected Behavior

CLI+ContractKit support for releasing gold

### LINKED ISSUE #2971: Upgrade ReleaseGold migrations script
### Expected Behavior
Various improvements suggested by @nambrot  should be implemented on `deploy_release_contracts.ts`:
* Validation on `releaseGoldContracts.json`-type files
* Confirmation on each contract creation
* Option to write output to a file

### Current Behavior

None of these behaviors are implemented.

lucasege:
### Description

Adding commands for all ReleaseGold fns. 

Also added some improvements to the deployment script to make the grant process safer and more understandable (and slower)

### Tested

Tested each fn with a local testnet and added some cli tests.

## Other changes

Typos, changed the jest timeout since I saw many tests failing from that - probably related to #3091

### Related issues

- Fixes #2820
- Fixes #2971 

### Backwards compatibility

Yes


packages/cli/src/commands/releasegold/authorize.ts
```diff
@@ -0,0 +1,87 @@
...
+    // tslint:disable-next-line
+    const { flags } = this.parse(Authorize)
+    const contractAddress = flags.contract
+    const role = flags.role
+    const releaseGoldWrapper = new ReleaseGoldWrapper(
+      this.kit,
+      newReleaseGold(this.kit.web3, contractAddress)
+    )
+
+    await newCheckBuilder(this)
```
mcortesi, line 45: maybe a good check would be to add a check if that address is actually a contract (getCodeSize > 0)

packages/cli/src/commands/releasegold/locked-gold.ts
```diff
@@ -0,0 +1,102 @@
...
+      }
+    } else if (flags.action === 'unlock') {
+      await checkBuilder
+        .isNotVoting(contractAddress)
+        .hasEnoughLockedGoldToUnlock(value)
+        .runChecks()
+      this.kit.defaultAccount = isRevoked ? releaseOwner : beneficiary
+      await displaySendTx('lockedGoldUnlock', releaseGoldWrapper.unlockGold(flags.value))
+    } else if (flags.action === 'withdraw') {
+      await checkBuilder.runChecks()
```
mcortesi, line 65: no checks?

packages/cli/src/commands/releasegold/set-account.ts
```diff
@@ -0,0 +1,66 @@
...
+    await newCheckBuilder(this)
+      .isAccount(releaseGoldWrapper.address)
+      .addCheck('Contract is not revoked', () => !isRevoked)
+      .runChecks()
+
+    let tx: any
+    if (flags.property === 'name') {
+      tx = releaseGoldWrapper.setAccountName(flags.value)
+    } else if (flags.property === 'dataEncryptionKey') {
+      tx = releaseGoldWrapper.setAccountDataEncryptionKey(flags.value as any)
```
mcortesi: why the any?

packages/cli/src/commands/releasegold/show.ts
```diff
@@ -0,0 +1,82 @@
...
+import {
+  ReleaseGoldWrapper,
+  ReleaseSchedule,
+  RevocationInfo,
+} from '@celo/contractkit/lib/wrappers/ReleaseGold'
+import { BaseCommand } from '../../base'
+import { printValueMapRecursive } from '../../utils/cli'
+import { Flags } from '../../utils/command'
+
+export interface BalanceState {
```
mcortesi: Probably a good fit to move to the wrapper inside contractKit

packages/cli/src/commands/releasegold/show.ts
```diff
@@ -0,0 +1,82 @@
...
+  totalWithdrawn: string
+  maxDistribution: string
+  totalBalance: string
+  remainingTotalBalance: string
+  remainingUnlockedBalance: string
+  remainingLockedBalance: string
+  currentReleasedTotalAmount: string
+}
+
+export interface ReleaseGoldInfo {
```
mcortesi: same

mcortesi (approved):
Lots of boilerplate.. hard to review. Seems correct, only mentioned a few things to improve.

Also, maybe a way to reduce boilerplate is to have some common CLI base command for all releaseGold calls, and thus reduce boilerplate

packages/cli/src/commands/releasegold/authorize.ts
lucasege: `Flags.address` appears to already do this, but I added an additional check in the common base command that the ABI of the contract at the address does match release gold abi

packages/cli/src/commands/releasegold/set-account.ts
lucasege: Good catch, this was required when I was using `value` for a different type earlier, but now it isn't necessary.

packages/cli/src/commands/releasegold/locked-gold.ts
lucasege: Yes that just runs the previously defined test `isAccount` since `withdraw` does not need any additional checks. Agree it looks a bit untidy

packages/cli/src/commands/account/authorize.ts
```diff
@@ -69,7 +69,6 @@ export default class Authorize extends BaseCommand {
       tx = await accounts.authorizeAttestationSigner(res.flags.signer, sig)
     } else {
       this.error(`Invalid role provided`)
-      return
```
asaj: Why remove this?

packages/cli/src/commands/releasegold/authorize.ts
```diff
@@ -0,0 +1,79 @@
...
+        sig,
+        flags.blsKey,
+        flags.blsPop!
+      )
+    } else if (role === 'validator') {
+      tx = await this.releaseGoldWrapper.authorizeValidatorSigner(flags.signer, sig)
+    } else if (role === 'attestation') {
+      tx = await this.releaseGoldWrapper.authorizeAttestationSigner(flags.signer, sig)
+    } else {
+      this.error('Invalid role provided')
```
asaj, line 75: return here?

packages/cli/src/commands/releasegold/locked-gold.test.ts
```diff
@@ -0,0 +1,33 @@
...
+      contractCanValdiate
+    )
+    const kit = newKitFromWeb3(web3)
+    const lockedGold = await kit.contracts.getLockedGold()
+    await CreateAccount.run(['--contract', contractAddress])
+    await LockedGold.run(['--contract', contractAddress, '--action', 'lock', '--value', '100'])
+    await LockedGold.run(['--contract', contractAddress, '--action', 'unlock', '--value', '50'])
+    await LockedGold.run(['--contract', contractAddress, '--action', 'lock', '--value', '75'])
+    await LockedGold.run(['--contract', contractAddress, '--action', 'unlock', '--value', '50'])
+    // Expected to fail "silently" because `locked-gold` uses ts time which we can't spoof
```
asaj: Why does this fail silently as opposed to loudly? why run it at all if it's going to fail?

packages/cli/src/commands/releasegold/locked-gold.ts
```diff
@@ -0,0 +1,95 @@
+import { flags } from '@oclif/command'
+import BigNumber from 'bignumber.js'
+import { newCheckBuilder } from '../../utils/checks'
+import { displaySendTx } from '../../utils/cli'
+import { Flags } from '../../utils/command'
+import { ReleaseGoldCommand } from './release-gold'
+
+export default class LockedGold extends ReleaseGoldCommand {
+  static description =
+    'Perform actions [lock, unlock, relock, withdraw] on the gold held in the given contract.'
```
asaj: Is this just about locked gold? Perhaps `Perform actions [lock, unlock, relock, withdraw] on Celo Gold that has been locked via the provided ReleaseGold contract.`

packages/cli/src/commands/releasegold/locked-gold.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+    } else if (flags.action === 'unlock') {
+      await checkBuilder
+        .isNotVoting(this.contractAddress)
+        .hasEnoughLockedGoldToUnlock(value)
+        .runChecks()
+      this.kit.defaultAccount = isRevoked ? releaseOwner : beneficiary
+      await displaySendTx('lockedGoldUnlock', this.releaseGoldWrapper.unlockGold(flags.value))
+    } else if (flags.action === 'withdraw') {
+      await checkBuilder.runChecks()
```
asaj, line 65: Why aren't these checks run for all subcommands?

packages/cli/src/commands/releasegold/locked-gold.ts
```diff
@@ -0,0 +1,95 @@
...
+      const currentTime = Math.round(new Date().getTime() / 1000)
+      while (true) {
```
asaj, line 67: Is this the same as the locked gold specific command? I recall a "wait" flag or something

packages/cli/src/commands/releasegold/refund-and-finalize.ts
```diff
@@ -0,0 +1,29 @@
+import { newCheckBuilder } from '../../utils/checks'
+import { displaySendTx } from '../../utils/cli'
+import { ReleaseGoldCommand } from './release-gold'
+
+export default class RefundAndFinalize extends ReleaseGoldCommand {
+  static description =
+    "Refund the given contract's balance to the appopriate parties and destroy the contact"
```
asaj: Consider including `Can only be called by the release owner of revokable ReleaseGold instances.`, or something to make it clear to users that they are unlikely to need this command

packages/cli/src/commands/releasegold/revoke-votes.ts
```diff
@@ -0,0 +1,76 @@
...
+    }),
+    group: Flags.address({
+      required: true,
+      description: 'Address of the group to revoke votes from',
+    }),
+    votes: flags.string({ required: true, description: 'The number of votes to revoke' }),
+  }
+
+  static examples = [
+    'revoke-votes --contract 0x47e172F6CfB6c7D01C1574fa3E2Be7CC73269D95 --type pending --group 0x5409ED021D9299bf6814279A6A1411A7e866A631 --votes 100',
```
asaj: Is this what we do in the normal revoke? I think it's more user friendly for them to just provide a value, and then automatically revoke pending votes first, then active votes.

packages/cli/src/commands/releasegold/revoke.ts
```diff
@@ -0,0 +1,24 @@
+import { newCheckBuilder } from '../../utils/checks'
+import { displaySendTx } from '../../utils/cli'
+import { ReleaseGoldCommand } from './release-gold'
+
+export default class Revoke extends ReleaseGoldCommand {
+  static description = 'Revoke the given contract instance'
```
asaj: Please elaborate, e.g. `Once revoked, any Locked Gold can be unlocked by the release owner. The beneficiary will then be able to withdraw any released Gold that had yet to be withdrawn, and the remainder can bet transferred by the release owner to the refund address. Note that not all ReleaseGold instances are revokable.`

packages/cli/src/commands/releasegold/revoke.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  static args = []
+
+  static examples = ['revoke --contract 0x5409ED021D9299bf6814279A6A1411A7e866A631']
+
+  async run() {
```
asaj, line 20: Consider adding a confirmation prompt to be extra sure that this isn't run accidentally

packages/cli/src/commands/releasegold/set-beneficiary.ts
```diff
@@ -0,0 +1,41 @@
...
+    const newBeneficiary = flags.beneficiary
+
+    await newCheckBuilder(this)
+      .addCheck(
+        'Owner argument is contract owner',
+        async () => owner === (await this.releaseGoldWrapper.getOwner())
+      )
+      .runChecks()
+
+    this.kit.defaultAccount = owner
```
asaj: This command won't work if the contract owner is a multisig.

packages/cli/src/commands/releasegold/set-beneficiary.ts
```diff
@@ -0,0 +1,41 @@
...
+      required: true,
+      description: 'Owner of `contract` capable of setting the new beneficiary (multisig)',
+    }),
+    beneficiary: Flags.address({ required: true, description: 'Address of the new beneficiary' }),
+  }
+
+  static args = []
+
+  static examples = [
+    'set-beneficiary --contract 0x5409ED021D9299bf6814279A6A1411A7e866A631 --owner 0x6Ecbe1DB9EF729CBe972C83Fb886247691Fb6beb --beneficiary 0x6Ecbe1DB9EF729CBe972C83Fb886247691Fb6beb',
```
asaj: Why do the owner and beneficiary need to be provided?

packages/cli/src/commands/releasegold/set-liquidity-provision.ts
```diff
@@ -0,0 +1,26 @@
+import { newCheckBuilder } from '../../utils/checks'
+import { displaySendTx } from '../../utils/cli'
+import { ReleaseGoldCommand } from './release-gold'
+
+export default class SetLiquidityProvision extends ReleaseGoldCommand {
```
asaj, line 7: ditto, consider adding a confirmation prompt

packages/cli/src/commands/releasegold/transfer.ts
```diff
@@ -0,0 +1,36 @@
+import { displaySendTx } from '../../utils/cli'
+import { Flags } from '../../utils/command'
+import { ReleaseGoldCommand } from './release-gold'
+
+export default class Transfer extends ReleaseGoldCommand {
```
asaj: Consider calling this command `transfer-dollars` or something like that. Rename the file as well to reflect this

packages/cli/src/commands/releasegold/transfer.ts
```diff
@@ -0,0 +1,36 @@
...
+  static description = 'Transfer stable tokens from the given contract address'
```
asaj: s/stable tokens/Celo Dollars

asaj: `Dollars may be accrued to the ReleaseGold contract via validator epoch rewards.`

packages/cli/src/commands/releasegold/transfer.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  static flags = {
+    ...ReleaseGoldCommand.flags,
+    to: Flags.address({
+      required: true,
+      description: 'Address of the recipient of stable token transfer',
+    }),
+    value: Flags.wei({ required: true, description: 'Value (in Wei) of stable token to transfer' }),
```
asaj: Celo Dollars

asaj (commented):
Nice work 👏

packages/cli/src/commands/releasegold/locked-gold.ts
lucasege: It is the same - must have changed since you saw it last

packages/cli/src/commands/releasegold/revoke-votes.ts
lucasege: You're right, I must have been asleep here, updated it

packages/cli/src/commands/releasegold/locked-gold.ts
lucasege: Which subcommands are you referring to here? Each command "lock", "unlock", and "withdraw" all run `checkBuilder.runChecks()` but `lock` and `unlock` just happen to add more checks

lucasege:
Still some flaky CLI tests - within `cli/releasegold` `locked-gold.test.ts` and `withdraw.test.ts`. They both pass when run individually but not when running the whole suite. Some CLI expertise would be greatly appreciated. 
They both fail in the test suite on the same test `checkBuilder.isAccount` which indicates to me that somehow the command `createAccount` on releaseGold is either getting overwritten or the suite isn't waiting for it to finish.
I suspect an issue either in my understanding/code or in `testWithGanache` or a deeper layer.

RESOLVED: needed to change some `beforeAll`s to `beforeEach` because of the way `testWithGanache` works, despite only having 1 test.

packages/cli/package.json
```diff
@@ -26,7 +26,7 @@
     "docs": "yarn oclif-dev readme --multi --dir=../docs/command-line-interface && yarn prettier ../docs/command-line-interface/*.md --write",
     "lint": "tslint -c tslint.json --project tsconfig.json",
     "prepack": "yarn run build && oclif-dev manifest && oclif-dev readme",
-    "test:reset": "yarn --cwd ../protocol devchain generate-tar .tmp/devchain.tar.gz --migration_override ../dev-utils/src/migration-override.json --upto 24 --release_gold_contracts scripts/truffle/releaseGoldContracts.json",
+    "test:reset": "yarn --cwd ../protocol devchain generate-tar .tmp/devchain.tar.gz --migration_override ../dev-utils/src/migration-override.json --upto 24 --release_gold_contracts scripts/truffle/releaseGoldDefaultContracts.json",
```
asaj, line 29: Maybe `releaseGoldExampleConfigs.json`?

packages/cli/src/commands/releasegold/locked-gold.ts
asaj: My bad must not have read carefully

asaj (approved):
Given tests are passing feel free to ignore the suggestion below

lucasege:
I'll include the name change in my next PR on RG

PR #2987, opened 2020-02-27
Fix PIN code on enabling Forno mode

i1skn:
### Description

PIN screen would tell, that PIN is incorrect even if PIN is correct, when enabling data saver. This is a fix. 

### Tested

iOS Simulator

### Other changes

I've change the background color of the view, which quickly show up before PIN Screen appears after app is back from background on iOS. Earlier when app is in background - we would show green screen, but after https://github.com/celo-org/celo-monorepo/commit/4c5633314cf8fbfcadb94db955e26e2b413b5d5c we just blur the content, so white screen would make more sense.

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/app/Background.tsx
+++ b/packages/mobile/src/app/Background.tsx
@@ -60,7 +60,7 @@ const styles = StyleSheet.create({
   container: {
     flex: 1,
     width: '100%',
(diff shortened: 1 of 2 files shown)


annakaz (approved):
LGTM!

PR #2988, opened 2020-02-27
Slashing fixes

mrsmkl:
### Description

Fix contractkit bugs relating to slashing.

### Tested

Adds e2e tests that exercise the modified code.


mcortesi:
This seems to have lots of commits that i suspect are not part of the PR? @mrsmkl

packages/utils/src/collections.ts
```diff
@@ -91,9 +91,17 @@ export function linkedListChanges(
...
   for (const it of changeList) {
-    const { lesser, greater } = _linkedListChange(sortedList, it)
+    const { lesser, greater } = _linkedListChange(list, it)
     lessers.push(lesser)
     greaters.push(greater)
   }
   return { lessers, greaters, list }
 }
+
+export function range(from: number, to: number, step: number) {
```
mcortesi: ❤️

packages/utils/src/collections.ts
```diff
@@ -91,9 +91,17 @@ export function linkedListChanges(
   const lessers: string[] = []
   const greaters: string[] = []
   for (const it of changeList) {
-    const { lesser, greater } = _linkedListChange(sortedList, it)
+    const { lesser, greater } = _linkedListChange(list, it)
```
mcortesi: consider doing `[].contact(sortedList)` since imo it's more idiomatic. and maybe renaming variable to `listClone` or something

mcortesi (approved):

packages/utils/src/collections.ts
jfoutts-celo: We already have this range function, just `import { range } from 'lodash'`

PR #2996, opened 2020-02-28
[phone-number-privacy] add BlindBLS library request and add test

aslawson:
### Description

### Tested

Tested locally and against remote firebase endpoint
eg:
 `curl -v "http://localhost:5000/celo-phone-number-privacy/us-central1/getSalt" -d '{"blindPhoneNumber": "5555555555"}' -H 'Content-Type: application/json'`

added unit test

### Other changes

NA

### Related issues

- Fixes #[328]

### Backwards compatibility

NA


packages/phone-number-privacy/src/config.ts
```diff
@@ -1,4 +1,10 @@
...
 const functions = require('firebase-functions')
-export const config = functions.config()
+export const functionConfig = functions.config()
 
-console.log(config)
+// Load environment variables from .env file
+dotenv.config()
+
+export const PHONE_NUMBER_PRIVACY_SECRET_KEY =
+  (process.env.SECRET_KEY as string) || functionConfig.envs.secret_key //TODO @aslawson what is best way to include private key to GCP?
```
jmrossy: We can chat about this in our next meeting. For now, hold off on putting the real key into firebase config

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,6 +1,9 @@
+import { BLINDBLS } from 'bls12377js-blind'
 import * as functions from 'firebase-functions'
+import { PHONE_NUMBER_PRIVACY_SECRET_KEY } from './config'
 
 export const getSalt = functions.https.onRequest((request, response) => {
-  console.log(request.body)
-  response.send(request.body)
+  const privateKey = new Buffer(PHONE_NUMBER_PRIVACY_SECRET_KEY)
```
jmrossy, line 6: Nit, have any needing processing of the key happen in config.ts so other files can just consume the key

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,6 +1,9 @@
...
+  const salt = BLINDBLS.computePRF(privateKey, new Buffer(request.body.blindPhoneNumber))
+  response.json({ success: true, salt })
```
jmrossy: Let's wrap this in a try catch with some more logging. Np if you'd rather do that in a follow-up PR

scripts/key_placer.sh
```diff
@@ -15,6 +15,7 @@ files=(
   "packages/mobile/android/sentry.properties"
   "packages/mobile/ios/GoogleService-Info.plist"
   "packages/mobile/ios/sentry.properties"
+  "packages/phone-number-privacy/.env"
```
jmrossy: I don't think we'll want this secret in the monorepo. @kobigurk WDYT?

packages/phone-number-privacy/package.json
```diff
@@ -10,14 +10,19 @@
     "deploy": "firebase deploy --only functions",
     "clean": "tsc -b . --clean",
     "build": "tsc -b .",
-    "lint": "tslint --project ."
+    "lint": "tslint --project .",
+    "test": "jest --runInBand",
+    "copy-env-to-firebase": "firebase functions:config:set envs.private_key=$(grep PRIVATE_KEY .env | cut -d '=' -f 2-)"
   },
   "dependencies": {
+    "bls12377js-blind": "https://github.com/celo-org/bls12377js-blind#4a2b67da441fca3fab254b6d787800f465a8dcee",
```
jmrossy: Use `https://github.com/celo-org/bls12377js-blind#3a5ac03` instead which has my package.json fix in the bls lib

jmrossy (commented):
Good progress! Some small comments below

packages/phone-number-privacy/test/index.test.ts
```diff
@@ -0,0 +1,88 @@
+import FunctionsTest from 'firebase-functions-test'
+import { BLINDBLS } from 'bls12377js-blind'
+import * as chai from 'chai'
```
jmrossy: We mostly use jest across the monorepo, please replace chai with jest

packages/phone-number-privacy/src/index.ts
aslawson: i'll leave as is for now. another file could also want the raw key rather than the buffer, so we can always extract this logic to the config based on how it ends up being used

packages/phone-number-privacy/src/index.ts
aslawson: will do

scripts/key_placer.sh
aslawson: i'll remove for now and can discuss with kobi

packages/phone-number-privacy/package.json
aslawson: kk

scripts/key_placer.sh
kobigurk: Is it encrypted or not committed? If yes, that's fine.
If not, we should put it in the server's database.

aslawson: @kobigurk -- we could do either, encrypt it or upload to servers db.  is there a preference?

jmrossy: The reason I'd rather not have it here is because unlike most of our secrets, we don't want all cLabs devs to have access to it

jmrossy (approved):
LGTM!

PR #2997, opened 2020-02-28
🚧WIP [Wallet] Refactor PIN entering strategy

### LINKED ISSUE #3104: [Wallet] Refactor PIN entering/verifying throughout the app
### Expected Behavior

PIN entering/verifying logic is stable and transparent throughout the app

### Current Behavior

PIN entering/verifying logic is unstable and cumbersome throughout the app

i1skn:
### Description

This PR aims to improve stability and transparency of PIN entering process throughout the wallet. 
* Split `getPincode` into two different methods: `getPincode (to just fetch pincode from whatever source is now available` and `ensurePincode (to ensure, that user know a correct PIN)`.
* `app/saga -> handleNavigatePinProtected` logic moved to `NavigationService -> navigateProtected` .
* Change the way, how app is locked, when it's background or closed with introducing `NavigationService -> lockCurrentScreen` method. 
* Rename `PincodeConfirmation.tsx` to `PincodeEnter.tsx` as it logically make more sense.
### Caveats
* **TL;DR**: app would be locked immediately, when feature is enabled and it was closed/backgrounded. Previously it was only locked, when PIN is not cached. 
**Full description:** `Background.tsx` is removed. It was an intermediate screen when app is open after it was backgrounded. Its purpose was to support delayed locking. But it was showing up for some small amount of time every time app is opened. I think it's beneficial to remove it to have clean and concise logic and implement it right later. To do it right is an interesting topic itself - we need to update the UI when app is backgrounded or suspended (phone is locked).
#### Android
- [x] Create wallet
- [x] View Back Keys
- [x] Enable in `Menu -> Security` App lock with PIN
- [x] Background app and open it
- [x] Close the app and open it
- [x] Send TX, when PIN is not cached
- [x] Send TX, when PIN is cached
- [ ] Turn Saver Mode ON

#### Android device with Phone PIN
- [ ] Create wallet
- [ ] View Back Keys
- [ ] Send TX, when PIN is not cached
- [ ] Turn Saver Mode ON

#### iOS
- [x] Create wallet
- [x] View Back Keys
- [x] Enable in `Menu -> Security` App lock with PIN
- [x] Background app and open it
- [x] Close the app and open it
- [x] Send TX, when PIN is not cached
- [x] Send TX, when PIN is cached
- [ ] Turn Saver Mode ON

### Related issues

- Fixes #3104

### Backwards compatibility

Yes


packages/mobile/src/app/reducers.ts
```diff
@@ -39,11 +39,15 @@ export const appReducer = (
       return {
         ...state,
         ...getRehydratePayload(action, 'app'),
-        doingPinVerification: initialState.doingPinVerification,
```
jmrossy, line 42: Should we be reseting `requestingAndroidPermission` here like we did for doingPinPerm?

packages/mobile/src/app/saga.ts
```diff
@@ -208,21 +180,24 @@ function* watchAppState() {
 
 function* handleSetAppState(action: SetAppState) {
   const appLocked = yield select(getAppLocked)
-  const cachedPin = getCachedPincode()
+  // When requesting Android Permissions app state
+  // would background, so we do not want to lock
+  // the in this case
```
jmrossy: why not in this case?

packages/mobile/src/app/reducers.ts
```diff
@@ -20,11 +20,11 @@ const initialState = {
   numberVerified: false,
   language: null,
   doingBackupFlow: false,
-  doingPinVerification: false,
   analyticsEnabled: true,
   lockWithPinEnabled: false,
   appState: AppState.Active,
   locked: false,
+  requestingAndroidPermission: false,
```
jmrossy: Is this definitely just an android issue?

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -49,12 +51,59 @@ export function navigate(routeName: string, params?: NavigationParams) {
...
-export function navigateAfterPinEntered(routeName: string, params?: NavigationParams) {
-  navigate('Background', {
-    onUnlock() {
-      navigate(routeName, params)
-    },
-  })
+export async function ensurePincode(disableGoingBack: boolean): Promise<boolean> {
+  const pincodeType = pincodeTypeSelector(store.getState())
+
+  if (pincodeType === PincodeType.CustomPin) {
```
jmrossy, line 64: What if the pincode type isn't custom?

packages/mobile/src/pincode/PincodeEnter.tsx
```diff
@@ -115,16 +116,16 @@ class PincodeConfirmation extends React.Component<Props, State> {
   }
 
   onPressConfirm = () => {
+    const { fornoMode, navigation, currentAccount } = this.props
     const { pin } = this.state
-    const fornoMode = this.props
-    if (this.props.currentAccount && !fornoMode) {
+    const withVerification = navigation.getParam('withVerification')
+    // TODO: Implement PIN verification for forno mode
```
jmrossy: Can we do this as part of this PR? Otherwise, is there a task open already for it?

packages/mobile/src/qrcode/QRScanner.tsx
```diff
@@ -61,10 +63,19 @@ class QRScanner extends React.Component<Props, State> {
     navigate(Screens.QRCode)
   }
 
+  componentDidMount() {
+    if (Platform.OS === 'android') {
+      this.props.setRequestingAndroidPermission(true)
```
jmrossy: I worry that needing to set this on every screen where we possibly request perms is error prone. Someone will surely forget one eventually. Is this absolutely necessary? If so, can this go into `packages/mobile/src/utils/permissions.android.ts`

packages/mobile/src/utils/permissions.android.ts
```diff
@@ -26,7 +28,9 @@ export async function checkContactsPermission() {
 }
 
 async function requestPermission(permission: Permission, title?: string, message?: string) {
+  const { dispatch } = store
```
jmrossy: This use of the store outside saga-land potentially problematic, especially for async functions like this.
I don't fully understand the implications of accessing the store this way but I believe it's frowned upon. Let's look into it

jmrossy (commented):
Looks like an overall better approach :)

packages/mobile/src/app/reducers.ts
i1skn: 👍

packages/mobile/src/app/reducers.ts
i1skn: Yes, see here: https://reactnative.dev/docs/appstate
`background`: [Android] on another Activity (even if it was launched by your app)

packages/mobile/src/navigator/NavigationService.ts
i1skn: I've added "if" for `PincodeType.Unset`, but if it's not either of them - we pass it through (`PincodeType.PhoneAuth` is not supported and will be removed soon)

packages/mobile/src/pincode/PincodeEnter.tsx
i1skn: I've implemented it as a part of this PR

packages/mobile/src/qrcode/QRScanner.tsx
i1skn: It is already in `packages/mobile/src/utils/permissions.android.ts`. Here I've added it just because we use `RNCamera`, which does not use `packages/mobile/src/utils/permissions.android.ts` and therefor require a separate implementation.

packages/mobile/src/utils/permissions.android.ts
i1skn: As far as I know `dispatch` can be used outside saga-land. As an example, we are using it here:
https://github.com/celo-org/celo-monorepo/blob/cf13c179263060a968ef1086bbabe6230680f6ec/packages/mobile/src/app/App.tsx#L58-L60

packages/mobile/src/app/saga.ts
i1skn: Cause, when user would get a request permission notification we do not want to lock the app (user has not left it or backgrounded really). I've updated the comment, so it's more specific on reasoning.

i1skn:
@jmrossy `requestingAndroidPermission` looks hacky for me, but I do not know how to solve this problem differently 😞This link https://stackoverflow.com/questions/49006987/react-native-android-app-state-when-on-another-activity pretty much summarises my research. Let me know if you have any ideas on how to solve this better.

jmrossy:
@i1skn I guess I still don't understand what the _desired_ behavior is for pin code + app backgrounding. Can you or @nityas clarify this?

If we're trying to auto-force pin every time the app backgrounds, remember that low-end phones sometimes background an app any time you switch to another app, so users would be locked out 3 times during verification

cmcewen:
Just tested on android and got this error:

```
NavigationService@navigateProtected/PIN ensure error:undefined is not a function in TypeError: undefined is not a function
    at ensurePincode$ (http://localhost:8081/index.bundle?pla
```

i1skn:
> Just tested on android and got this error:
> 
> ```
> NavigationService@navigateProtected/PIN ensure error:undefined is not a function in TypeError: undefined is not a function
>     at ensurePincode$ (http://localhost:8081/index.bundle?pla
> ```

@cmcewen  Hey, I've noticed that there was a couple of changes, which were braking the build and I assume were a result of bad merge. Fixed them, please try again.

i1skn:
> If we're trying to auto-force pin every time the app backgrounds, remember that low-end phones sometimes background an app any time you switch to another app, so users would be locked out 3 times during verification

That is a good case. Yes, if this PR stays untouched - they would be logged out 3 times, correct. I see the following ways on how to prevent this:
* Remove the caveat describe in the description and add delayed log out after certain interval of time. This one could solve the issue. Also, we can/should give user an ability to chose this interval. I saw this exact behaviour in 1Password, so this should be doable. I have a feeling, though, this would require some native code intervention.
* We create a temporary "window" for these screens, needed for verification, so user would not be logged out. This would open a security breach, but will solve the UX issue until we solve delayed log out.

@cmcewen @jmrossy @nityas let me know your thoughts on this. Thanks!

jmrossy:
> @cmcewen @jmrossy @nityas let me know your thoughts on this. Thanks!

Thanks for clarifying. To be honest, I don't really like either of those options. These solutions seem kind of unstable and difficult to test. The complexity and the risk they introduce may not be worth this small feature. That's just my 2 cents though. Happy to schedule a sync to chat more about this.

nityas:
> > If we're trying to auto-force pin every time the app backgrounds, remember that low-end phones sometimes background an app any time you switch to another app, so users would be locked out 3 times during verification
> 
> That is a good case. Yes, if this PR stays untouched - they would be logged out 3 times, correct. I see the following ways on how to prevent this:
> 
> * Remove the caveat describe in the description and add delayed log out after certain interval of time. This one could solve the issue. Also, we can/should give user an ability to chose this interval. I saw this exact behaviour in 1Password, so this should be doable. I have a feeling, though, this would require some native code intervention.
> * We create a temporary "window" for these screens, needed for verification, so user would not be logged out. This would open a security breach, but will solve the UX issue until we solve delayed log out.
> 
> @cmcewen @jmrossy @nityas let me know your thoughts on this. Thanks!

Thanks for raising this ivan! Is it possible to add a delay time between pin requires? perhaps a default of 5 minutes and allow this to be configurable to either 5 mins or every time(several bank apps I use do this)

i1skn:
@nityas we just had a productive brainstorm on this issue with @jmrossy and here is how I propose we move forward:
* We implement delayed lock (remove caveat described in the description), thus eliminating the problem described above. I will try a slightly different way of achieving this, so it would not require significant amount of time (thanks @jmrossy for the idea).
* We will use PIN cache duration to decide when to lock the wallet. This make sense, cause we are using the same PIN to unlock the wallet and to perform blockchain related operations. So, if the PIN cached - there is no reason to lock the app and if there is - we should decrease PIN cache duration.

LMK your opinion on this one, thanks!

i1skn:
I'm closing this one in favour of https://github.com/celo-org/celo-monorepo/pull/3207

PR #3001, opened 2020-02-28
Fast Intrasite Nav

### LINKED ISSUE #2994: Lightening Links
Use Next.js Link Component so that on local links it will only load the parts needed instead of a full page refresh. 

Add a top bar loading indicator like youtube has.  

it will look like green line with a coin on the head.

aaronmgdr:
### Description

next/Link provides a very nice way to navigate between pages while only loading the new assets that are needed (kind of like turbo links in rails i suppose) Recently I realized that the way `Button` was written prevented this from ever working. 

while this vastly increases speed there isnt any natural indicator that immediately shows its working. therefore I have added a thin green progress bar at top, similar to the ones used by youtube, github and figma. 

### Tested

Chromium based Browsers, safari, and Firefox. 

Tested navigating between pages, clicking on one link and then another before it could load, outbound links, and elements that look like links but which we just have an onPress handler

### Other changes

small update to mobile menu spacing. 

### Related issues

- Fixes #2994

### Backwards compatibility

navigation still works 👍


packages/web/src/_page-tests/__snapshots__/code-of-conduct.test.tsx.snap
```diff
@@ -1519,6 +1525,8 @@ exports[`CodeOfConduct renders 1`] = `
             data-focusable={true}
             dir="auto"
             href="https://creativecommons.org/licenses/by-sa/3.0/"
+            onClick={[Function]}
```
aaronmgdr, line 1528: now the proper handlers are being set :)

packages/web/src/shared/Button.3.tsx
```diff
@@ -137,7 +131,11 @@ export default class Button extends React.PureComponent<ButtonsProps, State> {
         <View
           style={[{ alignItems: align }, this.props.kind === BTN.INLINE && inlineStyle.container]}
         >
-          {href ? <Link href={href}>{renderedButton}</Link> : renderedButton}
+          <ButtonComponent status={this.getStatus()} {...this.props}>
```
aaronmgdr, line 134: Doing it this way meant Link couldnt wrap the anchor and do its magic.

packages/web/src/shared/Button.3.tsx
```diff
@@ -165,26 +163,28 @@ interface Props {
...
-          ? primaryStyles.darkText
-          : primaryStyles.text,
-        style,
-        status === BTNStates.disabled && baseStyles.notAllowed,
-      ]}
-    >
-      {children}
-    </Text>
+    <Link href={href} passHref={true}>
+      <Text
```
aaronmgdr, line 167: when        `accessibilityRole="link"` `Text` is render as anchor, `Link` sets some function handlers on that anchor so that for local hrefs a full page load is not initiated

packages/web/src/shared/Footer.tsx
```diff
@@ -108,15 +108,17 @@ const Navigation = React.memo(function _Navigation({
...
           return (
             <View
               key={index}
               style={[
                 styles.menuItem,
                 !isVertical && index === 0 && { marginLeft: 0 },
-                isVertical
-                  ? linkIsToCurrentPage && styles.currentMenuItem
-                  : styles.verticalMenuItem,
+                isVertical && verticalItemStyle,
```
aaronmgdr, line 127: logic wasn't right before

packages/web/src/shared/Link.tsx
```diff
@@ -3,14 +3,19 @@ import * as React from 'react'
...
-  )
+export default ({ href, children, passHref, prefetch = false }: Props) => {
+  if (href) {
+    return (
+      <Link prefetch={prefetch} href={href} passHref={passHref}>
+        {children}
+      </Link>
+    )
+  } else {
+    return <>{children}</>
```
aaronmgdr, line 19: this is so Link is not rendered when we are just setting a `onPress` as rendering without a `href` caused issues that the tests picked up

packages/web/src/shared/Progress.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  return { isPageTurning, hasError, route }
+}
+
+export default function Progress() {
+  const { isPageTurning, hasError, route } = usePageTurner()
+
+  if (isPageTurning) {
+    const speed = getEffectiveConnection(navigator)
+    const animationSpeed = styles[speed]
```
aaronmgdr, line 57: as am added bonus on chromium browsers speed up or slow down the animation based on estimated download speed

aaronmgdr: 🔥

jmrossy:
Gonna leave this one to @cmcewen as I've never worked with NextJS

cmcewen (approved):

PR #3003, opened 2020-03-01
Azure createUiDefinition plumbing and first pass

silasbw:
### Description

createUiDefinition.json and bundle building for managed Azure applications.

### Tested

Tested manually with [createUiDefinition sandbox](https://portal.azure.com/?feature.customPortal=false#blade/Microsoft_Azure_CreateUIDef/SandboxBlade) and by creating a service catalog entry.


packages/azure/createUiDefinition.json
```diff
@@ -0,0 +1,170 @@
...
+         "subLabel": {
+           "preValidation": "Configure cluster",
+           "postValidation": "Done"
+         },
+         "bladeTitle": "Cluster settings",
+         "elements": [
+           {
+             "name": "userObjectId",
+             "type": "Microsoft.Common.TextBox",
+             "label": "userObjectId",
```
codyborn, line 22: Could you provide a helpful hint for the users to understand better what these fields are used for?

packages/azure/createUiDefinition.json
```diff
@@ -0,0 +1,170 @@
...
+           },
+           {
+             "name": "validatorAccountAddress",
+             "type": "Microsoft.Common.TextBox",
+             "label": "validatorAccountAddress",
+             "visible": true
+           },
+           {
+             "name": "validatorPrivateKey",
+             "type": "Microsoft.Common.TextBox",
```
codyborn: I think you can make this a password input to hide the key

packages/azure/createUiDefinition.json
```diff
@@ -0,0 +1,170 @@
...
+         "elements": [
+           {
+             "name": "attesterPostgreSQLUsername",
+             "type": "Microsoft.Common.TextBox",
+             "label": "attesterPostgreSQLUsername",
+             "visible": true
+           },
+           {
+             "name": "attesterPostgreSQLPassword",
+             "type": "Microsoft.Common.TextBox",
```
codyborn: Password field as well

codyborn (approved):

packages/azure/createUiDefinition.json
silasbw: oh, I'll remove this completely. I originally thought I needed to add an access policy for the user to access the Key Vaults at creation time (*e.g.*, if the ever wanted to read a secret value).  The user can create an access policy at any time that allows them to access secrets, so this isn't actually needed.

codyborn: Oh sorry, I meant that we can add a "tooltip" for all of these fields to give a human-readable description. There's a separate property for this:
`"toolTip": "The object Id of the AAD identity that will have access to Key Vault"`

silasbw: Ah, OK -- yes, I'll add tooltips for everything.

Related: https://github.com/celo-org/celo-monorepo/pull/3022

PR #3011, opened 2020-03-02
Adds e2e tests for layer 1 gold transfer freezing

alecps:
### Description

This PR adds e2e tests to transfer_tests in /celotool to ensure that gold transfers are filtered correctly in layer 1 when gold transfers are frozen according to whether the participating addresses are whitelisted. 

### Tested

e2e tests added for layer 1 changes in #[890](https://github.com/celo-org/celo-blockchain/pull/890)

### Other changes

Updates migrations to register TransferWhitelist with registry contract (needed to access whitelist from tests). Also updates migrations to freeze gold on deployment according to the migrationsConfig. 

### Related issues

- Depends on #[890](https://github.com/celo-org/celo-blockchain/pull/890)
- Fixes #[251](https://github.com/celo-org/celo-labs/issues/251)

### Backwards compatibility

Yes.


packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -829,4 +846,74 @@ describe('Transfer tests', function(this: any) {
...
+  describe('Transfers Frozen >', () => {
+    before(restartWithCleanNodes)
+
+    for (const syncMode of syncModes) {
+      describe(`${syncMode} Node >`, () => {
+        before(`start geth on sync: ${syncMode}`, async () => {
+          try {
+            await startSyncNode(syncMode)
+          } catch (err) {
+            console.debug('some error', err)
```
codyborn: Are these errors always safe to ignore?

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -829,4 +846,74 @@ describe('Transfer tests', function(this: any) {
...
+              const feeCurrency = await kit.registry.addressFor(CeloContract.StableToken)
+              try {
+                const res = await transferCeloGold(FromAddress, ToAddress, TransferAmount, {
+                  gas,
+                  feeCurrency,
+                })
+                await res.getHash()
+                assert.fail('no error was thrown')
+              } catch (error) {
+                assert.include(error.toString(), 'Returned error: transfers are currently frozen')
```
codyborn: I can't find this string in our repos "transfers are currently frozen". Can you point me to where it's generated?

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -829,4 +860,77 @@ describe('Transfer tests', function(this: any) {
       })
     }
   })
+
+  describe('Transfers Frozen >', () => {
```
codyborn, line 889: Some other ideas for tests:
- Sender/receiver removed from whitelists are blocked again
- *unfreeze* allows all parties to once again transfer (regardless of whitelist state)

packages/celotool/src/e2e-tests/transfer_tests.ts
alecps: Good call. I copied that part from 'Transfer with changed intrinsic gas cost >' block of tests, but I think after looking at it again it was probably left in to catch errors from the call to setIntrinsicGas(..) which I'm not using. I think taking out the catch here is probably the right thing. Thanks!

packages/celotool/src/e2e-tests/transfer_tests.ts
alecps: Yea that's added in my other PR in celo-blockchain https://github.com/celo-org/celo-blockchain/pull/890

packages/celotool/src/e2e-tests/transfer_tests.ts
alecps: just added those suggestions. thanks!

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -829,4 +867,78 @@ describe('Transfer tests', function(this: any) {
...
+            })
+            it('should transfer succesfully', async () => {
+              testTransferToken({
+                expectedGas: INTRINSIC_TX_GAS_COST + ADDITIONAL_INTRINSIC_TX_GAS_COST,
+                transferToken: CeloContract.GoldToken,
+                feeToken: CeloContract.GoldToken,
+              })
+            })
+            after('remove sender from transfer whitelist', async () => {
+              await setAddressWhitelist('http://localhost:8545', validatorAddress, [])
```
codyborn: can we also test that transfers are blocked by making this change?

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -829,4 +867,78 @@ describe('Transfer tests', function(this: any) {
...
+            })
+          })
+
+          after('unfreeze gold transfers', async () => {
+            await unfreeze('http://localhost:8545', validatorAddress, CeloContract.GoldToken)
```
codyborn: Similarly, can we add a check to ensure that unfreezing the contract allows transfers?

packages/celotool/src/e2e-tests/transfer_tests.ts
alecps: We are checking that in the block described 'when neither sender nor receiver is whitelisted'

packages/celotool/src/e2e-tests/transfer_tests.ts
alecps: This is kinda already checked in "Normal Transfer >" but for the sake of testing the unfreeze helper I've added a test block that calls unfreeze in the before hook. The top-level before hook calls freeze(), so this should check that unfreeze does what it's supposed to.

packages/celotool/src/e2e-tests/transfer_tests.ts
codyborn: I think it's important to test the state transition and not just the state itself. For example, if it defaults to state A (whitelist empty) and we transition to state B (whitelist populated), then we still need to test transition back to A (re-empty list). If there was a bug that allowed the whitelist to be added to but not removed from then this type of test would showcase that bug.

codyborn (approved):

codyborn (approved):
Nice

packages/celotool/src/e2e-tests/transfer_tests.ts
```diff
@@ -414,12 +448,41 @@ describe('Transfer tests', function(this: any) {
...
+      try {
+        const res = await transferCeloGold(FromAddress, ToAddress, TransferAmount, {
+          gas,
+          feeCurrency,
+        })
+        await res.waitReceipt()
+        assert.fail('no error was thrown')
+      } catch (error) {
+        assert.include(error.toString(), expectedError)
+      }
```
lucasege, line 474: Have you tried `test-utils.assertRevert`? Should cover this logic. Not sure how usable it is within e2e tests. Lots of examples in `accounts.ts` for reference

alecps: That looks like it's for expecting VM reverts caused by failed require statements etc. whereas this is trying to expect an error that originates within geth

PR #3019, opened 2020-03-03
account:register-metadata sanity check

### LINKED ISSUE #2643: celocli account:register-metadata call should perform some sanity check for the provided URL
### Expected Behavior

calling: `celocli account:register-metadata` should error out by default if the contents of the provided URL aren't a valid json data. (it can still have a --force flag to override this check)

### Current Behavior

currently `celocli account:register-metadata` doesn't do any sanity checking so it is very easy for users to make mistakes and provide incorrect URLs accidentally and  break their metadata setups completely.

mrsmkl:
### Description

Before setting metadata URL, try to show it's content.

### Tested

Only manual testing.

### Related issues

- Fixes #2643


packages/cli/src/commands/account/register-metadata.ts
```diff
@@ -17,21 +20,34 @@ export default class RegisterMetadata extends BaseCommand {
       required: true,
       description: 'The url to the metadata you want to register',
     }),
+    force: flags.boolean({ description: 'Ignore checks' }),
```
nategraf: Adding a bit more description on _which_ checks. (The `isAccount` check will still run)
```suggestion
    force: flags.boolean({ description: 'Ignore metadata validity checks' }),
```

packages/cli/src/commands/account/register-metadata.ts
```diff
@@ -17,21 +20,34 @@ export default class RegisterMetadata extends BaseCommand {
...
+        const metadata = await IdentityMetadataWrapper.fetchFromURL(metadataURL)
+        console.info('Metadata contains the following claims: \n')
+        await displayMetadata(metadata, this.kit)
+        console.info()
+      }
+      const accounts = await this.kit.contracts.getAccounts()
+      await displaySendTx('registerMetadata', accounts.setMetadataURL(metadataURL))
+    } catch (error) {
+      console.error(`Cannot set metadata URL ${metadataURL}: ${error.toString()}`)
+    }
```
nategraf: I would recommend keeping this `try` block smaller. Maybe something like:
```suggestion
   if (!res.flags.force) {
     try {
        const metadata = await IdentityMetadataWrapper.fetchFromURL(metadataURL)
        console.info('Metadata contains the following claims: \n')
        await displayMetadata(metadata, this.kit)
        console.info() // Print a newline.
      } catch (error) {
        console.error(`Metadata could not be retrieved from ${metadataURL}: ${error.toString()}`)
      }
    }
   
    const accounts = await this.kit.contracts.getAccounts()
    await displaySendTx('registerMetadata', accounts.setMetadataURL(metadataURL))
```

nategraf (approved):
Generally looks good! 🙌 Please consider the following minor comments

PR #3023, opened 2020-03-03
[protocol] Add wrapper support for multisig Reserve spender and Gover…

aslawson:
…nance approver + tests

### Description

Added MultiSigWrapper in contractkit.
Updated reserve spender and governance approver tooling in contractkit and cli. 

### Tested

Added integration tests for reserve spender and reserve wrapper.  Updated governance wrapper testing.

### Other changes

NA

### Related issues

- Fixes #[issue number here]
Follows work on https://github.com/celo-org/celo-monorepo/pull/2910

### Backwards compatibility

Updates tooling so tooling is compatible with the contract changes to use multisig for the governance approver and reserve spender.


packages/contractkit/src/contract-cache.ts
```diff
@@ -119,18 +123,21 @@ export class WrapperCache {
   getGovernance() {
     return this.getContract(CeloContract.Governance)
   }
+  getGovernanceApproverMultiSig() {
+    return this.getContract(CeloContract.GovernanceApproverMultiSig)
```
asaj: Are you sure this will work? I think these getters look these contracts up from the registry, and these contracts probably shouldn't be registered as they're not part of the "core" suite

packages/contractkit/src/contract-cache.ts
```diff
@@ -119,18 +123,21 @@ export class WrapperCache {
...
-  // getMultiSig() {
-  //   return this.getWrapper(CeloContract.MultiSig, newMultiSig)
-  // }
   // getRegistry() {
   //   return this.getWrapper(CeloContract.Registry, newRegistry)
   // }
   getReserve() {
     return this.getContract(CeloContract.Reserve)
   }
+  getReserveSpenderMultiSig() {
```
asaj: Ditto

packages/contractkit/src/wrappers/MultiSig.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Contract for handling multisig actions
+ */
+export class MultiSigWrapper extends BaseWrapper<MultiSig> {
+  /**
+   * Allows an owner to submit and confirm a transaction.
+   * @param index The index of the pending withdrawal to withdraw.
+   */
+  submitTransaction = proxySend(
```
asaj: What if this function took a `TransactionObject`? That way we could leave the existing wrappers as-is (i.e. multisig agnostic, which I think is preferable), and users would be able to submit transactions more generally, leveraging the other contractkit wrappers, e.g.

```
const governance = await kit.getGovernanceWrapper()
const approverMultisig = await kit.getMultisigWrapperAt('0x1234567890...')
const tx = await governance.approve(10)
await approverMultisig.submitOrConfirmTransaction(tx)
```

packages/contractkit/src/wrappers/MultiSig.ts
```diff
@@ -0,0 +1,28 @@
...
+  destination: Address,
+  transaction: string
+): SubmitTransactionsParams => {
+  return [destination, 0, stringToBytes(transaction)]
+}
+
+/**
+ * Contract for handling multisig actions
+ */
+export class MultiSigWrapper extends BaseWrapper<MultiSig> {
```
asaj, line 9: I think this is missing the ability to confirm transactions. For example, in a 2/2 multisig for the spender, for example, one signatory would need to submit the tx and the other one would need to confirm it.

For usability, you could imagine a `submitOrConfirmTransaction` function in this wrapper, that takes a transaction object, and iterates through the multisig transactions from latest tx ID to earliest, looking for an unexecuted transaction that matches the provided parameters. If a match is found, it calls `confirmTransaction`, and if not, it calls `submitTransaction`.

Wdyt?

packages/dev-utils/src/migration-override.json
```diff
@@ -46,9 +48,6 @@
     ],
     "frozen": false
   },
-  "governanceApproverMultiSig": {
-    "useMultiSig": false
```
asaj, line 51: 👏

packages/protocol/migrations/07_reserve_spender_multisig.ts
```diff
@@ -14,7 +11,8 @@ const initializeArgs = async (): Promise<any[]> => {
   ]
 }
 
-module.exports = deploymentForProxiedContract<ReserveSpenderMultiSigInstance>(
+// TODO @amyslawson has to register address for contractKit -- see if this is necessary
```
asaj: Ah, indeed 😄 

I think a good goal would be to avoid registering this contract. We could instead look the address up via the reserve contract whenever we need to use it, e.g. in CLI tools or contractkit tests

packages/protocol/migrations/23_governance_approver_multisig.ts
```diff
@@ -14,7 +11,8 @@ const initializeArgs = async (): Promise<any[]> => {
   ]
 }
 
-module.exports = deploymentForProxiedContract<GovernanceApproverMultiSigInstance>(
+// TODO @amyslawson has to register address for contractKit -- see if this is necessary
```
asaj: Ditto

packages/protocol/runTests.js
```diff
@@ -35,6 +35,7 @@ const migrationOverrides = {
   },
   reserve: {
     initialBalance: 100000000,
+    otherAddresses: [network.from],
```
asaj: Assuming this is to test actually transferring gold with the reserve spender? It might be more readable to put an arbitrary, hard-coded address, and include a comment as to what this is for, e.g.

```
// Add an arbitrary "otherReserveAddress" so that reserve spending can be tested.
```

packages/protocol/test/common/integration.ts
```diff
@@ -466,4 +470,30 @@ contract('Integration: Exchange', (accounts: string[]) => {
       assert.isTrue(finalReserve.lt(originalReserve))
     })
   })
+
+  describe('When transferring gold', () => {
```
asaj, line 474: Nice

asaj:
Please populate the PR template 🙏

packages/contractkit/src/wrappers/MultiSig.ts
aslawson: I see what you mean...the second person would just end up adding a new identical reaction to the list rather than confirming in this implementation.  

Iterating through each submitted tx doesn't seem super efficient, but I can't think of another option off the top of my head.  I'll dig in a little.

aslawson:
weird... I did fill out the template.  i wonder why it did not save on publish....

packages/contractkit/src/wrappers/MultiSig.ts
aslawson: > we could leave the existing wrappers as-is 

^ are you suggesting we have the caller (test, cli, contractkit users) reference the multisig itself?  Is this a common pattern for multisig?  I feel it makes sense to imbed the multisig reference in the Governance and Reserve wrappers to abstract some of the complexity -- the caller would not have worry about how its implemented (multisig or not) and would not make the mistake of ever trying to perform the action forgetting it needs to be sent through the multisig contract.

i suppose the negative here is that the spender/approver's address could be changed to not use multisig (so multisig is not permanently inherent to the smart contract implementations of these roles).

aslawson: i guess...how would the client know they need to submit request through a multisig contract?

i suppose we also lose clarity that the actions may not actually be performed by rather submitted, awaiting to be confirmed by another signatory.  so the method name might need to be something like submitOrCommitGoldTransfer

nambrot: Sorry for the drive-by comment here, but my interpretation of Asa's suggestion was to leave the Governance/Reserve wrappers indeed independent of the multisig itself. The tooling/client in this case would be aware that it is being called through a multisig.

> i suppose the negative here is that the spender/approver's address could be changed to not use multisig (so multisig is not permanently inherent to the smart contract implementations of these roles).

How is that a problem? That actually seems like the benefit of not coupling the multisig to the Governance/Reserve wrappers?

aslawson: its not a problem -- that was me arguing in favor of asa's suggestion

i just threw out pros/cons without much clarity for which is which

packages/cli/src/commands/governance/approve.ts
```diff
@@ -21,6 +21,9 @@ export default class Approve extends BaseCommand {
     const id = res.flags.proposalID
     this.kit.defaultAccount = account
     const governance = await this.kit.contracts.getGovernance()
+    const governanceApproverMultiSig = await this.kit.contracts.getGovernanceApproverMultiSig(
```
asaj: Consider`this.kit.contracts.getMultisigAt`

packages/contractkit/src/base.ts
```diff
@@ -15,10 +15,12 @@ export enum CeloContract {
   GasPriceMinimum = 'GasPriceMinimum',
   GoldToken = 'GoldToken',
   Governance = 'Governance',
+  GovernanceApproverMultiSig = 'GovernanceApproverMultiSig',
```
asaj: Probably fine to just have a single `MultiSig` here, given the contract code is the same for all of them. Ditto everywhere else

packages/contractkit/src/wrappers/MultiSig.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Address } from '../base'
+import { MultiSig } from '../generated/types/MultiSig'
+import { BaseWrapper, proxyCall, stringToBytes, toTransactionObject } from './BaseWrapper'
+
+/**
+ * Contract for handling multisig actions
+ */
+export class MultiSigWrapper extends BaseWrapper<MultiSig> {
+  /**
+   * Allows an owner to submit and confirm a transaction.
```
asaj, line 11: nit: Please expand on this. e.g.
```
If an unexecuted transaction matching `txObject` exists on the multisig, adds a confirmation to that tx ID. Otherwise, submits the `txObject` to the multisig.
```

packages/contractkit/src/wrappers/MultiSig.ts
```diff
@@ -0,0 +1,34 @@
...
+   * @param index The index of the pending withdrawal to withdraw.
+   */
+  async submitOrConfirmTransaction(destination: string, txObject: TransactionObject<any>) {
+    const data = stringToBytes(txObject.encodeABI())
+    const transactionCount = await this.contract.methods.getTransactionCount(true, true).call()
+    let transactionId
+    for (transactionId = 0; transactionId < Number(transactionCount); transactionId++) {
+      const transaction = await this.contract.methods.transactions(transactionId).call()
+      if (transaction.data === data && transaction.destination === destination) {
```
asaj: Missing a check for `value`

packages/contractkit/src/wrappers/MultiSig.ts
```diff
@@ -0,0 +1,34 @@
...
+export class MultiSigWrapper extends BaseWrapper<MultiSig> {
+  /**
+   * Allows an owner to submit and confirm a transaction.
+   * @param index The index of the pending withdrawal to withdraw.
+   */
+  async submitOrConfirmTransaction(destination: string, txObject: TransactionObject<any>) {
+    const data = stringToBytes(txObject.encodeABI())
+    const transactionCount = await this.contract.methods.getTransactionCount(true, true).call()
+    let transactionId
+    for (transactionId = 0; transactionId < Number(transactionCount); transactionId++) {
```
asaj: nit: consider iterating in reverse, as users are likely going to be calling this on recently added txs

packages/contractkit/src/wrappers/MultiSig.ts
asaj: nit: Consider `return toTransactionObject(this.kit, this.contract.methods.confirmTransaction(transactionId))` here instead of breaking. IMO would make the next few lines mildly more readable

packages/contractkit/src/wrappers/Reserve.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { ReserveSpenderMultiSigWrapper } from './ReserveSpenderMultiSig'
+
+testWithGanache('Reserve Wrapper', (web3) => {
+  const kit = newKitFromWeb3(web3)
+  let accounts: string[] = []
+  let reserve: ReserveWrapper
+  let reserveSpenderMultiSig: ReserveSpenderMultiSigWrapper
+  const otherAddress = '0x91c987bf62D25945dB517BDAa840A6c661374402'
+  const otherSpender = '0x4404ac8bd8F9618D27Ad2f1485AA1B2cFD82482D'
+  // TODO @amyslawson will ganache always give same proxy address for testing?  can we ensure that?
```
asaj: Nope, the address a contract is deployed at is a function of the deploying account's nonce and the contract code being deployed. So if the migrations are changed such that transactions are added or removed, the contract would wind up being deployed to another address.

If you need to pull the reserve spender address, you should be able to search for the `ReserveSpenderAdded` on the `Reserve`. @lucasege had to do something similar recently for testing `ReleaseGold`

packages/contractkit/src/wrappers/Reserve.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    const tx2 = await reserve.transferGold(otherAddress, 10)
+    const multisigTx2 = await reserveSpenderMultiSig.submitOrConfirmTransaction(
+      reserve.address,
+      tx2.txo
+    )
+    const events2 = await (await multisigTx2.sendAndWaitForReceipt({ from: otherSpender })).events
+    expect(events2 && !events2.Submission && events2.Confirmation && events2.Execution)
+  })
+
+  test('test does not transfers gold if not spender', async () => {
```
asaj: nit: "transfer"

packages/contractkit/src/wrappers/Reserve.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { newKitFromWeb3 } from '../kit'
+import { ReserveWrapper } from './Reserve'
+import { ReserveSpenderMultiSigWrapper } from './ReserveSpenderMultiSig'
+
+testWithGanache('Reserve Wrapper', (web3) => {
+  const kit = newKitFromWeb3(web3)
+  let accounts: string[] = []
+  let reserve: ReserveWrapper
+  let reserveSpenderMultiSig: ReserveSpenderMultiSigWrapper
+  const otherAddress = '0x91c987bf62D25945dB517BDAa840A6c661374402'
```
asaj: nit: "otherReserveAddress"

asaj: nit: Where are these addresses coming from? Please add a comment

asaj (commented):
Very nice! Mostly nits at this point.

One thing I'm noticing is that we don't have a reserve spender CLI command, can you please add in a follow-up PR?

packages/contractkit/src/wrappers/MultiSig.ts
aslawson: yes -- much cleaner

packages/contractkit/src/base.ts
aslawson: The actual contract is different though... there is a GovernanceApproverMultiSig.sol and ReserveSpenderMultiSig.sol.  they both extend multisig, so it seems it would work, just feels a bit inconsistent

aslawson: i've updated to change to just generic multisig

packages/cli/src/commands/governance/approve.ts
```diff
@@ -21,6 +21,9 @@ export default class Approve extends BaseCommand {
     const id = res.flags.proposalID
     this.kit.defaultAccount = account
     const governance = await this.kit.contracts.getGovernance()
+    const governanceApproverMultiSig = await this.kit.contracts.getMultiSig(
```
asaj, line 24: Please consider making this command work both with and without a multisig, e.g. by adding a `--multisig` flag, or even adding an automatic check that looks up whether or not the current approver appears to be a smart contract or not. Fine to do as a follow-up PR

asaj (approved):

packages/cli/src/commands/governance/approve.ts
aslawson: kk -- i will do as follow up.  i have another pr in progress adding reserve cli tooling, so i'll include this

PR #3028, opened 2020-03-04
TGCSO - leaderboard adjustment 1.1

aaronmgdr:
### Description

updates points/btu for AGx1 to 1,187,607 as result of  correction of security audit score.

### Tested

deployed to dev.celo.org/validators 

### Other changes

use identity as `key` instead of index (which we know isnt a great pratice)

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


nategraf (approved):
LGTM! Thanks 🙌

PR #3030, opened 2020-03-05
Adding critical features for ReleaseGold contract

lucasege:
### Description

Added three critical features for the `ReleaseGold` contract:
* Added `transfer` fn so earned stable tokens can be transferred out of the contract.
* Added `expire` fn with corresponding hard-coded `EXPIRATION_TIME` value to allow recovery of funds if `beneficiary` loses their key. 2 year default.
* Changed use of openzeppelin's `reentrancyGuard` back to ours since `ReleaseGold` is intended as a proxied contract.

One consideration, ideally we would have added `expirationTime` to `ReleaseGold`'s initialize fn, but we already have 14 parameters and additional local variables so I was unable to add another parameter without hitting solc's `stack too deep` error (16 max local vars).

I also added two more wrapper fns that were previously overlooked `authorizeValidatorWithPublicKey` and `authorizeValidatorWithKeys`.

### Tested

Added unit tests to `releaseGold.ts`

### Other changes

Had to modify `mockStableToken` to allow for `transfer` testing.

Fixed parameter comment ordering.

### Backwards compatibility

Yes


packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -131,8 +137,26 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
...
+    _;
+  }
+
   function() external payable {} // solhint-disable no-empty-blocks
 
+  /**
+   * @notice Wrapper function for stable token transfer function.
+   */
+  function transfer(address to, uint256 value) external onlyWhenInProperState {
+    IERC20(registry.getAddressForOrDie(STABLE_TOKEN_REGISTRY_ID)).transfer(to, value);
```
nambrot, line 157: Does `getStableToken` from `UsingRegistry` not work?

nambrot (approved):

packages/protocol/contracts/governance/ReleaseGold.sol
lucasege: Unfortunately no, `getStableToken` returns `IStableToken` which doesn't expose the `transfer` fn. Since `StableToken` is ERC20 compliant, casting it to `IERC20` is safe here.
Asa and I decided to just use the typecast rather than expose transfer on `IStableToken`, if you think it's better to add to the interface let me know

packages/protocol/test/governance/release_gold.ts
```diff
@@ -930,6 +1037,77 @@ contract('ReleaseGold', (accounts: string[]) => {
...
+            assert.equal(isRevoked, true)
+          })
+
+          it('should set the released balance at revocation to total withdrawn', async () => {
+            const [, releasedBalanceAtRevoke] = await releaseGoldInstance.revocationInfo()
+            // 0 gold withdrawn at this point
+            assertEqualBN(releasedBalanceAtRevoke, 0)
+          })
+
+          it('should allow refund of all remaining gold', async () => {
```
codyborn, line 1097: Can you check if there's a test to ensure a non-releaseOwner can't withdraw after the revocation?

codyborn (approved):
LGTM

packages/protocol/test/governance/release_gold.ts
lucasege: There is a test that a beneficiary cannot withdraw more than their released gold at revoke time, which covers this logic, but potentially I can add a test later of a totally associated address, but there's no reason that wouldn't work.

PR #3031, opened 2020-03-05
Blockscout package fixes

jcortejoso:
### Description

Fix blockscout upgrade and migrate deployment commands.

### Tested

Tested on testnet

### Other changes

Removed a unused variable in terraform.

### Backwards compatibility

If the environment has an old version of blockscout helm package, it may require destroy&initial to be compatible.


packages/celotool/src/cmds/deploy/destroy/blockscout.ts
```diff
@@ -13,11 +13,9 @@ export const builder = {}
 export const handler = async (argv: DestroyArgv) => {
   await switchToClusterFromEnv()
 
-  const instanceName = `${argv.celoEnv}${fetchEnvOrFallback('BLOCKSCOUT_DB_SUFFIX', '')}`
-  const helmReleaseName = `${argv.celoEnv}-blockscout${fetchEnvOrFallback(
-    'BLOCKSCOUT_DB_SUFFIX',
-    ''
-  )}`
+  const dbSuffix = fetchEnvOrFallback('BLOCKSCOUT_DB_SUFFIX', '')
```
tkporter: Could you add `BLOCKSCOUT_DB_SUFFIX` to the `envVar` enum in `src/lib/env-utils.ts` and use that instead elsewhere as well?

packages/celotool/src/lib/blockscout.ts
```diff
@@ -68,6 +68,7 @@ async function helmParameters(
     `--set blockscout.db.username=${blockscoutDBUsername}`,
     `--set blockscout.db.password=${blockscoutDBPassword}`,
     `--set blockscout.db.connection_name=${blockscoutDBConnectionName.trim()}`,
+    `--set blockscout.db.drop=${fetchEnvOrFallback('BLOCKSCOUT_DROP_DB', 'false')}`,
```
tkporter: Ditto, adding this to `envVar` would be nice

packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -265,13 +265,17 @@ export async function grantRoles(serviceAccountName: string, role: string) {
...
-export async function retrieveCloudSQLConnectionInfo(celoEnv: string, instanceName: string) {
+export async function retrieveCloudSQLConnectionInfo(
+  celoEnv: string,
+  instanceName: string,
+  dbSuffix: string
+) {
   await validateExistingCloudSQLInstance(instanceName)
   const [blockscoutDBUsername] = await execCmdWithExitOnFailure(
-    `kubectl get secret ${celoEnv}-blockscout -o jsonpath='{.data.DATABASE_USER}' -n ${celoEnv} | base64 --decode`
+    `kubectl get secret ${celoEnv}-blockscout${dbSuffix} -o jsonpath='{.data.DATABASE_USER}' -n ${celoEnv} | base64 --decode`
```
tkporter: Could you pull out `${celoEnv}-blockscout${dbSuffix}` into a const so making changes in the future is a little more straightforward?

packages/helm-charts/blockscout/templates/blockscout-indexer.deployment.yaml
```diff
@@ -31,6 +31,7 @@ spec:
         - -c
         args:
         - |
+           sleep 5
```
tkporter, line 34: Curious (just for my own knowledge) what the need is for this & the other one in the web yaml?

packages/helm-charts/blockscout/templates/blockscout-migration.job.yaml
```diff
@@ -12,17 +12,24 @@ spec:
...
+           mix do ecto.migrate
         resources:
           requests:
             memory: 250M
             cpu: 200m
         env:
-{{ include "celo.blockscout-env-vars" .  | indent 10 }}
+        - name: DROP_DB
+          value: "{{ default "false" .Values.blockscout.db.drop }}"
+{{ include "celo.blockscout-env-vars" .  | indent 8 }}
```
tkporter, line 32: Is this indent supposed to change?

packages/helm-charts/blockscout/values.yaml
```diff
@@ -14,5 +14,6 @@ blockscout:
     # username: blockscout
     # password: password
     name: blockscout
+    drop: "False"
```
tkporter: nit: I think we should keep this as fully lowercase

packages/terraform-modules/testnet/modules/tx-node-load-balancer/variables.tf
```diff
@@ -18,11 +18,6 @@ variable forno_host {
   description = "The host name to use for the tx node forno setup"
 }
 
-variable gcloud_credentials_path {
```
tkporter, line 21: thanks!

packages/celotool/src/cmds/deploy/destroy/blockscout.ts
```diff
@@ -13,11 +13,9 @@ export const builder = {}
...
+  const instanceName = `${argv.celoEnv}${dbSuffix}`
+  const helmReleaseName = `${argv.celoEnv}-blockscout${dbSuffix}`
```
tkporter: Seems like in the destroy, initial, and upgrade command we're creating the same strings. Could you add the common code where the helm release name is generated into a shared function? That way future changes will be easier

tkporter (commented):
Some general comments, nothing major at all

packages/helm-charts/blockscout/templates/blockscout-indexer.deployment.yaml
jcortejoso: My intention was waiting for the migration job to be complete, or at least starting the migration job before any other db process starts, so the db could be dropped if needed. 
Really not very happy with this but cannot think on a simple approach to sync between different pods. Maybe a better approach would be an init container in the indexer that has some capabilities to detect when should the database be dropped and/or migrated.

packages/helm-charts/blockscout/templates/blockscout-migration.job.yaml
jcortejoso: [Both syntax are valid](https://yaml.org/spec/1.2/spec.html#id2799784):
```
env:
- name: myvar
  value: myvalue
- name: myvar2
  value: myvalue2
```
or
```
env:
  - name: myvar
    value: myvalue
  - name: myvar2
    value: myvalue2
```

I changed to use the same indenting as we are using in this file for the other keys (like `containers`) but both should be valid as far as we indent `DROP_DB` with the same spaces.

jcortejoso:
@tkporter thanks for reviewing and the comments. I think I addressed most of them.

H34D (approved):
LGTM

PR #3046, opened 2020-03-07
[Wallet] Fix problems with tx params in wallet dappkit tx signer

jmrossy:
### Description

Fix problems with tx params in wallet dappkit tx signer

### Tested

Locally with Savings Circle + Wallet

### Backwards compatibility

yes


marekolszewski (approved):
Looks good!

PR #3047, opened 2020-03-08
[web] Alliance Page  (PART A - client side)

aaronmgdr:
### Description

**Creation of New Top Level Page For the Celo Alliance.** 

- add darken on hover to transparent header
- alliance members by category and ability to filter them
- apply form send data to airtable (and mailing list)
- change version to 4.0, to keep inline with figma versions
- create checkbox which use a real html input type checkbox, while appearing nice. 

### Related issues
DEPENDS ON  https://github.com/celo-org/celo-monorepo/pull/3048

### Tested

Screen Shot 2020-03-08 at 2 07 54 PM
Screen Shot 2020-03-08 at 2 08 21 PM
Screen Shot 2020-03-08 at 2 08 37 PM
Screen Shot 2020-03-08 at 2 08 47 PM

Screen Shot 2020-03-08 at 2 08 54 PM
Screen Shot 2020-03-08 at 2 09 02 PM
Screen Shot 2020-03-08 at 2 09 10 PM
Screen Shot 2020-03-08 at 2 09 27 PM

### Other changes

* Switched Header to use screenSize Context instead of the `<Responsive>` as Responsive did not play well with `<Hoverable>` 

* make use of Next.js `api` (instead of making index.ts, place route handle files in `pages/api` 

* move validationErrors trans to common for easier sharing

* start enforcing style that local links are defined in `pagePaths` while links to other sites are stored in `CeloLinks` 

- Fixes https://github.com/celo-org/celo-labs/issues/191


packages/web/src/about/About.tsx
```diff
@@ -145,58 +149,13 @@ export class About extends React.Component<Props & I18nProps> {
   }
 }
 
-const BeautifulQuote = withScreenSize(
```
aaronmgdr, line 148: moved to own file

packages/web/src/alliance/Affirmations.tsx
```diff
@@ -0,0 +1,152 @@
...
+        </View>
+      </Cell>
+    </GridRow>
+  )
+}
+
+function TransItalic({ i18nKey }: { i18nKey: string }) {
+  return (
+    <Trans ns={NameSpaces.alliance} i18nKey={i18nKey}>
+      <Text style={textStyles.italic}>""</Text>
```
jmrossy, line 167: nit: not sure why you're using the Trans wrapper here instead of use `t`

packages/web/src/alliance/Affirmations.tsx
```diff
@@ -0,0 +1,152 @@
...
+      <Exemplar
+        belief={<TransItalic i18nKey="affirmations.wfpBelief" />}
+        copy={t('affirmations.wfpCopy')}
+        image={wfp}
+        preview={wfpPreview}
+        logo={<Image resizeMode="contain" source={wfpLogo} style={orgStyle.wfpLogo} />}
+        button={{
+          text: t('affirmations.wfpButton'),
+          href:
+            'https://medium.com/celoorg/how-to-design-for-all-stories-from-tanzania-refugees-8b34594d64ae',
```
jmrossy, line 72: here you use `celoorg`, below you use `celohq`

packages/web/src/alliance/Affirmations.tsx
```diff
@@ -0,0 +1,152 @@
...
+    <Trans ns={NameSpaces.alliance} i18nKey={i18nKey}>
+      <Text style={textStyles.italic}>""</Text>
+    </Trans>
+  )
+}
+
+const SMALLER = 0.8
+
+const mobileStyles = StyleSheet.create({
+  laboratoria: { width: 200 * SMALLER, height: 35 * SMALLER },
```
jmrossy: Nit: should these reference the regular styles? otherwise why have a SMALLER factor, why not just set the numbers directly?

packages/web/src/alliance/Main.tsx
```diff
@@ -0,0 +1,163 @@
...
+      background: `linear-gradient(90deg, rgba(255,255,255,1) 86%, rgba(12,218,110,1) 97%, rgba(255,101,83,1) 98%, rgba(159,105,255,1) 100%, rgba(82,182,255,1) 100%)`,
+    },
+    '100%': {
+      opacity: 0.9,
+      backgroundColor: colors.white,
+    },
+  },
+]
+
+const blendStyle = { mixBlendMode: 'darken' }
```
jmrossy, line 139: why is this out here?

packages/web/src/alliance/Members.tsx
```diff
@@ -0,0 +1,238 @@
...
+function initialState() {
+  return Object.keys(CategoryEnum).map((key: CategoryEnum) => {
+    return {
+      name: key,
+      records: [],
+    }
+  })
+}
+
+const ALL = 'all'
```
jmrossy, line 43: should this be part of an enum or something like that?

packages/web/src/alliance/Members.tsx
```diff
@@ -0,0 +1,238 @@
...
+  const [selectedFilter, setFilter] = React.useState(ALL)
+
+  const { isTablet, isMobile } = useScreenSize()
+
+  React.useEffect(() => {
+    const signal = { aborted: false }
+    // sometimes it is nessessary to break a rule
+    // https://github.com/facebook/react/issues/14326
+    // tslint:disable-next-line: no-floating-promises
+    gatherAllies(setAllies, signal)
```
jmrossy, line 57: I had a look at the react link you linked to. This doesn't match the async effect pattern Abramov provided. Why let this promise float?

packages/web/src/alliance/PinWheel.tsx
```diff
@@ -0,0 +1,213 @@
...
+    top: 0,
+    right: 0,
+    left: 0,
+    bottom: 0,
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+})
+
+function getKeyframes() {
```
jmrossy: Nit why is this wrapped in a getter if we don't have params for it? Should this be an array at module scope?

packages/web/src/alliance/RisingPinWheelForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { StyleSheet, View } from 'react-native-web'
+import CollectiveMission from 'src/alliance/CollectiveMission'
+import PinWheel from 'src/alliance/PinWheel'
+import SignupForm from 'src/alliance/SignupForm'
+import { Cell, GridRow, Spans } from 'src/layout/GridRow'
+import { standardStyles } from 'src/styles'
+
+export default function RisingPinWheelForm() {
+  return (
+    <View style={[standardStyles.darkBackground]}>
```
jmrossy: nit: array of 1

packages/web/src/forms/FormComponents.tsx
```diff
@@ -189,13 +205,93 @@ export function LabeledInput({ name, multiline, hasError, value, onInput, label
...
   )
 }
+
+interface CheckboxProps {
+  checked: boolean
+  name: string
+  onPress: (x: any) => void
+}
+
+export function Checkbox({ checked, onPress, name }: CheckboxProps) {
```
jmrossy, line 226: Long term, it'd be better for these to each be in their own file to enable better code splitting and lazy loading.

packages/web/src/header/Header.3.tsx
```diff
@@ -164,13 +150,40 @@ export class Header extends React.PureComponent<Props, State> {
     this.setState({ bannerHeight: height })
   }
 
+  allWhiteLogo = () => {
+    return this.props.router.pathname === menu.ABOUT_US.link && !this.state.belowFoldUpScroll
```
jmrossy, line 154: This seems hacky, is there a better way that doesn't hardcode in a pathname?

packages/web/src/shared/Outbound.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default function Outbound({ url }: { url: string }) {
+  return (
+    <a href={externalizeURL(url)} target="_blank">
+      <External size={12} color={colors.dark} />
+    </a>
+  )
+}
+
+export function externalizeURL(url: string) {
```
jmrossy, line 13: what's this for?

jmrossy (commented):
Overall seems good, just some minor comment below

packages/web/src/alliance/Affirmations.tsx
aaronmgdr: with `t` its not possible place different styles on parts of a phrase. IE only Italicize one word in a sentence. with `<Trans>` Can do it like `i18nkey: "I would like a <0>lot</0> of bread"  and the the 0th child given to <trans> replaces the `<0>`

packages/web/src/alliance/Affirmations.tsx
aaronmgdr: niice find. gracias.

packages/web/src/alliance/Main.tsx
aaronmgdr: oddly whenever blendMode is part of Stylesheet.create it gets lost

packages/web/src/alliance/Affirmations.tsx
aaronmgdr: you're very right. ha

packages/web/src/alliance/Members.tsx
aaronmgdr: here is my understanding.  `useEffect` is typed to take a function that returns void or a cleanup function. NOT a promise. so cant do await on this as that would require an async function which always return promises.

packages/web/src/header/Header.3.tsx
aaronmgdr: I could define a Set of all path names which should use an all white logo? at module scope

packages/web/src/shared/Outbound.tsx
aaronmgdr: if i throw `www.celo.org` straight into a `href` on an anchor its going to direct the person to `celo.org/www.celo.org`.

whereas `//www.celo.org` is known to be a domain. 

this way any URLs set in the airtable will go to the right spot.

packages/web/src/alliance/Members.tsx
aaronmgdr: i guess it could be part of the CategoryEnum but then thee will be several places it would need to be filtered out which seems extra complicated.

jmrossy (approved):
Think you forgot this comment: https://github.com/celo-org/celo-monorepo/pull/3047#discussion_r389437755

But otherwise lgtm

PR #3048, opened 2020-03-09
[web] ALLIANCE (Part B - node server side)

aaronmgdr:
### Description

This contains the necessary server side changes for [Alliance page](https://github.com/celo-org/celo-monorepo/pull/3047)

* fetching of alliance members from airtable

* adding new potential members to another airtable

### Other changes

moved contributors route to  its own file in pages/api/

### Related issues

- Fixes # https://github.com/celo-org/celo-labs/issues/191

### DIFF
--- a/packages/web/pages/api/alliance.ts
+++ b/packages/web/pages/api/alliance.ts
@@ -1,6 +1,6 @@
 import { NextApiRequest, NextApiResponse } from 'next'
 import getAllies from 'server/Alliance'
-import respondToError from 'server/respondToError'
+import respondError from 'server/respondError'
 
 export default async function(req: NextApiRequest, res: NextApiResponse) {
   try {
@@ -11,6 +11,6 @@ export default async function(req: NextApiRequest, res: NextApiResponse) {
       res.status(405)
     }
   } catch (e) {
-    respondToError(res, e)
+    respondError(res, e)
   }
 }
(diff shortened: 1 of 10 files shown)


packages/web/server/respondToError.ts
```diff
@@ -0,0 +1,11 @@
+interface Response {
+  status: (code: number) => Response
+  json: (jsonCompatible: object) => void
+}
+
+export default function respondToError(
```
jmrossy: Nit: This name makes me think it's a handler for when an error happens but looks like it's actually sending the error response itself. Maybe just `respondError`?

packages/web/pages/api/alliance.ts
```diff
@@ -0,0 +1,16 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import getAllies from 'server/Alliance'
+import respondToError from 'server/respondToError'
+
+export default async function(req: NextApiRequest, res: NextApiResponse) {
+  try {
+    if (req.method === 'GET') {
```
jmrossy, line 7: I don't know Next.JS well but seems like there should be a way to specify a `get` endpoint so you don't have to code in this 405 handling manually?

jmrossy (approved):
LGTM, two optional comments

packages/web/pages/api/alliance.ts
aaronmgdr: funnily enough this is pretty much what the docs show. 

If it was me id have it so that the handler for each method was a named export ie `export function get()`  `export function patch()`  `export function post()`

aaronmgdr: The second example of 
https://nextjs.org/docs/api-routes/introduction 

The first doesnt have it but it sounds like it will respond to just any method.

PR #3049, opened 2020-03-09
Fix codecov config

jeanregisser:
Trying to fix incorrect codecov status reported for the protocol test.

See https://community.codecov.io/t/codecov-reporting-unexpected-status-update-for-deleted-project/1069


jmrossy (approved):
Nice!

jeanregisser:
Though it still reports statuses for `codecov/patch` and `codecov/project` which are disabled in the config, I'm gonna merge this now as it's better than our current situation where all commits report a failing status for `codecov/project/protocol`.

I'll send a follow up PR once I hear back from codecov's support in https://community.codecov.io/t/codecov-reporting-unexpected-status-update-for-deleted-project/1069/16

PR #3058, opened 2020-03-10
[Wallet] Use new visual assets in backup and import wallet flows

### LINKED ISSUE #2716: Add new assets to backup key + safeguards flow
### Expected Behavior

New Assets for interstitial screen before safeguard+backup key flows added here: [Figma](https://www.figma.com/file/HGX4zx5XSUeejjfGGI5x3R/Backup-Key-and-Recovery?node-id=0%3A1)

### Current Behavior

currently using default phone icon

jmrossy:
### Description

Use new visual assets in backup and import wallet flows (see screenshots)

### Tested

Did import and backup flows on android and ios

### Other changes

Fix a spacing issue with backup phrase input field

### Related issues

- Fixes #2716 

### Backwards compatibility

Yes

IMG_6198
IMG_6200
IMG_6201
IMG_6202
IMG_6203
IMG_6204
IMG_6205


packages/mobile/src/backup/BackupIntroduction.tsx
```diff
@@ -97,16 +99,19 @@ class BackupIntroduction extends React.Component<Props> {
     return (
       <SafeAreaView style={styles.container}>
         <ScrollView contentContainerStyle={styles.scrollContainer}>
-          <Image source={backupIcon} style={styles.logo} />
-          <Text style={styles.h1}>{t('backupAndRecovery')}</Text>
           {!backupCompleted && (
             <>
+              <BackupIntroIcon style={styles.logo} width={180} height={138} />
```
jeanregisser, line 104: nit: what happens if you specify a width and height with a different ratio than the source svg?
wondering if we should compute the other dimension to prevent incorrect usages.

jeanregisser (approved):
LGTM :+1:

packages/mobile/src/backup/BackupIntroduction.tsx
jmrossy: Yeah I tested this. The SVG maintains it's scale but then includes a extra whitespace based on the values you set. Which may be helpful in certain cases so I think it'd be good to still let devs set the sizes

PR #3098, opened 2020-03-12
Adding genesis block accounts

lucasege:
### Description

Adding genesis accounts to `genesis_constants` in order to generate genesis block properly.

Also changed the address for the owner of RegistryProxy.

Created the `.env.rc0` and `.env.mnemonic.rc0.enc`, although both could use some review.

### Tested

Generated genesis block using celotool, looked Ok at a glance.

### Related issues

- Fixes #[issue number here]


packages/celotool/src/lib/genesis_constants.ts
```diff
@@ -40,3 +40,74 @@ export const CONTRACT_OWNER_STORAGE_LOCATION =
...
   '0x63746963616c2062797a616e74696e65206661756c7420746f6c6572616e6365'
+
+export const INITIAL_ACCOUNTS = [
+  {
+    address: '0x36940810BfDB329B31e38d3e97aFD673081B497C', // ReserveCo genesis address
+    balance: '120000000000000000000000000', // 120MM
+  },
+  {
+    address: '0xfCf982bb4015852e706100B14E21f947a5Bb718E', // LLC genesis address
+    balance: '200000000000000000000000000', // 200MM
```
aslawson: why aren't these split into several addresses?

lucasege: Had this same question - the answer I was given from Asa was that the split addresses are for amounts that are planned to be used often and so are vulnerable. 
The reserve address here will immediately be transferred to the reserve contract so its fine, and the LLC eventually gets transferred to the LLC treasury

aslawson: I see -- thanks for the explanation  👍

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+CELOSTATS_FRONTEND_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/celostats-frontend"
+CELOSTATS_FRONTEND_DOCKER_IMAGE_TAG="db16a1a4d9882cfed93ffa9f61cfa9f80cadfe18"
+CELOSTATS_TRUSTED_ADDRESSES=""
+CELOSTATS_BANNED_ADDRESSES=""
+CELOSTATS_RESERVED_ADDRESSES=""
+
+GETH_NODE_DOCKER_IMAGE_REPOSITORY="us.gcr.io/celo-testnet/geth"
+# When upgrading change this to latest commit hash from the master of the geth repo
+# `geth $ git show | head -n 1`
+GETH_NODE_DOCKER_IMAGE_TAG="2d730430a7e22f5890dfb6ac74cb9f7297a3630d"
```
asaj: Can we update this, and the one below it, to the latest commit hash from `celo-blockchain`

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+
+CELOTOOL_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/celo-monorepo"
+CELOTOOL_DOCKER_IMAGE_TAG="celotool-9548656d0411d8c3b3d525727e621442e21b055a"
+
+CELOCLI_STANDALONE_IMAGE_REPOSITORY="gcr.io/celo-testnet/celocli-standalone"
+CELOCLI_STANDALONE_IMAGE_TAG="0.0.30-beta2"
+
+# Schedule for an oracle deployed via celotool, expressed in crontab syntax
+# This schedule is "every 5th minute"
+ORACLE_CRON_SCHEDULE="*/5 * * * *"
```
asaj, line 49: We won't have an oracle for RC0, is there an option to deploy without oracles? cc @yerdua @tkporter

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+TRANSACTION_METRICS_EXPORTER_DOCKER_IMAGE_TAG="transaction-metrics-exporter-9548656d0411d8c3b3d525727e621442e21b055a"
+
+ATTESTATION_SERVICE_DOCKER_IMAGE_REPOSITORY="us.gcr.io/celo-testnet/celo-monorepo"
+ATTESTATION_SERVICE_DOCKER_IMAGE_TAG="attestation-service-9548656d0411d8c3b3d525727e621442e21b055a"
+
+GETH_EXPORTER_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet-production/geth-exporter"
+GETH_EXPORTER_DOCKER_IMAGE_TAG="ed7d21bd50592709173368cd697ef73c1774a261"
+
+# Genesis Vars
+NETWORK_ID=120320
```
asaj: I thought we were doing YYYYMMDD? Could be wrong

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+CONSENSUS_TYPE="istanbul"
+BLOCK_TIME=5
+# Minimum epoch length is 15 minutes
```
asaj: Let's make this 24 hours

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+EPOCH=720
+LOOKBACK=12
+ISTANBUL_REQUEST_TIMEOUT_MS=10000
+
+# the number of load test clients that will be given funds in the genesis & migrations
```
asaj, line 72: We're not funding these, are we? Probably want 0

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+LOAD_TEST_CLIENTS=20
+# every 7.2 seconds, so that 500 transactions are sent by a client every hour
+# to simulate 10,000 total transactions per hour
+LOAD_TEST_TX_DELAY_MS=7200
+
+# the amount in cUSD wei to give faucet, load test, and attestation bot accounts
+FAUCET_CUSD_WEI=60000000000000000000000
```
asaj, line 79: This also should be unused, right?

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+
+VALIDATORS=70
+PROXIED_VALIDATORS=35
+ELECTION_MIN_VALIDATORS=22
```
asaj: Hmm, where is this used? Do we override the migrationconfig? Suggest 50 for this, though I guess we don't really have a plan for how RC0 validators will get elected

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+LEADERBOARD_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/celo-monorepo"
+LEADERBOARD_DOCKER_IMAGE_TAG="leaderboard-e84c5086e94193fdef430008bd44129e4151bb6b"
+
+# Attestation Bot variables
+ATTESTATION_BOT_INITIAL_WAIT_SECONDS=600
+ATTESTATION_BOT_IN_BETWEEN_WAIT_SECONDS=600
+ATTESTATION_BOT_MAX_ATTESTATIONS=90
+
+# Number of gold-holding bots that vote for validator groups
+VOTING_BOTS=100
```
asaj: Probably want 0 of these

packages/celotool/src/lib/generate_utils.ts
```diff
@@ -213,7 +215,8 @@ export const generateGenesisFromEnv = (enablePetersburg: boolean = true) => {
   const lookbackwindow = parseInt(fetchEnvOrFallback(envVar.LOOKBACK, '12'), 10)
   const chainId = parseInt(fetchEnv(envVar.NETWORK_ID), 10)
 
-  const initialAccounts = getFaucetedAccounts(mnemonic)
+  const initialAccounts = INITIAL_ACCOUNTS
```
asaj: Is there a way we can keep this tool more general? As is, it would break deployments for all networks other than RC0

asaj (commented):
Can you include genesis block generated by these changes? I would guess that @tkporter 's tool to deploy a testnet with a custom genesis block will be what we want to use, and we probably need to point it to a file

.env.rc0
lucasege: Ah yes I did this but I guess I was on a branch locally, my bad

.env.rc0
lucasege: Didn't know there was a convention, I'll follow that

.env.rc0
lucasege: I believe it's unused unless `migrationOverride` is used in a sneaky way. Should I change this to be safe?

.env.rc0
lucasege: Again only used in `migrationOverride`

.env.rc0
lucasege: Quick search yields some kinda hacky ways to basically set a cron job that will never execute (i.e. set it to run on feb 31). Let me know if I should do this for oracle and voting_bot
https://stackoverflow.com/questions/8324306/cron-job-that-will-never-execute

packages/celotool/src/lib/generate_utils.ts
lucasege: Please review the new changes

packages/celotool/genesis_rc0.json
```diff
@@ -0,0 +1,297 @@
...
+  "difficulty": "0x1",
+  "coinbase": "0x0000000000000000000000000000000000000000",
+  "alloc": {
+    "04acedA1F7909187796D6d59F842ED4aEEa3E4b6": {
```
asaj: There are a bunch of balances here that shouldn't be there, these look like the ones that celotool adds automatically for testnets

asaj (commented):
One easy solution would be to only check in the genesis block, .env file, and encrypted mnemonic, and keep your changes on a separate branch. Would keep you from having to fix the tests now when it's not really relevant to RC0 deployment.

Wdyt?

packages/celotool/genesis_rc0.json
lucasege: These look like the 70 validators generated from the mnemonic and their balances are defined here 
https://github.com/celo-org/celo-monorepo/blob/0ea99f3a94dea0eb9bf9dd89c300cb4390006ee3/packages/celotool/src/lib/generate_utils.ts#L109

Should I change val0 to be the same as the rest?

lucasege:
Added a new branch at your suggestion, `lucasege/rc0` with the incompatible changes

.env.rc0
tkporter: The oracle is its own helm chart that is deployed using a different command, so we can just not deploy it and be totally fine. So we can remove this line or keep it, doesn't matter

.env.rc0
```diff
@@ -0,0 +1,150 @@
+# Don't use "//" for comments in this file.
+# This file is meant to be executed as a bash script for testing.
+ENV_TYPE="production"
+
+GETH_VERBOSITY=2
+
+VM_BASED=true
```
tkporter: Given that we are running this internally, I think it'd be cheaper ($$ wise) for us to not have this be VM_BASED, wdyt?

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+# the number of load test clients that will be given funds in the genesis & migrations
+LOAD_TEST_CLIENTS=0
+# every 7.2 seconds, so that 500 transactions are sent by a client every hour
+# to simulate 10,000 total transactions per hour
+LOAD_TEST_TX_DELAY_MS=7200
+
+# the amount in cUSD wei to give faucet, load test, and attestation bot accounts
+FAUCET_CUSD_WEI=60000000000000000000000
+
+VALIDATORS=70
```
tkporter: Given that the goal of this testnet is to test processes, do we need this to be so high?

Or if we want to model mainnet, should this be higher?

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+PROXIED_VALIDATORS=35
+ELECTION_MIN_VALIDATORS=50
+TX_NODES=5
```
tkporter: could probably lower this to like 2 or 3 just to save money

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+# STATIC_IPS_FOR_GETH_NODES=false
+# Whether tx_nodes/validators stateful set should use ssd persistent disks
+# GETH_NODES_SSD_DISKS=true
+IN_MEMORY_DISCOVERY_TABLE=false
+# PING_IP_FROM_PACKET=false
```
tkporter: if we make this not `VM_BASED`, make sure to uncomment this and make it true

packages/celotool/genesis_rc0.json
```diff
@@ -0,0 +1,87 @@
...
+  "difficulty": "0x1",
+  "coinbase": "0x0000000000000000000000000000000000000000",
+  "alloc": {
```
tkporter, line 26: I think we normally give validators some gold here otherwise the migrations might fail

.env.rc0
```diff
@@ -0,0 +1,150 @@
...
+LOAD_TEST_TX_DELAY_MS=7200
+
+# the amount in cUSD wei to give faucet, load test, and attestation bot accounts
+FAUCET_CUSD_WEI=60000000000000000000000
+
+VALIDATORS=70
+PROXIED_VALIDATORS=35
+ELECTION_MIN_VALIDATORS=50
+TX_NODES=5
+# STATIC_IPS_FOR_GETH_NODES=false
```
tkporter: If we make this not `VM_BASED`, this and the `GETH_NODES_SSD_DISKS` probably needs to be uncommented

packages/celotool/genesis_rc0.json
asaj: We won't be running the migration that registers validators in order to better mimic a mainnet deployment.

.env.rc0
asaj: Let's put it at 50

.env.rc0
asaj: Agreed this is probably not necessary

packages/celotool/genesis_rc0.json
asaj: None of them should have any balance

.env.rc0
asaj: But given genesis block LGTM, maybe we should leave at 70

asaj (approved):
Nice, LGTM. Consider adopting the suggestion from @tkporter of not making this VM based

PR #3099, opened 2020-03-12
[Wallet] Use async support in react-native-tcp for connectIPC on Android

etuleu:
### Description

The goal of this PR is to make use of the version of react-native-tcp that has support for async networking to communicate with the unix socket exposed by geth.

### Tested

yarn reset && yarn && cd packages/mobile && yarn dev:android
Made sure the app runs and using adb logcat looked for logs coming from the new version of react-native-tcp.

### Other changes

None.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


packages/mobile/package.json
```diff
@@ -115,7 +115,7 @@
     "react-native-splash-screen": "^3.2.0",
     "react-native-svg": "^9.11.1",
     "react-native-swiper": "^1.5.14",
-    "react-native-tcp": "https://github.com/cmcewen/react-native-tcp#65a1e45",
+    "react-native-tcp": "https://github.com/cmcewen/react-native-tcp#41caaa4",
```
jmrossy: Does the change correspond to this PR: https://github.com/celo-org/react-native-tcp/pull/5
If so, please get that merged to master on the RN-TCP repo before you use it here

jmrossy: if that PR is ready for review I'll have a look :)

etuleu: I believe @cmcewen approved that PR but it can't hurt to get an extra look. Also I am online and available for a call/chat if you have any questions. Thanks!

https://github.com/celo-org/react-native-tcp/pull/5

jeanregisser (approved):
Awesome 🚀👍

PR #3106, opened 2020-03-13
[web] Fetch more than just first 100 asset (Icons)

### LINKED ISSUE #3042: Load ALL icons
Purpose: Current celo.org/experience/brand/icons will only load 100 icons. We are quickly reaching that threshold and would like to change it to an infinite load.

Web Address: celo.org/experience/brand/icons

aaronmgdr:
### Description

original implementation quickly used the `.all` method which will only return the first 100 rows in an air-table, now page thru to actually get all possible rows. 

### Other changes

created a byMethod middleware to avoid having to write the error handling code on each route handler

### Related issues

- Fixes #3042 

### Backwards compatibility
no as routes  changed


packages/web/pages/api/experience/assets/[asset].ts
```diff
@@ -0,0 +1,25 @@
+import { NextApiRequest, NextApiResponse } from 'next'
```
Comment (jmrossy) [inline packages/web/pages/api/experience/assets/[asset].ts:1]: why does this filename has brackets in it? Is that a Next.JS thing?

Comment (aaronmgdr) [reply packages/web/pages/api/experience/assets/[asset].ts:1]: yeah its a way of doing what in express would be `/assets/:asset` essentially a url param

aaronmgdr:
note my first thought here was to be able to pass an page number from clientside. but airtable doesnt do explicit pages, just in memory paging.

jmrossy (approved):

PR #3108, opened 2020-03-13
fix slashing e2e tests

aslawson:
### Description

fix double signing slashing tests -- changed config slashing penalty from 5k to 9k

### Tested

NA

### Other changes

NA

### Related issues

follow up to https://github.com/celo-org/celo-monorepo/pull/2972 change

### Backwards compatibility

NA


packages/celotool/src/e2e-tests/slashing_tests.ts
```diff
@@ -314,7 +314,7 @@ describe('slashing tests', function(this: any) {
 
       // Penalty is defined to be 5000 cGLD in migrations, locked gold is 10000 cGLD for a validator
       const balance = await lockedGold.getAccountTotalLockedGold(signer)
-      assert.equal(balance.toString(10), '5000000000000000000000')
+      assert.equal(balance.toString(10), '1000000000000000000000')
```
lucasege: Nit: change comment to 9k penalty, and add ~ "so after slashing locked gold is 1000cGld"

packages/celotool/src/e2e-tests/slashing_tests.ts
```diff
@@ -347,7 +347,7 @@ describe('slashing tests', function(this: any) {
       // Penalty is defined to be 5000 cGLD in migrations, locked gold is 10000 cGLD for a validator
       const lockedGold = await kit.contracts.getLockedGold()
       const balance = await lockedGold.getAccountTotalLockedGold(signer)
-      assert.equal(balance.toString(10), '5000000000000000000000')
+      assert.equal(balance.toString(10), '1000000000000000000000')
```
lucasege: Same as above

lucasege (approved):
lgtm, 1 nit on a comment change

packages/celotool/src/e2e-tests/slashing_tests.ts
aslawson: good catch, thanks!

PR #3109, opened 2020-03-13
Fix web build by changing literal to string

cmcewen:
### Description

Fix web build - TS doesn't like string literals in enums apparently

### Tested

Build works

### DIFF
--- a/packages/web/src/styles.ts
+++ b/packages/web/src/styles.ts
@@ -9,7 +9,7 @@ export enum colors {
   placeholderDarkMode = '#838486',
   secondary = '#81868B',
   dark = '#2E3338',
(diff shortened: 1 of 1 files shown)


packages/web/src/styles.ts
```diff
@@ -9,7 +9,7 @@ export enum colors {
   placeholderDarkMode = '#838486',
   secondary = '#81868B',
   dark = '#2E3338',
-  darkTransparent = `rgba(46, 51, 56, 0.5)`,
+  darkTransparent = 'rgba(46, 51, 56, 0.5)',
```
aaronmgdr, line 12: oh

aaronmgdr (approved):

Issue #2787, opened 2020-02-14
Users SBAT add a claimed URL in their metadata 


clrblmt:
### Expected Behavior

**Make the `DOMAIN` claim verifiable**:

As part of the creation of the claim, the CLI should indicate instructions to add the following TXT record to their DNS settings:

`celo-site-verification=${base64encode(signatureOfClaimWithAddress)}`

Add a function similar to the [keybase verify function](https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/identity/claims/keybase.ts#L23), that will check for the signature on the TXT record by looking it up via the [`dns` package](https://nodejs.org/api/dns.html#dns_dns_resolvetxt_hostname_callback)

### Current Behavior

We have a `DOMAIN` claim that is not verifiable.

nambrot:
A modification that is worthwhile to think about is to allow the payload to specify an optional `URL` field that signifies a path to the page that should be displayed for users. I.e. I own `validator.com`, but I want users to redirect users to `http://validator.com/celo`. We would have to be careful however to ensure that people can't pretend to be someone else, i.e. `validator.com/alotofgarbage?redirect-to:www.google.com`. I believe this means that we have to very explicitly display the domain @cla-bel

aaitor:
**Requesting the domain claim:**
```
aitor@tijuana:~/Projects/Celo/Upstream/celo-monorepo/packages/cli (aaitor/cli-domain-claim)$ bin/run account:claim-domain ~/Projects/conf/Celo/metadata-orco.json --domain orco.dev --from  0x088e43db20b5606aa5a87af922ebc785599cbb49
Read Metadata from /home/aitor/Projects/conf/Celo/metadata-orco.json... done
Add claim... done
Write Metadata to /home/aitor/Projects/conf/Celo/metadata-orco.json... done
Please add the following TXT record to your domain:
celo-site-verification=MHgxZDE3ZTgyZjM2ZWI4MjFmZTJjMGJkYWM4ODA4MTRiYWQ5ZmYzZTViN2IxODNjMDNhZDRiOTIyYzhkZGRhZTRmMGNjYWMxYTAxMjFlYTEzNjg0MmZmZmNiNmE4ZjA0ZjAxODFlNWFjNjI0ZGRkMzgxZTIwMzE4NzI4ZmEyZjFiOTFi
```
At this point is necessary to add the `celo-site-verification` TXT record to the domain.
 
**Verifying the domain claim:**
```
[...]

PR #3113, opened 2020-03-16
Metadata crawler

mrsmkl:
### Description

Simple crawler for metadata.
TODO: something more intelligent to determine if metadata has to be verified.

### Related issues

- Related to #2787


mrsmkl:
I guess this was replaced by some other PR.

PR #3123, opened 2020-03-17
[L01] Duplicated onlyWhenFrozen modifier

mrsmkl:
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/394


packages/protocol/contracts/common/GoldToken.sol
```diff
@@ -197,7 +197,7 @@ contract GoldToken is Initializable, CalledByVm, Freezable, IERC20, ICeloToken {
    * @param value The amount of Celo Gold to transfer.
    * @return True if the transaction succeeds.
    */
-  function _transfer(address to, uint256 value) internal onlyWhenNotFrozen returns (bool) {
+  function _transfer(address to, uint256 value) internal returns (bool) {
```
m-chrzan: Nit: I think it might be cleaner to leave the modifier here, but remove it from `_transfer`'s callers.

asaj (approved):

PR #3125, opened 2020-03-17
[L04] Lack of indexed parameters in events

mrsmkl:
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/397


aslawson (commented):
I'm not super familiar on event indexing and whether numbers need indexing or not.

Also, is the issue suppose to be extensive AKA all event parameters should be index?  Just in case we use them?  Is there a cost to indexing params that we know we'll never need to sort/search by?

I do see some additional addresses as event params in Account.sol and some in the /identity and /stability packages as well that are unindexed

asaj (approved):

mrsmkl:
Indexed arg takes a bit more gas than data, 119 more for bytes32 and 215 more for address.
There can be only 3 indexed args for each event, the event name takes one of the four topic slots.
It's not too useful to set them to be indexed, something like blockscout is better for that purpose.

PR #3127, opened 2020-03-17
[L08] There are two different whitelist getters

mrsmkl:
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/401


alecps (approved):
LGTM. The only thing having this public gets us is the convenience of querying just the list of hardcoded addresses in the whitelist. Probably not worth the potential confusion.

asaj (approved):

PR #3128, opened 2020-03-17
[L11] Removed account signers cannot be reauthorized

mrsmkl:
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/404


packages/protocol/contracts/common/Accounts.sol
```diff
@@ -275,7 +275,8 @@ contract Accounts is IAccounts, Ownable, ReentrancyGuard, Initializable, UsingRe
   }
 
   /**
-   * @notice Removes the currently authorized vote signer for the account
+   * @notice Removes the currently authorized vote signer for the account.
+   * Note that the signers cannot be reauthorized after they have been removed.
```
aslawson, line 279: is this by design?  we don't want signers to be able to reauthorize?

aslawson (approved):

asaj (approved):

packages/protocol/contracts/common/Accounts.sol
mrsmkl: According to @asaj's comment in the issue, it's by design.

PR #3129, opened 2020-03-17
Small adjustments to reflect RC0 deployment

asaj:
### Description
Changes to .env file, genesis block, automatically whitelist contract deployer if cGLD price set in migrations config, don't send as many txs in parallel for governance constitution.
### Tested

Tested as part of RC0 deployment
### Other changes

None

### Related issues

None
### Backwards compatibility

None


packages/protocol/migrations/24_governance.ts
```diff
@@ -95,10 +92,8 @@ module.exports = deploymentForCoreContract<GovernanceInstance>(
       'Validators',
     ]
 
-    await Promise.all(
```
aslawson, line 98: does removing this make the tx happen sequentially instead of in parallel? is this related to the ledger flakiness?

packages/protocol/migrations/09_stabletoken.ts
```diff
@@ -60,10 +60,12 @@ module.exports = deploymentForCoreContract<StableTokenInstance>(
...
       const isOracle = config.stableToken.oracles.some((o) => eqAddress(o, fromAddress))
-      console.assert(
-        isOracle,
-        `Gold price specified in migration but ${fromAddress} not authorized as oracle`
-      )
+      if (!isOracle) {
+        console.warning(
+          `Gold price specified in migration but ${fromAddress} not explicitly authorized as oracle, authorizing...`
+        )
+        await sortedOracles.addOracle(stableToken.address, ensureLeading0x(fromAddress))
```
aslawson, line 67: are you doing this dynamically just so the contract deployer address doesn't have to be duplicated in both the the truffle config and migration config?

aslawson (approved):
LGTM -- though seems to not be building or installing dependencies...

packages/protocol/migrations/09_stabletoken.ts
asaj: Yeah I saw the migrations fail b/c the "from" address wasn't explicitly authorized, which is a pain. It's easier to whitelist the oracle now and remove it later via governance.

packages/protocol/migrations/24_governance.ts
asaj: Correct, now it will be in series, which while slower seems to be more reliable

PR #3130, opened 2020-03-17
[celocli] add reserve spender cli tooling, update governance approver tooling to be usable with and without multisig

aslawson:
### Description

celocli updates for multisig
- updated governance approver to have optional multisig
- added reserve spender tooling with optional multisig

### Tested

no running network with the multisig migrations to manually test this, will need to do so later

### Other changes

NA

### Related issues

- Follow up to https://github.com/celo-org/celo-monorepo/pull/3023

### Backwards compatibility

Improves tooling compatibility with migration changes


packages/cli/src/commands/governance/approve.ts
```diff
@@ -11,37 +11,47 @@ export default class Approve extends BaseCommand {
     ...BaseCommand.flags,
     proposalID: flags.string({ required: true, description: 'UUID of proposal to approve' }),
     from: Flags.address({ required: true, description: "Approver's address" }),
+    useMultiSig: flags.boolean({
```
asaj, line 14: The "from" flag is unnecessary if "useMultiSig" is not set, right?

packages/cli/src/commands/governance/approve.ts
```diff
@@ -11,37 +11,47 @@ export default class Approve extends BaseCommand {
...
     const res = this.parse(Approve)
     const account = res.flags.from
+    const useMultiSig = res.flags.useMultiSig
     const id = res.flags.proposalID
     this.kit.defaultAccount = account
     const governance = await this.kit.contracts.getGovernance()
-    const governanceApproverMultiSig = await this.kit.contracts.getMultiSig(
-      await governance.getApprover()
-    )
+    const multiSigAddress = await governance.getApprover()
```
asaj: It seems weird to me to do this if `useMultiSig` is false

packages/cli/src/commands/reserve/transfergold.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Flags } from '../../utils/command'
+
+export default class TransferGold extends BaseCommand {
+  static description = 'Transfers reserve gold to other reserve address'
+
+  static flags = {
+    ...BaseCommand.flags,
+    value: flags.string({ required: true, description: 'The unit amount of Celo Gold (cGLD)' }),
+    to: Flags.address({ required: true, description: 'Receiving address' }),
+    from: Flags.address({ required: true, description: "Sender's address" }),
```
asaj: This is the "spender's" address, yeah?

packages/cli/src/commands/reserve/transfergold.ts
```diff
@@ -0,0 +1,55 @@
...
+  async run() {
+    const res = this.parse(TransferGold)
+    const value = res.flags.value
+    const to = res.flags.to
+    const account = res.flags.from
+    const useMultiSig = res.flags.useMultiSig
+    this.kit.defaultAccount = account
+    const reserve = await this.kit.contracts.getReserve()
+    const spenders = await reserve.getSpenders()
+    // assumes that the multisig is the most recent spender in the spenders array
```
asaj, line 34: ditto, this reads weird to me when it's not guaranteed that we're using a multisig at all

asaj (commented):
> no running network with the multisig migrations to manually test this, will need to do so later

You can add unit tests for this in the CLI package

Also please update the PR title as it appears to be a truncated commit message

packages/cli/src/commands/governance/approve.ts
asaj, line 14: Please add an example for this below

aslawson: i thought the point was you wanted users to use this without multisig.  in which case there is still a an address the request is coming from (which may be the multisig address or non-smart contract account). there would always be a from, but useMultiSig would indicate that the caller wants the request they are making to go through the multisig

packages/cli/src/commands/governance/approve.ts
aslawson: ya it was a a tradeoff for overcomplexity, readability and avoiding duplicate logic.  i didn't want to be throwing if(useMultiSig) tertiaries on every statement.  and since this address is reused, it would be less efficient to re-call `await governance.getApprover` every time the address is needed.

i could obviously have a giant if/else block for the multisig vs non-multisig use cases, but then there is a lot of duplicate logic that is difficult to parse out into a separate method (eg. inserting an additional check in the checkBuilder).

if i don't explicitly instantiate the `governanceApproverMultiSig` when `useMultiSig` is false, then it requires in later logic needing to check if this variables is undefined prior to every use.

open to your further thoughts if you feel strongly about this.

packages/cli/src/commands/reserve/transfergold.ts
aslawson: ya...well technically they are also "sending" the transaction, which is why I keep messing this up.  I'll update

packages/cli/src/commands/reserve/transfergold.ts
aslawson: same

packages/cli/src/commands/governance/approve.ts
aslawson: sure thing

aslawson:
> so

I didn't see many examples of unit test in cli, but I'll take a look.

aslawson:
> > no running network with the multisig migrations to manually test this, will need to do so later
> 
> You can add unit tests for this in the CLI package
> 
> Also please update the PR title as it appears to be a truncated commit message

I didn't see many examples of unit test in cli, but I'll take a look.

packages/cli/src/commands/governance/approve.ts
yorhodes: I think a general philosophy for contractkit is deferring as much as possible to JS because of the latency/load on the blockchain client (esp if the node is remote)

packages/cli/src/commands/governance/approve.ts
yorhodes: oclif (the CLI framework) has some options for making flag dependencies and mutexes
https://oclif.io/docs/flags

packages/cli/src/commands/governance/approve.ts
```diff
@@ -11,37 +11,50 @@ export default class Approve extends BaseCommand {
...
-    const tx = await governance.approve(id)
+    const governanceTx = await governance.approve(id)
     const multiSigTx = await governanceApproverMultiSig.submitOrConfirmTransaction(
       governance.address,
-      tx.txo
+      governanceTx.txo
     )
-    await displaySendTx<any>('approveTx', multiSigTx, {}, 'ProposalApproved')
+    const tx = useMultiSig ? multiSigTx : governanceTx
+    await displaySendTx<any>('approveTx', tx, {}, 'ProposalApproved')
```
yorhodes: is the `any` annotation necessary here?

aslawson: yes, unfortunately it won't compile without it since the return from the 
multisig.submitOrConfirmTransaction() could be either a string or void, and the displaySendTx<A> requires that A is specified then (otherwise it just assumes its one of those types).  

although it looks like I can make it more specific by just doing displaySendTx<string | void> which seems better

packages/cli/src/commands/governance/approve.ts
aslawson: i don't want these to be mutually exclusive.  there will always be a from address needed.  we'll need to confirm the caller is a multisig signatory to perform the action

packages/cli/src/commands/governance/approve.ts
aslawson: ok -- i've updated to conditionally make the blockchain client request if multisig address is needed to be retrieved

packages/cli/src/commands/governance/approve.ts
asaj: Yeah I think my point was just that if they're not calling it via a multisig we can assume they're the approver and use that as the "from". Don't feel strongly

packages/cli/src/commands/governance/approve.ts
```diff
@@ -11,37 +11,54 @@ export default class Approve extends BaseCommand {
...
     // in case target is queued
     if (await governance.isQueued(id)) {
       await governance.dequeueProposalsIfReady().sendAndWaitForReceipt()
     }
 
     await newCheckBuilder(this)
-      .isApprover(account)
+      .isApprover(approver)
+      .addConditionalCheck(`${account} is multisig signatory`, useMultiSig, async () =>
+        governanceApproverMultiSig !== undefined
```
asaj, line 45: Could be more readable to check `useMultiSig` here

packages/cli/src/commands/governance/approve.ts
```diff
@@ -11,37 +11,54 @@ export default class Approve extends BaseCommand {
...
 
-    const tx = await governance.approve(id)
-    const multiSigTx = await governanceApproverMultiSig.submitOrConfirmTransaction(
-      governance.address,
-      tx.txo
-    )
-    await displaySendTx<any>('approveTx', multiSigTx, {}, 'ProposalApproved')
+    const governanceTx = await governance.approve(id)
+    const tx =
+      governanceApproverMultiSig === undefined
```
asaj, line 56: Ditto

packages/cli/src/commands/reserve/transfergold.test.ts
```diff
@@ -0,0 +1,33 @@
...
+        '--from',
+        accounts[0],
+        '--value',
+        '1',
+        '--to',
+        '0x91c987bf62D25945dB517BDAa840A6c661374402',
+      ])
+    ).rejects.toThrow("Some checks didn't pass!")
+  })
+  test('can transferGold with multisig option', async () => {
```
asaj, line 34: Can we confirm that gold actually gets transferred here? i.e. check the balance of `0x91c987bf62D25945dB517BDAa840A6c661374402` before and after?

packages/cli/src/commands/reserve/transfergold.ts
```diff
@@ -0,0 +1,57 @@
...
+    const multiSigAddress = spenders.length > 0 ? spenders[spenders.length - 1] : ''
+    const reserveSpenderMultiSig = useMultiSig
+      ? await this.kit.contracts.getMultiSig(multiSigAddress)
+      : undefined
+    const spender = useMultiSig ? multiSigAddress : account
+
+    await newCheckBuilder(this)
+      .addCheck(`${spender} is a reserve spender`, async () => reserve.isSpender(spender))
+      .addConditionalCheck(`${account} is a multisig signatory`, useMultiSig, async () =>
+        reserveSpenderMultiSig !== undefined
```
asaj, line 44: Ditto

packages/cli/src/commands/reserve/transfergold.ts
```diff
@@ -0,0 +1,57 @@
...
+          ? reserveSpenderMultiSig.isowner(account)
+          : new Promise<boolean>(() => false)
+      )
+      .runChecks()
+
+    const reserveTx = await reserve.transferGold(to, value)
+    const tx =
+      reserveSpenderMultiSig === undefined
```
asaj, line 53: ditto

packages/cli/src/commands/governance/approve.ts
```diff
@@ -11,37 +11,54 @@ export default class Approve extends BaseCommand {
     ...BaseCommand.flags,
     proposalID: flags.string({ required: true, description: 'UUID of proposal to approve' }),
     from: Flags.address({ required: true, description: "Approver's address" }),
+    useMultiSig: flags.boolean({
```
asaj, line 14: Can you add a unit test?

packages/cli/src/commands/governance/approve.ts
aslawson: the useMultiSig is included as the condition to run.

compiler forces me to added a check for governanceApproverMultiSig is undefined before using it

packages/cli/src/commands/governance/approve.ts
aslawson: same

packages/cli/src/commands/governance/approve.ts
aslawson: I did try -- its complicated because a proposal need to exist to be approve. To use the cli tooling to create a proposal it requires a json file with the proposal and url, so a bit complicated.  But it seems i can just use the contractkit directly to create one, so I will try that

packages/cli/src/commands/reserve/transfergold.test.ts
```diff
@@ -1,33 +1,47 @@
...
       accounts[0],
       '--value',
-      '1',
+      transferAmt.toString(10),
       '--to',
-      '0x91c987bf62D25945dB517BDAa840A6c661374402',
+      accounts[9],
       '--useMultiSig',
     ])
+    expect(await goldToken.balanceOf(accounts[9])).toEqual(initialBalance)
```
asaj: Wouldn't we expect this balance to go up by `transferAmt`?

packages/cli/src/commands/reserve/transfergold.test.ts
```diff
@@ -1,33 +1,47 @@
...
+    const initialBalance = await goldToken.balanceOf(accounts[9])
     await TransferGold.run([
       '--from',
       accounts[0],
       '--value',
-      '1',
+      transferAmt.toString(10),
       '--to',
-      '0x91c987bf62D25945dB517BDAa840A6c661374402',
+      accounts[9],
```
asaj, line 42: Can you transfer to this if it isn't an "otherReserveAddress"?

aslawson: it is the otherReserveAddress

packages/cli/src/commands/reserve/transfergold.test.ts
aslawson: ya...i thought that was what I had here...perhaps i removed the addition when debugging and then it passed...which is a bad sign

asaj (approved):

PR #3140, opened 2020-03-18
[Wallet] iOS multi env support + fastlane deployments

### LINKED ISSUE #1156: iOS app should use appropriate bundle identifiers for each supported env
### Expected Behavior

It should use appropriate bundle identifiers for each supported env.
Probably using an Xcode scheme for each of them. See https://github.com/luggit/react-native-config#ios-1

### Current Behavior

It uses the same bundle identifier not matter what env is active.

jeanregisser:
### Description

The basic idea is to have one scheme per environment file, so we can easily alternate between them, and have iOS specific config in the different env files so they remain the single source of thruth instead of being buried in Xcode itself.

This is similar to what has been done for Android.

Screenshot 2020-03-18 at 18 02 58

Here's the mapping of the schemes:
- `celo-dev` uses `.env`
- `celo-integration` uses `.env.integration`
- `celo-alfajores` uses `.env.alfajores`
- `celo-pilot` uses `.env.pilot`

These can be used from Xcode or the command line:
`yarn react-native run-ios --scheme "celo-dev"`

From day to day, devs would usually use the `celo-dev` scheme.

I also added new iOS specific lanes to easily build and deploy to TestFlight.

No more manual changes to the bundle identifier and google service plist needed 😉 

### Tested

Successfully built and deployed pilot app using `bundle exec fastlane ios pilotapp`

Screenshot 2020-03-18 at 17 03 35

### Related issues

- Fixes #1156

### Backwards compatibility

Yes.


jmrossy (approved):
All lgtm 👍

PR #3143, opened 2020-03-18
[L13] Comments still refer to removed functions

aslawson:
### Description

- removed comment references to maxFixedMul() and maxFixedDividend() methods which were removed.  
- defined value of maxFixedMul and maxFixedDividend in comments that explained testing performed on the multiply() and divide() methods so reference to them was clear.

### Tested

NA -- only affects comments

### Other changes

NA

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/406

### Backwards compatibility

NA


packages/protocol/contracts/common/FixidityLib.sol
```diff
@@ -230,9 +226,9 @@ library FixidityLib {
   }
 
   /**
-   * @notice x/y. If the dividend is higher than the max fixed dividend, it
+   * @notice x/y. If the dividend is higher than the max dividend value, it
    * might overflow. You can use multiply(x,reciprocal(y)) instead.
-   * @dev The maximum value that can be safely used as a dividend is defined as
+   * @dev The maximum value that can be safely used as a dividend (maxFixedDividend) is defined as
```
mrsmkl: Looks like `maxFixedDividend` is the same as `maxNewFixed`, perhaps this can be used to simplify the comments.

packages/protocol/contracts/common/FixidityLib.sol
```diff
@@ -163,10 +158,11 @@ library FixidityLib {
   }
 
   /**
-   * @notice x*y. If any of the operators is higher than maxFixedMul() it
+   * @notice x*y. If any of the operators is higher than the max multiplier value it
    * might overflow.
-   * @dev The maximum value that can be safely used as a multiplication operator is calculated is
-   * sqrt(maxUint256()*fixed1()), or 340282366920938463463374607431768211455999999999999
+   * @dev The maximum value that can be safely used as a multiplication operator
+   * (maxFixedMul) is calculated is sqrt(maxUint256()*fixed1()),
```
m-chrzan: Typo: is calculated *as*

m-chrzan (approved):
Just a small typo to fix

packages/protocol/contracts/common/FixidityLib.sol
aslawson: good point -- i'll update to just refer to it as maxNewFixed

packages/protocol/contracts/common/FixidityLib.sol
```diff
@@ -230,17 +226,17 @@ library FixidityLib {
   }
 
   /**
-   * @notice x/y. If the dividend is higher than the max fixed dividend, it
+   * @notice x/y. If the dividend is higher than the max dividend value, it
    * might overflow. You can use multiply(x,reciprocal(y)) instead.
-   * @dev The maximum value that can be safely used as a dividend is defined as
-   * divide(maxFixedDividend,newFixedFraction(1,fixed1())) is around maxUint256().
+   * @dev The maximum value that can be safely used as a dividend (maxNewFixed) is defined as
```
asaj, line 231: Why can you substitute `maxNewFixed` for `maxFixedDividend` everywhere here?

aslawson: they are calculated as the same value (maxUint256()/fixed1()). @mrsmkl suggested simplifying above ^^

PR #3144, opened 2020-03-18
[L12] Division by zero possible in Exchange.sol

aslawson:
### Description

Add a require verification that exchange rate denominator is greater than 0

### Tested

Ran tests

### Other changes

NA

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/405

### Backwards compatibility

NA


packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -342,6 +342,7 @@ contract Exchange is IExchange, Initializable, Ownable, UsingRegistry, Reentranc
       registry.getAddressForOrDie(SORTED_ORACLES_REGISTRY_ID)
     )
       .medianRate(stable);
+    require(rateDenominator > 0, "exchange rate denominator cannot be less than or equal to 0");
```
asaj: `exchange rate denominator cannot be 0` - no such thing as negative unsigned integers

asaj (approved):

packages/protocol/contracts/stability/Exchange.sol
aslawson: good point.  any reason to prefer `!==` over `>`?

aslawson: updated the error message

PR #3148, opened 2020-03-19
Switch wallet to use ContractKit for all txs

### LINKED ISSUE #475: Migrate Mobile App to the new contractkit
### Expected Behavior

Use ContractKit
- [x] View functions (exchange rate, fetching balances, contract decimals)
- [ ] State changing functions

### Current Behavior

Uses WalletKit

cmcewen:
### Description

This is the second attempt on #2364. It had forno issues, which are resolved in the same session, but it is still broken for new app sessions (currently reproducible in the Alfajores app)

### Tested

Verified from fresh install running both the light client and Forno

### Related issues

- Fixes #475

### Backwards compatibility

Don't expect any issues


packages/mobile/android/app/src/debug/res/xml/network_security_config.xml
```diff
@@ -7,4 +7,10 @@
     <domain includeSubdomains="true">10.0.2.2</domain>
     <domain includeSubdomains="true">10.0.3.2</domain>
   </domain-config>
+  <debug-overrides>
```
jmrossy, line 10: What was this for?

packages/mobile/src/fees/saga.ts
```diff
@@ -20,9 +20,10 @@ const TAG = 'fees/saga'
 // Prevents us from having to recreate txs and estimate their gas each time
 const feeGasCache = new Map<FeeType, BigNumber>()
 // Just use default values here since it doesn't matter for fee estimation
+
 const placeholderSendTx: BasicTokenTransfer = {
   recipientAddress: CeloDefaultRecipient.address,
-  amount: web3.utils.fromWei('1'),
+  amount: getContractKit().web3.utils.fromWei('1'),
```
jmrossy, line 26: a getWeb3() helper might be a good idea, seems like we drill down through CK in many places

packages/mobile/src/transactions/contract-utils.ts
```diff
@@ -0,0 +1,228 @@
+import { values } from 'lodash'
```
jmrossy, line 1: Can this be renamed to `send-utils`? I realize the name in walletKit was contract-utils but that's not really accurate

jmrossy (approved):
Looks great! a few optional comments below

packages/mobile/android/app/src/debug/res/xml/network_security_config.xml
cmcewen: Charles proxy settings

packages/mobile/src/transactions/contract-utils.ts
cmcewen: My feeling is that our send plumbing is still pretty ugly and could use some refactoring, which would hopefully remove a few of these files and function indirection. What do you think?

jmrossy: Agreed, could be better. Ideally we'd just rely on contract kit and extend it's capabilities to support any logging we need

PR #3149, opened 2020-03-19
[Wallet] Fix vertical alignment of text in inputs

i1skn:
### Description

Fixes the bug, when text in inputs is not centered. With this fix text will align to top, only when it's multiline.

### Tested

Android emulator

### Backwards compatibility

Yes

### Screenshot

xdSaLx925banyzno2bgx0SiMfjPi9wMt


jeanregisser (approved):
LGTM 👍

PR #3153, opened 2020-03-19
ReleaseGold various fixes and improvements

### LINKED ISSUE #3112: Missing event logging in ReleaseGold
## Missing event logging in ReleaseGold
Severity: Informational
Difficulty: Low
Type: Auditing and Logging

### Description
Several critical operations in the [ReleaseGold](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/ReleaseGold.sol) contract do not trigger events. As a result, it could be difficult to review the correct behavior of the contracts once deployed.

Critical operations that would benefit from triggering events include:
- [withdraw](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/ReleaseGold.sol#L310-L342): this function self-destructs once its balance reaches zero. It could be helpful to emit an event before triggering contract destruction.
- [refundAndFinalize](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/ReleaseGold.sol#L344-L354): this function also self-destructs when called after the release schedule has been revoked. While that revocation is logged with an event, since this is a separate function call it could also be useful to log separately as well.

Users and blockchain monitoring systems will not be able to easily detect suspicious behaviors without events.

### Exploit Scenario
Bob withdraws the last of his tokens from the contract, causing it to self-destruct. However, he miscalculated his remaining balance and believes there are still tokens remaining in the contract. Bob attempts to withdraw them again but the calls fail. He reviews the event log for the contract but does not see anything that would cause his calls to fail.

### Recommendation
Short term, add the missing events for the operations noted above. 

Long term, err on the side of logging events for any state transitions. Events help to monitor the contracts and trace suspicious behavior.

### LINKED ISSUE #3117: Missing SafeMath addition in ReleaseGold
## Missing SafeMath addition in ReleaseGold
Severity: Informational
Difficulty: Low
Type: Arithmetic

### Description
The [ReleaseGold](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/ReleaseGold.sol) contract makes use of the Open Zeppelin SafeMath library throughout to prevent integer overflow. However, there is one instance of non-SafeMath addition in one of the require checks in the [withdraw](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/governance/ReleaseGold.sol#L310-L342) function. This does not have obvious security implications, however it may be worth changing this to a call to SafeMath's add instead just to be consistent with the rest of the contract.

### Recommendation
Short term, add the missing SafeMath call for the function noted above. 

Long term, ensure consistent use of SafeMath for any instances of arithmetic where overflow may occur.

lucasege:
### Description

Fixes one crucial bug in RG where we don't account for pending withdrawal values, leaving the door open for potential early `selfdestruct`s and wonky releases.

Also added various changes/ fixes detailed below.

### Tested

Added unit testing for pending withdrawal accounting.

### Other changes

* Change deploy script to add months as multiples of 30 days rather than calendar months.
* Use safemath everywhere in RG
* Added events for self destruct events in RG

### Related issues

- Fixes #3117 
- Fixes #3112 

### Backwards compatibility

Yes


asaj (approved):

packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -407,11 +410,21 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
...
    * @notice Calculates remaining locked gold balance in the release schedule instance.
+   *         The returned amount also includes pending withdrawals to maintain consistent releases.
    * @return The remaining locked gold of the release schedule instance.
    * @dev The returned amount may vary over time due to locked gold rewards.
    */
   function getRemainingLockedBalance() public view returns (uint256) {
-    return getLockedGold().getAccountTotalLockedGold(address(this));
+    uint256 pendingWithdrawalSum = 0;
+    if (getAccounts().isAccount(address(this))) {
+      (uint256[] memory pendingWithdrawalValues, ) = getLockedGold().getPendingWithdrawals(
```
asaj: Consider defining `LockedGold.getTotalPendingWithdrawals` and using that here instead

packages/protocol/lib/proxy-utils.ts
```diff
@@ -1,14 +1,32 @@
...
 import Web3 from 'web3'
 
-import { ProxyInstance } from 'types'
 
 export async function setAndInitializeImplementation(
   web3: Web3,
   proxy: ProxyInstance,
   implementationAddress: string,
   initializerAbi: any,
+  from: Address,
```
asaj: Consider passing a `txOptions` object instead, here and elsewhere

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -204,10 +208,12 @@ async function handleJSONFile(err, data) {
 module.exports = async (callback: (error?: any) => number) => {
   try {
     argv = require('minimist')(process.argv.slice(5), {
-      string: ['network', 'grants', 'start_gold', 'output_file', 'really'],
+      string: ['network', 'from', 'grants', 'start_gold', 'output_file', 'really'],
     })
+    if (argv.from === '' || argv.from == null) {
```
asaj: Can we remove this? Would prefer this need to be specified explicitly

asaj (approved):

packages/protocol/scripts/truffle/deploy_release_contracts.ts
lucasege: I'll see if I can - this is used to allow for the `devchain` commands in `yarn;reset` to still be able to call thsi script without needing to access `accounts`. I'll see if I can access it there

PR #3156, opened 2020-03-20
Adding contractKit and cli domain claims support

### LINKED ISSUE #2787: Users SBAT add a claimed URL in their metadata

aaitor:
### Description

It includes the following changes:

* Add to contractKit the possibility of verifying domain claims
* Modify the CLI account:claim-domain command printing the TXT record to add to the domain claimed
* Provide a new CLI command identity:test-domain-claim verifying if a domain name has the right TXT record
* Metadata crawler agent to add the verification flag to the Blockscout database

### Tested

tested locally:

**Requesting the domain claim:**
```
aitor@tijuana:~/Projects/Celo/Upstream/celo-monorepo/packages/cli (aaitor/cli-domain-claim)$ bin/run account:claim-domain ~/Projects/conf/Celo/metadata-orco.json --domain orco.dev --from  0x088e43db20b5606aa5a87af922ebc785599cbb49
Read Metadata from /home/aitor/Projects/conf/Celo/metadata-orco.json... done
Add claim... done
Write Metadata to /home/aitor/Projects/conf/Celo/metadata-orco.json... done
Please add the following TXT record to your domain:
celo-site-verification=MHgxZDE3ZTgyZjM2ZWI4MjFmZTJjMGJkYWM4ODA4MTRiYWQ5ZmYzZTViN2IxODNjMDNhZDRiOTIyYzhkZGRhZTRmMGNjYWMxYTAxMjFlYTEzNjg0MmZmZmNiNmE4ZjA0ZjAxODFlNWFjNjI0ZGRkMzgxZTIwMzE4NzI4ZmEyZjFiOTFi
```
At this point is necessary to add the `celo-site-verification` TXT record to the domain.
 
**Verifying the domain claim:**
```
aitor@tijuana:~/Projects/Celo/Upstream/celo-monorepo/packages/cli (aaitor/cli-domain-claim)$ bin/run identity:test-domain-claim  ~/Projects/conf/Celo/metadata-orco.json --domain orco.dev --from  0x088e43db20b5606aa5a87af922ebc785599cbb49
Read Metadata from /home/aitor/Projects/conf/Celo/metadata-orco.json... done
Fetching domain orco.dev TXT records for verification
Reading TXT records for domain orco.dev
✓ TXT Record celo-site-verification correctly found in orco.dev
```

**Testing with an invalid domain**:
```
aitor@tijuana:~/Projects/Celo/Upstream/celo-monorepo/packages/cli (aaitor/cli-domain-claim)$ bin/run identity:test-domain-claim  ~/Projects/conf/Celo/metadata-orco.json --domain orco.com --from  0x088e43db20b5606aa5a87af922ebc785599cbb49   
Read Metadata from /home/aitor/Projects/conf/Celo/metadata-orco.json... done
Fetching domain orco.com TXT records for verification
Reading TXT records for domain orco.com
x Unable to find a valid celo-site-verification TXT record in orco.com domain
```

### Other changes

N/A

### Related issues

- Fixes #2787 

### Backwards compatibility

This change is backward compatible


H34D (approved):
LGTM

It has tests
It has Documentation

packages/cli/src/commands/account/claim-domain.ts
```diff
@@ -21,5 +21,10 @@ export default class ClaimDomain extends ClaimCommand {
     const metadata = this.readMetadata()
     await this.addClaim(metadata, createDomainClaim(res.flags.domain))
     this.writeMetadata(metadata)
+
+    const signature = JSON.parse(metadata.toString()).meta.signature
```
nambrot: I realized that my original write-up of the issue was incorrect. I don't think we should be putting the signature of the whole metadata in the TXT record, but just the hash of the claim itself, that avoids the need to have to actually pull the metadata, and one can verify the claim independently (assuming one verifies that the signature of the metadata proves the integrity of the claim itself)

packages/cli/src/commands/identity/test-domain-claim.ts
```diff
@@ -0,0 +1,49 @@
+import { createDomainClaim } from '@celo/contractkit/lib/identity/claims/claim'
+import { verifyDomainRecord } from '@celo/contractkit/lib/identity/claims/verify'
+import { flags } from '@oclif/command'
+import chalk from 'chalk'
+import { ClaimCommand } from '../../utils/identity'
+
+export default class TestDomainClaim extends ClaimCommand {
```
nambrot: Could we reuse our existing `get-metadata` and `show-metadata` commands?

packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -66,3 +70,80 @@ export const verifyAccountClaim = async (
...
+type dnsResolverFunction = (
+  hostname: string,
+  callback: (err: NodeJS.ErrnoException, addresses: string[][]) => void
+) => void
+
+/**
+ * It verifies if a DNS domain includes in the TXT records an entry with name
+ * `celo-site-verification` and a valid signature in base64
+ */
+export const verifyDomainClaim = async (
```
nambrot: If we move away from the signature of metadata to the hash of the claim itself in the TXT record, you can avoid most of this work with the metadataURL getter

packages/metadata-crawler/src/crawler.ts
```diff
@@ -0,0 +1,114 @@
...
+  try {
+    const query = `INSERT INTO celo_claims (address, type, element, verified, timestamp, inserted_at, updated_at) VALUES 
+        (decode($1, 'hex'), 'domain', $2, $3, now(), now(), now()) 
+        ON CONFLICT (address, type, element) DO 
+        UPDATE SET verified=$3, timestamp=now(), updated_at=now() `
+    const values = [address, domain, verified]
+
+    await client
+      .query(query, values)
+      .catch((error) => console.error(`Database error ${error}\n query: ${query}`))
```
nambrot: Can we use `bunyan` similarly to the `attestation-service` to get structured logs by default? https://github.com/celo-org/celo-monorepo/blob/master/packages/attestation-service/src/logger.ts#L1

packages/metadata-crawler/src/crawler.ts
```diff
@@ -0,0 +1,114 @@
...
+
+async function handleItem(item: { url: string; address: string }) {
+  try {
+    let metadata = await IdentityMetadataWrapper.fetchFromURL(item.url)
+    let claims = metadata.filterClaims(ClaimTypes.DOMAIN)
+
+    const numClaims = claims.length
+    for (let i = 0; i < numClaims; i++) {
+      const claim = claims[i]
+      const alreadyVerified = await isClaimAlreadyVerified(item.address, claim.domain)
```
nambrot: This will look up domain claims 1 by 1 in the DB, right? Is there anyway where we can flat map all domains first to get some parallelization?

packages/metadata-crawler/src/crawler.ts
```diff
@@ -0,0 +1,114 @@
+import { IdentityMetadataWrapper } from '@celo/contractkit'
```
nambrot: Do we already have a helm-chart for this/Added this to the blockscout helm chart?

nambrot (changes requested):
Thank you so much for doing this, sorry for the lack of specification :(

packages/cli/src/commands/account/claim-domain.ts
aaitor: As you said, I think makes sense to use the signature of claim itself and not the whole metadata. At the same time I think is useful to have the mechanism of given a domain and the metadata url:
1. download the metadata
2. find if there is a claim inside the metadata referenced in the url provided by the user
3. hash the claim found in the metadata

In that case you not only validate the user owns the domain (via TXT record), also indirectly validate that Metadata URL is valid as a user claims catalog.

While this part is discussed, I will change the code to use as signature the domain claim instead the whole metadata

nambrot: I agree that the whole metadata should be verified. However I would leave that to the caller. The primary means for users to verify is via `get-metadata` anyways which already verifies every single claim, right?

packages/contractkit/src/identity/claims/verify.ts
aaitor: changed the approach for using the hash of the claim instead

packages/metadata-crawler/src/crawler.ts
aaitor: :+1:

packages/metadata-crawler/src/crawler.ts
aaitor: :+1:

packages/metadata-crawler/src/crawler.ts
aaitor: I think the crawler can be executed as a cron. I suggest to do that as part of a different user story. What do you think?

packages/cli/src/commands/account/claim-domain.ts
aaitor: yep, this is using the same plumbing existing in the `IdentityMetadataWrapper` so now there is no much custom implementation about that

packages/contractkit/src/identity/claims/claim.ts
```diff
@@ -35,6 +35,7 @@ export const SignedClaimType = t.type({
   signature: SignatureType,
 })
 
+export const domainTxtHeader = 'celo-site-verification'
```
nambrot: Optional Nit: Consider making the name all CAPS

packages/contractkit/src/identity/claims/domain.test.ts
```diff
@@ -58,12 +59,28 @@ testWithGanache('Domain claims', (web3) => {
 
     describe('when we have a signature', () => {
       it('indicates that signature is correct', async () => {
-        const hasValidSiganture = await verifySignature(
-          hashOfClaims(metadata.claims),
-          signature,
-          address
+        console.log('serializedClaim: ' + serializeClaim(claim))
```
nambrot: Remove console.log

packages/contractkit/src/identity/claims/domain.test.ts
```diff
@@ -75,6 +92,7 @@ testWithGanache('Domain claims', (web3) => {
 
       it('indicates that the metadata does not contain the proper domain claim', async () => {
         const error = await verifyDomainClaim(claim, address, metadataUrlGetter)
+        // console.log(`The message is ${error}`)
```
nambrot: Remove

packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -21,6 +21,7 @@ export async function verifyClaim(
   claim: Claim,
   address: string,
   metadataURLGetter: MetadataURLGetter
+  // signer?: Signer
```
nambrot: Remove

packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -82,68 +83,71 @@ type dnsResolverFunction = (
...
     return `Metadata URL of ${claim.domain} could not be retrieved`
   }
 
   let metadata: IdentityMetadataWrapper
+  let claimFound
   try {
     metadata = await IdentityMetadataWrapper.fetchFromURL(metadataURL)
+    const existingClaims = metadata
+      .filterClaims(ClaimTypes.DOMAIN)
+      .filter((el: DomainClaim) => el.domain === domain)
```
nambrot: Is the explicit type neccessary?

packages/contractkit/src/identity/claims/verify.ts
nambrot: This comment still applies no? You do not need to fetch the metadata here, you can assume that the caller verifies the context of the claim passed

packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -82,68 +83,71 @@ type dnsResolverFunction = (
...
-                  found = verifySignature(hashOfClaims(metadata.claims), signature, address)
-                }
-              })
-            })
-          }
-          resolve()
+  const found = await new Promise((resolve) => {
+    dnsResolver(claim.domain, (error, domainRecords) => {
+      if (error) {
+        console.log(`Unable to fetch domain TXT records: ${error.toString()}`)
```
nambrot: Should we reject the promise here?

packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -82,68 +83,71 @@ type dnsResolverFunction = (
...
+      } else {
+        domainRecords.forEach((record) => {
+          record.forEach((entry) => {
+            if (entry.startsWith(domainTxtHeader)) {
+              console.log(`TXT Record celo-site-verification found`)
+              const signatureBase64 = entry.substring(domainTxtHeader.length + 1)
+              const signature = Buffer.from(signatureBase64, 'base64').toString('binary')
+              // console.log(`Record: ${record}`)
```
nambrot: Remove

packages/metadata-crawler/src/crawler.ts
```diff
@@ -41,60 +49,60 @@ async function addVerificationClaimToDatabase(address: string, domain: string, v
 
     await client
       .query(query, values)
-      .catch((error) => console.error(`Database error ${error}\n query: ${query}`))
-      .then(() => console.log(`Verification flag added to domain ${domain} and address ${address}`))
+      .catch((error) => logger.error(`Database error ${error}\n query: ${query}`))
```
nambrot: Since we are using a structured logger, I would recommend passing parameters structured as well

packages/metadata-crawler/src/crawler.ts
```diff
@@ -41,60 +49,60 @@ async function addVerificationClaimToDatabase(address: string, domain: string, v
...
+      // const alreadyVerified = await isClaimAlreadyVerified(item.address, claim.domain)
+      // logger.debug(`Is already verified? ${alreadyVerified}`)
+      // if (!alreadyVerified) {
+      logger.debug(`Verifying ${claim.domain} for address ${addressWith0x}`)
+
+      const verified = await verifyDomainRecord(addressWith0x, claim).catch((error) =>
+        logger.error(`Error in verifyDomainClaim ${error}`)
+      )
+      if (verified === undefined)
+        await createVerificationClaims(item.address, claim.domain, true, accounts)
```
nambrot: I thought verified indicates success, so shouldn't we only do this when this is true?

packages/cli/src/commands/identity/test-domain-claim.ts
nambrot: ping on this

packages/contractkit/src/identity/claims/claim.ts
aaitor: :+1:

packages/contractkit/src/identity/claims/domain.test.ts
aaitor: :+1:

packages/contractkit/src/identity/claims/domain.test.ts
aaitor: :+1:

packages/contractkit/src/identity/claims/verify.ts
aaitor: :+1:

packages/contractkit/src/identity/claims/verify.ts
aaitor: it's not necessary, removed

packages/contractkit/src/identity/claims/verify.ts
aaitor: :+1:

packages/contractkit/src/identity/claims/verify.ts
aaitor: :+1:

packages/metadata-crawler/src/crawler.ts
aaitor: :+1:

packages/metadata-crawler/src/crawler.ts
aaitor: Used same approach used in other places in contractKit. Keep in that way for consistency. 
I renamed in any case the variable for clarity

packages/cli/src/commands/identity/test-domain-claim.ts
aaitor: Integrated in those commands:
```
bin/run account:show-metadata  ~/Projects/conf/Celo/orco-dev-meta.json
Metadata at /home/aitor/Projects/conf/Celo/orco-dev-meta.json contains the following claims: 

TXT Record celo-site-verification found
Signature verified successfully
Type   Value            Status    Created At 
DOMAIN Domain: orco.dev Verified! 3 days ago 
```

packages/contractkit/src/identity/claims/verify.ts
aaitor: wiring in the way the verify command doesn't download the metadata

packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -66,3 +70,81 @@ export const verifyAccountClaim = async (
...
+  const found = await new Promise((resolve) => {
+    dnsResolver(claim.domain, (error, domainRecords) => {
+      if (error) {
+        console.log(`Unable to fetch domain TXT records: ${error.toString()}`)
+        resolve(false)
+      } else {
+        domainRecords.forEach((record) => {
+          record.forEach((entry) => {
+            if (entry.startsWith(DOMAIN_TXT_HEADER)) {
+              console.log(`TXT Record celo-site-verification found`)
```
nambrot: Given that we use this in user-facing CLI, I would like to reduce the debug statements to a minimum.

packages/contractkit/src/identity/metadata.ts
```diff
@@ -105,7 +106,11 @@ export class IdentityMetadataWrapper {
           throw new Error("Can't claim self")
         }
         break
-
+      case ClaimTypes.DOMAIN:
+        const existingClaims = this.data.claims.filter((el: any) => el.domain === claim.domain)
```
nambrot, line 134: Consider using `.find` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

packages/contractkit/src/identity/claims/verify.ts
aaitor: moved to `console.debug`, allowing to get those messages if the logger is initialized with the debug level for development purposes

packages/contractkit/src/identity/metadata.ts
aaitor: personally I like `filter` because returns always an array (with elements or empty), and find could return `undefined`. So from a readability point of view, next sentences where you use the result of this call looks more evident to me.

nambrot: Your call, imo 

```
const foundElement = find()
if (foundElement) return foundElement
```

just looks as readable to me?

nambrot (approved):

PR #3157, opened 2020-03-20
Fix: coinlist text not rendering

aaronmgdr:
### Description

issue: Coinlist text not rendering

cause: incorrect brakets in json file,

reason it was not found earlier: tests don't load the translations and this content is not viewable inside the US.

### Tested

pushed to staging and confirmed with Trever who is outside the US that it shows correctly


cmcewen (approved):
Could we add some mocks to the tests to ensure that this shows up?

aaronmgdr:
@cmcewen Was just about to create a ticket for having snapshot tests (or all tests) load the translation files.

aaronmgdr:
https://github.com/celo-org/celo-monorepo/issues/3159

jmrossy (approved):

PR #3164, opened 2020-03-23
[ContractKit] Ledger Wallet

### LINKED ISSUE #3075: Support use of Ledger Hardware wallets from CLI
### Expected Behavior

We need to be able to sign transaction from CLI using a ledger Nano

### Solution Idea

Define a new implementation for `Wallet` interface that is being used on `CeloProvider`. We might need to review a few regarding how `kit` uses `CeloProvider`. For example the `kit.addAccount()` method won't work out of the box, and need to have a different sematics; since it depends on the wallet being used.

gastonponti:
### Description

Adding support to use Ledger in the contractKit. Now the user can create a new wallet that will sign the local transactions using the ledger

### Tested

Manually

### Other changes

Refactor the singning-utils, found some issues with values that were not hex

### Related issues

- Fixes #3075 

### Backwards compatibility

Ok, the "normal" wallet is the default


packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -28,111 +28,136 @@ function makeEven(hex: string) {
...
-  let result: EncodedTransaction
+export interface RLPEncodedTx {
+  transaction: Tx
+  rlpEncode: any
+}
 
-  if (!txn) {
-    throw new Error('No transaction object given!')
-  }
+export function chainIdTransformationForSigning(chainId: number): number {
```
mcortesi, line 39: Add a comment about the EIP where this is enforced

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -20,19 +21,23 @@ enum InterceptedMethods {
...
+  constructor(readonly existingProvider: provider, wallet: Wallet = new DefaultWallet()) {
     this.rpcCaller = new DefaultRpcCaller(existingProvider)
     this.paramsPopulator = new TxParamsNormalizer(this.rpcCaller)
-    this.wallet = new DefaultWallet()
+    this.wallet = wallet
   }
 
   addAccount(privateKey: string) {
-    this.wallet.addAccount(privateKey)
+    if (hasProperty<{ addAccount: (privateKey: string) => void }>(this.wallet, 'addAccount')) {
```
yorhodes, line 36: +1

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -28,111 +28,136 @@ function makeEven(hex: string) {
...
-  }
+export function getHashFromEncoded(rlpEncode: string): string {
+  return Hash.keccak256(rlpEncode)
+}
 
-  // Resolve immediately if nonce, chainId and price are provided
-  if (txn.nonce !== undefined && txn.chainId !== undefined && txn.gasPrice !== undefined) {
-    return signed(txn)
-  }
+function stringNumberToHex(num: string | number): string {
```
yorhodes: I think we have some of this in the utils package?

packages/contractkit/src/wallets/wallet.ts
```diff
@@ -0,0 +1,11 @@
+import { EncodedTransaction, Tx } from 'web3-core'
+import { Address } from '../base'
+import { EIP712TypedData } from '../utils/sign-typed-data-utils'
+
+export interface Wallet {
```
yorhodes, line 5: +1

packages/contractkit/src/providers/celo-provider.ts
yorhodes: actually if wallet is strongly typed do we need this?

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -28,111 +28,136 @@ function makeEven(hex: string) {
...
+      s: signature.s,
+      hash,
+    },
+    raw: rawTransaction,
+  }
+  return result
+}
+
+export async function signTransaction(tx: Tx, privateKey: string): Promise<EncodedTransaction> {
+  if (!tx) {
```
yorhodes, line 166: do we need this?

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -28,111 +28,136 @@ function makeEven(hex: string) {
...
+export interface RLPEncodedTx {
+  transaction: Tx
+  rlpEncode: any
+}
 
-  if (!txn) {
-    throw new Error('No transaction object given!')
-  }
+export function chainIdTransformationForSigning(chainId: number): number {
+  return chainId * 2 + 35
```
yorhodes, line 40: what are these constants?

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -0,0 +1,164 @@
...
+   * [`${baseDerivationPath}/3`, `${baseDerivationPath}/99`, `${baseDerivationPath}/53`]
+   * @param baseDerivationPath base derivation path. Default: "44'/52752'/0'/0"
+   * @param devicePath path to the ledger device, otherwise will use the first one it finds
+   */
+  constructor(
+    readonly derivationPathIndexes: number[] = Array.from(Array(ADDRESS_QTY).keys()),
+    readonly baseDerivationPath: string = CELO_BASE_DERIVATION_PATH,
+    readonly devicePath: string = ''
+  ) {
+    const validDPs = derivationPathIndexes.reduce(
```
yorhodes: `filter`?

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -0,0 +1,164 @@
...
+      (valid, currentValue) => valid && Number.isInteger(currentValue) && currentValue >= 0,
+      true
+    )
+    if (!validDPs) {
+      throw new Error('ledger-wallet: Invalid address index')
+    }
+    this.setup()
```
yorhodes: I have been told that this isn't an idiomatic pattern
Any async initialization should happen outside of the constructor

yorhodes (changes requested):

packages/contractkit/src/utils/signing-utils.ts
gastonponti: no, hahaha, not this
👍

packages/contractkit/src/utils/signing-utils.ts
gastonponti: This was used before, it's not something weird that I added, haha
It's a protection from an specific attack:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

packages/contractkit/src/providers/celo-provider.ts
gastonponti: I'm not adding `addAccount` to the ledger
I could add the `addDerivationPath` but it's not the same as the account, that's why I removed it from the `Wallet` interface

packages/contractkit/src/utils/signing-utils.ts
yorhodes: add a comment with link? 🙏

packages/contractkit/package.json
```diff
@@ -28,6 +28,8 @@
   },
   "dependencies": {
     "@celo/utils": "0.1.9-dev",
+    "@ledgerhq/hw-app-eth": "^5.11.0",
+    "@ledgerhq/hw-transport-node-hid": "^5.11.0",
```
mcortesi: this can be an optional dependency

packages/contractkit/src/wallets/ledger-wallet.ts
mcortesi: agree

mcortesi (commented):
beside the `setup()` inside the constructor the rest looks ok, but added a few comments

packages/contractkit/src/wallets/ledger-types.d.ts
```diff
@@ -0,0 +1,83 @@
+import TransportNodeHid from '@ledgerhq/hw-transport-node-hid'
```
mcortesi: This file a `d.ts.` looks like it's in a bad location.. shouldn't this be part of `/types` 
how would others refer to this?

packages/contractkit/src/wallets/ledger-wallet.ts
gastonponti: filter will remove the bad ones, I'm verifying that everyone is ok
I'll use `some` (that's basically does that)

packages/contractkit/src/wallets/ledger-wallet.ts
gastonponti: Wasn't suppose to keep that on the commit (sorry)

packages/contractkit/src/wallets/ledger-types.d.ts
gastonponti: I found a really nice workaround. I removed it ;)

packages/contractkit/src/utils/signing-utils.ts
gastonponti: No, and this one asume the 0 as an empty

yorhodes (approved):
LGTM! excited to try using cli with my ledger

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -0,0 +1,222 @@
...
+  rlpEncodedTx,
+  signatureFormatter,
+} from '../utils/signing-utils'
+import { Wallet } from './wallet'
+
+export const CELO_BASE_DERIVATION_PATH = "44'/52752'/0'/0"
+const ADDRESS_QTY = 5
+
+export async function newLedgerWalletWithSetup(
+  transport: any,
```
cmcewen, line 22: I think you can make this the `Transport` type: `import Transport from '@ledgerhq/hw-transport'`

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -0,0 +1,222 @@
...
+      throw new Error('ledger-wallet: Invalid address index')
+    }
+  }
+
+  /**
+   * @param transport Transport to connect the ledger device
+   */
+  async init(transport: any) {
+    try {
+      if (this.setupFinished) {
```
codyborn, line 64: I believe this check is insufficient to prevent race condition. If this function is called twice before this.setupFinished is set, the initialization logic will run twice. This doesn't seem to cause any misbehavior in the current implementation, but may cause bugs in the future.

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -0,0 +1,222 @@
...
+    } else {
+      return false
+    }
+  }
+
+  async signTransaction(txParams: Tx): Promise<EncodedTransaction> {
+    this.initializationRequired()
+    try {
+      const rlpEncoded = rlpEncodedTx(txParams)
+      const zrxInfo = await byContractAddress(txParams.to)
```
codyborn, line 120: It looks like this is pulling data from a static base64 encoded file to provide mapping from address to ERC20 metadata:
https://github.com/LedgerHQ/ledgerjs/blob/0471ef03e975d1430be2637193d37e8f529baddc/packages/hw-app-eth/src/erc20.js#L39
If this is true, this metadata would only be applicable for Ethereum mainnet. The ERC20s deployed to Celo network will have different addresses than Ethereum. In the future we could implement our own mappping, but I think we should remove this for now.

yorhodes: 🧐
@gastonponti

gastonponti: Yes. I thought that I could add it for the future, and then if our ledger app supports contracts data, we could just make a fork of `hw-eth-app`, add the blob for the data of our contracts,  and just change the version in the `package.json`. But I removed it in other PR, just to make the code clearer. In the future if we need it, we will add it again. 👍

packages/contractkit/src/wallets/ledger-wallet.ts
gastonponti: Also, fixed in other PR #3256

packages/contractkit/src/wallets/ledger-wallet.ts
gastonponti: PR: #3256

PR #3169, opened 2020-03-23
[ContractKit] fix nonce parsing, fix gatewayFee default

gastonponti:
### Description

Changes to adapt types, broke how the nonce was parsed
Fixed now.

Also the default gatewayFee was an integer assumed as an hex, and that made a lower number

### DIFF
--- a/packages/contractkit/src/utils/tx-params-normalizer.ts
+++ b/packages/contractkit/src/utils/tx-params-normalizer.ts
@@ -1,3 +1,4 @@
+import { ensureLeading0x } from '@celo/utils/lib/address'
 import BigNumber from 'bignumber.js'
 import { Tx } from 'web3-core'
 import { RpcCaller } from './rpc-caller'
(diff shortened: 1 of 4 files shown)


packages/contractkit/src/utils/tx-params-normalizer.ts
```diff
@@ -42,7 +42,7 @@ export class TxParamsNormalizer {
     }
 
     if (!isEmpty(txParams.gatewayFeeRecipient) && isEmpty(txParams.gatewayFee)) {
-      txParams.gatewayFee = DefaultGatewayFee.toString(16)
+      txParams.gatewayFee = '0x' + DefaultGatewayFee.toString(16)
```
yorhodes: I know this is a short PR but can we use the `utils` functions?

yorhodes (approved):
see comment

packages/contractkit/src/utils/tx-params-normalizer.ts
gastonponti: hahaha, yes

mcortesi (approved):

packages/contractkit/src/utils/tx-params-normalizer.ts
```diff
@@ -64,7 +65,9 @@ export class TxParamsNormalizer {
   private async getNonce(address: string): Promise<number> {
     // Reference: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactioncount
     const result = await this.rpcCaller.call('eth_getTransactionCount', [address, 'pending'])
-    const nonce = parseInt(result.result.toString(), 10)
+
+    // tslint:disable-next-line: radix
+    const nonce = parseInt(result.result.toString())
```
jmrossy: Why not just provide the 16 as the radix instead of disabling tslint there?
Or even better, use BigNumber which has more robust parsing then parseInt

PR #3172, opened 2020-03-24
Reserve spend limits should not apply to CP-DOTO

asaj:
### Description

This PR removes the restriction on reserve spending from transfers initiated on behalf of the stability protocol.

### Tested

Unit tests.

### Other changes

None

### Related issues

None
### Backwards compatibility

Yes


aslawson (commented):
can you add details in the description?

I'm not sure what CP-DOTO stands for, but the transferExchangeGold method looks like it can only be accessed by registered contracts and the _transferGold internally, so seems to be a safe change that allows only specified contracts to bypass a daily spending limit

LGTM

aslawson (approved):

jose-compu:
CP-DOTO is the Exchange.sol, mainly around exchange().

rcroessmann (approved):
LGTM!

jose-compu:
Yes, is removing the daily spend limit from transferExchangeGold() that is called by CP-DOTO. Daily spend limit only checked for Custody transferGold() to otherAddresses.

PR #3174, opened 2020-03-24
Fund release gold signing keys

asaj:
### Description

This PR allows the first signing key of each type authorized by a release gold beneficiary to be funded with 1 cGLD for transaction fees, which is necessary for the signing keys to be usable while Celo Gold transfers remain disabled.
### Tested

Unit tests

### Other changes

None
### Related issues

None
### Backwards compatibility

Yes


packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -477,6 +477,18 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
...
   }
 
+  /**
+   * @notice Funds a signer address so that transaction fees can be paid.
+   * @param signer The signer address to fund.
+   */
+  function fundSigner(address payable signer) private {
+    // Fund signer account with 1 cGLD.
+    uint256 value = 1 ether;
+    totalWithdrawn = totalWithdrawn.add(value);
```
lucasege: Perhaps unintuitively, I don't think we want to update `totalWithdrawn` here. Making it 1 while `getCurrentReleasedTotalAmount` is still 0 will cause reverts whenenever we try `releasedAmount.sub(totalWithdrawn)` in `withdraw` and `refundAndFinalize`. 

I believe this will break the revoking flow - if `totalWithdrawn` is set to 1, but this contract gets revoked, our `releasedBalanceAtRevoke` will be 0 since the ratio `periodsSinceStart/numReleasePeriods` will be 0. Then in `refundAndFinalize` this line:
`uint256 beneficiaryAmount = revocationInfo.releasedBalanceAtRevoke.sub(totalWithdrawn);` will always revert, locking the funds. 

I'd suggest not updating `totalWithdrawn` and just treating this like it reduces the grant by 1CG, which it effectively does.

Let me know if that makes sense to you - still thinking out the proposed solution to make sure it doesn't introduce any other issues

lucasege (changes requested):

packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -504,12 +520,16 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
...
+  function authorizeValidatorSigner(address payable signer, uint8 v, bytes32 r, bytes32 s)
     external
     nonReentrant
     onlyCanValidate
     onlyWhenInProperState
   {
+    // If no previous signer has been authorized, fund the new signer so that tx fees can be paid.
+    if (getAccounts().getValidatorSigner(address(this)) == address(this)) {
+      fundSigner(signer);
+    }
```
aslawson, line 540: is this something we really only want to ReleaseGold?  And do other contracts even use these account contract methods?

Otherwise, I don't see the advantage of having this logic here here where you need to make two requests to the accounts contract rather than letting the accounts contract handle this.

aslawson (commented):
i'm not familiar with the consequences of changing the totalWithdrawn, so I'll let lucas give the shipit

packages/protocol/contracts/governance/ReleaseGold.sol
asaj: Makes total sense, nice catch.

packages/protocol/contracts/governance/ReleaseGold.sol
asaj: Yeah I think only for release gold. Once we hit mainnet, folks can just fund their signing keys themselves, but while transfers are disabled that won't be possible

packages/protocol/contracts/governance/ReleaseGold.sol
asaj:
```suggestion
```

packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -477,6 +477,17 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
     getLockedGold().withdraw(index);
   }
 
+  /**
+   * @notice Funds a signer address so that transaction fees can be paid.
+   * @param signer The signer address to fund.
```
asaj, line 490:
```suggestion
   * @param signer The signer address to fund.
   * @dev Note that this effectively decreases the total balance by 1 cGLD.
```

lucasege (approved):
LGTM pending failing test fix or flakiness acknowledgement

PR #3180, opened 2020-03-24
[Wallet] Add Colombian pesos to local currency list

jmrossy:
### Description

Add Colombian pesos to local currency list

### Tested

Locally on iOS

### Backwards compatibility

Yes


annakaz (approved):
LGTM

PR #3190, opened 2020-03-25
Snapshot the website text

### LINKED ISSUE #3159: snapshot Tests  should include translation values
currently snapshots only save the translate key, this leads possibility of a translation being messed up and not caught before deploying to prod

solution: Use i18n providor so that that translation values are recorded in snaoshots

aaronmgdr:
### Description

While we have snapshot test for each page of website they were previously saving the translation key like 'home.title'. This led to a huge possibility for translation files to get screwed up and not notice it till to late. Now we see any changes or issues in the tests!

HOW: Wrap each tested component in a new `TestProvider` which itself uses `next-i18next`'s appWithTranslate with a modified initializer that loads all translation files synchronously. 

### Tested
❇️
note that conflicts are inevitable and will be fixed regularly

### Other changes

also fixes some issues with fetch mocking. 

### Related issues

- Fixes #3159 

### Backwards compatibility

✅


aaronmgdr:
note that conflicts are inevitable and will be fixed regularly

cmcewen (approved):

PR #3207, opened 2020-03-26
[Wallet] Refactor PIN entering strategy throughout the wallet app

### LINKED ISSUE #3104: [Wallet] Refactor PIN entering/verifying throughout the app

### LINKED ISSUE #3223: iOS : App gets unlocked while entering incorrect pin after enabling security pin from ‘Settings’ page.
**Frequency:** 100%
**iOS App version:** iOS Playstore build : v1.6.2
**Repro on iOS Device:** iPhone 11 Pro (13.3)
**Testing Account:** 3155641283 /backup Key

**Repro Steps:** 
1) Launch the app and Login with test account 
2) Navigate to ‘Settings > Security’ page
3) Enter pin number and Enable ‘Require PINon App Open’ toggle 
4) Background-Foreground the app and Enter incorrect PIN number 
5) Observe 

**Current Behavior:** App gets unlocked while entering incorrect pin 

**Expected Behavior:** ‘Incorrect Pin’ error message should be shown after entering incorrect PIN number 

**Investigation:** 
1. User also redirects to ‘Security’ page after entering incorrect password 
2. ‘Incorrect Pin’ error message is shown when user tried to proceed with payment/gold transaction using incorrect PIN number

**Attachment:** [iOS_Incorrect pin is accepted.mp4](https://drive.google.com/file/d/1X8w57JlQCBMklLPmX-dBdbFBnNVYIMkG/view?usp=sharing) (Bug observed at 25 secs)

i1skn:
### Preamble

It's a second try. The first one is [here](https://github.com/celo-org/celo-monorepo/pull/2997). I've decided to create this one to start clean.

### Description

This PR aims to improve stability and transparency of PIN entering process throughout the wallet. 
* Split `getPincode` into two different methods: `getPincode (to just fetch pincode from whatever source is now available` and `ensurePincode (to ensure, that user know a correct PIN)`.
* `app/saga -> handleNavigatePinProtected` logic moved to `NavigationService -> navigateProtected` .
* Change the way, how app is locked: even if **PIN is not cached** the app would wait for 30 seconds until it will be locked. This is handy, when user switches between apps to enter verification codes, for example. Also, now we show lock screen as an overlay, not an actual screen using navigation.
* Unify `Pincode` view and use it in screens, when PIN entering is needed.
* Rename `PincodeConfirmation.tsx` to `PincodeEnter.tsx` as it logically make more sense.
* Lock wallet with PIN is now working in Forno

### Tested

Android and iOS devices

### Other changes

* Refactor `react-components/components/Button.tsx`. The reason is that `componentDidUpdate` used there were updating `debouncedOnPress` method **only for the next** render. I've done it with hooks, because it was naturally the case for them.
* Fix vertical alignement in backup phrase:
IMG_8660
IMG_8659

### Related issues

- Fixes #3104
- Fixes #3223

### Backwards compatibility

Yes


packages/mobile/src/app/saga.ts
```diff
@@ -207,23 +175,26 @@ function* watchAppState() {
...
-      },
-    })
+  if (
+    !cachedPin &&
+    lockWithPinEnabled &&
+    now - lastTimeBackgrounded > DO_NOT_LOCK_PERIOD &&
+    action.state === 'active' &&
+    !appLocked
+  ) {
+    yield put(lock())
```
jmrossy: nit, consider giving the `lock` action a more descriptive name

packages/mobile/src/backup/BackupPhraseContainer.tsx
```diff
@@ -155,6 +155,7 @@ const styles = StyleSheet.create({
     minHeight: 125,
     padding: 14,
     paddingTop: 16,
+    textAlignVertical: 'top',
```
jmrossy, line 158: Before merging this, please ensure this looks right on both iOS and Android. I believe I fixed a bug with backup phrase styling on iOS a week or two ago and I'm worried this may regress it.

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -50,12 +53,47 @@ export function navigate(routeName: string, params?: NavigationParams) {
     })
 }
 
-export function navigateAfterPinEntered(routeName: string, params?: NavigationParams) {
-  navigate('Background', {
-    onUnlock() {
-      navigate(routeName, params)
-    },
-  })
+export async function ensurePincode(): Promise<boolean> {
```
jmrossy: Not a big deal but I would expect this code to live next to other pin related logic

packages/mobile/src/pincode/Pincode.tsx
```diff
@@ -0,0 +1,90 @@
...
+  )
+
+  const onBackspacePress = useCallback(() => {
+    onChangePin(pin.substr(0, pin.length - 1))
+  }, [pin, onChangePin])
+
+  const validPin = useMemo(() => isPinValid(pin), [pin, isPinValid])
+
+  return (
+    <>
```
jmrossy, line 46: why this fragment?

packages/mobile/src/pincode/Pincode.tsx
```diff
@@ -0,0 +1,90 @@
...
+  title: string
+  placeholder: string
+  buttonText: string
+  maxLength?: number
+  isPinValid: (pin: string) => boolean
+  pin: string
+  onChangePin: (pin: string) => void
+}
+
+function Pincode(props: Props) {
```
jmrossy, line 25: Consider adding a comment here about the diff btwn this file, PincodeEnter, and PincodeLock. I had to look around for a bit to figure it out, wasn't immediately obvious

packages/react-components/components/Button.tsx
```diff
@@ -30,95 +30,95 @@ export interface ButtonProps {
...
+      backgroundColor = 'transparent'
+      borderColor = disabled ? colors.celoGreenInactive : colors.celoGreen
+      break
+    case BtnTypes.TERTIARY:
+      textColor = disabled ? colors.inactiveDark : colors.dark
+      backgroundColor = 'transparent'
+      borderColor = 'transparent'
+      break
+    default:
+      if (__DEV__) {
```
jmrossy, line 77: I would just throw, even in not dev mode. Helps us catch the error in the wild if this slips through cause it would make its way to Sentry

jmrossy (commented):
Nice work, seems like a good approach. And good job with the component test coverage! Just a few small comments

packages/mobile/src/pincode/Pincode.tsx
i1skn: Cause I wanted to group together `ScrollView` and `Button`, otherwise `View` would be needed.

packages/mobile/src/navigator/NavigationService.ts
i1skn: `ensurePincode` is only used in `navigateProtected`, so I've decided to put it here for now. `getPincode` lives also in `src/account/saga` and has a different purpose, so I decided to keep them separate and close to the places they've used. But I've removed `export` for this method, though.

packages/mobile/src/backup/BackupPhraseContainer.tsx
i1skn: `textAlignVertical` is Android specific (at least it say so here https://reactnative.dev/docs/text), but I've double checked it on iOS and backup phrase input LGTM:
IMG_8667
IMG_8668

packages/react-components/components/Button.tsx
i1skn: Because we provide default values, I think there is a chance it has been used already somewhere without a type. Hence, if we would want to make it strict - I would check all the usage of `Button` first. But this is a bit outside of the scope of this PR. WDYT?

packages/mobile/src/app/saga.ts
i1skn: Agree, renamed it to `appLock` and `appUnlock`

i1skn (commented):
@jmrossy thanks for the feedback! I've addressed the comments, PTAL.

packages/mobile/src/pincode/Pincode.tsx
i1skn: ✅

jmrossy (approved):

PR #3208, opened 2020-03-26
Show only expired governance proposals in "Expired" output

timmoreton:
### Description

Currently, when doing `governance:list` proposals that are not expired are listed in the 'Expired' section. Filter the output so this isn't so.

### Tested

By inspection

### Backwards compatibility

Yes


asaj (approved):
Whoops!

yorhodes (approved):

PR #3209, opened 2020-03-27
Remove duplicated words on the attestation docs

come-maiz:
### Description

There was a wrong sentence with duplicated words:
"without the without the"

### Tested

Just checked the md rendering in github.

### Other changes

None

### Related issues

None

### Backwards compatibility

Only documentation changes.


annakaz (approved):
Thanks for the contribution, @elopio !

PR #3214, opened 2020-03-27
Add setTargetVotingYield for EpochRewards contract accessible to governance

aslawson:
### Description

- a method for setting the target voting is needed to "unfreeze" voter rewards.  The target voting yield is initially set to 0

### Tested

NA

### Other changes

NA

### Related issues

NA

### Backwards compatibility

Will need to either upgrade contract in rc0 or just use in mainnet.


packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -287,6 +287,15 @@ contract EpochRewards is
     return true;
   }
 
+  function setTargetVotingYield(uint256 targetVotingYield) public onlyOwner {
```
asaj: Can we use this in `initialize`?

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -287,6 +287,15 @@ contract EpochRewards is
...
+    FixidityLib.Fraction memory target = FixidityLib.wrap(targetVotingYield);
+    require(
+      target.lte(targetVotingYieldParams.max),
+      "Target voting yield must be less than or equal to max"
+    );
+    targetVotingYieldParams.target = target;
```
asaj, line 303: Can we emit an event?

packages/protocol/contracts/governance/EpochRewards.sol
asaj: Missing natspec

asaj: Unit tests?

aslawson: i considered it, but it adds a requirement -- I didn't want to alter the current code implementation.

it does seems like it would be a benefit to the initialize to add the require statement though, so I can update

aslawson: will add

packages/protocol/contracts/governance/EpochRewards.sol
aslawson: sure

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -287,6 +287,15 @@ contract EpochRewards is
...
+  }
```
rcroessmann, line 306: Should we `return true;` here? Seems to be a convention we adopted.

aslawson: yup -- i updated

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -73,6 +73,7 @@ contract EpochRewards is
   event CarbonOffsettingFundSet(address indexed partner, uint256 fraction);
   event TargetValidatorEpochPaymentSet(uint256 payment);
   event TargetVotingYieldParametersSet(uint256 max, uint256 adjustmentFactor);
+  event TargetVotingYieldSet(uint256 target);
```
asaj, line 76: We have a TargetVotingYieldUpdated event, suggest we stick with one event

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -287,6 +288,22 @@ contract EpochRewards is
     return true;
   }
 
+  /**
+   * @notice Sets the target voting yield.
+   * @param targetVotingYield The relative target block reward for voters.
```
asaj, line 294: nit: mention fixidity

asaj (approved):

packages/protocol/contracts/governance/EpochRewards.sol
aslawson: decided we may want to differentiate between the target being set via the conditions vs governance, so will still with the new event `TargetVotingYieldSet`

rcroessmann (approved):

PR #3216, opened 2020-03-27
Remove gateway fee default values

nategraf:
### Description

In https://github.com/celo-org/celo-blockchain/pull/941 the default value for minimum gateway fee is changed to 0. This disables the mechanism by default. To match this, this PR removes the default values for gateway fee from the codebase and adjust the e2e test to continue to provide coverage for this feature.

### Tested

Existing tests

### Other changes

* Adjust the default minimum client version
* Rework parts of the gateway fee docs

### Backwards compatibility

Will not be able to submit LES transactions to nodes built before https://github.com/celo-org/celo-blockchain/pull/941 which run with default gateway fee values.


jmrossy (commented):
How should a light client retrieve the right value going forward?

nategraf:
We don't have support for retrieving the gateway fee value yet (which indeed does make the feature largely unusable without outside channels) but the default value was just masking that problem before.

packages/mobile/src/dappkit/dappkit.ts
```diff
@@ -70,7 +69,7 @@ function* produceTxSignature(action: RequestTxSignatureAction) {
       // In walletKit we use web3.eth.getCoinbase() to get gateway fee recipient
       // but that's throwing errors here. Not sure why, but txs work without it.
       const gatewayFeeRecipient = undefined
-      const gatewayFee = '0x' + new BigNumber(10000).toString(16)
+      const gatewayFee = undefined
```
jmrossy, line 72: Here we make the fee undefined but in walletkit you use `0x0`. Which is the recommended approach?

nategraf: I set it as `0x0` instead of `undefined` undefined there because the gateway fee recipient wanted one being defined to mean the other is also defined. (It doesn't matter top geth though because it will treat `nil` as 0)

jmrossy (approved):

PR #3218, opened 2020-03-27
Brandkit Downloads Tracking

### LINKED ISSUE #3217: track brandkit asset downloads
add tracking to anything that can be download from brandkit

icons
logos
graphics
etc

aaronmgdr:
### Description

Adds event tracking to BrandKit for when visitors download the following

Icons
Abstract Graphics
Illustrations
Logos
Celo Brand Voice Guide
Fonts

### Tested

changed the track function on dev to print to console and clicked around

### Other changes

on develop, analytics events now print to console

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/3217


packages/web/src/brandkit/tracking.ts
```diff
@@ -0,0 +1,22 @@
+import analytics from 'src/analytics/analytics'
```
jmrossy, line 1: Seems like this won't be just a brandkit requirement right? Maybe this should be further generalized

packages/web/src/brandkit/tracking.ts
```diff
@@ -0,0 +1,22 @@
...
+export enum AssetTypes {
+  logo = 'logo',
+  icon = 'icon',
+  font = 'font',
+  illustration = 'illustration',
+  graphic = 'graphic',
+}
+
+export const GLYPH_TRACKING = { name: 'Celo Glyphs', type: AssetTypes.logo }
```
jmrossy: Consider creating an interface for these

jmrossy (approved):
Approved with an optional suggestion

packages/web/src/brandkit/tracking.ts
aaronmgdr: thats a good point! although the trackDownload function itself is really just for brandkit.

PR #3220, opened 2020-03-28
[ContractKit] Add subscription to the CeloProvider

gastonponti:
### Description
Fix for the CeloProvider. Add subscription delegating to the provider wrapped

### DIFF
--- a/packages/contractkit/src/providers/celo-provider.ts
+++ b/packages/contractkit/src/providers/celo-provider.ts
@@ -131,6 +131,22 @@ export class CeloProvider {
     }
   }
 
+  on(type: string, callback: () => void): void {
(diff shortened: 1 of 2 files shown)


packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -131,6 +131,47 @@ export class CeloProvider {
     }
   }
 
+  on(type: string, callback: () => void): void {
```
mcortesi: i *think* that web3.js check `provider has "on"` to check if the provider supports subscriptions.

So, i would define these properties only if the existing provider supports them

mcortesi (approved):

PR #3222, opened 2020-03-30
Fix codecov statuses

jeanregisser:
### Description

Following up on https://github.com/celo-org/celo-monorepo/pull/3049#issuecomment-602592536,
though I haven't heard back from codecov support, I found a way to disable `codecov/project` and `codecov/status` statuses which we don't want.

### Tested

Statuses in the PR for codecov are only the expected `codecov/project/mobile` and `codecov/project/web`.


jeanregisser:
Looks like problem is fixed on codecov's side. I don't see the status appearing for `codecov/project` and `codecov/status` anymore.

PR #3226, opened 2020-03-30
[Wallet] Use decimal separator appropriate to localisation

### LINKED ISSUE #2320: Wallet displayed money amount should not show NaN
### Expected Behavior

Actual money value should be displayed correctly for very big numbers.

### Current Behavior

If you enter a very big amount in the exchange input, you end up with NaN:
image

### LINKED ISSUE #3221: [Wallet] Sending less than 1 cUSD, the decimal doesn't work
### Expected Behavior

User SBAT to send decimal amounts

### Current Behavior

Only `.` is accepted and not `,` as in some localisations.

As a part of https://github.com/celo-org/celo-monorepo/issues/3068

i1skn:
### Description

This PR changes the way of how we set decimal separator throughout the app. Instead of checking the language, we should rely on user's phone localisation, which provides decimal separator. 
Another reason to rely on it is that on iPhone - number keyboard would only show appropriate to localisation decimal separator. 

I've used `react-native-localize -> getNumberFormatSettings ()` to get `decimalSeparator`. I'm concerned, that some phones might not support it, but I have not found any info about that. So please let me know if you know otherwise.

### Screenshots

IMG_8674
IMG_8673

### Tested

 - [x] iPhone with German localisation
 - [x] iPhone with US localisation
 - [x] Android with German localisation
 - [x] Android with US localisation

### Related issues

- Fixes #3221
- Fixes #2320

### Backwards compatibility

Yes


jeanregisser (approved):
Awesome! 💯

packages/mobile/src/utils/numberFormat.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
```
jmrossy: consider merging this file with utils/formatting.ts or renaming it to make it clear its for the HOC

jmrossy (approved):
Nice work! One optional comment below

PR #3227, opened 2020-03-30
Modify active vote and unit calculations

asaj:
### Description

This PR modifies the election active vote units and value calculations to first derive the number of units, and then re-derive the corresponding amount of value. This prevents attacks that specify an amount so small that it rounds down to zero units.

### Tested

Added unit tests (WIP)

### Other changes

None
### Related issues

None
### Backwards compatibility

No


asaj:
Deprecated in favor of https://github.com/celo-org/celo-monorepo/pull/3258

PR #3233, opened 2020-03-31
[PhoneNumPriv] Add support for connecting to a Postgresql DB

jmrossy:
### Description

- Add support for connecting to a Postgresql DB
- Create DB migrations for the two tables
- Modify config to support running locally
- Cleanup test imports

### Tested

In our shiny new staging project:
https://console.cloud.google.com/sql/instances/phone-number-privacy-db/overview?project=celo-phone-number-privacy-stg&duration=PT1H&cloudshell=false

### Backwards compatibility

Yes


packages/phone-number-privacy/migrations/20200330212301_create-number-pairs-table.ts
```diff
@@ -0,0 +1,12 @@
+import * as Knex from 'knex'
+
+export async function up(knex: Knex): Promise<any> {
+  return knex.schema.createTable('number_pairs', (t) => {
```
codyborn: Can we define these strings ('number_pairs', 'accounts', 'user_phone_hash', 'contact_phone_hash', etc) in a constants file?

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,10 +1,17 @@
...
 import * as functions from 'firebase-functions'
-import { PHONE_NUMBER_PRIVACY_SECRET_KEY } from './config'
+import config, { connectToDatabase } from './config'
 
 export const getSalt = functions.https.onRequest((request, response) => {
-  const privateKey = new Buffer(PHONE_NUMBER_PRIVACY_SECRET_KEY)
+  const privateKey = new Buffer(config.salt.key)
   try {
+    // Adding this here as an example of how to connect to the DB
+    const knex = connectToDatabase()
```
codyborn, line 9: Instead of returning the raw knex object, it would nice to raise the abstraction to the caller. We can wrap the database with a new class that provides high level operations and provide a strongly typed class for PhoneNumber to perform any validation. Ex.
interface PhoneNumberDatabase {
  hasPhoneNumber(number: PhoneNumber): Promise<boolean>
  getLookupsByPhoneNumber(number: PhoneNumber): Promise<number>
  incrementPhoneNumberLookup(number: PhoneNumber, amount: number)
  ...
}

This will ensure that state management is not spread out throughout the code and allow us to:
1. Provide an interface that can be extended in the future to other state store types (ex. in mem cache)
2. Make it easier to reason about the state management (ex. which race conditions do we need to worry about)
3. Add/remove properties to the state object in a single place

packages/phone-number-privacy/src/config.ts
```diff
@@ -1,4 +1,57 @@
...
+      password: functionConfig.db_pass,
+      database: functionConfig.db_name,
+      host: `/cloudsql/${functionConfig.db_host}`,
+    },
+  }
+}
+
+export const connectToDatabase = () => {
+  console.debug('Creating knex instance')
+  return knex({
```
codyborn, line 50: What are the types of errors that can be thrown when creating this connection?

packages/phone-number-privacy/migrations/20200330212224_create-accounts-table.ts
```diff
@@ -0,0 +1,17 @@
+import * as Knex from 'knex'
+
+export async function up(knex: Knex): Promise<any> {
+  return knex.schema.createTable('accounts', (t) => {
+    t.increments('id')
+      .unsigned()
+      .primary()
+    t.dateTime('createdAt').notNullable()
+    t.string('address').notNullable()
```
aslawson: curious -- any reason the address itself can't be the primary key?

packages/phone-number-privacy/src/config.ts
```diff
@@ -1,4 +1,57 @@
...
+    user: string
+    password: string
+    database: string
+    host: string
+  }
+}
+
+let config: Config
+
+if (DEV_MODE) {
```
codyborn, line 20: What is the general strategy for supporting unit tests? Should we create an in-mem mock of the database? Should we run Postgres locally?

packages/phone-number-privacy/migrations/20200330212224_create-accounts-table.ts
```diff
@@ -0,0 +1,17 @@
...
+    t.integer('num_lookups').unsigned()
+    t.dateTime('did_matchmaking')
+  })
+}
+
+export async function down(knex: Knex): Promise<any> {
```
codyborn, line 15: Is it possible to limit the privileges of the application identity such that it doesn't have the ability to create/drop tables?

packages/phone-number-privacy/package.json
```diff
@@ -6,17 +6,23 @@
...
+    "deploy:staging": "firebase deploy --only functions --project celo-phone-number-privacy-stg",
+    "deploy:prod": "firebase deploy --only functions --project celo-phone-number-privacy",
+    "config:get:staging": "firebase functions:config:get --project celo-phone-number-privacy-stg",
+    "config:get:prod": "firebase functions:config:get --project celo-phone-number-privacy",
     "clean": "tsc -b . --clean",
     "build": "tsc -b .",
     "lint": "tslint --project .",
-    "test": "jest --runInBand"
+    "test": "jest --runInBand",
+    "db:migrate": "NODE_ENV=development ts-node scripts/runMigrations.ts"
```
aslawson, line 18: it looks like in dev mode, this migrates to `host: 'fakeHost'.  so does this even create anything?  if so which environment (locally vs stg vs prod)

packages/phone-number-privacy/src/index.ts
aslawson: I do think this would be nice -- perhaps there is existing tooling that integrates with knex which allows definition of the database schema in JSON with parser to generate the migrations and js wrappers.  otherwise could just create the wrappers manually

packages/phone-number-privacy/migrations/20200330212301_create-number-pairs-table.ts
```diff
@@ -0,0 +1,12 @@
...
+    t.string('user_phone_hash').notNullable()
+    t.string('contact_phone_hash').notNullable()
```
aslawson: do we need to tell postgres to index these or is that automatic?

packages/phone-number-privacy/migrations/20200330212301_create-number-pairs-table.ts
jmrossy: can do!

packages/phone-number-privacy/migrations/20200330212301_create-number-pairs-table.ts
jmrossy: I haven't addressed that yet, will do in a separate PR

packages/phone-number-privacy/package.json
jmrossy: Yeah you would have to populate those values with the ones from our staging DB but didn't want those checked in.

packages/phone-number-privacy/migrations/20200330212224_create-accounts-table.ts
jmrossy: Yep we could do that, makes sense

packages/phone-number-privacy/migrations/20200330212224_create-accounts-table.ts
jmrossy: Probably possible, I haven't looked into fine grained user controls but we could set up multiple DB users.
It's worth noting that these migrations are not run by the app but one time by us before deploying to app to prime the db

packages/phone-number-privacy/src/config.ts
jmrossy: I'm for in-mem db mock. WDYT?

packages/phone-number-privacy/src/config.ts
jmrossy: Not sure. Let's worry about robust error handling in a separate PR.

packages/phone-number-privacy/src/index.ts
jmrossy: I totally agree and do intend to create a `model` layer that abstracts away knex. I'll create a basic models layer now but not trying to fully address everything in this PR

packages/phone-number-privacy/src/config.ts
codyborn: I think that's probably best. The other option is to allow each of us to setup a dedicated postgres for unit testing and use an .env file for configuring the connection string. I think that the interface for us is straightforward enough for us to mock it properly and it forces us to think through the different errors that Postgres can throw.

packages/phone-number-privacy/migrations/20200330212224_create-accounts-table.ts
codyborn: Makes sense; thanks for clarifying. Let's see if we can create separate identities for the webapp and the deployment pipeline.

packages/phone-number-privacy/package.json
aslawson: kk makes sense

packages/phone-number-privacy/migrations/20200330212301_create-number-pairs-table.ts
aslawson: sounds good

aslawson (approved):

packages/phone-number-privacy/src/models/account.ts
```diff
@@ -0,0 +1,22 @@
...
+export const ACCOUNTS_TABLE = 'accounts'
+export enum ACCOUNTS_COLUMNS {
+  address = 'address',
+  createdAt = 'created_at',
+  numLookups = 'num_lookups',
+  didMatchmaking = 'did_matchmaking',
+}
+
+export class Account extends Model {
+  address: string
```
codyborn, line 12: Can we store user-defiend types? There exists an "Address" object already defined in the contrackit library that we can use.

codyborn (commented):
👍

PR #3237, opened 2020-03-31
Add note about expected state after setup to Baklava launch docs

nategraf:
### Description

This PR adds a note to clarify the expected state after setting up the validator infra for Baklava launch.

### Other Changes

* Change "dir" to "directory"
* Remove note about running the proxy on the same machine. We don't want people to do that.

### DIFF
--- a/packages/docs/getting-started/running-a-validator-in-baklava.md
+++ b/packages/docs/getting-started/running-a-validator-in-baklava.md
@@ -151,11 +151,15 @@ docker run --name celo-validator -it --restart always -p 30303:30303 -p 30303:30
 
 The `networkid` parameter value of `33120` indicates we are connecting to the new Baklava network.
 
-Note that if you are running the validator and the proxy on the same machine, then you should set the validator's listening port to something other than `30303`. (e.g. Use the Docker flags `-p 30313:30303` and `-p 30313:30303/udp` to export the node on port 30313 on the host).
+{% hint style="info" %}
+If you are running the validator and the proxy on the same machine, then you should set the validator's listening port to something other than `30303`. (e.g. Use the Docker flags `-p 30313:30303` and `-p 30313:30303/udp` to export the node on port 30313 on the host).
+{% endhint %}
+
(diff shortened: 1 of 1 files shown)


packages/docs/getting-started/running-a-validator-in-baklava.md
```diff
@@ -151,11 +151,15 @@ docker run --name celo-validator -it --restart always -p 30303:30303 -p 30303:30
 
 The `networkid` parameter value of `33120` indicates we are connecting to the new Baklava network.
 
-Note that if you are running the validator and the proxy on the same machine, then you should set the validator's listening port to something other than `30303`. (e.g. Use the Docker flags `-p 30313:30303` and `-p 30313:30303/udp` to export the node on port 30313 on the host).
+{% hint style="info" %}
+If you are running the validator and the proxy on the same machine, then you should set the validator's listening port to something other than `30303`. (e.g. Use the Docker flags `-p 30313:30303` and `-p 30313:30303/udp` to export the node on port 30313 on the host).
+{% endhint %}
+
+At this point your proxy should be peering with the bootnode and other nodes that come online. Your validator will not automatically peer with the proxy until the mining routine starts after the genesis timestamp on, so it will not have any peers. You should see a "Mining too far in the future" log message from the validator, which indicates it is waiting for the genesis timestamp to pass. This is the expected state until block production starts on April 6th at 1600 UTC.
```
tkporter: Small nit: we're still using the `cmd/bootnode` rather than a full node, so maybe it should be "should be peering with other nodes that come online through the bootnode". I played around with using a full node instead when setting up our tx nodes but it felt risky to change now

tkporter: Might be helpful to say something along the lines of "when that time is hit, there will be a few minute delay for the validators to identify each other and begin block production"

tkporter (approved):

packages/docs/getting-started/running-a-validator-in-baklava.md
nategraf: Thanks!

PR #3247, opened 2020-03-31
AKS prometheus deployment with celotool

### LINKED ISSUE #3246: Prometheus on AKS pushing to stackdriver
### Expected Behavior

AKS integration with celotool, the ability to deploy prometheus to AKS, and prometheus metrics being sent from AKS to Stackdriver

### Current Behavior

None

tkporter:
### Description

This includes azure integration with celotool and deployment commands for prometheus + the sidecar to send metrics to stackdriver on AKS. Down the road, once we have other azure deployment commands, I plan to make prometheus deployment a normal part of the cluster setup.

At a high level, this allows celotool to do some behind the scenes work to make sure kubectl is configured to use the proper AKS cluster for an environment. An example env would include:
```
AZURE_KUBERNETES_CLUSTER_NAME=nodes-aks
AZURE_KUBERNETES_CLUSTER_REGION=west-us-2
AZURE_KUBERNETES_RESOURCE_GROUP=cody-oracle-prototype
```

This also adds commands to deploy prometheus to a cluster in the `prometheus` namespace, as well as a sidecar to push metrics to stackdriver (using https://github.com/Stackdriver/stackdriver-prometheus-sidecar). It authenticates using a GCP service account that is created on a per-azure-cluster basis.

### Tested

Running `celotooljs deploy initial aks-prometheus` and `celotooljs deploy destroy aks-prometheus`

### Other changes

Refactored a little to separate GCP service account actions

### Related issues

- Fixes #3246

### Backwards compatibility

Yes


packages/helm-charts/prometheus-stackdriver/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for installing prometheus and the stackdriver-prometheus-sidecar
+name: oracle
```
jcortejoso: I find more clear a name related with prometheus (`prometheus-stackdiver` or similar)

packages/celotool/src/lib/azure.ts
```diff
@@ -0,0 +1,43 @@
...
+  } catch (error) {
+    console.info('No cluster currently set')
+  }
+
+  const clusterName = fetchEnv(envVar.AZURE_KUBERNETES_CLUSTER_NAME)
+  const resourceGroup = fetchEnv(envVar.AZURE_KUBERNETES_RESOURCE_GROUP)
+
+  if (currentCluster === null || currentCluster.trim() !== clusterName) {
+    await execCmdWithExitOnFailure(
+      `az aks get-credentials --resource-group ${resourceGroup} --name ${clusterName}`
```
jcortejoso: If the context is edited for some reason (for example switching the default namespace in the context), `az` shows a warning when you import again the credentials (`A different object named nodes-aks already exists in your kubeconfig file. Overwrite? (y/n):`). Maybe `--overwrite-existing` could be convenient here. What do you think?

jcortejoso (approved):
Awesome! Quite cool the integration of different cloud providers into celotool 👌

packages/helm-charts/prometheus-stackdriver/Chart.yaml
tkporter: Oops, copied it from another and forgot to change it. Thanks

packages/celotool/src/lib/azure.ts
tkporter: Nice catch

PR #3251, opened 2020-03-31
Slashing multiplier should impact validator rewards

### LINKED ISSUE #3241: Group and validator rewards don't factor in slashing multiplier
### Expected Behavior

PoS design doc states: 

> At the end of each epoch, provided that the validator and its group have the required minimum stake, Validators are rewarded Pv * Sv * M * (1 - C) Celo Dollars where C is group share for the group the validator was a member of when it was elected and Pv is the max payout to validators and is governable 
At the end of each epoch, for each validator that was elected, the group a validator was elected as a member of is rewarded Pv * Sv * C * M Celo Dollars 
where ....  M is the group’s current slashing penalty (M=1 initially, 0<M<=1)

### Current Behavior

Rewards do not factor in slashing multiplier.

mrsmkl:
### Description

Take slashing multiplier into account when computing validator and validator group rewards.

### Tested

Added a test case.

### Related issues

- Fixes #3241


asaj (approved):
👏

asaj:
Nice tests!

PR #3258, opened 2020-04-01
Increase precision of active vote units

### LINKED ISSUE #2564: Accessors for active vote units
### Expected Behavior

There is some way to access how many vote units each voter for a group has.
Because there is no limit on how many voters a group can have, this info will eventually be needed to compute the active vote values for voters in blockscout (that is, they too have to be computed on demand).

### Current Behavior

The information about units is private.

asaj:
### Description

This PR modifies the election active vote units to be stored with additional precision. This prevents attacks that specify an amount so small that it rounds down to zero units.

### Tested

Unit tests
### Other changes

None
### Related issues

- Fixes #2564
### Backwards compatibility

No


packages/contractkit/src/wrappers/Election.ts
```diff
@@ -302,7 +302,7 @@ export class ElectionWrapper extends BaseWrapper<Election> {
    * Activates any activatable pending votes.
    * @param account The account with pending votes to activate.
    */
-  async activate(account: Address): Promise<Array<CeloTransactionObject<boolean>>> {
+  async activate(account: Address): Promise<Array<CeloTransactionObject<string>>> {
```
mrsmkl: I guess these should be BigNumbers now.

packages/protocol/contracts/governance/Election.sol
```diff
@@ -453,6 +455,29 @@ contract Election is
     return pending.add(active);
   }
 
+  /**
+   * @notice Returns the active vote units for `group` made by `account`.
+   * @param group The address of the validator group.
+   * @param account The address of the voting account.
+   * @return The active vote units for `group` made by `account`.
+   */
+  function getActiveVoteUnitsForGroupByAccount(address group, address account)
```
mrsmkl, line 470: This should fix #2564

packages/protocol/contracts/governance/Election.sol
```diff
@@ -678,40 +716,64 @@ contract Election is
...
-  function getActiveVotesUnitsDelta(address group, uint256 value) private view returns (uint256) {
+  function votesToUnits(address group, uint256 value) private view returns (uint256) {
+    FixidityLib.Fraction memory fixedValue = FixidityLib.newFixed(value);
     if (votes.active.forGroup[group].total == 0) {
-      return value;
+      return fixedValue.unwrap();
+    } else {
+      return
+        fixedValue.multiply(FixidityLib.wrap(votes.active.forGroup[group].totalUnits)).unwrap().div(
+          votes.active.forGroup[group].total
```
mrsmkl: What is the difference between this and `value.mul(votes.active.forGroup[group].totalUnits).div(votes.active.forGroup[group].total)` ?
Anyway it looks like we are getting close to overflow, for example if we have 1e9 cGLD (1e27 wei), this gives 1e51 units, then when they are multiplied, we have 1e78 (overflow).

packages/protocol/contracts/governance/Election.sol
```diff
@@ -678,40 +716,64 @@ contract Election is
...
+  function unitsToVotes(address group, uint256 value) private view returns (uint256) {
+    FixidityLib.Fraction memory fixedValue = FixidityLib.wrap(value);
+    if (votes.active.forGroup[group].totalUnits == 0) {
+      return 0;
     } else {
       return
-        value.mul(votes.active.forGroup[group].totalUnits).div(votes.active.forGroup[group].total);
+        fixedValue.multiply(FixidityLib.newFixed(votes.active.forGroup[group].total)).unwrap().div(
+          votes.active.forGroup[group].totalUnits
+        );
```
mrsmkl: Since units have more precision than votes, perhaps we could round to nearest here?
I think that the effect for this would be that when votes are added, that would be exact (when epoch rewards are paid, I don't think that can be exact, clearly it's possible to get half a vote as an epoch reward).

packages/protocol/test/governance/election.ts
```diff
@@ -1704,4 +1739,297 @@ contract('Election', (accounts: string[]) => {
...
+            await Promise.all(voterAccounts.map(checkVoterInvariants))
+            await checkGroupInvariants(voterAccounts)
+            await mineBlocks(EPOCH, web3)
+          }
+        })
+      })
+
+      describe.only('when epoch rewards are distributed', () => {
+        it('actual and expected should always match within a small delta', async () => {
+          // TODO(asa): Failing here
```
mrsmkl: Have to read the votes from smart contract here...

mrsmkl (commented):
Worried about possible overflow, otherwise looks good.

packages/contractkit/src/wrappers/Election.ts
asaj: Leftover from a previous change

packages/protocol/contracts/governance/Election.sol
asaj: Good catch, reduced to precision to 1e20

packages/protocol/contracts/governance/Election.sol
asaj: Sorry I'm not totally clear on what you're suggesting?

mrsmkl (approved):

PR #3261, opened 2020-04-01
Changed celostats main domain to celostats-frontend

jcortejoso:
### Description

Changed the default dns https://{env}-celostats.{dns-domain} to the celostats-frontend service.

### Tested

Tested on rc0 environment


timmoreton (approved):
👍

PR #3266, opened 2020-04-01
Adjust migrationsConfig for RC1

asaj:
### Description

Slow down governance slightly, increase min electable validators.

### Tested

No


packages/protocol/migrationsConfig.js
```diff
@@ -188,7 +190,8 @@ const DefaultConfig = {
       exponent: 10,
       adjustmentSpeed: 0.1,
     },
-    membershipHistoryLength: 60,
+    // MUST BE KEPT IN SYNC WITH GROUP LOCKED GOLD DURATION
+    membershipHistoryLength: 180,
```
aslawson: "_ membershipHistoryLength  The max number of entries for validator membership history."

why does this have to be as long as locked gold duration?

aslawson: I'm unclear what it means by entries

asaj: Yeah you're totally right, oops.

packages/protocol/migrationsConfig.js
```diff
@@ -64,6 +64,7 @@ const DefaultConfig = {
     targetVotingGoldFraction: 2 / 3,
     maxValidatorEpochPayment: '205479452054794520547', // (75,000 / 365) * 10 ^ 18
     communityRewardFraction: 1 / 4,
+    // TODO(asa): Must be set before RC1
```
aslawson, line 79: There shouldn't be any urgency in setting this before RC1.  If its 0, no transaction will occur
https://github.com/celo-org/celo-blockchain/blob/dfd70edf41618d6b40aafd695087d0cf919e0f92/consensus/istanbul/backend/pos.go#L119

packages/protocol/migrationsConfig.js
```diff
@@ -84,14 +85,14 @@ const DefaultConfig = {
...
-    queueExpiry: WEEK, // Change to 4 weeks once mainnet activated
-    dequeueFrequency: MINUTE, // Change to 1 week once mainnet activated
-    concurrentProposals: 3, // Change to 10 once mainnet activated
-    approvalStageDuration: 30 * MINUTE, // Change to 3 days once mainnet activated
-    referendumStageDuration: HOUR, // Change to 1 week once mainnet activated
-    executionStageDuration: WEEK,
+    queueExpiry: 4 * WEEK,
+    dequeueFrequency: DAY, // Change to 1 week once mainnet activated
+    concurrentProposals: 3,
+    approvalStageDuration: DAY, // Change to 3 days once mainnet activated
```
aslawson, line 103: one thing that we may consider is allowing the proposal to move forward immediately to referendum once approved by the required number of signers

aslawson: same with referendum -- once it hits the required number of votes and gold value just move it forward (unless we think people are likely to change/retract their votes).  though this won't work if you change the participateBaseline to 0

packages/protocol/migrationsConfig.js
```diff
@@ -84,14 +85,14 @@ const DefaultConfig = {
...
+    referendumStageDuration: 2 * DAY, // Change to 1 week once mainnet activated
+    executionStageDuration: 3 * DAY,
     minDeposit: 100, // 100 cGLD
-    participationBaseline: 8 / 10,
+    participationBaseline: 0, // Start with no participation requirements, let the protocol adjust
```
aslawson: do you need to update the participationBaselineFloor as well?  looks like it will get set to the floor otherwise

https://github.com/celo-org/celo-monorepo/blob/8df0872d6b4f9468f0422a8e3fcfae9e7bf64f41/packages/protocol/contracts/governance/Governance.sol#L1159

packages/protocol/migrationsConfig.js
asaj: Nice

packages/protocol/migrationsConfig.js
asaj: > one thing that we may consider is allowing the proposal to move forward immediately to referendum once approved by the required number of signers

This was a deliberate choice in order to keep a predictable governance cadence

>  once it hits the required number of votes and gold value

What do you mean by this?

nategraf (approved):

packages/protocol/migrationsConfig.js
aslawson: i wonder how valuable that is to keep the approval stage for full length though.  you can still keep the cadence of how many get dequeued.  so the overall rate of approvals through the pipeline are still consistent.

> once it hits the required number of votes and gold value

I just mean if we hit a threshold of x/totalGold has voted yes, you can also accelerate to the next phase early

asaj: > you can still keep the cadence of how many get dequeued
But you'd lose the cadence of when folks need to vote, no?

packages/contractkit/package.json
```diff
@@ -21,7 +21,7 @@
     "build:gen": "yarn --cwd ../protocol build",
     "prepublishOnly": "yarn build:gen && yarn build",
     "docs": "typedoc && ts-node scripts/linkdocs.ts",
-    "test:reset": "yarn --cwd ../protocol devchain generate-tar .tmp/devchain.tar.gz --migration_override ../dev-utils/src/migration-override.json --upto 24",
+    "test:reset": "yarn --cwd ../protocol devchain generate-tar .tmp/devchain.tar.gz --migration_override ../dev-utils/src/migration-override.json --upto 25",
```
asaj:
```suggestion
    "test:reset": "yarn --cwd ../protocol devchain generate-tar .tmp/devchain.tar.gz --migration_override ../dev-utils/src/migration-override.json --upto 24",
```

PR #3281, opened 2020-04-02
[N04] MockReserve.sol has mintToken function

asaj:
### Description

The MockReserve contract has a mintToken function, while the contract it is based on, Reserve, does not.

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/439


mrsmkl (approved):
LGTM

PR #3282, opened 2020-04-02
[L13] Duplicated code in updatePublicKeys

asaj:
### Description

Refactor to reduce duplicated code.
### Tested

Unit tests
### Other changes

Moved `updatePublicKeys` down.

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/440

### Backwards compatibility

Yes


mrsmkl (approved):
Looks like these methods can only return `true`, perhaps the return value can be removed.

Issue #3078, opened 2020-03-11
Adversarial testing of stability protocol


timmoreton:
Create a table with columns described below and fill it with all attacks you can think of.

Table columns:
- attack category
- attack
- how easy attack is - CVSS score vector-  https://chandanbn.github.io/cvss/ )
- impact of attack (service availability, time, cost, reputation, etc)
- metric(s) by which this attack could be identified and by which an alert could be triggered
- remediation time 
- possible mitigations (and cost/effort of each)

jose-compu:
Added a possible attack (non-critical) and a solution:

**Testing timining attack with no bucket updates. https://github.com/celo-org/celo-monorepo/pull/3283**

diminator:
spreadsheet with possible attacks and network conditions ranked according to likelihood and impact:
https://docs.google.com/spreadsheets/d/1OjX6ahXjv3E89C3HNRmmYxHos2Ft9CNpDZk9fiipJTk/

PR #3283, opened 2020-04-02
Testing timining attack with no bucket updates.

jose-compu:
### Description

Not critical attack, that allows an attacker in some scenarios to avoid all bucket updated, leaving CP-DOTO just working as Uniswap with current buckets.

The proposed solution is to have `oracles.reportExpiry <= exchange.updateFreq` (now is the opposite).

Detailed description:

https://docs.google.com/document/d/1ePrxzpU7KclQTHzJbKDZR1GomSFnTeKCoceUJFHgFOo/edit#

### Tested

To be tested with new unit test.

### Other changes

Proposal: The solution condition can be also checked on chain on `setReportExpiry()` and `setUpdateFrequency()`, but probably is overkill because is configured initially and by governance, not something very frequent.

### Related issues

- Helps fixing #3078

### Backwards compatibility

Not smart contract were changed, only a unit test added.


H34D (approved):

aaronmgdr:
merge?

PR #3284, opened 2020-04-02
Fix incorrect line numbers in jest tests

### LINKED ISSUE #2961: jest tests don't show the correct line numbers for failing tests
### Expected Behavior

Jest tests should show the correct line numbers for failing tests.

### Current Behavior

Error is at a specific line with an `expect(...)` statement, but jest shows an incorrect line:

Screenshot 2020-02-26 at 10 27 56

jeanregisser:
### Description

Use ganache-core fork which bypasses the import of `source-map-support/register` during tests.

See:
- https://github.com/trufflesuite/ganache-core/issues/522
- https://github.com/kulshekhar/ts-jest/issues/727
- https://github.com/celo-org/ganache-core/commit/3dbc32b214b3bb8f0dda2bfd909850a752a73a55

### Tested

Failed jest tests now reports correct line numbers for failed expectations.

### Other changes

N/A

### Related issues

- Fixes #2961

### Backwards compatibility

Yes.


jeanregisser:
Some additional info:
I noticed the the failing expectations showing incorrect line numbers in the `mobile` project started to show correct line numbers since we deleted `walletkit`.
I tried on the `blockchain-api` project and got incorrect line numbers without this fix and removing the mock on `contractkit` (so `ganache-core` ends up being indirectly imported).

To sum up, this fix is needed whenever `ganache-core` ends up being imported during a jest test.

package.json
```diff
@@ -77,6 +77,7 @@
     "sha3": "1.2.3",
     "bignumber.js": "9.0.0",
     "prettier": "^1.19.1",
-    "@types/prettier": "^1.19.0"
+    "@types/prettier": "^1.19.0",
+    "ganache-core": "https://github.com/celo-org/ganache-core.git#3dbc32b"
```
jmrossy, line 81: I have a slight preference for using the github link inline where we declare a dep on ganache-core instead of overriding here at the root. It might confused devs who try to upgrade it. Consider using this link throughout the monorepo or moving the ganache-core dep up here to the root like we did for typescript

jeanregisser: Yes I'd prefer that too, but `ganache-core` is pulled by `@0x/subproviders`, so we'd need to fork that too.

```txt
❯ yarn why "ganache-core"
=> Found "ganache-core@2.7.0"
info Reasons this module exists
   - "_project_#@celo#protocol#@0x#subproviders" depends on it
   - Hoisted from "_project_#@celo#protocol#@0x#subproviders#ganache-core"
```

jmrossy: Ok thanks for clarifying

jmrossy (approved):

gastonponti:
I believe that the `0x/subproviders` in the `protocol/package.json` is not used anymore.
But other `0x` packages that we use, end up retrieving the `0x/subproviders`. We should probably make an issue to clean old unused dependencies, and see which ones could be avoided if we just extract some code. 

But this PR looks ok, nice catch!
LGTM

gastonponti (approved):

PR #3286, opened 2020-04-02
[Wallet] Update contactSyncBanner in design polish session

annakaz:
### Description

Fixed some bugs in the ContactSyncBanner during the weekly design polish session with @coreycelo . 

Currently the ContactSyncBanner scrolls with the list of contacts and appears for three seconds every time a user sends a new transaction, even when all contacts are synced. Now the Banner remains fixed below the search bar as users scroll through their contacts, and it is only shown when a user is actively importing their contacts

### Tested

On a pixel 2 before and after importing contacts

### Backwards compatibility

Yes


packages/mobile/src/send/ContactSyncBanner.tsx
```diff
@@ -24,6 +24,11 @@ export function ContactSyncBanner() {
     if (!isSynced) {
       return
     }
+    // Hide banner immediately if not loading contacts
+    if (!isLoadingContacts) {
```
jmrossy: I don't understand how this would work. When the screen loads, we haven't yet started loading, so it would be false, and we would then set hasSynced to true, meaning we wouldn't keep this banner shown for 3 seconds after syncing is done right? Has that requirement changed? If so let's remove the timer stuff and simplify this :)

annakaz: Right now, the banner is shown for 3 seconds after syncing is done, but this includes when a user has already imported their contacts. So every time someone opens this recipient picker, they see an "x of x contacts synced" for 3 seconds even though the contacts were previously imported. I think we didn't notice this bug previously since the syncing banner was less noticeable as it was hidden with the contacts scrolling

With this change, we still keep the banner up for 3 seconds if a user is actively importing their contacts, so we still need the timer. I tested that when importing contacts this banner will still appear for 3 seconds- is this different from the behavior you would expect?

jmrossy: I'm struggling a bit to see how the logic leads to the banner hanging around for 3 extra seconds but if it works then we're good 🤷‍♂

jmrossy (approved):

annakaz:
Going to merge this in without the banner timeout changes as I was testing it by importing contacts via invite, since verification is failing, but have not tested with the normal contact import path

PR #3292, opened 2020-04-02
[phoneNumberPrivacy] add ratelimiting logic with db connection [unit tests in progress]

aslawson:
### Description

- enforces ratelimiting on phone number salt generation
- query quota is calculated via: unverifiedQueryCount + verifiedQueryCount + (queryPerTransaction * transactionCount)
- number of queries already performed is stored in the `accounts` table in the DB as `num_lookups`
- on request, the `num_lookups` is incremented
- **on failure of db table right now, the error is just logged and the salt would still be returned** -- up for discussion on how we want to handle this

- currently pointing to `https://alfajores-forno.celo-testnet.org` for attestation contract interaction and `https://alfajores-blockscout.celo-testnet.org` for transaction counts.  this may change if we determine these services do not have high enough reliability for our use case.

### Tested

Heavily tested manually.  Unit Tests in progress

`curl -v "http://localhost:5000/celo-phone-number-privacy/us-central1/getSalt" -d '{"queryPhoneNumber": "5555555555", "account":"0x117ea45d497ab022b85494ba3ab6f52969bf6831", "phoneNumber":"+5555555555"}' -H 'Content-Type: application/json'`

### Other changes

NA

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/330

### Backwards compatibility

NA -- Service not used yet


packages/phone-number-privacy/package.json
```diff
@@ -18,6 +18,7 @@
     "db:migrate": "NODE_ENV=development ts-node scripts/runMigrations.ts"
   },
   "dependencies": {
+    "@celo/contractkit": "0.3.2-dev",
```
jmrossy: We can only use published versions in cloud functions, so this needs to be 0.3.1

packages/phone-number-privacy/src/config.ts
```diff
@@ -20,8 +27,16 @@ let config: Config
 if (DEV_MODE) {
   console.debug('Running in dev mode')
   config = {
+    blockchain: {
+      // TODO [amyslawson] figure out where to point these
```
jmrossy: These looks right to me for now :)

packages/phone-number-privacy/src/config.ts
```diff
@@ -4,8 +4,15 @@ import knex from 'knex'
 const DEV_MODE = process.env.NODE_ENV !== 'production' || process.env.FUNCTIONS_EMULATOR === 'true'
 
 interface Config {
+  blockchain: {
+    provider: string
+    blockscout: string
+  }
   salt: {
     key: string
+    unverifiedQueryCount: number
```
jmrossy: Consider renaming to `unverifiedQueryMax` and ditto below

packages/phone-number-privacy/src/web3/contracts.ts
```diff
@@ -0,0 +1,10 @@
+import { newKitFromWeb3 } from '@celo/contractkit'
+import Web3 from 'web3'
+import config from '../config'
+
+const web3: Web3 = new Web3(new Web3.providers.HttpProvider(config.blockchain.provider))
```
jmrossy: Given that we don't explicitly declare web3 as a dep of this project, it'd be better to let CK handle the interactions with web. Consider using `newKit` instead of `newKitFromWeb3`

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+ */
+function confirmUser() {
+  //TODO [amyslawson]
+  return
+}
+
+/*
+ * Computes the BLS Salt for the blinded phone number.
+ */
+function computeBLSSalt(queryPhoneNumber: string, response: functions.Response) {
```
jmrossy: Nit: I'd recommend we move all code out of index.ts except the API endpoints, which is a common structure for cloud functions I believe

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
-
-    const salt = BLINDBLS.computePRF(privateKey, new Buffer(request.body.blindPhoneNumber))
-    response.json({ success: true, salt })
+    const privateKey = new Buffer(config.salt.key)
+    return BLINDBLS.computePRF(privateKey, new Buffer(queryPhoneNumber))
   } catch (e) {
-    console.error('Failed to compute BLS salt', e)
-    response.status(500).send('Failed to compute BLS salt')
+    console.error('Failed to compute salt', e)
+    response.status(500).send('Failed to compute salt')
```
jmrossy: Don't handle the response directly here. Just let this function worry about BLS and then have the controllers layer (getSalt above) worry about dealing with HTTP codes and such.

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+
+export const NUM_ATTESTATIONS_REQUIRED = 3
+
+const contractKit = getContractKit()
+const knex = connectToDatabase()
 
 export const getSalt = functions.https.onRequest((request, response) => {
-  const privateKey = new Buffer(config.salt.key)
+  confirmUser()
+  Promise.resolve(getRemainingQueryCount(request.body.account, request.body.phoneNumber)).then(
```
jmrossy: Any reason we're using raw promises here instead of making the getSalt handler async?

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+    (remainingQueryCount) => {
+      if (remainingQueryCount <= 0) {
+        response.status(400).send('Requester exceeded salt service query quota')
+        return
+      }
+      const salt = computeBLSSalt(request.body.queryPhoneNumber, response)
+      Promise.resolve(incrementQueryCount(request.body.account))
```
jmrossy: ditto

jmrossy (commented):
Great progress! See comments below :)
Overall looks good but I think we could structure the code better to avoid having everything live in index.ts. Maybe you were already planning to do that in a follow-up PR, not sure.

packages/phone-number-privacy/src/index.ts
aslawson: good point -- I was just trying to get something functional out first, but starting to refactor may make the tests cleaner as well

packages/phone-number-privacy/src/index.ts
aslawson: I blame it on my lack of experience with typescript & firebase.  I didn't think it was possible with the firebase functions, but looks like it is
https://www.youtube.com/watch?v=Jr7pDZ1RAUg

I'll update -- that will clean up the code a lot

packages/phone-number-privacy/src/config.ts
```diff
@@ -33,8 +48,15 @@ if (DEV_MODE) {
...
   const functionConfig = functions.config()
   config = {
+    blockchain: {
+      provider: functionConfig.blockchain.provider,
+      blockscout: functionConfig.blockchain.blockscout,
+    },
     salt: {
       key: functionConfig.salt.key,
+      unverifiedQueryCount: functionConfig.salt.unverifiedQueryCount,
+      verifiedQueryCount: functionConfig.salt.verifiedQueryCount,
```
codyborn: Without looking at the code, I would assume that if the user is verified, we would go with the verifiedQueryCount and not (unverifiedQueryCount + verifiedQueryCount). We can rename it to "additionalVerifiedQueryCount" or remove the +unverifiedQueryCount to make it more intuitive.

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+  const queryQuota = await getQueryQuota(account, phoneNumber)
+  const performedQueryCount = await getPerformedQueryCount(account)
+  return queryQuota - performedQueryCount
+}
+
+/*
+ * Calculates how many queries the caller has unlocked based on the algorithm
+ * unverifiedQueryCount + verifiedQueryCount + (queryPerTransaction * transactionCount)
+ */
+async function getQueryQuota(account: string, phoneNumber: string) {
```
codyborn: This logic is going to grow quickly. We should create some classes to make this more extensible in the future. For example, each of these queryQuota changes can be generalized as a `QuotaRule` and we can use `QuotaRuleAggregator` to call each rule and aggregate them. Today we have 3 rules (unverified, verified, txCount). Each of these rules can be calculated in parallel and all add some number to the end queryQuota. I imagine we'll want to add more `QuotaRule` types in the future, but the aggregation logic shouldn't change.

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+  const numAttestationsRemaining = NUM_ATTESTATIONS_REQUIRED - numAttestationsCompleted
+  return numAttestationsRemaining <= 0
+}
+
+async function getTransactionCountFromAccount(account: string): Promise<number> {
+  // TODO [amyslawson] consider adding retry
+  const transactionUrl = `${config.blockchain.blockscout}/api?module=account&action=tokentx&address=${account}`
+  const resp = await fetch(transactionUrl)
+  const jsonResp = await resp.json()
+  if (jsonResp == null || jsonResp.result == null) {
```
codyborn: What is the max size of this response? We'll need to paginate processing if this grows linearly with number of txs.

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+  const attestationsWrapper: AttestationsWrapper = await contractKit.contracts.getAttestations()
+  const attestationStats = await attestationsWrapper.getAttestationStat(phoneNumber, account)
+  const numAttestationsCompleted = attestationStats.completed
+  const numAttestationsRemaining = NUM_ATTESTATIONS_REQUIRED - numAttestationsCompleted
+  return numAttestationsRemaining <= 0
+}
+
+async function getTransactionCountFromAccount(account: string): Promise<number> {
+  // TODO [amyslawson] consider adding retry
+  const transactionUrl = `${config.blockchain.blockscout}/api?module=account&action=tokentx&address=${account}`
```
codyborn: Using blockscout doesn't feel like a good long term solution. I'm concerned about SLA and security of the blockscout instance. The more services that we put in our critical path, the lower our SLA will be. I think a better short-term approach is to query the latest nonce for a given account from the full-node that we're running.

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+  return knex(ACCOUNTS_TABLE)
+    .where(ACCOUNTS_COLUMNS.address, account)
+    .select(ACCOUNTS_COLUMNS.numLookups)
+    .first()
+    .then((object) => {
+      return object === undefined ? 0 : object[ACCOUNTS_COLUMNS.numLookups]
+    })
+    .catch((reason) => {
+      // TODO [amyslawson] think of failure case here
+      console.error(reason)
```
codyborn: AFAIK we're planning on using Prometheus and Stackdriver across all CLabs services for monitoring. Does it make sense to integrate now so we can start logging these errors?

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+    console.error(`Failed to get valid response for ${transactionUrl}`)
+    return 0
+  }
+  return jsonResp.result.filter((transaction: any) => transaction.from === account).length
+}
+
+/*
+ * Returns how many queries the account has already performed.
+ */
+async function getPerformedQueryCount(account: string): Promise<number> {
```
codyborn: We should create a new class for wrapping the knex object and performing these high level operations. The more internals we expose, the harder it will be to guarantee our assumptions. Having a separate class will also make it easier to mock during testing. I recommend creating a new class (ex. `QueryStateStore`) that exposes `getPerformedQueryCount` and `incrementQueryCount`.

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
...
+      // TODO [amyslawson] think of failure case here
+      console.error(reason)
+      return 0
+    })
+}
+
+/*
+ * Increments query count in database.
+ */
+async function incrementQueryCount(account: string) {
```
codyborn: The query and increment logic should be atomic. For example, if I hit two servers simultaneously, both query my count as 32 and increment my count by 1, then my count will reach 34, which breaks the rules.

Some ideas to fix this:
1. Use serializable transactions. This will ensure that only one thread can read and modify the tx at a time. The second update will fail which is desired in our scenario.
See "SERIALIZABLE transactions" [here](https://www.2ndquadrant.com/en/blog/postgresql-anti-patterns-read-modify-write-cycles/).
2. Pass in the read value ('expectedLookups') into incrementQueryCount and fail if it's different than expected.
```
knex(ACCOUNTS_TABLE)
      .where(ACCOUNTS_COLUMNS.address, account)
      .where(ACCOUNTS_COLUMNS.numLookups, expectedLookups)
      .increment(ACCOUNTS_COLUMNS.numLookups, 1)
```
See "Optimistic concurrency controls" in the linked article.
3. Increment first, then read
This has the negative side-effect of potentially blocking a user too soon. However, if a user is spamming us perhaps they deserve to be blocked 😝

packages/phone-number-privacy/src/index.ts
codyborn: We should perform thorough validation of any content that's passed to us from an untrusted client.

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,148 @@
+import { AttestationsWrapper } from '@celo/contractkit/lib/wrappers/Attestations'
 import { BLINDBLS } from 'bls12377js-blind'
 import * as functions from 'firebase-functions'
+import fetch from 'node-fetch'
 import config, { connectToDatabase } from './config'
+import { ACCOUNTS_COLUMNS, ACCOUNTS_TABLE } from './models/account'
+import { getContractKit } from './web3/contracts'
+
+export const NUM_ATTESTATIONS_REQUIRED = 3
```
codyborn: Should this be configurable?

packages/phone-number-privacy/src/index.ts
jmrossy: These logs do go to stackdriver. Probably makes sense to hold off on any further work on logging + monitoring til we have the final infra setup planned out.

packages/phone-number-privacy/src/index.ts
jmrossy: Users can set arbitrary nonces in txs so long as it's greater than the previous one, right? . I think one way or another we need a reliable count on sent txs. Lets discuss it in our next sync

packages/phone-number-privacy/src/config.ts
aslawson: sure -- sounds good

packages/phone-number-privacy/src/index.ts
aslawson: interesting -- I was not aware that nonce can be set arbitrarily

packages/phone-number-privacy/src/index.ts
aslawson: its not in the mobile application, and I don't expect it to change much.

packages/phone-number-privacy/src/index.ts
aslawson: agreed

packages/phone-number-privacy/src/index.ts
codyborn: A user can set an arbitrary nonce, but the tx won't get mined unless it's +1 from the previous nonce. This is to ensure that someone doesn't censor a tx and lead you to taking actions that you didn't intend. There are two ways that we can query the RPC API for next nonce. One method for only taking into account on-chain txs and another method uses nonces in the mem-pool. For the latter method, I'm not sure if there's validation that the txs are strictly in order (non-skipping), however the former method definitely does.

packages/phone-number-privacy/src/index.ts
aslawson: this is a good point.
1. it seems we can pass this setting in via a raw query.  it says it lasts the current session (which I would think it just the full firebase request that creates the connection to the DB)
https://stackoverflow.com/questions/28181998/knex-js-postgresql-how-to-change-transaction-isolation-level
2. the only complication here with knex is that i'm already doing an update OR create, so if the update fails, its assuming its because the entry doesn't exist.  if I add this additional condition, it will also try to perform the create (which would also fail since it exists, but its just wasted requests)
3. ya -- we could do this and would only be an issue if there's an error later that prevents the user from getting the salt (we've essentially wasted their query)

i'm not sure how to test -- particularly the first one would be hard to tell if its been implemented correctly

aslawson: I added a sleep to simulate a race condition.  With adding
`  await knex.raw('BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;')`

it too also will fail on the update and then try the insert.  there really is no way to tell the difference between a failed update due to an entry nonexistence vs rejection to race condition.  the db returns just "0" to represent 0 rows updated rather than throw a parsable error :/

oddly the second request to "insert" then hangs and requires adding a timeout.

lets discuss in our next meeting

packages/phone-number-privacy/src/index.ts
aslawson: I like this idea as an improvement.  I'm going to leave as is for now due to time constraints and given that the current rules aren't exclusive -- we don't need to get all transactions if the user isn't even verified.

packages/phone-number-privacy/src/index.ts
aslawson: good point -- i'll wait until we discuss removing blockscout

packages/phone-number-privacy/src/common/identity.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Confirms that user is who they say they are and throws error on failure to confirm.
+ */
+export function confirmUser() {
```
jmrossy: `authenticateUser` may be a more descriptive and common name for this

packages/phone-number-privacy/src/database/database.ts
```diff
@@ -0,0 +1,11 @@
+import { connectToDatabase } from '../config'
+
+const knex = connectToDatabase()
+
+export function getDatabase() {
+  return knex
+}
+
+export async function setSerializable() {
+  await knex.raw('BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;')
```
jmrossy: what does this do?

packages/phone-number-privacy/src/database/database.ts
```diff
@@ -0,0 +1,11 @@
+import { connectToDatabase } from '../config'
+
+const knex = connectToDatabase()
```
jmrossy: Maybe move this function here now that we have a dedicated database file :)

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,55 @@
...
-export const getSalt = functions.https.onRequest((request, response) => {
-  const privateKey = new Buffer(config.salt.key)
-  try {
-    // Adding this here as an example of how to connect to the DB
-    const knex = connectToDatabase()
-    knex('accounts')
-      .first()
-      .then((val) => console.debug('account data', val))
-      .catch((reason) => console.error(reason))
+export const NUM_ATTESTATIONS_REQUIRED = 3
```
jmrossy: Consider moving to config.ts

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,21 +1,55 @@
...
+    const remainingQueryCount = await getRemainingQueryCount(
+      request.body.account,
+      request.body.phoneNumber
+    )
+    if (remainingQueryCount <= 0) {
+      response.status(400).send('Requester exceeded salt service query quota')
+      return
+    }
+    const salt = computeBLSSalt(request.body.queryPhoneNumber)
+    await incrementQueryCount(request.body.account).catch((error) => {
```
jmrossy, line 25: Should be no need for a `.catch` here given the outer try/catch in this function

packages/phone-number-privacy/src/salt-generation/query-quota.ts
```diff
@@ -0,0 +1,56 @@
+import { AttestationsWrapper } from '@celo/contractkit/lib/wrappers/Attestations'
+import fetch from 'node-fetch'
+import config from '../config'
+import { getPerformedQueryCount } from '../database/wrappers/account'
+import { getContractKit } from '../web3/contracts'
+
+const NUM_ATTESTATIONS_REQUIRED = 3
```
jmrossy: ditto, move to config and don't duplicate it

packages/phone-number-privacy/src/salt-generation/query-quota.ts
```diff
@@ -0,0 +1,56 @@
...
+const contractKit = getContractKit()
```
jmrossy: why create a module-scoped CK var here instead of calling getCK() where you need it?

packages/phone-number-privacy/src/salt-generation/query-quota.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/*
+ * Returns how many queries the account can make based on the
+ * calculated query quota and the number of queries already performed.
+ */
+export async function getRemainingQueryCount(account: string, phoneNumber: string) {
+  const queryQuota = await getQueryQuota(account, phoneNumber)
+  const performedQueryCount = await getPerformedQueryCount(account).catch((reason) => {
```
jmrossy: Similar to index, lets not mix promises and async/await.
If you're having to add the .catch to make the linter happy it means something else is wrong. `await` should already unwrap the promise

packages/phone-number-privacy/src/salt-generation/query-quota.ts
```diff
@@ -0,0 +1,56 @@
...
+  })
+  return queryQuota - performedQueryCount
+}
+
+/*
+ * Calculates how many queries the caller has unlocked based on the algorithm
+ * unverifiedQueryCount + verifiedQueryCount + (queryPerTransaction * transactionCount)
+ */
+async function getQueryQuota(account: string, phoneNumber: string) {
+  let queryQuota: number = config.salt.unverifiedQueryMax
```
jmrossy: nit, no need for explicit `: number` typing here right? TS should already know it

packages/phone-number-privacy/src/salt-generation/query-quota.ts
```diff
@@ -0,0 +1,56 @@
...
+  return numAttestationsRemaining <= 0
+}
+
+async function getTransactionCountFromAccount(account: string): Promise<number> {
+  // TODO [amyslawson] consider adding retry
+  const transactionUrl = `${config.blockchain.blockscout}/api?module=account&action=tokentx&address=${account}`
+  const resp = await fetch(transactionUrl)
+  const jsonResp = await resp.json()
+  if (jsonResp == null || jsonResp.result == null) {
+    console.error(`Failed to get valid response for ${transactionUrl}`)
```
jmrossy: maybe we should `throw here`. Not an urgent concern at all and we can make a pass at good error handling throughout the app in a later PR

jmrossy (commented):
Almost ready to merge! See comments below :)

packages/phone-number-privacy/src/salt-generation/query-quota.ts
aslawson: oh it was unused in the other location... but i'll put in config anyway

packages/phone-number-privacy/src/index.ts
aslawson: I suppose -- I was leaving it open in case we wanted to handle this differently.  right now if the db request fails we allow the query to be made.  if i allow the try/catch to be used it would return a 500.  it depends what we how we want to handle these errors

packages/phone-number-privacy/src/database/database.ts
aslawson: handle concurrent requests (see discussion with @codyborn above)

"This level emulates serial transaction execution, as if transactions had been executed one after another, serially, rather than concurrently."

https://www.postgresql.org/docs/current/sql-set-transaction.html
^ i says it should send an error, although in practice I just see it fail on update and return a 0 (as in 0 rows updated)

packages/phone-number-privacy/src/salt-generation/query-quota.ts
aslawson: copy pasta in the refactor -- will update

packages/phone-number-privacy/src/salt-generation/query-quota.ts
aslawson: i was originally part of the db request handling .then()/.catch() -- but I didn't want to handle it in wrapper of the account table itself since I felt error handling for the db request should be handled on a per-case basis.  what we want to do when there is a db error is circumstantial.

here it catches it and allows the transaction

packages/phone-number-privacy/src/salt-generation/query-quota.ts
aslawson: same thing -- it depends really if we want to refuse the user to query if there is a failure on our end.  if we want to avoid the potential of being a bottleneck to transactions, we'd want to error on allowing.  but if the risk for security is too high then we'd want to refuse the request

packages/phone-number-privacy/src/index.ts
aslawson: removed blockscout

packages/phone-number-privacy/src/index.ts
aslawson: removed blockscout

packages/phone-number-privacy/src/salt-generation/query-quota.ts
aslawson: removed

packages/phone-number-privacy/src/database/database.ts
```diff
@@ -0,0 +1,21 @@
...
+export const connectToDatabase = () => {
+  console.debug('Creating knex instance')
+  return knex({
+    client: 'pg',
+    connection: config.db,
+    debug: DEV_MODE,
+  })
+}
+
+const db = connectToDatabase()
```
jmrossy: Nit, no need for the connectToDatabase wrapper function at this point, may as well just assign this to the `knex`object.
If we find later that we should defer the knex object creation we can initiate it in getDatabase()

packages/phone-number-privacy/test/index.test.ts
```diff
@@ -1,76 +1,124 @@
...
+jest.mock('../src/salt-generation/query-quota')
+const mockReaminingQueryCount = queryQuota.getRemainingQueryCount as jest.Mock
+mockReaminingQueryCount.mockResolvedValue(2)
+
+jest.mock('../src/salt-generation/bls-salt')
+const mockBlsSalt = computeBLSSalt as jest.Mock
+mockBlsSalt.mockReturnValue(BLS_SALT)
+
+jest.mock('../src/database/wrappers/account', () => ({
+  incrementQueryCount: jest.fn().mockReturnValue(new Promise((resolve) => resolve())),
```
jmrossy, line 22: Nit a more common form for this would be  would be 
`jest.fn( () => {new Promise((resolve) => resolve())}),`

Also why the promise here? Keep in mind `await` can used on non-promises, Js handles it fine :)

packages/phone-number-privacy/test/salt-generation/query-quota.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+jest.mock('../../src/database/wrappers/account')
+const mockPerformedQueryCount = getPerformedQueryCount as jest.Mock
+
+// TODO (amyslawson) figure out how to change this mock for different unit tests
+// jest.mock('@celo/contractkit')
+// import { newKit } from '@celo/contractkit'
+// const mockNewKit = newKit as jest.Mock
+jest.mock('@celo/contractkit', () => {
+  const utils = require('../utils')
```
jmrossy, line 12: why do you import utils here via require instead of just `import`ing it at the top?

packages/phone-number-privacy/test/salt-generation/query-quota.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const ACCOUNT = '0x78dc5D2D739606d31509C31d654056A45185ECb6'
+const PHONE_NUMBER = '+1234567890'
+const queryQuota: QueryQuota = new QueryQuota()
+
+jest.mock('../../src/database/wrappers/account')
+const mockPerformedQueryCount = getPerformedQueryCount as jest.Mock
+
+// TODO (amyslawson) figure out how to change this mock for different unit tests
+// jest.mock('@celo/contractkit')
```
jmrossy: Maybe clean up the comments here a bit unless we need these?

jmrossy (approved):
Looks good! A few small comments but nothing major. 
Let's 🚢 it and keep building

packages/phone-number-privacy/test/salt-generation/query-quota.test.ts
aslawson: jest cannot find it otherwise...  apparently anything within the mock context needs to be defined there

packages/phone-number-privacy/test/index.test.ts
aslawson: RE: promises -- i didn't find that to always be the case -- if I had a .catch() after the promise call, then jest would complain that just passing the value back didn't have a .catch() method

packages/phone-number-privacy/test/salt-generation/query-quota.test.ts
aslawson: k

packages/phone-number-privacy/src/database/database.ts
aslawson: kk

packages/phone-number-privacy/test/salt-generation/query-quota.test.ts
jmrossy: Hmm, maybe because of how jest implicitly hoists all mocks to the top of the file, except for variables that contain 'mock' in the name (IIRC)

PR #3295, opened 2020-04-02
Add Baklava specific migrations configuration

nategraf:
### Description

This PR adds the baklava and baklava staging contract configurations. It does so by adding support for network specific configs in `migrationsConfig.js` and adds the constant params of the `testing` network as a way to cover this functionality in e2e tests and add a more interpretable common location for these configurations. 

### Tested

e2e tests and deployment to `baklavastaging`

### Other changes

* Remove some code from migrations-util that was not being used, but clobbers the setting for migrations when using celotool


packages/protocol/migrationsConfig.js
```diff
@@ -202,6 +203,65 @@ const DefaultConfig = {
...
+      frozen: false,
+    },
+    reserve: {
+      initialBalance: 100000000,
+    },
+  },
+  baklava: {
+    blockchainParameters: {
+      minimumClientVersion: {
+        major: 0,
```
asaj, line 233: Have we updated the version number in celo-blockchain?

packages/protocol/migrationsConfig.js
```diff
@@ -202,6 +203,65 @@ const DefaultConfig = {
...
+        minor: 10,
+        patch: 0,
+      },
+    },
+    election: {
+      minElectableValidators: '25', // About half of the expected genesis set.
+    },
+    governance: {
+      // Set to be able to complete a proposal in about a day, but give eveyone a chance to participate.
```
asaj: nit: typo

packages/protocol/migrationsConfig.js
```diff
@@ -202,6 +203,65 @@ const DefaultConfig = {
...
+      dequeueFrequency: 4 * HOUR,
+      approvalStageDuration: 4 * HOUR,
```
asaj, line 244: Just a heads up that there's a risk that folks submit governance proposals that dequeue and expire before the approvers are able to approve. A 4 hour SLA on governance proposal approval is not something I expect we can offer. Approvers will likely need advance notice of proposals on Baklava as long as this number is so low.

asaj (approved):

asaj:
This is nice! 👏

packages/protocol/migrationsConfig.js
nategraf: At least for the deployment phase, I am intending to make this trade off. We could increase it after the first set of proposals.

packages/protocol/migrationsConfig.js
nategraf: Yes with the block header deletion it is now `0.10.0`

PR #3296, opened 2020-04-03
[ReleaseGold] Adding canExpire extra configuration

### LINKED ISSUE #3265: ReleaseGold expiry should be configurable
### Expected Behavior

Expiry of ReleaseGold instances should be configurable with a boolean via `initialize`, and should not depend on validator status

lucasege:
### Description

Adding a `canExpire` flag on ReleaseGold that controls if the contract is expiration, rather than simply checking if the contract is registered as a validator.

Also added a fn for `beneficiary` addresses to control this flag, giving them further control to allow for less trusted setups, while still maintaining the option for fund recovery if a user loses their keys.

### Tested

Updated unit tests, running ci tests

### Other changes

Also added more clarity to the releasegold.lockedgold cli so the errors are more clear (checks if account exists before checking pending withdrawals)

### Related issues

- Fixes #3265 

### Backwards compatibility

Yes


packages/cli/src/commands/releasegold/set-can-expire.ts
```diff
@@ -0,0 +1,56 @@
...
+  static expireOptions = ['true', 'false', 'True', 'False']
+
+  static flags = {
+    ...ReleaseGoldCommand.flags,
+    value: flags.enum({
+      options: SetCanExpire.expireOptions,
+      required: true,
+      description: 'canExpire value',
+    }),
+    yesreally: flags.boolean({ description: 'Override prompt to set liquidity (be careful!)' }),
```
asaj: Fix description

packages/contractkit/src/wrappers/ReleaseGold.ts
```diff
@@ -396,6 +398,12 @@ export class ReleaseGoldWrapper extends BaseWrapper<ReleaseGold> {
    */
   setLiquidityProvision = proxySend(this.kit, this.contract.methods.setLiquidityProvision)
 
+  /**
+   * Sets the contract's `canExpire` field to `_canExpire`
+   * @param _canExpire If the contract may be expired after `EXPIRATION_TIME`
```
asaj: nit: if the contract can expire `EXPIRATION_TIME` after the release schedule finishes.

packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -203,6 +207,8 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
     releaseSchedule.releasePeriod = releasePeriod;
     releaseSchedule.releaseCliff = releaseStartTime.add(releaseCliffTime);
     releaseSchedule.releaseStartTime = releaseStartTime;
+    // Nonvalidators (employees) opt-out of Escheat, validators are opt-in.
```
asaj: nit: "Expiry is opt-in for folks who can validate, opt-out for folks who cannot. This is because folks who are running Validators or Groups are likely to want to keep cGLD in the ReleaseGold contract even after it becomes withdrawable."

packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -272,6 +277,20 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
     emit LiquidityProvisionSet(beneficiary);
   }
 
+  /**
+   * @notice Controls if the contract can be expired.
+   *         This would be called by beneficiarys who would like to validate with this contract
```
asaj: If you adopt the comment above I think you can remove these two lines lines.

packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -272,6 +277,20 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
...
+   *         without worrying about future expiration.
+   */
+  function setCanExpire(bool _canExpire) external onlyBeneficiary {
```
asaj, line 286: Missing docstring for `canExpire`

asaj, line 286: nit: do you need the `_`?

packages/protocol/test/governance/release_gold.ts
```diff
@@ -1062,86 +1062,88 @@ contract('ReleaseGold', (accounts: string[]) => {
...
+              const isRevoked = await releaseGoldInstance.isRevoked()
+              assert.equal(isRevoked, true)
             })
-          })
-          describe('registered validator', () => {
-            it('should revert', async () => {
-              await mockValidators.setValidatorGroup(releaseGoldInstance.address)
-              await assertRevert(releaseGoldInstance.expire({ from: releaseOwner }))
+
+            it('should set the released balance at revocation to total withdrawn', async () => {
```
asaj: Can we withdraw some cGLD from the beneficiary first?

packages/protocol/test/governance/release_gold.ts
```diff
@@ -1062,86 +1062,88 @@ contract('ReleaseGold', (accounts: string[]) => {
...
       })
     })
+
+    describe('when the contract is not expirable', () => {
+      beforeEach(async () => {
+        await createNewReleaseGoldInstance(releaseGoldDefaultSchedule, web3)
+        await releaseGoldInstance.setCanExpire(false, { from: beneficiary })
+      })
+
+      describe('when `expire` is called', () => {
```
asaj, line 1185: Can we time travel to the point where it *would* be expirable if the boolean was set? Otherwise we're testing a different condition here

asaj (commented):
Just some nits

packages/protocol/contracts/governance/ReleaseGold.sol
lucasege: to keep the same name, which I think is clean, yes

packages/protocol/test/governance/release_gold.ts
```diff
@@ -1062,86 +1068,122 @@ contract('ReleaseGold', (accounts: string[]) => {
...
           beforeEach(async () => {
-            await releaseGoldInstance.expire({ from: releaseOwner })
-          })
-
-          it('should revoke the contract', async () => {
-            const isRevoked = await releaseGoldInstance.isRevoked()
-            assert.equal(isRevoked, true)
+            const [releaseGoldStartTime, , , ,] = await releaseGoldInstance.releaseSchedule()
+            const expirationTime = await releaseGoldInstance.EXPIRATION_TIME()
+            const timeToTravel = releaseGoldStartTime.plus(expirationTime).toNumber()
```
asaj: Is this right? I thought timetravel takes a delta, but here it seems like you're passing an absolute value

asaj (approved):

packages/protocol/test/governance/release_gold.ts
lucasege: Fixed it, you're right

PR #3299, opened 2020-04-03
Azure HSM Signer

codyborn:
### Description
Introduces a new transaction signer type for Cloud Hardware Security Module (HSM). To reduce the code duplication, I've added more layers of abstraction with more separation of concerns. In addition to wallets there are now signers which focuses only on the different types of signing (in-memory, on a Ledger device, and in KeyVault). The wallet is now focused on account/signer management and less on the actual signing. This allowed me to put most of the wallet logic in a base class (WalletBase).

### Previous Architecture
Old UML

### Proposed Architecture
New UML

### Tested

I've added a set of tests for Azure KeyVault similar to the Local/Ledger tests. These tests run against a deployed Azure KeyVault. The tests will expect the developer to specify the connection parameters in the .env file. Will need to get instructions from @gastonponti on testing the ledger hardware device.


packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -5,6 +5,8 @@ import { account as Account, bytes as Bytes, hash as Hash, RLP } from 'eth-lib'
 import { EncodedTransaction, Tx } from 'web3-core'
 // @ts-ignore-next-line
 import * as helpers from 'web3-core-helpers'
+// @ts-ignore-next-line
```
tkporter: I see there are a few of these that were added, what was the typescript issue? Correct me if I'm wrong but seems like maybe these are just because typescript wants the imports paths to be alphabetically sorted

packages/contractkit/package.json
```diff
@@ -27,21 +27,29 @@
...
     "@ledgerhq/hw-app-eth": "^5.11.0",
     "@ledgerhq/hw-transport": "^5.11.0",
     "@types/debug": "^4.1.5",
+    "@types/secp256k1": "^3.5.3",
+    "bn.js": "^5.1.1",
     "bignumber.js": "^9.0.0",
     "cross-fetch": "3.0.4",
     "debug": "^4.1.1",
-    "ethereumjs-util": "^5.2.0",
+    "dotenv": "^8.2.0",
```
tkporter: Is this only used for testing purposes? If so it could be a `devDependency`

packages/contractkit/.env
```diff
@@ -0,0 +1,5 @@
+AZURE_CLIENT_ID=<CLIENT-APP-ID>
```
tkporter, line 1: Does this mean that in order to run tests we need to manually fill these in?
If so, I'm worried this will mean that only a select few people would have the ability/patience to test this, which imo is undesirable and could mean azure HSM signing could break without us knowing.

One way around this would be to have these values be filled in and encrypted into a `.env.enc` file, similar to the way we treat mnemonic files. That way we could restrict who can see these secrets-- the one downside is I'm not sure if CI has the ability to decrypt these, and if it does we risk people being about to print these secrets in CI

packages/contractkit/src/wallets/signers/azure-hsm-signer.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  async signTransaction(
+    addToV: number,
+    encodedTx: RLPEncodedTx
+  ): Promise<{ v: number; r: Buffer; s: Buffer }> {
+    const hash = getHashFromEncoded(encodedTx.rlpEncode)
+    const bufferedMessage = Buffer.from(trimLeading0x(hash), 'hex')
+    const signature = await this.keyVaultClient.signMessage(bufferedMessage, this.keyName)
+
+    const sigV = Nat.fromString(Bytes.fromNumber(addToV + signature.v))
```
tkporter: The logic here seems a little different from the one in the ledger signer, why's that?

packages/contractkit/src/wallets/signers/azure-hsm-signer.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  async signPersonalMessage(data: string): Promise<{ v: number; r: Buffer; s: Buffer }> {
+    const dataBuff = ethUtil.toBuffer(data)
+    const msgHashBuff = ethUtil.hashPersonalMessage(dataBuff)
+    const signature = await this.keyVaultClient.signMessage(Buffer.from(msgHashBuff), this.keyName)
+
+    const sigR = toPaddedBuffer(signature.r)
+    const sigS = toPaddedBuffer(signature.s)
+    const sigV = signature.v + 27
```
tkporter, line 47: Same, this is different from the ledger signer. Is this because the V the ledger returns will add the 27 and the Azure HSM doesn't?

Separately, just to confirm, only transactions are affected by EIP 155 and not personal messages, right?

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,181 @@
...
+  private readonly keyClient: KeyClient
+  private readonly SIGNING_ALGORITHM: string = 'ECDSA256'
+
+  // 0x04 prefix indicates that the key is not compressed
+  // https://tools.ietf.org/html/rfc5480#section-2.2
+  private readonly publicKeyPrefix: number = 0x04
+
+  constructor(vaultName: string) {
+    this.vaultName = vaultName
+    this.vaultUri = `https://${this.vaultName}.vault.azure.net`
```
tkporter, line 34: Is this uri accessible regardless of being executed within an Azure network?

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,181 @@
...
+    // DefaultAzureCredential supports service principal or managed identity
+    // If using a service principal, you must set the appropriate environment vars
+    this.credential = new DefaultAzureCredential()
```
tkporter, line 37: Could you explain how we expect someone using this to provide credentials?
I think it'd be nice to not rely on env vars in here and instead pass in specific config secrets to contractkit, but if that takes a big refactor I'd like to understand more what's expected in the env vars

tkporter (commented):
Didn't do a deep pass, mostly just questions for now

packages/contractkit/.gitignore
```diff
@@ -3,4 +3,5 @@ lib/
 .devchain/
 .devchain.tar.gz
 tmp/
-.tmp/
\ No newline at end of file
+.tmp/
+.env
```
yerdua, line 7: Since `.env` is here in `.gitignore`, why's it included above? Should it be?

packages/contractkit/src/utils/ledger-utils.ts
```diff
@@ -0,0 +1,16 @@
+import debugFactory from 'debug'
+// @ts-ignore-next-line
+import { BN } from 'bn.js'
```
yerdua: This doesn't appear to be used. Is this needed? If so, what does it do?

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+  })
+
+  beforeEach(async () => {
+    await wallet.init()
+  })
+
+  test('hasKey should return true for keys that are present', async () => {
+    // Invalid key should not be present
+    expect(await wallet.hasAccount('this is not a valid private key')).toBeFalsy()
```
yerdua: nit: this should be its own test, like "hasKey should return false for keys that are not present"

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,181 @@
...
+      if (e.message.includes('this is not a valid private key')) {
+        return false
+      }
+      throw e
+    }
+    return true
+  }
+
+  /**
+   * Returns true if the signature is in the "bottom" of the curve
```
yerdua, line 127: What does `"bottom" of  the curve` mean, and how does that map to the concept of being "Canonical"?

packages/contractkit/src/wallets/signers/signer.ts
```diff
@@ -0,0 +1,25 @@
+// @ts-ignore-next-line
+import { account as Account, bytes as Bytes, hash as Hash, nat as Nat, RLP } from 'eth-lib'
+// @ts-ignore-next-line
+import * as helpers from 'web3-core-helpers'
+// @ts-ignore-next-line
+import { BN } from 'bn.js'
```
yerdua: What's the reason for all the `@ts-ignore-next-line`s here?
I'm curious if these are needed, since they don't appear to be referenced in this file. But, the fact that you added the ts-ignore suggests that they _are_ necessary in a way that's not understood by the linter.

I'd like to see a note next to those briefly explaining why, to help some future editor of the file understand what's safe to change/remove.

packages/contractkit/src/utils/signing-utils.ts
codyborn: Thanks for bringing this up. I was getting the error:
`This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.`
I found that changing the line to the following resolved the issue:
`import BN = require('bn.js')`
I think it has to do with how bn.js is exporting the BN type and by performing the `require` it bypasses the typescript compiler. I haven't used typescript before so I'm still ramping up.. would appreciate any alternative suggestions/best-practices here.

packages/contractkit/package.json
codyborn: moved to dev dependencies, thanks

packages/contractkit/.gitignore
codyborn: I included a template for devs (above). The expectation is that each dev will modify their local copy with the credentials. Since we shouldn't be checking this in (contains secrets), I added it to the .gitignore to prevent tracking of future changes.

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: Correct, this is the public endpoint for the KeyVault. It's authenticated via MSI in this case.

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: The env vars needed are outlined in the .env template. I'll add a write-up in our dev readme on where to get these from. I just added a mock AzureKeyVaultClient for unit testing. This is convenient for devs not touching the HSM logic, but shouldn't be solely relied upon if modifying the signing logic.

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: It has to do with the ECDSA curve and the standard that Eth (and BTC) uses. I added this link in the comment above: https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures

If you're curious to dive deep on ECDSA signatures, I found this primer to be helpful:
https://blog.cloudflare.com/a-relatively-easy-to-understand-primer-on-elliptic-curve-cryptography/

packages/contractkit/src/wallets/signers/ledger-signer.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  async signPersonalMessage(
+    data: string
+  ): Promise<{ v: number; r: Buffer | Uint8Array; s: Buffer | Uint8Array }> {
+    try {
+      const dataBuff = ethUtil.toBuffer(data)
+      const msgHashBuff = ethUtil.hashPersonalMessage(dataBuff)
+      const trimmedMsgHash = trimLeading0x(msgHashBuff.toString('hex'))
```
codyborn: @gastonponti previously this was `trimLeading0x(data)` which I believe is incorrect. We also weren't calling `ethUtil.hashPersonalMessage(dataBuff)` before. Check out the ledger unit tests for example:
https://github.com/LedgerHQ/ledgerjs/blob/ae32cfff7930892c32e38f9c74ac6b4fa193b19b/packages/hw-app-eth/tests/Eth.test.js#L68

packages/contractkit/src/wallets/signers/ledger-signer.ts
```diff
@@ -0,0 +1,90 @@
...
+        transportErrorFriendlyMessage(error)
+      }
+      throw error
+    }
+  }
+
+  async signTypedData(typedData: EIP712TypedData): Promise<{ v: number; r: Buffer; s: Buffer }> {
+    try {
+      const dataBuff = generateTypedDataHash(typedData)
+      const trimmedData = trimLeading0x(dataBuff.toString('hex'))
```
codyborn, line 106: @gastonponti similarly, I updated it here as well

packages/contractkit/.env
codyborn: I agree that this is too much friction for devs who are not making changes to the HSM logic. I've implemented a mock HSM `AzureKeyVaultClient` that still allows us to test most of this logic while using an in-memory implementation. If a developer is making changes to any of the signing logic they should definitely provide a real cloud-HSM to test it E2E. I will add a section in the dev readme for how to populate these.

packages/contractkit/src/wallets/signers/azure-hsm-signer.ts
codyborn: I've consolidated the logic as much as I could across all three implementations. Most of the common logic is in wallet.ts. There is some small duplicate logic that allows us to have a stronger interface (ex. using Tx type instead of string). The main reason there are separate classes for each signer, is because each underlying library (ledger, azure, and ethUtil) all have different interfaces.

packages/contractkit/src/wallets/signers/azure-hsm-signer.ts
codyborn: Correct, Ledger and EthUtils both automatically add the 27 to the 'v' value.

RE: EIP155, that's my understanding as well. It's up to the application (smart contract) to enforce any specific qualities of the signed data.

packages/contractkit/src/utils/ledger-utils.ts
codyborn: Thanks for pointing this out.. I had many of these unused imports hidden from the ts-ignore comment. I've cleaned them up!

packages/contractkit/src/wallets/signers/signer.ts
codyborn: Addressed in the previous comment.. thanks again!

packages/contractkit/.env
tkporter: Good idea with the mock HSM. I think we should consider down the road a semi-automated way to run the e2e test before any releases

packages/contractkit/package.json
```diff
@@ -27,21 +27,28 @@
...
+    "ethereumjs-util": "^5.2.0",
     "fp-ts": "2.1.1",
     "io-ts": "2.0.1",
+    "ms-rest-azure": "^3.0.0",
+    "secp256k1": "^4.0.0",
     "web3": "1.2.4",
     "web3-core": "1.2.4",
     "web3-core-helpers": "1.2.4",
-    "web3-eth-abi": "1.2.4",
+    "web3-eth-abi": "^1.2.6",
```
gastonponti: Did this worked?
We fixed the `web3` versions across the repo because they don't care to be retro-compatible with their own...

packages/contractkit/package.json
```diff
@@ -27,21 +27,28 @@
...
   },
   "dependencies": {
+    "@azure/identity": "1.1.0-preview1",
+    "@azure/keyvault-keys": "^4.0.2",
     "@celo/utils": "0.1.9-dev",
     "@ledgerhq/hw-app-eth": "^5.11.0",
     "@ledgerhq/hw-transport": "^5.11.0",
     "@types/debug": "^4.1.5",
+    "@types/secp256k1": "^3.5.3",
+    "bn.js": "^5.1.1",
```
gastonponti: We are trying to use only one library for the bigNumbers. I know that we are having a lot of troubles because web3 types assumes to use BN, but really don't know the best solution here. @mcortesi maybe could know more about it.

packages/contractkit/src/wallets/ledger-wallet.test.ts
```diff
@@ -53,6 +56,46 @@ const ledgerAddreses: { [myKey: string]: { address: string; privateKey: string }
 
 const CHAIN_ID = 44378
 
+// Sample data from the official EIP-712 example:
+// https://github.com/ethereum/EIPs/blob/master/assets/eip-712/Example.js
```
gastonponti, line 66: GREAT!!!!!!!!!! ❤️ ❤️

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,177 @@
...
+      canonicalizedSignature,
+      publicKey,
+      message
+    )
+    return new Signature(recoveryParam, R, S)
+  }
+
+  public async hasKey(keyName: string): Promise<boolean> {
+    try {
+      await this.keyClient.getKey(keyName!)
```
gastonponti: `keyName` it's already an string, why the `!`?

gastonponti: You could even reject invalid strings before performing the client call (don't know if the `keyClient` already does that)
Don't know if that key should follow a specified pattern.

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private async getKey(keyName: string): Promise<KeyVaultKey> {
+    try {
+      let signingKey = await this.keyClient.getKey(keyName)
+
+      if (
+        typeof signingKey == 'undefined' ||
+        typeof signingKey.id == 'undefined' ||
+        typeof signingKey.key == 'undefined' ||
```
gastonponti: You could change to something like this:
```
if (typeof signingKey?.id == 'undefined' || typeof signingKey?.key?.x == 'undefined' || typeof signingKey?.key?.y == 'undefined')
```
the `?.` will allow you to collapse some checks

packages/contractkit/src/wallets/signers/ledger-signer.ts
gastonponti: Let me check this... I don't really remember if this was called before this sign method.
Check the `cli/src/commands/accounts/proof-of-possession.ts` as an example
Follow the `generateProofOfKeyPossession` method.
There is a `parseSignature` in the `signatureUtils` that invokes this (`hashMessageWithPrefix`)
```
export function hashMessageWithPrefix(message: string) {
  const prefix = '\x19Ethereum Signed Message:\n' + messageLength(message)
  return Web3Utils.soliditySha3(prefix, message)
}
```
This is something that I found the other day, but I thing that at least for our `cli` we were probably be making a doble header+hashing
I'll try to check it too, but take a look in that

packages/contractkit/src/wallets/signers/ledger-signer.ts
gastonponti: Same. Don't know if this will break some `cli` commands
I'm not telling that this is wrong, this should probably be the way to do it, but I think that the `cli` patched some older and wrong behaviour

gastonponti: Ok the local/default wallet has it like this. I'll check that to gather more info

gastonponti (changes requested):
Check the personalMessage sign for the double header (I'll do the same)
Because, we have 2 possibilities, we have a (wrong) cli working for the ledger and we have to change it, or the personalSign of the default wallet, never worked.

packages/contractkit/.env
codyborn: Agree, this can be part of our CI pipeline 👍

packages/contractkit/src/wallets/signers/ledger-signer.ts
codyborn: When stepping through the code, I found that previously this `dataBuff.toString()` was some binary blob but the ledger function is [expecting a hex string](https://github.com/LedgerHQ/ledgerjs/blob/a67978ffea959173b464f018174da7eded38779a/packages/hw-app-eth/src/Eth.js#L248). I'll follow the CLI logic you linked to above to test it out.

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: Awesome, thanks!

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: You're right, I'm not sure why it was there. Key names are arbitrary, user-provided strings. Azure naming restrictions sometime change and wouldn't want to unintentionally block some key names in the future.

packages/contractkit/package.json
codyborn: I was trying to use the BigNumber.js instead, but there were some operations that I needed that weren't supported. I've gone ahead and implemented the missing operations for BigNumber.js

packages/contractkit/package.json
codyborn: This was unintentional, I've reverted it back to 1.2.4

packages/contractkit/src/wallets/signers/ledger-signer.ts
codyborn: The message is first hashed in [Accounts.ts](https://github.com/celo-org/celo-monorepo/blob/16622e258bf68c4d56ffb347a3bc88d4daaee9ce/packages/contractkit/src/wrappers/Accounts.ts#L399)
 `const hash = Web3.utils.soliditySha3({ type: 'address', value: address })`
It then gets routed through web3->CeloProvider where it ends up calling signPersonalMessage on the LedgerWallet.
This then calls [ethUtil.hashPersonalMessage](https://github.com/ethereumjs/ethereumjs-util/blob/ae4c3009edd8307b1ef3bebc2dc03c9c2fdf1b40/src/signature.ts#L136) which formats it as an Ethereum message to be signed. This is consistent with the in-memory signer which performs the [hashPersonalMessage](https://github.com/celo-org/celo-monorepo/blob/8b54e4803a67eeb9ac772e8b9e261446623ede41/packages/contractkit/src/wallets/default-wallet.ts#L62) operation.

Once it's signed, `getParsedSignatureOfAddress()` calls `parseSignature(hash, signature, signer)` which takes the original hash, adds the Ethereum prefix (that you pointed out above), and performs ecrecover to verify the signature. Since the signed message and this validating message both have the prefix, the original signer can be recovered and the validation step should return true.

Given that ledger wallet was not adding the prefix previously, I don't know how the verification was working. I'll test it out now using a ledger device.

packages/contractkit/.env
```diff
@@ -0,0 +1,5 @@
...
+AZURE_TENANT_ID=7cb7628a-e37c-4afb-8332-2029e418980e
+AZURE_CLIENT_SECRET=<CLIENT_SECRET>
+VAULT_NAME=<AKV_NAME>
```
tkporter: Nit: could we prefix this and `KEY_NAME` with `AZURE_`?

packages/contractkit/src/utils/signing-utils.ts
tkporter: Hmm interesting-- have you tried adding `"esModuleInterop": true` to the contractkit `tsconfig.json`? Unsure of the consequences of adding that

packages/contractkit/.env
```diff
@@ -0,0 +1,5 @@
+AZURE_CLIENT_ID=<CLIENT-APP-ID>
+AZURE_TENANT_ID=7cb7628a-e37c-4afb-8332-2029e418980e
+AZURE_CLIENT_SECRET=<CLIENT_SECRET>
```
tkporter, line 3: Is this used?

packages/contractkit/src/utils/signing-utils.ts
codyborn: I ended up consolidating BigNumber libraries so this import was removed altogether

packages/contractkit/src/wallets/signers/ledger-signer.ts
codyborn: After more digging with @gastonponti, we found that:
1. The ledger signer app adds the "Ethereum" prefix to all `signPersonalMessage()` calls
2. [EIP712 expects this prefix for signed messages](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#eth_signtypeddata)

This means that the ledger signer is correct, but the in-memory signer had a bug where it left out the "Ethereum" prefix. That's been fixed in this PR.

packages/contractkit/src/wallets/signers/ledger-signer.ts
codyborn: As you know @gastonponti, it's all tested and working now 🙌

packages/contractkit/.env
codyborn: Good idea. I've prefixed both KEY_NAME and VAULT_NAME with "AZURE_"

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
```diff
@@ -0,0 +1,325 @@
...
+              const privKey = keyVaultAddresses.get(keyName)!.privateKey
+              const pubKey = ethUtil.privateToPublic(ethUtil.toBuffer(privKey))
+              return new BigNumber(ensureLeading0x(pubKey.toString('hex')))
+            },
+            signMessage: async (message: Buffer, keyName: string): Promise<Signature> => {
+              if (keyVaultAddresses.has(keyName)) {
+                const trimmedKey = trimLeading0x(keyVaultAddresses.get(keyName)!.privateKey)
+                const pkBuffer = Buffer.from(trimmedKey, 'hex')
+                const signature = ethUtil.ecsign(message, pkBuffer)
+                return new Signature(signature.v - 27, signature.r, signature.s)
```
tkporter, line 131: Could you add a comment why the `- 27` is happening?

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
```diff
@@ -0,0 +1,325 @@
...
+          }
+        })
+    }
+  })
+
+  describe('without initializing', () => {
+    const knownAddress = ACCOUNT_ADDRESS1
+    let celoTransaction: Tx
+    beforeEach(() => {
+      return new Promise(async (resolve) => {
```
tkporter: Any reason why this is & the other similar beforeEach blocks are using a promise?

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
```diff
@@ -0,0 +1,325 @@
...
+          gatewayFee: '0x5678',
+          data: '0xabcdef',
+        }
+        resolve()
+      })
+    })
+
+    test('fails calling getAccounts', () => {
+      try {
+        wallet.getAccounts()
```
tkporter, line 161: I’m fairly sure that if this doesn’t throw an error, the test case in the catch won’t be run and the test will still pass. Consider a check to make sure the test case in the catch was ran. There are a few other places like this elsewhere in the tests

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
```diff
@@ -0,0 +1,325 @@
...
+    test('fails calling signTypedData', async () => {
+      await expect(wallet.signTypedData(ACCOUNT_ADDRESS1, TYPED_DATA)).rejects.toThrowError()
+    })
+  })
+
+  describe('after initializing', () => {
+    beforeEach(async () => {
+      await wallet.init()
+    })
+    test('hasKey should return false for keys that are not present', async () => {
```
tkporter: Nit: `hasKey` -> `hasAccount`

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
```diff
@@ -0,0 +1,325 @@
...
+      // Invalid key should not be present
+      expect(await wallet.hasAccount('this is not a valid private key')).toBeFalsy()
+    })
+
+    test('hasKey should return true for keys that are present', async () => {
```
tkporter: Ditto

packages/contractkit/src/wallets/azure-hsm-wallet.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+    const keys = await this.keyVaultClient.getKeys()
+    const addressToSigner = new Map<Address, Signer>()
+    for (const key of keys) {
+      try {
+        const address = await this.getAddressFromKeyName(key)
+        addressToSigner.set(address, new AzureHSMSigner(this.keyVaultClient, key))
+      } catch (e) {
+        // Safely ignore non-secp256k1 keys
+        const message = e.message
```
tkporter: Nit: any need for this variable?

packages/contractkit/src/wallets/ledger-wallet.test.ts
```diff
@@ -51,12 +60,72 @@ const ledgerAddreses: { [myKey: string]: { address: string; privateKey: string }
...
+    if (USE_PHYSICAL_LEDGER) {
+      try {
+        // Use physical ledger if present
+        // Recreation of the connection will fail, therefore we use a single object
+        if (!hardwareWallet) {
+          const transport: Transport = await TransportNodeHid.open('')
+          hardwareWallet = await new LedgerWallet(undefined, undefined, transport)
+        }
+      } catch (e) {
+        throw new Error(
```
tkporter, line 172: Including `e` in here might be helpful

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
```diff
@@ -0,0 +1,325 @@
...
+                data: '0xabcdef',
+              }
+              resolve()
+            })
+          })
+
+          describe('when calling signTransaction', () => {
+            test('succeeds', async () => {
+              const signedTx: EncodedTransaction = await wallet.signTransaction(celoTransaction)
+              expect(signedTx).not.toBeUndefined()
```
tkporter, line 291: Should we verify the signature?

packages/contractkit/src/wallets/remote-wallet.ts
```diff
@@ -0,0 +1,109 @@
...
+  private setupFinished = false
+  private setupLocked = false
+  private INIT_TIMEOUT_IN_MS = 10 * 1000
+
+  /**
+   * Discovers wallet accounts and caches results in memory
+   * Idempotent to ensure multiple calls are benign
+   */
+  async init() {
+    if (this.setupLocked || this.setupFinished) {
```
tkporter, line 21: Wdyt about just having a separate condition if setupFinished is true that returns immediately? It would help with readability-- just feels a little weird to call `await initCompleted()` if the setup isn’t locked and is finished

packages/contractkit/package.json
```diff
@@ -27,17 +27,24 @@
     "lint": "tslint -c tslint.json --project ."
   },
   "dependencies": {
-    "@celo/utils": "0.1.9-dev",
+    "@azure/identity": "^1.0.2",
+    "@azure/keyvault-keys": "^4.0.2",
+    "@celo/utils": "^0.1.8",
```
gastonponti: Remember to point to `-dev`, that way we can always check against `master`

packages/contractkit/package.json
```diff
@@ -27,17 +27,24 @@
...
     "@ledgerhq/hw-app-eth": "^5.11.0",
     "@ledgerhq/hw-transport": "^5.11.0",
+    "@ledgerhq/hw-transport-node-hid": "^5.11.0",
     "@types/debug": "^4.1.5",
+    "@types/secp256k1": "^3.5.3",
```
gastonponti: the `@types` could be in the `devDependencies` (don't know why the `@types/debug` is there, haha, just noticed)

gastonponti (commented):
You have some tests failings.
I've added two comments for the `package.json`
Also, update master because I was needing to add my other changes for the ledger Team (sorry, shouldn't be a big change)

packages/contractkit/src/wallets/remote-wallet.ts
```diff
@@ -0,0 +1,109 @@
...
+  getAccounts(): Address[] {
+    this.initializationRequired()
+    return super.getAccounts()
+  }
+
+  /**
+   * Returns true if account is in the remote wallet
+   * @param address Account to check
+   */
+  hasAccount(address?: Address) {
```
tkporter: Might want to keep with the trend of being explicit about return types

packages/contractkit/src/wallets/signers/azure-hsm-signer.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  constructor(keyVaultClient: AzureKeyVaultClient, keyName: string) {
+    this.keyVaultClient = keyVaultClient
+    this.keyName = keyName
+  }
+
+  async signTransaction(
+    addToV: number,
+    encodedTx: RLPEncodedTx
+  ): Promise<{ v: string; r: string; s: string }> {
```
tkporter: Any reason for `{ v: string; r: string; s: string }` in the "lower level" and `{ v: number; r: Buffer; s: Buffer }` higher up?

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,191 @@
...
+      typeof signResult === 'undefined' ||
+      typeof signResult.result === 'undefined' ||
+      signResult.result.length !== 64
+    ) {
+      throw new Error(`Invalid signature returned from Azure: ${signResult}`)
+    }
+    const rawSignature = signResult.result
+
+    // Canonicalize signature
+    const secp256k1Curve = new EC('secp256k1')
```
tkporter: Looked briefly at `elliptic` to try to understand it a little better and they recommend creating this once and reusing if you can

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,191 @@
...
+  /**
+   * Attempts each recovery key to find a match
+   */
+  private static recoverKeyIndex(
+    signature: Uint8Array,
+    publicKey: BigNumber,
+    hash: Uint8Array
+  ): number {
+    for (let i = 0; i < 4; i++) {
+      const compressed = false
```
tkporter, line 142: nit: no need for this (unless if you want it for readability?)

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  private static bufferToBigNumber(input: Buffer): BigNumber {
+    return new BigNumber(ensureLeading0x(input.toString('hex')))
+  }
+  private static bigNumberToBuffer(input: BigNumber): Buffer {
+    return ethUtil.toBuffer(ensureLeading0x(input.toString(16))) as Buffer
+  }
+}
+
+export class Signature {
```
tkporter, line 202: Seems like something that's useful outside of `azure-key-vault-client.ts`, maybe it should live somewhere else?

packages/contractkit/src/wallets/wallet.ts
```diff
@@ -1,11 +1,120 @@
...
+   */
+  getAccounts(): Address[] {
+    return Array.from(this.accountSigners.keys())
+  }
+
+  /**
+   * Returns true if account has been registered
+   * @param address Account to check
+   */
+  hasAccount(address?: Address) {
```
tkporter: nit: return type

gastonponti (approved):
Great! LGTM
One little detail. Probably you have to do a `yarn install` and then a `yarn docs`
See that you removed the `contractkit` from every path. It happened the same to me today.
Install and generate the docs againt. Otherwise circle-ci will probably fail

packages/contractkit/package.json
codyborn: thanks!

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
codyborn: Done, thanks

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
codyborn: good catch ;)

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
codyborn: It's equivalent to an async beforeEach that I was using when trying to debug some test issues. I've consolidated to the async approach since it's cleaner 👍

packages/contractkit/src/wallets/signers/azure-hsm-signer.ts
codyborn: It was originally based on what the caller needed but I like the idea of having a uniform interface and having the caller modify the value to suit their needs. {number, Buffer, Buffer} is a stronger interface so I'll go with that.

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: This was intentional to be self-documenting. It doesn't look like TS supports named parameters.

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: Agree. I think this is part of larger work to consolidate the signing libraries/utilities. There exist some similar objects throughout the repo, but they don't provide much guarantees around the contents. I've added an issue #3410 to track this work.

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: Thanks for noticing this!

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -103,44 +105,30 @@ export function rlpEncodedTx(tx: Tx): RLPEncodedTx {
...
   r: string
   s: string
-}): { v: string; r: string; s: string } {
+}): { v: number; r: Buffer; s: Buffer } {
   return {
-    v: makeEven(trimLeadingZero(ensureLeading0x(signature.v))),
-    r: makeEven(trimLeadingZero(ensureLeading0x(signature.r))),
-    s: makeEven(trimLeadingZero(ensureLeading0x(signature.s))),
+    v: parseInt(makeEven(trimLeadingZero(ensureLeading0x(signature.v))), 16),
+    r: Buffer.from(makeEven(trimLeadingZero(ensureLeading0x(signature.r)))),
```
tkporter: This and the s below should be something like
`Buffer.from(trimLeading0x(makeEven(trimLeadingZero(ensureLeading0x(signature.r)))), 'hex')`

The way it is currently, `Buffer.from` is not reading them as hex

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -103,44 +105,30 @@ export function rlpEncodedTx(tx: Tx): RLPEncodedTx {
...
 
 export async function encodeTransaction(
   rlpEncoded: RLPEncodedTx,
-  signature: { v: string; r: string; s: string }
+  signature: { v: number; r: Buffer; s: Buffer }
 ): Promise<EncodedTransaction> {
   const hash = getHashFromEncoded(rlpEncoded.rlpEncode)
 
+  const v = stringNumberToHex(signature.v)
+  const r = signature.r.toString()
```
tkporter: This and s below should be something like `ensureLeading0x(signature.r.toString('hex'))`

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Attempts each recovery key to find a match
+   */
+  private static recoverKeyIndex(
+    signature: Uint8Array,
+    publicKey: BigNumber,
+    hash: Uint8Array
+  ): number {
+    for (let i = 0; i < 4; i++) {
```
tkporter, line 141: I think it can only be 0 and 1, right?

packages/contractkit/src/utils/signing-utils.ts
codyborn: I realized that since I'm converting the signature to {number, Buffer, Buffer} already, this util function is unnecessary. I've removed it. Thanks for bringing this up

packages/contractkit/src/utils/signing-utils.ts
codyborn: 👍

packages/contractkit/src/wallets/ledger-utils/tokens.ts
```diff
@@ -24,6 +24,23 @@ export interface TokenInfo {
   data: Buffer
 }
 
+/**
+ * @return
+ * -1: version1 < version2,
+ *  0: version1 == version2,
+ *  1: version1 > version2
+ */
+export function compareLedgerAppVersions(version1: string, version2: string): number {
```
gastonponti, line 33: Great! Yes, this should be the place 👍

packages/contractkit/src/wallets/signers/azure-key-vault-client.ts
codyborn: Apparently there's a small chance that the second bit can be set as well.
https://ethereum.stackexchange.com/questions/42455/during-ecdsa-signing-how-do-i-generate-the-recovery-id
> There's also a super rare chance that you need to set the second bit of the recovery id meaning the recovery id could in theory be 0, 1, 2, or 3. But there's a 0.000000000000000000000000000000000000373% of needing to set the second bit according to a question on Bitcoin.SE

I've looked at other implementations of this and confirmed that they check all four recovery ids.

codyborn: Ex. https://github.com/PegaSysEng/ethsigner/blob/a1f2ccda801dd040074824a1b68ba3538be4a40a/ethsigner/signer/azure/src/main/java/tech/pegasys/ethsigner/signer/azure/AzureKeyVaultTransactionSigner.java#L85

packages/contractkit/src/wallets/ledger-utils/tokens.ts
codyborn: 👍

packages/contractkit/src/wallets/azure-hsm-wallet.ts
codyborn: When I was debugging I couldn't see the contents of this variable until I created a temp var. Will clean it up. Thanks!

PR #3300, opened 2020-04-03
Fix e2e transfer tests

jeanregisser:
### Description

E2E transfers tests broke in b5dbd68ed8073a1f1f245feb1e3dd4d44c024619, likely due to the upgrade of BigNumber.
This PR makes sure the inflation number doesn't use exponential notation.

### Tested

E2E transfers tests pass (still some remaining flakiness though).

### Other changes

None

### Related issues

Discussed on Slack.

### Backwards compatibility

Yes


asaj (approved):
Nice find

PR #3307, opened 2020-04-03
Use Gold.mint() instead of state.AddBalance() for epoch rewards

mcortesi:
### Description

Companion PR to https://github.com/celo-org/celo-blockchain/pull/965

### DIFF
--- a/packages/protocol/contracts/common/GoldToken.sol
+++ b/packages/protocol/contracts/common/GoldToken.sol
@@ -140,7 +140,7 @@ contract GoldToken is Initializable, CalledByVm, Freezable, IERC20, ICeloToken {
    * @param to The account for which to mint tokens.
    * @param value The amount of cGLD to mint.
    */
-  function mint(address to, uint256 value) external onlyVm onlyWhenNotFrozen returns (bool) {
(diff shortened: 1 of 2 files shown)


packages/protocol/contracts/common/GoldToken.sol
```diff
@@ -135,6 +135,23 @@ contract GoldToken is Initializable, CalledByVm, Freezable, IERC20, ICeloToken {
     return true;
   }
 
+  /**
+   * @notice Mints new cGLD and gives it to 'to'.
+   * @param to The account for which to mint tokens.
+   * @param value The amount of cGLD to mint.
+   */
+  function mint(address to, uint256 value) external onlyVm onlyWhenNotFrozen returns (bool) {
```
asaj: Should this be `onlyWhenNotFrozen`? Otherwise epoch rewards will be frozen until gold transfers are unfrozen

mcortesi: I though they were the same thing.. I'll remove that

asaj (approved):
Please wait for https://github.com/celo-org/celo-monorepo/pull/3219 before merging

PR #3319, opened 2020-04-06
Remove buggy total balance from CLI

asaj:
### Description

Returning `total: 3.963877391197344453575983046348115674221700746820753546331534351508065746944e+75`

### Tested

Not tested

### Other changes

Remove `leaderboard` package

### Related issues

None
### Backwards compatibility

Yes


gastonponti:
The issue is to remove it, or fix that number?

asaj:
Well I'm not sure why we try to calculate this number in the first place..

mrsmkl (approved):
I think fixed the conversion a while ago, it was a problem with `quoteUsdSell` (the call reverted when 0 was passed as argument, but instead of throwing, some kind of error code was returned).

But at least the leaderboard should be removed to clean up the repo.

PR #3322, opened 2020-04-06
Add cli rewards:slasher

jfoutts-celo:
### Description

Add cli rewards:slasher command to manually slash a validator, or automatically monitor for and slash validators for downtime.

### Other changes

- Change slashValidator to slashSigner in DowntimeSlasher. This matches DoubleSigningSlasher and encourages better usage.
- Remove optional blockNumber from signerToAccount, it should never be needed.

### Tested

Tested on testnet that automatically slashes Validator for first block of slashable downtime.

### Backwards compatibility

Yes.


packages/cli/src/commands/rewards/slasher.ts
```diff
@@ -0,0 +1,269 @@
...
+              )
+            } else if (
+              sealBlockNumber - validatorDownSince[i] + 1 >=
+              slashableDowntime + extraWait
+            ) {
+              const downtime = sealBlockNumber - validatorDownSince[i] + 1
+              const downtimeEndBlock = sealBlockNumber - extraWait
+              const validator = validatorSet[i]
+              const signer = validatorSigners[i]
+              const alreadySlashedSigner = alreadySlashed[normalizeAddress(signer)]
```
jfoutts-celo, line 264: Only check alreadySlashed if startBlock has been supplied.  Explain why we don't need to check if we're caught up to tip.

nambrot:
@yorhodes I assume this is no longer applicable and closing

PR #3326, opened 2020-04-07
Add oracle address overrides for baklava contract migrations

tkporter:
### Description

Adds oracle addresses to be added at migration time for baklava.

These were generated essentially using
```
INDEX=0
while [ $INDEX -lt 10 ]; do
	celotooljs generate account-address --private-key `celotooljs generate bip32 --mnemonic "$MNEMONIC" --accountType price_oracle --index $INDEX`
	INDEX=$(( $INDEX + 1))
done
```
### Tested

Logged the config to make sure the proper overrides were used for baklava

### Other changes

Fixed a recent helm chart change that caused full nodes to crash

### Related issues

n/a

### Backwards compatibility

Yes


asaj:
What is the intended use case here? AFAIK `celotool deploy initial contracts` is not RC0/RC1/Baklava friendly, so this will not help with the Baklava contract deployment

timmoreton (approved):

asaj (approved):

PR #3341, opened 2020-04-08
Fix election:vote which uses an unstable sort

asaj:
## Description

This PR fixes the election:vote command by removing the unstable sort used in the contract wrapper, and instead relying on the fact that the contract will already return groups in a sorted order.

## Other changes

- Small docs fix
- Add `prompts` to CLI dependencies

## Tested

No
## Related issues

None
## Backwards compatibility

None


mrsmkl (approved):
Seems correct.

PR #3342, opened 2020-04-08
CLI fixes for new Baklava

mrsmkl:
## Description

Fixes problems found when checking if these work on the new baklava.
* If a validator is unregistered, return dummy data. This makes election:current work
* In election:run, if it fails, retry with minimum validators set to zero

### DIFF
--- a/packages/cli/src/commands/election/run.ts
+++ b/packages/cli/src/commands/election/run.ts
@@ -5,16 +5,12 @@ import { validatorTable } from '../validator/list'
 
 async function performElections(kit: ContractKit) {
   const election = await kit.contracts.getElection()
-  const election3 = await kit._web3Contracts.getElection()
   try {
     const signers = await election.electValidatorSigners()
     return signers
   } catch (err) {
     console.warn('Warning: error running actual elections, retrying with minimum validators at 0')
-    const config = await election.getConfig()
-    return election3.methods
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/election/run.ts
```diff
@@ -1,7 +1,23 @@
...
+async function performElections(kit: ContractKit) {
+  const election = await kit.contracts.getElection()
+  const election3 = await kit._web3Contracts.getElection()
+  try {
+    const signers = await election.electValidatorSigners()
+    return signers
+  } catch (err) {
+    console.warn('Warning: error running actual elections, retrying with minimum validators at 0')
+    const config = await election.getConfig()
+    return election3.methods
```
asaj: Can we add this to the contractKit wrapper here instead?

If it's easy, we could even consider making the `electValidatorSigners` in the wrapper take optional `min` and `max` arguments. If neither are provided, call `electValidatorSigners`, if at least one is provided, call `electNValidatorSigners`

asaj (approved):

asaj:
I think `validator:status` will still fail due to this:
https://github.com/celo-org/celo-monorepo/blob/master/packages/cli/src/commands/validator/status.ts#L109

mrsmkl:
I'll fix that too.

PR #3348, opened 2020-04-08
New chart for full-nodes in AKS

jcortejoso:
## Description

Added a chart for aks fullnodes.

## Other changes

Some refactor on `common` chart.

## Tested

Tested the chart on AKS. Pending to migrate/test the existing charts to the new common version.

## Related issues

- Fixes #aks


packages/celotool/src/lib/aks-fullnode.ts
```diff
@@ -0,0 +1,118 @@
...
+import {
+  deletePersistentVolumeClaims,
+  installGenericHelmChart,
+  removeGenericHelmChart,
+  upgradeGenericHelmChart,
+} from './helm_deploy'
+import { execCmdWithExitOnFailure } from './utils'
+
+const helmChartPath = '../helm-charts/celo-fullnode'
+// const releaseName = 'celo-fullnodes'
```
tkporter: nit: comments probably were intended to be removed

packages/celotool/src/lib/aks-fullnode.ts
```diff
@@ -0,0 +1,118 @@
...
+export function getReleaseName(celoEnv: string) {
+  return `${celoEnv}-fullnodes`
+}
+
+export function getKubeNamespace(celoEnv: string) {
+  return celoEnv
+}
+
+export function getReplicaCount() {
+  return parseInt(fetchEnv(envVar.AZURE_TX_NODES_COUNT), 3)
```
tkporter: This uses base 3 to parse `AZURE_TX_NODES_COUNT`, I assume we want base 10

packages/celotool/src/lib/aks-fullnode.ts
```diff
@@ -0,0 +1,118 @@
...
+async function helmParameters(celoEnv: string, kubeNamespace: string) {
+  const staticIps = (await createStaticIPs(celoEnv)).join(',')
+  const replicaCount = getReplicaCount()
+  return [
+    `--set namespace=${kubeNamespace}`,
+    `--set replicaCount=${replicaCount}`,
+    `--set storage.size=${parseInt(fetchEnv(envVar.AZURE_TX_NODES_DISK_SIZE), 10)}`,
+    `--set geth.image.repository=${fetchEnv(envVar.GETH_NODE_DOCKER_IMAGE_REPOSITORY)}`,
+    `--set geth.image.tag=${fetchEnv(envVar.GETH_NODE_DOCKER_IMAGE_TAG)}`,
+    // `--set geth.image.imagePullPolicy=Always`,
```
tkporter: nit: comment

packages/celotool/src/lib/aks-fullnode.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  const replicaCount = getReplicaCount()
+  const resourceGroup = fetchEnv(envVar.AZURE_KUBERNETES_RESOURCE_GROUP)
+
+  await Promise.all(
+    range(replicaCount).map((i) => deallocateIP(`${celoEnv}-validators-${i}`, resourceGroup))
+  )
+}
+
+async function deallocateIP(name: string, resourceGroup: string) {
```
tkporter: Wdyt about moving this & `registerIPAddress` to a more common file, like `src/lib/azure.ts`?

packages/celotool/src/lib/aks-fullnode.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export async function removeHelmRelease(celoEnv: string) {
+  const releaseName = getReleaseName(celoEnv)
+  await deallocateIPs(celoEnv)
+  await removeGenericHelmChart(releaseName)
+}
+
+async function helmParameters(celoEnv: string, kubeNamespace: string) {
+  const staticIps = (await createStaticIPs(celoEnv)).join(',')
```
tkporter: What happens when this is run multiple times? Would it do nothing if the IPs already exist or does it throw an error?

packages/celotool/src/lib/aks-fullnode.ts
```diff
@@ -0,0 +1,118 @@
...
+  console.info(`helmParams: ${params}`)
+  return installGenericHelmChart(
+    kubeNamespace,
+    releaseName,
+    helmChartPath,
+    await helmParameters(celoEnv, kubeNamespace)
+  )
+}
+
+export async function upgradeFullNodeChart(celoEnv: string, reset: boolean = false) {
```
tkporter, line 46: If `AZURE_TX_NODES_COUNT` changes when upgrading, we should probably allocate/deallocate as needed. Depending on how `createStaticIPs` works then the case where it increases is covered, but if it goes down I think we should deallocate the appropriate IP addresses. I know for the GCP testnet helm chart we look at the existing number of statefulset replicas and then compare that to the new value from the .env file, we should probably do something similar here

packages/celotool/src/lib/aks-fullnode.ts
```diff
@@ -0,0 +1,118 @@
...
+  const [address] = await execCmdWithExitOnFailure(
+    `az network public-ip create --resource-group ${resourceGroup} --name ${name} --allocation-method Static --sku Standard --query publicIp.ipAddress -o tsv`
+  )
+  return address.trim()
+}
+
+async function deallocateIPs(celoEnv: string) {
+  console.info(`Deallocating static IPs on Azure for ${celoEnv}`)
+
+  const replicaCount = getReplicaCount()
```
tkporter: What do you think about relying on the statefulset replica count rather than the .env value? To cover the edge case where someone changes the .env value and then runs the destroy command

packages/celotool/src/lib/azure.ts
```diff
@@ -13,6 +12,18 @@ export async function switchToClusterFromEnv(
     await doCheckOrPromptIfStagingOrProduction()
   }
 
+  // Azure subscription switch
+  const tenantId = fetchEnv(envVar.AZURE_TENANT_ID)
+  let currentTenantId = null
+  try {
+    ;[currentTenantId] = await execCmd('az account show --query tenantId -o tsv')
```
tkporter, line 19: Is the semicolon intentional?

codyborn:
@jmrossy @aslawson, we can use this for the phone number service as well

packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -701,7 +701,7 @@ export async function resetAndUpgradeHelmChart(celoEnv: string, useExistingGenes
   await scaleResource(celoEnv, 'StatefulSet', privateTxNodesSetname, 0, true)
   await scaleResource(celoEnv, 'Deployment', bootnodeName, 0)
 
-  await deletePersistentVolumeClaims(celoEnv)
+  await deletePersistentVolumeClaims(celoEnv, persistentVolumeClaimsLabels)
   await sleep(10000)
```
codyborn, line 708: I imagine this is here to ensure it's deleted prior to upgrade. Is it possible to query the state of the volumeClaims explicitly before continuing?

packages/helm-charts/celo-fullnode/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
+{{ range $index, $e := until (int .Values.replicaCount ) }}
```
codyborn, line 1: I'm not too familiar with helm charts. Are these coded up by hand or do you use a tool for building these?

packages/helm-charts/common/values.yaml
```diff
@@ -1 +1,33 @@
...
+      memory: "512Mi"
+      cpu: "1000m"
+  debug: false
+  ping_ip_from_packet: false
+  in_memory_discovery_table: false
+  syncmode: full
+  gcmode: full
+  verbosity: 2
+  public_ips:
+  - 52.149.51.48
```
codyborn, line 23: Where does this IP come from?

packages/helm-charts/celo-fullnode/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+      cpu: "1000m"
+  debug: false
+  syncmode: full
+  gcmode: full
+  ping_ip_from_packet: false
+  in_memory_discovery_table: false
+  verbosity: 2
+  # Azure static ips
+  public_ips:
+  - 1.2.3.4
```
codyborn, line 67: Should these be internal IP addresses? How does the Oracle service reach the load balancer?

packages/helm-charts/common/values.yaml
jcortejoso: These are the static ips created in Azure. The `values.yaml` file in the chart root serves as default value for the chart variables if no values are provided. In this case, the `geth.public_ips` variable is overwritten by celotool, and the static ips are created in the function `createStaticIPs` in packages/celotool/src/lib/aks-fullnode.ts. I will update these ips to make more clear that these ips have to be replaced to some ips that belongs to the same Resource Group of your aks nodes.

packages/helm-charts/celo-fullnode/values.yaml
jcortejoso: These are static public ips in the same resource group as the aks nodes (also can be used same approach in other cloud providers like gce or aws), and after creating them, they are assigned to a public Load Balancer using the property `loadBalancerIP`. We use this approach to expose the celo-blockchain service using the load balancer ip, and we need to create the ips in advance because the celo-blockchain pod/process needs to know its assigned ip in advance to use in the `--nat` option.

packages/helm-charts/celo-fullnode/templates/service.yaml
jcortejoso: It is possible to create a chart skeleton with `helm create <chart_name>`, but it will require in most cases a lot of changes. Helm charts are in essence the kubernetes yamls as templates, using Go templates and the `sprig` go template library. Also I find very useful checking existing charts (in this repository or publics in https://hub.helm.sh/) to see examples of different template approaches.

packages/celotool/src/lib/aks-fullnode.ts
```diff
@@ -0,0 +1,118 @@
...
+  ]
+}
+
+async function createStaticIPs(celoEnv: string) {
+  console.info(`Creating static IPs on Azure for ${celoEnv}`)
+  const resourceGroup = fetchEnv(envVar.AZURE_KUBERNETES_RESOURCE_GROUP)
+  const replicaCount = getReplicaCount()
+
+  const staticIps = await Promise.all(
+    range(replicaCount).map((i) => registerIPAddress(`${celoEnv}-validators-${i}`, resourceGroup))
```
tkporter: nit: maybe pull out `${celoEnv}-validators-${i}` into a function that's also used by `deallocateIPs` to make changing it easier in the future

packages/helm-charts/celo-fullnode/templates/NOTES.txt
```diff
@@ -0,0 +1,19 @@
+1. Get the application URL by running these commands:
```
tkporter: I like this file, very good idea

packages/helm-charts/celo-fullnode/templates/ingress.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if .Values.ingress.enabled -}}
+{{- $fullName := include "celo-fullnode.fullname" . -}}
+{{- $ingressPath := .Values.ingress.path -}}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{ $fullName }}
+  labels:
+    app: {{ template "celo-fullnode.name" . }}
```
tkporter: Should we use `standard.short_labels` or `standard.labels` here?

packages/helm-charts/celo-fullnode/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: {{ template "celo-fullnode.fullname" $ }}-{{ $index }}
+  labels:
+{{ include "standard.labels" $  | indent 4 }}
+spec:
+  type: {{ $.Values.service.type }}
+{{- if $.Values.geth.use_static_ips }}
+  loadBalancerIP: {{ index $.Values.geth.public_ips $index -}}
+{{- end }}
+  ports:
+    - port: 30303
```
tkporter, line 16: I know we talked about this directly, but we should add UDP support as well to accommodate discovery

packages/helm-charts/celo-fullnode/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
...
+      targetPort: discovery-tcp
```
tkporter: nit: because discovery only happens on UDP this name should probably change

packages/helm-charts/common/templates/_helpers.tpl
```diff
@@ -28,3 +82,96 @@
...
+    value: "{{ join "," .Values.geth.public_ips }}"
+  - name: REPLICA_NAME
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.name
+  ports:
+  - name: discovery-udp
+    containerPort: 30303
+    protocol: UDP
+  - name: discovery-tcp
```
tkporter: nit: same thing, maybe we should call this something different

tkporter (commented):
Just some general comments-- we should also try to iron out the issues we were seeing where remote peers were connecting to the gateway IP

packages/helm-charts/celo-fullnode/templates/NOTES.txt
jcortejoso: It is created automatically with the chart skeleton. It is the info shown just after deploying or upgrading the chart, we can complete with some meaningful information.

packages/celotool/src/lib/aks-fullnode.ts
jcortejoso: Do nothing with the ips. `az network public-ip create ...` is idempotent so any time you run will create the ip with the name in the resourcegroup if not exists, or nothing it it already exists, and returns the public ip value as output in both cases (with exit code 0 if there is not any other error).

packages/celotool/src/lib/azure.ts
jcortejoso: So I was checking and is inserted automatically by my VSCode when saving. I confirmed with Pedro and he told me that is correct that `;`. It is used to "break" the newline so the `[]` cannot be considered as part of the line before (probably you already knew it).

packages/helm-charts/celo-fullnode/templates/ingress.yaml
jcortejoso: Really the ingress is the default ingress resource. It can be fully customized with parameters, but do you think is a good idea to keep it if we do not plan to use it right now? If not we can directly remove this file (an associated values in default values.yaml).

.env.scenario1
```diff
@@ -0,0 +1,184 @@
+# These are the default values that should apply to most environments, especially dev ones
```
tkporter: Is this file included on purpose?

packages/celotool/src/lib/azure.ts
tkporter: Didn’t know that, that’s cool!

packages/helm-charts/celo-fullnode/templates/statefullset.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: apps/v1beta2
```
tkporter: Small nit: file name should be statefulset

packages/helm-charts/common/templates/_helpers.tpl
```diff
@@ -28,3 +82,98 @@
...
+    {{- end }}
+    {{- if .proxy_allow_private_ip_flag | default false }}
+    ADDITIONAL_FLAGS="${ADDITIONAL_FLAGS} --proxy.allowprivateip"
+    {{- end }}
+  
+    geth \
+      --bootnodes=$(cat /root/.celo/bootnodeEnode) \
+      --light.serve 90 \
+      --light.maxpeers 1000 \
+      --maxpeers 1100 \
```
tkporter: Could we make `light.serve`, `light.maxpeers`, and `maxpeers` configurable? I imagine down the road these 3 might need to be changed depending on how many peers we want for the full nodes

tkporter (approved):
A couple small things, but LGTM!

packages/celotool/src/lib/helm_deploy.ts
jcortejoso: Maybe it could be check but not sure if we need this or add any extra security.
There is a double abstraction layer for volumes in Kuberntes. The `PersistentVolumeClaim` represents a request for a volume. The `PersistentVolume` object represents a mapping between the physical volume and can be created automatically when creating a PVC resource.
Depending on some configuration on the `PersistentVolume` the physical volume will be deleted or not if the pvc is removed (for the storage class we are using here it will be deleted if nobody changes it manually).
The good part here of this abstraction is that recreating the PVC also force to create a new PV resource (with different name), so avoid any collision. Also the kubernetes controller should be responsible of deleting the old pv and azure disk.

packages/helm-charts/common/templates/_helpers.tpl
```diff
@@ -172,8 +188,129 @@ kind: ConfigMap
...
   labels:
-{{ include "standard.labels" .  | indent 4 }}
+{{ include "common.standard.labels" .  | indent 4 }}
 data:
   networkid: {{ .Values.genesis.networkId | quote }}
-  genesis.json: {{ .Values.genesis.fileBase64 | b64dec | quote }}
+  genesis.json: {{ .Values.genesis.genesisFileBase64 | b64dec | quote }}
+{{- end -}}
+
+{{- define "common.celotool-container" -}}
```
tkporter: Nit: maybe a name that indicates this is generating values for a validator would make this easier to understand

packages/helm-charts/common/templates/_helpers.tpl
```diff
@@ -1,21 +1,74 @@
...
+  - -c
   args:
-  - "init"
-  - "/var/geth/genesis.json"
+  - |
+      mkdir -p /var/geth /root/celo
+      if [ "{{ .Values.genesis.useGenesisFileBase64 | default false }}" == "true" ]; then
+        cp -L /var/geth/genesis.json /root/.celo/
+      else
+        wget -O /root/.celo/genesis.json "https://www.googleapis.com/storage/v1/b/genesis_blocks/o/{{ .Values.genesis.network }}?alt=media"
```
tkporter, line 55: What do you think about having the genesis/bootnodeEnode be retrieved by celotool from GCS at deploy time? That way if a pod gets restarted and then for some reason these files are inaccessible it wouldn't affect the pod

jcortejoso: So it should be the case if Values.genesis.useGenesisFileBase64=true. In that case the `config` volume is mounted at `/var/geth` and the genesis is then copied to `/root/.celo`. For the bootnodeEnode, if useGenesisFileBase64=true, should be generated including the `common.celotool-container`, which creates here an unclear dependency. This configuration is currently used for the testnet chart.
Maybe we can generate the bootenode address from celotool and pass it in the same configmap as the genesis file. I think that approach fits better now.
If useGenesisFileBase64!=true, in that case both the genesis and the bootenode is downloaded from gcp. It adds the dependency to download the files from gcp when the container starts, but also I think it helps to decouple the chart from celotool and helps portability. The aks fullnode chart is using this configuration.

PR #3353, opened 2020-04-08
Fix election:activate command in baklava doc

hqueue:
## Description

Fix election:activate command in baklava doc

Use vote signer address instead of releasegold contract address.

## Other changes

N/A

## Tested

No code change.

## Related issues

N/A

## Backwards compatibility

N/A

## Commit Message
Fix election:activate command in baklava doc

Use vote signer address instead of releasegold contract address.


timmoreton:
👍 Thanks for the contribution!

timmoreton (approved):

PR #3361, opened 2020-04-09
Adding attestation metadata verification by address and signer accounts

lucasege:
## Description

Allowing metadata to be signed by an account's authorized signers. Effectively this just involved adding to the verification side since you could always provide whatever `--from` address you desired. This change first checks the signature vs the account address, then against each of the account's signer addresses.

## Other changes

Changing a bunch of fns to async so that I can access async contract data in the metadata verification. Also changed these fns to pass around the `kit` so I can access the contracts.

## Tested

Added more unit tests to test this new functionality, and verified that overwriting a signer works as expected.

## Related issues

- Fixes #[issue number here]

## Backwards compatibility

Yes


packages/contractkit/src/identity/metadata.test.ts
```diff
@@ -4,31 +4,64 @@ import { NativeSigner } from '@celo/utils/lib/signatureUtils'
...
+    const validatorMetadata = IdentityMetadataWrapper.fromEmpty(address)
+    const attestationMetadata = IdentityMetadataWrapper.fromEmpty(address)
+    const accounts = await kit.contracts.getAccounts()
+    const voteSigner = ACCOUNT_ADDRESSES[2]
+    const validatorSigner = ACCOUNT_ADDRESSES[3]
+    const attestationSigner = ACCOUNT_ADDRESSES[4]
+    await accounts.createAccount().send({ from: address })
+    const testSigner = async (signer: any, action: any, metadata: any) => {
+      const pop = await accounts.generateProofOfKeyPossession(address, signer)
+      if (action === 'vote') {
```
nambrot, line 41: Optional: You could create a function map and then just call it that way. I.e.

```typescript
const functionByAction = { 'vote': acounts.authorizeVoteSigner }
await (await functionByAction[action]).send({ from: signer }))
```

packages/contractkit/src/identity/metadata.ts
```diff
@@ -34,19 +35,41 @@ export class IdentityMetadataWrapper {
...
   }
 
-  static fromFile(path: string) {
-    return this.fromRawString(readFileSync(path, 'utf-8'))
+  static fromFile(path: string, kit: ContractKit) {
+    return this.fromRawString(readFileSync(path, 'utf-8'), kit)
   }
 
-  static fromRawString(rawData: string) {
+  static async verifySigner(kit: ContractKit, hash: any, signature: any, metadata: any) {
```
nambrot, line 50: Optional Nit: It might be nice to be consistent in whether kit is the first or the last argument

packages/contractkit/src/identity/metadata.ts
```diff
@@ -34,19 +35,41 @@ export class IdentityMetadataWrapper {
...
+    // First try to verify on account's address
+    if (!verifySignature(hash, signature, metadata.address)) {
+      // If this fails, signature may still be one of `address`' signers
+      const accounts = await kit.contracts.getAccounts()
+      if (await accounts.isAccount(metadata.address)) {
+        const signers = [
```
nambrot: It would be faster to use Promise.all here, otherwise you are making these calls serially

packages/contractkit/src/identity/metadata.ts
```diff
@@ -34,19 +35,41 @@ export class IdentityMetadataWrapper {
...
+          await accounts.getAttestationSigner(metadata.address),
+          await accounts.getVoteSigner(metadata.address),
+          await accounts.getValidatorSigner(metadata.address),
+        ]
+        for (const signer of signers) {
```
nambrot: Optional Nit: What do you think about a more functional approach:

```
return signers.some(_ => verifySignature(hash, signature, _))
```

packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -18,13 +19,14 @@ import { ClaimTypes } from './types'
 export async function verifyClaim(
   claim: Claim,
   address: string,
-  metadataURLGetter: MetadataURLGetter
+  metadataURLGetter: MetadataURLGetter,
+  kit: ContractKit
```
nambrot: metadataURLGetter was basically me trying to avoid having to pass kit, but since we do that, we could get rid of metadataURLGetter?

nambrot (commented):
Just some small comments 🙏

packages/contractkit/src/identity/metadata.test.ts
lucasege: I tried something like this initially, my version was just via passing the fn as an argument, i.e. rather than passing `vote` I passed `accounts.authorizeVoteSigner`, but I can't get that or this suggestion to work - it complains about `kit` being `undefined` or similar issues, so I've stopped trying now since it seems low priority. let me know if you know what may be the problem and an easy fix

nambrot (approved):
🙏

PR #3347, opened 2020-04-08
New oracle helm chart

tkporter:
## Description

This PR moves the old cron-job oracle to be `mock-oracle`. In its old place, a new helm chart for an oracle statefulset has been introduced. All this does is run the oracles in a statefulset on Azure.

Introduced/repurposed some env vars:

```
MOCK_ORACLE_CRON_SCHEDULE - the cron schedule for the old mock oracle
MOCK_ORACLE_DOCKER_IMAGE_REPOSITORY - docker repo for old mock oracle
MOCK_ORACLE_DOCKER_IMAGE_TAG - tag for the image of the old mock oracle

ORACLE_DOCKER_IMAGE_REPOSITORY - docker repo for the new azure based oracle
ORACLE_DOCKER_IMAGE_TAG - tag for the image of the new azure based oracle
ORACLES - number of oracles to deploy in new azure
```

Future work will likely involve pod AAD identities, keyvault integration, and splitting oracle deployments across two kubernetes clusters.

## Other changes

n/a

## Tested

Deployed/destroyed new oracles and `mock-oracles`

## Related issues

n/a

## Backwards compatibility

Instead of using `celotooljs deploy initial|destroy oracle` for the old cronjob oracle, it's now `celotooljs deploy initial|destroy mock-oracle`

## Commit Message
New oracle helm chart

PR #3364, opened 2020-04-09
Oracle MVP infrastructure

tkporter:
## Description

Note: this builds upon https://github.com/celo-org/celo-monorepo/pull/3347, so the easiest way to review this PR until that one is merged is by looking at the diff between this branch and that PR's branch: https://github.com/celo-org/celo-monorepo/compare/trevor/oracle-helm-chart...trevor/oracle-mvp-infra?expand=1

This PR is the MVP for running the new oracle using private keys generated using the mnemonic. `rc0` and `baklava` will soon have 10 of these addresses added as oracles. Alfajoresstaging currently has 1 oracle address whitelisted that I was able to test this on.

Private keys for the oracles will be generated and put into a secret, and mounted as volumes onto the oracle client pods. Each pod will use their corresponding private key. Currently, only forno is used. This will change once https://github.com/celo-org/celo-monorepo/pull/3348 is merged in

## Other changes

n/a

## Tested

Ran on alfajoresstaging

## Related issues

n/a

## Backwards compatibility

Yes. The only thing to be aware of when deploying oracles is the necessary env variables must be present

## Commit Message
Oracle MVP infrastructure


jcortejoso (approved):
lgtm 🚀

PR #3369, opened 2020-04-09
[Wallet] Protect initial view of a backup key with PIN

i1skn:
## Description

Protect initial view of a backup key with PIN as well as just to see it in the future

## Tested

Android device

## Backwards compatibility

Yes

## Commit Message
[Wallet] Protect initial view of a backup key with PIN


annakaz (approved):
🚢

PR #3372, opened 2020-04-09
Docs: Ledger Setup

clrblmt:
Getting started docs for how to set up Ledger Nano S and X and get addresses.


packages/docs/getting-started/using-a-ledger-wallet.md
```diff
@@ -0,0 +1,111 @@
...
+
+![](https://storage.googleapis.com/celo-website/docs/ledger-settings.png)
+
+Click on the **Experimental features** menu.
+
+Turn on **Developer mode.** 
+
+![](https://storage.googleapis.com/celo-website/docs/ledger-settings-dev-mode.png)
+
+Exit the Settings menu by clicking on **Manager** on the left hand side bar.
```
gastonponti, line 29: @kobigurk correct me if I'm wrong, but we still need to set the provider to `4`, right?

kobigurk: Nope! It's published.

PR #3373, opened 2020-04-09
[Wallet] Fix invite redemption gas handling and gas price calculation

jmrossy:
## Description

Calculate gas for withdrawing from invite codes more precisely

## Other changes

Improve organization of some gas related things

## Tested

As part of alfajroes release prep

## Backwards compatibility

Yes

## Commit Message
[Wallet] Fix invite redemption gas handling and gas price calculation

The app now calculates invite code withdrawal fees dynamically and also inflates gas prices to match what Geth does.

### DIFF
--- a/packages/mobile/src/invite/saga.ts
+++ b/packages/mobile/src/invite/saga.ts
@@ -32,13 +32,14 @@ import {
 } from 'src/invite/actions'
 import { createInviteCode } from 'src/invite/utils'
 import { navigateHome } from 'src/navigator/NavigationService'
-import { getSendTxGas } from 'src/send/saga'
+import { getSendFee, getSendTxGas } from 'src/send/saga'
 import { fetchDollarBalance, transferStableToken } from 'src/stableToken/actions'
-import { createTransaction, fetchTokenBalanceInWeiWithRetry } from 'src/tokens/saga'
+import { createTokenTransferTransaction, fetchTokenBalanceInWeiWithRetry } from 'src/tokens/saga'
 import { generateStandbyTransactionId } from 'src/transactions/actions'
 import { waitForTransactionWithId } from 'src/transactions/saga'
 import { sendTransaction } from 'src/transactions/send'
 import { getAppStoreId } from 'src/utils/appstore'
+import { divideByWei } from 'src/utils/formatting'
 import Logger from 'src/utils/Logger'
 import { addLocalAccount, getContractKit } from 'src/web3/contracts'
 import { getConnectedUnlockedAccount, getOrCreateAccount, waitWeb3LastBlock } from 'src/web3/saga'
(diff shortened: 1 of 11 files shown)


packages/mobile/src/web3/utils.ts
```diff
@@ -1,9 +1,19 @@
...
+import BigNumber from 'bignumber.js'
+import { GAS_INFLATION_FACTOR } from 'src/config'
 import Logger from 'src/utils/Logger'
 import { getContractKit } from 'src/web3/contracts'
+import { Tx } from 'web3-core'
+import { TransactionObject } from 'web3-eth'
 
 const TAG = 'web3/utils'
 
+// Estimate gas taking into account the cofigured inflation factor
```
jeanregisser: nit: typo `configured`

packages/mobile/src/invite/saga.ts
```diff
@@ -353,19 +354,39 @@ export function* withdrawFundsFromTempAccount(
...
+  const sendTokenFee = divideByWei(sendTokenFeeInWei).times(5)
+
+  if (sendTokenFee.isGreaterThanOrEqualTo(tempAccountBalance)) {
+    throw new Error('Fee is too large for amount in temp wallet')
+  }
+
+  const netSendAmount = tempAccountBalance.minus(sendTokenFee)
+  Logger.debug(
+    TAG + '@withdrawFundsFromTempAccount',
+    `Withdrwaing net amount of ${netSendAmount.toString()}`
```
jeanregisser: nit: typo `Withdrawing`

packages/mobile/src/invite/saga.ts
```diff
@@ -353,19 +354,39 @@ export function* withdrawFundsFromTempAccount(
...
     recipientAddress: newAccount,
+    amount: tempAccountBalance,
     comment: SENTINEL_INVITE_COMMENT,
-    // TODO: appropriately withdraw the balance instead of using gas fees will be less than 1 cent
-    amount: tempAccountBalance.minus(0.01).toString(),
   })
+  const sendTokenFee = divideByWei(sendTokenFeeInWei).times(5)
+
+  if (sendTokenFee.isGreaterThanOrEqualTo(tempAccountBalance)) {
+    throw new Error('Fee is too large for amount in temp wallet')
```
jeanregisser, line 372: nice! this way we'll see the error more easily :+1:

Wondering, shouldn't we also check that the amount is enough to do verification as well?

cmcewen (approved):

jeanregisser (approved):
Much better 🎉 👍 
Left one question.

packages/mobile/src/invite/saga.ts
jmrossy: Good idea! It's a bit tough to know exactly what verification will cost but I suppose we already hard code an estimate in choosing the invite amount. I'll do this in a separate PR

PR #3377, opened 2020-04-10
Gastonponti/ledger address validation

gastonponti:
## Description
Adding the possibility to require ledger address validation for different cases


packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -100,21 +116,47 @@ export class LedgerWallet implements Wallet {
...
+        derivationPath,
+        validated: validationRequired,
+      })
     }
     return addresses
   }
 
+  private validationRequired(initialized: boolean, validated?: boolean): boolean {
+    if (!initialized) {
+      return this.ledgerAddressValidation === AddressValidation.onlyInitialization
```
codyborn: Just to clarify, `everyTransaction` and `oncePerAddress` should not validate upon initialization? `onlyInitialization` makes it sound like the other ones are upon initialization AND some other condition.

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -100,21 +116,47 @@ export class LedgerWallet implements Wallet {
...
+        return false
+      }
+      case AddressValidation.everyTransaction: {
+        return true
+      }
+      case AddressValidation.oncePerAddress: {
+        return !validated
+      }
+    }
+    return false
```
codyborn: We should throw an exception here since it should always be defined

packages/contractkit/src/wallets/ledger-wallet.test.ts
```diff
@@ -232,4 +243,107 @@ describe('LedgerWallet class', () => {
...
+        wallet = new LedgerWallet(undefined, undefined, AddressValidation.never)
+        mockForceValidation = jest.fn((): void => {
+          // do nothing
+        })
+        mockLedger(wallet, mockForceValidation)
+      })
+
+      it("won't validate", async () => {
+        await wallet.init({})
+        expect(mockForceValidation.mock.calls.length).toBe(0)
```
codyborn, line 261: Nice!

packages/contractkit/src/wallets/ledger-wallet.test.ts
```diff
@@ -232,4 +243,107 @@ describe('LedgerWallet class', () => {
       })
     })
   })
+
+  describe('asking for addresses validations', () => {
+    describe('never', () => {
+      const knownAddress = ACCOUNT_ADDRESS1
+
+      beforeEach(() => {
```
codyborn, line 251: Since there's only one test after each of these beforeEach clauses.. does it make sense to just include the content in the test case?

packages/contractkit/src/wallets/ledger-wallet.test.ts
```diff
@@ -232,4 +243,107 @@ describe('LedgerWallet class', () => {
...
+        })
+        mockLedger(wallet, mockForceValidation)
+      })
+
+      it('will validate only in the first transaction of the address', async () => {
+        await wallet.init({})
+        expect(mockForceValidation.mock.calls.length).toBe(0)
+        await wallet.signPersonalMessage(knownAddress, ACCOUNT_ADDRESS_NEVER)
+        expect(mockForceValidation.mock.calls.length).toBe(1)
+        await wallet.signPersonalMessage(knownAddress, ACCOUNT_ADDRESS_NEVER)
```
codyborn, line 328: Can you add another check in this test that a new address increments again:
```
await wallet.signPersonalMessage(otherAddress, ACCOUNT_ADDRESS_NEVER)
expect(mockForceValidation.mock.calls.length).toBe(2)
```

codyborn (approved):

packages/contractkit/src/wallets/ledger-wallet.ts
gastonponti: could it be something like
`everyTransaction`
`firstTransactionPerAddress`
`initializationOnly`
or is still unclear?

gastonponti: The thing is that if I retrieve 10 addresses, but only use one, the `onlyInitialization` will need to validate the 10...
 I could add some comments too

packages/contractkit/src/wallets/ledger-wallet.ts
gastonponti: Yes, before the onlyInitialization was part of the switch and this was a valid option (but not now, I'll throw an exception)

packages/contractkit/src/wallets/ledger-wallet.test.ts
gastonponti: I always saw the before each as the way to initialise the context that you are going to test. And every test as the actions that have to be asserted later

packages/contractkit/src/wallets/ledger-wallet.test.ts
gastonponti: Of course!

packages/contractkit/src/wallets/ledger-wallet.test.ts
codyborn: Sounds good

PR #3381, opened 2020-04-10
[Doc] Unlock an account on proxy and new celostats address on baklava

jcortejoso:
## Description
Currently celostats require an unlocking account on the proxy to report correctly. This account can be any and do not have to be related in any way with any other address.
Additionally, updated the celostats reporting address. Websocket implementations are not required to follow redirects (https://tools.ietf.org/html/rfc6455) and seems that geth is not able to follow the redirects when setting up the websocket connection to celostats.


timmoreton (approved):

abellinii:
Hi Guys,

Is there a reason the creation of a new account is different here than from the rest of the running a validator document?

``docker run --name celo-proxy-password -it --rm  -v $PWD:/root/.celo $CELO_IMAGE account new --password /root/.celo/.password``

and not 

``docker run -v $PWD:/root/.celo --rm -it $CELO_IMAGE account new
``

PR #3386, opened 2020-04-10
Fix bug in exchange CLI

asaj:
## Description

This PR fixes a bug introduced in the exchange CLI when I tried to optimize it to batch both the approval tx and the exchange tx in the same block, by going back to the old, slower way (but using `increaseAllowance` rather than `approve`).

## Other changes

None
## Tested

Manual verification on RC0 that exchanges work in both directions.

## Related issues

None
## Backwards compatibility

Yes


packages/cli/src/commands/exchange/gold.ts
```diff
@@ -35,9 +35,12 @@ export default class ExchangeGold extends BaseCommand {
     const goldToken = await this.kit.contracts.getGoldToken()
     const exchange = await this.kit.contracts.getExchange()
 
-    await displaySendTx('approve', goldToken.approve(exchange.address, sellAmount.toFixed()))
```
aslawson, line 38: it seems like before you were calling both approve and increaseAllowance but they seem to both update the same `allowance` field.  so wouldn't the allowance become twice as much?  why would we get an error that says the allowance isn't high enough?

aslawson (approved):

PR #3387, opened 2020-04-10
[ReleaseGold] Simple fix for cli bug

lucasege:
## Description

Fixed a bug that was breaking `releasegold:show` because the `ReleaseGold` contracts deployed on `baklava` and `rc0` were deployed before the `canExpire` update to RG, so the mismatch between that and locally compiled master contracts was causing contractkit to error. This just adds a try/catch since the relevant info (`revocationInfo`) isn't relevant to current players on `baklava` anyway, but this bug would annoy them/me.

## Other changes
None

## Tested

Ran myself

## Backwards compatibility

Yes

## Commit Message
_If using automerge, specify desired commit title_

_And desired commit message body_


packages/contractkit/src/wrappers/ReleaseGold.ts
```diff
@@ -149,13 +149,24 @@ export class ReleaseGoldWrapper extends BaseWrapper<ReleaseGold> {
...
+    try {
+      const revocationInfo = await this.contract.methods.revocationInfo().call()
+      return {
+        revocable: revocationInfo.revocable,
+        canExpire: revocationInfo.canExpire,
+        releasedBalanceAtRevoke: valueToBigNumber(revocationInfo.releasedBalanceAtRevoke),
+        revokeTime: valueToInt(revocationInfo.revokeTime),
+      }
+    } catch (_) {
+      // This error is caused by a mismatch between the deployed contract and the locally compiled version.
```
asaj, line 161: Can we be sure to track this hack in an issue so it can be removed?

asaj (approved):

packages/contractkit/src/wrappers/ReleaseGold.ts
lucasege: Yup will do

PR #3392, opened 2020-04-12
Fix bad ethstats in url for proxy in docs

lahabana:
## Description

The ethstats url in both the proxy and the validator should be the same.
It seems like a "-proxy" suffix was left from a previous version.


asaj:
Thanks @lahabana :)

PR #3394, opened 2020-04-12
Update celo-dapp-gallery.md

adrianmitter:
## Description

We have developed a prototype for the Celo Camp 2020 and would like to briefly describe it in the DApp Gallery (as recommended by the camp mentors).

## Other changes

No other changes, since it only affects a single documentation file.


critesjosh (approved):
looks good, thanks!

PR #3367, opened 2020-04-09
[contractkit] SortedOracles: check for and remove expired reports

yerdua:
## Description

In order to facilitate cleaning up after oracles that have stopped reporting, this PR adds two functions to the SortedOracles wrapper: `removeExpiredReports` and `isOldestReportExpired`.

## Other changes

Fixed a misleading typo in a function comment

## Tested

This PR includes tests for the new functions.

## Backwards compatibility

Yes! Functions are added to the wrapper, not modified or removed.

## Commit Message
[contractkit] SortedOracles: check for and remove expired reports

In order to facilitate cleaning up after oracles that have stopped reporting, two functions are added to the SortedOracles wrapper: `removeExpiredReports` and `isOldestReportExpired`.

PR #3397, opened 2020-04-13
Add oracle:remove-expired-reports CLI command

asaj:
## Description

New command to remove expired oracle reports and minor change to existing oracle:rates command.

## Other changes

None
## Tested

Manually

## Related issues

None
## Backwards compatibility

Yes
## Commit Message
_If using automerge, specify desired commit title_

_And desired commit message body_


packages/cli/src/commands/oracle/rates.ts
```diff
@@ -25,12 +25,13 @@ export default class GetRates extends BaseCommand {
     const res = this.parse(GetRates)
     const sortedOracles = await this.kit.contracts.getSortedOracles()
 
-    const rates = await sortedOracles.getRates(res.args.token)
+    const reports = await sortedOracles.getReports(res.args.token)
```
yerdua, line 28: Should the command also be renamed? `oracle:rates` -> `oracle:reports`

packages/contractkit/src/wrappers/SortedOracles.ts
```diff
@@ -105,6 +109,12 @@ export class SortedOraclesWrapper extends BaseWrapper<SortedOracles> {
     valueToBigNumber
   )
 
+  removeExpiredReports = proxySend(
+    this.kit,
+    this.contract.methods.removeExpiredReports,
+    tupleParser(stringIdentity, valueToString)
+  )
+
```
yerdua: This function was added in another PR that merged today: https://github.com/celo-org/celo-monorepo/pull/3367

The main difference is that it takes the name of the token, rather than the address, to be consistent with the rest of the functions here.

packages/cli/src/commands/oracle/remove-expired-reports.ts
```diff
@@ -0,0 +1,43 @@
...
+    'remove-expired-reports StableToken --from 0x8c349AAc7065a35B7166f2659d6C35D75A3893C1',
+    'remove-expired-reports --from 0x8c349AAc7065a35B7166f2659d6C35D75A3893C1',
+  ]
+
+  async run() {
+    const res = this.parse(RemoveExpiredReports)
+    this.kit.defaultAccount = res.flags.from
+    const sortedOracles = await this.kit.contracts.getSortedOracles()
+    const reports = await sortedOracles.getReports(res.args.token)
+    const tokenAddress = await this.kit.registry.addressFor(res.args.token)
```
yerdua: With the current master version of the `SortedOracles` wrapper, you can skip this address lookup.

packages/cli/src/commands/oracle/rates.ts
asaj: Yes, good call

packages/contractkit/src/wrappers/SortedOracles.ts
asaj: Okay cool, I'll use that one instead

packages/contractkit/src/wrappers/SortedOracles.ts
```diff
@@ -124,9 +124,12 @@ export class SortedOraclesWrapper extends BaseWrapper<SortedOracles> {
    */
   async removeExpiredReports(
     token: CeloToken,
-    numReports: number
+    numReports?: number
   ): Promise<CeloTransactionObject<void>> {
     const tokenAddress = await this.kit.registry.addressFor(token)
+    if (!numReports) {
+      numReports = (await this.getReports(token)).length - 1
+    }
```
yerdua, line 132: Nice! This will simplify the logic in the oracle client as well. Other than the case when _all_ reports are expired, it's hard to imagine a scenario where we don't want to remove all the expired ones. So, this seems like the best default behavior.

yerdua (approved):

PR #3403, opened 2020-04-13
[ContractKit] Ledger token data

gastonponti:
## Description

Sends the contract data to the ledger so it can be shown properly 

## Tested

Manually


packages/contractkit/src/wallets/ledger-utils/data.ts
```diff
@@ -0,0 +1 @@
+export = 'AAAAaARjVVNES4TC75SidNv4Pi8ewWCEVsm2LZYAAAASAACcuDBFAiEAqkKjjSN4G+snqF6HY3oxwh5vmslOh/gZVG3MGg5vgfgCIGuGPU8jhvxdXF4bMSQc4vLs2iXgWzzj/1fOLX2kmDoTAAAAaARjR0xERPQ06Doxefzt4olBs6gZU/tXUhcAAAASAACcuDBFAiEAlnEhhgsbClErGxR6SmBTuGEwIE3I+gofDEufqkB6+ewCIFbdxoPuDvl6mb5xzYMM0cCeJwRvaOT7/jTj0RWHIWOZAAAAaARjVVNEpWETGhyKwlkl+4SLykWnSvYeWjgAAAASAACu8jBFAiEAxVAZSsitVIBd7Ov1wGYgZGVI68gA1tlFkPDaTzAppEACIAlPr01KDlmG1pUo7hd4nD4rN2qLdeJaT8sDk23cLuS5AAAAZwRjR0xE8ZSv31CwPmm9fQV8GqnhDJlU5MkAAAASAACu8jBEAiBkJSVHwDoQclm3VNsr1LB0JrXIUy8hQJEBDZJeKS3gsQIgNfXfisKOh/yjw5HM0kv6zYyv3Ea+XI9wVPDQ4OtaPig='
```
codyborn, line 1: How did you generate this? Does it include addresses for Alfajores, Baklava, and Mainnet?

packages/contractkit/src/wallets/ledger-utils/tokens.ts
```diff
@@ -0,0 +1,83 @@
...
+  signature: Buffer
+  data: Buffer
+}
+
+export interface API {
+  byContract: (arg0: string) => TokenInfo | null | undefined
+  list: () => TokenInfo[]
+}
+
+const asContractAddress = (addr: string) => {
```
codyborn: `normalizeAddressWith0x()` does this

packages/contractkit/src/wallets/ledger-utils/tokens.ts
```diff
@@ -0,0 +1,83 @@
...
+      const entry: TokenInfo = {
+        ticker,
+        contractAddress,
+        decimals,
+        chainId,
+        signature,
+        data: item,
+      }
+      entries.push(entry)
+      byContract[[contractAddress, chainId].join('-')] = entry
```
codyborn: It'd be useful to wrap this in a function to prevent errors in the future. `getContractKey(address: string, chainId: number)` or something.

packages/contractkit/src/wallets/ledger-utils/data.ts
gastonponti: Alfajores and Baklava for now.
@kobigurk does the magic and I'm just adding it, haha.
Basically when we add some support for contracts to the ledger, we will need to change this file too. I thought of retrieving this info from somewhere else, but we could have some security issues. And we could also extract this as another package, but don't seems to be really necessary for now.
We could add in the future a way to change this automatically (@kobigurk, like when we generate the contract's classes)

packages/contractkit/src/wallets/ledger-utils/tokens.ts
gastonponti: Yes nice catch!

packages/contractkit/src/wallets/ledger-utils/tokens.ts
gastonponti: Great!

packages/contractkit/src/wallets/ledger-utils/data.ts
kobigurk: We have a private repo with code that generates it, will share.

codyborn: Sounds good.. thanks for the explanation!

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -301,4 +325,15 @@ export class LedgerWallet implements Wallet {
...
     throw error
   }
+
+  private compareVersionWithTheLedgerApp(version: string): number {
+    const numberV = this.stringVersionToNumber(version)
+    const numberAppV = this.stringVersionToNumber(this.appConfiguration.version)
+    return numberV < numberAppV ? -1 : numberAppV === numberV ? 0 : 1
+  }
+
+  private stringVersionToNumber(version: string): number {
```
codyborn: Can be static

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -301,4 +325,15 @@ export class LedgerWallet implements Wallet {
     }
     throw error
   }
+
+  private compareVersionWithTheLedgerApp(version: string): number {
```
codyborn: To make the order more explicit can you take in two string params and make this static. Intuitively I'm not sure what -1 means since the order isn't strictly defined.

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -185,6 +200,15 @@ export class LedgerWallet implements Wallet {
     try {
       const rlpEncoded = rlpEncodedTx(txParams)
       const path = await this.getDerivationPathFor(txParams.from!.toString())
+      if (this.compareVersionWithTheLedgerApp('1.0.2') <= 0) {
```
codyborn: Can we store '1.0.2' as a constant somewhere?

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -326,13 +332,19 @@ export class LedgerWallet implements Wallet {
...
+   *  1: version1 > version2
+   */
+  static compareLedgerAppVersions(version1: string, version2: string): number {
+    const numberV1 = this.stringVersionToNumber(version1)
+    const numberV2 = this.stringVersionToNumber(version2)
+    return numberV1 < numberV2 ? -1 : numberV1 === numberV2 ? 0 : 1
   }
 
-  private stringVersionToNumber(version: string): number {
+  static stringVersionToNumber(version: string): number {
```
codyborn: Looks good! We can still keep this as `private`

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -113,6 +120,15 @@ export class LedgerWallet implements Wallet {
     }
   }
 
+  private async retrieveAppConfiguration() {
+    this.appConfiguration = await this.ledger!.getAppConfiguration()
+    if (!this.appConfiguration.arbitraryDataEnabled) {
+      console.log(
```
codyborn: console.warn()

codyborn (approved):

PR #3406, opened 2020-04-13
[ReleaseGold] Fix bug in RG.transfer_dollars

lucasege:
## Description

Fixes bug causing `transfer-dollars` to not work due to a backwards ternary operator

## Other changes

Typo

## Tested

Working on adding `transfer-dollars.test.ts` but this PR is high priority so it will come later. Tested manually and command works as intended now.

## Backwards compatibility

yes


packages/cli/src/commands/releasegold/set-beneficiary.ts
```diff
@@ -12,38 +13,52 @@ export default class SetBeneficiary extends ReleaseGoldCommand {
...
+        console.info('Aborting due to user response')
+        process.exit(0)
+      }
     }
+    // Multisig is the owner of the ReleaseGold contract
     const owner = await this.releaseGoldWrapper.getOwner()
     const releaseGoldMultiSig = await this.kit.contracts.getMultiSig(owner)
+    const releaseOwner = await this.releaseGoldWrapper.getReleaseOwner()
+    const beneficiary = await this.releaseGoldWrapper.getBeneficiary()
+    console.log(releaseOwner, beneficiary)
```
asaj: remove?

packages/cli/src/commands/releasegold/set-beneficiary.ts
```diff
@@ -12,38 +13,52 @@ export default class SetBeneficiary extends ReleaseGoldCommand {
...
     const tx = this.releaseGoldWrapper.setBeneficiary(newBeneficiary)
-    const multiSigTx = await releaseGoldMultiSig.submitOrConfirmTransaction(
+    const multiSigTxReleaseOwner = await releaseGoldMultiSig.submitOrConfirmTransaction(
+      this.contractAddress,
+      tx.txo
+    )
+    await displaySendTx<any>('setBeneficiaryOneOfTwo', multiSigTxReleaseOwner, {
+      from: releaseOwner,
```
asaj: In what circumstances would you expect the user to have both keys accessible? IMO this command should take a `--from` flag, add a "check" that the "from" flag is a signatory on the multisig, and then submit or confirm a single multisig tx from that "from" address

asaj (approved):

PR #3407, opened 2020-04-14
Small e2e transfer test fix

asaj:
## Description

Assertion that a token is frozen is failing, perhaps b/c we are not waiting for the freeze tx to be mined. This PR moves to using contractkit to ensure the tx is mined. 

## Other changes

- Removes unused code from governance e2e test
- Adds some logging to help diagnose flakes

## Tested

No

## Related issues

None

## Backwards compatibility

Yes


mrsmkl (approved):
Looking good.

PR #3417, opened 2020-04-14
Setters for the tobin tax

rcroessmann:
## Description

Creates setters for the tobin tax and the ratio at which the tobin tax kicks in.

## Tested

No


packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -109,6 +112,15 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
     emit TobinTaxSet(value);
   }
 
+  /**
+   * @notice Sets the tobin tax.
+   * @param value The tobin tax.
+   */
+  function setTobinTax(uint256 value) public onlyOwner {
```
rcroessmann: Copy pasted this in again by accident, removed this duplicate block in next commit.

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -438,15 +469,14 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
...
   /**
    * @notice Computes a tobin tax based on the reserve ratio.
-   * @return The numerator of the tobin tax amount, where the denominator is 1000.
+   * @return The tobin tax expresesed as a fixidity fraction.
    */
   function computeTobinTax() private view returns (FixidityLib.Fraction memory) {
-    // The protocol calls for a 0.5% transfer tax on Celo Gold when the reserve ratio <= 2.
     FixidityLib.Fraction memory ratio = FixidityLib.wrap(getReserveRatio());
-    if (ratio.gte(FixidityLib.newFixed(2))) {
+    if (ratio.gte(FixidityLib.wrap(tobinTaxReserveRatio))) {
```
yorhodes, line 476: any reason why `tobinTaxReserveRatio` isn't also a Fixidity frac?

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -438,15 +469,14 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
...
       return FixidityLib.wrap(0);
     } else {
-      return FixidityLib.wrap(TOBIN_TAX_NUMERATOR);
+      return FixidityLib.wrap(tobinTax);
```
yorhodes, line 479: ditto with `tobinTax`? we would save gas by computing this in the setters

yorhodes (approved):

packages/protocol/contracts/stability/Reserve.sol
yorhodes: in fact we do compute it in the setter but don't store the value

yorhodes: (mixed up `wrap` and `newFixed` but we would still save some albeit minimal amt of gas

packages/protocol/contracts/stability/Reserve.sol
rcroessmann: Given the timing, I tried to stay as close as possible to the previous implementation.

packages/protocol/contracts/stability/Reserve.sol
rcroessmann: See comment above.

PR #3424, opened 2020-04-15
Fix celocli accounts:proof-of-possession command

sharppanda:
## Description

When trying command celocli accounts:proof-of-possession, celocli reports that it does not exists..
I think it should be celocli account:proof-of-possession instead

## Tested

celocli account:proof-of-possession works

## Commit Message
Fix typo in  celocli accounts:proof-of-possession command


asaj (approved):

asaj:
Thanks @sharppanda! Looks like there's a merge conflict so merging this fix in https://github.com/celo-org/celo-monorepo/pull/3479

PR #3430, opened 2020-04-15
Adding BucketsUpdated event.

rcroessmann:
## Description

Adds a BucketsUpdated event to the exchange which is emitted whenever the exchange buckets are updated.

## Tested

No


packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -303,6 +304,7 @@ contract Exchange is IExchange, Initializable, Ownable, UsingRegistry, Reentranc
       lastBucketUpdate = now;
 
       (goldBucket, stableBucket) = getUpdatedBuckets();
+      emit BucketsUpdated(goldBucket, stableBucket);
```
asaj, line 307: Can you add a unit test that confirms this event is emitted when expected?

asaj (approved):

Issue #3330, opened 2020-04-07
User Education screens around tx limits


jmrossy:
### Expected Behavior

1. When user hits send if the amount entered will put them over the daily limit show the red banner (see figma) 

2. on onboarding the legal disclaimers are now on their own screen

Part of #3063

### Current Behavior

1. no tx limit warnings

2.disclaimers on same screen as the user info

nityas:
designs here: https://www.figma.com/file/PyK5VI7mE2fpHLeRPL0yxB/Transaction-Limit?node-id=0%3A1

cc @aaronmgdr @jmrossy @coreycelo

jmrossy:
@nityas is that the right link? Don't see anything about tx limits

nityas:
Oops! Updated with the correct link

aaronmgdr:
This about the red banner that shows how much of daily limit has been used on  (each or just when approaching limit?) send?

jmrossy:
@aaronmgdr This is just about the new screens. @etuleu is working on the underlying changes in the saga code to enforce the limit

PR #3434, opened 2020-04-15
[wallet] new Terms Screen

aaronmgdr:
## Description

Creates a new Terms Screen and moves disclaimer onto it

## Other changes

fixes mock `Trans` to be able to render when children given is just one

## Tested

add 1 snapshot test, 1 interaction test 

Screenshot_20200424-163709

## Related issues

- part of #3330 

## Commit Message

add Terms screen to user onboarding


packages/mobile/locales/es-419/nuxNamePin1.json
```diff
@@ -4,11 +4,10 @@
   "welcomeCelo": "Te damos la bienvenida a Celo",
   "chooseCountryCode": "Código de país",
   "chooseCountry": "País",
-  "joinText": {
-    "0": "El Monedero Celo te ayuda a enviar, recibir y guardar valor en la red de Celo.",
-    "1":
-      "Al unirte a esta red, nos das permiso para recopilar información anónima sobre cómo utilizas la aplicación. Además, si verificas tu número de teléfono, se almacenará una copia ofuscada en la Red de Celo. Más información en ",
-    "2": "celo.org/terms"
+  "joinText": "El Monedero Celo te ayuda a enviar, recibir y guardar valor en la red de Celo.",
+  "terms": {
```
aaronmgdr, line 8: just moved these

packages/mobile/src/app/saga.ts
```diff
@@ -123,6 +126,8 @@ export function* navigateToProperScreen() {
     navigate(Screens.SetClock)
   } else if (!e164Number) {
     navigate(Screens.JoinCelo)
+  } else if (!acceptedTerms) {
+    navigate(Screens.RegulatoryTerms)
```
aaronmgdr, line 130: dont let them skip it!

packages/mobile/src/invite/JoinCelo.tsx
```diff
@@ -105,15 +103,11 @@ export class JoinCelo extends React.Component<Props, State> {
...
   }
 
-  onPressGoToTerms = () => {
-    navigateToURI(CELO_TERMS_LINK)
-  }
-
   goToNextScreen = () => {
     const nextScreen =
       this.props.pincodeType === PincodeType.Unset
         ? Screens.PincodeEducation
```
aaronmgdr: im not sure what the path is for PinCode.UNset and if in that flow we would also go to the terms screen, but im assuming that it would only happen after they have onboarded

packages/mobile/src/invite/JoinCelo.tsx
```diff
@@ -186,12 +180,6 @@ export class JoinCelo extends React.Component<Props, State> {
               this.props.cachedNumber !== '' ? this.props.cachedNumber : undefined
             }
           />
-          <Text style={styles.disclaimer}>
```
aaronmgdr, line 189: moved to terms screen

packages/mobile/locales/en-US/nuxNamePin1.json
```diff
@@ -4,13 +4,16 @@
...
-    "0": "The Celo Wallet helps you to send, receive, and save value on the Celo Network.",
-    "1":
-      "By joining this network, you give us permission to collect anonymous information about how you use the app. Additionally, if you verify your phone number, an obfuscated copy of it will be stored on the Celo Network. Learn more at ",
-    "2": "celo.org/terms"
-  },
+  "joinText": "The Celo Wallet helps you to send, receive, and save value on the Celo Network.",
   "phoneNumber": "000 000 0000",
+  "terms": {
+    "title": "Terms",
+    "heading1": "Heading",
```
aaronmgdr: TODO need to know what headings should be

packages/mobile/src/invite/RegulatoryTerms.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <SafeAreaView style={styles.container}>
+        <DevSkipButton nextScreen={Screens.PincodeEducation} />
+        <ScrollView contentContainerStyle={styles.scrollContainer}>
+          <View style={styles.terms}>
+            <Text style={fontStyles.h1} testID="Terms">
+              {t('terms.title')}
+            </Text>
+            <Text style={styles.header}>{t('terms.heading1')}</Text>
+            <Text style={styles.disclaimer}>
+              <Trans ns={Namespaces.nuxNamePin1} i18nKey={'terms.privacy'}>
```
aaronmgdr: use `Trans` rather than two separate `i18n.t `calls as this allows for languages with different word order typology

packages/mobile/locales/en-US/nuxNamePin1.json
```diff
@@ -26,6 +29,7 @@
   "fullName": "Full Name",
   "importIt": "Restore Existing Wallet",
   "cancel": "Cancel",
+  "accept": "Accept",
```
i1skn: I think Spanish translation for this is missing

packages/mobile/src/invite/JoinCelo.tsx
i1skn: So, desired behaviour is Terms screen would be after user sets the PIN, right?

packages/mobile/locales/en-US/nuxNamePin1.json
aaronmgdr: how do you typically handle getting the spanish? online robo translator?

packages/mobile/src/invite/JoinCelo.tsx
aaronmgdr: Its before the PIN. they enter it right after the Name Country Phone Number Screen

packages/mobile/locales/en-US/nuxNamePin1.json
i1skn: For simple stuff - just G Translator, for more complicated phrases - channel in Slack #spanish

packages/mobile/src/invite/JoinCelo.tsx
i1skn: When app was launched for first time `pincodeType === PincodeType.Unset // true`, so right after `JoinCelo` screen user would be redirected to `PincodeEducation`, where he/she sets a Pincode and goes straight to `EnterInviteCode` screen. Thus the only way to get to `RegulatoryTerms` screen would be to reload the app, then in `navigateToProperScreen` saga we would catch it earlier than `PincodeType.Unset` condition. 

Probably, I'm missing something here - let me know. But if my understanding is correct I would suggest to add another condition to this "if" to check `acceptedTerms === true` and only then navigate to  `PincodeEducation ` screen. This, though, would require duplicating this "if" in `RegulatoryTerms` as well.

aaronmgdr: oh! this is very helpful. Thanks man. I mistook Unset to mean we were unsetting it.

aaronmgdr: I think i got it now

aaronmgdr: logic like this
```
    if (!this.props.acceptedTerms) {
      navigate(Screens.RegulatoryTerms)
    } else if (this.props.pincodeType === PincodeType.Unset) {
      navigate(Screens.PincodeEducation)
    } else {
      navigate(Screens.EnterInviteCode)
    }

```

aaronmgdr: although im a bit confused why after pincode education the following screen is InviteCode not PineCodeEnter. but that seems to be how it has been

i1skn (approved):
Great work on this one! 🚢 it!

packages/mobile/e2e/Transfer.spec.js
```diff
@@ -53,7 +53,12 @@ describe('Transfer Works', () => {
     await element(by.id('JoinCeloContinueButton')).tap()
   })
 
-  it('NUX->Pin', async () => {
+  it.skip('NUX-Terms', async () => {
```
jmrossy: Did you mean to leave these .skips here?

packages/mobile/locales/en-US/nuxNamePin1.json
jmrossy: consider putting single-word strings like this in global.json instead so we can use it throughout the app

packages/mobile/src/invite/RegulatoryTerms.tsx
```diff
@@ -0,0 +1,125 @@
+import Button, { BtnTypes } from '@celo/react-components/components/Button'
```
jmrossy: I think it was a mistake to put JoinCelo.tsx in the `invite` folder, not sure when that happened. I'd rather not compound that though. Can you put this is a different folder?

jmrossy (commented):
Can you include a screenshot?

packages/mobile/src/invite/RegulatoryTerms.tsx
aaronmgdr: Im not seeing an obvious place. there used to be a nux folder, something like that would make most sense to me

aaronmgdr: Or maybe the account folder would be a good spot to move both this and Join to?

ha noticing there is an Invite component in the account folder.

packages/mobile/e2e/Transfer.spec.js
aaronmgdr: oh right. 

been having a lot of trouble getting e2e tests running to get these fixed

packages/mobile/src/invite/RegulatoryTerms.tsx
jmrossy: Yeah that Invite component should here and these two nux screens should be in a folder like `registration` or something

aaronmgdr: moved those two to regitration

PR #3438, opened 2020-04-16
Update gas fee payment natspecs

asaj:
## Description

To clarify how these functions are used and the rationale behind event emission only in `creditGasFees`.

## Other changes

None
## Tested

No
## Related issues

None
## Backwards compatibility

Yes

### DIFF
--- a/packages/protocol/contracts/stability/StableToken.sol
+++ b/packages/protocol/contracts/stability/StableToken.sol
@@ -480,6 +480,10 @@ contract StableToken is
    * @notice Reserve balance for making payments for gas in this StableToken currency.
    * @param from The account to reserve balance from
    * @param value The amount of balance to reserve
+   * @dev Note that this function is called by the protocol when paying for tx fees in this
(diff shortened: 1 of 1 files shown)


packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -502,6 +506,10 @@ contract StableToken is
    * @param tipTxFee Coinbase fee
    * @param baseTxFee Community fund fee
    * @param gatewayFee Gateway fee
+   * @dev Note that this function is called by the protocol when paying for tx fees in this
+   * currency. Before the tx is executed, gas is debited from the sender via a call to
+   * `debitGasFees`. Note too that the events emitted by `creditGasFees` reflect the *net* gas fee
+   * payments for the transaction.
```
mcortesi, line 512: maybe mention that this must emit events?

I guess there's a difference between specifying the interface every ERC20 with gas fee support should implement, and what stabletoken does.. Of course should be the same, but from documenting perspective it's different (interface specifies what implementers should do)

mcortesi (approved):

PR #3443, opened 2020-04-16
[utils] Add Account generation util functions

### LINKED ISSUE #3435: CLI and mobile are not using derivation paths for creating/showing accounts
### Expected Behavior

CLI and Mobile should use the Celo derivation path, to show/create addresses from a specific mnemonic.
This way, the addresses generated, can be retrieved by any tool (cli, wallet, ledger), and be used as it should.

### Current Behavior

A user creates an account using the CLI, gets a mnemonic, an address, and a pair of public/private keys.
The user lost all its data excepto the mnemonic. If the user tries to generate the same address from the mnemonic using the Celo derivation path, it won't reach the same address

gastonponti:
## Description

Add account functions to the utils, to generate mnemonics and public/private keys using the Celo Derivation Path (m/44'52752'0'/0)

## Other features

Add password and language options for the generation of the mnemonic and the keys in the cli command 

## Tested

Add tests for the library and manually check generation the same account from the mnemonic of a Ledger for testing

## Related issues

- Fixes #3435 (the CLI part)

## Backwards compatibility

We change the way to generate accounts from the CLI, so technically, no


packages/cli/src/commands/account/new.ts
```diff
@@ -1,52 +1,60 @@
...
+import { privateKeyToAddress } from '@celo/utils/lib/address'
+import { flags } from '@oclif/command'
 import { LocalCommand } from '../../base'
 import { printValueMap } from '../../utils/cli'
-import { ReactNativeBip39MnemonicGenerator } from '../../utils/key_generator'
 
 export default class NewAccount extends LocalCommand {
   static description =
-    'Creates a new account locally and print out the key information. Save this information for local transaction signing or import into a Celo node.'
+    "Creates a new account locally using the Celo Derivation Path (m/44'/52752'/0/0) and print out the key information. Save this information for local transaction signing or import into a Celo node."
```
kobigurk: Maybe `m/44'/52752'/0/0/index`?

packages/cli/src/commands/account/new.ts
```diff
@@ -1,52 +1,60 @@
...
   static flags = {
     ...LocalCommand.flags,
+    password: flags.string({
+      description: 'Choose a password to generate the keys',
+    }),
+    indexAddress: flags.integer({
+      default: 0,
+      description: 'Choose the index address of the derivation path',
+    }),
+    language: flags.string({
```
kobigurk, line 20: Does Ledger support mneomnics in other languages?

packages/utils/src/account.ts
```diff
@@ -0,0 +1,63 @@
...
+): string {
+  return bip39.generateMnemonic(strength, undefined, getWordList(language))
+}
+
+export function generateKeys(
+  mnemonic: string,
+  password?: string,
+  addressIndex: number = 0
+): { privateKey: string; publicKey: string } {
+  const seed = bip39.mnemonicToSeedSync(mnemonic, password)
```
kobigurk, line 37: We should test that it works with the Ledger passphrase as expected and add some words about it to the docs I think.

gastonponti: I checked using the same mnemonic of a ledger that I have for testing, and it derived all the index right (at least the first 4, haha)
I would check the other way. Upload a mnemonic to the Ledger

packages/cli/src/commands/account/new.ts
gastonponti: Mmmm, I couldn't find anything. We could add a warning in the help command.  For someone who is not a english native, could be useful, not everyone will use this only for the ledger

packages/utils/src/account.ts
gastonponti: Ok, I verified the other way... (at least 20 minutes adding words, hahaha)
But everything works. Generates the same addresses the cli and the ledger from the mnemonic generated by the cli

kobigurk (approved):

packages/cli/src/commands/account/new.ts
```diff
@@ -1,52 +1,61 @@
...
 
 export default class NewAccount extends LocalCommand {
   static description =
-    'Creates a new account locally and print out the key information. Save this information for local transaction signing or import into a Celo node.'
+    "Creates a new account locally using the Celo Derivation Path (m/44'/52752'/0/0/indexAddress) and print out the key information. Save this information for local transaction signing or import into a Celo node. Ledger: this command has been tested swapping mnemonics with the Ledger successfully (only supports english)"
 
   static flags = {
     ...LocalCommand.flags,
+    password: flags.string({
+      description: 'Choose a password to generate the keys',
```
jmrossy, line 14: Should this be optional?

packages/cli/src/commands/account/new.ts
```diff
@@ -1,52 +1,61 @@
...
-    const ecPublicKey: string = ecKeyPair.getPublic('hex')
-    return ecPublicKey
-  }
-
-  static generateAccountAddressFromPrivateKey(privateKey: string): string {
-    if (!privateKey.startsWith('0x')) {
-      privateKey = '0x' + privateKey
+  static languageOptions(language: string): MnemonicLanguages | undefined {
+    if (language) {
+      return (MnemonicLanguages[language as any] as unknown) as MnemonicLanguages
```
jmrossy: Can this type casting be cleaned up? It seems like a ts-ignore may be better than 3x casts if necessary

packages/metadata-crawler/.gitignore
```diff
@@ -0,0 +1,3 @@
+lib/
```
jmrossy, line 1: Did you mean to include this? Not a problem, just checking

packages/utils/src/account.ts
```diff
@@ -0,0 +1,67 @@
...
+  chinese_traditional,
+  english,
+  french,
+  italian,
+  japanese,
+  korean,
+  spanish,
+}
+
+export function generateMnemonic(
```
jmrossy, line 21: Please package these functions up into a containing object and export it in index.ts (see utils/src/phoneNumbers.ts for an example) so it can be more easily consumed by non-TS consumers as well.

jmrossy (commented):
Mostly looks great! Just some small comments

packages/cli/src/commands/account/new.ts
gastonponti: It's optional. `@oclif`  has a `required` field that is false for default

gastonponti: Aw, I understand. The old version didn't have the option. This is more of how the user want to use it

packages/cli/src/commands/account/new.ts
gastonponti: Haha, you are right

packages/metadata-crawler/.gitignore
gastonponti: Yes, the metadata-crawler/lib was making me crazy. I know that it's not the way to do it, but this will avoid pushing things that shouldn't be in the repo

jmrossy (approved):

PR #3446, opened 2020-04-16
Show different networks on validator explorer

Pedro-vk:
## Description

_Show different networks on validator explorer and be able to switch between them_

## Commit Message
Show different networks on validator explorer


packages/web/env-config.js.template
```diff
@@ -11,6 +11,9 @@ const environments = {
     BLOCKSCOUT: {
       pageLength: 5,
       uri: 'http://localhost:4000/graphiql',
+      mainnet: 'http://localhost:4000/graphiql',
+      baklava: 'http://localhost:4000/graphiql',
```
aaronmgdr, line 15: opinion: I know they were here before, but im not sure these URLs need to be in the ENV config file. only real purpose would be to either obscure them or to have different environments point to different networks but i think that is a recipe for not noticing a network isnt working until production.

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -1,14 +1,24 @@
...
+import Button, { BTN } from 'src/shared/Button.3'
+import menu from 'src/shared/menu-items'
+import OvalCoin from 'src/shared/OvalCoin'
 import { colors, standardStyles, textStyles } from 'src/styles'
 import { weiToDecimal } from 'src/utils/utils'
 
+const networkMenu = [
+  // ['Mainnet', menu.VALIDATORS_LIST.link],
+  ['Baklava', menu.VALIDATORS_LIST__BAKLAVA.link],
+  ['Baklavastaging', menu.VALIDATORS_LIST_BAKLAVASTAGING.link],
```
aaronmgdr: I believe we will want alfajores here too based on what @timmoreton said

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -266,6 +289,26 @@ class ValidatorsList extends React.PureComponent<ValidatorsListProps & I18nProps
...
+            <View style={[styles.links]}>
+              {networkMenu.map(([name, link]) => (
+                <View
+                  key={name}
+                  style={[
+                    styles.linkWrapper,
+                    this.props.router.pathname !== link ? styles.linkWrapperInactive : {},
+                  ]}
+                >
+                  <Button kind={BTN.DARKNAV} href={link} text={name} />
```
aaronmgdr: btw
consider extending https://github.com/celo-org/celo-monorepo/blob/master/packages/web/src/shared/navigation.tsx to take an optional coinColor Prop. and using.

aaronmgdr:
Playing around on the preview site Im wondering if we can have the 
```
"Validator Explorer"
[list of networks]
```

part be rendered outside of the Apollo so when switching back and forth only the data gets refreshed. 

could be a different PR. 

what do you think?

packages/web/env-config.js.template
Pedro-vk: How do you think that we should handle that?

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: Yes, I'm waiting for the rest of the information like mainnet to finish this part. Right now is more for feedback purposes.

packages/web/env-config.js.template
aaronmgdr: the network => url object could be kept either in the ValidatorList component or its own file.

unless you see a benefit of having them inside the ENV file? 

up to you.

Pedro-vk: It's true that probably we won't need different variables by environment. So ok, I'll move to the component

Pedro-vk:
> Playing around on the preview site Im wondering if we can have the
> 
> ```
> "Validator Explorer"
> [list of networks]
> ```
> 
> part be rendered outside of the Apollo so when switching back and forth only the data gets refreshed.
> 
> could be a different PR.
> 
> what do you think?

It's a good idea :muscle:

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: I didn't see this component, so will be good to use it.

I think that I'm going to work on a new PR with this change and move the header outside the GraphQL query.

aaronmgdr: sweet

Pedro-vk:
@aaronmgdr I moved the menu to the parent component to improve the UX and used Navigation component :+1:

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -291,20 +295,14 @@ class ValidatorsList extends React.PureComponent<Props, State> {
...
-                  ]}
-                >
-                  <Button kind={BTN.DARKNAV} href={link} text={name} />
-                  {this.props.router.pathname === link && (
-                    <View style={styles.activeTab}>
-                      <OvalCoin color={colors.gold} size={10} />
-                    </View>
-                  )}
+              {networkMenuList.map(([name, link, navigate]: any) => (
+                <View key={name} style={[styles.linkWrapper]}>
```
aaronmgdr: nit: unneeded array around linkWrapper

packages/web/src/shared/navigation.tsx
```diff
@@ -4,26 +4,35 @@ import Button, { BTN } from 'src/shared/Button.3'
...
+  LIGHT,
+  DARK,
+}
+
 interface Props {
   link?: string
   onPress?: () => void
   text: string
   selected: boolean
+  theme?: NavigationTheme
```
aaronmgdr, line 17: btw I think from a design perspective there is no reason that you couldnt have gold and light or green on dark. 

but we can always tackle that if time comes.

aaronmgdr (approved):

packages/web/src/shared/navigation.tsx
Pedro-vk: I did in this way because is necessary to set the button theme too

PR #3451, opened 2020-04-16
Add initContainer for adding a label to the pods with the name

jcortejoso:
Add initContainer for adding a label to the pods with the name


tkporter (approved):
Thanks!!

PR #3469, opened 2020-04-17
[phone-number-privacy] refactor and handle network/db errors with consistent logging for alerts

aslawson:
## Description

- refactored each endpoint to have its handler in a separate file for readability
- added enum for consistent error messaging to add alerting for network/db errors

## Other changes

- removed serializable setting since it seems to work differently then thought.  it performs all the db request on a snapshot local to the current db connection.  thus, it wouldn't prevent concurrent requests from separate service instances nor write the changes to the actual db.  will need to address the concurrency issue another way

## Tested

updated tests and manually tested

## Related issues

NA

## Backwards compatibility

NA

## Commit Message
[phone-number-privacy] refactor and handle network/db errors with consistent logging for alerts


packages/phone-number-privacy/src/common/utils.ts
```diff
@@ -0,0 +1,15 @@
+import { Response } from 'firebase-functions'
+
+export enum ErrorMessages {
+  DATABASE_UPDATE_FAILURE = 'Failed to update database entry',
```
jmrossy: Nice! Let's also include unique codes in the messages to make them easier to set up alerts for. How about `CELO_PNP_ERR_0X`

packages/phone-number-privacy/src/index.ts
```diff
@@ -1,96 +1,12 @@
...
-      await getNumberPairContacts(request.body.userPhoneNumber, request.body.contactPhoneNumbers)
-    ).map((numberPair) => ({ phoneNumber: numberPair, salt: computeBLSSalt(numberPair) }))
-    await setNumberPairContacts(request.body.userPhoneNumber, request.body.contactPhoneNumbers)
-    // TODO (amyslawson) return salts with contact
-    response.json({ success: true, matchedContacts })
-  } catch (e) {
-    console.log('Failed to getContactMatches', e)
-    response.status(500).send(e)
-  }
+  await handleGetContactMatches(request, response)
```
jmrossy: No need to await if it's the only thing in this method. Or to put it another way, there's no benefit to awaiting if something has no lines after it.

jmrossy (commented):
Mostly LGMT! I'll have another look once the base PR is merged :)

packages/phone-number-privacy/src/database/wrappers/number-pairs.ts
```diff
@@ -8,13 +9,18 @@ export async function getNumberPairContacts(
...
+    const contentPairs = await getDatabase()(NUMBER_PAIRS_TABLE)
+      .select(NUMBER_PAIRS_COLUMN.userPhoneHash)
+      .where(NUMBER_PAIRS_COLUMN.contactPhoneHash, userPhone)
 
-  return contentPairs
-    .map((contractPair) => contractPair[NUMBER_PAIRS_COLUMN.userPhoneHash])
-    .filter((number) => contactPhones.includes(number))
+    return contentPairs
+      .map((contractPair) => contractPair[NUMBER_PAIRS_COLUMN.userPhoneHash])
+      .filter((number) => contactPhones.includes(number))
```
codyborn: `.filter((number) => contactPhones.includes(number))`
This is O(n^2). Each number (N) is compared to contactPhones (N) via a linear search. If we build a map/object and use that instead, the lookup will be constant.

codyborn (approved):

packages/phone-number-privacy/src/database/wrappers/number-pairs.ts
aslawson: oh that's a good point -- i should change it to the js equivalent of a hashset

jmrossy: @aslawson in JS one would typically use either an object `{}` or a map `new Map<type1, type2>()`, whichever you prefer.

packages/phone-number-privacy/src/common/utils.ts
```diff
@@ -0,0 +1,43 @@
+import { Response } from 'firebase-functions'
```
jmrossy: This file feels a bit like a bunch of random stuff.
Consider breaking out ErrorMessages into its own file

packages/phone-number-privacy/src/common/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  EXCEEDED_QUOTA = 'CELO_PNP_ERR_05 Requester exceeded salt service query quota',
+  SALT_COMPUTATION_FAILURE = 'CELO_PNP_ERR_06 Failed to compute BLS salt',
+  UNKNOWN_ERROR = 'CELO_PNP_ERR_07 Something went wrong',
+}
+
+export function respondWithError(res: Response, statusCode: number, error: string) {
+  res.status(statusCode).json({ success: false, error })
+}
+
+export class HashSet<T> {
```
jmrossy: What's this for? Seems like a recreation of `Map` which node already has

jmrossy: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

packages/phone-number-privacy/src/database/wrappers/number-pairs.ts
```diff
@@ -0,0 +1,50 @@
...
+  contactPhones: string[]
+): Promise<string[]> {
+  try {
+    const contentPairs = await getDatabase()(NUMBER_PAIRS_TABLE)
+      .select(NUMBER_PAIRS_COLUMN.userPhoneHash)
+      .where(NUMBER_PAIRS_COLUMN.contactPhoneHash, userPhone)
+
+    const contactPhonesHashSet = hashSetBuilder(contactPhones)
+    return contentPairs
+      .map((contractPair) => contractPair[NUMBER_PAIRS_COLUMN.userPhoneHash])
```
jmrossy: typo contractPair

packages/phone-number-privacy/src/database/wrappers/number-pairs.ts
```diff
@@ -0,0 +1,50 @@
...
+      [NUMBER_PAIRS_COLUMN.contactPhoneHash]: contactPhone,
+    }
+    rows.push(data)
+  }
+  try {
+    await getDatabase().batchInsert(NUMBER_PAIRS_TABLE, rows)
+  } catch (e) {
+    if (e.code !== '23505') {
+      // ignore duplicate insertion error (23505)
+      console.error(ErrorMessages.DATABASE_INSERT_FAILURE, e)
```
jmrossy, line 47: Comment says ignore but we still console.error here? Should this be warn and not use that error code? Keep in mind we'll be alerting based on the error codes

packages/phone-number-privacy/src/salt-generation/get-salt.ts
```diff
@@ -0,0 +1,54 @@
...
+    const remainingQueryCount = await queryQuota.getRemainingQueryCount(
+      request.body.account,
+      request.body.phoneNumber
+    )
+    if (remainingQueryCount <= 0) {
+      respondWithError(response, 400, ErrorMessages.EXCEEDED_QUOTA)
+      return
+    }
+    const salt = computeBLSSalt(request.body.queryPhoneNumber)
+    await incrementQueryCount(request.body.account).catch((error) => {
```
jmrossy: remove .catch here, use async/await patterns

packages/phone-number-privacy/src/database/wrappers/number-pairs.ts
aslawson: i can move the comment above -- but it only logs the error if its not that error code.  there's no real insert-if-not-exist, and I figured we don't care if someone tries to duplicate an entry

packages/phone-number-privacy/src/common/utils.ts
aslawson: i'm trying to get O(N) lookup. that Map.has() looks promising. if its underlying data structure is a javascript map, then we are probably good

aslawson: O(1)*

aaronmgdr: yeah Map.has is O(1)

jmrossy (approved):

PR #3470, opened 2020-04-17
Add Celo genesis message

codyborn:
## Description

Add community elected genesis block message to the genesis block extra-data field.


packages/celotool/src/lib/generate_utils.ts
```diff
@@ -249,12 +250,12 @@ export const generateGenesisFromEnv = (enablePetersburg: boolean = true) => {
 }
 
 export const generateIstanbulExtraData = (validators: Validator[]) => {
-  const istanbulVanity = 32
+  const istanbulVanity = GENESIS_MSG_HASH
```
asaj, line 253: Can we throw an error if istanbulVanity is not exactly equal to 64 characters?

asaj (approved):

packages/celotool/src/lib/generate_utils.ts
codyborn: Good idea to be safe

PR #3472, opened 2020-04-17
Create alfajores-ultralight.md

critesjosh:
## Description

Add a section to the documentation dedicated to setting up an Alfajores ultralight client for the purpose of connecting to the SDK. An improvement over the existing solution in "Hello Contracts" because it has its own dedicated page that is easier to find and has more focused information.


barbaraliau:
@critesjosh is this still relevant?

PR #3473, opened 2020-04-17
add contract deployment info

critesjosh:
## Description

Adds an explanation of how to deploy a contract using ContractKit.

## Other changes

A small addition about how to get accounts from a node and set it to the kit.defaultAccount.

## Tested

The method described is already used in [another section of the docs](https://docs.celo.org/developer-guide/start/hello-contract-remote-node#custom-node-js-deployment). Adding it here makes it easier to find.


packages/docs/developer-resources/contractkit/usage.md
```diff
@@ -1,6 +1,6 @@
 # ContractKit Usage
 
-The following are some examples of the capabilities of the `ContractKit`
+The following are some examples of the capabilities of the `ContractKit`, assuming it is already connected to a node.
```
nityas: perhaps add a link to docs on how to connect to a node if that step is not yet completed?

nityas (approved):
looks good!

packages/docs/developer-resources/contractkit/usage.md
```diff
@@ -20,9 +20,20 @@ await kit.setFeeCurrency(CeloContract.StableToken)
...
 ```
 
+## Deploy a contract
+
+Deploying a contract with the default account already set. Simply send a transaction with no `to:` field. See more about [sending custom transactions](https://docs.celo.org/developer-guide/overview/introduction/contractkit/contracts-wrappers-registry#sending-custom-transactions)
+
+```ts
+let bytecode = '0x608060405234...' // compiled Solidity deployment bytecode
+
+let tx = await kit.sendTransaction({
```
nityas, line 35: maybe add a link to blockscout and/or how to check that the deploy was successful?

PR #3474, opened 2020-04-17
Update key-rotation.md

mattharrop:
## Description

Fixed error which confused which private key is required to execute the proof of possession commands.


asaj (approved):

asaj:
Thanks @mattharrop!

PR #3476, opened 2020-04-17
[Wallet] Account creation using the Celo derivation path

### LINKED ISSUE #3435: CLI and mobile are not using derivation paths for creating/showing accounts

gastonponti:
## Description

The account creation was generating addresses that where not following the Celo's derivation path. This fixes that

## Related issues

- Fixes #3435 

## Commit Message
[Wallet] Account creation using the Celo derivation path

## Description

The account creation was generating addresses that where not following the Celo's derivation path.


gastonponti:
@jmrossy @cmcewen I couldn't test it in the app.

One thing that is weird is that I didn't found anything using `createQuizWordList`, don't know if you prefer removing that code to make it cleaner or leave it like that

packages/mobile/src/web3/saga.ts
```diff
@@ -148,7 +149,7 @@ export function* getOrCreateAccount() {
     Logger.debug(TAG + '@getOrCreateAccount', 'Creating a new account')
 
     const wordlist = getWordlist(yield select(currentLanguageSelector))
-    let mnemonic: string = yield call(generateMnemonic, MNEMONIC_BIT_LENGTH, null, wordlist)
+    let mnemonic: string = yield call(generateMnemonic, MNEMONIC_BIT_LENGTH, wordlist)
```
jmrossy: @gastonponti I think there's been a disconnect on our plan for this part. I thought we wanted to continue to use `react-native-bip39` for *creating* the mnemonics but that we would use code from utils to *derive* the keys from those mnemonics. The benefit to using RN-Bip39 is that it runs in native code so it's more performant on low-end phones.

@cmcewen I believe you have a bit more context since IIRC you were the one to integrate RN-Bip39 into the app. Were we seeing long waits on low-end phones when generating mnemonics in JS-land?

cmcewen: Yup so the issue is pbkdf2 is real slow on android in hermes. We can fix this by switching the bip39 version (https://github.com/bitcoinjs/bip39/blob/master/src/index.js#L4) similar to what we did with the rn version: https://github.com/celo-org/react-native-bip39/commit/8d007387fc8bcc18a2d0e2344ec986a0c6f15834

gastonponti: Ok, I got dizzy by the bip32 talk (the one deriving), sorry.
I'm changing it right now

gastonponti: The `generateMnemonic` it's fine, it's not using `pbkdf2`, but it's used to generate the seed that is input for the `bip32`.
Let me check if I could extract that, otherwise I could switch between libs (but I'm idea was to unify and reduce libs, haha)

jmrossy: @gastonponti Yes, I've considered replacing our web of crypto dependencies with a single Celo one to make things more secure and auditable. Probably won't happen before mainnet though!
We can chat and figure out the best solution for now :)

packages/mobile/src/import/saga.ts
```diff
@@ -31,14 +31,15 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
   Logger.debug(TAG + '@importBackupPhraseSaga', 'Importing backup phrase')
   yield call(waitWeb3LastBlock)
   try {
-    if (!validateMnemonic(phrase)) {
+    if (!validateMnemonic(phrase, undefined, bip39)) {
```
gastonponti, line 34: @jmrossy this is the place where a language is required (I'll add an issue)

packages/cli/src/commands/account/new.ts
```diff
@@ -53,8 +53,11 @@ export default class NewAccount extends LocalCommand {
 
   async run() {
     const res = this.parse(NewAccount)
-    const mnemonic = generateMnemonic(undefined, NewAccount.languageOptions(res.flags.language!))
-    const keys = generateKeys(mnemonic, res.flags.password, res.flags.indexAddress)
+    const mnemonic = await generateMnemonic(
+      undefined,
```
jmrossy: Should the param order be swapped to avoid this undefined in the middle?

packages/mobile/src/backup/utils.ts
```diff
@@ -12,7 +12,10 @@ export const MNEMONIC_SPLITTER = 'celo'
 export async function createQuizWordList(mnemonic: string, language: string | null) {
   const disallowedWordSet = new Set(mnemonic.split(' '))
   const languageWordList = getWordlist(language)
-  const wordOptions: string = await generateMnemonic(1000, null, languageWordList)
+  const wordOptions: string = await generateMnemonic(
+    MnemonicStrength.s256_24words,
+    languageWordList
+  )
```
jmrossy, line 20: We should be passing in RN-bip39 here right?

packages/mobile/src/web3/saga.ts
```diff
@@ -48,8 +49,9 @@ import { Block } from 'web3-eth'
 
 const TAG = 'web3/saga'
 
-const ETH_PRIVATE_KEY_LENGTH = 64
-const MNEMONIC_BIT_LENGTH = (ETH_PRIVATE_KEY_LENGTH * 8) / 2
+// const ETH_PRIVATE_KEY_LENGTH = 64
```
jmrossy: You can remove these comments, pretty sure we won't need them

packages/utils/src/account.ts
```diff
@@ -1,5 +1,9 @@
 import * as bip32 from 'bip32'
 import * as bip39 from 'bip39'
+// tslint:disable-next-line: no-duplicate-imports
+import { wordlists } from 'bip39' // Unify the wordlists (otherwise depends on the instance of the bip39)
```
jmrossy: Why not access this through `bip39.wordlists` instead of creating a dupe import?

packages/utils/src/account.ts
```diff
@@ -18,23 +22,79 @@ export enum MnemonicLanguages {
...
+        } else {
+          resolve(bip39.entropyToMnemonic(randomBytesBuffer.toString('hex'), wordlist))
+        }
+      })
+    })
+  },
+  validateMnemonic: bip39.validateMnemonic,
+}
+
+interface Bip39 {
```
jmrossy, line 28: Nit move this up before bip39Wrapper

packages/utils/src/account.ts
```diff
@@ -18,23 +22,79 @@ export enum MnemonicLanguages {
   spanish,
 }
 
-export function generateMnemonic(
+const bip39Wrapper: Bip39 = {
+  mnemonicToSeedSync: bip39.mnemonicToSeedSync,
+  mnemonicToSeed: bip39.mnemonicToSeed,
+  generateMnemonic: (
```
jmrossy: Nit: having this method defined inline like this is a bit tough to read

packages/utils/src/account.ts
```diff
@@ -18,23 +22,79 @@ export enum MnemonicLanguages {
...
+  mnemonic: string,
+  password?: string,
+  addressIndex: number = 0,
+  bip39ToUse: Bip39 = bip39Wrapper
+): Promise<{ privateKey: string; publicKey: string }> {
+  const seed = await bip39ToUse.mnemonicToSeed(mnemonic, password)
+  const node = bip32.fromSeed(seed)
+  const newNode = node.derivePath(`${CELO_DERIVATION_PATH_BASE}/${addressIndex}`)
+  return {
+    privateKey: newNode.privateKey!.toString('hex'),
```
jmrossy: Is there a type problem or can the PK possibly be null? Maybe we should check it and throw if not?

jmrossy (commented):
Mostly looks good, just some small comments!
Lmk if you need help testing this in the app

packages/mobile/src/backup/utils.ts
gastonponti: Yes. Nice catch 👍

packages/utils/src/account.ts
gastonponti: Because I didn't had the wrapper before, hahaha. But you are totally right!

packages/cli/src/commands/account/new.ts
gastonponti: Well, It depends of which is the parameter that will be more often the default.
The `bip39` and `react-native-bip39`, both have first the strength, that's why I've choose this way

gastonponti: To make it prettier I could add the `MnemonicStrength.s256_24words` that is the default value, to make everything clearer

jmrossy: Up to you, I'm fine either way :)

packages/utils/src/account.ts
gastonponti: I checked the code. The only possible way that the privateKey is throwing null/undefined, it's if you are trying to generate the node from a publicKey.
And only two functions lead to that `bip32.fromBase58` and `bip32.fromPublicKeyLocal`
The, `bip.fromSeed` (the only one we are using) will return a valid node with a privateKey, and will avoid that error

gastonponti: I could add a throw and a comment

jmrossy (approved):
Ship it!

PR #3478, opened 2020-04-18
Lower the block gas limit after deploying contracts

asaj:
## Description

This PR allows for the block gas limit to be reduced after the contracts are deployed.

Some contracts (e.g. `Validators` and `Governance`) require more than the 10M block gas limit planned for RC1 at deployment time, so to work around this the block gas limit is lowered in a folllow-up script.

## Other changes

- The truffle config now sets the transaction gas amount based on whether or not it appears to be an initial contract deployment.

## Tested

- End to end tests

## Related issues

None

## Backwards compatibility

No


packages/protocol/scripts/bash/init_network.sh
```diff
@@ -6,4 +6,4 @@ set -euo pipefail
 # Flags:
 # -n: Name of the network to migrate to
 
-yarn run migrate -r "$@"
+yarn run migrate -r "$@" && yarn run set_block_gas_limit "$@"
```
m-chrzan, line 9: Would it make sense to have this be a last migration script instead of a new independent script?

m-chrzan (commented):
One thing I worry about: should any of these large contracts need an update (in the worst case scenario - due to a bug) the upgrade process would require a vote to temporarily increase the gas limit (leaving a window when anyone can submit costlier transactions), upload the contract, a vote to actually upgrade contract logic, and another vote to decrease gas limit. This is especially cumbersome in the case of a hotfix that needs to pass quickly.

Don't see any quick fixes, but wanted to point this risk out.

asaj:
> Don't see any quick fixes, but wanted to point this risk out.

Yeah thanks for flagging. `Governance.sol` and `Validators.sol` are the two "offenders" here. I suspect the block gas limit will be raised in the short term above 10M as the network matures, bounding the time in which upgrading these contracts would be painful.

packages/protocol/scripts/bash/init_network.sh
asaj: Rationale behind a separate script was that truffle will send final transactions to the Migrations.sol contract after every migration, and the gas amount truffle uses for all contracts is not easily configured. So for migrations it's configured to allow for 20M gas per tx, and after the gas limit is lowered, any further truffle txs will fail.

packages/protocol/scripts/bash/set_block_gas_limit.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Sets block gas limit and turns ownership of contract over to governance
+#
+# Flags:
+# -n: Name of the network to migrate to
+
+TRUFFLE_OVERRIDE=""
+MIGRATION_OVERRIDE=""
+NETWORK=""
+# https://github.com/trufflesuite/truffle-migrate/blob/develop/index.js#L161
```
m-chrzan, line 12: This looks like remnants from a copy and paste?

packages/protocol/scripts/bash/set_block_gas_limit.sh
```diff
@@ -0,0 +1,34 @@
...
+# Default to larger than the number of contracts we will ever have
+
+while getopts 'n:rt:f:c:m:' flag; do
+  case "${flag}" in
+    n) NETWORK="$OPTARG" ;;
+    c) TRUFFLE_OVERRIDE="$OPTARG" ;;
+    m) MIGRATION_OVERRIDE="$OPTARG" ;;
+    t) ;;
```
m-chrzan, line 20: Ditto

packages/protocol/scripts/bash/set_block_gas_limit.sh
```diff
@@ -0,0 +1,34 @@
...
+# Flags:
+# -n: Name of the network to migrate to
+
+TRUFFLE_OVERRIDE=""
+MIGRATION_OVERRIDE=""
+NETWORK=""
+# https://github.com/trufflesuite/truffle-migrate/blob/develop/index.js#L161
+# Default to larger than the number of contracts we will ever have
+
+while getopts 'n:rt:f:c:m:' flag; do
```
m-chrzan, line 15: Ditto - superfluous flags listed here.

m-chrzan (approved):
Looks good pending green tests.

PR #3480, opened 2020-04-18
Update Accounts.sol

mattharrop:
Clarify error when attempting to re-authorize an account that has been authorized in the past.

## Description

When rotating VALIDATOR_SIGNER key, it is prohibited to re-use a kay.  If re-use is attempted, this error message is displayed.  The old error message was not clear, this proposed error message is more clear.


asaj (approved):

asaj:
Thanks @mattharrop!

PR #3494, opened 2020-04-20
[Wallet] Replace QR scanner screen instead of keeping it in stack

### LINKED ISSUE #3425: [Wallet] iOS redirect on transaction is buggy
### Expected Behavior

User stays on `Review payment` after they entered amount and pressed "Send"

### Current Behavior

User sometimes navigated back to `Send or Request` screen. 
Video: https://celo-org.slack.com/archives/CE48N92M7/p1583445768004700

--

As a part of #3068

i1skn:
## Description

This is needed to unmount the QR code scanner screen, so it does not continue to scan even though a user has moved to other screen.

## Other changes

* Remove `.flowconfig` cause we are using `TypeScript`
* Ensure that `navigator` exists in `navigateBack()`

## Tested

Android device & iPhone

## Related issues

- Fixes #3425

## Backwards compatibility

Yes

## Commit Message
[Wallet] Replace QR scanner screen instead of keeping it in stack

This is needed to unmount the QR code scanner screen, so it does not continue to scan even though a user has moved to other screen.


annakaz (approved):
👍 LGTM

packages/mobile/src/qrcode/QRScanner.tsx
```diff
@@ -23,50 +24,31 @@ interface DispatchProps {
...
-  state = {
-    isScanningEnabled: true,
-  }
+  state = {}
 
   camera: RNCamera | null = null
 
   // This method would be called multiple times with the same
   // QR code, so we need to catch only the first one
   onBardCodeDetected = (rawData: any) => {
```
annakaz: Nit: not a new typo but "onBarcodeDetected" (removing d)

jmrossy (approved):
lgtm 👍

packages/mobile/src/qrcode/QRScanner.tsx
i1skn: Fixed!

PR #3495, opened 2020-04-20
Added a command that's helpful for checking the status of key rotation

mrsmkl:
## Description

`election:current` might be confusing because it doesn't necessarily show the current signers if they are being changed.

## Tested

Only manual testing.


asaj:
Could this be folded into election:current somehow?

asaj (approved):

PR #3507, opened 2020-04-20
HSM support + aad-pod-identity for oracles

tkporter:
## Description

This PR moves the oracle helm chart to use HSMs in separate key vaults via https://github.com/Azure/aad-pod-identity. Celotool will create an azure identity for each oracle in a deployment that only has access to the key vault it requires. Oracle addresses are given using ORACLE_AZURE_KEY_VAULT_ADDRESSES, which is of the form:

```
<address>:<keyVaultName>,<address>:<keyVaultName>
eg: 0x0000000000000000000000000000000000000000:keyVaultName0,0x0000000000000000000000000000000000000001:keyVaultName1 
```

## Other changes

n/a

## Tested

Ran oracles on a private network `trevor4` that had different key vaults.

## Related issues

n/a

## Backwards compatibility

No-- the non HSM oracle deployment doesn't work anymore

## Commit Message
HSM and aad-pod-identity support for oracles


packages/celotool/src/lib/oracle.ts
```diff
@@ -17,25 +23,124 @@ export async function installHelmChart(celoEnv: string) {
...
+//   eg: 0x0000000000000000000000000000000000000000:keyVault0,0x0000000000000000000000000000000000000001:keyVault1
+// into an array in the same order
+function oracleAddressesAndVaults(): {
+  address: string
+  keyVaultName: string
+}[] {
+  const vaultNamesAndAddresses = fetchEnv(envVar.ORACLE_ADDRESS_KEY_VAULTS).split(',')
+  const addressesAndVaults = []
+  for (const nameAndAddress of vaultNamesAndAddresses) {
+    const [address, keyVaultName] = nameAndAddress.split(':')
```
codyborn: We should add a check here and fail if input isn't as expected (ex. address missing)

codyborn (approved):
Great work!

packages/helm-charts/oracle/templates/statefulset.yaml
```diff
@@ -43,7 +73,16 @@ spec:
...
+
+            # Get the correct key vault name
+            AZURE_KEY_VAULTS={{- range $index, $identity := .Values.oracle.identities -}}{{ $identity.azure.keyVaultName }},{{- end }}
+            export AZURE_KEY_VAULT=`echo -n $AZURE_KEY_VAULTS | cut -d ',' -f $((RID + 1))`
+
+            # Get the correct oracle account address
+            ORACLE_ACCOUNT_ADDRESSES={{- range $index, $identity := .Values.oracle.identities -}}{{ $identity.address }},{{- end }}
+            export ORACLE_ACCOUNT_ADDRESS=`echo -n $ORACLE_ACCOUNT_ADDRESSES | cut -d ',' -f $((RID + 1))`
+
+            sleep 120
```
jcortejoso: I guess the `sleep` is for waiting to `aad-pod-identity` to assign the label identity, am I right?
Not sure if it can be refactored to use a reactive way using the metadata ip, but maybe I can check that.

jcortejoso (approved):
Great work!

packages/helm-charts/oracle/templates/statefulset.yaml
tkporter: Oops, that was a hack so I didn't have to rebuild an image that would have a longer amount of time between reports. I'll remove- thanks!

PR #3508, opened 2020-04-20
Use KeyVault client for retrieving BLS key

### LINKED ISSUE #429: User SBAT to faucet themselves with a proper number
### Expected Behavior

User SBAT to faucet themselves with a proper number. And in case of a failure, the error should be less robotic. Most people do not know what E.164 number is.

### Current Behavior

The checker is too strict and even `(` or `)` with a robotic message demanding that `E.164 number should be supplied`.

codyborn:
## Description
The BLS key will be used to generate the salt for a given phone number mapping. As such, leaking this key can potentially compromise the privacy of users on the network. This change allows us to store the BLS key within a secure KeyVault instead of as an environment variable. 

## Other changes

Migrate the AzureKeyVaultClient to the contractkit utils directory since it's used by more than just account signing.

## Tested

Tested via mock object as well as using an actual KeyVault.

## Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/429

## Backwards compatibility

These changes are not backwards compatible. We will need to deploy a KeyVault, create the corresponding key, and update the CloudFunction config before we can deploy the update.


packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
```diff
@@ -0,0 +1,35 @@
...
+      const privateKey = await BLSCryptographyClient.getPrivateKey()
+      const pkBuff = new Buffer(privateKey)
+      return BLINDBLS.computePRF(pkBuff, new Buffer(queryPhoneNumber))
+    } catch (e) {
+      console.error('Failed to compute salt', e)
+      throw e
+    }
+  }
+
+  private static privateKey: string
```
aslawson: does this work to have this static variable named the same as the const variable in the above method?

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  /**
+   * Get singleton privateKey
+   */
+  private static async getPrivateKey(): Promise<string> {
+    if (BLSCryptographyClient.privateKey) {
+      return BLSCryptographyClient.privateKey
```
aslawson: when would `BLSCryptographyClient.privateKey` already be set?  i don't see it being set anywhere in this diff

packages/phone-number-privacy/src/config.ts
```diff
@@ -20,6 +19,13 @@ interface Config {
     database: string
     host: string
   }
+  keyVault: {
+    azureClientID: string
+    azureClientSecret: string
+    azureTenant: string
```
aslawson, line 25: i only see `azureVaultName` and `azureSecretName` used, where do these fields come into play?

packages/phone-number-privacy/package.json
```diff
@@ -18,7 +18,7 @@
     "db:migrate": "NODE_ENV=development ts-node scripts/runMigrations.ts"
   },
   "dependencies": {
-    "@celo/contractkit": "0.3.2",
+    "@celo/contractkit": "0.3.9-dev",
```
aslawson: we can only used published versions in firebase, so we may have to wait if your contractkit changes aren't published yet

packages/phone-number-privacy/package.json
```diff
@@ -27,7 +27,8 @@
     "pg": "^7.18.2"
   },
   "devDependencies": {
-    "firebase-functions-test": "^0.1.6"
+    "firebase-functions-test": "^0.1.6",
+    "jest": "^25.4.0"
```
aslawson: we already pull in jest in the collective package.json -- are you just including it here for the version bump?

packages/phone-number-privacy/.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
```
aslawson: what is this for?  should the .vscode path be added to .gitignore?

aslawson (commented):
one thing i'm unclear on -- if these keys can only be generated in the keyvault, how do we generate the threshold keys?  

https://github.com/celo-org/blind-threshold-bls-wasm
```
// Generate the secret shares for a 3-of-4 threshold scheme
const t = 3;
const n = 4;
const keys = threshold.thresholdKeygen(n, t, crypto.randomBytes(32))
```

^ it looks like all keys get generated at once

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
codyborn: Yeah, the method above can distinguish between them `privateKey` is the local var and `BLSCryptographyClient.privateKey` is the static one.

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
codyborn: `getPrivateKey()` can be called multiple times and it doesn't make sense to make a round-trip network request just to get the same key. This just checks if it's already populated the cached value and returns it if it's present.

packages/phone-number-privacy/src/config.ts
codyborn: Good catch! I forgot to set these as env variables for the `AzureKeyVaultClient` to use them.

packages/phone-number-privacy/.vscode/launch.json
codyborn: This file allows us to debug tests locally in VSCode. I figured others would find it useful but we can always have folks generate it on-demand. Do you have any preference?

packages/phone-number-privacy/package.json
codyborn: Having a project level jest was necessary for the local debugging to work (see launch.json comment).

packages/phone-number-privacy/package.json
codyborn: Thanks! I'll split this into two separate PRs then. Once the contractkit can be published then I'll update this version number.

codyborn:
> one thing i'm unclear on -- if these keys can only be generated in the keyvault, how do we generate the threshold keys?
> 
> https://github.com/celo-org/blind-threshold-bls-wasm
> 
> ```
> // Generate the secret shares for a 3-of-4 threshold scheme
> const t = 3;
> const n = 4;
> const keys = threshold.thresholdKeygen(n, t, crypto.randomBytes(32))
> ```
> 
> ^ it looks like all keys get generated at once

@aslawson since cloud KeyVaults don't support BLS keys, I'm using the generic `secret` functionality of KeyVault. It acts like a secure Key/Value dictionary and allows us to provide any format for the secret object. 
Example: https://docs.microsoft.com/en-us/azure/key-vault/secrets/quick-create-portal

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
aslawson: ya but I don't see you ever populating the cache value, so this would always be null

packages/phone-number-privacy/.vscode/launch.json
aslawson: i don't have a preference -- i typically see the .vscode stuff added to .gitignore but am not an expert on the best practices here

more from a curiosity, how is it used?

aslawson:
> @aslawson since cloud KeyVaults don't support BLS keys, I'm using the generic secret functionality of KeyVault. It acts like a secure Key/Value dictionary and allows us to provide any format for the secret object.
> Example: https://docs.microsoft.com/en-us/azure/key-vault/secrets/quick-create-portal

I see -- so we can generate the keys externally and upload them to the vaults.  I had thought you mentioned a limitation that keys had to be generated in the vault, but perhaps that was for other key types.

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
aslawson: 👍

packages/phone-number-privacy/.vscode/launch.json
jmrossy: Cool addition! Don't think we have many of these included yet. For this to work, does someone need to open VSCode on this package specifically or does it work when opened from the monorepo as most people do?

packages/phone-number-privacy/package.json
jmrossy: We avoid introducing differing versions of packages in the monorepo, it leads to unexpected behavior in unexpected places. Why was it necessary for local debugging? The jest package would get hoisted to the monorepo root's node_module's regardless

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
```diff
@@ -0,0 +1,40 @@
+import { AzureKeyVaultClient } from '@celo/contractkit/src/utils/azure-key-vault-client'
+import { BLINDBLS } from 'bls12377js-blind'
+import config from '../config'
+
+export class BLSCryptographyClient {
```
jmrossy: Is there a particular reason for making this a class? Like does it have some state it should encapsulate.
This is more of a style point but it's not very Javascripty to create a class and then fill it with static functions. There's no real downside to just putting those functions at the root scope and then exporting them. Node's module system (a file is a module) will already encapsulate any variables you declare in this scope

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
```diff
@@ -0,0 +1,40 @@
...
+  /*
+   * Computes the BLS Salt for the blinded phone number.
+   */
+  public static async computeBLSSalt(queryPhoneNumber: string) {
+    try {
+      const privateKey = await BLSCryptographyClient.getPrivateKey()
+      const pkBuff = new Buffer(privateKey)
+      return BLINDBLS.computePRF(pkBuff, new Buffer(queryPhoneNumber))
+    } catch (e) {
+      console.error('Failed to compute salt', e)
```
jmrossy: Can we use one of Amy's new error codes here?

jmrossy (commented):
Mostly looks good, just some small comments :)

packages/phone-number-privacy/.vscode/launch.json
codyborn: VSCode expects the .vscode to be at the root of the directory that is currently open in VSCode. However, we can add multiple Launch.json files at the root of monorepo and the dev can select whichever one she wants to run: https://code.visualstudio.com/Docs/editor/debugging

codyborn: I'll add this as a separate PR and get broader eyes on it since it'll be setting a precedence across the repo.

packages/phone-number-privacy/package.json
codyborn: Will move to the root of monorepo as part of the launch.json migration in a future PR (removing here for now).

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
codyborn: you're totally right, thanks!

codyborn:
> > @aslawson since cloud KeyVaults don't support BLS keys, I'm using the generic secret functionality of KeyVault. It acts like a secure Key/Value dictionary and allows us to provide any format for the secret object.
> > Example: https://docs.microsoft.com/en-us/azure/key-vault/secrets/quick-create-portal
> 
> I see -- so we can generate the keys externally and upload them to the vaults. I had thought you mentioned a limitation that keys had to be generated in the vault, but perhaps that was for other key types.

Yeah, that was a limitation for HSM keys. Unfortunately, HSMs don't support BLS keys so we have to use the less secure approach of storing/retrieving the raw key value.

packages/phone-number-privacy/src/salt-generation/bls-cryptography-client.ts
codyborn: I'm new to nodeJS so I'm probably just falling back to my C# habits 😄 I made it a class since it has the `privateKey` state which should be a private static property of the class. Is it recommended to just have this in a module?

codyborn: Spoke offline; due to a limitation with Jest, we'll lose the ability to dynamically mock (test against mock and actual KV) if we remove the class and make it a module. Will keep as is.

packages/phone-number-privacy/test/salt-generation/bls-signature.test.ts
```diff
@@ -1,11 +1,30 @@
...
-import { computeBlindedSignature } from '../../src/bls/bls-signature'
+import { BLSCryptographyClient } from '../../src/bls/bls-cryptography-client'
+import config from '../../src/config'
+
+const USING_MOCK = config.keyVault.azureClientSecret === 'useMock'
 
 describe(`BLS service computes signature`, () => {
+  beforeEach(() => {
+    // Use mock client if env vars not specified
+    if (!USING_MOCK) {
```
aslawson, line 10: is it not expected that the test will always run with the DEV environment configs?  seems we could just expect that we are using the mock, no?

aslawson (approved):
looks good!

packages/phone-number-privacy/test/salt-generation/bls-signature.test.ts
codyborn: Correct, it will always use mocks as-is. The idea is that when a dev wants to locally test the an actual KV instance, they would add the keyvault variables to the config file. Devs making changes that affect this logic would be expected to test out the non-mocked scenario to ensure it all works. I'll add this to our dev readme to make it more clear.

PR #3509, opened 2020-04-21
Updates G2 point compression

kobigurk:
## Description

In combination with https://github.com/celo-org/celo-blockchain/pull/1000.


jmrossy (commented):
Given that celo-blockchain changed, do we need to update the @celo/client lib in mobile or does this issue not effect geth light clients?

kobigurk:
> Given that celo-blockchain changed, do we need to update the @celo/client lib in mobile or does this issue not effect geth light clients?

Very good point. It does affect them, and although unlikely, should be updated.

jmrossy:
@kobigurk Okay, we can merge your two PRs then and open a new one for mobile once celo-blockchain has autodeployed the new build to npm (as `@celo/client`)

jmrossy (approved):

PR #3510, opened 2020-04-21
Add cli identity:current-attestation-services

nambrot:
## Description

This PR adds a new cli command `identity:current-attestation-services` which will take the current validator set and print out which validator has 
1) opted into attestations
2) has an attestation service URL claimed
3) lists the status information returned by it

## Other changes

- Adds a proxyCall to Accounts from the smart contract for `hasAuthorizedAttestationSigner`

## Tested

- Ran against baklava


packages/cli/src/commands/identity/current-attestation-services.ts
```diff
@@ -0,0 +1,91 @@
...
+      smsProviders: [],
+      blacklistedRegionCodes: [],
+      rightAccount: false,
+    }
+
+    try {
+      const metadata = await IdentityMetadataWrapper.fetchFromURL(this.kit, metadataURL)
+      const attestationServiceURLClaim = metadata.findClaim(ClaimTypes.ATTESTATION_SERVICE_URL)
+
+      if (attestationServiceURLClaim === undefined) {
```
jmrossy: nit, `if (! attestationServiceURLClaim)`

jmrossy (approved):

PR #3511, opened 2020-04-21
Fix verification issue in mobile app

cmcewen:
## Description

Changes to contractkit caused an error in the app - `this.kit` was null since the function wasn't bound

## Tested

Verification failed before this change but works now


i1skn (changes requested):
Great work on catching this!!! JS is so JS 🤦 

Regarding **Podfile** changes - please check that `bundle exec pod --version` returns `1.9.1` and try to remove `~/.cocoapods`, then try `pod install` again. After this `Podfile.lock` should not have any changes.

jeanregisser (approved):
Awesome! 👍

i1skn (approved):

PR #3512, opened 2020-04-21
TME update for RC1

### LINKED ISSUE #3070: Protocol Economics Monitor Dashboard with Celo Stack
### Commit
- [x] access public views with exporter
- [x] make sure all events are being reported
- [x] export relevant data to bigquery
- [ ] create Reserve dashboard in datastudio
- [ ] create PoS dashboard in datastudio

### Stretch
- [x] failover strategy with replay and catchup

diminator:
## Description

Improvements to TME:
- add public views
- catch-up mode: `FROM_BLOCK`, `TO_BLOCK`, `BLOCK_INTERVAL`
- internal transactions
- make sure block parsing doesnt crash

## Tested

deployed on:
- baklava stake-off (bigquery + [notebook](https://colab.research.google.com/drive/1jEsnLerzf6ndIRgFcJ3yh8tq3ghr4ZWO))
- baklava testnet (bigquery + [notebook](https://colab.research.google.com/drive/1osi7d4vbFItKiozUVB7ASidmtvR0eMG9))

## Related issues

- Fixes #3070


packages/contractkit/src/address-registry.ts
```diff
@@ -33,7 +33,8 @@ export class AddressRegistry {
 
       debug('Fetched address:  %s = %s', address)
       if (!address || address === NULL_ADDRESS) {
-        throw new Error(`Failed to get address for ${contract} from the Registry`)
+        // TODO: Remove this!
+        console.error(`Failed to get address for ${contract} from the Registry`)
```
diminator: @Pedro-vk can we remove this?

dockerfiles/transaction-metrics-exporter/Dockerfile
```diff
@@ -18,13 +18,19 @@ COPY scripts/ scripts/
 # Copy only pkg.json
 COPY packages/typescript/package.json packages/typescript/
 COPY packages/utils/package.json packages/utils/
+COPY packages/protocol/package.json packages/protocol/
```
diminator, line 14: @Pedro-vk we probably want to point here to latest contractkit?

diminator: or maybe after merge changes we made here to contractKit and released a new version?

packages/contractkit/src/address-registry.ts
Pedro-vk: I think so

dockerfiles/transaction-metrics-exporter/Dockerfile
Pedro-vk: I think that point to a specific version to avoid retrocompatibility issues

packages/contractkit/src/wallets/ledger-wallet.ts
```diff
@@ -9,7 +9,7 @@ import { LedgerSigner } from './signers/ledger-signer'
 import { Signer } from './signers/signer'
 import { Wallet } from './wallet'
 
-export const CELO_BASE_DERIVATION_PATH = CELO_DERIVATION_PATH_BASE.slice(2)
+export const CELO_BASE_DERIVATION_PATH: string = CELO_DERIVATION_PATH_BASE.slice(2)
```
nambrot, line 12: Is this necessary?

packages/transaction-metrics-exporter/src/blockchain.ts
```diff
@@ -46,16 +40,23 @@ export async function metricExporterWithRestart(providerUrl: string) {
...
+  reason: 'connection-error' | 'subscription-error' | 'not-listening'
+  error?: any
+}
 
 export async function runMetricExporter(kit: ContractKit): Promise<EndReason> {
-  const blockProcessor = await newBlockHeaderProcessor(kit)
+  ;(kit.web3.currentProvider as any).enableSubscriptions()
+
+  // Start exporting metrics
+  const fromBLock = Number.parseInt(process.env.FROM_BLOCK || '', 10)
```
nambrot: Typo

packages/transaction-metrics-exporter/src/block-processor.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  async init() {
+    if (this.initialized) {
+      throw new Error('BlockProcessor is running')
+    }
+    this.initialized = true
+
+    await this.loadContracts()
+
+    if (this.fromBlock?.toFixed?.()) {
```
nambrot: Is this to check if `fromBlock` is Nan?

packages/transaction-metrics-exporter/src/blockchain.ts
```diff
@@ -46,16 +40,23 @@ export async function metricExporterWithRestart(providerUrl: string) {
...
+  const toBlock = Number.parseInt(process.env.TO_BLOCK || '', 10)
```
nambrot: How were you planning on deploying this?

nambrot: I.e. kuberentes/helm does not know?

packages/transaction-metrics-exporter/src/block-processor.ts
```diff
@@ -0,0 +1,266 @@
...
+              contract,
+              function: method,
+              args: JSON.stringify(args),
+              blockNumber,
+              values: transformValues(returnData),
+            }
+            this.logEvent(LoggingCategory.State, data)
+            this.logHistogram({ ...data, args }, maxBucketSize)
+          })
+          .catch(() => '') as Promise<any>
```
nambrot: I don't think we should be just failing silently?

dockerfiles/transaction-metrics-exporter/Dockerfile
nambrot: Yeah we'll want to use the published version as much as possible to avoid bloating the docker image here

packages/transaction-metrics-exporter/src/block-processor.ts
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  async getToBlock() {
+    if (this.toBlock) {
+      return this.toBlock
+    }
+    return (await this.kit.web3.eth.getBlock('latest')).number
+  }
+
+  async onNewBlock(blockNumber: number, blockIntervalNumber?: number) {
```
nambrot, line 122: What's the usecase for not just processing every single block?

packages/transaction-metrics-exporter/src/block-processor.ts
```diff
@@ -0,0 +1,266 @@
...
+          id: '2',
+        },
+        (err: any, result: any) => {
+          if (!err) {
+            result?.result
+              ?.filter?.(
+                ({ type, callType }: any) => type === 'create' || callType === 'delegatecall'
+              )
+              ?.forEach?.((data: any) =>
+                this.logEvent(LoggingCategory.InternalTransaction, {
```
nambrot, line 213: What exactly does this log? I thought we would try to parse the internal transaction as well?

packages/transaction-metrics-exporter/src/block-processor.ts
```diff
@@ -0,0 +1,266 @@
...
+            })
+
+            // @ts-ignore We want to rename event => eventName to avoid overwriting
+            event.eventName = event.event
+            delete event.event
+
+            this.logEvent(LoggingCategory.ParsedLog, event)
+          }
+          // tslint:disable-next-line
+        } catch {}
```
nambrot: I don't think we should just silently fail here?

packages/transaction-metrics-exporter/src/block-processor.ts
```diff
@@ -0,0 +1,266 @@
...
+            this.contracts[contract].setDefaultBlock('latest')
+            const data: DataResult = {
+              contract,
+              function: method,
+              args: JSON.stringify(args),
+              blockNumber,
+              values: transformValues(returnData),
+            }
+            this.logEvent(LoggingCategory.State, data)
+            this.logHistogram({ ...data, args }, maxBucketSize)
```
nambrot, line 152: This means that every function call that we want to do is always going to be logged as a histogram? Is that reasonable?

packages/transaction-metrics-exporter/src/states.ts
```diff
@@ -0,0 +1,108 @@
...
+  ) => { [key in RV]: any },
+  args:
+    | ArgumentTypes<Contracts[T][M] & Function>
+    | ((blockNumber: number) => ArgumentTypes<Contracts[T][M] & Function>) = [] as any,
+  maxBucketSize: { [key in RV]: any } = {} as any
+): StateGetter {
+  return { contract, method, args, transformValues, maxBucketSize }
+}
+
+export const stateGetters: StateGetter[] = [
```
nambrot, line 45: If I see correctly here, you are trying to generalize all state getters, but could you maybe elaborate on why you did so? As far as I can see, you are losing the Type benefits, the code becomes harder to read, and the metrics that are being logged become harder to reason about.

packages/contractkit/src/wallets/ledger-wallet.ts
Pedro-vk: Yes, it produces an error on documentation if we don't add it

packages/transaction-metrics-exporter/src/blockchain.ts
Pedro-vk: I'm not sure, but the environment variables will work on both scenarios

packages/transaction-metrics-exporter/src/block-processor.ts
Pedro-vk: It is exactly what we do. This line is logging each of the internal transactions that have been parsed previously.

packages/transaction-metrics-exporter/src/states.ts
Pedro-vk: @nambrot 
- I did it to ensure that each metric is correct, uses the contract, method, arguments that are specified in the contract method
- I did this in this way because the previous code was too long and repetitive. With this getters, we have only the necessary information to add a metric, without repeating the same code several times.
- If the metrics that are logged are hard to understand, we can work on the result that is reported. Let me know how do you think that we should report it.

packages/transaction-metrics-exporter/src/block-processor.ts
diminator: For doing quick sampling long range for analysis (stability, reserve, rewards, exchange, ...)

packages/transaction-metrics-exporter/src/block-processor.ts
diminator: probably the metric format needs to become part of the state view definition / config

@Pedro-vk can you log a new issue for this?

Pedro-vk: @diminator Done: https://github.com/celo-org/celo-monorepo/issues/3542

packages/transaction-metrics-exporter/src/blockchain.ts
diminator: see #3543

packages/transaction-metrics-exporter/src/states.ts
nambrot: What repetition are we exactly avoiding? Maybe I'm missing something, but actually just calling these methods to me does not sound that repetitive. I get that the motivation here was to get to a config-like state, however to a reader this is IMO more confusing since one has to jump back and forth to really grasp how this works.

Separately, we totally lose the compiler enforcement of the contractkit methods

barbaraliau:
Is this PR still relevant?

PR #3516, opened 2020-04-21
[Wallet] Improve support flow

### LINKED ISSUE #3150: Minor Support Flow Improvements
### Expected Behavior

1) On successful send of a report, redirect them back to the app on the settings page and flash a toast with the following success message:
"Your issue has been successfully reported! Expect an email response soon!"

2) add testnet name and address to the JSON blob in the email text
### Current Behavior

What is the current behavior?

1) we redirect back to the Contact form

2) no testnet name or address in JSON blob

i1skn:
## Description

Improving UX on Support flow

## Other changes

Fixing issue reported by Trevor: Tapping the "Get Backup Key" banner which takes you to the "Backup and Recovery" page and pressing the back arrow takes you to the settings view no matter what the original view was

## Tested

iOS device

## Related issues

- Fixes #3150

## Backwards compatibility

Yes

## Commit Message
[Wallet] Improve support flow


packages/mobile/src/account/SupportContact.tsx
```diff
@@ -65,8 +83,14 @@ const Support = () => {
...
     setInProgress(false)
-    Mailer.mail(email, (error: any, event: any) => {
-      Logger.showError(error + ' ' + event)
+    Mailer.mail(email, (error: any, event: string) => {
+      if (event === 'sent') {
+        navigate(Screens.Account)
+        props.showMessage(t('contactSuccess'))
+      }
+      if (error) {
+        Logger.showError(error + ' ' + event)
```
jmrossy, line 92: Should we use the store's showError here instead of Logger's toast?

i1skn: As far as I understand `showError` is used in places, where we sort of expect an error and we create a translation for this:
https://github.com/celo-org/celo-monorepo/blob/083604438dd0277684bd0ca4a28cc629f83ecf79/packages/mobile/src/alert/actions.ts#L36-L46
Here we just log if something goes wrong. WDYT?

jmrossy: I think `showError` would be a bit better but it's really nbd :)

jmrossy (approved):

PR #3521, opened 2020-04-21
ReleaseGold deployment script handles shards

lucasege:
## Description

Handling balance sharding on the deploy side. This is done by taking an initial address via `-f`, then once this is insufficient for the next grant, it sends the balance to your next provided address.

Some extra safe guards are taken (verifying 1 cGLD transfer first) to make this safer.

## Other changes

Various style

## Tested

Testing myself locally, will be testing a real grant file after publishing this

## Backwards compatibility

Yes


packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -144,6 +147,92 @@ async function handleGrant(releaseGoldConfig: any, currGrant: number) {
   }
 }
 
+async function checkBalance(releaseGoldConfig: any) {
+  const weiAmountReleasedPerPeriod = new BigNumber(
+    web3.utils.toWei(releaseGoldConfig.amountReleasedPerPeriod.toString())
+  )
+  const grantTotal = weiAmountReleasedPerPeriod
```
asaj: Can we add 1 cGLD here for tx fees?

Maybe call this variable `grantDeploymentCost` instead of `grantTotal` to reflect that

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -144,6 +147,92 @@ async function handleGrant(releaseGoldConfig: any, currGrant: number) {
...
+          "'s balance will not cover the next grant with identifier " +
+          releaseGoldConfig.identifier +
+          '.'
+      )
+    )
+    const addressResponse = await prompts({
+      type: 'text',
+      name: 'newFromAddress',
+      message:
+        'If you would like to continue deployment from a new address please provide it now. If you would like to cancel deployment just press enter. Grants already deployed will persist.',
```
asaj: minor nit: I worry about accidentally hitting enter here, can we instead require "exit" to be typed in to exit?

asaj: Maybe include "Note that the key for the current `from` address needs to remain accessible in order to roll funds over to the new `from` address"

asaj (approved):

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -144,6 +147,92 @@ async function handleGrant(releaseGoldConfig: any, currGrant: number) {
...
+      continue
+    }
+    console.info(
+      '\nSending 1 cGLD as a test from ' +
+        fromAddress +
+        ' to ' +
+        addressResponse.newFromAddress +
+        ' to verify ownership.\n'
+    )
+    await web3.eth.sendTransaction({
```
asaj: Just flagging that we are assuming here that we will never wind up with a leftover balance of < 1 cGLD. Seems like an assumption that is probably safe but may be worth skipping the test transaction if we have < 1 cGLD

packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -144,6 +147,92 @@ async function handleGrant(releaseGoldConfig: any, currGrant: number) {
...
+    if (!confirmResponse.confirmation) {
+      console.info(chalk.red('Setting new address failed.\nRetrying.'))
+      continue
+    }
+    console.info(chalk.green('\nTransfer confirmed, sending the remaining balance.\n'))
+    const fromBalancePostTransfer = new BigNumber(await web3.eth.getBalance(fromAddress))
+    await web3.eth.sendTransaction({
+      from: fromAddress,
+      to: addressResponse.newFromAddress,
+      value: fromBalancePostTransfer,
```
asaj: Wait actually I'm surprised this works, we should not be able to transfer our entire balance since we need to pay for tx fees

PR #3524, opened 2020-04-21
[ContractKit] Add secret retrieval capability to AzureKeyVaultClient

codyborn:
## Description

Add secret retrieval capability to AzureKeyVaultClient. This is needed for the phone number service to use BLS keys in KeyVault.

## Tested

Tested as part of this PR:
https://github.com/celo-org/celo-monorepo/pull/3508


tkporter (approved):
LGTM!

jmrossy (approved):
LGTM too but may be good to get signoff from @mcortesi or @gastonponti as well :)

packages/contractkit/src/utils/azure-key-vault-client.ts
```diff
@@ -123,6 +126,14 @@ export class AzureKeyVaultClient {
     return true
   }
 
+  public async getSecret(secretName: string): Promise<string> {
+    const secret = await this.secretClient.getSecret(secretName)
+    if (typeof secret.value === 'undefined') {
```
gastonponti: you could check with just `if (secret.value)`
for example an empty string it's a valid one?

gastonponti (commented):
All the other things looks good, but I have my doubts with that typeof === 'undefined'

packages/contractkit/src/utils/azure-key-vault-client.ts
codyborn: Technically empty string is valid, but unlikely correct.

gastonponti (approved):
Great, LGTM

PR #3529, opened 2020-04-21
More redirects to celo docs

marekolszewski:
Fixes and adds more redirects to the celo docs website.


timmoreton (commented):
The original URL we shared was `operations-manual/using-a-ledger-wallet.md` - does this cover that case? (https://github.com/celo-org/celo-monorepo/commit/8b3d5e12800665022f69aed4e8232ae77df5bde5#diff-bbec661115decc42a52ca40e2890ee13L21 )

timmoreton (approved):

PR #3533, opened 2020-04-22
Allow minting of zero value for GoldToken and StableToken

nategraf:
## Description

Minting of zero value is currently disallowed in the GoldToken and StableToken contracts. This PR allows minting zero value.

### DIFF
--- a/packages/protocol/contracts/common/GoldToken.sol
+++ b/packages/protocol/contracts/common/GoldToken.sol
@@ -141,7 +141,6 @@ contract GoldToken is Initializable, CalledByVm, Freezable, IERC20, ICeloToken {
    * @param value The amount of cGLD to mint.
    */
   function mint(address to, uint256 value) external onlyVm returns (bool) {
-    require(value >= 0, "mint value must be >= 0");
     if (value == 0) {
       return true;
     }
--- a/packages/protocol/contracts/common/GoldToken.sol
+++ b/packages/protocol/contracts/common/GoldToken.sol
@@ -141,7 +141,11 @@ contract GoldToken is Initializable, CalledByVm, Freezable, IERC20, ICeloToken {
    * @param value The amount of cGLD to mint.
    */
   function mint(address to, uint256 value) external onlyVm returns (bool) {
-    require(value > 0, "mint value must be > 0");
+    require(value >= 0, "mint value must be >= 0");
+    if (value == 0) {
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/common/GoldToken.sol
```diff
@@ -141,7 +141,11 @@ contract GoldToken is Initializable, CalledByVm, Freezable, IERC20, ICeloToken {
    * @param value The amount of cGLD to mint.
    */
   function mint(address to, uint256 value) external onlyVm returns (bool) {
-    require(value > 0, "mint value must be > 0");
+    require(value >= 0, "mint value must be >= 0");
```
asaj: unnecessary check for unsigned int, please remove

packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -218,7 +218,11 @@ contract StableToken is
    */
   function _mint(address to, uint256 value) private returns (bool) {
     require(to != address(0), "0 is a reserved address");
-    require(value > 0, "mint value must be > 0");
+    require(value >= 0, "mint value must be >= 0");
```
asaj: unnecessary check for unsigned int, please remove

packages/protocol/test/stability/stabletoken.ts
```diff
@@ -146,6 +146,18 @@ contract('StableToken', (accounts: string[]) => {
...
 
+    it('should allow the minting 0 value', async () => {
+      await stableToken.mint(validators, 0, { from: validators })
+      const balance = (await stableToken.balanceOf(validators)).toNumber()
+      assert.equal(balance, 0)
+      const supply = (await stableToken.totalSupply()).toNumber()
+      assert.equal(supply, 0)
+    })
+
+    it('should not allow the minting negative value', async () => {
```
asaj: I'm not sure whats' being tested here

asaj (approved):

packages/protocol/contracts/common/GoldToken.sol
nategraf: Ah, right!

packages/protocol/test/stability/stabletoken.ts
nategraf: Absolutely nothing!

PR #3536, opened 2020-04-22
Minor adjustments to RG deploy script

asaj:
## Description

- startGold for beneficiary deducted from RG balance
- minor refactor for `ONE_CGLD`
- Use `toFixed()` when sending txs with bignumbers

## Other changes

None
## Tested

No
## Related issues

None
## Backwards compatibility

Yes


lucasege (changes requested):
I think we also need to change the gold amount passed into the ReleaseGold initialization Params, right?

Also there are some places where we print the grant balance, can you verify those include this reduction too?

lucasege (approved):
LGTM

PR #3538, opened 2020-04-22
Add celo protocol tx URI scheme and qr codes to contractkit

### LINKED ISSUE #3052: Design/develop standard URI scheme for SDK transactions/QR code usage (Gitcoin bounties)
### Expected Behavior

### Current Behavior

What is the current behavior?

yorhodes:
## Description

- Adds `contractkit/src/utils/tx-uri.ts` to provide 2-way conversion between transaction URIs and transaction objects
- Adds regexer for URI scheme
- Adds qr code generator from URI scheme (svg default, terminal option)

## Other changes

Adds qr code dependency

## Tested

See `utils/tx-uri.test.ts` for cGLD and cUSD transfer tests

## Related issues

- Fixes #3052 

## Backwards compatibility

Yes

## Commit Message
Add celo protocol tx URI scheme and qr codes to contractkit


nityas:
from product perspective, params/design LGTM 👍

packages/contractkit/src/utils/tx-uri.ts
```diff
@@ -0,0 +1,119 @@
...
+          args = args.concat(queryParams[idx])
+          delete queryParams[idx]
+        })
+        const callSig = abi.encodeParameters(abiTypes, args)
+
+        tx.data += trimLeading0x(callSig)
+      }
+    }
+
+    tx = { ...tx, ...queryParams }
```
gastonponti: This `queryParams` will be empty if you have inputTypes, because you are deleting every... Is that something expected?
and assume that queryParams is a `key: number`, and not a `key: string`, right? (just checking)

packages/contractkit/src/utils/tx-uri.ts
```diff
@@ -0,0 +1,119 @@
...
+  let uri = `celo:${tx.to!}`
+
+  if (tx.chainId) {
+    uri += `@${tx.chainId}`
+  }
+
+  let paramsAppended = false
+  if (tx.data !== undefined) {
+    if (!functionName || !abiTypes) {
+      throw new Error("Cannot decode tx 'data' without 'functionName' and 'abiTypes'")
```
gastonponti: Again, don't know if it's a bug or not, but in the error message you are saying that you need both to continue, and here it fails if one of those are missing. (probably just the message)

gastonponti: yes just the message, haha

packages/contractkit/src/utils/tx-uri.ts
```diff
@@ -0,0 +1,119 @@
...
+    const functionSig = trimLeading0x(abi.encodeFunctionSignature(functionSelector))
+    const funcEncoded = trimLeading0x(tx.data).slice(0, 8)
+
+    if (functionSig !== funcEncoded) {
+      throw new Error("'functionName' does not match first 4 encoded bytes of 'tx.data'")
+    }
+
+    uri += `/${functionSelector}`
+
+    if (tx.data.length > 8) {
```
gastonponti: You should `trimLeading0x`, otherwise you could be getting data from the first "4 encoded bytes of tx.data"

packages/contractkit/src/utils/tx-uri.ts
```diff
@@ -0,0 +1,119 @@
...
+      const argsEncoded = trimLeading0x(tx.data).slice(8)
+      const decoded = abi.decodeParameters(abiTypes, argsEncoded)
+
+      delete decoded.__length__
```
gastonponti: 👀

packages/contractkit/src/utils/tx-uri.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  const value = '100'
+  const simpleTransferTx: Tx = {
+    value,
+    to: recipient,
+  }
+  const simpleTransferUri = `celo:${recipient}?value=${value}`
+  let stableTokenTransferUri: string
+  let stableTokenTransferTx: Tx
+
+  beforeAll(async () => {
```
gastonponti, line 24: could you add a test for a function without parameters? (or this will be only for transfers?)

gastonponti: could a "-1" params, or a third param for a 2 params function be a problem? It's curiosity, no one will generate something that does not work

gastonponti (commented):
The only possible issue is the `trim0x`, the other are questions

packages/contractkit/src/utils/tx-uri.ts
yorhodes: lol so I can spread the annoying web3 type

packages/contractkit/src/utils/tx-uri.ts
yorhodes: queryParams is a `interface ParsedUrlQuery { [key: string]: string | string[]; }`
I am attempting to remove all of the number keys (contract function args) so they don't get added to the Tx object when spreading (see the tests)
Let me know if you have a better way to do this

packages/contractkit/src/utils/tx-uri.test.ts
yorhodes: any additional params outside of `0:abiParamLength` will end up in the tx object... see my question above

packages/contractkit/src/utils/tx-uri.ts
gastonponti: hahaha 👍

gastonponti (approved):
I have my doubts on the key casting to string, but, it the test is working... hahaha
👍

gastonponti (approved):
Yeah, definitely cleaner!

PR #3539, opened 2020-04-22
Segment TME logs to allow for separate deployments

nambrot:
## Description

This PR adds a suffix to the TME helm chart name and container name to allow for seprate deployments of TME to write different logs. 

## Other changes

- Deployed to RC0


jcortejoso (approved):
lgtm

PR #3546, opened 2020-04-22
[Wallet] Adding recent payments to the state so we can check for 24 hour limit

### LINKED ISSUE #3063: FATF Compliance and User Education- transaction limits $500 cUSD/day
### Expected Behavior

https://www.figma.com/file/PyK5VI7mE2fpHLeRPL0yxB/Transaction-Limit?node-id=0%3A1

### Current Behavior

What is the current behavior?

### LINKED ISSUE #3329: Wallet transaction limits $500 cUSD/day
### Expected Behavior

Wallet can't send more than $500 USD / day
Part of #3063

### Current Behavior

No limits

etuleu:
## Description

This change adds a limit of $500 per 24 hours for making payments.
To accomplish this we need to store a list of recent payments with their timestamps, which we append to after payment is initiated. The interface then checks and displays an error if the limit is reached.

## Tested

Added unit tests for utils.

## Related issues

- fixes #3329 
- fixes https://github.com/celo-org/celo-monorepo/issues/3063

## Backwards compatibility

This change should be backwards compatible because we are only adding new properties to the state. We will not be able to check for payments sent before we started storing this so it will only function going forward. Since the limit is daily this should be fine.

## Commit Message

Adding recent payments to the redux state so we can check for 24 hour limit.


packages/mobile/src/send/reducers.ts
```diff
@@ -48,13 +56,25 @@ export const sendReducer = (
   }
 }
 
+const sendPaymentOrInvite = (state: State, amount: BigNumber) => {
+  const latestPayment = { timestamp: Date.now(), amount }
+  const recentPayments = [...state.recentPayments, latestPayment]
```
etuleu: @jmrossy basically here we could filter out entries that are older than 24 hours. But my worry is would that violate the "pureness" of redux? If we pass in a timestamp through the action would that make it "more pure"?

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -206,6 +215,21 @@ export class SendAmount extends React.Component<Props, State> {
...
+      (p: PaymentInfo) => timeDeltaInHours(now, p.timestamp) < 24
+    )
+
+    const initial = new BigNumber(this.state.amount)
+    const amount: BigNumber = paymentsLast24Hours.reduce(
+      (sum, p: PaymentInfo) => sum.plus(p.amount),
+      initial
+    )
+
+    return amount <= DAILY_PAYMENT_LIMIT_USD
```
aaronmgdr: i'ts reached if amount is greater than DAILY_LIMIT

 Doesnt this return true if amount is less than or equal to DAILY_LIMIT ?

packages/mobile/src/send/reducers.ts
```diff
@@ -48,13 +56,25 @@ export const sendReducer = (
   }
 }
 
+const sendPaymentOrInvite = (state: State, amount: BigNumber) => {
+  const latestPayment = { timestamp: Date.now(), amount }
```
aaronmgdr: Before we talked about  using the payment success action. otherwise failed unsent payments count against your total which is not so fun. 

also you were gonna pass timestamp in to the action from the saga, which i thought was a great idea

packages/mobile/src/app/ErrorMessages.ts
```diff
@@ -29,6 +29,7 @@ export enum ErrorMessages {
   INVITE_FAILED = 'inviteFailed',
   INVITE_OPEN_APP_FAILED = 'inviteOpenAppFailed',
   SEND_PAYMENT_FAILED = 'sendPaymentFailed',
+  PAYMENT_LIMIT_REACHED = 'paymentLimitReached',
```
jmrossy, line 35: This string needs to map to an error message string in /mobile/locales/en-US/global.json and /mobile/locales/es-419/global.json

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -58,16 +58,21 @@ import {
...
 import { fetchDollarBalance } from 'src/stableToken/actions'
+import { timeDeltaInHours } from 'src/utils/time'
 import { withDecimalSeparator } from 'src/utils/withDecimalSeparator'
 
 const AmountInput = withDecimalSeparator(
   withTextInputLabeling<ValidatedTextInputProps<DecimalValidatorProps>>(ValidatedTextInput)
 )
 const CommentInput = withTextInputLabeling<TextInputProps>(TextInput)
 
+const DAILY_PAYMENT_LIMIT_USD: BigNumber = new BigNumber(500)
```
jmrossy: Please move this to Config.ts and just make it a `Number`

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -206,6 +215,21 @@ export class SendAmount extends React.Component<Props, State> {
     }
   }
 
+  isPaymentLimitReached = (now: number) => {
```
jmrossy: Can we add a unit test for this? It may be easier to do if you pull this method out of the component and move it to somewhere in `/utils`

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -206,6 +215,21 @@ export class SendAmount extends React.Component<Props, State> {
...
+    // we are only interested in the last 24 hours
+    const paymentsLast24Hours = this.props.recentPayments.filter(
+      (p: PaymentInfo) => timeDeltaInHours(now, p.timestamp) < 24
+    )
+
+    const initial = new BigNumber(this.state.amount)
```
jmrossy: Use `parseInputAmount` here, see elsewhere in this file for examples

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -206,6 +215,21 @@ export class SendAmount extends React.Component<Props, State> {
...
+    const amount: BigNumber = paymentsLast24Hours.reduce(
+      (sum, p: PaymentInfo) => sum.plus(p.amount),
```
jmrossy: Keep in mind that the amount is entered in a local currency, whereas the hard limit is in USD. If you need help figuring out the conversion utilities I think @jeanregisser can help :)

packages/mobile/src/send/SendAmount.tsx
jmrossy: Yeah, I think this is the inverse of the function name

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -253,6 +277,13 @@ export class SendAmount extends React.Component<Props, State> {
       return
     }
 
+    const now = Date.now()
```
jmrossy, line 287: Why pull now here and then pass it in instead of getting it in `isPaymentLimitReached`?

packages/mobile/src/send/reducers.ts
jmrossy: It's not perfect but I think calling for Date.now() here and filtering old ones is the most reasonable approach. Any others would probably be overly complex for no tangible benefit

jmrossy (commented):
Good first pass! Please see comments below

packages/mobile/src/send/reducers.ts
aaronmgdr: @jmrossy im surprised at that sense it would mean the reducer wouldnt be pure which I thought was pretty foundational to redux?

jmrossy: @aaronmgdr I disagree. The pureness of a reducer is intended to mean that it has no side-effects other than the state modifications for actions it handles. This is just a state modification for a particular action, seems well within realm of intended use of reducers

aaronmgdr: ah cool,

packages/mobile/src/send/SendAmount.tsx
etuleu: good catch! just testing you guys :) jk

packages/mobile/src/send/reducers.ts
etuleu: Yes, I am still not completely sure about the best place but here is my thinking, after discussing a bit with Connor also. If we store in payment success only then someone in theory could go into airplane mode, then enqueue a bunch of transactions and then turn airplane off. That would allow him to go over the limit.

This is why I think doing it in sendPayment, and then removing failed transactions (I still need to implement that) is a bit better. But, I am open to suggestions of course.

packages/mobile/src/app/ErrorMessages.ts
etuleu: Done

etuleu: I will work with Aaron on top of this PR to make sure we can pass in the data we need and the formatting looks good. He has a change that allows passing the data to the error component I think.

packages/mobile/src/send/SendAmount.tsx
etuleu: Done.

packages/mobile/src/send/SendAmount.tsx
etuleu: Moved to utils. Working on unit tests next.

etuleu: I also passed the recent payments as a parameter to make unit testing nicer.

packages/mobile/src/send/SendAmount.tsx
etuleu: I switched to using the existing method this.getDollarsAmount() and then calling toNumber() on it. I will ping Jean as well to make sure this is the right method to use.

etuleu: @jeanregisser could you please take a look at this PR, in particular to the local currency conversions?

packages/mobile/src/send/SendAmount.tsx
etuleu: @jeanregisser I am now using this.getDollarsAmount local method. Is that a good choice?

packages/mobile/src/send/SendAmount.tsx
etuleu: I think this would make unit testing nicer, instead of having to deal with time traveling libraries? it is kind of weird but then the function itself is more deterministic was my thinking. wdyt?

packages/mobile/src/send/reducers.ts
etuleu: yea I kept looking for examples of how other people handle time series or time based data structures but couldn't find anything serious. So, filter here? What about Aaron's suggestion to pass the timestamp from the saga in the action? That seems pretty good as well.

packages/mobile/locales/en-US/global.json
```diff
@@ -113,5 +113,6 @@
   "timeframes": {
     "30d": "30 days"
   },
-  "message": "Message"
+  "message": "Message",
+  "paymentLimitReached": "The daily limit is {{dailyLimit}} {{dailyRemaining}} remaining"
```
etuleu: I am not sure about how to pass these parameters but @aaronmgdr is looking into it.

packages/mobile/locales/es-419/global.json
```diff
@@ -113,5 +113,6 @@
   "timeframes": {
     "30d": "30 dias"
   },
-  "message": "Mensaje"
+  "message": "Mensaje",
+  "paymentLimitReached": "El límite diario es {{dailyLimit}} {{dailyRemaining}} restante"
```
etuleu: What is the process for translations usually? This is most likely wrong.

aaronmgdr: There is spanish channel on  slack

packages/mobile/src/send/reducers.ts
aaronmgdr: One Direct benefit I see of passing in a timestamp is that it makes testing the reducer much simpler.

aaronmgdr:
How UI looks in various languages / currencies

EDIT: I fixed the missing CUSD 

Screen Shot 2020-04-29 at 5 19 56 PM
Screen Shot 2020-04-29 at 5 21 21 PM
Screen Shot 2020-04-29 at 5 22 18 PM

packages/mobile/src/send/SendAmount.tsx
aaronmgdr: another benefit is that we need the time when getting the amountRemaining and by passing it into both we are certain to use the same time in both place otherwise we could technically have people who try and send 24.5999 hours after sending enough to put them over get an error but the the error message will say they have some remaining.

packages/mobile/src/alert/actions.ts
```diff
@@ -33,11 +33,15 @@ export const showMessage = (
   return showAlert(AlertTypes.MESSAGE, message, dismissAfter, buttonMessage, title)
 }
 
-export const showError = (error: ErrorMessages, dismissAfter?: number): ShowAlertAction => {
+export const showError = (
+  error: ErrorMessages,
+  dismissAfter?: number | null,
+  options?: object
```
jmrossy: Consider renaming to i18n options for clarity

jmrossy (approved):
LGTM, one optional comment below

packages/mobile/src/send/reducers.ts
```diff
@@ -48,13 +57,32 @@ export const sendReducer = (
   }
 }
 
+const sendPaymentOrInvite = (state: State, amount: BigNumber) => {
+  const timestamp = Date.now()
```
i1skn: Calling `now()` here would make the reducer not pure. What do you think if we make `timestamp` a part of `SEND_PAYMENT_OR_INVITE` action?

i1skn (commented):
Great work! Overall looks goof to me! Left just one little comment

packages/mobile/src/send/reducers.ts
etuleu: That makes sense. I think Aaron suggested that but I forgot to implement it. Will do now and commit. Thanks for the review!

PR #3548, opened 2020-04-22
Point to Blockscout of RC1 on Validator Explorer

Pedro-vk:


aaronmgdr (approved):
trusting this is right.

PR #3555, opened 2020-04-22
ReleaseGold script should compare left over balance to 2 cGLD

lucasege:
## Description
This is a copy of a PR on rc1 that is important to add here.

Left over balance from collecting shards must be greater than 2cGLD to account for the 1cGLD transfer verification and 1cGLD for TX fees.

## Tested

Yes

## Backwards compatibility

Yes


lucasege:
This was lumped in with a separate PR.
Closing.

Issue #1777, opened 2019-11-19
Secure Send


nityas:
### Expected Behavior

three text messages should be treated as verified, any less should be considered 'unverified'

[Figma](https://www.figma.com/file/i3KEZWz9sKiIl3gcwQeqHn/Secure-Send?node-id=178%3A0)

### Current Behavior

one text treated as verified, three texts required to redeem from escrow

jmrossy:
Related: https://github.com/celo-org/celo-monorepo/issues/2438

coreycelo:
Added Figma link

PR #3558, opened 2020-04-23
Phone Number Privacy in Wallet (Part 1)

jmrossy:
## Description

- Implement wallet-side changes for phone number privacy
- Add react-native-bls-threshold lib to wallet
- Update ContractKit Attestation wrapper to work with salted hashes
- Add partial support for phone numbers with multiple addresses 
- Remove contact sync code from dappkit as its no longer viable
- Cleanup some other identity related code in the wallet
- De-dupe mocks for payment requests
- Add Redux store migration for the schema changes

This adds the new RN-bls-threshold lib to the wallet, for that see here:
https://github.com/celo-org/react-native-blind-threshold-bls

And that lib is itself based on this rust project:
https://github.com/celo-org/celo-threshold-bls-rs

## Tested

Tested full onboarding in app using the new PGPNP service to generate a salt.

## Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/329
- Facilitates #1777

## Backwards compatibility

No. The phone hashing is different than before, verification now fetches salts before starting, and contact mapping in the app is removed


packages/mobile/.env
```diff
@@ -2,7 +2,7 @@ ENVIRONMENT=local
 DEFAULT_TESTNET=alfajores
 # If FORNO_ENABLED_INITIALLY, local geth will not run initially.
 # If toggled on, it will use DEFAULT_SYNC_MODE. See src/geth/consts.ts for more info
-FORNO_ENABLED_INITIALLY=false
+FORNO_ENABLED_INITIALLY=true
```
jmrossy: TODO revert this

packages/dappkit/src/index.ts
```diff
@@ -163,122 +159,3 @@ export async function requestTxSig(
 
   Linking.openURL(serializeDappKitRequestDeeplink(request))
 }
-
-function isValidPhoneNumber(phoneNumber: PhoneNumber): E164Number | undefined {
```
jmrossy, line 167: Lookups are no longer feasible in dappkit given the quota-restricted salt queries required. I'm removing this until we figure out some alternative.

packages/dappkit/README.md
```diff
@@ -66,34 +66,6 @@ Once you have the account address, you can make calls against your own smart con
   this.setState({ cUSDBalance, isLoadingBalance: false })
 ```
 
-## Get Contact List Addresses
```
jmrossy, line 69: Lookups are no longer feasible in dappkit given the quota-restricted salt queries required. I'm removing this stuff until we figure out some alternative.

mcortesi (commented):
contract kit changes lgtm!

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -12,18 +12,28 @@ import { AttestationModel, AttestationStatus } from '../models/attestation'
...
   phoneNumber: E164PhoneNumberType,
   account: AddressType,
   issuer: AddressType,
 })
 
+const AttestationRequestOptional = t.partial({
+  salt: SaltType,
+})
+
+export const AttestationRequestType = t.intersection([
```
aslawson: i'm not super familiar with this library, but it seems like this should be a union rather than an intersection...?

packages/celotool/src/cmds/account/lookup.ts
```diff
@@ -13,24 +13,37 @@ export const describe = 'command for lookup of accounts for a given identifier'
...
+    })
 }
 
 export const handler = async (argv: LookupArgv) => {
   await switchToClusterFromEnv(false)
   console.log(`Looking up addresses attested to ${argv.phone}`)
+  if (!argv.salt) {
+    console.warn(
+      `Warning: you have not specified a salt so this phone number's hash will not be the same as the one used by the wallet app`
+    )
```
aslawson, line 40: ultimately this returns the address, not the phone number hash, so I would think this command would just fail without the salt.
 
would it be possible that there would be a mapping for a non-salted number?  eg. are we expecting other applications to use the attestations service to add unsalted mappings (and if so, should we be worried that those app's would expose phone numbers to harvesting)

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -105,164 +93,65 @@ function* updateRecipientsCache(
...
   try {
-    Logger.debug(TAG, `Fetch and store address mapping for ${e164Numbers.length} phone numbers`)
-
-    const addresses: Array<string | null> = yield call(retryAsync, getAddresses, 3, [
-      e164Numbers,
-      attestationsWrapper,
+    Logger.debug(TAG + '@fetchPhoneAddresses', `Fetching addresses for number`)
+    // Clear existing entries for those numbers so our mapping consumers
+    // know new status is pending.
+    yield put(updateE164PhoneNumberAddresses({ [e164Number]: undefined }, {}))
```
aslawson, line 102: i'm a little confused how this works but seems this action updates the e164Number object.

why is this not fetching the salt before getting Attestations with updateE164PhoneNumberSalts?

packages/mobile/src/identity/privacy.ts
```diff
@@ -0,0 +1,138 @@
...
+      const phoneHash = getPhoneHash(e164Number, cachedSalt)
+      return { e164Number, phoneHash, salt: cachedSalt }
+    }
+
+    Logger.debug(`${TAG}@fetchPrivatePhoneHash`, 'Salt was not cached, fetching')
+    const details: PhoneNumberHashDetails = yield call(getPhoneHashPrivate, e164Number)
+    yield put(updateE164PhoneNumberSalts({ [e164Number]: details.salt }))
+    return details
+  } catch (error) {
+    if (error.message === ErrorMessages.SALT_QUOTA_EXCEEDED) {
```
aslawson, line 41: should we be syncing on this error messages?

aslawson (commented):
wow quite long!  only got halfway through -- i'll continue looking tomorrow

packages/celotool/src/lib/attestation.ts
```diff
@@ -45,6 +45,7 @@ export async function requestAttestationsFromIssuers(
     try {
       const response = await attestations.revealPhoneNumberToIssuer(
         phoneNumber,
+        '',
```
i1skn: I see that in `attestation-service/src/requestHandlers/attestation.ts` salt is optional, shall we make it optional here as well to avoid passing an empty string?

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -64,6 +60,9 @@ export interface UnselectedRequest {
   attestationRequestFeeToken: string
 }
 
+// Map of identifier -> (Map of address -> AttestationStat)
+export type IdentifierLookupResult = Record<string, Record<string, AttestationStat>>
```
i1skn: Should it be `Record<string, Record<Address, AttestationStat>>` ?

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -105,164 +93,65 @@ function* updateRecipientsCache(
...
-      } else {
-        // Save invalid/0 addresses to avoid checking again
-        // null means a contact is unverified, whereas undefined means we haven't checked yet
-        e164NumberToAddressUpdates[e164Number] = null
-      }
+    if (!addresses) {
+      Logger.debug(TAG + '@fetchPhoneAddresses', `No addresses for for number`)
+      // Save invalid/0 addresses to avoid checking again
+      // null means a contact is unverified, whereas undefined means we haven't checked yet
+      e164NumberToAddressUpdates[e164Number] = null
```
i1skn, line 119: I feel a bit concerned about using `null` and `undefined` to represent different logical states. Even though, that is probably how people should've been using them, in reality it would be hard to catch some potentially dangerous code like this one in the future:
 ```javascript
if (!e164NumberToAddressUpdates[e164Number]) {
    // do stuff
 }
```
WDYT?

File: packages/mobile/src/identity/contactMapping.ts [hunk 11]
```diff
@@ -271,7 +160,30 @@ export function getAddressFromPhoneNumber(
...
+
+  if (!addresses) {
+    return addresses
+  }
+
+  if (addresses.length === 0) {
+    throw new Error('Phone addresses array should never be empty')
+  }
+
+  if (addresses.length === 1) {
```
i1skn: Should it be `addresses.length > 1`?

packages/mobile/src/identity/privacy.ts
```diff
@@ -0,0 +1,138 @@
...
+      blindPhoneNumber: base64BlindedMessage,
+    }),
+  })
+
+  if (!res.ok) {
+    handleSignMessageFailure(res)
+  }
+
+  Logger.debug(`${TAG}@postToSignMessage`, 'Response ok. Parsing.')
+  const signResponse = (await res.json()) as SignMessageResponse
```
i1skn, line 120: Do we want to `try ... catch` here?

packages/mobile/src/recipients/RecipientPicker.tsx
```diff
@@ -176,11 +174,10 @@ export class RecipientPicker extends React.Component<RecipientProps> {
   }
 
   render() {
-    const { sections, listHeaderComponent, showContactSyncBanner } = this.props
+    const { sections, listHeaderComponent } = this.props
 
     return (
       <View style={style.body} testID={this.props.testID}>
-        {showContactSyncBanner && <ContactSyncBanner />}
```
i1skn, line 183: I have not found any other place we use `ContactSyncBanner`, if this is correct do we want to remove it?

i1skn (commented):
A great work overall! Really exciting to see this coming to live! Left just some minor comments.

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -97,17 +106,14 @@ class AttestationRequestHandler {
     const address = getAccountAddress()
 
     // TODO: Check with the new Accounts.sol
-    if (address.toLowerCase() !== this.attestationRequest.issuer.toLowerCase()) {
+    if (address.toLowerCase() !== issuer.toLowerCase()) {
```
gastonponti: Minor detail. You could use `eqAddress` from `utils -> addresses` (this also compares `0x34` === `34` => true, which could help you with some weird cases)

packages/attestation-service/src/requestHandlers/attestation.ts
jmrossy: You would think so, but it seems like this is the way:
https://github.com/gcanti/io-ts/issues/56

packages/celotool/src/cmds/account/lookup.ts
jmrossy: Yes, if someone verifies with celotool, then they could also lookup with celotool. I don't want to totally prevent verification without salts, we just want the wallet to opt-in by default

packages/mobile/src/identity/contactMapping.ts
jmrossy: See the comment on lines 99-100 :)
This clears the cached phone number so we can check it again. No need to clear salt caches since salts never change.

packages/mobile/src/identity/privacy.ts
jmrossy: As in syncing with what we do in the cloud service? No this is a wallet-specific thing. The place where the wallet + cloud need to be in sync is the error code (403)

packages/mobile/src/recipients/RecipientPicker.tsx
jmrossy: I was considering that but Nitya recommended we keep it, she suspects the new designs may still use it. I'll add a TODO on ContactSyncBanner to capture this

packages/mobile/src/identity/privacy.ts
jmrossy: I'd rather let it propagate and have the try/catch in `fetchPhoneHashPrivate` handle it

packages/mobile/src/identity/contactMapping.ts
jmrossy: You're right, good catch thanks!!

packages/celotool/src/lib/attestation.ts
jmrossy: Good idea!

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: Yep! this was copy pasted from existing code but that's a nice improvement

packages/mobile/src/identity/contactMapping.ts
jmrossy: Yeah I don't love it either but I liked it slightly better than using some other Sentinel value like `-1`.
Then intention is not to have anyone access `e164NumberToAddressUpdates` directly, they should use the `getAddressFromPhoneNumber` helper. How about I add a comment explaining that?

i1skn: Yeah, sounds good!

gastonponti (commented):
LGTM

packages/mobile/src/identity/contactMapping.ts
aslawson: oh i see -- you mean the comments that I've literally highlighted and failed to read :P

ya that makes sense now

packages/attestation-service/src/requestHandlers/attestation.ts
jmrossy: Good idea! Will do

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -58,12 +54,22 @@ type AttestationServiceRunningCheckResult =
...
+}
+
 export interface UnselectedRequest {
   blockNumber: number
   attestationsRequested: number
   attestationRequestFeeToken: string
 }
 
+// Map of identifier -> (Map of address -> AttestationStat)
+export type IdentifierLookupResult = Record<string, Record<Address, AttestationStat>>
```
jeanregisser: It looks like the value can be `undefined` in the `Record` (or maybe `null` also?)

What do you think of changing it to something like:

```typescript
export type IdentifierLookupResult = Record<string, Record<Address, AttestationStat | undefined> | undefined>
```

To catch errors at compile when accessing lookup results.

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -105,164 +93,65 @@ function* updateRecipientsCache(
...
+
+  const addresses = Object.keys(results[phoneHash])
+    .filter(isValidAddress)
+    .map((a) => a.toLowerCase())
+  return addresses.length ? addresses : null
 }
 
+const isValidAddress = (address: string) =>
+  typeof address === 'string' && address.length && !new BigNumber(address).isZero()
+
```
jeanregisser: Should we use our `isValidAddress` from `utils` instead?

packages/mobile/src/import/saga.ts
```diff
@@ -78,10 +77,7 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
     yield put(redeemInviteSuccess())
     yield put(refreshAllBalances())
 
-    // Check if the account was verified
-    const isVerified = yield call(checkVerification)
-
-    navigate(isVerified ? Screens.WalletHome : Screens.VerificationEducationScreen)
+    navigate(Screens.VerificationEducationScreen)
```
jeanregisser, line 81: I'm missing some context here, we can't do the previous logic anymore in the new system?

packages/mobile/src/paymentRequest/OutgoingPaymentRequestSummaryNotification.test.tsx
```diff
@@ -2,89 +2,40 @@ import * as React from 'react'
 import 'react-native'
 import { Provider } from 'react-redux'
 import * as renderer from 'react-test-renderer'
-import { PaymentRequestStatus } from 'src/account/types'
-import { SHORT_CURRENCIES } from 'src/geth/consts'
 import OutgoingPaymentRequestSummaryNotification from 'src/paymentRequest/OutgoingPaymentRequestSummaryNotification'
 import { createMockStore } from 'test/utils'
+import { mockE164Number, mockPaymentRequests } from 'test/values'
```
jeanregisser, line 7: Nice factoring this 👍

packages/mobile/src/identity/privacy.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  Logger.debug(`${TAG}@getPhoneNumberSalt`, 'Retrieving blinded message')
+  const base64BlindedMessage = await BlindThresholdBls.blindMessage(e164Number)
+  const base64BlindSig = await postToSignMessage(base64BlindedMessage)
+  Logger.debug(`${TAG}@getPhoneNumberSalt`, 'Retrieving unblinded signature')
+  const base64UnblindedSig = await BlindThresholdBls.unblindMessage(base64BlindSig)
+  Logger.debug(`${TAG}@getPhoneNumberSalt`, 'Converting sig to salt')
+  return getSaltFromThresholdSignature(base64UnblindedSig)
+}
```
jeanregisser: This looks like the core logic in this new system.
I understand the output but I honestly don't understand what it does internally.
Could we explain this in a somewhat approachable way? 😅 
I know this is very new, but would be cool to document it publicly.

jeanregisser (commented):
Awesome work! 🎉 
Left some comments mostly because I don't understand everything around this.

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: Will do!

packages/mobile/src/identity/contactMapping.ts
jmrossy: We need to also check that it's not 0 but I'll use utils isValidAddress here too

packages/mobile/src/import/saga.ts
jmrossy: Correct, because we don't have the salt needed to get the hash to check verification status. Simply allowing the user to do verification again, which will be super quick if they're using the same number, is the simplest solution for now

packages/mobile/src/identity/privacy.ts
jmrossy: One of the epics for this milestone is to create public documentation for this :)

packages/attestation-service/src/requestHandlers/attestation.ts
nambrot: As mentioned in the linked issue, what do you think about just making `salt` a required key and have that be a union of `t.union([SaltType, t.undefined])`. That's what we are doing here as well https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/identity/claims/account.ts#L13-L14

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -110,29 +119,29 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
   )
 
+  /**
+   * @notice Waits for appropriate block numbers for before issuer can be selected
+   * @param identifier Attestation identifier (e.g. phone hash)
+   * @param account Address of the account
+   */
   waitForSelectingIssuers = async (
-    phoneNumber: string,
+    identifier: string,
```
nambrot, line 143: What do you think about making the salting the responsibility of the wrapper instead of the consumer? IMO it would allow consumers to just pass things through without having to manage the salting themselves

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -110,29 +119,29 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
     valueToInt
   )
 
-  /**
```
nambrot, line 113: Was removing this documentation block intentional?

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -392,70 +398,76 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
 
   revealPhoneNumberToIssuer(
     phoneNumber: string,
     account: Address,
     issuer: Address,
-    serviceURL: string
+    serviceURL: string,
+    salt?: string
   ) {
+    const body: AttesationServiceRevealRequest = {
```
nambrot, line 433: Would it be not fine to just pass `salt,` even if it is undefined?

packages/dappkit/README.md
nambrot: Can you also remove it from our docs or at least ensure that @nityas or @critesjosh are aware to do so?

nambrot (commented):
Great work!

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: Given the complex interplay between the app, the native libraries and the phone number privacy service, I'm not sure how that would be possible

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: Just tested, you're right. Will clean this up

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: It was not, nice catch

packages/contractkit/src/wrappers/Attestations.ts
nambrot: Maybe I'm misunderstanding, but can't `identifier: string` be replaced by `phoneNumber: string, salt?: string`?

jmrossy: Ah, I thought you wanted the the wrapper to handle actually *getting* the salts.
We _could_ do that, I guess it's just a matter of taste. I rather a simpler, generic interface that actually matches what the contract does. Also nice that it doesn't assume the id is a phone number for when we support other types.

jmrossy: Oh, also also we needed it this way for the phone number privacy service, which for matchmaking will handle hashed phone numbers but not salts

nambrot: Up to you. What do you think of having two methods for this? My understanding of having wrappers in the first place is to have them abstract transformation from the contract to the consumer in the first place, but @mcortesi would know better

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -97,17 +100,14 @@ class AttestationRequestHandler {
     const address = getAccountAddress()
 
     // TODO: Check with the new Accounts.sol
-    if (address.toLowerCase() !== this.attestationRequest.issuer.toLowerCase()) {
+    if (eqAddress(address, issuer)) {
```
codyborn: Should be:
`if (!eqAddress(address, issuer))`

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -97,17 +100,14 @@ class AttestationRequestHandler {
...
       throw new Error(`Mismatching issuer, I am ${address}`)
     }
 
     const attestations = await kit.contracts.getAttestations()
-    const state = await attestations.getAttestationState(
-      this.attestationRequest.phoneNumber,
-      this.attestationRequest.account,
-      this.attestationRequest.issuer
-    )
+    const identifer = PhoneNumberUtils.getPhoneHash(phoneNumber, salt)
```
codyborn: nitpick: identifier

packages/attestation-service/src/requestHandlers/attestation.ts
jmrossy: Good catch thanks!

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -119,20 +119,24 @@ class AttestationRequestHandler {
...
+      salt
     )
 
     return kit.web3.eth.sign(message, getAttestationSignerAddress())
   }
 
   async validateAttestation(attestationCode: string) {
+    const { phoneNumber, account, salt } = this.attestationRequest
     const address = getAccountAddress()
+    const identifer = PhoneNumberUtils.getPhoneHash(phoneNumber, salt)
```
codyborn: same here

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -58,12 +54,25 @@ type AttestationServiceRunningCheckResult =
   | { isValid: true; result: ActionableAttestation }
   | { isValid: false; issuer: Address }
 
+interface AttesationServiceRevealRequest {
+  account: string
```
codyborn: Should this be type `Address`?

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -151,48 +165,45 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
   getAttestationState: (
-    phoneNumber: string,
+    identifier: string,
     account: Address,
     issuer: Address
   ) => Promise<AttestationStateForIssuer> = proxyCall(
     this.contract.methods.getAttestationState,
-    tupleParser(PhoneNumberUtils.getPhoneHash, stringIdentity, stringIdentity),
-    (state) => ({ attestationState: parseInt(state[0], 10) })
+    undefined,
```
codyborn, line 184: Why are we passing in `undefined` as input? I'm not sure I follow.

jmrossy: because this param (the input pre-parser) is not needed but the next one (output pre-parser) is needed

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: Good idea!

packages/attestation-service/src/requestHandlers/attestation.ts
jmrossy: Thanks again!

i1skn (approved):
Excellent work! From wallet perspective 🚢 it!

PR #3570, opened 2020-04-23
[phone-number-privacy] add ratelimiting to matchmaking, update salt/bls references

aslawson:
##  Description

- add ratelimiting to matchmaking -- can only matchmake one time
- add error message for monitoring
- refactored to separate salt logic (which includes rate limiting) and bls signature logic
- change to return 403 for invalid requests, includes: matchmaking more than once, exceeded salt query quota
- updates BLS signature testing to include signature verification

## Other changes

- Changed db table wrappers to use class models

## Tested

- Added test to assure user gets an error on attempting to matchmake more than once
- manual testing

## Related issues

https://github.com/celo-org/celo-labs/issues/424

## Backwards compatibility

NA not in production

## Commit Message
[phone-number-privacy] add ratelimiting to matchmaking, update salt/bls references


packages/phone-number-privacy/src/common/error-utils.ts
```diff
@@ -8,6 +8,7 @@ export enum ErrorMessages {
   EXCEEDED_QUOTA = 'CELO_PNP_ERR_05 Requester exceeded salt service query quota',
   SIGNATURE_COMPUTATION_FAILURE = 'CELO_PNP_ERR_06 Failed to compute BLS signature',
   UNKNOWN_ERROR = 'CELO_PNP_ERR_07 Something went wrong',
```
jmrossy: Nit: Could we make this error number 0?

packages/phone-number-privacy/src/database/wrappers/account.ts
```diff
@@ -19,7 +19,23 @@ export async function getPerformedQueryCount(account: string): Promise<number> {
...
- * Increments query count in database.
+ * Returns whether account has already performed matchmaking
+ */
+export async function getDidMatchmaking(account: string): Promise<boolean> {
+  try {
+    const didMatchmaking = await getDatabase()(ACCOUNTS_TABLE)
+      .where(ACCOUNTS_COLUMNS.address, account)
+      .select(ACCOUNTS_COLUMNS.didMatchmaking)
+      .first()
+    return didMatchmaking !== undefined && didMatchmaking[ACCOUNTS_COLUMNS.didMatchmaking] !== null
```
jmrossy: Nit: Unless you need to check for `undefined` or `null` *specifically*, the more common thing in JS would be just:
`didMatchmaking && didMatchmaking[ACCOUNTS_COLUMNS.didMatchmaking]` which will coalesce types for you

packages/phone-number-privacy/src/database/wrappers/account.ts
```diff
@@ -37,11 +53,36 @@ async function incrementExistingQueryCount(account: string) {
...
+  const data = new Account(account)
+  data[ACCOUNTS_COLUMNS.numLookups] = 1
+  return updateRecord(data)
+}
+
+/*
+ * Set did matchmaking to true in database.  If record doesn't exist, create one.
+ */
+export async function setDidMatchmaking(account: string) {
+  await ((await setExistingDidMatchmaking(account)) || (await addNewRecordWithMatchmaking(account)))
```
jmrossy: what's the intention here? Is it to try setting it and then add record if it fails?

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
```diff
@@ -1,26 +1,30 @@
...
     const matchedContacts: ContactMatch[] = (
       await getNumberPairContacts(request.body.userPhoneNumber, request.body.contactPhoneNumbers)
     ).map((numberPair) => ({ phoneNumber: numberPair }))
     await setNumberPairContacts(request.body.userPhoneNumber, request.body.contactPhoneNumbers)
-    // TODO (amyslawson) return salts with contact
+    await setDidMatchmaking(request.body.account)
     response.json({ success: true, matchedContacts })
   } catch (e) {
-    console.error('Failed to getContactMatches', e)
+    console.error(ErrorMessages.UNKNOWN_ERROR + ' Failed to getContactMatches', e)
```
jmrossy: With console, you can just provide multiple arguments. No need to `+` them together yourself

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
```diff
@@ -30,7 +34,15 @@ interface ContactMatch {
...
-  return hasValidUserPhoneNumberParam(requestBody) && hasValidContractPhoneNumbersParam(requestBody)
+  return (
+    hasValidAccountParam(requestBody) &&
+    hasValidUserPhoneNumberParam(requestBody) &&
+    hasValidContractPhoneNumbersParam(requestBody)
+  )
+}
+
+function hasValidAccountParam(requestBody: any): boolean {
+  return requestBody.account && (requestBody.account as string).startsWith('0x')
```
jmrossy: Consider using the utilities in https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/address.ts for stuff like this

jmrossy (commented):
looks good, just some small nits

packages/phone-number-privacy/src/database/wrappers/account.ts
aslawson: correct -- from my google searching, I could not find another way to do update-if-exists + insert-if-does-not-exist.

if you have thoughts on a better way, let me know!

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
aslawson: k

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
aslawson: there's no check for leading 0x here -- just 'ensureLeading0x' which modified the input to add a leading 0x if there is none

packages/phone-number-privacy/src/database/wrappers/account.ts
jmrossy: I would recommend just querying first rather than relying on update to return falsy values to get this or to work. 
Something like:
`getDatabase().select('address').from<Account>(ACCOUNTS_TABLE).where(ACCOUNTS_COLUMNS.address, account)`
then if that returns something, update it, otherwise create it. 

Semi-related, just as a point of style using multiple awaits on one line is possible but not great for readability and debugability :)

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
jmrossy: I was thinking you'd use `isValidAddress` instead of just checking that the account param starts with `0x`

jmrossy (approved):

packages/phone-number-privacy/src/database/wrappers/account.ts
aslawson: sure --
i guess the benefit of the current implementation is that if it updates, then it only ever makes one db request.  if it queries first -- then there are always 2 request.

but by querying first, it is more clear what is happening, and that other update errors wouldn't cause subsequent insert request

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
aslawson: oh i see -- its exported from ethereumjs-util

kk i'll update in next pr (as the mergify has already pushed this one -- should have waited for the shipit before adding the label...)

aslawson: oh you added the label :P

PR #3565, opened 2020-04-23
[BlockchainApi] Fix exchange transactions

jeanregisser:
## Description

This PR restores functionality for the wallet activity feed regarding exchange transactions.

The existing API we used broke because there are now events emitted when paying for fees in cUSD. See discussion in https://celo-org.slack.com/archives/CMX4JFMQW/p1586526953169900

A new API has been added to blockscout in order to get all cUSD and cGLD transfers happening in transactions.
This work has been done by @mrsmkl in https://github.com/celo-org/blockscout/pull/102 which has already been deployed on alfajores.

For transactions paying for fees in cUSD we try to detect transfers related to fees (validator, gateway fee recipient, community fund). 
This has risks of breaking, so I've put assertions so we can detect if it happens in the future.

## Other changes

- Removed obsolete feed events query (had been deprecated)

## Tested

Deployed on alfajores and tested successfully.

## Related issues

- Fixes #3395
- Fixed #2509

## Backwards compatibility

Yes

PR #3583, opened 2020-04-24
[BlockchainApi] Fix transfers processing when gateway fee recipient is null

jeanregisser:
## Description

Follow up on #3565, fix transfers processing when gateway fee recipient is null.

## Tested

Tested that transactions with a null gateway fee recipient are processed correctly.

## Backwards compatibility

Yes.

### DIFF
--- a/packages/blockchain-api/src/blockscout.ts
+++ b/packages/blockchain-api/src/blockscout.ts
@@ -15,6 +15,7 @@ export interface BlockscoutTransferTx {
   gasUsed: string
   feeToken: string
   gatewayFee: string
+  gatewayFeeRecipient: string
   input: string
   celoTransfers: BlockscoutCeloTransfer[]
 }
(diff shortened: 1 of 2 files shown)


packages/blockchain-api/src/blockscout.ts
```diff
@@ -44,33 +45,31 @@ export class BlockscoutAPI extends RESTDataSource {
...
-    const feeTransfer2 = celoTransfers[transfersCount - 2]
-    const feeTransfer3 = celoTransfers[transfersCount - 3]
-    const fee1Value = new BigNumber(feeTransfer1.value)
-    const fee2Value = new BigNumber(feeTransfer2.value)
-    const fee3Value = new BigNumber(feeTransfer3.value)
-    const gasValue = new BigNumber(transferTx.gasUsed).multipliedBy(transferTx.gasPrice)
-    const gatewayFeeValue = new BigNumber(transferTx.gatewayFee)
-    const expectedTotalFeeValue = gasValue.plus(gatewayFeeValue)
+    let totalFee = new BigNumber(0)
+    for (let i = 1; i <= expectedFeeTransfersCount; i++) {
```
i1skn, line 56: Should we use `i <= transfersCount;` here?
Otherwise, in case `transfersCount > expectedFeeTransfersCount` we would not count all the fees, right?

jeanregisser: I think the logic is correct.
The fee transfers are always the last 2 or 3 transfers.

Let's say it's the case with 3 fee transfers, we have the following transfers for the transaction we're looking at:
```
[
  transfer1,
  transfer2,
  feeTransfer1,
  feeTransfer2
  feeTransfer3
]
```
We want to add only `feeTransfer1`, `feeTransfer2`, `feeTransfer3`, so we iterate from 1 to 3 taking transfers from the end of the array.

Does this make it clearer for you?

i1skn: Yeah, thanks!

i1skn (approved):
🚢 it!

PR #3586, opened 2020-04-24
Update mergify to use PR title and body as commit message

jmrossy:
## Description

Change mergify config and PR template to use PR body and title in automerge commits.


jeanregisser (approved):
Looking good 👍

PR #3587, opened 2020-04-24
[cli] Verify Proof of Possession

joeabbey:
## Description

A common use-case has emerged in which a client submits a proof-of-possession signature for authorization of signers to a custodian.  In that case, prior to performing on-chain operations it is desirable to be able to easily validate that the signature is correct.

Moreover this PR serializes the parsed signature to the form that the contract expects.

*Signtaure is VRS*
```
./bin/run  account:verify-proof-of-possession --account 0x199eDF79ABCa29A2Fa4014882d3C13dC191A5B58 --signer 0x0EdeDF7B1287f07db348997663EeEb283D70aBE7 --signature 0x1c5efaa1f7ca6484d49ccce76217e2fba0552c0b23462cff7ba646473bc2717ffc4ce45be89bd5be9b5d23305e87fc2896808467c4081d9524a84c01b89ec91ca3
valid: true
signature: 0x1c5efaa1f7ca6484d49ccce76217e2fba0552c0b23462cff7ba646473bc2717ffc4ce45be89bd5be9b5d23305e87fc2896808467c4081d9524a84c01b89ec91ca3
```
*Signature is RSV*
```
./bin/run  account:verify-proof-of-possession --account 0x199eDF79ABCa29A2Fa4014882d3C13dC191A5B58 --signer 0x0EdeDF7B1287f07db348997663EeEb283D70aBE7 --signature 0x5efaa1f7ca6484d49ccce76217e2fba0552c0b23462cff7ba646473bc2717ffc4ce45be89bd5be9b5d23305e87fc2896808467c4081d9524a84c01b89ec91ca301
valid: true
signature: 0x1c5efaa1f7ca6484d49ccce76217e2fba0552c0b23462cff7ba646473bc2717ffc4ce45be89bd5be9b5d23305e87fc2896808467c4081d9524a84c01b89ec91ca3
```
> Note: the signature has been serialized to VRS

### DIFF
--- a/packages/cli/src/utils/command.ts
+++ b/packages/cli/src/utils/command.ts
@@ -2,6 +2,7 @@ import { ensureLeading0x, trimLeading0x } from '@celo/utils/lib/address'
 import { BLS_POP_SIZE, BLS_PUBLIC_KEY_SIZE } from '@celo/utils/lib/bls'
 import { URL_REGEX } from '@celo/utils/lib/io'
 import { isE164NumberStrict } from '@celo/utils/lib/phoneNumbers'
+import { POP_SIZE } from '@celo/utils/lib/signatureUtils'
 import { flags } from '@oclif/command'
 import { CLIError } from '@oclif/errors'
 import { IArg, ParseFn } from '@oclif/parser/lib/args'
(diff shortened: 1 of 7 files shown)


asaj (approved):

packages/cli/src/utils/command.ts
```diff
@@ -111,6 +115,11 @@ export const Flags = {
     description: 'Phone Number in E164 Format',
     helpValue: '+14152223333',
   }),
+  proofOfPossession: flags.build({
```
asaj, line 118: Love this! Can we use this flag type in other commands where relevant?

joeabbey-anchor: Absolutely, its specifically for a 65 byte signature as you can see it even validates that it is 65-bytes (`POP_SIZE`)

```
const parseProofOfPossession: ParseFn<string> = (input) => {
  return parseBytes(input, POP_SIZE, `${input} is not a proof-of-possession`)
}
```

asaj (approved):

Issue #3410, opened 2020-04-14
Clean up signing libraries


codyborn:
### Expected Behavior

Remove usage of `eth-lib` (should use `ethereumjs-util` instead).
Consolidate definition of Signature object (v,r,s) that performs type validation. 

### Current Behavior

Currently we have a variety of different signing libraries/utilities (signatureUtils & signing-utils) that perform similar operations.

oneeman:
@barbaraliau I happened upon that library and saw this in its README:

> Lightweight Ethereum libraries. This is a temporary repository which will be used as the basis of an implementation on Idris (or similar).
>
> Edit: please, don't use this lib. I worked on it for a few days and stopped to create Formality. Eventually I'll make Ethereum libraries in Formality and compile them to JavaScript. Right now, if you need Ethereum stuff on the browser, consider alternatives (web3, ethers, ethereumjs, etc.). Also, for account signing and creation, consider http://github.com/maiavictor/nano-ethereum-signer

Should this issue be prioritized given that the library's author recommends it not be used?

PR #3591, opened 2020-04-24
Test signatures that need canonicalization

codyborn:
## Description

There was a bug introduced where signatures failed to be RLP decoded because they were not canonical. In this case, signatures with s/r values that were prefixed by "0x00" should have the first byte removed to fit the canonicalization standard that geth expects. I used our existing celo test transaction and cycled through nonces until I hit the "0x00" prefix and added this nonce+tx as a test case.

I searched for nodejs libraries that perform full-signature validation and unfortunately couldn't find one, hence the explicit check for "0x00" in the test.

## Other changes

Account.recover() doesn't support signature canonicalization so I pad the r/s values as needed during recovery for testing.


packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -171,7 +171,13 @@ export function recoverTransaction(rawTx: string): [Tx, string] {
     data: rawValues[8],
     chainId,
   }
-  const signature = Account.encodeSignature(rawValues.slice(9, 12))
+  let r = rawValues[10]
+  let s = rawValues[11]
+  // Account.recover cannot handle canonicalized signatures
+  // A canonicalized signature may have the first byte removed if its value is 0
+  r = ensureLeading0x(trimLeading0x(r).padStart(64, '0'))
```
gastonponti, line 178: I think you could also use `r = Bytes.pad(32, r)`
this is the code:
```
const pad = (l,hex) =>
  hex.length === l*2+2 ? hex : pad(l,"0x"+"0"+hex.slice(2));
```
It's less optimal, but it's a library that we are already using (both work for me, haha, your choice)

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
```diff
@@ -298,6 +298,18 @@ describe('AzureHSMWallet class', () => {
                 normalizeAddressWith0x(knownAddress)
               )
             })
+
+            // https://github.com/ethereum/go-ethereum/blob/38aab0aa831594f31d02c9f02bfacc0bef48405d/rlp/decode.go#L664
+            test('signature with 0x00 prefix is canonicalized', async () => {
+              // nonce 65 with the mock private keys will produce an S value with the first byte as 0x00
```
gastonponti: Maybe you could extract, and coy here the transaction. I know it will duplicated code, but if in the future, if someone change the example, because it needs to add something, could lead it to a nightmare to replicate the double 0, haha.
In this particular example, I prefer to know that the Tx belongs only to this tests, for it's specific requirements

packages/contractkit/src/wallets/ledger-wallet.test.ts
```diff
@@ -335,6 +335,22 @@ describe('LedgerWallet class', () => {
               },
               TEST_TIMEOUT_IN_MS
             )
+
+            // https://github.com/ethereum/go-ethereum/blob/38aab0aa831594f31d02c9f02bfacc0bef48405d/rlp/decode.go#L664
+            test(
+              'signature with 0x00 prefix is canonicalized',
+              async () => {
+                // nonce 65 with the mock private keys will produce an S value with the first byte as 0x00
+                celoTransaction.nonce = 65
```
gastonponti: Same

packages/contractkit/src/wallets/local-wallet.test.ts
```diff
@@ -158,6 +160,18 @@ describe('Local wallet class', () => {
               normalizeAddressWith0x(knownAddress)
             )
           })
+
+          // https://github.com/ethereum/go-ethereum/blob/38aab0aa831594f31d02c9f02bfacc0bef48405d/rlp/decode.go#L664
+          test('signature with 0x00 prefix is canonicalized', async () => {
+            // nonce 65 with the mock private keys will produce an S value with the first byte as 0x00
+            celoTransaction.nonce = 65
```
gastonponti: same

gastonponti (approved):
Just details. 👍

packages/contractkit/src/utils/signing-utils.ts
codyborn: I'm trying to reduce the usage of the eth-lib wherever possible due to the message you pointed out:
> please, don't use this lib
https://github.com/MaiaVictor/eth-lib

packages/contractkit/src/wallets/azure-hsm-wallet.test.ts
codyborn: 👍

packages/contractkit/src/utils/signing-utils.ts
codyborn: https://github.com/celo-org/celo-monorepo/issues/3410

PR #3594, opened 2020-04-24
Add --estimate to cli rewards show

jfoutts-celo:
## Description

Add --estimate flag to cli rewards:show to avoid using archival state for results older than 1024 blocks.

## Other changes

Corrects usages of archival state in contractkit, and fixes voter rewards calculation.

## Tested

e2e tests.

## Backwards compatibility

Yes


packages/contractkit/src/wrappers/Election.ts
```diff
@@ -475,9 +472,13 @@ export class ElectionWrapper extends BaseWrapper<Election> {
...
    */
-  async getVoterRewards(address: Address, epochNumber: number): Promise<VoterReward[]> {
+  async getVoterRewards(
+    address: Address,
+    epochNumber: number,
+    archive?: boolean
+  ): Promise<VoterReward[]> {
     const blockNumber = await this.kit.getLastBlockNumberForEpoch(epochNumber)
-    const voter = await this.getVoter(address, blockNumber)
+    const voter = await this.getVoter(address, archive ? blockNumber : undefined)
```
nategraf: This implementation is misleading because if the user passes in a past epoch and `archive=false` it will give them incorrect information. I don't think we should add the archive option. Instead the epochNumber option should be optional and if specified will attempt, but possibly fail, to retrieve past data.

jfoutts-celo: Thanks. Very good points. I've updated to call `estimate` and default false. Without an archive node (or employing approximation) there's only 85 minute (1024 * 5 seconds) window to retrieve exact voter rewards information.

packages/contractkit/src/wrappers/Election.ts
```diff
@@ -487,9 +484,13 @@ export class ElectionWrapper extends BaseWrapper<Election> {
    * @param address The address to retrieve VoterRewards for.
    * @param epochNumber The epoch to retrieve VoterRewards at.
    */
-  async getVoterRewards(address: Address, epochNumber: number): Promise<VoterReward[]> {
+  async getVoterRewards(
+    address: Address,
+    epochNumber: number,
+    estimate?: boolean
+  ): Promise<VoterReward[]> {
```
nategraf: I find this function signature confusion because it will mislead the reader when `estimate` is true. If it is set to true, the `epochNumber` parameter is ignored in calculation, but included in the result, which makes the result effectively inconsistent. (i.e. although is says it's with regard to epoch `epochNumber`, its actually calculated from current data. Instead I would strongly suggest having the indication to use current epoch data be to set `epochNumber` to `undefined`.

packages/cli/src/commands/rewards/show.ts
```diff
@@ -24,6 +24,7 @@ export default class Show extends BaseCommand {
 
   static flags = {
     ...BaseCommand.flagsWithoutLocalAddresses(),
+    estimate: flags.boolean({ description: 'Estimate voter rewards from current state' }),
```
nategraf: Because non-archive is the default configuration for a node, the defaults for this command should not rely on it. Instead of adding this estimate flag to opt-out of archival state use, the option should be opt-in. My suggestion would be to overload the `epochs` flag by making it default to `0` or `undefined`, in which case only current state will be used, and if the user specifies a number there, then use archival state. It should also be explicit in the command or flag description that the target node must have the desired state in it's archives.

nategraf (commented):
A couple of comments on interfaces. Sorry for letting this PR languish. Feel free to ping me when you address these so I get to approving the final state faster.

packages/contractkit/src/wrappers/Election.ts
jfoutts-celo: Added missing param comment.  When `estimate` is true, the voter's share of the (actual) group voter rewards for the supplied `epochNumber` is estimated based on current votes.  This can be misleading, but only when you opt-in by supplying `estimate` = true.  Having `epochNumber` = undefined imply the latest epoch with estimation works, but I would also like to be very explicit (i.e. dedicated parameter) that the result is an estimation.

packages/cli/src/commands/rewards/show.ts
jfoutts-celo: Added some explanatory error handling.  The defaults work for 85 minutes after each epoch.  After that an archive node is needed to determine exact voter rewards (archival state issue only applies when --voter supplied).  If you haven't changed your votes, then your current share of votes for a group should be a good estimation of your share 90 minutes or 23 hours ago.  But we can't always show exact information for the latest epoch. I think an `estimate` flag makes that clear.

packages/contractkit/src/wrappers/Election.ts
nategraf: I see your point about making it quite explicit that the returned value is an estimate. My biggest qualm here is that is that the `epochNumber` is included in the return value when using an estimate, which is misleading. I also find it misleading that the `epochNumber` argument must be supplied, but is ignored. My best suggestion here might be to split the function signature such that there is an "estimate" method that does not take the epoch reward parameter and this function which does, but does not have the estimate parameter.

packages/cli/src/commands/rewards/show.ts
nategraf: I like the error handling! 👍

packages/contractkit/src/wrappers/Election.ts
jfoutts-celo: Removed the notion of estimation from contractkit by factoring out getVoterShare() from getVoterRewards().

nategraf (approved):
Thanks for working with all my feedback!

PR #3595, opened 2020-04-25
Add account:lock and duration flag to account:unlock

nategraf:
## Description

Currently there is no way to relock an account from the CLI. This PR adds an `account:lock` command as well as adding a `--duration` flag to `account:unlock` to unlock an account temporarily.


mrsmkl (approved):
Tested locally, LGTM.

PR #3482, opened 2020-04-18
[web] Upgrade next.js

aaronmgdr:
## Description

upgrades next.js. most benefitially this means we have better bundling/ bundle reporting. 

Screen Shot 2020-04-18 at 1 46 29 PM

## Other changes

removed experimental stuff as its now done stable way

## Tested

visited most pages in dev and after deploying (need to do more)

## Backwards compatibility
should be

## Commit Message

Upgrade Next.js for Website to latest Version

PR #3596, opened 2020-04-25
remove SASS Hamburger

aaronmgdr:
## Description

Replaces  sass-based hamburger with a RNS implementation so we can remove sass as a dependency

This is necessary (well this or something else) to upgrade to the next version of next.js

## Tested

By hand and Eye
hamberger

## Related issues
#3482 

## Backwards compatibility
YES


cmcewen (approved):
woot!

PR #3598, opened 2020-04-25
Bump wallet to 1.7.1

i1skn:
## Description

Bump wallet to 1.7.1


packages/mobile/circle_deploy.sh
```diff
@@ -4,6 +4,7 @@ set -euo pipefail
 . ../../scripts/check_if_changed.sh
 
 CHANGED=$(checkIfChangedFolder)
+BUILD=""
```
jmrossy, line 7: We don't use this script any more, no need for changes here

jmrossy (approved):

PR #3603, opened 2020-04-26
[Phone Number Privacy] Restructure query quota and improve logging

jmrossy:
## Description

- Restructure query quota to simplify it
- Improve query quota tests by mocking contract kit instead of local functions
- Add a logging wrapper (logger) that includes a prefix id for each request
- Add more logging throughout

## Tested

Locally and in staging

## Backwards compatibility

Yes


packages/phone-number-privacy/test/salt-generation/query-quota.test.ts
```diff
@@ -1,62 +1,62 @@
...
-})
-
-describe('get remaining query count', () => {
-  beforeEach(() => {
-    jest.clearAllMocks()
-  })
-  it('calculates remaining query count', async () => {
+describe(getRemainingQueryCount, () => {
+  it('Calculates remaining query count for verified account', async () => {
+    const contractKitVerifiedNoTx = createMockContractKit(
```
jmrossy, line 19: By mocking contractKit instead of spying on and mocking the internal methods of the former Class, we can exercise the code flows more thoroughly and accurately.

packages/phone-number-privacy/src/common/logger.ts
```diff
@@ -0,0 +1,14 @@
+// tslint:disable: no-console
+import { v4 as uuidv4 } from 'uuid'
+
+// This allows us to differentiate parallel instantiations of this function
+const prefix = uuidv4().slice(0, 8)
```
aslawson, line 5: oh interesting

packages/phone-number-privacy/src/database/wrappers/account.ts
```diff
@@ -10,42 +11,47 @@ function accounts() {
...
+    logger.error(ErrorMessages.DATABASE_GET_FAILURE, e)
     return 0
   }
 }
 
+async function getAccountExists(account: string): Promise<boolean> {
+  const existingAccountRecord = await accounts()
+    .where(ACCOUNTS_COLUMNS.address, account)
+    .first()
+  return !!existingAccountRecord
```
aslawson, line 31: so this double negation is a pattern intended to return true if the record exists, correct?

packages/phone-number-privacy/src/database/wrappers/account.ts
```diff
@@ -10,42 +11,47 @@ function accounts() {
...
-      .where(ACCOUNTS_COLUMNS.address, account)
-      .increment(ACCOUNTS_COLUMNS.numLookups, 1)
+    if (await getAccountExists(account)) {
+      return accounts()
+        .where(ACCOUNTS_COLUMNS.address, account)
+        .increment(ACCOUNTS_COLUMNS.numLookups, 1)
+    } else {
+      const newAccount = new Account(account)
+      newAccount[ACCOUNTS_COLUMNS.numLookups] = 1
+      return updateRecord(newAccount)
```
aslawson: seems like this method should be renamed `insertRecord`

probably my bad here, but lets update

packages/phone-number-privacy/test/salt-generation/query-quota.test.ts
aslawson: wow -- i'm happy to see you figured how to successfully mock the contractkit.

aslawson (approved):
lgtm

special thanks for cleaning up the query quota testing 🙏

packages/phone-number-privacy/src/database/wrappers/account.ts
jmrossy: Yeah, it's a JS way of coalescing a type to boolean

packages/phone-number-privacy/src/database/wrappers/account.ts
jmrossy: Will do

packages/phone-number-privacy/src/common/logger.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const logger = {
```
nambrot, line 7: Sorry for the drive-by, but have you considered using a library like bunyan? I personally find structured logs very helpful, as it allows operators to extract relevant information without much gymanistics. Here is an example for the attestation-service https://github.com/celo-org/celo-monorepo/blob/master/packages/attestation-service/src/logger.ts#L1

packages/phone-number-privacy/src/database/wrappers/account.ts
```diff
@@ -10,42 +11,47 @@ function accounts() {
...
-}
-
-async function incrementExistingQueryCount(account: string) {
+export async function incrementQueryCount(account: string) {
+  logger.debug('Incrementing query count')
   try {
-    return await accounts()
-      .where(ACCOUNTS_COLUMNS.address, account)
-      .increment(ACCOUNTS_COLUMNS.numLookups, 1)
+    if (await getAccountExists(account)) {
```
nambrot, line 40: What do you think about using upsert here?

packages/phone-number-privacy/src/common/logger.ts
jmrossy: Thanks for the suggestions! Have to disagree with you though.
1. I'm not in favor of adding a lib to replace 10 lines of JS that is doing the job, especially in a security sensitive project
2. Why learn a new abstraction layer for logging when console is working!

packages/phone-number-privacy/src/database/wrappers/account.ts
jmrossy: Agreed, we'll do that in a follow-up

packages/phone-number-privacy/src/common/logger.ts
nambrot: 1. very fair
2. Operators of this service might appreciate structured logging/easy integration with logging providers

PR #3604, opened 2020-04-26
Add retry unreliable tests

codyborn:
## Description

Add retries to two unreliable test suites. These tests frequently fail unrelated to the current changes adding a lot of delay to the check-in process.

### DIFF
--- a/packages/celotool/ci_test_governance.sh
+++ b/packages/celotool/ci_test_governance.sh
@@ -12,6 +12,7 @@ export TS_NODE_FILES=true
 if [ "${1}" == "checkout" ]; then
     # Test master by default.
     BRANCH_TO_TEST=${2:-"master"}
+    echo "TEST"
     echo "Checking out geth at branch ${BRANCH_TO_TEST}..."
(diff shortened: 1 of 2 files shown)


packages/celotool/ci_test_governance.sh
```diff
@@ -13,7 +13,7 @@ if [ "${1}" == "checkout" ]; then
     # Test master by default.
     BRANCH_TO_TEST=${2:-"master"}
     echo "Checking out geth at branch ${BRANCH_TO_TEST}..."
-    ../../node_modules/.bin/mocha -r ts-node/register src/e2e-tests/governance_tests.ts --branch ${BRANCH_TO_TEST}
+    ../../node_modules/.bin/mocha --retries 3 -r ts-node/register src/e2e-tests/governance_tests.ts --branch ${BRANCH_TO_TEST}
```
gastonponti, line 16: I would reduce it to 2, or try how hard is to bail 2 and make a dry run if both failed

gastonponti (approved):

packages/celotool/ci_test_governance.sh
codyborn: Thanks Gaston, still looking into whether this will work. Mocha retry will not re-run the `before` step, implying we'll be starting tests with the current state which I don't believe can be expected to pass.

PR #3609, opened 2020-04-27
Updates token list for RC1

kobigurk:
## Description

Updates the Ledger token list for RC1 so cGLD and cUSD transfers will display properly on a Ledger.

## Tested

Tested on celo-blockchain with the same token list and on a custom version @gastonponti provided that doesn't check balances on RC1.


kobigurk:
@gastonponti is there a way to test it on RC1?

gastonponti:
> @gastonponti is there a way to test it on RC1?

You could use it against rc1-forno
`celocli transfer:dollars --from XXX --to YYY --value ZZZ  --node https://rc1-forno.celo-testnet.org --useLedger`
You won't make a success transaction because you don't have any funds, but at least you could see that everything is displayed as it should.
You HAVE to change the `const CELO_APP_ACCEPTS_CONTRACT_DATA_FROM_VERSION = '1.0.2'` variable in `contractkit/src/wallets/signers/ledger-signer.ts` to 1.0.1 if your ledger is not on that version.

If you want to verify all the cycle with a valid transaction, we should probably can't faucet anything, so we should ask a validator, maybe?

gastonponti (approved):

kobigurk:
> > @gastonponti is there a way to test it on RC1?
> 
> You could use it against rc1-forno
> `celocli transfer:dollars --from XXX --to YYY --value ZZZ --node https://rc1-forno.celo-testnet.org --useLedger`
> You won't make a success transaction because you don't have any funds, but at least you could see that everything is displayed as it should.
> You HAVE to change the `const CELO_APP_ACCEPTS_CONTRACT_DATA_FROM_VERSION = '1.0.2'` variable in `contractkit/src/wallets/signers/ledger-signer.ts` to 1.0.1 if your ledger is not on that version.
> 
> If you want to verify all the cycle with a valid transaction, we should probably can't faucet anything, so we should ask a validator, maybe?

@gastonponti provided a custom version for testing, and it works!

PR #3611, opened 2020-04-27
Use the correct account in releasegold:locked-gold

asaj:
## Description

`withdraw` was not being sent from the correct address.

## Other changes

None
## Tested

No
## Related issues

None
## Backwards compatibility

Yes


lucasege (approved):
lgtm

PR #3612, opened 2020-04-27
Include invite code in SMS when user reminds invitee of escrowed funds

### LINKED ISSUE #3436: When reminding an invited user from notification, the text message is blank
### Expected Behavior

The text message should contain the original invite code along with the reminder message(I believe one used to exist -- can provide copy if not!)

### Current Behavior
When reminding an invited user from notification, the text message is blank

cc @coreycelo

tarikbellamine:
### Description

The goal was to send a link to the app + an invite code in the body of a SMS message when a user was prompted to remind their invitee that they have funds in escrow. Given we were not storing invite codes, this necessitated a refactor of our app state to store the relevant information.

### Other changes

I made addressToE164Number the source of truth for address look ups (previously we were using invite.invitee data in some cases). I did this by storing the invitee address in addressToE164Number whenever an invite is sent out.

### Tested

Built tests to for the "remind" button for both iOS and Android. Added tests for regular invite SMS to test for Android and iOS separately.

### Related issues

- Closes #3436 

### Backwards compatibility

NO: given the structure of app state was changed, app data needs to be wiped.


packages/mobile/src/invite/actions.ts
```diff
@@ -16,8 +16,17 @@ export enum Actions {
   SKIP_INVITE_FAILURE = 'INVITE/SKIP_INVITE_FAILURE',
 }
 
-export interface Invitees {
-  [tempAddress: string]: string // tempAddress -> e164Number
+export interface InviteDetails {
```
tarikbellamine, line 19: @jmrossy Another option to eliminate redundancy between EscrowedPayment and InviteDetails is to eliminate tx related data (timestamp, escrowAmmount, escrowTxId, etc) from InviteDetails and just do a join on the two structures when we need to. What do you think?

packages/mobile/src/invite/saga.ts
```diff
@@ -105,14 +110,16 @@ export async function generateInviteLink(inviteCode: string) {
   return shortUrl
 }
 
-async function sendSms(toPhone: string, msg: string) {
+export async function sendSms(toPhone: string, msg: string) {
   return new Promise((resolve, reject) => {
     try {
       if (Platform.OS === 'android') {
+        // TODO: look into how to configure react-native-sms for Android so we can cut this package
```
tarikbellamine: I tried to do this but got an ambiguous error. Someone who's more comfortable with android and the config settings should be able to do this pretty quickly. Who would be  a good person to take a look?

packages/mobile/locales/es-419/walletFlow5.json
```diff
@@ -56,6 +56,8 @@
   "escrowedPaymentReminder": "Reacuerda al receptor aceptar el pago",
   "escrowedPaymentReminderSms":
     "¡Un recordatorio amistoso de que aún no ha canjeado sus dólares de celo!",
+  "escrowedPaymentReminderSmsNoData":
+    "¡Un recordatorio amistoso de que aún no ha canjeado sus dólares de celo! ¡Descargue la aplicación móvil para comenzar!",
```
tarikbellamine: This was done on Google translate. Is there a process to double check translations?

cmcewen: We have the #spanish channel 😄

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
```diff
@@ -13,33 +13,56 @@ import { EscrowedPayment } from 'src/escrow/actions'
...
 
 export class EscrowedPaymentListItem extends React.PureComponent<Props> {
-  onRemind = () => {
-    const { payment, t } = this.props
+  onRemind = async () => {
+    const { payment, t, invitees } = this.props
     const recipientPhoneNumber = payment.recipientPhone
     CeloAnalytics.track(CustomEventNames.clicked_escrowed_payment_send_message)
-    // TODO: open up whatsapp/text message slider with pre populated message
+    // TODO(Tarik): add a UI that allows user to choose between SMS and Whatsapp (currently only SMS) for reminder message
```
cmcewen: I think this is probably worth a github issue since it's more of a feature level thing than a comment, which I usually think of as like code janitorial tasks

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
```diff
@@ -13,33 +13,56 @@ import { EscrowedPayment } from 'src/escrow/actions'
...
     try {
-      if (Platform.OS === 'android') {
-        SendIntentAndroid.sendSms(recipientPhoneNumber, t('walletFlow5:escrowedPaymentReminderSms'))
+      const inviteDetails = invitees.find(
+        (inviteeObj) => recipientPhoneNumber === inviteeObj.e164Number
+      )
+      // OPEN QUESTION: Appears to me we won't have inivtee data when local data is wiped. is this an appropriate compromise in that case?
```
cmcewen: When would local data be wiped? To me that implies the whole app storage is gone so I don't think we would have to worry about it

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
```diff
@@ -13,33 +13,56 @@ import { EscrowedPayment } from 'src/escrow/actions'
...
+      if (!inviteDetails) {
+        message = t('walletFlow5:escrowedPaymentReminderSmsNoData')
       } else {
-        // TODO look into using MFMessageComposeViewController to prefill the body for iOS
-        navigateToURI(`sms:${recipientPhoneNumber}`)
+        const { inviteCode } = inviteDetails
+        // OPEN QUESTION: this function creates a unique link every time a reminder is sent. is this desirable?
+        const link = await generateInviteLink(inviteCode)
+        message = t('walletFlow5:escrowedPaymentReminderSms', {
+          // OPEN QUESTION: why do we have to send inviteCode and link if link is derived from inviteCode?
```
cmcewen: Believe this is so someone can copy and paste the text message if the link doesn't work

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
```diff
@@ -13,33 +13,56 @@ import { EscrowedPayment } from 'src/escrow/actions'
...
+      )
+      // OPEN QUESTION: Appears to me we won't have inivtee data when local data is wiped. is this an appropriate compromise in that case?
+      let message
+      if (!inviteDetails) {
+        message = t('walletFlow5:escrowedPaymentReminderSmsNoData')
       } else {
-        // TODO look into using MFMessageComposeViewController to prefill the body for iOS
-        navigateToURI(`sms:${recipientPhoneNumber}`)
+        const { inviteCode } = inviteDetails
+        // OPEN QUESTION: this function creates a unique link every time a reminder is sent. is this desirable?
```
cmcewen: If it's easy to save the link, might be worthwhile but no strong preference

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
```diff
@@ -125,5 +148,7 @@ const styles = StyleSheet.create({
 })
 
 export default componentWithAnalytics(
-  withTranslation(Namespaces.inviteFlow11)(EscrowedPaymentListItem)
+  connect<StateProps, {}, {}, RootState>(mapStateToProps)(
```
cmcewen: This isn't a hard rule, but for things that might appear in a list it is often nice to only connect to redux at the "list" level and then pass down relevant state/props. So in this case we probably won't have that many, but just an fyi

packages/mobile/src/escrow/saga.ts
```diff
@@ -268,16 +267,20 @@ function* doFetchSentPayments() {
...
-      const recipientPhoneNumber = tempAddresstoRecipientPhoneNumber[id]
+      const address = sentPaymentIDs[i]
+      const recipientPhoneNumber = addressToE164Number[address]
+      const id = sentPaymentIDs[i].toLowerCase() // OPEN QUESTION: why are we making this lowercase? seems inconsistent
       const payment = sentPaymentsRaw[i]
       if (!payment) {
         continue
       }
 
+      // OPEN QUESTION: EscrowedPayment and InviteDetails appear to store redundant data but invitee data is stored on send and escrow data is fetched from the smart contract.
```
cmcewen: I don't think we can fetch the phone number from the contract, right?

packages/mobile/src/invite/saga.ts
cmcewen: What error did you get? I thought maybe there was some issue here with not using the user's default messaging app, but I'm not 100% sure

packages/mobile/src/invite/saga.ts
```diff
@@ -192,6 +197,30 @@ export function* sendInvite(
       Logger.error(TAG, 'Currently only dollar escrow payments are allowed')
     }
 
+    // OPEN QUESTION - does storing invitee data only in the event of a successful transaction have
```
cmcewen: I think there is potentially an issue where the transaction might be broadcast to the network and added to a block but you lose network connectivity or something so the error handling fails in web3

cmcewen (commented):
so there is a way to write migrations of redux state - @jmrossy has the most experience here and can maybe point you in the right direction. 

in terms of tests, i'm not sure if this is what you mean but i think having some test that the reminder shows up when provided the right data in redux would be good

packages/mobile/src/invite/saga.ts
tarikbellamine: I didn't screen grab the exact error but I believe it had to do with the `SendSMSPackage.getInstance()` that you need to set in `getPackages` getting overwritten by some internal process. Couldn't find good documentation for solving and wasn't comfortable enough tinkering so I abandoned fairly quickly. Worth recreating the error message and creating an issue?

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
tarikbellamine: Makes sense. I'll create one.

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
tarikbellamine: Makes sense. Wasn't sure if there was potentially another way local data could be wiped.

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
tarikbellamine: Got it, thanks.

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
tarikbellamine: Yeah pretty easy to do and is slightly preferable to creating a new one every time. Only thing that gave me pause was if there was some sort of security reason I wasn't aware of.

packages/mobile/src/escrow/saga.ts
tarikbellamine: Another option to eliminate redundancy between EscrowedPayment and InviteDetails is to eliminate tx related data (timestamp, escrowAmount, escrowTxId, etc) from InviteDetails and just do a join on the two structures when we need to. What do you think? Is there ever a time when InviteDetails would have tx data we couldn't get from the smart contract?

packages/mobile/src/invite/saga.ts
tarikbellamine: That's true. Doesn't storing the invitee data before we know the transaction was successful open us up to some unintended consequences (e.g., prompting the user to remind their invitee to redeem their escrowed payment with info that isn't valid)? As I see it, downside of failing to store invitee data (e.g., no reminders) is probably more palatable? Also from a frequency perspective, I'd assume it's more likely that one of the two blockchain transactions we are making fails than a connectivity issue arises in between a successful transaction and writing to the store.

If we agree that we don't need to store tx data in the InviteeDetails, I actually think the ideal location to store invite data is in between the invite transaction and and the escrow transaction.

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
tarikbellamine: I think I follow conceptually but not sure how to execute. Is there an example in the code you could point me to?

packages/mobile/locales/es-419/walletFlow5.json
tarikbellamine: ¡fantástico!

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
jmrossy: +1 for avoiding connecting list items (components that get rendered as items in a flat list).
The solution would be to connect this into the parent component and pass in the data as a prop

packages/mobile/src/escrow/saga.ts
```diff
@@ -268,16 +267,20 @@ function* doFetchSentPayments() {
...
-    const tempAddresstoRecipientPhoneNumber: Invitees = yield select(inviteesSelector)
+
+    const addressToE164Number = yield select(addressToE164NumberSelector)
     const sentPayments: EscrowedPayment[] = []
     for (let i = 0; i < sentPaymentsRaw.length; i++) {
-      const id = sentPaymentIDs[i].toLowerCase()
-      const recipientPhoneNumber = tempAddresstoRecipientPhoneNumber[id]
+      const address = sentPaymentIDs[i]
+      const recipientPhoneNumber = addressToE164Number[address]
+      const id = sentPaymentIDs[i].toLowerCase() // OPEN QUESTION: why are we making this lowercase? seems inconsistent
```
jmrossy: We lower case it because data coming in/out of contracts + web3 seem to have inconsistent casing. 
Unrelated, why do you assign address to this just above and then another var for id?

packages/mobile/src/invite/actions.ts
jmrossy: Yep that works! Up to you, it's not a significant amount of data so should be fine either way

packages/mobile/src/invite/saga.ts
```diff
@@ -192,6 +197,30 @@ export function* sendInvite(
...
+    // unintended consequences? Seems right to me but previous position was before the generateStandbyTransactionId call
+
+    // OPEN QUESTION: EscrowedPayment and InviteDetails appear to store redundant data but invitee data is stored on send and escrow data is fetched from the smart contract.
+    // Should we consolidate into one data structure that we update when appropriate?
+    const inviteDetails: InviteDetails = {
+      timestamp: '1587621489078', // PLACEHOLDER - we probably want to read from the actual tx data for this
+      e164Number,
+      tempWalletAddress: temporaryAddress,
+      tempWalletPrivateKey: temporaryWalletAccount.privateKey, // what could we need this for?
```
jmrossy: For reclaiming the $ if the invite doesn't get used.

packages/mobile/src/invite/saga.ts
```diff
@@ -192,6 +197,30 @@ export function* sendInvite(
       Logger.error(TAG, 'Currently only dollar escrow payments are allowed')
     }
 
+    // OPEN QUESTION - does storing invitee data only in the event of a successful transaction have
+    // unintended consequences? Seems right to me but previous position was before the generateStandbyTransactionId call
+
+    // OPEN QUESTION: EscrowedPayment and InviteDetails appear to store redundant data but invitee data is stored on send and escrow data is fetched from the smart contract.
+    // Should we consolidate into one data structure that we update when appropriate?
+    const inviteDetails: InviteDetails = {
+      timestamp: '1587621489078', // PLACEHOLDER - we probably want to read from the actual tx data for this
```
jmrossy: Date.now() should be fine, it would be used in the UI that shows users their history

packages/mobile/src/invite/saga.ts
```diff
@@ -192,6 +197,30 @@ export function* sendInvite(
...
+      e164Number,
+      tempWalletAddress: temporaryAddress,
+      tempWalletPrivateKey: temporaryWalletAccount.privateKey, // what could we need this for?
+      tempWalletRedeemed: false, // no logic in place to toggle this yet
+      inviteCode,
+      escrowAmount: amount ? amount.toString() : undefined, // this is not being read anywhere yet
+      escrowCurrency: currency, // this is not being read anywhere yet
+      escrowTxId, // the tx id we are saving is wrong (paymentId in sentEscrowedPayments is wrong too)
```
jmrossy: yeah you'd need to get it from web3 after the tx has been hashed. But we may need need this field anyway.

packages/mobile/test/values.ts
```diff
@@ -45,7 +45,60 @@ export const mockCountryCode = '+1'
 
 export const mockQrCodeData = `{"address":"${mockAccount}","e164PhoneNumber":"${mockE164Number}","displayName":"${mockName}"}`
 
-export const mockInvitableRecipient: RecipientWithContact = {
+const mockNameInvite = 'Marcus Zuckerberg'
```
jmrossy: Maybe lets use faker names here :)

jmrossy (commented):
Mostly looks good! You're on the right track :)

packages/mobile/src/invite/actions.ts
tarikbellamine: Great, I am going that route. All things equal, I'd rather the data sets be MECE.

packages/mobile/src/invite/saga.ts
tarikbellamine: Makes sense

packages/mobile/src/invite/saga.ts
tarikbellamine: yup, decided not to include escrow data in InviteDetail

packages/mobile/test/values.ts
tarikbellamine: haha fair point

packages/mobile/src/escrow/saga.ts
tarikbellamine: Should I be storing recipientPhoneNumber as lowercase in addressToE164Number?

That was my bad...I inserted those two lines above id without changing anything below it. I condensed it.

jmrossy: > Should I be storing recipientPhoneNumber as lowercase in addressToE164Number?

Phone numbers don't have casing and you shouldn't be modifying addressToE164Number

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
tarikbellamine: Done

packages/mobile/src/escrow/saga.ts
tarikbellamine: My bad, I misreferenced what I was talking about. I am calling `updateE164PhoneNumberAddresses` to store the mapping of the invitee phone number to their temporary wallet address and I am not making it lowercase when I do that.

https://github.com/celo-org/celo-monorepo/blob/tarikbellamine/invite-reminder-sms/packages/mobile/src/invite/saga.ts#L215

packages/mobile/src/escrow/EscrowedPaymentListItem.test.tsx
```diff
@@ -1,21 +1,59 @@
...
+        <EscrowedPaymentListItem
+          payment={mockEscrowedPayment}
+          invitees={[mockInviteDetails, mockInviteDetails2]}
+        />
       </Provider>
     )
     expect(tree).toMatchSnapshot()
   })
+
+  it('sends an SMS reminder on Andorid as expected', async () => {
```
cmcewen: Nice! Small typo 😄

packages/mobile/src/home/__snapshots__/NotificationBox.test.tsx.snap
```diff
@@ -400,6 +400,7 @@ exports[`NotificationBox renders correctly for with all notifications 1`] = `
                       onResponderTerminate={[Function]}
                       onResponderTerminationRequest={[Function]}
                       onStartShouldSetResponder={[Function]}
+                      testID="undefined/backupKeyFlow6:getBackupKey/Button"
```
cmcewen, line 401: Not a huge deal but these guys are all `undefined`

packages/mobile/src/invite/saga.ts
cmcewen: I think we can still handle this by storing before we try to send, and then if we are confident the transaction failed we can remove the stored data. I think we want to avoid the case where you send some money and it's not recoverable, but if we at least store the data that means we could get it back at some point. You could guard against this by checking that the account actually has money and it hasn't been redeemed before showing the reminder.

cmcewen (commented):
Overall looks good to me. 

I'm still concerned about losing the private key to the new account after a successful transaction has gone through, so I think showing a reminder that doesn't exist is still a "less bad" failure case than funds disappearing in the ether. If you want to merge this and handle that ghost reminder case in another PR that's fine too

cmcewen:
Also check your lint/build 😄

packages/mobile/src/home/__snapshots__/NotificationBox.test.tsx.snap
tarikbellamine: Yeah so I needed to add a testID to notification buttons in order to properly test the "remind" button and to do that I needed to pass in the testID from the parent component as a prop to ensure uniqueness. Components elsewhere that aren't passing in a testID prop will show up with an "undefined" in the string (instead of not having a testID at all).

I could go back and pass through props from all parent components but that seemed out of scope.

packages/mobile/src/invite/saga.ts
tarikbellamine: That's a good point I hadn't considered. I'll move `storeInvite` function back to where it was, before any transactions are made. I'll also open up issue to deal with removal in the event of a transaction failure.

packages/mobile/package.json
```diff
@@ -112,7 +112,7 @@
     "react-native-splash-screen": "^3.2.0",
     "react-native-svg": "^9.11.1",
     "react-native-swiper": "^1.5.14",
-    "react-native-tcp": "https://github.com/celo-org/react-native-tcp#68567d7",
+    "react-native-tcp": "https://github.com/celo-org/react-native-tcp#96cdfeb",
```
jmrossy: Did you mean to update this? Seems very tangential to this PR.

packages/mobile/src/escrow/saga.ts
```diff
@@ -268,18 +267,19 @@ function* doFetchSentPayments() {
...
       sentPaymentIDs.map((paymentID) => call(getEscrowedPayment, escrow, paymentID))
     )
-    const tempAddresstoRecipientPhoneNumber: Invitees = yield select(inviteesSelector)
+
+    const addressToE164Number = yield select(addressToE164NumberSelector)
     const sentPayments: EscrowedPayment[] = []
     for (let i = 0; i < sentPaymentsRaw.length; i++) {
-      const id = sentPaymentIDs[i].toLowerCase()
-      const recipientPhoneNumber = tempAddresstoRecipientPhoneNumber[id]
+      const address = sentPaymentIDs[i]
```
jmrossy: Should you be `toLowerCase`ing here?

packages/mobile/src/invite/saga.ts
```diff
@@ -192,6 +210,9 @@ export function* sendInvite(
       Logger.error(TAG, 'Currently only dollar escrow payments are allowed')
     }
 
+    const e164NumberToAddress = { [e164Number]: null } // null means the number is unverified
```
jmrossy: This would replace any existing mappings a user may have for this number, which we don't want. Instead, do this:
`yield put(updateE164PhoneNumberAddresses({}, addressToE164Number))`

packages/mobile/src/escrow/saga.ts
```diff
@@ -268,18 +267,19 @@ function* doFetchSentPayments() {
...
+      const recipientPhoneNumber = addressToE164Number[address]
       const payment = sentPaymentsRaw[i]
       if (!payment) {
         continue
       }
 
       const escrowPaymentWithRecipient: EscrowedPayment = {
-        paymentID: id,
+        paymentID: address.toLowerCase(),
```
jmrossy: instead of here.

jmrossy (commented):
Almost ready! Just a few small comments left

packages/mobile/src/invite/saga.ts
tarikbellamine: I see. Is this true because we don't have any logic to stop someone from inviting an existing user? Or is there a use case I am missing where a user might want to invite someone with a number already in the mapping?

packages/mobile/src/escrow/saga.ts
tarikbellamine: This was an open question for me. Should I be making `address` lowercase before attempting to look it up in `addressToE164Number`? And generally speaking, is the move to always convert to lowercase when dealing with addresses?

packages/mobile/package.json
tarikbellamine: Yeah...I was getting a weird CI build error until I changed this. Do you think it could be a problem?

jmrossy: Looks like @jeanregisser has update this anyway since you created this PR (see merge conflicts).
Recommend reverting

packages/mobile/src/escrow/saga.ts
jmrossy: Yeah that's recommended.

packages/mobile/src/invite/saga.ts
jmrossy: Someone could invite an existing user via the Invite screen from the Account/Settings menu. And more broadly, there's no reason for us wipe away any existing mappings that may exist for this number.

tarikbellamine: Got it

packages/mobile/src/escrow/saga.ts
tarikbellamine: Ok makes sense. In that case, I also needed to do it here: https://github.com/celo-org/celo-monorepo/blob/tarikbellamine/invite-reminder-sms/packages/mobile/src/invite/saga.ts#L211 and here: https://github.com/celo-org/celo-monorepo/blob/tarikbellamine/invite-reminder-sms/packages/mobile/src/invite/saga.ts#L171

jmrossy (commented):
Code changes look good but we almost forgot: you need a redux migration to make this change backwards compatible.
See src/redux/migrations.ts and here:
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/redux/store.ts#L11

jmrossy (dismissed):

jmrossy (approved):

PR #3613, opened 2020-04-27
Changes order of operations in the ledger and considers fee currency

kobigurk:
### Description

* The Ledger must do the token information injection as the last operation before sending the transaction, as other operations override the relevant values. Currently the token injection is done before `getValidatedDerivationPath`, where if `ledgerConfirmAddress` is used will override some details and cause the token data not to be displayed. The PR changes the order of operations.
* If feeCurrency is defined, injects its token info as well.

### Tested

Tested with the Ledger app version 1.0.3.

### Backwards compatibility

Backwards compatible.


gastonponti (approved):
Nice catch!

packages/contractkit/src/wallets/signers/ledger-signer.ts
```diff
@@ -169,6 +170,15 @@ export class LedgerSigner implements Signer {
       if (tokenInfo) {
         await this.ledger!.provideERC20TokenInformation(tokenInfo)
       }
+      if (rlpEncoded.transaction.feeCurrency) {
```
gastonponti: this is a sanitised value, so you should check against `0x`
replace it for this line:
`if (rlpEncoded.transaction.feeCurrency && rlpEncoded.transaction.feeCurrency !== '0x') {`

kobigurk: Oh, thanks!

PR #3617, opened 2020-04-28
[phone-number-privacy] min balance for unverified salt callers, require verification for matchmake

aslawson:
### Description

- add min balance 10 cents for unverified callers for salt query to receive quota of 2
- add requirement for phone number verification prior to match making request
- make caller phone number hash param for salt optional (for pre-attestation case where caller needs their own salt)

### Other changes

NA

### Tested

Updated tests for these cases

### Related issues

- https://github.com/celo-org/celo-labs/issues/448

### Backwards compatibility

NA


packages/phone-number-privacy/src/common/identity.ts
```diff
@@ -8,3 +11,14 @@ export function authenticateUser() {
...
 }
+
+export async function isVerified(account: string, hashedPhoneNumber: string): Promise<boolean> {
+  // TODO (amyslawson) wrap forno request in retry
+  // TODO (aslawson) update to work with hashed phoneNumber
+  const attestationsWrapper: AttestationsWrapper = await getContractKit().contracts.getAttestations()
+  const attestationStats = await attestationsWrapper.getAttestationStat(hashedPhoneNumber, account)
+  const numAttestationsCompleted = attestationStats.completed
+  const numAttestationsRemaining =
+    config.attestations.numberAttestationsRequired - numAttestationsCompleted
```
codyborn, line 22: Broader question: should we also take into account success rate (in addition to min attestations). For example, if I control a validator node, I may be able to fake 1/100 attestations so I could reach the min attestation threshold but my success rate would be 1%.

packages/phone-number-privacy/src/config.ts
```diff
@@ -38,6 +40,7 @@ if (DEV_MODE) {
       unverifiedQueryMax: 2,
       additionalVerifiedQueryMax: 30,
       queryPerTransaction: 2,
+      minDollarBalance: new BigNumber(100000000000000000),
```
codyborn: Can we have a constant like cUSDToWei (whatever the smallest unit is called) and derive this value like:
`minDollarBalance: new BigNumber(.1 * cUSDToWei)`

This way it's self documenting and less error prone.

packages/phone-number-privacy/test/salt-generation/query-quota.test.ts
```diff
@@ -13,50 +16,95 @@ jest.mock('../../src/web3/contracts')
...
       },
       createMockWeb3(100)
     )
     mockPerformedQueryCount.mockImplementation(() => new Promise((resolve) => resolve(0)))
+    mockIsVerified.mockReturnValue(false)
     mockGetContractKit.mockImplementation(() => contractKitVerifiedNoTx)
     expect(await getRemainingQueryCount(mockAccount, mockPhoneNumber)).toEqual(2)
   })
+  it('Calculates remaining query count for unverified account without any balance', async () => {
+    const contractKitVerifiedNoTx = createMockContractKit(
```
codyborn, line 84: 👍

codyborn (approved):
Looks good! Left some small comments.

packages/phone-number-privacy/src/common/identity.ts
aslawson: technically yes -- the `getAttestationStat()` also returns `total` attestations.

my worry is what if there are user issues -- SMS get lost, the user gets confused and then resubmits for new attestation requests?  What would be the appropriate percentage to balance affect on user versus potential malicious request?

also, considering the number of require attestations is currently 3, you'd need to run 3 nodes that all get randomly selected to perform the attestation to hack this.

for reference, this is how the wallet confirms a phone number is verified currently.  though I'm open to the suggestion.

packages/phone-number-privacy/src/config.ts
aslawson: good idea -- i'm not sure what we call the smallest unit.  

in our migrationconfig we seem just to include all the zeros every time, but this idea seems better

packages/phone-number-privacy/src/common/identity.ts
jmrossy: @codyborn We definitely should take into account success rate and we intend to do the same in the wallet. The exact tuning of that isVerified is still TBD but we'll probably want to move that logic to ContractKit when we have it

packages/phone-number-privacy/src/config.ts
```diff
@@ -1,9 +1,11 @@
+import BigNumber from 'bignumber.js'
 import * as functions from 'firebase-functions'
 import logger from './common/logger'
 
 export const DEV_MODE =
   process.env.NODE_ENV !== 'production' || process.env.FUNCTIONS_EMULATOR === 'true'
 
+const CUSD_VALUE = 1000000000000000000
```
jmrossy: Rather than hardcoding this, consider using the `toWei` method in web3

jmrossy (approved):
Approved with an optional comment

packages/phone-number-privacy/src/common/identity.ts
aslawson: @jmrossy do you suggest we should add success rate now?  and what do we think would be a good starting point for the threshold?

packages/phone-number-privacy/src/config.ts
aslawson: cool thanks

packages/phone-number-privacy/src/common/identity.ts
codyborn: Makes sense. We can take a look at the success rates for non-malicious users and base it off of that.

PR #3618, opened 2020-04-28
Add broken link linter with remark

nategraf:
### Description

Currently there are a lot of broken links in the docs. This PR installs the `remark-cli` and the link validator plugin to add a `lint` rule to catch broken links. It also fixes all the broken links currently caught by the linter.

### Other changes

* Add additional page references to `developer-resources/contractkit`


yorhodes (approved):
This is so cool!

Could you add a check to CI?

nategraf:
It's already in CI now because `yarn lint` runs on every PR! 🙌

PR #3621, opened 2020-04-28
Add more info to signature error message

mrsmkl:
### Description

Validators keep seeing this message.
Adding the address against which the signature is checked to the error message.

Changed domain verification to check using all signer addresses like the others.


packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -70,37 +69,45 @@ type dnsResolverFunction = (
...
   address: string,
   dnsResolver: dnsResolverFunction = resolveTxt as any
 ) => {
   const found = await new Promise((resolve) => {
-    dnsResolver(claim.domain, (error, domainRecords) => {
+    dnsResolver(claim.domain, async (error, domainRecords) => {
       if (error) {
         console.debug(`Unable to fetch domain TXT records: ${error.toString()}`)
-        resolve(false)
+        resolve(`Unable to fetch domain TXT records: ${error.toString()}`)
```
jmrossy: We should be using `reject` here. Resolving truthy values that are meant to be a failures can lead to misinterpreted results in consumers.

packages/contractkit/src/identity/claims/verify.ts
```diff
@@ -70,37 +69,45 @@ type dnsResolverFunction = (
...
                 resolve(true)
               }
             }
-          })
-        })
+          }
+        }
       }
-      resolve(false)
+      resolve(`Unknown failure with address ${address}`)
```
jmrossy: ditto

jmrossy (commented):
Mostly looks good but promise handling needs a fix

packages/contractkit/src/identity/claims/verify.ts
mrsmkl: Agree, but a different pattern is used here for some reason (undefined means success and returned strings are error messages). Anyway cleaned it a bit up to make it easier to understand.

gastonponti (approved):
LGTM

nambrot (approved):

PR #3633, opened 2020-04-28
Keybase fixes

mrsmkl:
### Description

Looks like more fixes are needed to get keybase working.


nambrot (approved):

packages/cli/src/utils/identity.ts
```diff
@@ -34,6 +35,7 @@ export abstract class ClaimCommand extends BaseCommand {
     try {
       cli.action.start(`Read Metadata from ${filePath}`)
       const data = IdentityMetadataWrapper.fromFile(this.kit, filePath)
+      // Perhaps should check here that `--from` and data.data.meta.address match
```
nambrot: That's no longer the case right? From can be a signer?

mrsmkl: Yeah, so should check that it's the address or one of the signers.

packages/cli/src/utils/identity.ts
```diff
@@ -20,20 +21,37 @@ export abstract class ClaimCommand extends BaseCommand {
...
+  }
+
+  protected readMetadata = async () => {
+    const { args, flags } = this.parse(this.self)
     const filePath = args.file
     try {
       cli.action.start(`Read Metadata from ${filePath}`)
-      const data = IdentityMetadataWrapper.fromFile(this.kit, filePath)
+      const data = await IdentityMetadataWrapper.fromFile(this.kit, filePath)
+      if (flags.from) {
```
nambrot: This is required right, so you can just assume it is there?

packages/contractkit/src/identity/metadata.ts
```diff
@@ -139,11 +139,19 @@ export class IdentityMetadataWrapper {
...
         break
-      case ClaimTypes.DOMAIN:
+      case ClaimTypes.DOMAIN: {
         const existingClaims = this.data.claims.filter((el: any) => el.domain === claim.domain)
         if (existingClaims.length > 0) {
           return existingClaims[0]
         }
+        break
+      }
+      case ClaimTypes.KEYBASE: {
```
nambrot, line 154: One should be able to claim multiple keybase accounts, right?

mrsmkl: The idea here is to have only claim for each username, similarly to the domains above.

nambrot: Ah right my bad

nambrot (approved):

PR #3638, opened 2020-04-29
[celocli] Add ledger to the config command

gastonponti:
### Description

Add the possibility to configure every "base command" flag, for later use in each command.

### Tested

Manually verification

### Backwards compatibility

Yes


gastonponti:
@cla-bel @asaj We should probably check if we want to still show those flags in every command...

The same happens with the `--node` flag. Technically you could add it to every command, but we are not showing it because we can `set` it

packages/cli/src/commands/config/set.ts
```diff
@@ -1,28 +1,79 @@
...
+    return "Can't have two or more of these flags stored: '--useLedger', '--useAKV' or '--privateKey'."
+  }
+  if (
+    (celoConfig.ledgerAddresses ||
+      celoConfig.ledgerCustomAddresses ||
+      celoConfig.ledgerConfirmAddress) &&
+    !celoConfig.useLedger
+  ) {
+    console.log(
+      "You will need to use the '--useLedger' flag in your transactions, to use the Ledger flags that you set. You could also store the '--useLedger' flag to your configuration"
```
clrblmt, line 70: > to use the Ledger flags that you set.

 Unclear what this means

packages/cli/src/commands/config/set.ts
```diff
@@ -1,28 +1,79 @@
...
+    )
+  }
+  if (celoConfig.azureVaultName && !celoConfig.useAKV) {
+    console.log(
+      "You will need to use the '--useAKV' flag in your transactions, to use the Azure flags that you set. You could also store the '--useAKV' flag to your configuration"
```
clrblmt, line 75: > to use the Azure flags that you set

Same as above. Can guess what it may mean but not clear to me.

packages/cli/src/base.ts
```diff
@@ -116,30 +113,42 @@ export abstract class BaseCommand extends LocalCommand {
     }
 
     const res: ParserOutput<any, any> = this.parse()
-    if (res.flags && res.flags.privateKey && !res.flags.useLedger && !res.flags.useAKV) {
+    if (
+      res.flags &&
+      this.flagOrConfig(res, 'privateKey') &&
+      !this.flagOrConfig(res, 'useLedger') &&
```
codyborn, line 119: We should make these constants somewhere

packages/cli/src/commands/config/set.ts
```diff
@@ -1,28 +1,79 @@
...
+    }),
   }
 
-  static examples = ['set  --node ws://localhost:2500', 'set  --node <geth-location>/geth.ipc']
+  static examples = [
+    'set  --node ws://localhost:2500',
+    'set  --node <geth-location>/geth.ipc',
+    'set  --node https://somehost.com  --useLedger',
+    'set  --useLedger  --ledgerCustomAddress "[0, 47, 99]"  --add',
+    'set  --node ws://localhost:2500  --useLedger  --ledgerAddresses 3 --ledgerConfirmAddress',
```
codyborn, line 33: nit: some extra spaces between args

codyborn (approved):
Thanks for adding this @gastonponti!

packages/cli/src/commands/config/set.ts
gastonponti: This is a friendly warning because the user is setting the `--ledgerAddresses 5` for example, but is not adding the `--useLedger`.
I didn't required to add that flag, because there could be a use case where the user doesn't want to use the ledger in every transaction, so it could just add which are the address in the ledger that it will use, but then for every `celocli` command add or not the `--useLedger`
Example:
```
> celocli config:set --node <myNode> --ledgerCustomAddresses "[3,6]"
> celocli transfer:gold --privateKey <privateKeyForAnotherAddress> --from <privateKeyAddress> --to <someAddress> --value 1000
> celocli transfer:gold --from <ledgerAddressIndex3> --to <someAddress> --value 1000 --useLedger
```

If the user set the `useLedger` in the first command, this will tell that the user wants to use the EVERY TIME, and the second command will fail.

gastonponti: Writing this response I realise that I could add a very small change that could make everything easier, hahaha.
So, let me add something first...

aaronmgdr:
friendly reminder that this was approved but not merged

barbaraliau:
@gastonponti if this PR is still relevant, can you fix the merge conflicts? sorry there are like one million merge conflicts.

PR #3639, opened 2020-04-29
Docs: say that a full node can serve celocli

tromer:
In the "Running a Full Node" tutorial, say and show that the now-running full node can serve `celocli`.


timmoreton (approved):
👍

PR #3320, opened 2020-04-06
[Wallet] Upgrade to firebase v6

jeanregisser:
### Description

This PR upgrades `react-native-firebase` to v6.4.0.

I had to disable Hermes on Android. Problem was with support of ES6 `Proxy`.
We can re-enable later when Hermes supports it or wait to merge this PR until it's supported.

See https://github.com/facebook/hermes/issues/33

### Tested

App works as expected (notifications, rtdb read/write).
Not tested: full dynamic links flow with installation from the store (Play Store / App Store).

### Other changes

N/A

### Related issues

- Fixes #2571

### Backwards compatibility

Yes

PR #3650, opened 2020-04-29
[Wallet] Fix Firebase regressions on iOS

### LINKED ISSUE #3622: iOS: ‘Could not connect to firebase’ error is shown for fraction of seconds on ‘Invitation code’ page while processing with invitation code.
**Frequency:** 100%
**iOS App version:** iOS Testflight build : v1.7.1 (28)
**Repro on iOS device:** iPhone 11 Pro (13.3)

**Repro Steps:** 
1) Launch the app and Enter the phone number
2) Tap on ‘Continue’ button and Add security pin 
3) Navigate to ‘Invitation Code’ page and Enter invitation code 
4) Make sure process gets start and Observe while processing 

**Current Behavior:** ‘Could not connect to firebase’ error is shown for fracation of seconds

**Expected Behavior:** ‘Could not connect to firebase’ error should be not be shown 

**Investigation:** 
1) Issue does not reproduce in Android 
2) Randomly, Same error is shown when user navigating into the app 

**Attachment:** [iOS_Couldn’t connect to Firebase error is shown](https://drive.google.com/file/d/10A0fqnkMTde73j3IZpMjDgDs39NU11tk/view?usp=sharing) (Observe the issue between 0:32 to 0:37 secs)

### LINKED ISSUE #3623: iOS: App gets crashed by killing-relaunching the app after logging into the app.
**Frequency:** 80%
**iOS App version:** iOS Testflight build : v1.7.1 (28)
**Repro on iOS device:** iPhone 11 Pro (13.3)
**Testing Account:** 3155641283 /backup Key

**Repro Steps:** 
1) Launch the app and Login with test account
2) Kill - Relaunch the app 
3) Make sure green screen is shown and wait for some time 
4) Observe 

**Current Behavior:** App gets crashed by killing-relaunching the app

**Expected Behavior:** App should be working as an expected

**Investigation:** 
1. App continuously crashed by back-grounding-foregrounding the app once issue starts reproing 
2. ‘Opss..’ screen is shown while foregrounding the app randomly 

**Impact:** User should re-installed the app and re-login into the app
 
**Attachment:** [iOS_App get crashed by launching the app](https://drive.google.com/file/d/1BPe5Z_qTSjgKHt1tZW0pH9J8asr_8tOQ/view?usp=sharing) (Observe the issue between 0:47 to 2:00 min)

**Crashlog:** [iOS_App get crashed by launching the app.ips](https://drive.google.com/file/d/103dYObI8Y1Qea4Vb7LcLLaeJH8apKCmo/view?usp=sharing)

jeanregisser:
### Description

Follow up PR to #3320

This PR fixes crashes on iOS related to the database module.
I had already patched those for `react-native-firebase` v5 (see https://github.com/invertase/react-native-firebase/pull/2734).
Turns out a similar patch is still needed for v6. I will submit a PR upstream too for v6.

Also fixed broken push registration on iOS. There was a new call needed with v6.
I upgraded `react-native-firebase` to v6.7.1 since it changed push registration too, this way we're already on top again (but for how long? 😉).

These issues are reproducible only when using a real device.

### Tested

- App doesn't crash anymore on iOS in release mode.
- Push registration works

I could NOT test push being received successfully as `notification-service` currently has a similar issue as `blockchain-api` did a couple of days ago (fixed in #3583).

### Related issues

- Fixes #3622
- Fixes #3623

### Backwards compatibility

Yes.


packages/mobile/firebase.json
```diff
@@ -0,0 +1,5 @@
+{
+  "react-native": {
+    "messaging_ios_auto_register_for_remote_messages": false
```
jeanregisser, line 3: This is because we don't want the system popup with asking for push permissions right when the user runs the app for the first time.
Push permission is asked later on when we create the user account.

jmrossy (approved):
Looks like a tricky one, thanks for looking into it!

PR #3666, opened 2020-04-29
[Wallet] Updating e2e tests

### LINKED ISSUE #3059: Extend E2E testing to complete transactions
### Expected Behavior

### Current Behavior

What is the current behavior?

etuleu:
### Description

End to end tests. This starts from restoring a wallet and goes all the way to making a transaction. Other paths could be added as well in the future, such as verifying the phone number (or at least simulating it).

### Other changes

I also updated the config for circle ci to update the android target because it was out of date.

Added a new flag NET_DELAY to the emulator options, which was used in testing bandwidth, which is where I started and then got into end to end testing.
Also changed emulator options cold boot to off and fast boot to on, which is what the detox docs recommend. I am not entirely sure what the difference is but the emulator seems to work fine on my machine with these.

### Tested

`yarn test:e2e:android`

### Related issues

- Fixes #3059 

### Backwards compatibility


packages/mobile/scripts/run_e2e.sh
```diff
@@ -124,28 +128,28 @@ if [ $PLATFORM = "android" ]; then
...
   do
-    sleep 3 
+    sleep 3
   done
 
   runTest
 
   echo "Closing emulator (if active)"
-  adb devices | grep emulator | cut -f1 | while read line; do adb -s $line emu kill; done
+  #adb devices | grep emulator | cut -f1 | while read line; do adb -s $line emu kill; done
```
jmrossy: why comment this out?

jmrossy (commented):
Looks great! Just one question below

packages/mobile/scripts/run_e2e.sh
etuleu: oops! was not supposed to add this file. I am commenting this out while testing so the emulator stays open after something fails.

jmrossy (commented):
The changes LGTM so far, but waiting for the CI test to pass before approval :)

cmcewen (approved):

PR #3672, opened 2020-04-29
Move vscode launch files to monorepo root

codyborn:
### Description

Introducing launch.json for local VSCode unit test debugging. Adding this to monorepo so that a developer can run tests from one place. [See this document](https://jestjs.io/docs/en/troubleshooting#debugging-in-vs-code) for more details on VSCode and Jest integration.
image
In the debug tab, you can select which test suite you want to run. While running you can add breakpoints, inspect and modify variables.
image

If this pattern looks reasonable to folks we can begin adding it for all projects within monorepo.


jmrossy (approved):
Nice!

aslawson (approved):
super cool!  can't wait to test this out :)

is there a way to have this debug mode listening to local running application port for live debugging?  eg. if running the phone number privacy application locally, we could set breakpoints and inspect/change variables ect.

PR #3675, opened 2020-04-30
React Native 0.62 Upgrade

cmcewen:
### Description

Upgrading RN version and basically all deps


i1skn:
Great work on this one!!! One thing this PR brakes for me is hot-reloading on both platforms. I guess, it's related to Geth as it was before, until @jmrossy fixed it, but I'm not sure.

Also, I've created a PR (against this branch) to add Flipper support to iOS https://github.com/celo-org/celo-monorepo/pull/3697 (+ couple of minor cosmetic fixes, related to upgrade of `react-navigation`).

packages/mobile/android/app/src/main/java/org/celo/mobile/MainApplication.java
```diff
@@ -60,7 +59,7 @@ public ReactNativeHost getReactNativeHost() {
   public void onCreate() {
     super.onCreate();
     SoLoader.init(this, /* native exopackage */false);
-    initializeFlipper(this); // Remove this line if you don't want Flipper enabled
+    initializeFlipper(this, getReactNativeHost().getReactInstanceManager());
```
jmrossy, line 62: Should this be disabled somehow when app is built for release?

packages/mobile/package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "tslint -c tslint.json --project tsconfig.json",
     "build": "yarn run build:ts && yarn run build:metro",
     "build:ts": "tsc --noEmit",
-    "build:metro": "echo 'NOT WORKING RIGHT NOW'",
+    "build:metro": "yarn run react-native bundle --entry-file index.js --bundle-output bundle.js; rm bundle.js",
```
jmrossy, line 11: Nice! I'm wondering if we should remove the mobile-test-build-app job in CI if we have this working now. They're not exactly the same but there's a decent amount of overlap there. 
https://github.com/celo-org/celo-monorepo/blob/master/.circleci/config.yml#L846

packages/mobile/package.json
```diff
@@ -60,80 +61,78 @@
...
     "graphql": "^14.1.1",
-    "i18next": "^19.0.2",
-    "js-sha3": "^0.7.0",
+    "i18next": "^19.4.3",
+    "js-sha3": "^0.8.0",
     "lodash": "^4.17.14",
-    "lottie-ios": "3.1.3",
+    "lottie-ios": "^3.1.8",
     "lottie-react-native": "^3.3.2",
-    "moment": "^2.22.1",
```
jmrossy, line 74: 👍

jmrossy (approved):
Looks great!!! A few small comments below

packages/mobile/android/app/src/main/java/org/celo/mobile/MainApplication.java
cmcewen: This is handled below in the initializeFlipper method - if the class isn't found, the call fails

packages/mobile/package.json
cmcewen: Yeah good call

cmcewen:
@i1skn tested on both iOS and android devices and hot reloading works for me, so i'm gonna merge

i1skn:
> @i1skn tested on both iOS and android devices and hot reloading works for me, so i'm gonna merge

Good, that means that something is wrong with my setup! Gonna try to clean all the caches and stuff.

PR #3686, opened 2020-04-30
Add downtimeSlasher in NetworkConfig

zviadm:
### Description

Exposes downtimeslasher configuration in NetworkConfig

### Tested

yarn build (inside ./cli)
/bin/run network:parameters


zviadm:
I seem to have two issues:
1. `yarn test:prepare` doesn't exist anymore, i am guessing it is just `yarn test:reset` now? 
2. I can't generate docs using `yarn docs` because it will use my github repo url and will cause a massive changelog. So i have to change remote run docs and maybe change it back

nategraf (approved):
Looks good to me! 🙌 I'll see what I can do about those issues

PR #3428, opened 2020-04-15
Parse revert reason from failed contractkit txs

asaj:
## Description

This PR causes contractkit to fetch the revert reason from smart contract calls for which gas estimation fails. This prevents celocli from logging the following extremely generic error, which may mislead folks to think that not enough gas is being provided:

```
Error: Returned error: gas required exceeds allowance (20000000) or always failing transaction
```

## Other changes

None
## Tested

On Baklava/Alfajores by running the following commands with a keystore and ledger account, respectively:

```
Asas-MacBook-Pro:cli asaj$ ./bin/run transfer:dollars --from 0x456f41406b32c45d59e539e4bba3d7898c3584da --to 0x456f41406b32c45d59e539e4bba3d7898c3584da --value 99999999999999999999999999
Sending Transaction: transfer... !
Error: Gas estimation failed: transfer value exceeded balance of sender
    at ContractKit.<anonymous> (~/Documents/celo/celo-monorepo/packages/contractkit/src/kit.ts:312:15)
    at step (~/Documents/celo/celo-monorepo/packages/contractkit/lib/kit.js:44:23)
    at Object.throw (~/Documents/celo/celo-monorepo/packages/contractkit/lib/kit.js:25:53)
    at rejected (~/Documents/celo/celo-monorepo/packages/contractkit/lib/kit.js:17:65)


Asas-MacBook-Pro:cli asaj$ ./bin/run transfer:dollars --from 0x456f41406b32c45d59e539e4bba3d7898c3584da --to 0x456f41406b32c45d59e539e4bba3d7898c3584da --value 99999999999999999999999999 --useLedger
Retrieving derivation Paths [ 0, 1 ]
Sending Transaction: transfer... !
Error: Gas estimation failed: transfer value exceeded balance of sender
    at ContractKit.<anonymous> (~/Documents/celo/celo-monorepo/packages/contractkit/src/kit.ts:312:15)
    at step (~/Documents/celo/celo-monorepo/packages/contractkit/lib/kit.js:44:23)
    at Object.throw (~/Documents/celo/celo-monorepo/packages/contractkit/lib/kit.js:25:53)
    at rejected (~/Documents/celo/celo-monorepo/packages/contractkit/lib/kit.js:17:65)
```
## Related issues

Would have been helpful in debugging many other issues e.g. https://github.com/celo-org/celo-monorepo/issues/2301
 
## Backwards compatibility

Yes

PR #3688, opened 2020-04-30
Show revert reason in mobile

cmcewen:
### Description

Building off the work here: http://github.com/celo-org/celo-monorepo/pull/3428, we should now get the revert reason when gas estimation fails


jmrossy (approved):
Dope

PR #3689, opened 2020-04-30
Move libusb dependency to dev env

codyborn:
### Description

`@ledgerhq/hw-transport-node-hid` is only needed for testing (using ledger device) and shouldn't be a non-dev dependency. This is currently blocking services from using contractkit without installing libusb (which we can't do in Google Cloud Function).

### Tested

Built and ran tests locally.


dockerfiles/attestation-service/Dockerfile
```diff
@@ -1,9 +1,6 @@
 FROM circleci/node:10
```
gastonponti: The PR is using again the `circle/node:10`? Or it was not merged yet?

codyborn: Good catch, it just needed to be merged. Thanks

jmrossy (approved):
Ship it!

PR #3697, opened 2020-05-01
[Wallet] Flipper iOS support

### LINKED ISSUE #3912: Add Flipper support for iOS

i1skn:
### Description

This PR brings Flipper support for iOS. 
I've added a couple SWIFT related flags to build process and ideally I would like @jeanregisser to take a look on them to verify that it would not brake anything.

### Other changes

* Fix default background issue introduced with `react-navigation` upgrade https://github.com/react-navigation/react-navigation/issues/6819
* Remove grey border under the header https://github.com/react-navigation/react-navigation/issues/865
* `react-native-fast-crypto` were conflicting with `Flipper` and because we need it only on Android (I've checked with @cmcewen) I've added it to ignore list on iOS.
* Remoe `tsOS`, `tvOSTest` and `celoTest` targets (AFAIK we do not use/plan to use them)

### Tested

* iOS Debug/Release build on a real device
* Android build on a real device

### Related issues

- Fixes #3912

### Backwards compatibility

Yes


annakaz:
Can we set this up for Android too?

i1skn:
> Can we set this up for Android too?

I think this was already done by @cmcewen in #3675

packages/mobile/react-native.config.js
```diff
@@ -7,4 +7,11 @@ module.exports = {
       },
     },
   },
+  dependencies: {
+    'react-native-fast-crypto': {
+      platforms: {
```
jmrossy, line 12: What does this do?

jmrossy (commented):
Looks great! Just one question below

packages/mobile/react-native.config.js
i1skn: It disables `react-native-fast-crypto` for iOS

jmrossy: Why would we want to do that?

jmrossy: @cmcewen I disagree that we shouldn't be using rn-fast-crypto on ios. The lib seems to support ios. 
@i1skn if fixing it would be a lot of work, it's fine to do it in a separate PR but you should leave a TODO here and open an issue

i1skn: Well, `react-native-fast-crypto` and `flipper` relies on conflicting libraries. I have not checked how hard it would be to fix `react-native-fast-crypto`, but I can do it if having disabled for iOS would be a significant impact onto UX. Another option is to not have Flipper on iOS until we find a better solution. So, @cmcewen @jmrossy what do you think?

cmcewen: Sorry @i1skn I wasn't clear - I meant that android was the reason we added this library, not that we could turn it off on iOS. I believe bip39 still relies on it and I think it would potentially break some features. I think it should not be that hard to only include on copy of those .a files, no?

i1skn: Gotcha, I will investigate then :)

jmrossy:
@i1skn can we close this?

i1skn:
> @i1skn can we close this?

I will come back to it in the next milestone, will convert it to Draft for now.

barbaraliau:
Closing this issue as it is still in draft stage and stale. Please reopen if it is still relevant and being worked on!

PR #3700, opened 2020-05-01
[Wallet] Wait for rehydrate to fix DataSaver restart issue

### LINKED ISSUE #2985: [Wallet] Users SBAT turn off data saver on iOS
### Expected Behavior

Users can turn off data saver on iOS on a default geth build of the app

### Current Behavior

When users toggle data saver off, the app restarts but data saver is still on

### LINKED ISSUE #3147: Data saver mode doesn't work after restarting the app
### Expected Behavior

TXs can be sent after restarting the app

### Current Behavior

To reproduce:
- Turn on data saver mode
- Force quit the app
- Start again and try to send a transaction - hangs forever

Version: 1.6.2, build 104794504, Pixel 2 XL

### LINKED ISSUE #3155: “Transaction Failed” error message is shown when trying to make a payment with Data Saver toggled on.
**Frequency:** 100%
**App version:** Test flight v1.6.2(24)
**Repro on:** iPhone 7+( 13.3.1)
**Does not Repro on:** iPhone 5S(12.41.)
**Testing Account:** 6504908232/backup Key
**Pre-condition:** Make sure Data Saver toggle is ON.

**Repro Steps:**
1. Tap on  + icon from home page. . 
2. Initiate a payment. 
3. Enter amount and tap on Send button. . 
4. Wait for the confirmation of the payment. 
5. Observe 

**Current Behavior:** Transaction failed Error message is shown after 2 minutes of initiating the transaction.  
**Expected Behavior:** Payment should be done successfully. 
 
**Investigation:** Issue is not repro if we request for any payment with Data Saver toggle on. 
 
**Attachment:**[ iOS_TransactionFailed.mp4](https://drive.google.com/open?id=1jt3SmlMnGQN1AZ30bFT6UkwmCnmpN0_v) (Bug observe at 10 sec)

annakaz:
### Description

Currently, when the app is restarted, there is a moment before state is rehydrated when we don’t know whether app is in DataSaver mode. We assume the initial state FORNO_ENABLED_INITIALLY in the .env file, then toggle DataSaver on or off if needed. With the default of FORNO_ENABLED_INITIALLY=false, geth starts every time the app is restarted, even when DataSaver is on. This has made it easy for the following bugs to emerge: #3155, #3147, related to a user starting a default geth build of the app, turning on DataSaver, then restarting the app but seeing geth still sync as the geth saga starts before state is rehydrated and is not cancelled.

This PR waits until state is rehydrated to start any web3/contractkit related actions, so that the correct provider can be used based on DataSaver mode and to avoid unnecessarily starting the light client.

### Other changes

- Re enable DataSaver in Account screen
- Remove "Restart with false PIN warning" on DataSaver screen. Users can toggle on DataSaver with a false PIN, they just won't be able to send until correct PIN is entered
- Fixes Pincode navigation bug in ExchangeScreen when DataSaver enabled
- Clarified DataSaver text that users will have to restart app and return to DataSaver screen

### Tested

Built default geth APK and tested on pixel 2: restored wallet from backup key, got verified, bought gold, turned on DataSaver, sold gold, killed app, sold gold, turned on geth, turned off geth, turned on geth to prompt app restart, sold gold with geth on
Also tested invite

(Built APK as we've seen restart behavior can differ slightly in debug mode)

Will send Ankit both a default geth and a default forno APK

### Related issues

- Fixes #3147 
- Fixes #3155
- Should fix #2985 (iOS restart issue), though not yet tested on iOS


packages/mobile/src/account/__snapshots__/Account.test.tsx.snap
```diff
@@ -598,6 +598,83 @@ exports[`Account renders correctly 1`] = `
             </View>
           </View>
         </View>
+        <View
```
jmrossy, line 594: Why did this file change?

packages/mobile/src/web3/contracts.ts
```diff
@@ -2,25 +2,61 @@ import { newKitFromWeb3 } from '@celo/contractkit'
...
+import { contractKitReadySelector, fornoSelector } from 'src/web3/selectors'
 import Web3 from 'web3'
 import { provider } from 'web3-core'
 
-// Logging tag
 const tag = 'web3/contracts'
 
-const web3: Web3 = getWeb3()
-let contractKit = newKitFromWeb3(web3)
+export const web3ForUtils: Web3 = new Web3() // Web3 with no provider
```
jmrossy: Are there potential downsides to having two web3 instances created at the same time?
Instead of this, I'd recommend creating a `getWeb3` function that proxies to the kit's existing web3 instance

packages/mobile/src/web3/contracts.ts
```diff
@@ -2,25 +2,61 @@ import { newKitFromWeb3 } from '@celo/contractkit'
...
+  while (!contractKitReadySelector(store.getState())) {
+    Logger.debug(`getContractKitOutsideGenerator`, `ContractKit is still locked...`)
+    await sleep(250) // Every 0.25 seconds
+  }
+
+  Logger.debug(`${tag}@getContractKitOutsideGenerator`, 'ContractKit ready, returning')
+  return getContractKitBasedOnFornoInStore()
+}
+
+export function* getContractKit() {
```
jmrossy, line 51: Can you clarify why CK has to wait for that action? As you've already seen, we call this *a lot*, and if we can somehow avoid sagafying this it would be much cleaner.

packages/mobile/src/web3/contracts.ts
```diff
@@ -2,25 +2,61 @@ import { newKitFromWeb3 } from '@celo/contractkit'
...
+  Logger.info(
+    `${tag}@getWeb3`,
+    `Initializing web3, platform: ${Platform.OS}, forno mode: ${fornoMode}, provider: ${
+      fornoMode ? DEFAULT_FORNO_URL : 'geth'
+    }`
+  )
+  return fornoMode ? new Web3(getHttpProvider(DEFAULT_FORNO_URL)) : new Web3(getIpcProvider())
+}
+
+export async function getContractKitOutsideGenerator() {
```
jmrossy, line 35: Needing this feels very hacky, can we find a more elegant solution?

packages/mobile/src/account/__snapshots__/Account.test.tsx.snap
annakaz: DataSaver was behind a feature flag so didn't show up in Account before

packages/mobile/src/web3/contracts.ts
annakaz: CK has to wait for that action only if there is a lock on contractKit (contractKitReady = false), as it must wait for it to be unlocked. 
There is a lock on contractKit when:
1. the app was just opened and store is not yet persisted, so geth vs forno mode is not known and we must wait until it is known to start geth or forno
2. the app is in the middle of switching providers (turning on or off forno mode)

packages/mobile/src/web3/contracts.ts
annakaz: Agreed 😃  this would be removed as we finish sagafying contractKit logic. Added a comment explaining its temporary context- I had looked at doing the migration in this PR (moved everything straightforward over) but some will require a bigger refactor. 

The ability to block contractKit calls seems worthwhile beyond just switching providers for forno- for example if there are connectivity issues

cmcewen: Yeah Anna and I talked about this a bit, and I think similar to what we found with blocking till geth is ready, it's better to have that ability at the contractkit level too. Theoretically when switching we would want to wait for all "in flight" transactions to complete (waiting for a confirmation, for example) and then swap over

cmcewen (commented):
Overall looks pretty good! Did you test the relevant flows like invite/verification with both modes? And is there anywhere you can capture the rest of the work that should be done here?

packages/mobile/src/web3/contracts.ts
annakaz: Switched to avoid creating an additional web3 instance. 

There are still two instances of contractKit (which each have their own web3 instance) for each mode, but this was suggested by Mariano (https://celo-org.slack.com/archives/CL7BVQPHB/p1581115835017900) as a better way than switching providers

cmcewen (approved):
LGTM!

annakaz:
Opened issue https://github.com/celo-org/celo-monorepo/issues/3727 and added a TODO reminding to remove `getContractKitOutsideGenerator()` when it is resolved

PR #3712, opened 2020-05-04
Fixing total balance

### LINKED ISSUE #3705: Wrong total balance in celocli
command: celocli account:balance 0x81a9CC4103e2c66a652b2C17BaaF134FdaeAfCA9
### expected output:
All balances expressed in units of 10^-18.
gold: 999676394500000000
lockedGold: 0
usd: 0
total: 999676394500000000
pending: 0

### current output:
All balances expressed in units of 10^-18.
gold: 999676394500000000
lockedGold: 0
usd: 0
total: 3.963877391197344453575983046348115674221700746820753546332534027902565746944e+75
pending: 0

The error works with all accounts whose balance has not changed after the genesis block.

The bug is most likely in contactkit because it repeats in my js code as well.

mrsmkl:
### Description

When calling `getBuyTokenAmount` or `setSellTokenAmount` with 0 value, the result will be a division by zero. Returning 0 instead in contractkit.

### Tested

Only manual testing

### Related issues

- Fixes #3705

### DIFF
--- a/packages/contractkit/src/wrappers/Exchange.ts
+++ b/packages/contractkit/src/wrappers/Exchange.ts
@@ -65,14 +65,14 @@ export class ExchangeWrapper extends BaseWrapper<Exchange> {
    * @param sellGold `true` if gold is the sell token
    * @return The corresponding buyToken amount.
    */
-  getBuyTokenAmount: (
-    sellAmount: BigNumber.Value,
-    sellGold: boolean
-  ) => Promise<BigNumber> = proxyCall(
-    this.contract.methods.getBuyTokenAmount,
-    tupleParser(valueToString, identity),
-    valueToBigNumber
-  )
+  async getBuyTokenAmount(buyAmount: BigNumber.Value, sellGold: boolean): Promise<BigNumber> {
+    const buy = valueToString(buyAmount)
+    if (new BigNumber(buy).eq(new BigNumber(0))) {
+      return new BigNumber(0)
+    }
+    const res = await this.contract.methods.getBuyTokenAmount(buy, sellGold).call()
+    return valueToBigNumber(res)
+  }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/contractkit/src/wrappers/Exchange.ts
```diff
@@ -81,14 +81,14 @@ export class ExchangeWrapper extends BaseWrapper<Exchange> {
...
    */
-  getSellTokenAmount: (
-    buyAmount: BigNumber.Value,
-    sellGold: boolean
-  ) => Promise<BigNumber> = proxyCall(
-    this.contract.methods.getSellTokenAmount,
-    tupleParser(valueToString, identity),
-    valueToBigNumber
-  )
+  async getSellTokenAmount(buyAmount: BigNumber.Value, sellGold: boolean): Promise<BigNumber> {
```
gastonponti, line 84: the parameter `buyAmount` should be called `sellAmount` to keep the same name from the contract.

packages/contractkit/src/wrappers/Exchange.ts
```diff
@@ -81,14 +81,14 @@ export class ExchangeWrapper extends BaseWrapper<Exchange> {
...
+    const buy = valueToString(buyAmount)
+    if (new BigNumber(buy).eq(new BigNumber(0))) {
```
gastonponti: detail: `new BigNumber(buy).eq(0)` will also work and it's cleaner

gastonponti (commented):
Although I understand the fix, (and probably this is the fastest way to fix it), this should be fixed in the exchange contract. Otherwise every implementation (possible contractKits in other languages, for example) will have to add this previous check.

mrsmkl:
Ok, I'll add the fix to the contracts too. It will probably take some time before we update the contracts though.

gastonponti (approved):

PR #3715, opened 2020-05-04
Oracle prometheus metrics, HSM wallet init retries, update aad-pod-identity

tkporter:
### Description

* Exposes prometheus metrics on oracle pods that are scraped & sent to SD
* Sets appropriate env vars for Azure HSM Wallet init retries (see https://github.com/celo-org/celo-oracle/pull/41)
* Updates aad-pod-identity version
* Sets `syncRetryDuration` to be 1 minute-- this is how often the aad-pod-identity MIC will check all resources to see if any identity bindings are not as they should be. This defaults to 1 hour, but aad-pod-identity is not robust to pod label changes (which we rely on), so this is the easy fix to ensure that oracle pod identity inconsistencies are identified in at most 1 minute

### Other changes

n/a

### Tested

Ran it on AKS, works great

### Related issues

n/a

### Backwards compatibility

No, there were some config renames on the aad-pod-identity side


jcortejoso (approved):
Nice. Looking good

PR #3717, opened 2020-05-04
[Wallet] Disable manual setting of nonce in transactions

jmrossy:
### Description

 Disable manual setting of nonce in transactions. This was introduced when the retry/cancel/timeout wrapper was added but it may be causing failures during verification.

### Tested

Did onboarding flow including verification.

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/transactions/send.ts
+++ b/packages/mobile/src/transactions/send.ts
@@ -15,7 +15,6 @@ import Logger from 'src/utils/Logger'
 import { assertNever } from 'src/utils/typescript'
 import { getGasPrice } from 'src/web3/gas'
 import { fornoSelector } from 'src/web3/selectors'
-import { getLatestNonce } from 'src/web3/utils'
 import { TransactionObject } from 'web3-eth'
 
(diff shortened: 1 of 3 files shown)


packages/mobile/src/transactions/send.ts
```diff
@@ -25,7 +24,7 @@ const TAG = 'transactions/send'
 // causing failures when a tx times out (rare but can happen on slow devices)
 const TX_NUM_TRIES = 1 // Try txs up to this many times
 const TX_RETRY_DELAY = 2000 // 2s
-const TX_TIMEOUT = 40000 // 40s
+const TX_TIMEOUT = 45000 // 45s
```
annakaz, line 27: Why increase tx timeout when nonces are no longer manual?

annakaz (approved):
LGTM!

packages/mobile/src/transactions/send.ts
jmrossy: Just a precaution since Ankit also saw a case where the tx timed out

PR #3725, opened 2020-05-05
Update contractkit ver to 4.5

codyborn:
### Description

Increase ContractKit version to 4.5-dev. Will be publishing ver 4.4 which removes the dependency on libusb in the environment.

### DIFF
--- a/packages/attestation-service/package.json
+++ b/packages/attestation-service/package.json
@@ -27,7 +27,7 @@
     "lint": "tslint -c tslint.json --project ."
   },
   "dependencies": {
-    "@celo/contractkit": "0.4.4-dev",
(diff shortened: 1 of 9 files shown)


annakaz (approved):
🚢

gastonponti (approved):
Just a question for checking purpose. We are not using some new `utils` feature from the `-dev` version, right?

codyborn:
> Just a question for checking purpose. We are not using some new `utils` feature from the `-dev` version, right?

Correct, the last change to `@celo/utils` was a version bump:
https://github.com/celo-org/celo-monorepo/commit/744502bf1e03abab8faa7a611645109accfd2733

PR #3733, opened 2020-05-06
Small updates while connecting PGPNP to AKV

codyborn:
### Description

- Firebase config keys can only be lowercase. 
- Updating ContractKit version (new version doesn't rely upon libusb which couldn't be installed on Cloud Functions)
- Reference AKV util via `lib` path and not `src` path (only for local)


packages/phone-number-privacy/package.json
```diff
@@ -29,7 +29,8 @@
   },
   "devDependencies": {
     "@types/uuid": "^7.0.3",
-    "firebase-functions-test": "^0.2.0"
+    "firebase-functions-test": "^0.2.0",
+    "query-string": "^6.12.1"
```
jmrossy: Where do we use this?

jmrossy (commented):
LGTM, just not sure about the query-string dep

packages/phone-number-privacy/package.json
codyborn: Ah I added this because I was writing some e2e tests.. will exclude from this PR

jmrossy (approved):

PR #3735, opened 2020-05-06
[NotificationService] Fix notifications

### LINKED ISSUE #1339: [Android] Push notification is not received for user B when user A proceed for payment

jeanregisser:
### Description

The notification service was picking the wrong address for the transfer, because of the events which changed at the transaction layer, similar to what we had to fix in `blockchain-api` in #3565.

Each transaction contains the actual transfer and also fee transfers. The previous code was only considering the last transfer emitted for the transaction, which was a fee transfer.

Note: we could also filter out fee transfers like we did for `blockchain-api`, but there's no harm leaving them for now (they won't have any matching fcm token).

### Other changes

- Don't create a Sentry event when a notification is received while in background. There is nothing that needs to be done. (Example: https://sentry.io/organizations/celo/issues/1648609728/?project=1250733&query=is%3Aunresolved)

### Tested

- Deployed on Alfajores:
  - Push notifications are sent again to the correct user address when sending money.

### Related issues

- Fixes #1339

### Backwards compatibility

Yes.


jmrossy (approved):
Ship it!

Issue #3601, opened 2020-04-25
Attestation service should not store attestation info for expired attestations


nambrot:
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

Issue #3600, opened 2020-04-25
Attestation service should not store phone numbers in plaintext


nambrot:
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

PR #3753, opened 2020-05-07
[phone-number-privacy][attestation] update attestation service to store hashed identifiers and dynamically purge expired records

aslawson:
### Description

- attestation service now stores hashed (phoneNumber+salt) identifiers rather than phoneNumbers
- dynamically purges expired attestation records from the db

### Other changes

NA

### Tested

- [x]  Manually test DB update with dev DB
`curl -v "http://localhost:3000/attestations" -d '{"phoneNumber":"+14143335555", "account":"0x2a14b55240cc82ed0128f20cd01520a3dd499644", "issuer":"0xE6e53b5fc2e18F51781f14a3ce5E7FD468247a15"}'  -H 'Content-Type: application/json'`
<s> - [ ]  Update and add test for record purging</s>

### Related issues

https://github.com/celo-org/celo-monorepo/issues/3600
https://github.com/celo-org/celo-monorepo/issues/3601

### Backwards compatibility

No -- we will need to ask validators to restart attestation service.


packages/attestation-service/migrations/20200506234935-attestation-v2.js
```diff
@@ -0,0 +1,43 @@
...
+    const transaction = await queryInterface.sequelize.transaction()
+
+    try {
+      await queryInterface.removeColumn('Attestations', 'phoneNumber')
+
+      await queryInterface.addColumn('Attestations', 'identifier', {
+        type: Sequelize.STRING,
+      })
+
+      await queryInterface.removeIndex(
```
nambrot, line 7: Did removing the index after removing the column work? I would have thought that one needed to remove the index first

packages/attestation-service/migrations/20200506234935-attestation-v2.js
```diff
@@ -0,0 +1,43 @@
...
+        'Attestations',
+        ['account', 'identifier', 'issuer'],
+        { fields: ['account', 'identifier', 'issuer'], unique: true },
+        { transaction }
+      )
+
+      await queryInterface.addIndex(
+        'Attestations',
+        ['createdAt'],
+        { fields: ['createdAt'], unique: true },
```
nambrot: I'm not sure how common it is to add a uniqueness constraint on a timestamp. We could have 2 records at the same time, right?

packages/attestation-service/migrations/20200506234935-attestation-v2.js
```diff
@@ -0,0 +1,43 @@
...
+      )
+
+      await transaction.commit()
+    } catch (error) {
+      await transaction.rollback()
+      throw error
+    }
+  },
+  down: (queryInterface, Sequelize) => {
+    return queryInterface.dropTable('Attestations')
```
nambrot: In the `down` method, one would generally specify the inverse of `up` in case one would like to reverse the migration. So i.e. you would probably want to 
- remove the `identifier` column
- add the `phoneNumber` column
- change the appropriate indicies

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -77,17 +79,44 @@ async function ensureLockedRecord(
...
+
 class AttestationRequestHandler {
   logger: Logger
+  identifier: string
   sequelizeLogger: (_msg: string, sequelizeLog: any) => void
   constructor(public readonly attestationRequest: AttestationRequest, logger: Logger) {
     this.logger = logger.child({ attestationRequest })
     this.sequelizeLogger = (msg: string, sequelizeLogArgs: any) =>
       this.logger.debug({ sequelizeLogArgs, component: 'sequelize' }, msg)
+    this.identifier = PhoneNumberUtils.getPhoneHash(
```
nambrot, line 137: I forget if the salt is optional, but if it is, I assume `getPhoneHash` works appropriately?

nambrot (commented):
Looks good, just minor comments!

packages/attestation-service/src/requestHandlers/attestation.ts
jmrossy: Yes, salt is optional for this service and getPhoneHash works if its undefined

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -77,17 +79,44 @@ async function ensureLockedRecord(
...
+          [Op.lte]: moment()
+            .subtract(expiryTimeInSeconds, 'seconds')
+            .toDate(),
+        },
+      },
+      transaction,
+    })
+    await transaction.commit()
+  } catch (err) {
+    await transaction.rollback()
```
jmrossy, line 110: Some logging here would be good. We'll want alerting for this

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -77,17 +79,44 @@ async function ensureLockedRecord(
   return attestationRecord
 }
 
+async function purgeExpiredRecords() {
+  const attestations = await kit.contracts.getAttestations()
+  const expiryTimeInSeconds = (await attestations.attestationExpiryBlocks()) * 5
```
jmrossy: Let's cache this value. It would only change through governance, so almost never.

jmrossy (commented):
Mostly looks good to me too!

packages/attestation-service/migrations/20200506234935-attestation-v2.js
aslawson: it did work... but that's a good point.  perhaps i should just inverse these to be safe

packages/attestation-service/migrations/20200506234935-attestation-v2.js
aslawson: good point -- will update

packages/attestation-service/migrations/20200506234935-attestation-v2.js
aslawson: gotcha -- that makes sense

packages/attestation-service/src/requestHandlers/attestation.ts
aslawson: ya -- i can add the sequelizeLogger in the transaction which seems to be how logging is handled in this service

nambrot (approved):

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -30,6 +31,18 @@ export const AttestationRequestType = t.type({
 
 export type AttestationRequest = t.TypeOf<typeof AttestationRequestType>
 
+const ATTESTATION_EXPIRY_TIMEOUT_MS = 60 * 60 * 24 * 100 // 1 day
```
jmrossy: I think you're off by a factor of 10 here

aslawson: nice catch

PR #3756, opened 2020-05-07
Remove enzyme as a testing library

cmcewen:
### Description

We have switched to `react-native-testing-library`, so this just moves everything over to that library and removes the enzyme dependency

### Tested

This is just tests


jmrossy (approved):
💯

PR #3761, opened 2020-05-07
(WIP) build: better selection of tests run in CI

### LINKED ISSUE #3746: Smarter selection of which tests should be run

AlexBHarley:
### Description

graph

Here's a visualisation of our dependencies, this is now source controlled in the `dependency-graph.json` file. This PR improves our `ci_check_if_test_should_run_v2.sh` script so that we now only run tests for a package when:

- the yarn.lock file has changed
- the circleci config has changed
- the package has changed
- a dependency of the package has changed.

The `dependency-graph.json` file also gets automatically generated in CI and will exit early if it notices any differences from the checked in version. This is similar to how we cache our `node_modules` and confirm they match with our `yarn.lock`.

### Other changes

Moved some utility functions to their own file in the celotool repository. This is so we don't need to install the bls-zexe stuff just to run our `execCmd` helpers. 

### Tested

Ran a bunch of builds locally and on CI.

Working correctly in the following scenarios, with a dependency graph A->B->C (A depends on B depends on C).

- A has changed 
  - `./scripts/ci_check_if_test_should_run_v2.sh A` = true
  - `./scripts/ci_check_if_test_should_run_v2.sh B` = false
  - `./scripts/ci_check_if_test_should_run_v2.sh C` = false
- B has changed
  - `./scripts/ci_check_if_test_should_run_v2.sh A` = true
  - `./scripts/ci_check_if_test_should_run_v2.sh B` = true
  - `./scripts/ci_check_if_test_should_run_v2.sh C` = false
- C has changed
  - `./scripts/ci_check_if_test_should_run_v2.sh A` = true
  - `./scripts/ci_check_if_test_should_run_v2.sh B` = true
  - `./scripts/ci_check_if_test_should_run_v2.sh C` = true

### Related issues

- Fixes #3746


.circleci/config.yml
```diff
@@ -641,8 +682,7 @@ jobs:
       - run:
           name: Check if the test should run
           command: |
-            FILES_TO_CHECK="${PWD}/packages/celotool,${PWD}/packages/protocol,${PWD}/.circleci/config.yml"
-            ./scripts/ci_check_if_test_should_run_v2.sh ${FILES_TO_CHECK}
+            ./scripts/ci_check_if_test_should_run_v2.sh celotool
```
asaj, line 683: For these e2e tests we should ensure that celotool has a dependency on the `protocol` package

packages/celotool/src/lib/cmd-utils.ts
```diff
@@ -0,0 +1,124 @@
+import { exec, spawn, SpawnOptions } from 'child_process'
```
asaj, line 1: Any changes in this file? Or just copy / paste

scripts/check_if_test_should_run_v2.ts
```diff
@@ -39,33 +53,64 @@ async function checkIfTestShouldRun() {
...
-    if (anyPathsChanged) {
+
+  const circleciChangeCommit = await getChangeCommit(join(rootDirectory, '.circleci', 'config.yml'))
+  const yarnLockChangeCommit = await getChangeCommit(join(rootDirectory, 'yarn.lock'))
+  if (
+    branchCommits.includes(yarnLockChangeCommit) ||
+    branchCommits.includes(circleciChangeCommit)
+  ) {
+    // always run tests when yarn.lock or circlici config have changed
+    console.info('true')
```
asaj: nit: consider `circleci config or yarn.lock file changed, tests should run`

scripts/check_if_test_should_run_v2.ts
```diff
@@ -39,33 +53,64 @@ async function checkIfTestShouldRun() {
...
+  const changedPackages = await getChangedPackages(branchCommits)
+  logMessage(`Found ${changedPackages.length} changed packages (${changedPackages.join(', ')})`)
+
+  function hasChangedDependencies(packageName: string): Boolean {
+    if (changedPackages.includes(packageName)) {
+      return true
+    }
+
+    for (const dep of dependencyGraph[packageName]) {
+      return hasChangedDependencies(dep)
```
asaj: Does this work? It seems like for a package A that depends on packages B and C, `hasChangedDependencies(A)` would return `false` if C changed but B did not.

Perhaps instead you could do `dependencyGraph[packageName].map(hasChangedDependencies).some(identity)` or just `or` every return value together

scripts/dependency-graph.json
```diff
@@ -0,0 +1,30 @@
+{
```
asaj: Can we add a CI step to generate this automatically and fail if this file is out of date? We do something similar with `yarn.lock` which is very helpful for making sure it doesn't become stale

asaj (commented):
This is awesome!! 👏

packages/celotool/src/lib/cmd-utils.ts
AlexBHarley: Just copy paste

scripts/dependency-graph.json
AlexBHarley: Done

.circleci/config.yml
AlexBHarley: Added protocol as a dev dep of the celo tool

scripts/check_if_test_should_run_v2.ts
```diff
@@ -2,15 +2,21 @@
 // Say the caller is in the celotool dir and wants to check if celotool or protocol dir has changed then
 // the sample usage will be
 // node -r ts-node/register scripts/check_if_test_should_run_v2.ts --dirs packages/protocol,packages/celotool
-// Prints "true" to stdout if the tests should run
-// Prints "false" otherwise
+// Prints "false" if tests shouldn't run, anything else will cause tests to run
```
asaj, line 5: Will "logMessage" avoid this?

IMO this is unnecessarily fragile, could we instead explicitly call `process.exit` to decide whether or not the tests should run?

scripts/dependency-graph-utils.ts
```diff
@@ -0,0 +1,67 @@
...
+import { createHash } from 'crypto'
+import { existsSync, readdirSync, readFileSync, writeFileSync } from 'fs'
+import { join } from 'path'
+
+const packagesDirectory = join(__dirname, '..', 'packages')
+export const filename = 'dependency-graph.json'
+
+// Easier to operate on folder names of packages rather than the name
+// from the package.json file. People are more familiar with folder names as well
+function getFoldersFromPackageNames(): { [x: string]: string } {
```
asaj: nit: s/folder/directory

scripts/dependency-graph-utils.ts
```diff
@@ -0,0 +1,67 @@
...
+const hash = (input: string): string =>
+  createHash('SHA256')
+    .update(input)
+    .digest('base64')
+
+export const graphHasChanged = async (): Promise<boolean> => {
+  const oldGraph = readFileSync(filename).toString()
+  const newGraph = await buildGraph()
+
+  if (hash(oldGraph) !== hash(newGraph)) {
```
asaj: nit: consider `return hash(oldGraph) !== hash(newGraph)`

asaj (approved):
Sweet!

scripts/check_if_test_should_run_v2.ts
AlexBHarley: I agree that it's kind of weird but `process.exit(1)` for a script that runs successfully is also not great.

PR #3762, opened 2020-05-07
Two AKS clusters for oracle infra

tkporter:
### Description

This PR includes 2 main things:
* Support for 2 AKS oracle clusters in a single env
* Oracle pods using full nodes in the same namespace of their cluster for RPCs

#### 2 AKS Clusters

An AKS cluster is associated with an `OracleAzureContext`:
```
enum OracleAzureContext {
  PRIMARY = 'PRIMARY',
  SECONDARY = 'SECONDARY',
}
```

When running any celotool command to do with deploying oracles or oracle full nodes, the context must be specified. For a cleaner interface, these 2 boolean flags are used: `--primary`, `--secondary`. As an example, this is how one would deploy oracle full nodes, and then oracle pods all to the secondary cluster:

```
$ celotooljs deploy initial oracle-fullnode -e myenv --secondary
$ celotooljs deploy initial oracle -e myenv --secondary --useFullNodes
```

The celotool deploy commands `aks-fullnode` have been renamed to `oracle-fullnode` to indicate they are only meant for oracle use and have the flags `--primary` and `--secondary`. All `aks-fullnode` logic still remains the same as it only needs to be aware of 1 AKS cluster. Oracle deployment is separate from oracle full node deployment so that we can safely decouple upgrades of the oracle client from the full nodes

Now there are azure-related env vars that are specific to their context:

```
ORACLE_PRIMARY_AZURE_SUBSCRIPTION_ID=...
ORACLE_PRIMARY_AZURE_TENANT_ID=...
ORACLE_PRIMARY_AZURE_KUBERNETES_CLUSTER_NAME=...
ORACLE_PRIMARY_AZURE_KUBERNETES_RESOURCE_GROUP=...
ORACLE_PRIMARY_ADDRESS_AZURE_KEY_VAULTS=<address>:<key vault name>,<address>:<key vault name>

ORACLE_SECONDARY_AZURE_SUBSCRIPTION_ID=...
ORACLE_SECONDARY_AZURE_TENANT_ID=...
ORACLE_SECONDARY_AZURE_KUBERNETES_CLUSTER_NAME=...
ORACLE_SECONDARY_AZURE_KUBERNETES_RESOURCE_GROUP=...
ORACLE_SECONDARY_ADDRESS_AZURE_KEY_VAULTS=<address>:<key vault name>,<address>:<key vault name>
```

I introduced a new type, `AzureClusterConfig`, that has the basic info on an AKS cluster that's constructed using those above env vars for a specific context.

To make oracle identities more clear, I created a new type `OracleIdentity` to contain info on the address/keyvault/managed identity for a specific oracle.

### Other changes

* Removed the aks-prometheus commands, and instead added installing prometheus as a part of the cluster setup checks that are made upon every AKS related command
* Managed identities now have the name: `<key vault name>-<address>` to more elegantly make them distinct
* Added a component label to the `celo-fullnode` helm chart to properly delete PVCs when upgrading & resetting
* Updated port-forwarding to accommodate the new app label for the testnet helm chart
* Changed PVCs for the testnet helm chart from 2GB to 15GB-- 2GB was really low

### Tested

Successfully running oracles across 2 AKS clusters using the full nodes for RPCs

### Backwards compatibility

`aks-fullnode` deploy commands are replaced, and all oracle deploy cmds must specify one of `--primary` or `--secondary`. Most Azure related env vars are replaced as well


packages/celotool/src/cmds/deploy/initial/oracle-fullnode.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const builder = (argv: yargs.Argv) => {
+  return addOracleMiddleware(argv)
+}
+
+export const handler = async (argv: OracleFullNodeInitialArgv) => {
+  const oracleAzureContext = getOracleAzureContext(argv.primary)
+  await switchToAzureContextCluster(argv.celoEnv, oracleAzureContext)
+  const clusterConfig = getAzureClusterConfig(oracleAzureContext)
+  await installFullNodeChart(argv.celoEnv, clusterConfig)
```
codyborn, line 23: What do you think about wrapping these three methods in a single call that returns the cluster config taking into account the context?
  

```
const oracleAzureContext = getOracleAzureContext(argv.primary)
await switchToAzureContextCluster(argv.celoEnv, oracleAzureContext)
const clusterConfig = getAzureClusterConfig(oracleAzureContext)
```

packages/celotool/src/lib/oracle.ts
```diff
@@ -37,86 +107,167 @@ async function helmParameters(celoEnv: string) {
...
   const servicePrincipalClientId = rawServicePrincipalClientId.trim()
   await execCmdWithExitOnFailure(
     `az role assignment create --role "Managed Identity Operator" --assignee ${servicePrincipalClientId} --scope ${identity.id}`
   )
   // Allow the oracle identity to access the correct key vault
   await execCmdWithExitOnFailure(
-    `az keyvault set-policy --name ${keyVaultName(
-      index
-    )} --key-permissions {get,list,sign} --object-id ${identity.principalId} -g ${resourceGroup()}`
+    `az keyvault set-policy --name ${oracleIdentity.keyVaultName} --key-permissions {get,list,sign} --object-id ${identity.principalId} -g ${clusterConfig.resourceGroup}`
```
codyborn: Do we need to clean these up when an Oracle is moved or deleted?

codyborn (approved):

packages/celotool/src/lib/oracle.ts
tkporter: Great idea! Added that

packages/celotool/src/cmds/deploy/initial/oracle-fullnode.ts
tkporter: I kind of like having the separate functions, mostly because it's only for the oracle full nodes, they're separate things, and I don't like relying on `argv.primary`. If we find ourselves needing to do this in more places then I'd be up for merging them into a more portable function

packages/celotool/src/lib/azure.ts
```diff
@@ -74,29 +86,39 @@ async function installAADPodIdentity() {
...
 }
 
-export function resourceGroup() {
-  return fetchEnv(envVar.AZURE_KUBERNETES_RESOURCE_GROUP)
-}
-
-export function clusterName() {
-  return fetchEnv(envVar.AZURE_KUBERNETES_CLUSTER_NAME)
+/**
+ * deleteIdentity gets basic info on an existing identity
```
codyborn, line 101: Comment needs updating

PR #3764, opened 2020-05-07
[Wallet] Gold screen redesign

### LINKED ISSUE #3659: New Wallet Design: Gold
Parent Issue: https://github.com/celo-org/celo-monorepo/issues/3331
Figma: https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=2995%3A35982

i1skn:
### Description

This PR implements [Celo Gold](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=2995%3A35982) redesigned screens with some exceptions:
- Fees brake-down is not implemented (first we need to change how we fetch fees form `blockchain-api` and de-hardcode other fees in Wallet codebase).
- Sticky header with the Balance (currently it says just "Gold")

### Other changes

- Change `Button` logic with `BtnSize.FULL` a bit, so it's not needed to add manually styles every time we want a full width button.
- Change `Button` logic, so it accepts `string | Element` (needed when we use <CurrencyDisplay /> inside the button text)
- Small fixes after the upgrade of react navigation to v5

### Tested

Android

### Related issues

- Fixes #3659

### Backwards compatibility

Yes


packages/react-components/components/Button.v2.tsx
```diff
@@ -27,7 +27,7 @@ export enum BtnSizes {
 export interface ButtonProps {
   onPress: () => void
   style?: StyleProp<ViewStyle>
-  text: string
+  text: string | JSX.Element
```
i1skn: This is needed, when we show <CurrencyDisplay /> inside the button text.

packages/react-components/styles/fonts.v2.tsx
```diff
@@ -67,6 +67,12 @@ const fontStyles = StyleSheet.create({
   center: {
     textAlign: 'center',
   },
+  mediumNumber: {
```
i1skn, line 70: This is not in the Design Kit, but we use it more than twice on Gold Screen - I figure we could use it the future. Cc @coreycelo

packages/mobile/src/exchange/CeloGoldHistoryChart.tsx
```diff
@@ -230,24 +262,28 @@ function CeloGoldHistoryChart({ t, testID, i18n }: Props) {
...
-            %)
+            {rateWentUp ? '▴' : '▾'} {rateChangeInPercentage.toFormat(2)}%
           </Text>
         </View>
       </View>
       <VictoryGroup
         domainPadding={CHART_DOMAIN_PADDING}
         singleQuadrantDomainPadding={false}
-        padding={{ left: 0 }}
+        padding={{ left: variables.contentPadding, right: variables.contentPadding }}
```
aaronmgdr: nit create a constant to avoid object recreation

aaronmgdr (approved):

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
```diff
@@ -13,41 +22,115 @@ export interface ExchangeConfirmationCardProps {
...
-    <View style={styles.container}>
-      <View style={styles.exchange}>
-        <CurrencyDisplay type={DisplayType.Big} amount={makerAmount} size={36} useColors={true} />
-        <View style={styles.arrow}>
-          <RoundedArrow />
-        </View>
-        <CurrencyDisplay type={DisplayType.Big} amount={takerAmount} size={36} useColors={true} />
+    <SafeAreaView style={styles.container}>
+      <View style={styles.paddedContainer}>
+        <ScrollView>
```
jeanregisser: For fullscreen views, it's nicer to put `<SafeAreaView>` inside the `<ScrollView>` so when scrolling up and down the content is not clipped before the bottom/top of the screen.

packages/mobile/src/navigator/Headers.v2.tsx
```diff
@@ -52,19 +52,19 @@ export const emptyHeader = {
   headerTitle: '',
   headerTitleStyle: [styles.headerTitle, componentStyles.screenHeader],
   headerTitleContainerStyle: {
-    alignItems: 'center',
+    alignItems: 'center' as 'center',
   },
-  headerTitleAlign: 'center',
+  headerTitleAlign: 'center' as 'center',
```
jeanregisser: 😅

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -75,27 +87,24 @@ import VerificationSuccessScreen from 'src/verify/VerificationSuccessScreen'
 
 const Stack = createStackNavigator<StackParamList>()
 
-export const headerArea = {
+export const defaultScreenOptions = {
   // Force this for now on iOS so screen transitions look normal
   // given we intentionally hide the bottom separator from the nav bar
```
jeanregisser: Looks like this comment can be removed since it was related to the `headerMode` property.

packages/react-components/components/Button.v2.tsx
jeanregisser: It's preferable to use `React.ReactNode` 
https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

packages/react-components/components/Button.v2.tsx
```diff
@@ -129,3 +126,10 @@ function getStyle(size: BtnSizes | undefined, backgroundColor: Colors) {
       return { ...styles.button, ...styles.medium, backgroundColor }
   }
 }
+
+function getStyleForWrapper(
+  size: BtnSizes | undefined,
+  style: StyleProp<ViewStyle>
+): StyleProp<ViewStyle> {
+  return [{ flexDirection: size === BtnSizes.FULL ? 'column' : 'row' }, style]
+}
```
jeanregisser, line 136: This doesn't need to be addressed in the PR but I think the current `BtnSize` API is limiting.
In another PR I needed a button having both `FULL` (full width) and `SMALL` (40pt height) behaviors.
I don't have a proposition right now but wanted to raise this.

packages/react-components/components/SectionHead.v2.tsx
```diff
@@ -0,0 +1,30 @@
...
+    paddingBottom: 8,
+    paddingHorizontal: 16,
+  },
+  text: {
+    ...fontStyles.h2,
+    color: colors.dark,
+  },
+})
+
+export default SectionHead
```
jeanregisser: I've updated `SectionHeadNew` in a previous master that's already been merged:
https://github.com/celo-org/celo-monorepo/blob/ee5242cb59ef61a35718d76c8dc2447d0c71c7a0/packages/react-components/components/SectionHeadNew.tsx
We can use that one now.

packages/mobile/src/transactions/GoldTransactionFeedItem.tsx
```diff
@@ -0,0 +1,155 @@
...
+            <>
+              <Text style={styles.exchangeRate}> @ </Text>
+              <CurrencyDisplay
+                amount={exchangeRateAmount}
+                hideSymbol={false}
+                hideCode={true}
+                showLocalAmount={true}
+                style={styles.exchangeRate}
+              />
+            </>
```
jeanregisser, line 65: If you're only interested in displaying a specific value and not doing conversions, it looks like you can simplify to this:

```jsx
const localAmount = (isSellGoldTx ? makerAmount : takerAmount).localAmount!
const exchangeRateAmount = {
  value: localAmount.exchangeRate,
  currencyCode: localAmount.currencyCode,
  localAmount: null
}

// Then
<CurrencyDisplay
  amount={exchangeRateAmount}
  hideSymbol={false}
  hideCode={true}
  showLocalAmount={false}
  style={styles.exchangeRate}
/>
```

jeanregisser (changes requested):
This looks awesome!! 👍 🎉 
I've suggested a couple of small changes.

packages/react-components/components/Button.v2.tsx
i1skn: Yeah, I was thinking about this as well to split this into two verticals: font size and horizontal filling. And started to think that this is needed, when I saw your workaround on this, when you were needed full width with small font size.

packages/react-components/components/SectionHead.v2.tsx
i1skn: On the Gold tab they have a different styling, so I will rename to `SectionHeadGold.tsx` for now.

packages/mobile/src/transactions/GoldTransactionFeedItem.tsx
i1skn: Unfortunately this would not work with currency symbols. See here:
https://github.com/celo-org/celo-monorepo/blob/337d28c94e418afea2d6ec05f5a7930916fc8371/packages/mobile/src/components/CurrencyDisplay.tsx#L151-L155

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
i1skn: Would content overlap with the bar in the bottom on iPhone X in this case?

packages/mobile/src/transactions/GoldTransactionFeedItem.tsx
i1skn: I think we should decouple conversion and representation inside `CurrencyDisplay`, but that is for a different PR :)

packages/react-components/components/SectionHead.v2.tsx
jeanregisser: Oh ok, all good then 👍

packages/mobile/src/transactions/GoldTransactionFeedItem.tsx
jeanregisser: ah 😢, I intended to merge LocalCurrency and CURRENCIES to avoid this kind of mess.
All good then 👍

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
jeanregisser: Yes but that's what is preferred visually, the difference is that there's some padding at the bottom so you can still fully view all the content from the scrollview without overlap.

jeanregisser (approved):
🚀

packages/mobile/src/navigator/Headers.v2.tsx
```diff
@@ -59,9 +59,10 @@ export const emptyHeader: StackNavigationOptions = {
   headerShown: true,
   headerTitleStyle: [styles.headerTitle, componentStyles.screenHeader],
   headerTitleContainerStyle: {
-    alignItems: 'center',
+    alignItems: 'center' as 'center',
```
aaronmgdr, line 62: btw since this is now typed as `StackNavigationOptions` the `as 'center'` should be no longer needed

PR #3765, opened 2020-05-07
Add PGPNP support to iOS

### LINKED ISSUE #3698: App signature hashes in Attestation Service
### Expected Behavior

Want the values hardcoded in for simplicity.
Need values for both the Alfajores and Mainnet wallet builds

### Current Behavior

Gets it from the `APP_SIGNATURE` config

jmrossy:
### Description

* Update to react-native-treshold-bls version that includes iOS support
* Add private hash logic to invite saga for escrow
* Fix minor issue with quota calculation in PGPNP service
* Have app inform attestation service of desired sms retriever sig
* Cleanup now dead sms retriever configs

### Tested

Retrieved salt on iOS
No regressions on Android.
Tested update to Att. service on Alfa Staging + Alfajores

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/463
- Fixes https://github.com/celo-org/celo-monorepo/issues/3698

### Backwards compatibility

Yes (feature still behind a feature flag)


packages/mobile/package.json
```diff
@@ -81,15 +81,15 @@
...
+    "react-native-blind-threshold-bls": "https://github.com/celo-org/react-native-blind-threshold-bls#e071514",
     "react-native-camera": "^3.23.1",
     "react-native-clock-sync": "^1.0.0",
     "react-native-config": "https://github.com/luggit/react-native-config#89a602b",
     "react-native-confirm-device-credentials": "git://github.com/celo-org/react-native-confirm-device-credentials#436afa9",
     "react-native-contacts": "https://github.com/celo-org/react-native-contacts#e473717",
     "react-native-device-info": "^5.5.6",
     "react-native-exit-app": "^1.1.0",
-    "react-native-fast-crypto": "git+https://github.com/celo-org/react-native-fast-crypto#3946f9",
+    "react-native-fast-crypto": "https://github.com/celo-org/react-native-fast-crypto#df1c171",
```
jmrossy, line 92: No logical changes to this package. I just merged the branch we were using there to master to make things easier to follow.

packages/mobile/ios/Podfile
```diff
@@ -102,6 +102,8 @@ target "celo" do
 
   pod 'Segment-Firebase'
 
+  pod 'react-native-blind-threshold-bls', :path => '../../../node_modules/react-native-blind-threshold-bls'
```
cmcewen: This shouldn't be necessary with the packager autolinking - can you try to get that working?

packages/mobile/package.json
cmcewen: I think react-native-bip39 also has a dep there which we should update. Feel like this is another argument to move things into the monorepo 😄

yarn.lock
```diff
@@ -26942,6 +26942,13 @@ react-native-exit-app@^1.1.0:
     buffer "^5.0.8"
     rfc4648 "^1.0.0"
 
+"react-native-fast-crypto@https://github.com/celo-org/react-native-fast-crypto#df1c171":
```
cmcewen: this should be a swap, no?

packages/mobile/package.json
```diff
@@ -17,10 +17,10 @@
     "dev:clean-android": "cd android && ./gradlew clean",
     "dev:send-text": "adb shell input text",
     "dev:reload": "adb shell input text \"RR\"",
-    "dev:emulator": "bash ./scripts/start_emulator.sh &",
-    "dev:remote": "remotedev --hostname=localhost --port=8000",
+    "dev:emulator:list": "$ANDROID_HOME/emulator/emulator -list-avds",
+    "dev:emulator:run": "$ANDROID_HOME/emulator/emulator -avd",
```
cmcewen, line 21: I think if the functionality doesn't change it's nice to leave the name the same

jmrossy: It was broken before so doubt anyone was using it. I prefer it like this

packages/mobile/ios/Podfile
jmrossy: Why are RN-geth and CeloBlockchain here then?

yarn.lock
jmrossy: Something else must still be using the old hash

packages/mobile/package.json
jmrossy: Okay, fixed the reference in bip39

packages/celotool/src/cmds/deploy/initial/verification-pool.ts
```diff
@@ -11,16 +11,10 @@ export const builder = {}
 export const handler = async (argv: InitialArgv) => {
   const vpConfig = await getVerificationPoolConfig(argv.celoEnv)
   const projName = fetchEnvOrFallback(envVar.TESTNET_PROJECT_NAME, 'celo-testnet')
-  const appSignature = fetchEnvOrFallback(envVar.SMS_RETRIEVER_HASH_CODE, '')
+  // const appSignature = fetchEnvOrFallback(envVar.SMS_RETRIEVER_HASH_CODE, '')
```
nambrot: We should just remove this now, right?

nambrot (approved):
Attestation service changes look great, genius idea to make the requestor configure it!

packages/celotool/src/cmds/deploy/initial/verification-pool.ts
jmrossy: We could but I'd rather not further grow the scope of this PR by touching verification pool code

jmrossy: I'm open to removing the pool at this point, though I still think the verifier should stay for now

cmcewen (approved):
🙌 nice work on this

PR #3768, opened 2020-05-08
PGPNP: Server-side && Client-side authentication check

codyborn:
### Description

New auth check to ensure that the account making the request has custody over the private key matching the account.

Some considerations:
- Protocol should be fwd compatible
- Auth should not be re-usable to make a different request (ex. use previously signed payload to query new phone number salt)

To accommodate both of these requirements, we sign the entire message body and include the signature in the authorization header. Authorization header was used since the standard includes use for custom auth scheme. If we add new fields to the body, the server auth will not break since it just checks that the user signed the `request.body`. If the signed message is ever leaked, a new user can only replay the same query and cannot request new phone numbers as the old user.

When performing the signing, I intentionally used a method that includes the Ethereum header `\x19Ethereum Signed Message:\n` since this is compatible with most ECDSA signing libs including our Ledger app.

Adding E2E tests that can be ran locally or against a deployed instance. These E2E tests will eventually be ran as part of our CICD process.

### Tested

See e2e tests. Deployed PGPNP to staging and tested emulated android device.
image

### Backwards compatibility

Not backwards compatible.


packages/phone-number-privacy/test/end-to-end/get-blinded-salt.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      const response = await postToSignMessage(
+        BLINDED_PHONE_NUMBER,
+        PRIVATE_KEY1,
+        ACCOUNT_ADDRESS1,
+        'invalid'
+      )
+      expect(response.status).toBe(401)
+    })
+
+    it('With auth header signer mismatch', async () => {
```
aslawson, line 60: might also be nice to have test with happy path -- correct auth header

packages/phone-number-privacy/test/end-to-end/get-blinded-salt.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    })
+  })
+
+  it('Address salt querying out of quota', async () => {
+    const response = await postToSignMessage(BLINDED_PHONE_NUMBER, PRIVATE_KEY1, ACCOUNT_ADDRESS1)
+    expect(response.status).toBe(403)
+  })
+})
+
+async function postToSignMessage(
```
aslawson, line 97: so does running this require we have the service running locally?

packages/phone-number-privacy/src/common/identity.ts
```diff
@@ -1,15 +1,26 @@
...
+
+  // https://tools.ietf.org/html/rfc7235#section-4.2
+  const messageSignature = request.get('Authorization')
+  const signer = request.body.account
+  if (!messageSignature || !signer) {
+    return false
+  }
+
+  const message = JSON.stringify(request.body)
+  return verifySignature(message, messageSignature, signer)
```
aslawson, line 23: nice

aslawson (approved):

packages/mobile/package.json
```diff
@@ -17,7 +17,7 @@
     "dev:clean-android": "cd android && ./gradlew clean",
     "dev:send-text": "adb shell input text",
     "dev:reload": "adb shell input text \"RR\"",
-    "dev:emulator": "bash ./scripts/start_emulator.sh &",
+    "dev:emulator": "$ANDROID_HOME/emulator/emulator -avd Nexus_5X_API_28_x86 -no-boot-anim",
```
jmrossy: Funny, I also fixed this in my PR here, adding a list and a run command:
https://github.com/celo-org/celo-monorepo/pull/3765/files#diff-4bbd284d4cdf8b13bb5dc9aed93f552bR20

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
```diff
@@ -13,7 +13,11 @@ export async function handleGetContactMatches(request: Request, response: Respon
       respondWithError(response, 400, ErrorMessages.INVALID_INPUT)
       return
     }
-    authenticateUser()
+    if (!authenticateUser(request)) {
+      logger.warn(ErrorMessages.UNAUTHENTICATED_USER)
```
jmrossy: No need for this, respondWithError logs the error already

packages/phone-number-privacy/src/salt-generation/get-salt.ts
```diff
@@ -13,7 +13,11 @@ export async function handleGetBlindedMessageForSalt(request: Request, response:
       respondWithError(response, 400, ErrorMessages.INVALID_INPUT)
       return
     }
-    authenticateUser()
+    if (!authenticateUser(request)) {
+      logger.warn(ErrorMessages.UNAUTHENTICATED_USER)
```
jmrossy: ditto

packages/phone-number-privacy/test/utils.ts
```diff
@@ -43,3 +45,16 @@ export function createMockWeb3(txCount: number) {
     },
   }
 }
+
+export function getBlindedPhoneNumber(phoneNumber: string, blindingFactor: Buffer): string {
+  const blindedPhoneNumber = threshold.blind(Buffer.from(phoneNumber), blindingFactor).message
+  return _uint8ArrayToBase64(blindedPhoneNumber)
+}
+
+function _uint8ArrayToBase64(bytes: Uint8Array) {
```
jmrossy: It's not a problem but small nit, no need for the leading underscore. We already know this function is 'private' to this file cause it's not exported

packages/phone-number-privacy/test/utils.ts
```diff
@@ -1,5 +1,7 @@
 import BigNumber from 'bignumber.js'
+import threshold from 'blind-threshold-bls'
 import Web3 from 'web3'
+const btoa = require('btoa')
```
jmrossy: Can we use es6 import syntax here? Should be
`import btoa from 'btoa'`
or in rare cases depending on the lib's export style:
`import * as btoa from 'btoa'`

jmrossy (dismissed):
Looks great! Just some optional comments below

packages/phone-number-privacy/test/end-to-end/get-blinded-salt.test.ts
codyborn: Agreed! I wanted to add this but the account will hit the quota limit unless it's prefunded (hence the out of quota limit test is really the auth happy path). I added a new, disabled test to test the 200 case but it requires the tester to prefund the test account to get past the quota limit.

packages/phone-number-privacy/test/end-to-end/get-blinded-salt.test.ts
codyborn: Yes, there must be a service running for these tests. The tester can run it locally and just run the tests, or point the .env parameter to a deployed instance. This is an easy way to perform validation after you've deployed an upgrade to staging for example. I want this to eventually be automated as part of our CICD process.

packages/mobile/package.json
codyborn: Awesome

packages/mobile/src/config.ts
```diff
@@ -55,7 +55,7 @@ export const FAQ_LINK = 'https://celo.org/faq'
 export const FORUM_LINK = 'https://forum.celo.org/c/support'
 export const CELO_SUPPORT_EMAIL_ADDRESS = 'support@celo.org'
 export const DEFAULT_FORNO_URL = `https://${DEFAULT_TESTNET}-forno.celo-testnet.org`
-// TODO configure per envrionment
+// TODO configure per environment
```
jmrossy, line 59: Thank you! I have a bad habit of writing typo-ridden comments haha

packages/mobile/src/identity/privacy.ts
```diff
@@ -97,20 +112,30 @@ interface SignMessageResponse {
...
   Logger.debug(`${TAG}@postToSignMessage`, `Posting to ${SIGN_MESSAGE_ENDPOINT}`)
-  // Cody TODO: Sign payload using account privkey
+  const body = JSON.stringify({
+    hashedPhoneNumber,
+    blindedQueryPhoneNumber: base64BlindedMessage.trim(),
+    account,
+  })
+
+  // Sign payload using account privkey
+  const authHeader = await contractKit.web3.eth.sign(body, account)
```
jmrossy, line 123: Can you confirm this works when the app is in forno mode. You can enable it via the Settings page ('Data-saver mode')

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
```diff
@@ -48,12 +49,15 @@ function isValidGetContactMatchesInput(requestBody: any): boolean {
...
-    hasValidContractPhoneNumbersParam(requestBody)
+    hasValidContractPhoneNumbersParam(requestBody) &&
+    isBodyReasonablySized(requestBody)
   )
 }
 
 function hasValidAccountParam(requestBody: any): boolean {
-  return requestBody.account && isValidAddress(requestBody.account)
+  return (
+    requestBody.account && isValidAddress(requestBody.account) && isBodyReasonablySized(requestBody)
```
jmrossy: Why check `isBodyReasonablySized(requestBody)` here when we also check it above?

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
```diff
@@ -63,3 +67,7 @@ function hasValidUserPhoneNumberParam(requestBody: any): boolean {
 function hasValidContractPhoneNumbersParam(requestBody: any): boolean {
   return requestBody.contactPhoneNumbers && Array.isArray(requestBody.contactPhoneNumbers)
 }
+
+function isBodyReasonablySized(requestBody: any): boolean {
+  return JSON.stringify(requestBody).length <= REASONABLE_BODY_CHAR_LIMIT
```
jmrossy: Not a big deal but I wonder if there's a way to check this without re-stringifying the request body, since it gets parsed automatically by express.

packages/phone-number-privacy/test/end-to-end/get-blinded-salt.test.ts
```diff
@@ -17,6 +19,7 @@ const PHONE_NUMBER = '+15555555555'
 const IDENTIFIER = PhoneNumberUtils.getPhoneHash(PHONE_NUMBER)
 const BLINDING_FACTOR = new Buffer('0IsBvRfkBrkKCIW6HV0/T1zrzjQSe8wRyU3PKojCnww=', 'base64')
 const BLINDED_PHONE_NUMBER = getBlindedPhoneNumber(PHONE_NUMBER, BLINDING_FACTOR)
+const DEFAULT_FORNO_URL = `https://alfajores-forno.celo-testnet.org`
```
jmrossy: Could you get this from the DEV_MODE config? it's already specified there

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
```diff
@@ -48,12 +49,15 @@ function isValidGetContactMatchesInput(requestBody: any): boolean {
   return (
     hasValidAccountParam(requestBody) &&
     hasValidUserPhoneNumberParam(requestBody) &&
-    hasValidContractPhoneNumbersParam(requestBody)
+    hasValidContractPhoneNumbersParam(requestBody) &&
+    isBodyReasonablySized(requestBody)
```
aslawson, line 54: what is this protected against? or validating for explicitly?

packages/phone-number-privacy/src/common/identity.ts
```diff
@@ -1,15 +1,26 @@
...
+ * Authorization header should contain the EC signed body
  */
-export function authenticateUser() {
+export function authenticateUser(request: Request): boolean {
   logger.debug('Authenticating user')
-  // TODO [amyslawson]
-  return
+
+  // https://tools.ietf.org/html/rfc7235#section-4.2
+  const messageSignature = request.get('Authorization')
```
aslawson, line 16: can we update the curl examples in index.js to include what the header should look like?

aslawson (approved):
lgtm

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
codyborn: Thanks for noticing this. I've moved these checks a new file, common/input-validation.ts.

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
codyborn: Good question, I wasn't able to find a way to do this using the request object from Express 🤔

packages/mobile/src/identity/privacy.ts
codyborn: Verified that I can get the salt in data-saver mode:
image

packages/phone-number-privacy/src/match-making/get-contact-matches.ts
codyborn: Since we pass the request body (user input) into the EC recover function, I wanted to be sure that nothing the user gave us could break the function. I'm not sure how the hash calculation scales with the size of the input so I tried to set reasonable size restrictions.

packages/phone-number-privacy/src/common/identity.ts
codyborn: Updated!

jmrossy (approved):

PR #3767, opened 2020-05-08
[refresh] Account Key Start / Post Setup  (part 1)

aaronmgdr:
### Description

creates refresh version BackupKeyIntroduction used as Account Key Intro and Account Key PostSetup

*note that I am deliberately using the same Component name so that its clear that this can be directly swapped with the old version. 

### What is missing

1. Screen is not hooked up
   1.1 Button / actions don't necessarily do whats need / navigate

2. Does not handle the top bar except for where specific to this screen. ie wait an hour button is there but hamburger is not

3. handle getting real mneumonic 

#### what is not planned
. convert all components to functions (although it does use mostly functional)

### Other changes

reusable v2 Components
`Logo` (now in SVG)

`Switch`
Screen Shot 2020-05-07 at 5 09 53 PM

`TopBarButton` 
Shown here with gray text, but defaults to ui green
Screen Shot 2020-05-07 at 5 54 36 PM
`CancelButton`

`TextButton`
### Tested

Screen Shot 2020-05-07 at 5 12 16 PM

Screen Shot 2020-05-07 at 5 11 49 PM

### Related issues

- Relates to #walletrefresh

PR #3772, opened 2020-05-09
[refresh] Components for wallet 

aaronmgdr:
### Description

Creates refresh versions of CancelButton, TextButton, TopBarButton

These Changes are also included in https://github.com/celo-org/celo-monorepo/pull/3767

however i extracted out here in hopes they can be merged faster for use by others

### Tested

check


packages/react-components/components/TextButton.v2.tsx
```diff
@@ -0,0 +1,44 @@
...
+import fontStyles from '@celo/react-components/styles/fonts.v2'
+import * as React from 'react'
+import { StyleProp, StyleSheet, Text, TextStyle } from 'react-native'
+
+type Props = TouchableProps & {
+  style?: StyleProp<TextStyle>
+  testID?: string
+}
+// unstyled Touchable Text, good for making other Text Buttons such as TopBarButton
+export function TextButtonGeneric(props: Props) {
```
jeanregisser: nit: Maybe this could be named `BorderlessButton`? and moved to its own file since it's used elsewhere.

packages/mobile/src/navigator/TopBarButton.v2.tsx
```diff
@@ -0,0 +1,32 @@
...
+  testID?: string
+}
+
+export default function TopBarButton({ onPress, style, children, disabled, testID }: Props) {
+  return (
+    <TextButtonGeneric
+      disabled={disabled}
+      onPress={onPress}
+      testID={testID}
+      style={style ? { ...styles.text, ...style } : styles.text}
```
jeanregisser, line 19: nit: how about `style={[styles.text, style]}` instead?

jeanregisser (approved):
Great! 👍 
Left some small nits.

packages/mobile/src/navigator/TopBarButton.v2.tsx
aaronmgdr: @jmrossy usually says to avoid using `[]` in for styles for perf reasons although here if i am destrutcting into this object maybe it's not really any different.

PR #3774, opened 2020-05-09
fix rare bug in Azure HSM Wallet implementation. issue: #3498

### LINKED ISSUE #3498: ContractKit Azure HSM: Unable to generate recovery key from signature
### Expected Behavior

No errors when signing with an Azure HSM

### Current Behavior

I found this error when using an Azure HSM via contractkit:

```
(node:1) UnhandledPromiseRejectionWarning: Error: Unable to generate recovery key from signature.
    at Function.AzureKeyVaultClient.recoverKeyIndex (/celo-oracle/node_modules/@celo/contractkit/lib/wallets/signers/azure-key-vault-client.js:257:15)
    at AzureKeyVaultClient.<anonymous> (/celo-oracle/node_modules/@celo/contractkit/lib/wallets/signers/azure-key-vault-client.js:202:61)
    at step (/celo-oracle/node_modules/@celo/contractkit/lib/wallets/signers/azure-key-vault-client.js:33:23)
    at Object.next (/celo-oracle/node_modules/@celo/contractkit/lib/wallets/signers/azure-key-vault-client.js:14:53)
    at fulfilled (/celo-oracle/node_modules/@celo/contractkit/lib/wallets/signers/azure-key-vault-client.js:5:58)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:1) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 164)
```

At the moment, @codyborn and I haven't been able to reproduce it but he has seen it as well

zviadm:
### Description

bigNumberToBuffer function wasn't returning strings with deterministic length.
I think this should fix issue: #3498

### Tested

> yarn test (but some unrelated errors show up due to local setup)

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/3498


packages/contractkit/src/utils/azure-key-vault-client.ts
```diff
@@ -192,8 +186,13 @@ export class AzureKeyVaultClient {
   private static bufferToBigNumber(input: Buffer): BigNumber {
     return new BigNumber(ensureLeading0x(input.toString('hex')))
   }
-  private static bigNumberToBuffer(input: BigNumber): Buffer {
-    return ethUtil.toBuffer(ensureLeading0x(input.toString(16))) as Buffer
+  private static bigNumberToBuffer(input: BigNumber, size: number): Buffer {
```
tkporter: @codyborn I wonder if we should move `bigNumberToBuffer` and `bufferToBigNumber` into a utils file later on

tkporter (approved):
🙏🙏 this is super helpful, thanks for looking into this

Deferring to @codyborn

packages/contractkit/src/utils/azure-key-vault-client.ts
```diff
@@ -192,8 +186,13 @@ export class AzureKeyVaultClient {
...
+    let hex = input.toString(16)
```
tkporter, line 190: wdyt of a one liner `input.toString(16).padStart(64, '0')`?

zviadm: if you look through my commit history i tried padStart first, but the way contractkit is compiled, it doesnt have support for padStart since it is too new of a javascript.

packages/contractkit/src/utils/azure-key-vault-client.ts
codyborn: Agreed. We can do this in a follow up PR.

packages/contractkit/src/utils/azure-key-vault-client.ts
```diff
@@ -192,8 +186,13 @@ export class AzureKeyVaultClient {
...
+    const hexSize = size * 2
```
codyborn: Can you add a comment for where this '2' comes from:
// Two hex chars per byte

packages/contractkit/src/utils/azure-key-vault-client.ts
codyborn: Can you include units in this variable name: `lengthInBytes`

packages/contractkit/src/utils/azure-key-vault-client.ts
```diff
@@ -151,18 +151,12 @@ export class AzureKeyVaultClient {
...
-      try {
-        const recoveredPublicKeyByteArr = ecdsaRecover(signature, i, hash, compressed)
-        const publicKeyBuff = Buffer.from(recoveredPublicKeyByteArr)
-        const recoveredPublicKey = AzureKeyVaultClient.bufferToBigNumber(publicKeyBuff)
-        debug('Recovered key: ' + recoveredPublicKey)
-        if (publicKey.eq(recoveredPublicKey)) {
-          return i
-        }
-      } catch (e) {
-        if (e.message === 'Public key could not be recover') {
```
codyborn, line 163: I'd still like to keep this try/catch here since the user may be attempting to sign using a non-ECDSA key and this message might help them locate the issue.

codyborn:
Thanks for hunting this down @zviadm! I've ran these tests using an Azure HSM and attempted to repro the bug - everything looks good ✅

packages/contractkit/src/utils/azure-key-vault-client.ts
zviadm: try catch as it was before was silently swallowing error messages because it wasn't re-throwing errors when message didnt match this constant. 
I think it is cleaner without the try/catch and 'ecdsaRecover' will show up in stack trace anyways for anyone who would be debugging issues. It is pretty clear that it is talking about ecdsa keys.

zviadm: also anyone using this, would have to have at least some understanding of what kind of key to use anyways, because ECDSA isn't explicit. There are different types of ECDSA keys, and user must be using specifically 'secp256k1' version, others wont work.

zviadm:
@codyborn btw i just realized, i think there is same bug right here too:
https://github.com/celo-org/celo-monorepo/blob/11814bdb3b44bddb3c19a9ec0f439ad45e86b6bc/packages/contractkit/src/wallets/azure-hsm-wallet.ts#L54

isValidPublicKey can probably fail if first 8 bits of public key happens to be zero and causes it to create 1 byte short public key stirng.

should i fix this in this same diff or fix it in separate diff?

packages/contractkit/src/utils/azure-key-vault-client.ts
codyborn: That's fair. We can also keep the original logic and just rethrow the error in the case it doesn't match this message. I'll leave it up to you.

codyborn:
> @codyborn btw i just realized, i think there is same bug right here too:
> 
> https://github.com/celo-org/celo-monorepo/blob/11814bdb3b44bddb3c19a9ec0f439ad45e86b6bc/packages/contractkit/src/wallets/azure-hsm-wallet.ts#L54
> 
> isValidPublicKey can probably fail if first 8 bits of public key happens to be zero and causes it to create 1 byte short public key stirng.
> 
> should i fix this in this same diff or fix it in separate diff?

Good point; I think it's worth verifying that 
`signingKey.key!.x!` and `signingKey.key!.y!` are both 32 bytes in AzureKeyVaultClient class `getPublicKey()` method. I think including it in this same PR is fine if you're willing to add that.  Thanks again!

codyborn (approved):

zviadm:
> > @codyborn btw i just realized, i think there is same bug right here too:
> > https://github.com/celo-org/celo-monorepo/blob/11814bdb3b44bddb3c19a9ec0f439ad45e86b6bc/packages/contractkit/src/wallets/azure-hsm-wallet.ts#L54
> > 
> > isValidPublicKey can probably fail if first 8 bits of public key happens to be zero and causes it to create 1 byte short public key stirng.
> > should i fix this in this same diff or fix it in separate diff?
> 
> Good point; I think it's worth verifying that
> `signingKey.key!.x!` and `signingKey.key!.y!` are both 32 bytes in AzureKeyVaultClient class `getPublicKey()` method. I think including it in this same PR is fine if you're willing to add that. Thanks again!

i will send that as a separate diff for simplicity after this one gets merged.

PR #3780, opened 2020-05-11
Make DowntimeSlasher easier to use from contractkit

nategraf:
### Description

I went to go slash some validators on Baklava today and realized that figuring out which validators were down was a non-trivial task. In response I added a `isValidatorDown` function to help determine which validators are slashable and adjusted the existing slashing wrapper function to take the signer or validator account address instead of just the account address.

###

* Add missing CLI commands to the headings in `SUMMARY.md`

### Tested

* Manual testing by slashing validators on Baklava


mrsmkl (approved):
Looking good

PR #3786, opened 2020-05-12
[Wallet] Fix gold history chart after library update

i1skn:
### Description

Fix gold history chart after library update

### Tested

Android emulator

### Backwards compatibility

Yes


jeanregisser (approved):
Thanks!! 👍

PR #3790, opened 2020-05-12
Catch fetch error

nambrot:
### Description

If the service is unreachable, it would prevent the command from finishing


mrsmkl (approved):
LGTM

PR #3777, opened 2020-05-11
Enable PGPNP and increment wallet version

jmrossy:
### Description

* Enable prefix for phone hashing everywhere
* Enable PGPNP in the wallet
* Increment wallet version
* Implement basic version of exhausted lookup query handling
* Add ability in wallet to wait for tx failure (currently only waits for success)

### Tested

End-to-end In Alfa staging

### Related Issues

Fixes https://github.com/celo-org/celo-labs/issues/465

### Backwards compatibility

No. This invalidates any existing verifications.

Screen Shot 2020-05-13 at 16 27 01

PR #3791, opened 2020-05-12
[Wallet] Secure Send

### LINKED ISSUE #1777: Secure Send

tarikbellamine:
### Description
When there are multiple addresses associated with a single phone number, which can happen if someone gets SIM swapped or forgets their private key and opens a new account, a manual check by the user is needed to ensure they are transacting with the correct account. They do this by going through the Secure Send flow where they are asked to contact the recipient and confirm either the last 4 digits of their account number or the full thing. 

This comes into play in three different scenarios in which new addresses could be added to the phone number mapping:
1) Send payment
2) Request payment
3) Fulfill a payment request

### Other changes
This necessitated a rework of some of our data flow and logic so I refactored things as needed.

### Tested
Yes

### Related issues
- Fixes #1777 

### Backwards compatibility
Yes


packages/mobile/locales/en-US/sendFlow7.json
```diff
@@ -80,5 +81,28 @@
...
+    "header": "Please confirm {{displayName}}'s account",
+    "button": "Confirm Account Number"
+  },
+  "secureSendExplanation": {
+    "1": "{{e164Number}} has multiple accounts associated with it. {{displayName}} could have created a new Celo account, or may have lost their phone or had their account compromised.",
+    "2": "Before transacting, you must confirm {{displayName}}'s account. You can do this by scanning their QR code or confirming their account number with them."
+  },
+  "confirmAccountNumber": {
+    "title": "Confirm Account Number",
+    "1a": "Please ask {{displayName}} for the last four digits of their account number and enter it below.",
```
annakaz: Nit: can you use more informative keys than 1a 1b?

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -96,11 +104,24 @@ function* updateRecipientsCache(
...
-    // know new status is pending.
+    Logger.debug(
+      TAG + '@fetchPhoneAddressesAndCheckIfRecipientValidationRequired',
+      `Fetching addresses for number`
+    )
+    const oldE164NumberToAddress = yield select(e164NumberToAddressSelector)
+    const oldAddresses = oldE164NumberToAddress[e164Number]
+      ? [...oldE164NumberToAddress[e164Number]]
+      : oldE164NumberToAddress[e164Number]
+    // const oldAddresses = [
```
annakaz: Remove comment?

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -110,12 +131,19 @@ export function* fetchPhoneAddresses({ e164Number }: FetchPhoneAddressesAction)
     ])
 
     const addresses: string[] | null = yield call(getAddresses, e164Number, attestationsWrapper)
+    // const addresses = [
```
annakaz: Remove comment?

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -22,25 +24,57 @@ interface OwnProps {
...
+  fullValidationRequired: boolean
 }
 
 type Props = OwnProps & WithTranslation
 
 const AVATAR_SIZE = 40
 
 export class IncomingPaymentRequestListItem extends React.Component<Props> {
+  componentDidMount() {
+    // need to check latest mapping to prevent user from accepting fradulent requests
```
annakaz: Small nit: capitalize comment

packages/mobile/src/qrcode/QRScanner.tsx
```diff
@@ -88,18 +133,7 @@ class QRScanner extends React.Component<Props> {
...
-            text={t('showYourQRCode')}
-            standard={false}
-            type={BtnTypes.SECONDARY}
-          >
-            <View style={styles.footerIcon}>
-              <QRCode />
-            </View>
-          </Button>
-        </View>
+        {this.renderUsersOwnQRCode()}
```
annakaz: Instead of returning null above, can you do `!scanIsForSecureSend && this.renderUsersOwnQRCode()` or just `!scanIsForSecureSend && [original tsx for rendering own QR coe]`

packages/mobile/src/qrcode/utils.ts
```diff
@@ -48,22 +54,59 @@ export async function shareSVGImage(svg: SVG) {
...
   let data: { address: string; e164PhoneNumber: string; displayName: string } | undefined
+
   try {
     data = JSON.parse(barcode.data)
   } catch (e) {
     Logger.warn(TAG, 'QR code read failed with ' + e)
   }
   if (typeof data !== 'object' || isEmpty(data.address)) {
-    yield put(showError(ErrorMessages.QR_FAILED_NO_ADDRESS))
+    // yield put(showError(ErrorMessages.QR_FAILED_NO_ADDRESS))
```
annakaz: Remove comment?

packages/mobile/src/qrcode/utils.ts
```diff
@@ -48,22 +54,59 @@ export async function shareSVGImage(svg: SVG) {
...
   }
   if (!isValidAddress(data.address)) {
     yield put(showError(ErrorMessages.QR_FAILED_INVALID_ADDRESS))
     return
   }
+  try {
+    if (scanIsForSecureSend) {
+      if (!transactionData) {
+        throw Error(
+          'Error passing through data: Transaction and recipient information not received'
```
annakaz: Is there an error shown to the user when this is thrown?

packages/mobile/src/qrcode/utils.ts
```diff
@@ -48,22 +54,59 @@ export async function shareSVGImage(svg: SVG) {
...
+        )
+      }
+
+      const targetAddress = data.address
+      const { e164PhoneNumber } = transactionData.recipient
+      if (!e164PhoneNumber) {
+        throw Error('Error passing through data: Phone number not part of recipient data')
```
annakaz: Same here is an error shown to the user when this is thrown?

packages/mobile/src/send/ConfirmRecipient.tsx
```diff
@@ -0,0 +1,180 @@
...
+interface OwnProps {
+  navigation: Navigation
+}
+
+interface StateProps {
+  recipient: Recipient
+  transactionData: TransactionData
+  fullValidationRequired: boolean
+  displayName: string
+  startOfSentenceDisplayName: string
```
annakaz: Nit: rename startOfSentenceDisplayName? I don't understand what this means

packages/mobile/src/send/ConfirmRecipientAccount.tsx
```diff
@@ -0,0 +1,301 @@
...
+        </View>
+      )
+    }
+
+    const singleDigitInputComponentArr = ['k', '0', 'F', '4'].map((placeholderValue, index) => (
+      <SingleDigitInput
+        key={placeholderValue}
+        inputValue={singleDigitInputValueArr[index]}
+        inputPlaceholder={placeholderValue}
+        // tslint:disable-next-line:jsx-no-lambda
```
annakaz: Is the lambda necessary?

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -0,0 +1,301 @@
...
+        </View>
+      )
+    }
+
+    const singleDigitInputComponentArr = ['k', '0', 'F', '4'].map((placeholderValue, index) => (
+      <SingleDigitInput
+        key={placeholderValue}
+        inputValue={singleDigitInputValueArr[index]}
+        inputPlaceholder={placeholderValue}
+        // tslint:disable-next-line:jsx-no-lambda
```
annakaz: Is the lambda necessary?

packages/mobile/src/send/actions.ts
```diff
@@ -29,7 +35,7 @@ export interface SendPaymentOrInviteAction {
   amount: BigNumber
   reason: string
   recipient: Recipient
-  recipientAddress?: string | null
+  recipientAddress: string | null | undefined
```
annakaz: Why not keep as optional instead of adding undefined?

packages/mobile/src/send/reducers.ts
```diff
@@ -42,6 +74,33 @@ export const sendReducer = (
...
     case Actions.STORE_LATEST_IN_RECENTS:
       return storeLatestRecentReducer(state, action.recipient)
+    case Actions.VALIDATE_RECIPIENT_ADDRESS:
+      return {
+        ...state,
+        isValidRecipient: false,
+      }
+    case Actions.VALIDATE_RECIPIENT_ADDRESS_SUCCESS:
+      const { manuallyValidatedE164NumberToAddress } = state
+      // overwrite the previous mapping every time a new one is validated
```
annakaz: Small nit: capitalize

packages/mobile/src/send/saga.ts
```diff
@@ -169,12 +184,53 @@ function* sendPaymentOrInviteSaga({
...
+  recipient,
+}: ValidateRecipientAddressAction) {
+  Logger.debug(TAG, 'Starting Recipient Address Validation')
+  try {
+    const userAddress = yield select(userAddressSelector)
+    const e164NumberToAddress = yield select(e164NumberToAddressSelector)
+    const { e164PhoneNumber } = recipient
+
+    // should never happen, if it does it means there's an error passing through recipient object
+    if (!e164PhoneNumber) {
```
annakaz: Is this not already checked by the type signature?

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+import { Recipient, RecipientKind } from 'src/recipients/recipient'
+import {
+  ConfirmationInput,
+  ManuallyValidatedE164NumberToAddress,
+  TransactionData,
+} from 'src/send/reducers'
+import Logger from 'src/utils/Logger'
+
+export const formatDisplayName = (displayName: string) => {
+  if (displayName !== 'Mobile #') {
```
annakaz: Is 'Mobile #' hardcoded for some empty case? Can you add a comment explaining the "your contact" case below

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+  if (!addresses) {
+    return addresses
+  }
+
+  if (addresses.length === 0) {
+    throw new Error('Phone addresses array should never be empty')
+  }
+
+  if (addresses.length > 1) {
+    Logger.warn('Number mapped to multiple addresses, need to disambiguate')
```
annakaz: Include tag

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+    const validatedAddress = manuallyValidatedE164NumberToAddress[e164Number]
+    if (!validatedAddress) {
+      throw new Error(
```
annakaz: Is this error caught somewhere?

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+  manuallyValidatedE164NumberToAddress: ManuallyValidatedE164NumberToAddress
+): ConfirmationInput => {
+  const { recipient } = transactionData
+  let recipientAddress: string | null | undefined
+
+  if (recipient.kind === RecipientKind.QrCode || recipient.kind === RecipientKind.Address) {
+    recipientAddress = recipient.address
+  } else {
+    if (!recipient.e164PhoneNumber) {
+      throw new Error('Phone number missing')
```
annakaz: Can you handle this through typing by forcing an e164 phone number to exist on recipient when it's expected?

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+export const getVerificationStatus = (
+  recipient: Recipient,
+  e164NumberToAddress: E164NumberToAddressType
+): RecipientVerificationStatus => {
+  let recipientVerificationStatus
+  if (recipient.kind === RecipientKind.QrCode || recipient.kind === RecipientKind.Address) {
+    recipientVerificationStatus = RecipientVerificationStatus.VERIFIED
+  } else {
+    if (!recipient.e164PhoneNumber) {
+      throw new Error('Phone number missing')
```
annakaz: Same as above can this be handled with types instead?

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  return recipientVerificationStatus
+}
+
+export const getFeeType = (
+  recipientVerificationStatus: RecipientVerificationStatus
+): FeeType | null => {
+  switch (recipientVerificationStatus) {
+    case RecipientVerificationStatus.UNKNOWN:
+      return null
```
annakaz, line 51: If feeType is null then no fee is estimated in SendAmount, is that the expected behavior or should there be a default when it's unknown?

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+    case RecipientVerificationStatus.UNVERIFIED:
+      return FeeType.INVITE
+    case RecipientVerificationStatus.VERIFIED:
+      return FeeType.SEND
+  }
+}
+
+// Given addresses can be added and deleted we can't rely on changes in length to signal address changes
+// Not sure if address order im array is consistent so not assuming it
+const checkIfItemsOfChildArrAreSubsetOfParentArr = (parentArr: string[], childArr: string[]) => {
```
annakaz: Can use sets instead of sorting?

annakaz: Nit: remove checkIf in function name

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+  for (let i = 0; i < childArrSorted.length; i += 1) {
+    if (parentArrSorted[i] !== childArrSorted[i]) {
+      return false
+    }
+  }
+
+  return true
+}
+
+const checkIfNewAddressesAdded = (
```
annakaz: nit: remove checkIf in function name

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+  oldAddresses: string[] | null | undefined,
+  newAddresses: string[]
+) => {
+  if (oldAddresses && checkIfItemsOfChildArrAreSubsetOfParentArr(oldAddresses, newAddresses)) {
```
annakaz: Is the oldAddresses check necessary? `checkIfItemsOfChildArrAreSubsetOfParentArr` could cover the empty case, then could just return `!(checkIfItemsOfChildArrAreSubsetOfParentArr(oldAddresses, newAddresses))`

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+    return false
+  }
+
+  return true
+}
+
+const checkIfLast4DigitsAreUnique = (addressArr: string[]) => {
```
annakaz: Nit: I would find this helper functions easier to read if checkIf were removed from the beginning of the function names, but it's a matter of personal preference

packages/mobile/src/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
annakaz: I don't think this should be added? As the celo-monorepo root yarn.lock is what's used

packages/mobile/test/schemas.ts
```diff
@@ -291,6 +291,153 @@ export const v1Schema = {
   },
 }
 
+export const v2Schema = {
```
annakaz: This schema migration should only include changes from v1 schema. Here is an example of a migration: https://github.com/celo-org/celo-monorepo/blob/5c4f07edfb57f5161ff2622b3bdffe0ed7bbb515/packages/mobile/test/schemas.ts#L181

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+}
+
+export const getAddressFromPhoneNumber = (
+  e164Number: string,
+  e164NumberToAddress: E164NumberToAddressType,
+  manuallyValidatedE164NumberToAddress: ManuallyValidatedE164NumberToAddress
+): string | null | undefined => {
+  const addresses = e164NumberToAddress[e164Number]
+
+  if (!addresses) {
```
annakaz: Is this the unverified case?

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+    const validatedAddress = manuallyValidatedE164NumberToAddress[e164Number]
+    if (!validatedAddress) {
+      throw new Error(
+        'Multiple addresses but none were manually validated. Should have routed through secure send.'
+      )
+    }
+
+    return validatedAddress
+  }
+  // Normal verified case, return the first address
```
annakaz: Can you add comments like this for all the cases in this function

packages/mobile/src/send/saga.ts
```diff
@@ -169,12 +184,53 @@ function* sendPaymentOrInviteSaga({
...
+  fullAddressValidationRequired,
+  recipient,
+}: ValidateRecipientAddressAction) {
+  Logger.debug(TAG, 'Starting Recipient Address Validation')
+  try {
+    const userAddress = yield select(userAddressSelector)
+    const e164NumberToAddress = yield select(e164NumberToAddressSelector)
+    const { e164PhoneNumber } = recipient
+
+    // should never happen, if it does it means there's an error passing through recipient object
```
annakaz: Nit: capitalize

packages/mobile/src/send/saga.ts
```diff
@@ -169,12 +184,53 @@ function* sendPaymentOrInviteSaga({
...
+    if (!e164PhoneNumber) {
+      throw Error('There is no phone number associated with this recipient')
+    }
+
+    const possibleRecievingAddresses = e164NumberToAddress[e164PhoneNumber]
+    // should never happen since secure send is initiated due to there being several possible addresses
```
annakaz: Nit: capitalize

packages/mobile/src/send/saga.ts
```diff
@@ -169,12 +184,53 @@ function* sendPaymentOrInviteSaga({
...
+}: ValidateRecipientAddressAction) {
+  Logger.debug(TAG, 'Starting Recipient Address Validation')
+  try {
+    const userAddress = yield select(userAddressSelector)
+    const e164NumberToAddress = yield select(e164NumberToAddressSelector)
+    const { e164PhoneNumber } = recipient
+
+    // should never happen, if it does it means there's an error passing through recipient object
+    if (!e164PhoneNumber) {
+      throw Error('There is no phone number associated with this recipient')
```
annakaz: Can this be checked with types instead

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,267 @@
...
+  )
+
+  if (!targetAddress) {
+    throw Error(ErrorMessages.ADDRESS_VALIDATION_NO_MATCH)
+  }
+
+  return targetAddress
+}
+
+export const checkForValidationErrorsAndReturnMatch = (
```
annakaz: Nit: rename to validateAndReturnMatch

packages/mobile/src/send/utils.ts
annakaz: Nit: rename to newAddressesAdded ?

packages/mobile/src/send/ValidateRecipientAccount.tsx
annakaz: Can you do this without the lambda function?

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -0,0 +1,301 @@
...
+            inputValue={inputValue}
+            inputPlaceholder={'0xf1b1d5a6e7728g309c4a025k122d71ad75a61976'}
+            onInputChange={this.onInputChange}
+            shouldShowClipboard={this.shouldShowClipboard}
+          />
+        </View>
+      )
+    }
+
+    const singleDigitInputComponentArr = ['k', '0', 'F', '4'].map((placeholderValue, index) => (
```
annakaz: Can you clarify why there is a hardcoded value here?

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -0,0 +1,301 @@
...
+          <Text style={styles.body}>
+            {t('confirmAccountNumber.2b', {
+              displayName,
+            })}
+          </Text>
+          <Text style={styles.codeHeader}>{t('accountInputHeaderB')}</Text>
+          <CodeRow
+            status={CodeRowStatus.INPUTTING}
+            inputValue={inputValue}
+            inputPlaceholder={'0xf1b1d5a6e7728g309c4a025k122d71ad75a61976'}
```
annakaz: Why is a hardcoded placeholder needed here? If it is, can you make these placeholders constants above so it's clear it's intentional?

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -157,24 +138,23 @@ export class SendAmount extends React.Component<Props, State> {
...
-    if (recipient.kind === RecipientKind.QrCode || recipient.kind === RecipientKind.Address) {
-      // Skip for QR codes or Addresses
-      return
-    }
-    if (!recipient.e164PhoneNumber) {
-      throw new Error('Missing recipient e164Number')
+  fetchLatestPhoneAddressesAndRecipientVerificationStatus = () => {
+    const { recipient } = this.props
+    // Skip phone number fetch for QR codes or Addresses
+    if (recipient.kind !== RecipientKind.QrCode && recipient.kind !== RecipientKind.Address) {
```
annakaz: Instead of the ! could you list which RecipientKinds would need a phone number?

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -157,24 +138,23 @@ export class SendAmount extends React.Component<Props, State> {
...
+      if (!recipient.e164PhoneNumber) {
+        throw new Error('Missing recipient e164Number')
```
annakaz: Can this be checked through types

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -22,25 +24,57 @@ interface OwnProps {
...
+    this.fetchLatestPhoneAddressesAndRecipientVerificationStatus()
+  }
+
+  fetchLatestPhoneAddressesAndRecipientVerificationStatus = () => {
+    const { requester: recipient } = this.props
+
+    // Skip phone number fetch for QR codes or Addresses
+    if (recipient.kind !== RecipientKind.QrCode && recipient.kind !== RecipientKind.Address) {
```
annakaz: I would find this easier to read if it was checking whether the RecipientKind matched one that required a phone number rather than skipping it for ones that do not

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -96,11 +104,24 @@ function* updateRecipientsCache(
   yield put(setRecipientCache({ ...e164NumberToRecipients, ...otherRecipients }))
 }
 
-export function* fetchPhoneAddresses({ e164Number }: FetchPhoneAddressesAction) {
+export function* fetchPhoneAddressesAndCheckIfRecipientValidationRequired({
```
annakaz: Could this name be shortened at all

packages/mobile/src/qrcode/utils.ts
tarikbellamine: Nope, there is no use legitimate use case where this should happen. Every time the scan is for secure send, transactionData should be passed through as well. Is there a way to link two optional variables in typescript?

packages/mobile/src/qrcode/utils.ts
tarikbellamine: Nope, there is no use case where this should happen. The issue stems from having a "rolled up" Recipient type that we use kinda loosely and not being able to type check navigation props, which is how recipient data is largely passed around.

I believe what @cmcewen is working on right now will unlock our ability to be better about this moving forward. I'll change how this error is handled to make it explicit that it's a Recipient type thing.

packages/mobile/src/send/utils.ts
tarikbellamine: I believe that is the expected behavior. RecipientVerificationStatus should only be unknown when the mapping is in the process of updating, otherwise it should always be either Verified or Unverified. It's impossible to send a transaction with an Unknown verification status.

packages/mobile/src/send/saga.ts
tarikbellamine: Same issue as above. I'll change how this error is handled to make it explicit that it's a Recipient type thing.

packages/mobile/src/send/utils.ts
tarikbellamine: Yup, this is either unverified (null value) or unknown (undefined). Unknown should only happen momentarily, if at all, while addresses are being updated.

packages/mobile/src/send/utils.ts
tarikbellamine: Good suggestion. Done.

packages/mobile/src/send/saga.ts
tarikbellamine: Made the error a Recipient one instead so it should be easy to typecheck when we have that option.

packages/mobile/src/send/SendAmount.tsx
tarikbellamine: Good suggestion, much less confusing that way

packages/mobile/src/qrcode/utils.ts
tarikbellamine: Thought of a way around this!

packages/mobile/src/send/ValidateRecipientAccount.tsx
tarikbellamine: Yup, good suggestion

packages/mobile/src/send/utils.ts
tarikbellamine: I think it's necessary to cover the case where oldAddresses is null or undefined (i.e., when there were no old verified addresses)

packages/mobile/src/components/SingleDigitInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+  inputValue,
+  inputPlaceholder,
+  onInputChange,
+  shouldShowClipboard,
+}: Props) {
+  return (
+    <TextInput
+      value={inputValue}
+      placeholder={inputPlaceholder}
+      shouldShowClipboard={shouldShowClipboard}
```
i1skn: Is there the case we would want to have `shouldShowClipboard` equals true here?

packages/mobile/src/qrcode/utils.ts
```diff
@@ -48,9 +54,13 @@ export async function shareSVGImage(svg: SVG) {
 export function* handleBarcode(
   barcode: QrCode,
   addressToE164Number: AddressToE164NumberType,
-  recipientCache: NumberToRecipient
+  recipientCache: NumberToRecipient,
+  e164NumberToAddress: E164NumberToAddressType,
+  scanIsForSecureSend?: true,
```
i1skn: Should here be `boolean` instead?

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -206,89 +186,78 @@ export class SendAmount extends React.Component<Props, State> {
...
+    this.setState({ reason })
   }
 
   onSend = () => {
+    const {
+      recipientVerificationStatus,
+      manualAddressValidationRequired,
+      fullValidationRequired,
+    } = this.props
+    console.log('Recipient Verification Status: ', recipientVerificationStatus)
```
i1skn: nit: `console.log`

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -0,0 +1,301 @@
...
+import { Screens } from 'src/navigator/Screens'
+import { Recipient } from 'src/recipients/recipient'
+import { RootState } from 'src/redux/reducers'
+import { validateRecipientAddress } from 'src/send/actions'
+import { TransactionData } from 'src/send/reducers'
+
+type Navigation = NavigationInjectedProps['navigation']
+
+interface OwnProps {
+  navigation: Navigation
```
i1skn: nit: you can use `NavigationInjectedProps` instead. We usually put unique to a component props inside `OwnProps`

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -0,0 +1,301 @@
...
+  validateRecipientAddress: typeof validateRecipientAddress
+}
+
+const mapDispatchToProps = {
+  validateRecipientAddress,
+}
+
+const mapStateToProps = (state: RootState, ownProps: NavigationInjectedProps): StateProps => {
+  const { navigation } = ownProps
+  const transactionData = navigation.getParam('transactionData')
```
i1skn: not a big deal, but can we move `navigation.getParam` outside of `mapStateToProps` cause they are not related to `redux` store?

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+  state: State = {
+    inputValue: '',
+    singleDigitInputValueArr: [],
+    isModalVisible: false,
+  }
+
+  componentDidUpdate = async () => {
+    const { isValidRecipient, isPaymentRequest, transactionData } = this.props
+    if (isValidRecipient) {
```
i1skn: nit: check here that previous `isValidRecipient` was `false` and only then do redirect, otherwise we can trigger it multiple times when any of the props were updated.

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -66,13 +64,23 @@ const mapDispatchToProps = {
   completePaymentRequest,
 }
 
-const mapStateToProps = (state: RootState): StateProps => {
+const mapStateToProps = (state: RootState, ownProps: NavigationInjectedProps): StateProps => {
+  const { navigation } = ownProps
+  const transactionData = navigation.getParam('transactionData')
```
i1skn: ditto

i1skn (commented):
Really great work done here 💯 ! See some minor comments I've left.

packages/mobile/src/components/SingleDigitInput.tsx
tarikbellamine: I don't think so. The user is being asked to enter in just the last 4 digits of the address and we are asking them to do this confirmation over phone or in person, so don't think copy/paste is something we want to encourage. What do you think?

packages/mobile/src/qrcode/utils.ts
tarikbellamine: The way I am using it, it's only ever true or undefined. I could make this a required boolean but given SecureSend is a bit of an edge case, I didn't want to require the more common path to have to be explicit about passing in a `false` property. What do you think?

packages/mobile/src/send/ValidateRecipientAccount.tsx
tarikbellamine: good call

packages/mobile/src/send/ValidateRecipientAccount.tsx
tarikbellamine: Yup, good point

packages/mobile/src/send/ValidateRecipientAccount.tsx
tarikbellamine: That's a good point, but I kind of prefer to leave it in. Mind if I open this up to discussion?

tarikbellamine: here's the thread if you want to join in: https://celo-org.slack.com/archives/CL7BVQPHB/p1589486122236100

packages/mobile/src/send/SendConfirmation.tsx
tarikbellamine: leaving as is for now: https://celo-org.slack.com/archives/CL7BVQPHB/p1589486122236100

packages/mobile/src/send/ValidateRecipientAccount.tsx
tarikbellamine: Kind of? I think this is a rare instance where it actually makes sense. I considered creating 4 separate state variables but that didn't seem to be a better solution. Open to suggestions for how I could do it without the lambda

packages/mobile/locales/en-US/global.json
```diff
@@ -116,5 +118,12 @@
   "message": "Message",
   "addressLookupFailure": "Failed to look up phone number address",
   "saltFetchFailure": "Failed to retrieve phone number secret",
-  "saltQuotaExceededError": "Your phone number lookup quota is used up"
+  "saltQuotaExceededError": "Your phone number lookup quota is used up",
+  "addressValidationNoMatch": "The account information entered does not match any associated with your contact",
```
jmrossy, line 157: These are only used in the Send flow right? putting them under a `addressValidation` key in sendFLow7.json may make more sense. In ErrorMessages you can access it by specifying the namespace first. E.g. `sendflow7:addressvalidation.noMatch`

packages/mobile/src/components/SingleDigitInput.tsx
jmrossy: The `shouldShowClipboard` properly only applies when using a TextInput wrapped with the `withTextInputPasteAware` HOC. Suggest removing

packages/mobile/src/identity/actions.ts
```diff
@@ -17,7 +17,8 @@ export enum Actions {
   COMPLETE_ATTESTATION_CODE = 'IDENTITY/COMPLETE_ATTESTATION_CODE',
   UPDATE_E164_PHONE_NUMBER_ADDRESSES = 'IDENTITY/UPDATE_E164_PHONE_NUMBER_ADDRESSES',
   UPDATE_E164_PHONE_NUMBER_SALT = 'IDENTITY/UPDATE_E164_PHONE_NUMBER_SALT',
-  FETCH_PHONE_ADDRESSES = 'IDENTITY/FETCH_PHONE_ADDRESSES',
+  FETCH_ADDRESSES_AND_VALIDATION_STATUS = 'IDENTITY/FETCH_ADDRESSES_AND_VALIDATION_STATUS',
+  STORE_RECIPIENT_VERIFICATION_STATUS = 'IDENTITY/STORE_RECIPIENT_VERIFICATION_STATUS',
```
jmrossy: Where is this used?

packages/mobile/src/web3/reducer.ts
```diff
@@ -91,3 +92,5 @@ export const reducer = (
       return state
   }
 }
+
+export const userAddressSelector = (state: RootState) => state.web3.account
```
jmrossy: This is a dupe of the existing currentAccountSelector: 
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/web3/selectors.ts#L3

packages/mobile/test/schemas.ts
```diff
@@ -291,6 +291,15 @@ export const v1Schema = {
...
```
jmrossy: Don't create a new schema version when you're only adding properties. We would only do that when we make backwards incompatible changes

packages/mobile/test/utils.ts
```diff
@@ -83,6 +83,17 @@ export function createMockNavigationProp(params: any): NavigationScreenProp<any>
   }
 }
 
+export function createMockNavigationProp2(params: any): NavigationScreenProp<any> {
```
jmrossy: Can you just replace the old one instead of adding this second? Or at least make the differentiating more clear

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -96,11 +104,15 @@ function* updateRecipientsCache(
...
 
-export function* fetchPhoneAddresses({ e164Number }: FetchPhoneAddressesAction) {
+export function* fetchAddressesAndValidateSaga({ e164Number }: FetchAddressesAndValidateAction) {
   try {
-    Logger.debug(TAG + '@fetchPhoneAddresses', `Fetching addresses for number`)
-    // Clear existing entries for those numbers so our mapping consumers
-    // know new status is pending.
+    Logger.debug(TAG + '@fetchAddressesAndValidate', `Fetching addresses for number`)
+    const oldE164NumberToAddress = yield select(e164NumberToAddressSelector)
+    const oldAddresses = oldE164NumberToAddress[e164Number]
```
jmrossy: Tip: these 3 lines could be:
`const oldAddresses = oldE164NumberToAddress[e164Number] || []`

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -124,11 +136,26 @@ export function* fetchPhoneAddresses({ e164Number }: FetchPhoneAddressesAction)
       addresses.map((a) => (addressToE164NumberUpdates[a] = e164Number))
     }
 
+    const userAddress = yield select(userAddressSelector)
```
jmrossy: See other comment, please replace all `userAddressSelector` uses with the existing `currentAccountSelector`

packages/mobile/src/send/reducers.ts
```diff
@@ -1,18 +1,50 @@
...
+}
+
+export interface ConfirmationInput {
+  recipient: Recipient
+  amount: BigNumber
+  reason: string
+  recipientAddress: string | null | undefined
+  type: TokenTransactionType
+  firebasePendingRequestUid?: string | null
+}
```
jmrossy: I don't think these belong here

packages/mobile/src/send/reducers.ts
```diff
@@ -1,18 +1,50 @@
...
+
+export interface SecureSendPhoneNumberMapping {
+  [e164Number: string]: {
+    address: string | undefined
+    addressValidationRequired: boolean
+    fullValidationRequired: boolean
```
jmrossy: What's the difference between these two fields? Could one be implied by the fact that address may be undefined, and thereby save on storage space?

packages/mobile/src/send/reducers.ts
```diff
@@ -42,7 +74,43 @@ export const sendReducer = (
...
+        ...state,
+        isValidRecipient: true,
+        secureSendPhoneNumberMapping,
+      }
+    case Actions.VALIDATE_RECIPIENT_ADDRESS_FAILURE:
+      return {
+        ...state,
+        isValidRecipient: false,
+      }
+    case Actions.SECURE_SEND_REQUIRED:
```
jmrossy: Nit: Actions names should start with an action verb

packages/mobile/src/send/reducers.ts
```diff
@@ -1,18 +1,50 @@
...
+    addressValidationRequired: boolean
+    fullValidationRequired: boolean
+  }
+}
+
 export interface State {
   isSending: boolean
   recentRecipients: Recipient[]
+  isValidRecipient: boolean
+  secureSendPhoneNumberMapping: SecureSendPhoneNumberMapping
```
jmrossy: I think the identity reducer is a better place for this. It's coupled more closely to contact mapping than to the send flow IMO. Keep in mind this will eventually affect other parts of the app

packages/mobile/src/send/reducers.ts
```diff
@@ -42,7 +74,43 @@ export const sendReducer = (
...
+        secureSendPhoneNumberMapping,
+      }
+    case Actions.VALIDATE_RECIPIENT_ADDRESS_FAILURE:
+      return {
+        ...state,
+        isValidRecipient: false,
+      }
+    case Actions.SECURE_SEND_REQUIRED:
+      const newSecureSendMapping = state.secureSendPhoneNumberMapping
+      newSecureSendMapping[action.e164Number] = {
```
jmrossy: This is breaking the redux immutability pattern. You should be creating a new object for `SecureSendMapping` instead of just grabbing the old one. `dot-prop-immutable` lib can help with this.

packages/mobile/src/send/actions.ts
```diff
@@ -17,6 +19,17 @@ export enum Actions {
   SEND_PAYMENT_OR_INVITE = 'SEND/SEND_PAYMENT_OR_INVITE',
   SEND_PAYMENT_OR_INVITE_SUCCESS = 'SEND/SEND_PAYMENT_OR_INVITE_SUCCESS',
   SEND_PAYMENT_OR_INVITE_FAILURE = 'SEND/SEND_PAYMENT_OR_INVITE_FAILURE',
+  VALIDATE_RECIPIENT_ADDRESS = 'SEND/VALIDATE_RECIPIENT_ADDRESS',
+  VALIDATE_RECIPIENT_ADDRESS_SUCCESS = 'SEND/VALIDATE_RECIPIENT_ADDRESS_SUCCESS',
+  VALIDATE_RECIPIENT_ADDRESS_FAILURE = 'SEND/VALIDATE_RECIPIENT_ADDRESS_FAILURE',
+  SECURE_SEND_REQUIRED = 'SEND/SECURE_SEND_REQUIRED',
+}
+
+export interface HandleBarcodeDetected {
```
jmrossy: Should be `HandleBarcodeDetectedAction`

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,288 @@
...
+  // Default displayName is "Mobile #" when no contact is saved
+  // To make the copy nicer, we make the displayName "your contact" isntead
+  if (displayName === 'Mobile #') {
+    return { displayName: 'your contact', displayNameCapitalized: 'Your contract' }
+  }
+
+  return { displayName, displayNameCapitalized: displayName }
+}
+
+export const getAddressFromPhoneNumber = (
```
jmrossy: Please move these back. They aren't utils for the send flow. They're utils for access identity related data, which are used by places in code other than the send flow

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,288 @@
...
+  e164Number: string,
+  e164NumberToAddress: E164NumberToAddressType,
+  secureSendPhoneNumberMapping: SecureSendPhoneNumberMapping
+): string | null | undefined => {
+  const TAG = 'send/utils/getAddressFromPhoneNumber'
```
jmrossy: we use TAG to represent the file name, not the method name usually

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,288 @@
+import { ErrorMessages } from 'src/app/ErrorMessages'
+import { FeeType } from 'src/fees/actions'
+import { E164NumberToAddressType, RecipientVerificationStatus } from 'src/identity/reducer'
+import { Recipient, RecipientKind } from 'src/recipients/recipient'
+import { ConfirmationInput, SecureSendPhoneNumberMapping, TransactionData } from 'src/send/reducers'
+import Logger from 'src/utils/Logger'
+
+export const formatDisplayName = (displayName: string) => {
```
jmrossy: Consider moving to `formatting.ts`

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -160,57 +187,3 @@ function* getAddresses(e164Number: string, attestationsWrapper: AttestationsWrap
 
 const isValidNon0Address = (address: string) =>
   typeof address === 'string' && isValidAddress(address) && !new BigNumber(address).isZero()
-
-export function getAddressFromPhoneNumber(
```
jmrossy, line 168: See other comment, please move these back here

packages/mobile/src/send/saga.ts
```diff
@@ -169,12 +188,52 @@ function* sendPaymentOrInviteSaga({
...
+
+    const validatedAddress = validateAndReturnMatch(
+      userInputOfFullAddressOrLastFourDigits,
+      possibleRecievingAddresses,
+      userAddress,
+      fullAddressValidationRequired
+    )
+    yield put(validateRecipientAddressSuccess(e164PhoneNumber, validatedAddress))
+  } catch (error) {
+    Logger.error(TAG, 'validateRecipientAddressSaga/Address validation error: ', error.message)
```
jmrossy: Logger.error can take the error as the third param, no need to strip out `message`

packages/mobile/src/send/saga.ts
```diff
@@ -169,12 +188,52 @@ function* sendPaymentOrInviteSaga({
...
+  Logger.debug(TAG, 'Starting Recipient Address Validation')
+  try {
+    if (!recipient.e164PhoneNumber) {
+      throw Error(`Invalid recipient type for Secure Send: ${recipient.kind}`)
+    }
+
+    const userAddress = yield select(userAddressSelector)
+    const e164NumberToAddress = yield select(e164NumberToAddressSelector)
+    const { e164PhoneNumber } = recipient
+    const possibleRecievingAddresses = e164NumberToAddress[e164PhoneNumber]
```
jmrossy: Use the `getAdressFrom....` helpers where you can to help cover edge cases

jmrossy (changes requested):
Nice first pass but I think we'll need some revisions here. I've reviewed about half. Please ping when when you've addressed the feedback and I'll look through the rest :)

packages/mobile/src/identity/actions.ts
tarikbellamine: good catch. no longer used.

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: That wasn't what I was originally thinking because `oldE164NumberToAddress[e164Number]` could evaluate to `null` or `undefined` as well. But, I could evaluate those to equal `[]` instead and handle it differently downstream, which is cleaner

packages/mobile/src/send/reducers.ts
tarikbellamine: I believe we need both. `address` is `undefined` until validation is successful so its value can't be used as a leading indicator.

true + true  = full address validation needed
true + false = last 4 digits of address validation needed
false + false = no validation needed

packages/mobile/locales/en-US/global.json
tarikbellamine: The issue I ran into is that `showError` [forces use of the global namespace](https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/alert/actions.ts#L40). I can either leave as is or switch to using `showAlert` and move everything from `global` into `sendFlow7`. What do you think?

As an FYI - I don't think past decisions should necessarily dictate the future but looks like there are quite a few other instances of flow-specific error messages included in `global`

packages/mobile/src/send/utils.ts
tarikbellamine: I am open to moving this wherever, but think it's worth pointing out that `getAddressFromPhoneNumber` is only used in `getConfirmationInput`, which is only used in `SendConfirmation.tsx` and `PaymentRequestConfirmation.tsx`

More broadly, `getAddressFromPhoneNumber` as it is designed probably shouldn't be used outside of the context of preparing to send a transaction. Attempting to use it outside of this context is kind of worthless because we have not yet gathered the data required to disambiguate a phone number that maps to multiple addresses, which is where Secure Send comes in.

packages/mobile/src/send/saga.ts
tarikbellamine: This is an example where using `getAddressFromPhoneNumber` wouldn't work because there are multiple addresses mapped to a phone number and we have not yet asked the user to validate the "correct" one.

packages/mobile/src/send/reducers.ts
tarikbellamine: Where else do you think this could come into play? See comments below about why I think using this outside of Secure Send is an issue.

packages/mobile/test/utils.ts
tarikbellamine: I wanted to replace it but it's a bit of a can of worms because `createMockNavigationProp` looks wrong to me and it is used very widely. Maybe I am missing something...can you take a look?

The only difference in `createMockNavigationProp2` is that I am passing the `property` through the `getParam` method so that I can grab the specific property rather than whole object of parameters.

packages/mobile/src/send/reducers.ts
tarikbellamine: I could make `fullValidiationRequired` optional if we want to optimize space (effectively replacing `false` with `undefined`). What do you think?

packages/mobile/test/utils.ts
tarikbellamine: Took a closer look and I am pretty sure `createMockNavigationProp` doesn't work how it's intended. Fixing it would require fixing ~12 different test files, majority of which are not related to Secure Send. Do you think it's worth taking on in this PR or opening a separate issue for?

packages/mobile/locales/en-US/global.json
jmrossy: Hmm, it shouldn't be forcing global. If you check ErrorMessage.ts you'll see that many include a namespace, which we should be doing more of going forward.
I see that line you linked to was changed only 3 days ago. Maybe @etuleu's recent commit broke the i18n namespacing for error. @etuleu can you check this please?

packages/mobile/src/send/reducers.ts
jmrossy: How about an enum representing the diff possible states instead of two flags?

packages/mobile/src/send/utils.ts
jmrossy: Valid points, thanks! Not sure why I thought there may be other consumers of this atm. 

My concern is mainly to prevent identity related complexity from leaking to multiple places in the app. Keeping it confined to the `identity` folder seems more sustainable to me. Also I find it's bets to avoid putting meaningful logic into `utils` files which become a dumping ground over time.

What's the issue with having it be in `identity` even if send flow is the only consumer for now? Just an idea but we could have this structured as a saga that navigates to the Secure Send screen as necessary, like what I've done for phone privacy here: https://github.com/celo-org/celo-monorepo/pull/3777/files#diff-2f75b310de275b594e6658f9d0eb23bfR215

packages/mobile/test/utils.ts
jmrossy: Fair enough. How about you rename this to `createMockNavigationProp` and the other to `createMockNavigationPropDeprecated` and leave a comment to clarify?

packages/mobile/src/send/saga.ts
jmrossy: I see, agreed :)

packages/mobile/src/send/reducers.ts
jmrossy: I guess it's more a matter of opinion really but the name 'Secure Send' is just marketing. It's really 'Attestation contract mapping disambiguation'. All of the other code dealing with the Attestation contract, verification, and the creation of number<->address mappings lives in the identity folder.

packages/mobile/src/send/saga.ts
```diff
@@ -169,12 +187,52 @@ function* sendPaymentOrInviteSaga({
...
+    const validatedAddress = validateAndReturnMatch(
+      userInputOfFullAddressOrLastFourDigits,
+      possibleRecievingAddresses,
+      userAddress,
+      fullAddressValidationRequired
+    )
+    yield put(validateRecipientAddressSuccess(e164PhoneNumber, validatedAddress))
+  } catch (error) {
+    Logger.error(TAG, 'validateRecipientAddressSaga/Address validation error: ', error)
+    yield put(showError(error.message))
```
jmrossy: Note, if error.message isn't in `ErrorMessages` this may show garbage. See here for an example of how to handle that:
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/identity/verification.ts#L198

packages/mobile/src/send/actions.ts
```diff
@@ -87,3 +134,36 @@ export const sendPaymentOrInviteSuccess = (): SendPaymentOrInviteSuccessAction =
...
+  type: Actions.VALIDATE_RECIPIENT_ADDRESS_SUCCESS,
+  e164Number,
+  validatedAddress,
+})
+
+export const validateRecipientAddressFailure = (): ValidateRecipientAddressFailureAction => ({
+  type: Actions.VALIDATE_RECIPIENT_ADDRESS_FAILURE,
+})
+
+export const secureSendRequired = (
```
jmrossy: Nit: looks like you renamed the action but not the action create + action type

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,269 @@
...
+import { TransactionDataInput } from 'src/send/SendAmount'
+import { ConfirmationInput } from 'src/send/SendConfirmation'
+import Logger from 'src/utils/Logger'
+
+export const getAddressFromPhoneNumber = (
+  e164Number: string,
+  e164NumberToAddress: E164NumberToAddressType,
+  secureSendPhoneNumberMapping: SecureSendPhoneNumberMapping
+): string | null | undefined => {
+  const TAG = 'send/utils'
```
jmrossy: Nit move to top of file?

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,269 @@
...
+        )
+      }
+
+      return validatedAddress
+    }
+
+    // Normal case when there is only one address in the mapping
+    return addresses[0]
+  } catch (error) {
+    Logger.error(TAG, error.message)
```
jmrossy: What should happen on error? Should we propagate it? Return null?

packages/mobile/src/send/utils.ts
```diff
@@ -0,0 +1,269 @@
...
+  newAddresses: string[] | null,
+  e164Number: string,
+  secureSendPhoneNumberMapping: SecureSendPhoneNumberMapping
+) => {
+  const validatedAddress =
+    secureSendPhoneNumberMapping[e164Number] || secureSendPhoneNumberMapping[e164Number].address
+  return newAddresses && newAddresses.length > 1 && !validatedAddress
+}
+
+export const checkIfValidationRequired = (
```
jmrossy: How about we move all this stuff to `identity/secureSend.ts`?

packages/mobile/test/utils.ts
tarikbellamine: Works for me!

packages/mobile/src/send/reducers.ts
tarikbellamine: Yeah, that's probably a better solution. Will do.

packages/mobile/src/send/utils.ts
tarikbellamine: Sounds good

packages/mobile/src/send/utils.ts
tarikbellamine: That's a good question and I am not sure what the answer is. This is a true error in the sense that it really should never happen. If it does, that means we are not using the function correctly (i.e., it's being called before the SecureSend flow has collected the data).

Both `null` (unverified) and `undefined` (loading) have a reserved meaning here. Given this is not really meant to be used at the top of the stack, I don't necessarily think it warrants a user facing error.  How do we feel about making this a breaking error? If so, what's the best way to implement that?

packages/mobile/src/send/utils.ts
tarikbellamine: Cool that all makes sense to me. I see your point about grouping similar logic and agree. No issue with moving it (will do that), just wanted to make sure we were on the same page about how it was being utilized.

Pairing the address fetching and the navigation together into one saga was something I considered and actually liked more from a technical perspective as the logic is more contained. The hangup for me was that there would be a 1-2 second lag (due to the attestation lookup) between the user action and the eventual navigation. Did the PGPNP implementation have a similar trade off?

packages/mobile/src/send/saga.ts
tarikbellamine: Nice, I like that

packages/mobile/src/send/utils.ts
jmrossy: I don't think PGPNP had that lag. Can you clarify what caused that 1-2 second lag and why it doesn't happen with the current implementation? Def not saying you need to change it btw, just so I understand.

packages/mobile/src/send/utils.ts
jmrossy: Yeah if it's a true error and we don't have a clean way to recover, just let it propagate and our React error boundary will end up catching it

packages/mobile/src/send/utils.ts
tarikbellamine: I believe the lag is just due to the async nature of the call we are making. In the current implementation, we are making this call on `componentDidMount` when the user hits the `SendAmount` screen. That way, by the time they press "send", we already know which direction we need to route them so they'll never really notice any type of loading time.

packages/mobile/locales/en-US/global.json
tarikbellamine: Checked the commit history and looks like this has been an issue since initial commit? I first noticed this ~3 weeks ago and mentioned it in that large thread of observations I posted to the wallet channel.

I don't know much about`i18n`... does it have built in functionality that parses namespaces? If the fix isn't a quick one, I am more inclined to address in a separate issue so we can get this merged.

packages/mobile/src/redux/migrations.ts
```diff
@@ -41,4 +41,14 @@ export const migrations = {
       },
     }
   },
+  3: (state: any) => {
```
jmrossy: This just adds new state right? No need for a new migration. Also if you did add one, you'd have to also rev the version in redux/store.ts and preferably also create a test

packages/mobile/locales/en-US/global.json
jmrossy: Yep totally reasonable

packages/mobile/src/identity/contactMapping.test.ts
```diff
@@ -90,4 +108,90 @@ describe('Fetch Addresses Saga', () => {
       )
       .run()
   })
+
+  it('requires SecureSend with partial verification when a new adddress is added and last 4 digits are unique', async () => {
+    const contractKit = await getContractKitOutsideGenerator()
```
jmrossy, line 113: Why is this here? Seems like you don't use it and pass in a mockContractKit instead, which seems right to me

packages/mobile/src/identity/contactMapping.test.ts
```diff
@@ -90,4 +108,90 @@ describe('Fetch Addresses Saga', () => {
...
+            [mockAccount.toLowerCase()]: mockE164Number,
+            [mockAccount2.toLowerCase()]: mockE164Number,
+          }
+        )
+      )
+      .run()
+  })
+
+  it('requires SecureSend with full verification when a new adddress is added and last 4 digits are not unique', async () => {
+    const contractKit = await getContractKitOutsideGenerator()
```
jmrossy, line 156: ditto

packages/mobile/src/identity/secureSend.ts
```diff
@@ -0,0 +1,161 @@
+import { ErrorMessages } from 'src/app/ErrorMessages'
+import { AddressValidationType, SecureSendPhoneNumberMapping } from 'src/identity/reducer'
+import { Recipient } from 'src/recipients/recipient'
+import Logger from 'src/utils/Logger'
+
+const TAG = 'identity/secureSend'
+
+// Given addresses can be added and deleted we can't rely on changes in length to signal address changes
+// Not sure if address order in array is consistent so not assuming it
+const newAddressesAdded = (oldAddresses: string[], newAddresses: string[]) => {
```
jmrossy: This is nbd but the convention we try to keep it is arrow functions in classes (like for handlers) but classic functions for things defined top-level in modules. This facilities recursion and keeps consistency with generators which can't be expressed as lamdbas

packages/mobile/src/qrcode/utils.ts
```diff
@@ -64,6 +69,35 @@ export function* handleBarcode(
...
+    if (secureSendTxData) {
+      if (!secureSendTxData.recipient.e164PhoneNumber) {
+        throw Error(`Invalid recipient type for Secure Send: ${secureSendTxData.recipient.kind}`)
+      }
+
+      const userScannedAddress = data.address.toLowerCase()
+      const { e164PhoneNumber } = secureSendTxData.recipient
+      const possibleRecievingAddresses = e164NumberToAddress[e164PhoneNumber]
+      // This should never happen. Secure Send is triggered when there are
+      // mutliple addrresses for a given phone number
```
jmrossy: typo `mutliple addrresses`

packages/mobile/src/qrcode/utils.ts
```diff
@@ -64,6 +69,35 @@ export function* handleBarcode(
...
+      )
+      if (!possibleRecievingAddressesFormatted.includes(userScannedAddress)) {
+        yield put(showError(ErrorMessages.QR_FAILED_INVALID_RECIPIENT))
+        return
+      }
+
+      yield put(validateRecipientAddressSuccess(e164PhoneNumber, userScannedAddress))
+    }
+  } catch (error) {
+    Logger.error(TAG + '@handleBarcode', `Error with Secure Send: `, error)
```
jmrossy: Should this saga proceed if there's an error?

packages/mobile/src/qrcode/utils.ts
```diff
@@ -64,6 +69,35 @@ export function* handleBarcode(
     yield put(showError(ErrorMessages.QR_FAILED_INVALID_ADDRESS))
     return
   }
+  try {
```
jmrossy: Nit: would be nice if this stuff was in a separate function if possible

packages/mobile/src/recipients/recipient.ts
```diff
@@ -140,19 +124,29 @@ export function getRecipientFromAddress(
   return e164PhoneNumber ? recipientCache[e164PhoneNumber] : undefined
 }
 
-export function getRecipientVerificationStatus(
+export const getRecipientVerificationStatus = (
```
jmrossy: why change this?

packages/mobile/src/recipients/recipient.ts
```diff
@@ -140,19 +124,29 @@ export function getRecipientFromAddress(
...
   e164NumberToAddress: E164NumberToAddressType
-): RecipientVerificationStatus {
+): RecipientVerificationStatus => {
   if (recipient.kind === RecipientKind.QrCode || recipient.kind === RecipientKind.Address) {
     return RecipientVerificationStatus.VERIFIED
   }
 
-  if (!recipient.e164PhoneNumber) {
-    throw new Error('No recipient e164Number found')
+  if (recipient.e164PhoneNumber) {
```
jmrossy: Really small nit but have you heard of the zen-of-python reduce nesting principle? TLDR, look for opportunities to reduce nesting by inverting `if`s. Instead of the bulk of logic being in this if, check the inverse and return UNKNOWN in that case

packages/mobile/src/send/actions.ts
```diff
@@ -54,9 +63,15 @@ export const storeLatestInRecents = (recipient: Recipient): StoreLatestInRecents
   recipient,
 })
 
-export const handleBarcodeDetected = (data: QrCode) => ({
+export const handleBarcodeDetected = (
+  data: QrCode,
+  scanIsForSecureSend?: true,
+  transactionData?: TransactionDataInput
+): HandleBarcodeDetectedAction => ({
```
jmrossy, line 70: 💯

jmrossy (approved):
Approved with some optional comments, please have a look before merging. Congrats on this big change!

packages/mobile/src/identity/contactMapping.test.ts
tarikbellamine: I think I need it. I pass in `mockAttestationWrapper` but still use `contractKit` in the provide method.

packages/mobile/src/identity/secureSend.ts
tarikbellamine: Gotcha. Will change all the functions in this file and keep that in mind moving forward.

packages/mobile/src/qrcode/utils.ts
tarikbellamine: I actually don't think there should be a catch here at all. Thanks for pointing out

packages/mobile/src/qrcode/utils.ts
tarikbellamine: Good suggestion. Done.

packages/mobile/src/recipients/recipient.ts
tarikbellamine: Will implement. Useful tactic, thanks!

packages/mobile/src/recipients/recipient.ts
tarikbellamine: Part of a larger refactor...basically folded 3 functions across 3 different files into 1. Think it's a net positive. Will change back to non-arrow function as per earlier style comment.

PR #3792, opened 2020-05-12
[cli] Proposals for proxy

mrsmkl:
### Description

Looks like there's no support for generating this from the json yet, now it should be handled the same way as other methods.

### Tested

Only manual testing.


mrsmkl:
Figured out how to do it without changes.

PR #3794, opened 2020-05-12
Set blockperiod when running nodes with celotool

asaj:
### Description

Needed for https://github.com/celo-org/celo-blockchain/pull/985, otherwise in e2e tests the validator nodes have a block period of 1 and the non-validator nodes have a block period of 5. 

### Other changes

None

### Tested

e2e tests
### Related issues

None
### Backwards compatibility

Yes


packages/celotool/src/lib/geth.ts
```diff
@@ -915,6 +915,8 @@ export async function startGeth(
     'extip:127.0.0.1',
     '--allow-insecure-unlock', // geth1.9 to use http w/unlocking
     '--gcmode=archive', // Needed to retrieve historical state
+    '--istanbul.blockperiod',
+    blocktime.toString(),
```
m-chrzan, line 919: If `gethConfig.genesisConfig.blocktime` is undefined, do we still want the default to be 1s (set on line 890)?

m-chrzan (approved):

packages/celotool/src/lib/geth.ts
asaj: Yeah I think it's fine for now since this is mostly used for e2e tests

Issue #3660, opened 2020-04-29
New Wallet Design: Account Key


nityas:
Parent Issue: #3645 
Figma: https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=2995%3A33668

PR #3795, opened 2020-05-13
redesign account key (Part 1 - Core Flow)

aaronmgdr:
### Description

new Account Key Flow (formerly known as backup Key)

Coming in Part 2 & 3

1. Account Key Guide
2. back button hide when no words selected
3. Modal that pops up when you go to cancel
4. The Screenshot Prevention. 

### Other changes

add svg version of rings Logo
add Switch.v2 

### Tested

mostly already tested with some no longer relevant rests removed. 

Screen Shot 2020-05-13 at 2 01 05 PM
Screen Shot 2020-05-13 at 2 01 43 PM
Screen Shot 2020-05-13 at 2 01 50 PM
Screen Shot 2020-05-13 at 2 01 57 PM
Screen Shot 2020-05-13 at 2 03 03 PM

### Related issues

- part of #3660 

### Backwards compatibility

removes references to social backup


aaronmgdr:
TODO spanish and remove unused translations

packages/mobile/src/backup/BackupIntroduction.tsx
```diff
@@ -58,170 +52,104 @@ class BackupIntroduction extends React.Component<Props> {
...
-    paddingBottom: 30,
+    paddingHorizontal: Sizes.THICK,
     justifyContent: 'center',
   },
-  logo: {
-    alignSelf: 'center',
-    marginBottom: 30,
+  postSetupContainer: {
+    flex: 1,
+    paddingTop: Sizes.THICK,
```
jeanregisser: I was wondering at first what was the purpose of the `Sizes` enum given the generic name.
Is its main use for spacing things? If so we might want to rename it to `Spacing`.

packages/mobile/src/backup/BackupIntroduction.tsx
```diff
@@ -58,170 +52,104 @@ class BackupIntroduction extends React.Component<Props> {
...
+    paddingHorizontal: Sizes.REG,
```
jeanregisser: nit: how about using a full name `REGULAR` instead of an abbreviation?

packages/mobile/src/backup/BackupPhrase.tsx
```diff
@@ -49,9 +51,10 @@ const mapStateToProps = (state: RootState): StateProps => {
 class BackupPhrase extends React.Component<Props, State> {
   // TODO(Rossy): Show modal when cancelling if backup flow incomplete
   static navigationOptions = () => ({
-    ...headerWithBackButton,
+    ...headerWithCancelButton,
+    headerLeft: <CancelButton color={colors.gray4} />,
+    headerRight: <HeaderRight />,
```
jeanregisser: Shouldn't we have a title here? I saw `Account Key` in figma.

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -27,7 +26,18 @@ const TAG = 'backup/BackupQuiz'
...
+
+export enum Mode {
+  entering,
+  checking,
+  failed,
+}
+
+// TODO Add states for, checking, failed, success
 interface State {
+  is: Mode
```
jeanregisser: nit: Can we use a more descriptive variable name?

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -27,7 +26,18 @@ const TAG = 'backup/BackupQuiz'
...
 const MNEMONIC_BUTTONS_TO_DISPLAY = 6
 
+// miliseconds
+const CHECKING_DURATION = 2.6 * 1000
+
+export enum Mode {
+  entering,
+  checking,
+  failed,
+}
```
jeanregisser: nit: looks like we're inconstantly naming enum values. I've seen UPPERCASE, PascalCase, and lowercased here.
@celo-org/wallet can we pick one? 
I'd vote for PascalCase.

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -144,168 +156,178 @@ export class BackupQuiz extends React.Component<Props, State> {
...
+
 export default componentWithAnalytics(
   connect<{}, DispatchProps, {}, RootState>(null, { setBackupCompleted, showError })(
     withTranslation(Namespaces.backupKeyFlow6)(BackupQuiz)
   )
 )
 
-const styles = StyleSheet.create({
+export const styles = StyleSheet.create({
```
jeanregisser: Do we need to export all of these styles?

packages/mobile/src/backup/DelayButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const dispatchToProps = {
+  setBackupAsDelayed: setBackupDelayed,
+}
+
+export default connect<DelayStateProps, DelayDispatchProps, {}, RootState>(
+  mapStateToHeaderProps,
+  dispatchToProps
+)(function DelayButton(props: DelayDispatchProps & DelayStateProps) {
```
jeanregisser: nit: would be cleaner with our `useSelector` hook.

packages/mobile/src/backup/utils.ts
```diff
@@ -105,6 +107,17 @@ export async function getStoredMnemonic(): Promise<string | null> {
...
+export function useAccountKey() {
+  const [accountKey, setAccountKey] = useState<string | null>(null)
+
+  useAsync(async () => {
+    const mnemonic = accountKey || (await getStoredMnemonic())
+    setAccountKey(mnemonic)
+  }, [])
+
+  return accountKey
+}
```
jeanregisser: I think `useState` is unnecessary here, can be simplified to:
```suggestion
export function useAccountKey() {
  asyncAccountKey = useAsync(getStoredMnemonic, [])
  return asyncAccountKey.result;
}
```

packages/mobile/src/components/CancelButton.v2.tsx
```diff
@@ -1,3 +1,4 @@
+import { colorsEnum } from '@celo/react-components/styles/colors.v2'
```
jeanregisser, line 1: nit: should just be named `colors`

packages/react-components/components/Switch.v2.tsx
```diff
@@ -0,0 +1,19 @@
...
+    <RNSwitch
+      trackColor={switchTrackColors}
+      thumbColor={colors.gray2}
+      ios_backgroundColor={colors.gray3}
+      {...props}
+    />
+  )
+}
+
+const switchTrackColors = {
```
jeanregisser: nit: we usually use UPPERCASE for constants like this.

jeanregisser (commented):
Very nice work! 👍💯 
Left a couple of small comments

packages/mobile/src/backup/BackupIntroduction.tsx
```diff
@@ -58,170 +52,104 @@ class BackupIntroduction extends React.Component<Props> {
...
+}
+
+function AccountKeyIntro({ onPrimaryPress }: AccountKeyStartProps) {
+  const { t } = useTranslation(Namespaces.backupKeyFlow6)
+  return (
+    <ScrollView contentContainerStyle={styles.introContainer}>
+      <Logo height={32} />
+      <Text style={styles.h1}>{t('introTitle')}</Text>
+      <Text style={styles.body}>{t('introBody')}</Text>
+      <Button text={t('introPrimaryAction')} onPress={onPrimaryPress} />
```
i1skn, line 85: Can we this Button out of `ScrollView`, so on the small screen it would still be visible without scrolling?

packages/mobile/src/account/Account.tsx
```diff
@@ -88,8 +90,12 @@ export class Account extends React.Component<Props, State> {
     navigate(Screens.Profile)
   }
 
-  goToBackupScreen() {
-    navigate(Screens.BackupIntroduction)
+  goToBackupScreen = () => {
+    if (this.props.backupCompleted) {
+      navigateProtected(Screens.BackupIntroduction)
```
i1skn, line 95: I see, that `Screens.BackupIntroduction` has here two different purposes depend on `backupCompleted`. Do you think we can split this screen into two and keep `Screens.BackupIntroduction` only for Introduction?

packages/mobile/src/backup/BackupIntroduction.tsx
```diff
@@ -58,170 +52,104 @@ class BackupIntroduction extends React.Component<Props> {
...
+  return (
+    <ScrollView contentContainerStyle={styles.postSetupContainer}>
+      <View>
+        <Text style={fontStyles.h2}>{t('introTitle')}</Text>
+        <View style={styles.keyArea}>
+          <Text style={fontStyles.large}>{accountKey}</Text>
+        </View>
+        <Text style={styles.postSetupBody}>{t('postSetupBody')}</Text>
+      </View>
+      <View style={styles.postSetupCTA}>
```
i1skn, line 104: ditto about moving button out of `ScrollView`

packages/mobile/src/backup/BackupQuiz.tsx
i1skn: For enums I'd vote for PascalCase as well

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -44,19 +54,23 @@ interface DispatchProps {
 type Props = WithTranslation & DispatchProps & NavigationInjectedProps
 
 export class BackupQuiz extends React.Component<Props, State> {
-  static navigationOptions = () => ({
-    ...headerWithBackButton,
+  static navigationOptions = ({ navigation }: NavigationInjectedProps) => ({
+    ...headerWithCancelButton, // TODO switch to v2 once Ivans PR is in
+    headerRight: <DeleteWord onPressBackspace={navigation.getParam('onPressBackspace')} />,
```
i1skn: we could add a check here that `navigation.getParam('onPressBackspace')` is not null, at least until TS can do it for us instead.

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -111,28 +119,32 @@ export class BackupQuiz extends React.Component<Props, State> {
...
-  onPressSubmit = () => {
+  afterCheck = () => {
     const { userChosenWords, mnemonicLength } = this.state
     const mnemonic = this.getMnemonicFromNavProps()
-    if (
-      userChosenWords.length === mnemonicLength &&
-      userChosenWords.map((w) => w.word).join(' ') === mnemonic
-    ) {
+    const lengthsMatch = userChosenWords.length === mnemonicLength
+
```
i1skn, line 135: nit: we can have here `const contentMatch = userChosenWords.map((w) => w.word).join(' ') === mnemonic` and later `if (lengthsMatch && contentMatch)`. Not a big deal, but would improve readability a bit.

packages/mobile/src/backup/DelayButton.tsx
i1skn: can we also have `DelayButton` outside of `connect` function? Not a big deal, but to be consistent with rest of the codebase and also, if would need to define some statics on it later.

packages/mobile/src/backup/DelayButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const { backupTooLate, backupDelayedTime, setBackupAsDelayed } = props
+
+  const onPressDelay = React.useCallback(() => {
+    setBackupAsDelayed()
+    CeloAnalytics.track(CustomEventNames.delay_backup)
+    navigateBack()
+  }, [])
```
i1skn: should we have `[setBackupAsDelayed]` here?

i1skn (commented):
Great work on this one 💪 !!! Left just some minor comments

aaronmgdr:
> TODO spanish and remove unused translations

Done

packages/mobile/src/backup/BackupQuiz.tsx
aaronmgdr: nope

packages/mobile/src/backup/BackupQuiz.tsx
aaronmgdr: oh yeah this is something i've been infamously inconsistent on.

packages/mobile/src/backup/DelayButton.tsx
aaronmgdr: good call on the useSelector. I have refactored

packages/mobile/src/backup/BackupIntroduction.tsx
aaronmgdr: Given that the design has all of this being centered vertically together I dont think we should

packages/mobile/src/backup/BackupQuiz.tsx
aaronmgdr: hmm if i do that then it will always do this mapping even if lengths tont match which seems wasteful

aaronmgdr: I extracted out a function. best of both worlds!

packages/mobile/src/account/Account.tsx
aaronmgdr: My thought is that while yes it could be two, its better to draw a line somewhere as to what we will and will not change in the redesign. for me that includes not creating new screens if any exisiting one is hooked up like we need. 

Without this Line I would have also renamed this component, and the i18n key file at leas, maybe more.

packages/mobile/src/backup/BackupIntroduction.tsx
aaronmgdr: done

packages/mobile/src/backup/BackupIntroduction.tsx
aaronmgdr: confirmed with Cory it should not be taken out

packages/mobile/src/backup/BackupQuiz.tsx
aaronmgdr: done

packages/mobile/src/components/CancelButton.v2.tsx
aaronmgdr: will handle in a separate PR ;)

packages/mobile/src/backup/DelayButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const mapStateToHeaderProps = (state: RootState): DelayStateProps => {
+  return {
+    backupTooLate: isBackupTooLate(state),
+    backupDelayedTime: state.account.backupDelayedTime,
+  }
+}
+
+export default function DelayButton() {
+  const { backupTooLate, backupDelayedTime } = useSelector(mapStateToHeaderProps)
```
jeanregisser: nit: `useSelector` has a subtle difference compared to `connect`, it uses `===` comparison to decide if it needs re-rendering. Here since the `mapStateToProps` is returning a new object it will always re-render whenever the redux state changes. See https://react-redux.js.org/api/hooks#equality-comparisons-and-updates

```suggestion
  const { backupTooLate, backupDelayedTime } = useSelector(mapStateToHeaderProps, shallowEqual)
```

packages/mobile/src/backup/DelayButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+import TopBarButton from 'src/navigator/TopBarButton.v2'
+import { RootState } from 'src/redux/reducers'
+import { isBackupTooLate } from 'src/redux/selectors'
+
+interface DelayStateProps {
+  backupTooLate: boolean
+  backupDelayedTime: number
+}
+
+const mapStateToHeaderProps = (state: RootState): DelayStateProps => {
```
jeanregisser: nit: rename `mapStateToHeaderProps` -> `mapStateToProps`

packages/react-components/styles/styles.v2.ts
```diff
@@ -1,8 +1,8 @@
...
-export enum Sizes {
-  SMALLEST = BASE_UNIT * 1.5,
-  SMALLER = BASE_UNIT,
-  REG = BASE_UNIT * 2,
-  THICK = BASE_UNIT * 3,
+export enum Spacing {
+  Smallest8 = BASE_UNIT,
+  Small12 = BASE_UNIT * 1.5,
+  Regular16 = BASE_UNIT * 2,
+  Thick24 = BASE_UNIT * 3,
```
jeanregisser, line 7: 👍

jeanregisser (approved):
Nice!!

PR #3798, opened 2020-05-13
Added field timestamp to the testnet config

kevjue:
### Description

This PR adds the ability to specify the `TIMESTAMP` genesis field in the testnet env config.  The value of that field will be the time when the network will start consensus.

### Other changes

None

### Tested

Ran a testnet and ensured it respected the `TIMESTAMP` field.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes


tkporter (approved):
lgtm!

PR #3799, opened 2020-05-13
Update React Navigation

cmcewen:
### Description

This updates react-navigation to the latest version. It brings nice features like hooks, legitimate typing of params, and hopefully an easier structure overall. Still need a bit of work here, but just as an fyi in case you will be doing some work here.

We also have this `componentWithAnalytics` function that is just worth getting rid of imo. It stops our type checking, and we can get the current nav state using the navigator, so no reason to do it on a screen level.


packages/mobile/src/account/DollarEducation.tsx
```diff
@@ -21,7 +20,7 @@ export class DollarEducation extends React.Component<Props> {
   goToSend = () => {
     this.props.setEducationCompleted()
     CeloAnalytics.track(CustomEventNames.send_dollar_nux)
-    navigate(Stacks.SendStack)
+    navigate(Screens.Send)
```
aaronmgdr, line 23: I noticed on some screens you changed to `this.props.navigation.naviate` but not here?

packages/verifier/src/components/TransactionReview/VerificationReviewScreen.tsx
```diff
@@ -25,7 +25,7 @@ export class VerificationReviewScreen extends React.PureComponent<Props> {
   static navigationOptions = { header: null }
 
   getNavigationProps = (): ReviewProps => {
-    const { timestamp, value, phoneNumbers } = this.props.navigation.getParam('reviewProps')
+    const { timestamp, value, phoneNumbers } = this.props.navigation.w('reviewProps')
```
aaronmgdr: checkin just  `w`

packages/mobile/src/account/__snapshots__/DataSaver.test.tsx.snap
```diff
@@ -108,7 +108,7 @@ exports[`DataSaver renders correctly 1`] = `
       }
       swipeThreshold={100}
       transparent={true}
-      visible={true}
+      visible={false}
```
aaronmgdr, line 111: This may indicate a problem

packages/mobile/src/firebase/firebase.ts
```diff
@@ -14,9 +14,10 @@ import { currentLanguageSelector } from 'src/app/reducers'
 import { FIREBASE_ENABLED } from 'src/config'
 import { WritePaymentRequest } from 'src/firebase/actions'
 import { handleNotification } from 'src/firebase/notifications'
-import { navigate } from 'src/navigator/NavigationService'
-import { Screens } from 'src/navigator/Screens'
+import { navigateHome } from 'src/navigator/NavigationService'
 import Logger from 'src/utils/Logger'
+// We can't combine the 2 imports otherwise it only imports the type and fails at runtime
+// tslint:disable-next-line: no-duplicate-imports
```
aaronmgdr: I dont see a next line?

aaronmgdr (commented):
left a few comments. WOW what a good one. nice ratio of adds to deletes!

packages/mobile/src/account/DollarEducation.tsx
cmcewen: Yeah I just didn't bother changing it everywhere but I do prefer that

packages/verifier/src/components/TransactionReview/VerificationReviewScreen.tsx
cmcewen: didn't mean to change the verifier so reverted 😄

PR #3800, opened 2020-05-13
Validator Explorer feedback and improvements

Pedro-vk:
### Description

_Validator Explorer feedback and improvements_

### Tested

_On Chrome_

### Related issues

- Related to https://github.com/celo-org/celo-labs/issues/489


packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -267,16 +276,26 @@ class ValidatorsList extends React.PureComponent<Props, State> {
...
+  isPinned({ address }: any) {
+    const list = (localStorage.getItem(localStoragePinnedKey) || '').split(',') || []
+    return +list.includes(address)
   }
 
   render() {
     const { expanded, orderBy, orderAsc } = this.state
     const { data } = this.props
     const validatorGroups = !data ? ([] as CeloGroup[]) : this.sortData(this.cleanData(data))
+    const onPinned = () => this.setState({ update: Math.random() } as any)
```
aaronmgdr: can you make this method on the class?

packages/web/src/dev/ValidatorsListRow.tsx
```diff
@@ -12,6 +12,8 @@ import { cutAddress, formatNumber } from 'src/utils/utils'
 const unknownGroupName = 'Unnamed Group'
 const unknownValidatorName = 'Unnamed Validator'
 
+const localStoragePinnedKey = 'pinnedValidators'
```
aaronmgdr: this is the same key as in the other file right? 

might be good to define once and import

packages/web/src/dev/ValidatorsListRow.tsx
```diff
@@ -91,19 +100,46 @@ class ValidatorsListRow extends React.PureComponent<Props & I18nProps, State> {
     this.removeDocumentListener()
   }
 
+  isPinned(toggle?: boolean) {
+    const { address } = this.props.group
+    let list = (localStorage.getItem(localStoragePinnedKey) || '').split(',') || []
+    let isPinned = list.includes(address)
+    if (toggle) {
+      list = list.filter((_) => _ !== address)
```
aaronmgdr: could we  only filter when `isPinned` is true  ?

packages/web/src/dev/ValidatorsListRow.tsx
```diff
@@ -91,19 +100,46 @@ class ValidatorsListRow extends React.PureComponent<Props & I18nProps, State> {
...
-    const { tooltip } = this.state
+    const { tooltip, isPinned } = this.state
     const stopPropagation = (event: any) => {
       event.stopPropagation()
     }
     const toggleTooltip = (event: any) => {
       stopPropagation(event)
       this.setState({ tooltip: !tooltip })
     }
+    const togglePinned: any = (event: any) => {
```
aaronmgdr: nit: remove `any`

packages/web/src/dev/ValidatorsListRow.tsx
```diff
@@ -57,21 +61,26 @@ export interface CeloGroup {
...
     tooltip: false,
+    isPinned: false,
   }
   tooltipRef = React.createRef<any>()
   removeDocumentListener: () => void
 
   constructor(...args) {
     super(...(args as [any]))
 
+    this.state.isPinned = this.isPinned()
```
aaronmgdr, line 82: do not update state directly use setState

aaronmgdr: rather than setting in constructor its better to make a call to setState in the componentDidMount function

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -267,16 +276,26 @@ class ValidatorsList extends React.PureComponent<Props, State> {
       .sort((a, b) => b.id - a.id)
       .sort((a, b) => compare(dAccessor(a), dAccessor(b)))
       .sort((a, b) => dir * compare(accessor(a), accessor(b)))
+      .sort((a, b) => this.isPinned(b) - this.isPinned(a))
+  }
+
+  isPinned({ address }: any) {
```
aaronmgdr, line 281: nit: type as `{addres: string}`

packages/web/src/dev/ValidatorsListRow.tsx
Pedro-vk: It produces an error because technically the view doesn't emit the event (but it does). Another solution is removing the `any` and start using a `div`

packages/web/src/dev/ValidatorsList.tsx
Pedro-vk: Not possible at all, because it is used in the sorting that doesn't have enough typing information (and I didn't want to over typed it, in the end, the data is an array of key/values).

packages/web/src/dev/ValidatorsListRow.tsx
```diff
@@ -87,23 +94,55 @@ class ValidatorsListRow extends React.PureComponent<Props & I18nProps, State> {
...
+    const { tooltip, isPinned } = this.state
+    const stopPropagation = (event) => {
       event.stopPropagation()
     }
-    const toggleTooltip = (event: any) => {
+    const toggleTooltip = (event) => {
       stopPropagation(event)
       this.setState({ tooltip: !tooltip })
     }
+    const togglePinned: any = (event) => {
```
aaronmgdr: im having trouble understanding why this and the  `toggleTooltip` above are defined in render? 

same with the `stopPropagation`.

Pedro-vk: I can move it to a method, but I'll need to create a new function inside render or bind the context to the method to be able to use `this` context

aaronmgdr: if you want a method to be alwasy lexicaly bound to its class you just need to define it an arrow function on the class like this

```
class ValidatorsListRow extends React.PureComponent<Props & I18nProps, State> {

togglePinned = () => {}
```

i recommend changing all methods on the class to be defined as such

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -264,19 +273,32 @@ class ValidatorsList extends React.PureComponent<Props, State> {
...
+    const list = (localStorage.getItem(localStoragePinnedKey) || '').split(',') || []
+    return +list.includes(address)
+  }
+
+  onPinned() {
```
aaronmgdr, line 286:
```suggestion
  onPinned = () =>  {
```

really all our class instance methods should be defined in this way.

packages/web/src/dev/ValidatorsListRow.tsx
```diff
@@ -87,23 +94,57 @@ class ValidatorsListRow extends React.PureComponent<Props & I18nProps, State> {
...
+        list = list.filter((_) => _ !== address)
+      }
+      isPinned = !isPinned
+      localStorage.setItem(localStoragePinnedKey, list.join(','))
+      this.props.onPinned()
+    }
+    return isPinned
+  }
+
+  stopPropagation = (event) => {
```
aaronmgdr, line 122: yes!

packages/web/src/dev/ValidatorsList.tsx
aaronmgdr: ah ok

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -264,19 +273,32 @@ class ValidatorsList extends React.PureComponent<Props, State> {
...
+    this.setState({ update: Math.random() } as any)
```
aaronmgdr, line 287: I dont quite follow this? 

it doesnt look like update is a property of State, and why a random value?

Pedro-vk: I the solution that I found to force to sort the data again after pinning a group. The group emits an event and then is forced to sort again.

aaronmgdr (approved):

PR #3802, opened 2020-05-13
Filter kube label metrics for AKS prometheus

tkporter:
### Description

Stackdriver allows a maximum of 10 custom labels. `kube-state-metrics` has some metrics of the form `"kube_.+_labels"` that provide the labels of k8s resources as metric labels. If some k8s resources have too many labels, this results in a bunch of errors when the sidecar tries to send metrics to Stackdriver. This is the easiest fix I found, as far as I can tell there's no way to filter out metrics by how many labels they have

### Other changes

Fixed an issue with the cluster name

### Tested

Deployed prometheus on an AKS cluster

### Related issues

n/a

### Backwards compatibility

Yes, we don't rely on any `kube_.+_labels` metrics


jcortejoso (approved):
lgtm

PR #3803, opened 2020-05-13
[Wallet] Redesign - Add header components

i1skn:
### Description
* add `TopBarIconButton` and `TopBarTextButton`.
* add generic `HeaderTitleWithSubtitle` 
* refresh header with balance

### Tested

Android emulator

### Backwards compatibility

Yes

#### 1
Screenshot_1589378972
#### 2
Screenshot_1589379000
#### 3
Screenshot_1589378963
#### 4
Screenshot_1589378978
#### 5 (when top bar button is pressed)
Screenshot_1589459370


packages/mobile/src/components/BackButton.v2.tsx
```diff
@@ -0,0 +1,8 @@
+import BackButton from '@celo/react-components/components/BackButton.v2'
+import React, { useCallback } from 'react'
+import { navigateBack } from 'src/navigator/NavigationService'
+
+export default function MobileBackButton(props: { height?: number; color?: string }) {
+  const goBack = useCallback(() => navigateBack(), [navigateBack])
```
aaronmgdr: why not just pass navigateBack directly to BackButton?

packages/mobile/src/navigator/TopBarButton.v2.tsx
```diff
@@ -1,31 +1,32 @@
...
   return (
-    <BorderlessButton
-      disabled={disabled}
-      onPress={onPress}
-      testID={testID}
-      style={style ? { ...styles.text, ...style } : styles.text}
-    >
-      {children}
-    </BorderlessButton>
+    <TouchableOpacity onPress={onPress} disabled={disabled} testID={testID}>
```
aaronmgdr: why this change? 

why remove Touchable? it handles having android phones have the nice ripple.

packages/react-components/components/BackButton.v2.tsx
```diff
@@ -0,0 +1,35 @@
...
+      </TouchableOpacity>
+    </View>
+  )
+}
+
+const SIZE = 16
+
+const styles = StyleSheet.create({
+  container: {
+    paddingLeft: variables.contentPadding + 7,
```
aaronmgdr: im not sure about using these variables in new design. since everything is multiples of 8. the +7 worries me

packages/mobile/src/navigator/TopBarButton.v2.test.tsx
```diff
@@ -5,7 +5,7 @@ import TopBarButton from 'src/navigator/TopBarButton.v2'
 
 describe('TopBarButton', () => {
   it('displays children', () => {
-    const { queryByText } = render(<TopBarButton>label</TopBarButton>)
+    const { queryByText } = render(<TopBarButton title={'label'} />)
```
aaronmgdr: hmm I think we should avoid changing interfaces

packages/mobile/src/navigator/TopBarButton.v2.tsx
```diff
@@ -1,31 +1,32 @@
...
+      <Text style={textStyle ? [styles.text, textStyle] : styles.text}>{title}</Text>
+    </TouchableOpacity>
   )
 }
 
 const styles = StyleSheet.create({
   text: {
-    ...fontStyles.regular500,
```
aaronmgdr, line 28: hmm the buttons i reference in the header when i made this were def `500`

i1skn: @coreycelo did some changes in the design, so I've reflected them here.

aaronmgdr: ok cool

packages/react-components/icons/BackChevron.v2.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react'
+import Svg, { Path } from 'svgs'
+
+export interface Props {
+  height: number
+  color: string
+}
+
+export default function BackChevron({ color, height }: Props) {
```
jmrossy: Setting default static props here like we do for some other icons may be a nice addition

packages/react-components/components/BackButton.v2.tsx
jmrossy: +1, setting up new variables may be nice.
But also, I find variables for things like padding to be unhelpful. How often will we say 'reduce padding across the app by 2'. Even if we did, unlikely that most things would line up right after the var change

packages/react-components/components/BackButton.v2.tsx
```diff
@@ -0,0 +1,35 @@
...
+type Props = TouchableOpacityProps & {
+  height?: number
+  color?: string
+  testID?: string
+}
+export default function BackButton({ onPress, color, height, disabled, testID }: Props) {
+  return (
+    <View style={styles.container}>
+      <TouchableOpacity
+        hitSlop={variables.iconHitslop}
```
jmrossy: Maybe not a problem but just a note, I think hitslop only works if the parent container is sufficiently large (ie small parent will cut off hitslop area)

packages/mobile/src/navigator/TopBarButton.v2.test.tsx
jmrossy: I think it should be case-by-case. A title prop does make more sense to me than a child component

packages/mobile/src/navigator/Headers.v2.tsx
```diff
@@ -0,0 +1,117 @@
...
+        <CurrencyDisplay
+          amount={{
+            value: balance,
+            currencyCode: CURRENCIES[token].code,
+          }}
+        />{' '}
+        available
+      </Trans>
+    ) : (
+      // TODO: a null balance doesn't necessarily mean it's loading
```
jmrossy, line 93: I find the possibility of null balance creates ugly conditionals all across the app and I wonder if we'd be better off just defaulting to 0. We'd lose a signal that says 'I don't know the balance' but it would simplify many places

jmrossy (commented):
Mostly looks great!

packages/mobile/src/navigator/TopBarButton.v2.test.tsx
i1skn: I agree here with Rossy, plus it was introduced only in `v2`, so potentially will only affect our WIP branches. @aaronmgdr WDYT?

packages/mobile/src/navigator/Headers.v2.tsx
i1skn: I think, this is a bit out of the scope of this PR. But I agree, we should come up with a convenient approach how we represent unknown balance.

i1skn:
@aaronmgdr I've reworked `TopBarButton` and split it into `TopBarIconButton` and `TopBarTextButton`. I know that you have an opened PR, which relies on it and I'm happy to wait until we merge your one first.

packages/react-components/components/BackButton.v2.tsx
i1skn: I've moved this logic to `mobile` package. Regarding the padding itself I've changed it to `+6` (correlates with the design) to the standard padding we have.

Screenshot 2020-05-14 at 12 56 26

aaronmgdr (approved):

Issue #3653, opened 2020-04-29
Instrument wallet performance metrics


cmcewen:

jose-compu:
1. How many Txs are failing?
2. What is the average time of confirmation/receipt received?

jose-compu:
3. What is the Start time?
4. What is the Sync time?

jose-compu:
5. How much time to get Peers?

jose-compu:
Time To Initialize = tti

https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/app/App.tsx#L56

`CeloAnalytics.track(DefaultEventNames.appLoadTTIInMilliseconds, { tti }, true)`

tarikbellamine:
Address by #4599

PR #3810, opened 2020-05-13
[Wallet] Adding time metrics to sync and tx events, also new events.

jose-compu:
### Description
Adding time elapsed to Sync Event and Tx Events.

### Tested
Manually on Debug console.

### Related issued
Part of #3653 

### Backwards compatibility
Some event names where changed, this is not compatible with existing data on Analytics.


packages/mobile/src/transactions/saga.ts
```diff
@@ -54,18 +55,31 @@ export function* sendAndMonitorTransaction<T>(
     yield call(wrapSendTransactionWithRetry, txId, sendTxMethod)
     yield put(transactionConfirmed(txId))
 
+    txTime = Date.now() - txTime
     if (currency === CURRENCY_ENUM.GOLD) {
+      Logger.debug(
+        TAG + '@sendAndMonitorTransaction',
+        `send_gold_transaction_confirmed Performance (ms) = ` + txTime.toString()
+      )
+      CeloAnalytics.track(CustomEventNames.send_gold_transaction_confirmed, { txTime })
```
jmrossy: These seem redundant with the analytics tracking above. How about we just make currency a property of the `CustomEventNames.send_transaction_confirmed` event?

packages/mobile/src/transactions/saga.ts
```diff
@@ -54,18 +55,31 @@ export function* sendAndMonitorTransaction<T>(
...
       yield put(fetchGoldBalance())
     } else if (currency === CURRENCY_ENUM.DOLLAR) {
-      CeloAnalytics.track(CustomEventNames.send_dollar_transaction_confirmed)
+      Logger.debug(
+        TAG + '@sendAndMonitorTransaction',
+        `send_dollar_transaction_confirmed Performance (ms) = ` + txTime.toString()
+      )
+      CeloAnalytics.track(CustomEventNames.send_dollar_transaction_confirmed, { txTime })
```
jmrossy: note, you need to add txTime to the property whitelist in packages/mobile/src/analytics/constants.ts

packages/mobile/src/web3/saga.ts
```diff
@@ -77,13 +79,17 @@ export function* checkWeb3SyncProgress() {
         if (latestBlock && latestBlock.number > 0) {
           yield put(completeWeb3Sync(latestBlock.number))
           Logger.debug(TAG, 'checkWeb3SyncProgress', 'Sync is complete')
+          millisecs = Date.now() - initTime
+          CeloAnalytics.track(CustomEventNames.sync_complete, { millisecs })
```
jmrossy: This method gets called potentially many times throughout the lifetime of the app. We probably don't want to spam out events for each micro-sync right? We should find a different strategy for tracking sync time

jmrossy (commented):
Good progress and congrats on your first wallet PR!

packages/mobile/src/web3/saga.ts
jose-compu: We can do a Moving Average every N micro-sync and then post every M micro-sync. What do you think.

packages/mobile/src/transactions/saga.ts
jose-compu: Okey, got it.

packages/mobile/src/transactions/saga.ts
jose-compu: Thanks

packages/mobile/src/web3/saga.ts
jose-compu: I did it now with only one parameter:  N = M.

jose-compu: For example track() is called on start sync and once every 100 syncs. Mean and spot time elapsed since sync start is reported.

jose-compu:
For loading geth and sync registering	only a few events 1 of 100 and also moving average, for tx still all events.

packages/mobile/src/web3/saga.ts
```diff
@@ -75,6 +81,13 @@ export function* checkWeb3SyncProgress() {
 
         const latestBlock: Block = yield call(getLatestBlock)
         if (latestBlock && latestBlock.number > 0) {
+          ;[meanMillisecs, trackCount] = trackMeanMillisecs(
```
cmcewen: If I'm reading this correctly, this is the only event function we need to use trackMeanMilliseconds for, right?

packages/mobile/src/geth/geth.ts
```diff
@@ -120,9 +129,23 @@ async function initGeth() {
     }
 
     if (!(await ensureGenesisBlockWritten())) {
+      ;[meanMillisecs, trackCount] = trackMeanMillisecs(
+        initTime,
+        meanMillisecs,
+        trackCount,
+        TRACK_EVERY_GETH,
+        CustomEventNames.geth_failed_genesis_block
```
cmcewen: For these kinds of events - I'm not sure we would expect this to happen very often, so I don't think we really need to call this function, right?

packages/mobile/src/utils/time.ts
```diff
@@ -318,3 +320,29 @@ function quickFormat(timestamp: number, i18next: i18nType, formatRule: string) {
...
+
+// warning: this function has side-effects on module
+//          vars meanMillisecs and trackCount.
+export const trackMeanMillisecs = (
+  initTime: number,
+  meanMillisecs: number,
+  trackCount: number,
+  onceEvery: number,
+  eventName: CustomEventNames,
+  attachDeviceInfo: boolean = false
```
cmcewen: this param isn't used anywhere is it?

packages/mobile/src/utils/time.ts
```diff
@@ -318,3 +320,29 @@ function quickFormat(timestamp: number, i18next: i18nType, formatRule: string) {
...
 }
+
+// update moving average for time ellapsed
+export function updateMeanMillisecs(meanMillisecs: number, millisecs: number, onceEvery: number) {
+  const multiplier = meanMillisecs > 0.0 ? meanMillisecs / onceEvery : 1.0
+  meanMillisecs = meanMillisecs * (1.0 - multiplier) + multiplier * millisecs
+  return meanMillisecs
+}
+
+// warning: this function has side-effects on module
```
cmcewen: I think this abstraction could be improved so that you don't mutate external vars in the function. If you create an object in this file with keys that are the event name and values that store data, you could track meanMilliseconds and trackCount there

packages/mobile/src/web3/saga.ts
jose-compu: okey, CustomEventNames.sync_complete then.

packages/mobile/src/geth/geth.ts
jose-compu: Yep, these are more abnormal conditions.

jose-compu: Okey, good idea, then only CustomEventNames.sync_complete

packages/mobile/src/utils/time.ts
jose-compu: Okey, got it, thanks.

packages/mobile/src/utils/time.ts
jose-compu: Yes, on CustomEventNames.sync_blockchain_corruption and CustomEventNames.sync_unexpected_error.

packages/mobile/src/web3/saga.ts
jose-compu: okey, perfect -> CustomEventNames.sync_complete

tarikbellamine (changes requested):
Nice work! Left a couple of comments. It seems like you are mostly tracking duration with your timers but FYI we have timestamps included by default and can calculate the duration between any two events by subtracting the timestamps.

Also worth nothing that we have a `trackSubEvents` method that allows easy grouping of related events and will track duration between subEvents for us.

packages/mobile/src/analytics/constants.ts
```diff
@@ -47,6 +46,21 @@ export enum CustomEventNames {
...
   edit_send_invite = 'edit_send_invite',
 
+  // Geth events
+  geth_failed_genesis_block = 'geth_failed_genesis_block',
+  geth_failed_static_nodes = 'geth_failed_static_nodes',
+  geth_error_corrupt_chain = 'geth_error_corrupt_chain',
+  geth_error_already_running = 'geth_error_already_running',
+  geth_error_unexpected = 'geth_error_unexpected',
+
+  // Sunc events
```
tarikbellamine: nit: Typo - "Sync"

packages/mobile/src/analytics/constants.ts
```diff
@@ -282,6 +300,7 @@ export enum PropertyPathWhitelist {
   countryCode = 'countryCode',
   cta = 'cta',
   currentScreen = 'currentScreen',
+  ccyTicker = 'ccyTicker',
```
tarikbellamine: Can we please rename to `currencyCode`?

packages/mobile/src/analytics/constants.ts
```diff
@@ -307,7 +327,9 @@ export enum PropertyPathWhitelist {
   makerAmount = 'makerAmount',
   makerToken = 'makerToken',
   makerTokenAmount = 'makerTokenAmount',
+  meanMillisecs = 'meanMillisecs',
   method = 'method',
+  millisecs = 'millisecs',
```
tarikbellamine: nit: Can we rename this to something more descriptive such as `timeToFailure`?

packages/mobile/src/transactions/saga.ts
```diff
@@ -80,18 +81,24 @@ export function* sendAndMonitorTransaction<T>(
...
     } else {
       // Fetch both balances for exchange
       yield put(fetchGoldBalance())
       yield put(fetchDollarBalance())
     }
+    CeloAnalytics.track(CustomEventNames.send_transaction_confirmed, { millisecs, ccyTicker })
   } catch (error) {
     Logger.error(TAG + '@sendAndMonitorTransaction', `Error sending tx ${txId}`, error)
+    millisecs = Date.now() - millisecs
+    CeloAnalytics.track(CustomEventNames.send_transaction_failed, { millisecs })
```
tarikbellamine: Can we pass the currency code here as well?

packages/mobile/src/web3/saga.ts
```diff
@@ -55,10 +56,18 @@ export const SYNC_TIMEOUT = 2 * 60 * 1000 // 2 minutes
 const BLOCK_CHAIN_CORRUPTION_ERROR = "Error: CONNECTION ERROR: Couldn't connect to node on IPC."
 const SWITCH_TO_FORNO_TIMEOUT = 15000 // if syncing takes >15 secs, suggest switch to forno
 const WEB3_MONITOR_DELAY = 100
+// track sync time first time and then only one every so many events.
+const TRACK_SAMPLING_PERIOD_SYNC = 1
+
+// Tracking meanMilliseconds and count per analytics event that need sampling (too frequent).
+const performanceTracking = {
```
tarikbellamine: I believe you can use `CeloAnalytics.startTracking` + `trackSubEvents` + `stopTracking` to serve this purpose

jose-compu:
> Nice work! Left a couple of comments. It seems like you are mostly tracking duration with your timers but FYI we have timestamps included by default and can calculate the duration between any two events by subtracting the timestamps.
> 
> Also worth nothing that we have a `trackSubEvents` method that allows easy grouping of related events and will track duration between subEvents for us.

I am measuring the time with regard to the init of a task, is not between two events. I am only submitting one event per task.

packages/mobile/src/web3/saga.ts
```diff
@@ -65,10 +69,10 @@ export function* checkWeb3SyncProgress() {
 
       if (typeof syncProgress === 'boolean' && !syncProgress) {
         Logger.debug(TAG, 'checkWeb3SyncProgress', 'Sync maybe complete, checking')
-
         const latestBlock: Block = yield call(getLatestBlock)
         if (latestBlock && latestBlock.number > 0) {
           yield put(completeWeb3Sync(latestBlock.number))
+          web3SyncAnalyticsTracker.log(syncStartTime)
```
annakaz, line 75: It seems like `web3SyncAnalyticsTracker` is only used here, so the data is never submitted. Would it be possible to just remove it and use the existing CeloAnalytics?

tarikbellamine:
@timmoreton, @joigno and I have done some work on geth analytics instrumentation but it recently occurred to me that this is probably a subset of what you are working on

Issue #3645, opened 2020-04-29
New design system


i1skn:
https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=2930%3A0

- [ ] Design Kit
  - [x] colors, fonts, Buttons (#3691)
    - [x] Button
  - [x] Avatars
  - [x] Loading Spinner
  - [x] Inputs fields
    - [x] Switch
  - [x] Guide (aka Education)
  - [x] Top bar elements
     - [x] CancelButton #3772
     - [x] TopBarButton #3772
     - [x] BackButton  #3803
     - [x] Central header with Subheader #3803
  - [x] Banners  (AKA Alerts in the code)
  - [ ] Toast
  - [x] Messaging #3840
  - [x] Dialogs #3863
  - [x] Card in #3840 (A View with shadow)
  - [x] Custom Keyboard https://github.com/celo-org/celo-monorepo/pull/3879
  - [x] Lists 
     - [x] Transactions / People #3815
     - [x] Settings
     - [x] Receipt Layout (@i1skn is working on it as a part of https://github.com/celo-org/celo-monorepo/pull/3764, can extract it earlier if needed)
     - [x] On Celo Gold tab and Cash In/Out (it should be similar to one in Gold Screen - https://github.com/celo-org/celo-monorepo/pull/3764)
- [ ] Account Key (#3660)
  - [x] Core Flow
  - [x] Guide
  - [x]  back button hide when no words selected
  - [x]  Modal that pops up when you go to cancel
   - [ ] The Screenshot Prevention.
- [x] Gold (#3659)
- [x] Home
  - [x] Home Screen #3815
  - [x] Notifications #3840
  - [x] Transactions drilldowns #3910
  - [x] QR Scanning
  - [x] Send flow 1st part #3895 
  - [x] Send flow 2nd part #3904
  - [x] Secure Send flow #3904 
  - [x] Request flow #3953
  - [x] Invite flow #3953
- [x] Cash In (#3848)
- [x] Settings
- [x] Help (https://github.com/celo-org/celo-monorepo/pull/4274)
- [ ] ~~Invite (deprecated for now)~~
- [ ] Onboarding flow
- [x] Navigation
   - [x] react-navigation v5 #3799
   - [x] drawer menu #3894
- [x] Pin code screen

aaronmgdr:
This is a good start, we should just be aware that there will be things in progress and that we might complete part of these

[...]

PR #3815, opened 2020-05-14
[Wallet][ReDesign] Transaction feed and Send/Request bottom bar

jeanregisser:
### Description

This adds/updates the following elements on the Home screen with the new design:
- Transaction feed
- Send/Request bar at the bottom

I've left the settings on the right and balance at the top until we can add the drawer menu.
Also `Avatar` is the old one.

Next steps: 
- Transaction drilldown
- Notification panel
- Send/Request flow
- Animate hide/show of Send/Request bottom bar as the user scrolls

### Other changes

- Fix visual glitch in the nav bar when scrolling down at the beginning of the home screen.

### Related issues

- Epic #3645 

### Backwards compatibility

Yes.


packages/mobile/locales/en-US/walletFlow5.json
```diff
@@ -72,14 +75,28 @@
   "feedItemVerificationRewardInfo": "Verifier Reward",
   "feedItemFaucetTitle": "Celo",
   "feedItemFaucetInfo": "{{faucet}} Faucet",
-  "feedItemFaucetInfo_missingTestnet": "Faucet",
+  "feedItemFaucetInfo_noTestnet": "Faucet",
   "feedItemInviteSentTitle": "Celo",
   "feedItemInviteSentInfo": "Invited {{nameOrNumber}}",
-  "feedItemInviteSentInfo_missingInviteeDetails": "$t(inviteFlow11:inviteSent)",
+  "feedItemInviteSentInfo_noInviteeDetails": "$t(inviteFlow11:inviteSent)",
```
aaronmgdr, line 81: oh what is happening with the value here? is this some i18n magic?

jeanregisser: It's the [`context` feature of `i18next`](https://www.i18next.com/translation-function/context), it's appended to the key.

packages/mobile/src/transactions/transferFeedUtils.ts
```diff
@@ -27,11 +27,15 @@ export function getTransferFeedParams(
   recipientCache: NumberToRecipient,
   address: string,
   addressToE164Number: AddressToE164NumberType,
-  comment: string | null,
+  rawComment: string | null,
```
annakaz, line 30: 🎨

packages/mobile/src/transactions/transferFeedUtils.ts
```diff
@@ -52,45 +56,64 @@ export function getTransferFeedParams(
...
     }
-
     case TokenTransactionType.InviteReceived: {
       title = t('feedItemInviteReceivedTitle')
       info = t('feedItemInviteReceivedInfo')
       break
     }
+    case TokenTransactionType.Sent: {
+      title = t('feedItemSentTitle', {
+        context: !nameOrNumber ? 'noReceiverDetails' : null,
```
annakaz, line 79: Nit: could use an enum for the contexts to prevent typos?

annakaz (approved):
👍 LGTM!

packages/mobile/locales/en-US/walletFlow5.json
aaronmgdr: 😲

packages/mobile/src/transactions/transferFeedUtils.ts
jeanregisser: yes, managing translation strings is a bit too manual and error prone, I've mentioned [js-lingui](https://github.com/lingui/js-lingui) in the past and want to propose using it.
It would solve this kind of issue as it automatically extracts strings from the source code.
Though i18next has something similar which could be worth looking at.

packages/mobile/src/transactions/TransactionFeedItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const { type, amount, title, info, icon, status, onPress } = props
+
+  const isReceived = new BigNumber(amount.value).isPositive()
+  const isPending = status === TransactionStatus.Pending
+
+  let subtitle = info
+  if (isPending) {
+    subtitle = t('confirmingTransaction')
+  }
```
i1skn: nit: I think this would be a bit more readable
```suggestion
  const subtitle = isPending ? t('confirmingTransaction') : info
```

i1skn (approved):
Great work 💯 !

Issue #3747, opened 2020-05-06
[Wallet] Light client timeout issue fetching state on older devices


annakaz:
### Expected Behavior

Users with older Android devices can fetch balances without timing out

### Current Behavior

On older Android devices, seeing a timeout issue when fetching state. For example: `Error fetching balance:Returned error: execution aborted (timeout = 5s)`. Only happens sometimes

See logs here: https://github.com/celo-org/celo-monorepo/issues/3280#issuecomment-624618721

Note that this caused https://github.com/celo-org/celo-monorepo/issues/3280 to be observed, but that verification issue is tracking reveal-related issues. This timeout issue does not seem to be specific to verification

PR #3819, opened 2020-05-15
[Wallet] Retry verification with forno if it fails using geth

annakaz:
### Description

Verification has been occasionally failing due to a geth timeout (#3747 ). Since this does not occur in forno mode, users who experience a verification failure are prompted to retry using forno. This is only for failures that occur before requesting attestations, which requires a transaction, so users may not have sufficient balance to reverify. Behavior remains the same for failures that occur after requesting attestations. 

### Other changes

Created modal component (not yet updated with new design)

### Tested

On Android emulator, tested that modal throws when failure happens before requesting attestations, and that verification is retried in forno mode. 

Confirmed with Ankit that this addresses the verification issue he was observing, though I can't test locally as I can't reproduce his problem locally.

### Related issues

- Part of epic #3644 

### Backwards compatibility

Yes


annakaz:
Screen Shot 2020-05-14 at 4 50 54 PM
Screen Shot 2020-05-14 at 4 50 33 PM
(One other case with "enter codes" text, unchanged from before)

packages/mobile/locales/en-US/nuxVerification2.json
```diff
@@ -63,8 +57,7 @@
   },
   "missingCodesModal": {
     "header": "Missing Codes?",
-    "body":
```
annakaz, line 66: Just spacing changes here and above

packages/mobile/src/account/__snapshots__/DataSaver.test.tsx.snap
```diff
@@ -238,11 +238,11 @@ exports[`DataSaver renders correctly 1`] = `
                       "lineHeight": 18,
```
annakaz, line 738: Snapshot changed as Modal component is being used. Note that redesign has not yet been applied

packages/mobile/src/components/WarningModal.tsx
```diff
@@ -0,0 +1,80 @@
+import TextButton from '@celo/react-components/components/TextButton'
```
annakaz, line 1: Copied over from Modal originally used in DataSaver with small changes to allow for one or two buttons. Note that the redesign has not been applied- leaving that for a future PR

packages/mobile/src/verify/VerificationFailedModal.tsx
```diff
@@ -1,24 +1,30 @@
...
 }
 
 export function VerificationFailedModal(props: Props) {
   const { t } = useTranslation(Namespaces.nuxVerification2)
-  const [isDismissed, setIsDismissed] = React.useState(false)
+  const [isDismissed, setIsDismissed] = React.useState(true)
+
+  React.useEffect(() => {
+    setIsDismissed(false)
+  }, [setIsDismissed])
```
annakaz: This change was necessary to prevent a ghost modal from showing up briefly when a user reopens the Verification loading page and the modal was dismissed. This use of useEffect is like `componentDidMount()`

packages/mobile/src/verify/VerificationFailedModal.tsx
```diff
@@ -29,56 +35,54 @@ export function VerificationFailedModal(props: Props) {
     navigateHome()
   }, [props.cancelVerification])
 
+  const onRetry = React.useCallback(() => {
+    props.toggleFornoMode(true)
+    props.setRetryVerificationWithForno(false) // Only prompt retry with forno once
+    setIsDismissed(true)
+    navigate(Screens.VerificationEducationScreen)
```
annakaz, line 42: Renavigate uesrs to `VerificationEducationScreen` when they retry, so they will have to click start again. Lmk if anyone has thoughts on instead redirecting to `VerificationLoadingScreen`- this is more direct but was concerned users wouldn't know that verification had been started again

packages/mobile/src/verify/VerificationLoadingScreen.tsx
```diff
@@ -117,6 +127,10 @@ class VerificationLoadingScreen extends React.Component<Props> {
         </View>
         <VerificationFailedModal
           verificationStatus={verificationStatus}
+          retryWithForno={retryWithForno}
+          fornoMode={fornoMode}
```
annakaz, line 130: Have to keep track of forno mode to avoid prompting users who are already in forno mode

packages/mobile/src/verify/__snapshots__/VerificationLoadingScreen.test.tsx.snap
```diff
@@ -278,15 +278,15 @@ exports[`VerificationLoadingScreen renders correctly 1`] = `
         <Text
```
annakaz, line 278: Minor changes to modal because modal component is used- note that this will change when redesign is applied

packages/mobile/test/schemas.ts
```diff
@@ -291,6 +291,14 @@ export const v1Schema = {
   },
 }
 
+export const v2Schema = {
```
annakaz: Added v2 schema to avoid conflicting with some issues in v1 schema migration, which I believe are being fixed in https://github.com/celo-org/celo-monorepo/pull/3791

packages/mobile/src/verify/VerificationFailedModal.tsx
tarikbellamine: I like sending users back to `VerificationEducationScreen`

tarikbellamine (commented):
Looks pretty good to me! Two questions:
1. Based on my understanding, users will only ever get one shot at verifying in forno mode given the initial state is `true` and every call to `setRetryVerificationWithForno` passes in `false` as an argument.  Should we consider resetting the value to `true` after every non-forno verification fail so that users always have the option to try forno again?
2. It looks like we are avoiding the forno reprompt after a transaction fails to be conservative about not prompting the user to undergo another transaction they may not have a balance for. Would it be worth it to do a quick check of their balance and make the prompt conditioned on that?

annakaz:
> 1. Based on my understanding, users will only ever get one shot at verifying in forno mode given the initial state is `true` and every call to `setRetryVerificationWithForno` passes in `false` as an argument.  Should we consider resetting the value to `true` after every non-forno verification fail so that users always have the option to try forno again?

The idea here is to have users switch to forno if verification fails, and stay in forno as the observed failures (https://github.com/celo-org/celo-monorepo/issues/3747) seem to be non-verification specific (thanks for raising, added a comment clarifying this). Once a user retries with forno, they would have had to manually toggle forno off in order for them to need the prompt again, so I think it makes sense to only show the modal once 

> 2. It looks like we are avoiding the forno reprompt after a transaction fails to be conservative about not prompting the user to undergo another transaction they may not have a balance for. Would it be worth it to do a quick check of their balance and make the prompt conditioned on that?

We have seen verification fail for a few different reasons after a transaction fails, and only some of them would be fixed in forno mode. I agree if we could know that the failure was light-client specific then it would be nice to check the balance, but for now I want to avoid having users retry and see another failure- updated the comment to clarify this

cmcewen (approved):
looks good to me! would be great to get a thumbs up from nitya or corey on the UX side too

packages/mobile/src/verify/VerificationFailedModal.tsx
jmrossy: Might be nice to have this in a comment in code :)

packages/mobile/src/verify/VerificationFailedModal.tsx
jmrossy: I'm fine with that too. Verification is a complex flow. Starting it again from the beginning seems safer.

jmrossy (approved):
Lgtm 👍 
One optional comment

annakaz:
Waiting on tests from Ankit after phone number privacy changes

annakaz:
Confirmed with Ankit that this resolves the verification failure he was seeing!

PR #3820, opened 2020-05-15
Reintroducing Ripple Buttons on Android

aaronmgdr:
### Description

Back when `Button.v2` came out It used `TouchableOpacity` as i couldn't get it to keep the ripple affect inside the curves, instead it ignored them, but not in a cool way. it stoped on the square edge, which looked super tacky. 

but no more!

Now it does it the right way. rippling to the curvaceous border and no further

### Other changes

renames `colorsEnum` to `Colors`

### Tested

confirmed it looks right

### Related issues

- Epic #3645 

### Backwards compatibility

might have some merge issue with the colorsEnum change but hopefully worth it


packages/react-components/components/Button.v2.tsx
```diff
@@ -47,20 +49,21 @@ export default React.memo(function Button(props: ButtonProps) {
...
-        activeOpacity={0.75}
-        onPress={debouncedOnPress}
-        disabled={disabled}
-        style={getStyle(size, backgroundColor)}
-        testID={testID}
-      >
-        <Text
-          accessibilityLabel={accessibilityLabel}
-          style={{ ...fontStyles.regular600, color: textColor }}
+      <View style={styles.containRipple}>
```
jeanregisser: nit: we could probably combine the 2 containers `<View>` at the top.

jeanregisser (approved):
Cool! 👍

packages/react-components/components/Button.v2.tsx
aaronmgdr: nope doesnt work.

PR #3821, opened 2020-05-15
Oracle upgrade and readiness probe

codyborn:
### Description

Adds capability to upgrade Oracle nodes. To ensure that a node is fully operational before moving to the next node, we implore the readinessProbe [here](https://github.com/celo-org/celo-oracle/pull/53). Please see [this document for broader context](https://docs.google.com/document/d/1WMppUMnnb8x_-SHqVH6v7rUVwme6WIx6b-pf3K4oL9U/edit?usp=sharing).

### Tested

Validated using oracles deployed in our AKS cluster.

### Backwards compatibility

No, these changes require that the new docker image is used which contains the readiness probe.


packages/celotool/src/cmds/deploy/upgrade/oracle.ts
```diff
@@ -0,0 +1,33 @@
...
+import yargs from 'yargs'
+
+export const command = 'oracle'
+
+export const describe = 'upgrade the oracle(s) on an AKS cluster'
+
+type OracleUpgradeArgv = UpgradeArgv &
+  OracleArgv & {
+    useFullNodes: boolean
+    reset: boolean
```
tkporter: nit: if we aren't using the `--reset` flag we can take this out

packages/celotool/src/lib/oracle.ts
```diff
@@ -94,6 +98,20 @@ export async function installHelmChart(
...
+  context: OracleAzureContext,
+  useFullNodes: boolean
+) {
+  await upgradeGenericHelmChart(
+    celoEnv,
+    releaseName(celoEnv),
+    helmChartPath,
+    await helmParameters(celoEnv, context, useFullNodes)
+  )
+  return
```
tkporter: nit: we can make this function non async and just have it:
```
return upgradeGenericHelmChart(
    celoEnv,
    releaseName(celoEnv),
    helmChartPath,
    await helmParameters(celoEnv, context, useFullNodes)
  )
```

tkporter (approved):
Asked this question in the google doc, but x posting here:
You said if the readiness probe returns false too many times, k8s will restart the pod. What's the threshold for this? It could be minutes before a new pod is ready and I want to make sure this doesn't send the pod into a crash loop

codyborn:
> Asked this question in the google doc, but x posting here:
> You said if the readiness probe returns false too many times, k8s will restart the pod. What's the threshold for this? It could be minutes before a new pod is ready and I want to make sure this doesn't send the pod into a crash loop

When the readinessProbe is failing, K8s will simply wait indefinitely for it to return a success. It will halt the deployment (no new pods will be upgraded), but it will not restart the pod. Upgrade is an asynchronous process (celotooljs command returns immediately), so we should setup some utility in our deployment tooling to monitor the status of a live upgrade.

packages/celotool/src/lib/oracle.ts
codyborn: Good point. I still need to keep it async since I'm await-ing the helmParameters() call, but I updated it to just return the upgradeGenericHelmChart promise.

PR #3825, opened 2020-05-15
[wallet/azure-hsm] fix bug when converting big number to a buffer

zviadm:
### Description

Fix similar bug that was in AzureKeyClient in AzureHSMWallet class

### Tested

yarn test src/wallets


zviadm:
actually nvm, this isn't needed because public key always starts with byte: "0x04" (because it is not compressed public key), so it can never start with a zero byte, thus converting BigNumber as it was before isn't really a problem.

PR #3826, opened 2020-05-15
[Wallet] Move wallet key management from disk/forno to local geth keystore

### LINKED ISSUE #3651: Complete Wallet Audit
### Expected Behavior

Complete an audit of the wallet codebase

yorhodes:
### Description

Eliminates private key persistence outside of mobile geth keystore. Changes `fornoMode` to sign transactions with local unsynced geth client and submit to forno cluster for sync features. 

### Other changes

- Adds `RpcWallet` and `RpcSigner` to contractkit for signing with a "local" JSON rpc client with a provider pointing at a different RPC.
- Changes `fornoMode` to launch geth client with `maxPeers=0` (do not sync). Toggles of `fornoMode` respawn the geth process with this setting toggled. 

### Tested

- `RpcWallet` has tests in contractkit against ganache for import/unlock
- `RpcWallet` is mocked in mobile tests and invite/backup/transfer saga tests pass

### Related issues

- Opens #4183
- Closes #3651

### Backwards compatibility


packages/mobile/src/geth/saga.ts
```diff
@@ -192,3 +212,167 @@ export function* gethSagaIfNecessary() {
...
+    const accountAddress = yield call(assignAccountFromPrivateKey, privateKey)
+    if (!accountAddress) {
+      throw new Error('Failed to assign account from private key')
+    }
+
+    yield call(setKey, 'mnemonic', mnemonic)
+
+    return accountAddress
+  } catch (error) {
+    // Capturing error in sentry for now as we debug backup key issue
```
cmcewen: We should probably make sure the error doesn't contain the key

packages/mobile/src/geth/saga.ts
```diff
@@ -192,3 +212,167 @@ export function* gethSagaIfNecessary() {
...
+    yield call(waitForGethConnectivity)
+    account = yield call(gethWallet.addAccount, key, pincode)
+    Logger.debug(
+      TAG + '@addAccountToGethKeystore',
+      `Successfully imported raw key for account ${account}`
+    )
+    yield put(setAccountInGethKeystore(account))
+  } catch (e) {
+    Logger.error(TAG + '@addAccountToGethKeystore', 'Failed to import raw key', e)
+    if (e.toString().includes('account already exists')) {
```
cmcewen: think this is what you're talking about in terms of error parsing. i'm not really sure there is any other way to check this save to look for an enum in the blockchain codebase, but if you think of one it would definitely be nicer

packages/mobile/src/invite/saga.ts
```diff
@@ -346,25 +350,13 @@ export function* skipInvite() {
...
-      // Import account into the local geth node
-      tempAccount = yield call(
-        contractKit.web3.eth.personal.importRawKey,
-        trimLeading0x(inviteCode),
-        TEMP_PW
-      )
-    }
+    yield call(waitForGethConnectivity)
+    // Import account into the local geth node
+    const tempAccount: string | null = yield call(
```
cmcewen: if this is possible null we should probably throw, no?

packages/mobile/src/pincode/utils.ts
```diff
@@ -13,10 +12,11 @@ export function isPinCorrect(
   currentAccount: string
 ): Promise<typeof pin> {
   return new Promise((resolve, reject) => {
+    // TODO(yorke: fix
```
cmcewen: you'll just remove this, right?

cmcewen (commented):
curious to hear more about manual testing here, but overall seems to be looking good

packages/contractkit/src/wallets/rpc-wallet.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    })
+
+    it('SBAT to unlock added account', async () => {
+      const account = await rpcWallet.addAccount(keys.privateKey, passphrase)
+      await rpcWallet.unlockAccount(account, passphrase, 5)
+      const isUnlocked = rpcWallet.isAccountUnlocked(account)
+      expect(isUnlocked).toBeTruthy()
+    })
+
+    // TODO: enable when ganache supports eth_signTransaction
```
cmcewen: This seems pretty important imo. Can you link to an issue or any more info about this?

packages/mobile/src/web3/saga.ts
```diff
@@ -396,11 +137,9 @@ export function* switchToGethFromForno() {
     yield put(setContractKitReady(false)) // Lock contractKit during provider switch
     yield put(setFornoMode(false))
     yield spawn(gethSaga)
+    // yield call(ensureAccountInGethKeystore)
```
annakaz: Nit: remove comment?

packages/mobile/src/web3/contracts.ts
```diff
@@ -15,8 +15,11 @@ import { provider } from 'web3-core'
 
 const tag = 'web3/contracts'
 
-const contractKitForno = newKitFromWeb3(getWeb3(true))
-const contractKitGeth = newKitFromWeb3(getWeb3(false))
+// TODO(yorke): merge instances of contractkit and refactor retrieval of providers
+const IpcProvider = getIpcProvider()
+export const gethWallet = new RpcWallet(IpcProvider)
+const contractKitForno = newKitFromWeb3(getWeb3(true), gethWallet)
```
annakaz, line 22: Previously contractKitForno did not depend on the IPC provider- will this be defined if the app is started in forno mode?

annakaz (commented):
Does this require that geth runs for signing in forno mode? If yes, it would be good to test running geth on the lower end devices that forno is designed for

packages/mobile/src/escrow/saga.ts
```diff
@@ -146,22 +140,10 @@ function* withdrawFromEscrow() {
...
-    } else {
-      // Unlock temporary account
-      yield call(contractKit.web3.eth.personal.unlockAccount, tempWalletAddress, TEMP_PW, 600)
-    }
-
-    const msgHash = contractKit.web3.utils.soliditySha3({ type: 'address', value: account })
-
-    Logger.debug(TAG + '@withdrawFromEscrow', `Signing message hash ${msgHash}`)
+    // Unlock temporary account
+    yield call(gethWallet.unlockAccount, tempWalletAddress, TEMP_PW, 600)
```
gastonponti: This `600` could be useful to extract it as a configuration parameter

packages/mobile/src/web3/reducer.ts
```diff
@@ -5,9 +5,6 @@ import { Actions, ActionTypes, Web3SyncProgress } from 'src/web3/actions'
 export interface State {
   syncProgress: Web3SyncProgress
   latestBlockNumber: number
-  account: string | null
```
jmrossy: Why move these?

packages/contractkit/src/wallets/rpc-wallet.ts
```diff
@@ -0,0 +1,112 @@
...
+  addPersonal = 'personal_importRawKey',
+  unlockPersonal = 'personal_unlockAccount',
+  signPersonal = 'personal_sign',
+  signTransaction = 'eth_signTransaction',
+  signTypedData = 'eth_signTypedData',
+}
+
+const currentTimeInSeconds = () => Math.round(Date.now() / 1000)
+
+export class RpcWallet implements Wallet {
```
gastonponti: I prefer that you follow the same pattern that we already have in the contractKit.

The `Wallet` is the one that store the account info. And the `Signer` the one that know how to sign. This made easier every test, and will make all the wallets extend from the same base. Otherwise will depend on the implementation and that will be hard to follow in the future.

This could be a case of `RemoteWallet`, because the RPC call depends on the provider, and this wallet could allow to call a remote node.

packages/mobile/src/web3/reducer.ts
yorhodes: good question
I guess it wasn't strictly necessary but wanted to make clear that web3 was no longer managing the account state

packages/mobile/src/web3/contracts.ts
yorhodes: hmm, good point
this does require geth is running although we can turn off syncing for the lower end devices
is there a process for testing on lower end hardware?

packages/mobile/src/dappkit/dappkit.ts
```diff
@@ -87,7 +86,7 @@ function* produceTxSignature(action: RequestTxSignatureAction) {
         params.to = tx.to
       }
       Logger.debug(TAG, 'Signing tx with params', JSON.stringify(params))
-      const signedTx = await contractKit.web3.eth.signTransaction(params)
```
gastonponti: My only concern about these cases (they are all over the refactor), is that your signing directly from the wallet.

So if in the future this changes, you are going to need a refactor for this.
The idea behind using the `web3.eth.sign` is that you could switch the wallet whenever you want, without changing the code implementation.

Again, don't know if it's a problem now, but one of the advantages of using web3 is that will act as a interface, where you'll be switching the wallets, without changing anything.

Another thing, is that this asumes that the every time you want a Tx to be signed, will be from the local `geth` node . The idea of the `celoProvider` of catching the Tx that you have a local way of signing it or otherwise sent it to the node, is lost using the wallet this way. Again, is it something that could happen? A user using the remote node keystore to sign? I don't know, but this way you are removing that option.

gastonponti (changes requested):
The Wallet/Signer it's something that you should probably change (I'm adding @codyborn that was the leading engine behind that change)

The other comment, the one about using `web3.eth` it's probably something that you have to discuss with the wallet team. Why was that in the beginning? And if they think that it's the right way of doing it (@cmcewen)

packages/mobile/src/web3/reducer.ts
jmrossy: I appreciate that. But if you're going to move it, you'll also need to create a redux migration for existing users on alfajores, and a test for that migration. It's not a lot of work but it'll further increase conflicts with my PR here:
https://github.com/celo-org/celo-monorepo/pull/3777
Which will merge hopefully in a day or two

packages/mobile/src/invite/saga.ts
yorhodes: the `addAccount` will throw and never return null

codyborn:
Can you provide some background on the motivation behind the migration from web3 to geth form managing accounts?

packages/contractkit/src/wallets/rpc-wallet.test.ts
```diff
@@ -0,0 +1,68 @@
+import { testWithGanache } from '@celo/dev-utils/lib/ganache-test'
+import { generateKeys, generateMnemonic } from '@celo/utils/src/account'
+import { normalizeAddressWith0x, privateKeyToAddress } from '@celo/utils/src/address'
+import { RpcWallet } from './rpc-wallet'
+
+testWithGanache('rpc-wallet', (web3) => {
```
codyborn, line 87: There are some pretty good wallet tests here that should be pulled into these tests:
https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/wallets/local-wallet.test.ts

yorhodes:
> Can you provide some background on the motivation behind the migration from web3 to geth form managing accounts? @codyborn 

@cmcewen can speak to this but it's less about migrating from web3 to geth and more about generally consolidating keystore management. A big motivation here is eliminating privateKey read/writes to disk outside geth and transmitting secrets over network to forno.
As @gastonponti mentioned, the refactor from web3 to gethwallet usage could be shortsighted. I am considering reverting some of that.

packages/mobile/src/dappkit/dappkit.ts
yorhodes: reverted these changes

packages/mobile/src/web3/reducer.ts
yorhodes: reverted this change

packages/contractkit/src/wallets/rpc-wallet.test.ts
```diff
@@ -0,0 +1,241 @@
...
+          try {
+            await rpcWallet.unlockAccount(ACCOUNT_ADDRESS1, 'wrong_passphrase', DURATION)
+          } catch (e) {
+            expect(e.message).toContain('RpcSigner unlock failed')
+          }
+        })
+      })
+
+      // TODO(yorke): ENABLE WHEN GANACHE SUPPORTS THESE METHODS PROPERLY
+      // describe('signing', () => {
```
codyborn: You can use `xdescribe` or `describe.skip` to add tests that aren't yet passing.

packages/contractkit/src/wallets/rpc-wallet.ts
```diff
@@ -0,0 +1,72 @@
...
+    const signer = this.getSigner(address) as RpcSigner
+    return signer.isUnlocked()
+  }
+
+  /**
+   * Gets the signer based on the 'from' field in the tx body
+   * @param txParams Transaction to sign
+   * @dev overrides WalletBase.signTransaction
+   */
+  async signTransaction(txParams: Tx): Promise<EncodedTransaction> {
```
codyborn: Why can't we use the signTransaction() in the WalletBase? It seems the only difference is the `v` value derived from the chainID is set to 0 here. I assume we'd still need to base it off the chain ID, right?

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -0,0 +1,75 @@
+// @ts-ignore-next-line
+import { account as Account } from 'eth-lib'
+import * as ethUtil from 'ethereumjs-util'
+import { InterceptedMethods } from '../../providers/celo-provider'
+import { RpcCaller } from '../../utils/rpc-caller'
+import { RLPEncodedTx } from '../../utils/signing-utils'
+import { Signer } from './signer'
+
+const decodeSig = (sig: any) => {
```
codyborn: Thanks for modularizing this. Can you move this to `signing-utils.ts` and update `local-signer` to use this?

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -0,0 +1,75 @@
...
+const currentTimeInSeconds = () => Math.round(Date.now() / 1000)
+
+export class RpcSigner implements Signer {
+  constructor(
+    protected rpc: RpcCaller,
+    protected account: string,
+    protected unlockBufferSeconds = 5,
+    protected unlockTime = -1,
+    protected unlockDuration = -1
+  ) {}
```
codyborn, line 66: Neat! I didn't realize you can do this in TS.

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+const currentTimeInSeconds = () => Math.round(Date.now() / 1000)
+
+export class RpcSigner implements Signer {
+  constructor(
+    protected rpc: RpcCaller,
+    protected account: string,
+    protected unlockBufferSeconds = 5,
```
codyborn, line 63: What's the purpose of `unlockBufferSeconds`? Can you add a class header with parameter descriptions?

packages/contractkit/src/wallets/rpc-wallet.ts
yorhodes: added a comment which speaks to this

packages/contractkit/src/wallets/rpc-wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  async addAccount(privateKey: string, passphrase: string): Promise<string> {
+    privateKey = ensureLeading0x(privateKey)
+    const address = normalizeAddressWith0x(privateKeyToAddress(privateKey))
+    if (this.hasAccount(address)) {
+      throw new Error(`RpcWallet: account already exists`)
+    }
+    const signer = new RpcSigner(this.rpc, address)
+    const resp = await signer.init(privateKey, passphrase)
```
cmcewen: If there's an error on resp, throw here with the error

packages/mobile/src/invite/saga.ts
```diff
@@ -346,26 +344,10 @@ export function* skipInvite() {
...
-      // Import account into the local geth node
-      tempAccount = yield call(
-        contractKit.web3.eth.personal.importRawKey,
-        trimLeading0x(inviteCode),
-        TEMP_PW
-      )
-    }
-    Logger.debug(TAG + '@addTempAccountToWallet', 'Account added', tempAccount!)
+    // Import account into the local geth node
+    const wallet = yield call(getWallet)
```
cmcewen: add `: RpcWallet` here so we get the type

packages/mobile/src/invite/saga.ts
```diff
@@ -346,26 +344,10 @@ export function* skipInvite() {
...
+    const tempAccount = yield call(wallet.addAccount, inviteCode, TEMP_PW)
```
cmcewen: need to have `yield call([wallet, wallet.addAccount] , inviteCode, TEMP_PW)` otherwise the context isn't set. i can tell you didn't run this code 😄

packages/mobile/src/web3/saga.ts
```diff
@@ -204,31 +199,19 @@ export function* assignAccountFromPrivateKey(privateKey: string) {
...
-      } catch (e) {
-        if (e.toString().includes('account already exists')) {
-          Logger.warn(TAG + '@assignAccountFromPrivateKey', 'Attempted to import same account')
-        } else {
-          Logger.error(TAG + '@assignAccountFromPrivateKey', 'Error importing raw key')
-          throw e
-        }
+    const wallet = yield call(getWallet)
+    try {
+      yield call(wallet.addAccount, privateKey, pincode)
```
cmcewen: same on the `[wallet,...` deal

packages/mobile/src/invite/saga.ts
```diff
@@ -382,11 +364,8 @@ export function* withdrawFundsFromTempAccount(
   newAccount: string
 ) {
   Logger.debug(TAG + '@withdrawFundsFromTempAccount', 'Unlocking temporary account')
-  const fornoMode = yield select(fornoSelector)
-  const contractKit = yield call(getContractKit)
-  if (!fornoMode) {
-    yield call(contractKit.web3.eth.personal.unlockAccount, tempAccount, TEMP_PW, 600)
-  }
+  const wallet = yield call(getWallet)
+  yield call(wallet.unlockAccount, tempAccount, TEMP_PW, 600)
```
cmcewen: same on context

packages/contractkit/src/wallets/rpc-wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+export class RpcWallet extends WalletBase {
+  protected readonly rpc: RpcCaller
+
+  constructor(protected _provider: provider) {
+    super()
+    this.rpc = new DefaultRpcCaller(_provider)
+  }
+
+  async addAccount(privateKey: string, passphrase: string): Promise<string> {
+    privateKey = ensureLeading0x(privateKey)
```
cmcewen: c'mon man. new variable

packages/contractkit/src/wallets/rpc-wallet.ts
cmcewen: This fails with 0x in privatekey, so you need to trim the 0x

cmcewen (changes requested):
i did some testing and got the import working, but txs still do not send
```
Possible Unhandled Promise Rejection (id: 0):
Error: missing-tx-params-populator@getGasPrice: gas price for currency 0xa561131a1C8aC25925FB848bCa45A74aF61e5A38 cannot be computed pass it explicitly
```

packages/contractkit/src/wallets/rpc-wallet.ts
yorhodes: weird that it worked with ganache...

packages/mobile/src/invite/saga.ts
yorhodes: I don't understand why the tests are passing?... I mocked these functions but maybe incorrectly

packages/contractkit/src/wallets/rpc-wallet.ts
yorhodes: I totally agree with this but this is present all over contractkit utils functions and I think I borrowed it...

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -205,3 +205,12 @@ export function verifyEIP712TypedDataSigner(
   const valid = verifySignature(ensureLeading0x(trimmedData), signedData, expectedAddress)
   return valid
 }
+
+export function decodeSig(sig: any) {
```
jmrossy, line 209: Just wondering if any of this is duping what's already in `https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/utils/signing-utils.ts`?

packages/contractkit/src/wallets/rpc-wallet.test.ts
```diff
@@ -0,0 +1,242 @@
...
+export const PRIVATE_KEY1 = '1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef'
+export const ACCOUNT_ADDRESS1 = normalizeAddressWith0x(privateKeyToAddress(PRIVATE_KEY1))
+export const PRIVATE_KEY2 = '0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890fdeccc'
+export const ACCOUNT_ADDRESS2 = normalizeAddressWith0x(privateKeyToAddress(PRIVATE_KEY2))
+
+const PASSPHRASE = 'ce10'
+const DURATION = 10000
+
+// ./build/bin/geth --alfajores --syncmode=lightest  --rpcapi=net,eth,web3,personal --verbosity 4
+describe.skip('rpc-wallet', () => {
```
jmrossy, line 62: Should this skip be here?

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -0,0 +1,105 @@
...
+   * @param rpc RPC caller instance
+   * @param account Account address derived from the private key to be called in init
+   * @param unlockBufferSeconds Number of seconds to shrink the unlocked duration by to account for
+   * latency and timing inconsistencies on the node
+   * @param unlockTime Timestamp in seconds when the signer was last unlocked
+   * @param unlockDuration Number of seconds that the signer was last unlocked for
+   *
+   */
+  constructor(
+    protected rpc: RpcCaller,
```
jmrossy, line 61: Nit, why have an empty constructor? You can just declare the vars with default values above this function

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -0,0 +1,105 @@
...
+    protected account: string,
+    protected unlockBufferSeconds = 5,
+    protected unlockTime = -1,
+    protected unlockDuration = -1
+  ) {}
+
+  async init(privateKey: string, passphrase: string) {
+    const response = await this.rpc.call('personal_importRawKey', [
```
jmrossy: It would be great to get some typing for these call. An enum with the set of possible calls would work

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -0,0 +1,105 @@
...
+      trimLeading0x(privateKey),
+      passphrase,
+    ])
+    this.checkResponse(response, 'init')
+    return response.result!
```
jmrossy: It's certain that result isn't null/empty? Or should we check and throw?

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  async signRawTransaction(tx: Tx): Promise<any> {
+    if (normalizeAddressWith0x(tx.from! as string) !== this.account) {
```
jmrossy, line 75: what's `from`'s type if not string?

packages/mobile/src/app/ErrorMessages.ts
```diff
@@ -65,4 +62,7 @@ export enum ErrorMessages {
   ADDRESS_VALIDATION_PARTIAL_POORLY_FORMATTED = 'addressValidationPartialPoorlyFormatted',
   ADDRESS_VALIDATION_FULL_OWN_ADDRESS = 'addressValidationFullOwnAddress',
   ADDRESS_VALIDATION_PARTIAL_OWN_ADDRESS = 'addressValidationPartialOwnAddress',
+  PGPNP_QUOTA_ERROR = 'PGPNP_QUOTA_ERROR',
```
jmrossy: Please restore these or the error messages won't work. Also I think their previous positions made more sense

packages/mobile/src/geth/geth.ts
```diff
@@ -73,19 +73,23 @@ function getFolder(filePath: string) {
...
+async function createNewGeth(sync: boolean = true): Promise<typeof RNGeth> {
   Logger.debug('Geth@newGeth', 'Configure and create new Geth')
   const { nodeDir, syncMode } = networkConfig
   const genesis: string = await readGenesisBlockFile(nodeDir)
   const networkID: number = GenesisBlockUtils.getChainIdFromGenesis(genesis)
 
   Logger.debug('Geth@newGeth', `Network ID is ${networkID}, syncMode is ${syncMode}`)
 
+  // 25 is react native geth default
+  const maxPeers = sync ? 25 : 0
```
jmrossy: Nit: consider hoisting this 25 up to a const at the top for clarity

packages/mobile/src/geth/geth.ts
```diff
@@ -149,38 +153,41 @@ async function initGeth() {
...
-          `the node will not be able to sync with the network till restart`,
-        error
-      )
-      return false
-    }
-    if (enodes != null) {
-      await writeStaticNodes(nodeDir, enodes)
-      return true
-    }
+  // if (await staticNodesAlreadyInitialized(nodeDir)) {
```
jmrossy: Will you be restoring this?

packages/mobile/src/geth/saga.ts
```diff
@@ -177,18 +179,19 @@ function* monitorGeth() {
...
+    const gethRelatedSagas = yield fork(monitorGeth)
+    yield take(Actions.CANCEL_GETH_SAGA)
+    yield cancel(gethRelatedSagas)
+  } else {
+    // TODO(yorke): monitor to make sure RPC is still available
+    yield put(setGethConnected(true))
+    yield take(Actions.CANCEL_GETH_SAGA)
   }
+
+  yield put(setGethConnected(false))
```
jmrossy, line 221: Seems like when we reach here, we should restart or exit the app right? Maybe there's logic elsewhere I'm missing but do we ever recover from this spot and reconnect somehow? Feel free to just leave a TODO

packages/mobile/src/app/ErrorMessages.ts
```diff
@@ -65,4 +62,7 @@ export enum ErrorMessages {
...
+  INSUFFICIENT_BALANCE = 'INSUFFICIENT_BALANCE',
+  MATCHMAKING_QUOTA_EXCEEDED = 'MATCHMAKING_QUOTA_EXCEEDED',
```
jmrossy: Thanks for the typo fix here ;)

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -1,5 +1,5 @@
 import { ContractKit } from '@celo/contractkit'
-import { getPhoneHash, isE164Number, PhoneNumberUtils } from '@celo/utils/src/phoneNumbers'
+import { getPhoneHash, isE164Number, PhoneNumberUtils } from '@celo/utils/lib/phoneNumbers'
```
jmrossy: The mobile app intentionally uses `src`. Tbh I can't remember why but I remember we did this on purpose. @cmcewen do you remember? Was it for better typescript support?

packages/mobile/src/invite/saga.ts
```diff
@@ -356,26 +354,10 @@ export function* skipInvite() {
...
-      // Import account into the local geth node
-      tempAccount = yield call(
-        contractKit.web3.eth.personal.importRawKey,
-        trimLeading0x(inviteCode),
-        TEMP_PW
-      )
-    }
-    Logger.debug(TAG + '@addTempAccountToWallet', 'Account added', tempAccount!)
+    // Import account into the local geth node
+    const wallet: RpcWallet = yield call(getWallet)
```
jmrossy: Much nicer :)

packages/mobile/src/pincode/utils.ts
```diff
@@ -1,30 +1,14 @@
...
-      getContractKitOutsideGenerator()
-        .then((contractKit: any) =>
-          contractKit.web3.eth.personal
-            .unlockAccount(currentAccount, pin, 1)
-            .then((result: boolean) => (result ? resolve(pin) : reject()))
-        )
-        .catch(reject)
-    }
-  })
+export async function isPinCorrect(pin: string, currentAccount: string): Promise<typeof pin> {
```
jmrossy: Can we rename this to something like `attemptWalletUnlock` or `ensureCorrectPin`? Seems closer to what it does.
Not your code, I know, but while we're here :)

packages/mobile/src/web3/contracts.ts
```diff
@@ -27,7 +32,18 @@ function getWeb3(fornoMode: boolean): Web3 {
       fornoMode ? DEFAULT_FORNO_URL : 'geth'
     }`
   )
-  return fornoMode ? new Web3(getHttpProvider(DEFAULT_FORNO_URL)) : new Web3(getIpcProvider())
+  return fornoMode ? new Web3(getHttpProvider(DEFAULT_FORNO_URL)) : new Web3(ipcProvider)
+}
+
+export const getWalletAsync = async () => promisifyGenerator(getWallet())
+
+export function* getWallet() {
```
jmrossy: To keep consistency with our other functions like this, maybe `getConnectedWallet` would be more accurate here

packages/mobile/src/web3/saga.ts
```diff
@@ -189,8 +184,10 @@ export function* getOrCreateAccount() {
 
     return accountAddress
   } catch (error) {
+    if (privateKey && !error.message.contains(privateKey)) {
+      Sentry.captureException(error)
+    }
     // Capturing error in sentry for now as we debug backup key issue
```
jmrossy: Move comment up?

packages/mobile/src/web3/saga.ts
```diff
@@ -189,8 +184,10 @@ export function* getOrCreateAccount() {
 
     return accountAddress
   } catch (error) {
+    if (privateKey && !error.message.contains(privateKey)) {
```
jmrossy, line 188: Good idea to check for PK here

packages/mobile/src/web3/saga.ts
```diff
@@ -203,44 +200,37 @@ export function* assignAccountFromPrivateKey(privateKey: string) {
...
-          Logger.warn(TAG + '@assignAccountFromPrivateKey', 'Attempted to import same account')
-        } else {
-          Logger.error(TAG + '@assignAccountFromPrivateKey', 'Error importing raw key')
-          throw e
-        }
+    const wallet: RpcWallet = yield call(getWallet)
+    try {
+      yield call([wallet, wallet.addAccount], privateKey, pincode)
+    } catch (e) {
+      if (e.toString().includes('account already exists')) {
```
jmrossy: Can we strongly type the errors here?

jmrossy (commented):
Overall nice work here! Please see comments below

packages/contractkit/src/wallets/signers/rpc-signer.ts
yorhodes: `checkResponse` does this

packages/contractkit/src/wallets/signers/rpc-signer.ts
yorhodes: this is a product of extending web3's types
apparently a number is a valid from because in some ethereum clients it is used as an index for the account keystore
we don't use this anywhere though

packages/mobile/src/app/ErrorMessages.ts
yorhodes: ! woah
this was auto-formatted somehow

packages/mobile/src/geth/geth.ts
yorhodes: nope
I will delete it
we need to explicitly write an empty array of enodes because even with a `maxPeers` of 0, geth will use these static nodes

packages/contractkit/src/wallets/rpc-wallet.test.ts
yorhodes: yes, unfortunately we don't have good testing infrastructure for this type of test right now

yorhodes: but I'm leaving it in so at least devs can use it locally

packages/contractkit/src/utils/signing-utils.ts
yorhodes: there are a bunch of incompatibilities but I don't think so anyway
just for example, `stringNumberToHex` makes hex strings an even number of bytes by prepending a 0 byte which the geth RPC complains about

packages/mobile/src/identity/privateHashing.ts
cmcewen: I think just because it's the actually TS source as opposed to the built source, so no need to rebuild when making changes

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -0,0 +1,130 @@
...
+   * latency and timing inconsistencies on the node
+   * @param unlockTime Timestamp in seconds when the signer was last unlocked
+   * @param unlockDuration Number of seconds that the signer was last unlocked for
+   *
+   */
+  constructor(
+    protected rpc: RpcCaller,
+    protected account: string,
+    protected unlockBufferSeconds = 5,
+    protected unlockTime = -1,
```
gastonponti, line 64: extracted because is it easy for testing? Otherwise could be more of an internal variable instead of something set from the dev

gastonponti (approved):

packages/mobile/src/geth/geth.ts
```diff
@@ -149,38 +153,36 @@ async function initGeth() {
...
-      return false
-    }
-    if (enodes != null) {
-      await writeStaticNodes(nodeDir, enodes)
-      return true
-    }
+  Logger.debug('Geth@ensureStaticNodesInitialized', 'initializing static nodes')
+  let enodes: string | null = null
+  try {
+    enodes = await StaticNodeUtils.getStaticNodesAsync(DEFAULT_TESTNET)
```
cmcewen: we don't need to fetch these if sync is disabled, right? we can just set it to `[]`

cmcewen (approved):
one small change about fetching static nodes (also maybe we should cache the values so we don't have to fetch every time?)

packages/contractkit/src/wallets/signers/rpc-signer.ts
yorhodes: I think `unlockTime` should be internal

packages/mobile/src/app/ErrorMessages.ts
jmrossy: Still need to fix these

jmrossy (dismissed):

yorhodes:
retest

PR #3829, opened 2020-05-15
[contractKit][mobile] add percent of attestations completed as part of heuristic for phone number validation (will also be used in PGPNP)

aslawson:
### Description

- Add percent of attestations completed as part of heuristic for phone number validation
- when new version of contractKit is published, this will also be used in PGPNP

`const DEFAULT_ATTESTATION_THRESHOLD = 0.8`
Value `.8` open for discussion, I feel this value is a bit aggressive.  I ran a script to calculate the average fraction of attestations completed from alfajores which resulted in `0.9153` with 479 identifiers.  Not sure how great of a proxy this is as many users are likely developers in areas where resources may be less constraint.

### Other changes

- Updated PGPNP get-salt to move the database call inside the try block in case of db failure

### Tested

Updated tests

### Related issues

- https://github.com/celo-org/celo-labs/issues/506

### Backwards compatibility

Backwards compatible change


packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -200,6 +210,42 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
+   * @param attestationThreshold Optional threshold for fraction attestations completed. Will
+   *  default to hardcoded value if absent.
+   */
+  async getAttestationStatus(
+    identifier: string,
+    account: Address,
+    numAttestationsRequired?: number,
+    attestationThreshold?: number
+  ): Promise<AttestationsStatus> {
+    if (!numAttestationsRequired) {
```
nambrot: I think you can just set the defaults in the parameters of the function I believe?

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -200,6 +210,42 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
+   * Returns the attestation status of an identifer/account pair indicating whether the pair
+   * is verified.
+   * @param identifier Attestation identifier (e.g. phone hash)
+   * @param account Address of the account
+   * @param numAttestationsRequired Optional number of attestations required.  Will default to
+   *  hardcoded value if absent.
+   * @param attestationThreshold Optional threshold for fraction attestations completed. Will
+   *  default to hardcoded value if absent.
+   */
+  async getAttestationStatus(
```
nambrot: What do you think about just expanding the existing `getAttestationStat` function with this functionality?

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -200,6 +210,42 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
+      numAttestationsRequired = DEFAULT_NUM_ATTESTATIONS_REQUIRED
+    }
+    if (!attestationThreshold) {
+      attestationThreshold = DEFAULT_ATTESTATION_THRESHOLD
+    }
+    const attestationStats = await this.getAttestationStat(identifier, account)
+    const numAttestationsRemaining = numAttestationsRequired - attestationStats.completed
+    const fractionAttestation =
+      attestationStats.total < 1 ? 0 : attestationStats.completed / attestationStats.total
+    const isVerified = numAttestationsRemaining <= 0 && fractionAttestation > attestationThreshold
```
nambrot: I don't have the full context here on where this heuristic is being used, but what does `verified` stand for in this function? This might be the ESL in me, but isn't verification technically a possibly misleading term? If I understand correctly, it's a transfer from the mobile code which defines `verified` as "this account has received sufficient attestations to move on".

nambrot (commented):
Just some small comments!

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -15,11 +15,21 @@ import {
...
 export interface AttestationStat {
   completed: number
   total: number
 }
 
+export interface AttestationsStatus {
+  isVerified: boolean
+  numAttestationsRemaining: number
+  total: number
+  fractionAttestation: number
```
jmrossy: including `total` instead of the ratio may be more useful for some clients. I can imagine scenarios where we'd want to surface these stats in the app and i'd be nice to not need to work backwards from the fraction

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: +1, if would be like 
`numAttestationsRequired?: number = DEFAULT_NUM_ATTESTATIONS_REQUIRED`

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: I kind of like having them separate, one is a pure proxy call and the other does some post-processing.
But i'd be nice if we rename this one to `getVerifiedStatus`. Also `getAttestationStat` should be `getAttestationStats`.

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -200,6 +210,42 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
     (stat) => ({ completed: valueToInt(stat[0]), total: valueToInt(stat[1]) })
   )
 
+  /**
+   * Returns the attestation status of an identifer/account pair indicating whether the pair
```
jmrossy: Can we add a bit more about what 'verified' means? Namely that it's that an ID is above some threshold of confidence.

packages/contractkit/src/wrappers/Attestations.ts
jmrossy: yeah, we should clarify that 'verified' is a term of convenience to mean that the attestation stats for a given ID are beyond a certain threshold of confidence

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -15,11 +15,21 @@ import {
   valueToInt,
 } from './BaseWrapper'
 
+// TODO (amyslawson) is there a proper config to put this in?
+const DEFAULT_NUM_ATTESTATIONS_REQUIRED = 3
+const DEFAULT_ATTESTATION_THRESHOLD = 0.8
```
codyborn: Let's put these in the src/config.ts

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -187,7 +197,7 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
   )
 
   /**
-   * Returns the attestation stats of a phone number/account pair
+   * Returns the attestation stats of a identifer/account pair
```
codyborn, line 200: nit: identifier

packages/contractkit/src/wrappers/Attestations.ts
codyborn: nit: identifier

packages/mobile/src/identity/verification.test.ts
```diff
@@ -106,11 +106,21 @@ const mockActionableAttestations: ActionableAttestation[] = [
 ]
```
codyborn, line 107: I think it also makes sense to add some tests that mock the `getAttestationStat` call and test that `getAttestationStatus` behaves correctly.

codyborn (approved):
Looks good; signing off pending the threshold discussion.

packages/contractkit/src/wrappers/Attestations.ts
aslawson: makes sense, i'll update

packages/mobile/src/identity/verification.test.ts
aslawson: ya -- I was looking for examples of contractKit wrapper tests and thought it was weird I didn't see a test directory.  just realized all the tests are alongside the files in the wrappers directory 🤦‍♀️

i'll add some testing there

packages/contractkit/src/wrappers/Attestations.ts
aslawson: i think you mean `completed` as the `total` is already included.  but you'd need to reverse calculate to get `completed`

packages/contractkit/src/wrappers/Attestations.ts
aslawson: there is no contractkit/src/config.ts :/

nambrot (approved):

packages/contractkit/src/wrappers/Attestations.test.ts
```diff
@@ -0,0 +1,56 @@
+import { testWithGanache } from '@celo/dev-utils/lib/ganache-test'
```
jmrossy, line 1: Thanks for adding this!

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -200,6 +210,38 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
+    numAttestationsRequired: number = DEFAULT_NUM_ATTESTATIONS_REQUIRED,
+    attestationThreshold: number = DEFAULT_ATTESTATION_THRESHOLD
+  ): Promise<AttestationsStatus> {
+    const attestationStats = await this.getAttestationStat(identifier, account)
+    const numAttestationsRemaining = numAttestationsRequired - attestationStats.completed
+    const fractionAttestation =
+      attestationStats.total < 1 ? 0 : attestationStats.completed / attestationStats.total
+    // 'verified' is a term of convenience to mean that the attestation stats for a
+    // given identifier are beyond a certain threshold of confidence
+    const isVerified = numAttestationsRemaining <= 0 && fractionAttestation > attestationThreshold
```
jmrossy:
```suggestion
    const isVerified = numAttestationsRemaining <= 0 && fractionAttestation >= attestationThreshold
```

jmrossy (approved):
Looks great! One optional comment below

packages/contractkit/src/wrappers/Attestations.test.ts
codyborn: +1

PR #3830, opened 2020-05-16
Add local flag to account:list command to avoid needing an external node

nategraf:
### Description

I tried to pull the some addresses form my Ledger wallet with `account:list` and realized it would throw an exception if not connected to a node. This PR adds a `--local` flag to skip the node addresses and only return the ones from the local wallet.

### Tested

Used it to get my wallet addresses from my Ledger.

### Backwards compatibility

100%


packages/cli/src/commands/account/list.ts
```diff
@@ -8,24 +8,32 @@ export default class AccountList extends BaseCommand {
...
+    const res = this.parse(AccountList)
 
     const celoProvider: CeloProvider = this.kit.web3.currentProvider as any
-    const addresses = await this.kit.web3.eth.getAccounts()
+
+    // Retreive accounts from the connected Celo node.
+    const addresses = []
+    if (res.flags.local) {
+      addresses.push(...(await this.kit.web3.eth.getAccounts()))
+    }
```
gastonponti: This could be easier to read in the future (and probably you want the !local for retrieving the node)
```suggestion
    const addresses = res.flags.local? [] : await this.kit.web3.eth.getAccounts()
```

packages/cli/src/commands/account/list.ts
```diff
@@ -8,24 +8,32 @@ export default class AccountList extends BaseCommand {
...
       .map((value) => toChecksumAddress(value))
-    let localName = 'Local'
-    const res: ParserOutput<any, any> = this.parse()
 
-    if (res.flags.useLedger) {
-      localName = 'Ledger'
-    }
+    // Display the addresses.
+    const localName = res.flags.useLedger ? 'Ledger' : 'Local'
     console.log('All Addresses: ', addresses)
```
gastonponti: This line shows every address, and with this you are showing [] for local

What you could do, is to change the order, retrieve first the local wallet addresses, and then all of them.
And instead of the last recommendation add this:

```
    // Get addresses from the local wallet.
    const localAddresses = celoProvider.wallet
      .getAccounts()
      .map((value) => toChecksumAddress(value))

   const addresses = res.flags.local? localAddresses : await this.kit.web3.eth.getAccounts()
```

packages/cli/src/commands/account/list.ts
nategraf: Does https://github.com/celo-org/celo-monorepo/pull/3830/commits/c24e42aa5cf206d410b7c301b8513935ed585699 address this comment?

packages/cli/src/commands/account/list.ts
nategraf: Does https://github.com/celo-org/celo-monorepo/pull/3830/commits/c24e42aa5cf206d410b7c301b8513935ed585699 address this comment?

packages/cli/src/commands/account/list.ts
```diff
@@ -8,29 +8,38 @@ export default class AccountList extends BaseCommand {
...
+      res.flags.local ?? true
+        ? celoProvider.wallet.getAccounts().map((value) => toChecksumAddress(value))
+        : []
 
-    if (res.flags.useLedger) {
-      localName = 'Ledger'
+    // Display the addresses.
+    const localName = res.flags.useLedger ? 'Ledger' : 'Local'
+    if (res.flags.local === undefined) {
+      console.log('All Addresses: ', nodeAddresses.concat(localAddresses))
```
gastonponti: The `this.kit.web3.eth.getAccounts()` retrieves the node + the local, because basically we are showing the user the addresses that have with all the info to sign a Tx (locally or via a node, but for the user it doesn't matter).
so, `nodeAddresses.concat(localAddresses)` will duplicate the local ones
That's why I was removing the locals from all the address to show the "keystore addresses" (or node addresses)

packages/cli/src/commands/account/list.ts
```diff
@@ -8,29 +8,38 @@ export default class AccountList extends BaseCommand {
...
+    // Get addresses from the local wallet.
     const celoProvider: CeloProvider = this.kit.web3.currentProvider as any
-    const addresses = await this.kit.web3.eth.getAccounts()
-    const localAddresses = celoProvider.wallet
-      .getAccounts()
-      .map((value) => toChecksumAddress(value))
-    let localName = 'Local'
-    const res: ParserOutput<any, any> = this.parse()
+    const localAddresses =
+      res.flags.local ?? true
```
gastonponti, line 29: this is like having the flag `local` with a default: true
If I'm understanding what you wanted to do, basically this gives you the possibility of later, hide the keystore address
I'm not really sure why that's needed. I totally agree on the "only local" because you won't need a node to check your ledger for example, but in the other case IMO it seems the same (remember that the `getAccounts` already do the work of getting the local addresses)

packages/cli/src/commands/account/list.ts
```diff
@@ -8,29 +8,38 @@ export default class AccountList extends BaseCommand {
...
 
-    if (res.flags.useLedger) {
-      localName = 'Ledger'
+    // Display the addresses.
+    const localName = res.flags.useLedger ? 'Ledger' : 'Local'
+    if (res.flags.local === undefined) {
+      console.log('All Addresses: ', nodeAddresses.concat(localAddresses))
+    }
+    if (!res.flags.local) {
+      console.log('Keystore Addresses: ', nodeAddresses)
```
gastonponti, line 40: Same problem, `getAccounts` has everything

gastonponti (approved):

PR #3835, opened 2020-05-18
Migrate from prom-to-sd to Prometheus server+Sidecar

jcortejoso:
### Description

This PR includes the migration for Stackdriver metric format, from using [stackdriver-prometheus](https://github.com/Stackdriver/stackdriver-prometheus) to [stackdriver-prometheus-sidecar](https://github.com/Stackdriver/stackdriver-prometheus-sidecar). The goal of the change is expanding current [Stackdriver metric quotas](https://cloud.google.com/monitoring/quotas).

### Other changes

- Adapted `prometheus-stackdriver` helm chart introduced by @tkporter to optionally including the service Account (as it should not be needed by default to gke clusters)
- Included prometheus+sidecar with new gke cluster deployments.
- Removed prom-to-sd stuff
- Included new celotool deployment commands `gke-prometheus` useful for existing cluster and for development clusters where prometheus+sidecar is not deployed when the cluster is created.

### Tested

- [x] Tested the GKE deployment.
- [ ] Pending to test with AKS

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/484

### Backwards compatibility
Metrics will change the name/path, which will affect existing dashboards and/or alerts.


packages/celotool/src/lib/prometheus.ts
```diff
@@ -47,30 +51,48 @@ export async function removeHelmRelease() {
...
+      `--set stackdriver_metrics_prefix=external.googleapis.com/prometheus/${clusterConfig.clusterName}`,
+      `--set gcloudServiceAccountKeyBase64=${await getPrometheusGcloudServiceAccountKeyBase64forAKS(
+        clusterConfig.clusterName
+      )}`
+    )
+  } else {
+    const clusterName = fetchEnv(envVar.KUBERNETES_CLUSTER_NAME)
+    params.push(
+      `--set cluster=${clusterName}`,
+      `--set stackdriver_metrics_prefix=external.googleapis.com/prometheus/${clusterName}`
```
nambrot, line 86: Why are we setting the cluster as part of the prefix, shouldn't the cluster_name be a label of the metric itself?

jcortejoso: So with current prometheus config the cluster_name is not a label, but we can it if we find that approach better.

nambrot: I guess I'm not exactly sure what is deployed where, but at least the eksportisto metrics have the `cluster_name` label and they seem correct

jcortejoso: I guess it was added by eksportio. I did not see that label for other metrics. I added it as an extra label

jcortejoso: I guess it was added by eksportisto itself. Added as an extra label for all metrics

nambrot (approved):

PR #3840, opened 2020-05-19
[Wallet][ReDesign] Notifications

jeanregisser:
### Description

This implements the Notifications redesign.

Note:  `Avatar` used is the old one. Copies and icons are not entirely final.

Next steps: 
- Transaction drilldown
- Send/Request flow
- Animate hide/show of Send/Request bottom bar as the user scrolls

### Other changes

- Added a generic `Card` component with shadow and optional rounded corners

### Tested

Some screenshots (though there are more variations).

**`<SimpleMessagingCard />`**

**`<RequestMessagingCard />`**

**`<AggregatedRequestsMessagingCard />`**

**Screens**

Navigation bar is currently broken with the update to `react-navigation` v5. Will update soon.

### Related issues

- Epic #3645 

### Backwards compatibility

Yes.


packages/react-components/components/Card.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { elevationShadowStyle } from '@celo/react-components/styles/styles'
+import React from 'react'
+import { StyleSheet, View, ViewProps } from 'react-native'
+
+export interface Props extends ViewProps {
+  rounded?: boolean
+  children?: React.ReactNode
+}
+
+export default function Card({ style, rounded = false, ...props }: Props) {
```
aaronmgdr, line 11: The Card!

packages/react-components/components/Card.tsx
```diff
@@ -0,0 +1,24 @@
+import colors from '@celo/react-components/styles/colors.v2'
+import { elevationShadowStyle } from '@celo/react-components/styles/styles'
+import React from 'react'
+import { StyleSheet, View, ViewProps } from 'react-native'
+
+export interface Props extends ViewProps {
+  rounded?: boolean
+  children?: React.ReactNode
```
aaronmgdr, line 8: when would use use without children?

packages/react-components/components/MessagingCard.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { StyleSheet } from 'react-native'
+
+// Card used by all messaging cards
+export default function MessagingCard({ style, ...props }: Props) {
+  return <Card style={[styles.container, style]} rounded={true} {...props} />
+}
+
+const styles = StyleSheet.create({
+  container: {
+    height: 144,
```
aaronmgdr, line 12: I think its best to avoid an explicit height unless there is no other solution. if we must have one consider a min-height. better to have height be determined by content + padding. 

explicit height  can lead to overflow due to differences in language content length, or user chosen font size.

packages/react-components/components/RequestMessagingCard.tsx
```diff
@@ -0,0 +1,72 @@
...
+    flex: 1,
+    marginRight: 16,
+  },
+  title: {
+    ...fonts.label,
+    color: colors.gray4,
+    marginBottom: 4,
+  },
+  amount: {
+    // TODO: font should be bold
```
aaronmgdr, line 62: handle now or later?

packages/react-components/components/Card.tsx
jeanregisser: I don't know, mostly wanted the interface to be exactly like `<View />` 😉

packages/react-components/components/MessagingCard.tsx
jeanregisser: I agree with you 💯 

Here's a special case because those cards are displayed in a carousel in the home feed and we need all of them to have the same height so it doesn't look weird when swiping through them.
I did that mostly as a shortcut for now. 

I'll see if I can make the carousel cards grow to reach the height of the longest card.

packages/react-components/components/RequestMessagingCard.tsx
jeanregisser: I'll add it now!

packages/react-components/components/Card.tsx
aaronmgdr: gotcha

packages/react-components/components/RequestMessagingCard.tsx
jeanregisser: Actually I'll add it later 😉

packages/mobile/locales/es-419/walletFlow5.json
```diff
@@ -38,7 +38,7 @@
   "welcomeToCeloPayments": "Bienvenido a Pagos Celo",
   "letsGetStartedByFirstTransaction": "Empezemos enviando tu primera transacción",
   "sendMoney": "Enviar Dinero",
-  "verificationFee": "Tasa de Verificación",
+  "verificationFee": "Tasa de Verificación",
```
aaronmgdr, line 41: i wonder what is different?

packages/mobile/src/notifications/SummaryNotification.tsx
```diff
@@ -1,69 +1,54 @@
...
-  if (total - PREVIEW_SIZE > 0) {
-    return (
-      <View>
-        <Text style={styles.moreWithCountText}>
-          {i18n.t('global:moreWithCount', { count: total - PREVIEW_SIZE + 1 })}
-        </Text>
-      </View>
-    )
+function getContext(count: number) {
+  if (count === 2) {
```
cmcewen, line 20: do i care if this is a `switch` statement? no.
does this look exactly like a `switch` statement? yes.
is this comment pointless? yes.
😄

packages/mobile/src/paymentRequest/utils.ts
```diff
@@ -49,7 +49,7 @@ export function getSenderFromPaymentRequest(
   } else {
     return {
       kind: RecipientKind.MobileNumber,
-      address: paymentRequest.requesterAddress,
+      address: paymentRequest.requesteeAddress,
```
cmcewen, line 55: This seems like it was broken before maybe?

packages/react-components/components/CallToActionsBar.tsx
```diff
@@ -0,0 +1,41 @@
+import TextButton from '@celo/react-components/components/TextButton.v2'
+import React from 'react'
+import { StyleSheet, View } from 'react-native'
+
+export interface CallToAction {
+  onPress: () => unknown
```
cmcewen, line 6: when would this have a return type besides void?

cmcewen (approved):
Love to see more lines deleted than added 🥇

packages/mobile/locales/es-419/walletFlow5.json
jeanregisser: I resolved the conflict within the GitHub UI, maybe it changed the underlying unicode char 🤷‍♂️

packages/mobile/src/notifications/SummaryNotification.tsx
jeanregisser: Ahah, I remember starting with a `switch` statement, and then switched (no pun intended) to `if` thinking I'd make a comparison.

packages/mobile/src/paymentRequest/utils.ts
jeanregisser: Yes! This looks like it might have cause a bug, not sure if we actually used that address though.

packages/react-components/components/CallToActionsBar.tsx
jeanregisser: I copied this from the existing code we had, I think if this was used with an `async` handler, `void` wouldn't work.

PR #3846, opened 2020-05-20
allow AzureKeyVaultClient to take in custom credential

zviadm:
### Description

Allow creating AzureKeyVaultClient with custom credentials


zviadm:
ping? i already have a local fork for myself, but i think this is useful change for everyone, since sometimes it is better to use EnvironmentCredentials instead of DefaultCredentials (when you dont want to have a fallback, otherwise you get really confusing error messages).

barbaraliau:
@zviadm @codyborn is this PR still relevant?

zviadm:
> @zviadm @codyborn is this PR still relevant?

i think it is still relevant and useful but up to you all if you want to pick it up or not. For my own use cases, as I mentioned, I have already made a fork anyways so I don't care if it gets merged or not.

AlexBHarley (approved):
I think this makes sense to merge. It's inline with what we do in the [AWS HSM client](https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/wallets/aws-hsm-wallet.ts#L29)

codyborn (approved):
Apologies for the delay. I'm quite bad about checking Github notifications. Looks good!

PR #3867, opened 2020-05-21
[Wallet] Account Key Cancel Confirmation

aaronmgdr:
### Description

show a dialog when user goes to cancel prompting them to complete instead

### Tested

Screen Shot 2020-05-21 at 11 20 26 AM

### Related issues

- #3645 

### Backwards compatibility

### DIFF
--- a/packages/mobile/locales/en-US/backupKeyFlow6.json
+++ b/packages/mobile/locales/en-US/backupKeyFlow6.json
@@ -46,6 +46,10 @@
     "confirmation": "I have sent each Safeguard phrase to a trusted friend.",
     "yourSafeguards": "Your Safeguards"
   },
+  "cancelDialogTitle": "Finish setting up your Account Account Key",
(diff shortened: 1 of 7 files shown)


packages/mobile/locales/en-US/backupKeyFlow6.json
```diff
@@ -46,6 +46,10 @@
     "confirmation": "I have sent each Safeguard phrase to a trusted friend.",
     "yourSafeguards": "Your Safeguards"
   },
+  "cancelDialogTitle": "Finish setting up your Account Account Key",
```
jmrossy, line 49: Consider putting these under a shared property (e.g. `cancelDialog`) with 4 children. We have a bad habit of dumping everything at the top level but we shouldn't do that going forward, it's getting messy!

jmrossy (commented):
Changes mostly lgtm, one comment about i18n string structure

packages/mobile/locales/en-US/backupKeyFlow6.json
aaronmgdr: ha tots i usually do

i1skn (approved):
LGTM

Issue #3868, opened 2020-05-21
Validator explorer group uptime issues


mrsmkl:
### Expected Behavior

A useful and intuitive behaviour.

### Current Behavior

One problem is that affialiated and unelected non-elected validators are included in calculation.

Another issue is that it is confusing that new validators have low uptime.

Some feedback:

zviad | wotrust.us Yesterday at 8:05 PM
it would probably be most accurate if score is calculated by only considering elected validators from "election:current", or "election:run", taking even fully accepted members would still be incorrect if some of the validators arent getting elected.

zviad | WOTrust | celovote.com Today at 12:28 AM
one suggestion would be to completely remove value from group uptime column, and only keep "per validator" uptime-s, that you show only if you expand the group row. The group's average uptime isn't really relevant for anything, and because new validators start with 0% uptime it only creates confusion unless you are really familiar with all the different scores that exist in the system.

Peter [ChainLayer.io] Today at 12:29 AM
Or just the (average) last epoch uptime of the validators in the group since that is what the voter rewards are based on anyway if im not mistaking

marekolszewski:
Since the uptime column is not accurate, can we hide it temporarily? CC @aaronmgdr

aaronmgdr:
sure Ill do a quick hide

Fouda-volt:
1. There is no explanation for all the metrics that are shown in simple words for newcomers. An info pop-up can be really useful.
2. The percentages listed in the three votes columns are very confusing and I feel don't give any insight about the validator 

![image](https://user-images.githubusercontent.com/62042003/82624543-d8ad3200-9bb0-11ea-8c74-b708a1186d73.png)

cfl0ws:
Can someone please point me to the uptime calculation?

cfl0ws:
> 1. There is no explanation for all the metrics that are shown in simple words for newcomers. An info pop-up can be really useful.
> 2. The percentages listed in the three votes columns are very confusing and I feel don't give any insight about the validator
> 
> ![image](https://user-images.githubusercontent.com/62042003/82624543-d8ad3200-9bb0-11ea-8c74-b708a1186d73.png)

I agree that tooltips or something like that explaining each metric would be very helpful.

zviadm:
To give more suggestions, first step would be identifying who the validators page is for. I can see two distinct groups that would care about validator and groups but in very different ways:
* Celo gold holders.
* Group and validator owners.

Celo Gold holders mostly care about:
[...]

PR #3872, opened 2020-05-22
Temp Hide Uptime

aaronmgdr:
### Description

Uptime isnt really helpful right now so we are hiding it

### Other changes

also adds an overflow scroll so that this is usable on screens that are not huge

### Related issues

- sweeps under rug  #https://github.com/celo-org/celo-monorepo/issues/3868


marekolszewski (approved):
Looks good!

PR #3875, opened 2020-05-22
Fixes to Oracle deployment

codyborn:
### Description

Fixing some issues I hit while deploying from a service principal account.

### Tested

Tested deployment/upgrade/destroy with service principal account.


jcortejoso (approved):
lgtm

PR #3877, opened 2020-05-22
[Wallet] Secure Send Pt 2

tarikbellamine:
### Description
Added small Secure Send features:
- ability for a user to force secure send by hitting "edit" on the `SendConfirmation` screen
- text on the transaction drilldown screen to let the user know if a recipient's address has changed via Secure Send

### Other changes
Removed an action we no longer needed: `validateRecipientAddressFailure`

### Tested
Wrote tests for the new features I added.

### Related issues
Add on to #3791 

### Backwards compatibility
Yes


packages/mobile/src/identity/saga.ts
```diff
@@ -54,12 +53,11 @@ export function* validateRecipientAddressSaga({
     yield put(validateRecipientAddressSuccess(e164PhoneNumber, validatedAddress))
   } catch (error) {
     Logger.error(TAG, 'validateRecipientAddressSaga/Address validation error: ', error)
-    if (error.message in ErrorMessages) {
+    if (Object.values(ErrorMessages).includes(error.message)) {
```
cmcewen, line 56: is this a stylistic preference, or different behavior?

packages/mobile/src/transactions/TransactionReview.tsx
```diff
@@ -3,26 +3,64 @@ import * as React from 'react'
...
+
+  const newAddress = secureSendPhoneNumberMapping[e164PhoneNumber].address
+  if (!newAddress || newAddress === address) {
+    return false
+  }
+
+  return true
+}
+
+const mapStateToProps = (state: RootState, ownProps: OwnProps): StateProps => {
```
cmcewen, line 59: mr. ownProps 😂

packages/mobile/src/navigator/types.tsx
```diff
@@ -123,7 +123,8 @@ export type StackParamList = {
   [Screens.TabNavigator]: {}
   [Screens.TransactionReview]: {
     reviewProps: ReviewProps
-    confirmationProps: TransferConfirmationCardProps | ExchangeConfirmationCardProps
+    transferConfirmationProps?: TransferConfirmationCardProps
```
cmcewen: Is there a reason to change these? This changes from "one of these two is required" to "one, both, or neither of these two" will exist

packages/mobile/src/transactions/TransactionReview.tsx
```diff
@@ -46,7 +84,9 @@ class TransactionReview extends React.PureComponent<Props> {
   }
 
   getConfirmationProps = () => {
-    const confirmationProps = this.props.route.params.confirmationProps
+    const confirmationProps =
+      this.props.route.params.transferConfirmationProps ||
```
cmcewen: Similar question here - how do we know that one and only one of the right type is passed?

packages/mobile/src/transactions/TransferFeedItem.tsx
```diff
@@ -35,13 +35,15 @@ function navigateToTransactionReview({
   }
 
   const recipient = getRecipientFromAddress(address, addressToE164Number, recipientCache)
+  const e164PhoneNumber = addressToE164Number[address] || undefined
```
cmcewen, line 38: is this just to make extra clear that it could be undefined?

packages/mobile/src/identity/saga.ts
tarikbellamine: I believe it's different behavior. Correct me if I am wrong but I think the original checks property names in `ErrorMessages` whereas the latter checks values, which is what I actually want here.

I could also change the error messages I am throwing to be the properties if that's how we normally do it, but this was less work for me.

packages/mobile/src/navigator/types.tsx
tarikbellamine: The reason why I did was that these two types are drastically different. Wasn't a problem when all we were doing was passing them through but I ran into issues when I needed to operate on them before passing them on.

Is there an easy way to do conditionals based on types if we have one property with two potential types? For example, I was trying unsuccessfuly to do `if (typeof confirmationProps === TransferConfirmationCardProps)` but it doesn't seem `typeof` works that way?

packages/mobile/src/transactions/TransferFeedItem.tsx
tarikbellamine: It's to force `null` to `undefined`. I don't really care about the difference between the two but all of the type checking downstream is expecting `string` or `undefined`

packages/mobile/src/transactions/TransactionReview.tsx
tarikbellamine: Admittedly less apparent on this screen but the only way to navigate here is by using either `navigateToPaymentTransferReview` or `navigateToExchangeReview`. Each of these functions requires their respective `confirmationProps` as an argument so it would be impossible to get to this screen with anything other than exactly 1

packages/mobile/src/identity/saga.ts
tarikbellamine: To clarify, I wrote the original and noticed it wasn't working when adding on in this PR.

packages/mobile/src/navigator/types.tsx
cmcewen: I think you're probably looking for a type guard: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

packages/mobile/src/transactions/TransactionReview.tsx
cmcewen: I guess probably convention would be using those functions, but I think in the future someone could add another type to TransactionReview and just call an explicit navigate. Basically, there is nothing stopping someone from just calling `this.props.navigation.navigate(Screens.TransactionReview)`

packages/mobile/src/identity/saga.ts
cmcewen: Ah makes sense. Yeah my intuition was that it was different behavior, but just checking that it meant the previous behavior was broken 😄

packages/mobile/src/transactions/TransactionReview.tsx
tarikbellamine: That's true, only did it this way because I thought I had too. Using typeguards is definitely preferable

packages/mobile/src/navigator/types.tsx
tarikbellamine: Indeed I am! Will revert these changes and use this

packages/mobile/src/transactions/TransactionReview.tsx
```diff
@@ -55,12 +95,13 @@ class TransactionReview extends React.PureComponent<Props> {
     return confirmationProps
   }
 
-  renderCard = (type: TokenTransactionType, confirmationProps: any) => {
+  renderCard = (type: TokenTransactionType, props: any) => {
```
cmcewen: small nit i missed - could we type these props?

cmcewen (approved):
🚢:shipit:

packages/mobile/src/transactions/TransactionReview.tsx
tarikbellamine: done. also cleaned up those methods we dont need

PR #3880, opened 2020-05-22
[Wallet] Type safe navigate

jeanregisser:
### Description

Following the upgrade to `react-navigation` which enables type safe navigation params, we still had our global `navigate` wrapper which wasn't checking navigation params.
This PR fixes it and caught a couple of small issues related to them.

Note: the default `navigate` method typing is slightly incorrect and won't catch when the parameters for a route are needed but they aren't provided.
See https://github.com/react-navigation/react-navigation/issues/7936

Our global `navigate` helper fixes that.

### Other changes

- Fix crash when opening the import wallet screen from the enter invite code screen during onboarding
- Add `getMockStackScreenProps` helper to create both `navigation` and `route` props in a type safe way for tests.

### Tested

Global `navigate(...)` calls are now type safe for the route name and its params!

### Related issues

None.

### Backwards compatibility

Yes.


packages/mobile/src/account/saga.ts
```diff
@@ -69,7 +69,8 @@ export function* getPincode(withVerification = true) {
     const pin = yield new Promise((resolve, reject) => {
       navigate(Screens.PincodeEnter, {
         onSuccess: resolve,
-        onFail: reject,
+        // Looks like this param is not used/needed anymore
+        // onFail: reject,
```
jeanregisser: @i1skn TypeScript caught that this param wasn't used anymore. Could you confirm it's the intended behaviour?

packages/mobile/src/import/ImportWallet.tsx
```diff
@@ -72,7 +72,7 @@ export class ImportWallet extends React.Component<Props, State> {
 
   checkCleanBackupPhrase = () => {
     const { route, navigation } = this.props
-    if (route && route.params.clean) {
+    if (route.params?.clean) {
```
jeanregisser, line 75: This is the bug that triggered the crash when navigating to the import screen.

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -43,7 +47,12 @@ export function replace(routeName: string, params?: object) {
     })
 }
 
-export function navigate(routeName: string, params?: object) {
+export function navigate<RouteName extends keyof StackParamList>(
+  ...args: undefined extends StackParamList[RouteName]
+    ? [RouteName] | [RouteName, StackParamList[RouteName]]
+    : [RouteName, StackParamList[RouteName]]
+) {
```
jeanregisser, line 54: This is the TypeScript magic which checks the right params are passed!

```tsx
// with the following config:
type StackParamList = {
  [Screens.MyRoute]: undefined
  [Screens.AnotherRoute]: { something: string }
  [Screens.MyWay]: { something: number } | undefined
}

// MyRoute doesn't take any param, it accepts both
navigate(Screens.MyRoute);
// and
navigate(Screens.MyRoute, undefined);

// AnotherRoute takes params, it only accepts the following and raises 
// a compile time error if one of the route param is missing or incorrect:
navigate(Screens.AnotherRoute, { something: "something" });
// This will also raise an error:
navigate(Screens.AnotherRoute)

// MyWay params are optional, it accepts both
navigate(Screens.MyWay, { something: 3 });
// and
navigate(Screens.MyWay)
// but errors with:
navigate(Screens.MyWay, { });
```

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -42,7 +42,7 @@ export class IncomingPaymentRequestListItem extends React.Component<Props> {
       firebasePendingRequestUid: id,
     }
 
-    if (addressValidationType !== AddressValidationType.NONE) {
+    if (addressValidationType && addressValidationType !== AddressValidationType.NONE) {
```
jeanregisser, line 45: Looks like this caught a possible bug, @tarikbellamine is this still correct?

packages/mobile/src/transactions/actions.ts
```diff
@@ -153,6 +153,6 @@ export const navigateToExchangeReview = (
   })
 }
 
-export const navigateToRequestedPaymentReview = (confirmationInput: ConfirmationInput) => {
-  navigate(Screens.SendConfirmation, { confirmationInput })
+export const navigateToRequestedPaymentReview = (transactionData: TransactionDataInput) => {
+  navigate(Screens.SendConfirmation, { transactionData })
```
jeanregisser, line 157: And this is another bug prevented!

packages/mobile/test/values.ts
```diff
@@ -176,6 +177,23 @@ export const mockNavigation: StackNavigationProp<any, any> = ({
...
 
+export function getMockStackScreenProps<RouteName extends keyof StackParamList>(
+  ...args: undefined extends StackParamList[RouteName]
+    ? [RouteName] | [RouteName, StackParamList[RouteName]]
+    : [RouteName, StackParamList[RouteName]]
+): StackScreenProps<StackParamList, RouteName> {
+  const [name, params] = args
+  return {
+    navigation: mockNavigation,
+    // @ts-ignore
```
jeanregisser: The params are type checked already, this is fine.

packages/mobile/src/account/saga.ts
i1skn: It seems it is unused! Nice to see new reactnavigation v6 in action 💪 Also, `reject` on line 69 probably would be unused as well.

packages/mobile/test/utils.ts
```diff
@@ -115,6 +122,23 @@ export function getMockI18nProps() {
...
+export function getMockStackScreenProps<RouteName extends keyof StackParamList>(
+  ...args: undefined extends StackParamList[RouteName]
+    ? [RouteName] | [RouteName, StackParamList[RouteName]]
+    : [RouteName, StackParamList[RouteName]]
+): StackScreenProps<StackParamList, RouteName> {
+  const [name, params] = args
+  return {
+    navigation: mockNavigation,
+    // @ts-ignore
+    route: {
```
jeanregisser, line 134: The params are type checked already, this is fine.

aaronmgdr:
nice work!

packages/mobile/src/account/Profile.tsx
```diff
@@ -19,9 +19,8 @@ interface StateProps {
   userContact: UserContactDetails
 }
 
-interface OwnProps {
-  navigation: any
-}
+// tslint:disable-next-line: no-empty-interface
+interface OwnProps {}
```
jmrossy, line 23: Why not just remove OwnProps from here and the next line?

packages/mobile/src/import/ImportWallet.tsx
jmrossy:
```suggestion
    if (route?.params?.clean) {
```

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
jmrossy: Looks ok to me :)

jmrossy (approved):
Looks great! A few optional comments

packages/mobile/src/account/Profile.tsx
jeanregisser: Right! Left this in case we wanted to add some props again. This screen will get refactored with the redesign and this will probably go away too.

Issue #3658, opened 2020-04-29
Wallet Analytics: Account Key


nityas:
Parent Epic: #3641 
### Expected Behavior

https://docs.google.com/document/d/1OGNJBijiWp7bWc2I0p8yyJ72EMCj9pNRT1VWxJJu5JY/edit#heading=h.crz3fi6kkwp8

### Current Behavior

What is the current behavior?

aaronmgdr:
- [x] backup_start: ‘set up now’ button click
- [x] backup_snooze: ‘wait one hour’ button click

- [x] backup_setup_info: info button is clicked
- [x] backup_setup_toggle_complete: “I have written down my key” is toggled to enabled state
- [x] backup_setup_toggle_incomplete: “I have written down my key” is toggled to disabled state

- [x] backup_setup_cancel: cancel button is clicked
	params: name of screen
- [x] backup_setup_cancel_confirm: action is taken on cancel modal
	params: option selected
- [ ] backup_setup_screenshot_warning: ‘okay’ button click
	Q: possible to log an event when the modal is launched?
- [x] backup_quiz_submit:
	param: number of deletes, time from first input word to submit
- [x]  backup_quiz_success:
	param: time from submit to validate
- [x]  backup_quiz_incorrect:
	param: time from submit to validate

- [x]  backup_quiz_cancel: cancel button is clicked
	param: name of screen
- [x]  backup_quiz_cancel_confirm: action is taken on cancel modal
	param: option selected
- [ ]  backup_quiz_screenshot_warning: ‘okay’ button click
	Q: possible to log an event when the modal is launched?

- [ ] backup_error: error in backup flow
	params: error identifier(or message)
- [x] backup_educate_1_next: next button on 1st edu screen
- [x] backup_educate_2_next: next button on 2nd edu screen
- [x] backup_educate_3_next: next button on 3rd edu screen
- [x] backup_educate_4_next: next button on 4th edu screen

- [x] backup_educate_1_cancel: cancel button on 1st edu screen
- [x] backup_educate_2_cancel: cancel button on 2nd edu screen
- [x] backup_educate_3_cancel: cancel button on 3rd edu screen
- [x] backup_educate_4_cancel: cancel button on 4th edu screen

aaronmgdr:
TODO: tracking of error in analytics.

add an error boundary? 

add lots of try catchs? 

both?

neither?

PR #3884, opened 2020-05-22
[Wallet] Account Key Analytics ++

aaronmgdr:
### Description

adds analytics for Account Key / Backup Screens

### Other changes

fixes lack of margin above quiz on Quiz screen

reduces wait time

### Tested

add test to ensure that when a word is pressed it is moved to the selected words area

Screen Shot 2020-05-22 at 2 27 24 PM

### Related issues

- #3658


packages/mobile/src/backup/BackupIntroduction.tsx
```diff
@@ -54,7 +54,7 @@ class BackupIntroduction extends React.Component<Props> {
   }
 
   onPressBackup = () => {
-    CeloAnalytics.track(CustomEventNames.set_backup_phrase)
```
tarikbellamine, line 57: is this deprecated or still being used elsewhere? if deprecated, can we delete from the constants file?

packages/mobile/src/backup/BackupPhrase.tsx
```diff
@@ -87,10 +86,26 @@ class BackupPhrase extends React.Component<Props, State> {
...
     })
+
+    CeloAnalytics.track(
+      value
+        ? CustomEventNames.backup_setup_toggle_enable
+        : CustomEventNames.backup_setup_toggle_disable
+    )
   }
 
   onPressConfirmArea = () => {
```
tarikbellamine, line 97: do we need the same tracking on both the `onPressConfirmArea` and `onPressConfirmSwitch`? seems kind of redundant but i dont know enough about how the component works to say for sure

packages/mobile/src/analytics/constants.ts
```diff
@@ -88,6 +88,18 @@ export enum CustomEventNames {
...
+
+  backup_quiz_submit = 'backup_quiz_submit',
+  backup_quiz_success = 'backup_quiz_success',
+  backup_quiz_incorrect = 'backup_quiz_incorrect',
+
+  backup_setup_toggle_complete = 'backup_setup_toggle_complete',
+  backup_setup_toggle_incomplete = 'backup_setup_toggle_incomplete',
+
   // Screen name: Backup_Phrase, Backup_Insist, Backup_Share, Backup_Set
   set_backup_phrase = 'set_backup_phrase', // (count # of taps on “Set Backup Phrase” in Backup_Phrase) [we should not track the actual value of this field, just whether the user filled it out]
```
tarikbellamine: would be great to remove all the old constants we no longer need so we dont crowd this file

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -124,6 +132,10 @@ export class BackupQuiz extends React.Component<Props, State> {
       mnemonicWords: mnemonicWordsUpdated,
       userChosenWords: userChosenWordsUpdated,
     })
+    CeloAnalytics.trackSubEvent(
+      CustomEventNames.backup_quiz_submit,
```
tarikbellamine, line 136: not sure how `trackSubEvent` works but should we be tracking `backup_quiz_submit` on every backspace given we are using `startTracking` and `endTracking` to track the duration?

packages/mobile/src/navigator/Headers.v2.tsx
```diff
@@ -59,7 +59,7 @@ export const emptyHeader: StackNavigationOptions = {
   headerShown: true,
   headerTitleStyle: [styles.headerTitle, componentStyles.screenHeader],
   headerTitleContainerStyle: {
-    alignItems: 'center' as 'center',
```
tarikbellamine, line 62: lol

tarikbellamine (commented):
nice work. think it's most of the way there just have a couple of questions and there might be a bit of cleanup to do with old analytics constants and getting the tests to pass

packages/mobile/src/backup/BackupQuiz.tsx
tarikbellamine: i see now, first argument is what you want the second argument to be attached to

packages/mobile/src/analytics/constants.ts
aaronmgdr: oh thats a good point.

packages/mobile/src/backup/BackupPhrase.tsx
aaronmgdr: yes because either might be pressed. 

*Area for when the text is pressed, *Switch for the switch itself. I would like to use one function but Switch component requires a function that takes boolean param

I could de duplicate the logic though

packages/mobile/src/navigator/Headers.v2.tsx
aaronmgdr: a good trick but not needed

packages/mobile/src/backup/BackupPhrase.tsx
aaronmgdr: ok much better now

packages/mobile/src/analytics/constants.ts
```diff
@@ -88,8 +88,19 @@ export enum CustomEventNames {
   invite_friends_sms = 'invite_friends_sms', // to count the # of taps on “Invite with SMS" button on Invite_Friends_Review
   invite_friends_whatsapp = 'invite_friends_whatsapp', // to count the # of taps on “Invite with WhatsApp" button on Invite_Friends_Review
 
+  backup_start = 'backup_start', // ‘set up now’ button click
```
annakaz, line 91: Nit: May be beyond the scope of this PR but it would be nice if we had a consistent comment format throughout this file (for example, "(count ... " or " Count quiz backspace pressed" etc)

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -27,7 +29,7 @@ const TAG = 'backup/BackupQuiz'
 const MNEMONIC_BUTTONS_TO_DISPLAY = 6
 
 // miliseconds
-const CHECKING_DURATION = 2.6 * 1000
+const CHECKING_DURATION = 1.8 * 1000
```
annakaz, line 32: Why did this const change? Also can you add a comment explaining what checking duration is

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -92,7 +94,7 @@ export class BackupQuiz extends React.Component<Props, State> {
   getMnemonicFromNavProps() {
     const mnemonic = this.props.route.params.mnemonic
     if (!mnemonic) {
-      throw new Error('Mnemonic missing form nav props')
+      throw new Error('Mnemonic missing from nav props')
```
annakaz, line 97: 🎨

packages/mobile/src/analytics/constants.ts
```diff
@@ -88,8 +88,19 @@ export enum CustomEventNames {
...
+  backup_setup_info = 'backup_setup_info',
+
+  backup_quiz_backspace = 'backup_quiz_backspace', // whenever the backspace is pressed
+
+  backup_quiz_submit = 'backup_quiz_submit',
+  backup_quiz_success = 'backup_quiz_success',
+  backup_quiz_incorrect = 'backup_quiz_incorrect',
+
+  backup_setup_toggle_complete = 'backup_setup_toggle_complete',
+  backup_setup_toggle_incomplete = 'backup_setup_toggle_incomplete',
```
annakaz: Are these used anywhere?

annakaz (approved):
LGTM assuming the unused analytics consts are removed!

packages/mobile/src/backup/BackupQuiz.tsx
aaronmgdr: oh i thought it felt too long. I was the one who originally set it to 2.6 but regretted it being so long.

Issue #3624, opened 2020-04-28
Android & iOS: ‘Confirming..’ text is shown for current and previous transaction on ‘Wallet’ and ‘Gold’ tab when first transaction is in confirming state and user scroll up-down the page.


Lss-Ankit:
**Frequency:** 50%
**iOS App version:** iOS Testflight build : v1.7.1 (28)
**Android App version:** Android Celo Alfa build : v1.7.1
**Repro on iOS device:** iPhone 11 Pro (13.3)
**Repro on Android device:** Samsung Galaxy J7 Prime (6.0.1)
**Testing Account:** 3155641283 /backup Key

**Repro Steps:** 
1) Launch the app and Login with test account
2) Proceed with buy/sell gold and make sure ‘confirming..’ text is shown
3) Scroll up & down the page and Observe 

**Current Behavior:** ‘Confirming..’ text is shown for current and previous transaction on ‘Wallet’ and ‘Gold’ tab

**Expected Behavior:** ‘Confirming..’ text should be shown for current transaction only

**Attachment:** [iOS_Confimring text is shown](https://drive.google.com/file/d/1gfkg9o8gD2UrzpO0vqCWmYeyrFrN7X4l/view?usp=sharing) (Observe the issue between 0:02 to 0:14 secs)

jmrossy:
Should be fixed by #3887

Issue #2704, opened 2020-02-07
“Confirming...” text is continuously shown on transaction from Wallet > Activity section while proceeding for transaction when “Data saver” toggle is ON. 


Lss-Ankit:
**Frequency:** 100%
**App version:** Internal Android Celo Alfa v1.6.0. 
**Repro on:** Google Pixel XL 2 (9.0), Samsung Galaxy A8+ (8.0.0), 
**Testing Account:** 7228953526/[backup Key](https://docs.google.com/spreadsheets/d/1Iy41GGA6e2pES2XHjBn7nb8X3n8zZsu7EW1jvwD2e70/edit?usp=sharing)

**Pre-condition:** Make sure Data Saver toggle is ON. 

**Repro Steps:** 
1) Launch the app and login with testing account.
2) Tap on + tab. 
3) Tap on any celo user. 
4) proceed for transaction. 
5) Navigate to the Wallet tab and observe “Confirming...” text is shown continuously.

**Investigation:** 
1. I have waited until 5-10 minutes but still unable to proceed for the transaction. 
2. User is able to proceed for transactions when Data saver toggle is OFF. 
3. Issue does not repro in IOS test flight build 1.6.0 (21). 
4. After cold launch the app confirming Transaction activity is removed.  

**Current Behavior:** “Confirming...” text is continuously shown
**Expected Behavior:** Users should be able to proceed for transactions. 
**Impact:** User is blocked to proceed for transaction when data saver toggle is ON
**Attachment:**  [Android_ConfirmingIssue.mp4 ](https://drive.google.com/file/d/11gO59sM-XJC3F3Sb0z-vgBq3msgjQxp6/view?usp=sharing)(Bug observe at 19 sec)

Lss-Ankit:
Verified on iOS Celo Alfajores Wallet (1.6.0 (21)) and Android Celo Alfa v1.6.0 and found the following 

Continuous loading indicator is shown when user invite any user if  “Data saver” toggle is ON
Impact: user blocked to invite any user when data Saver toggle is ON. 
Attachment: [Continous_LoadingIsShown.mp4 ](https://drive.google.com/file/d/1T_wcGS_mmsKWeS0lubf1aRPhYKEadX-Q/view?usp=sharing)

Lss-Ankit:
[...]

PR #3887, opened 2020-05-24
[Wallet] Add identity metadata to outgoing payments

### LINKED ISSUE #1335: [Android] ‘cUSD’ balance is not updated automatically when user receives an amount from other users.

### LINKED ISSUE #2647: [Wallet] Sync incoming transaction and balance update
This issue is a result of discussion started here: https://github.com/celo-org/celo-monorepo/pull/2548#issuecomment-579541351

### Expected Behavior

When user receives incoming transaction, we need to show it on the list and also update the balance. We should present these two actions to a user at the same time or explicitly communicate in the UI, that one of them is about to get refreshed. Exact behavior TBD (@nityas @coreycelo)

### Current Behavior

When user receives incoming transaction, first user would see it in the list and a couple of seconds later balance would update itself.

### LINKED ISSUE #2878: Wallet should refresh balance on receiving payment notification
### Expected Behavior

Refreshed balance

### Current Behavior

Stale balance, and autorefresh wasn't working for some reason (tested in Nigeria on my iPhone)

jmrossy:
### Description

The main goal of this changeset is to add identity metadata (phone number and salt) to outgoing payment in the encrypted comments. This enables recipients to find corresponding contact information without having to do mass lookups. To achieve this, several changes where required / beneficial:

- Improve type handling of Solidity `bytes` in ContractKit
- Reduce duplication of utils in ContractKit
- Confirm users are truly 'verified' in contact mapping
- Add transaction feed query watcher
- Trigger balance refresh on tx feed update
- Scan new tx feed items for identity metadata in the comments
- Persist useful identity metadata to the phone number mappings 
- Fix some edge cases that break comment decryption memoization
- Update contractKit docs
- Some related refactoring and cleanup

### Tested

- Sent btwn two verified numbers on two phones and confirmed txs get labelled with contact info correctly.
- Confirmed from logs behavior is correct
- Added lots of new unit tests

### Related issues

- Fixes #2878
- Fixes #2647
- Fixes #1335 
- Fixes ##3624
- Fixes https://github.com/celo-org/celo-labs/issues/473

### Backwards compatibility

Yes


packages/mobile/src/home/saga.ts
```diff
@@ -43,13 +45,6 @@ export function* refreshBalances() {
   yield put(fetchSentEscrowPayments())
 }
 
-export function* refreshBalancesWithLoadingSaga() {
```
jmrossy, line 46: Removed because its unused and redundant with `watchRefreshBalances`

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -0,0 +1,316 @@
+// Wallet utilities for comment encryption and comment key management
```
jmrossy, line 1: The file is actually a rename of the `identity/commentKey.ts` but it's been mostly rewritten.
It handles comment key management tasks, encryption/decryption, and extracting identity metadata from comments

packages/mobile/src/identity/commentEncryption.test.ts
```diff
@@ -0,0 +1,100 @@
+import {
```
jmrossy, line 8: Actually a rename of `identity/commentKey.ts` but I'll be significantly increasing the coverage we had there before merging this.

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -109,19 +109,15 @@ function* updateRecipientsCache(
 export function* fetchAddressesAndValidateSaga({ e164Number }: FetchAddressesAndValidateAction) {
```
jmrossy, line 110: Changes in this file are just a bit of restructuring to allow some useful code to be leverage by `identity/commentEncryption.ts` too

packages/mobile/src/qrcode/QRScanner.tsx
```diff
@@ -85,7 +88,7 @@ class QRScanner extends React.Component<Props> {
               }}
               style={styles.preview}
               type={RNCamera.Constants.Type.back}
-              onBarCodeRead={memoize(this.onBarCodeDetected, (qr) => qr.data)}
```
jmrossy, line 88: The way it was before, we were creating a new memoize function on each render, defeating the purpose of memoization I believe.

packages/mobile/src/transactions/actions.ts
```diff
@@ -16,81 +16,97 @@ export enum Actions {
   ADD_HASH_TO_STANDBY_TRANSACTIONS = 'TRANSACTIONS/ADD_HASH_TO_STANDBY_TRANSACTIONS',
   TRANSACTION_CONFIRMED = 'TRANSACTIONS/TRANSACTION_CONFIRMED',
   TRANSACTION_FAILED = 'TRANSACTIONS/TRANSACTION_FAILED',
+  NEW_TRANSACTIONS_IN_FEED = 'TRANSACTIONS/NEW_TRANSACTIONS_IN_FEED',
 }
 
-export interface AddStandbyTransaction {
+export interface AddStandbyTransactionAction {
```
jmrossy, line 22: I renamed the action interfaces for consistency with our others

packages/mobile/src/transactions/reducer.ts
```diff
@@ -1,60 +1,43 @@
-import { TokenTransactionType } from 'src/apollo/types'
-import { CURRENCY_ENUM } from 'src/geth/consts'
+import { getRehydratePayload, REHYDRATE, RehydrateAction } from 'src/redux/persist-helper'
+import { RootState } from 'src/redux/reducers'
 import { Actions, ActionTypes } from 'src/transactions/actions'
-
-export interface ExchangeStandby {
```
jmrossy, line 5: Moved to `transactions/types.ts`

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -223,23 +213,15 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
     identifier: string,
     account: Address,
-    numAttestationsRequired: number = DEFAULT_NUM_ATTESTATIONS_REQUIRED,
-    attestationThreshold: number = DEFAULT_ATTESTATION_THRESHOLD
-  ): Promise<AttestationsStatus> {
+    numAttestationsRequired?: number,
+    attestationThreshold?: number
+  ) {
     const attestationStats = await this.getAttestationStat(identifier, account)
-    const numAttestationsRemaining = numAttestationsRequired - attestationStats.completed
```
jmrossy, line 230: This stuff was moved to `utils/src/attestations.ts` to make it easier to use in certain parts of the wallet

packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -209,28 +209,7 @@ function mapInvite(tx: FeedItem): FeedItem {
   return tx
 }
 
-function getTransactions(data: UserTransactionsQuery | undefined) {
```
jmrossy, line 212: Moved to transferFeedUtils

packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -209,28 +209,7 @@ function mapInvite(tx: FeedItem): FeedItem {
...
-  return data?.tokenTransactions?.edges.map((edge) => edge.node).filter(isPresent) ?? []
-}
-
 export class TransactionsList extends React.PureComponent<Props> {
-  txsFetched = (data: UserTransactionsQuery | undefined) => {
```
jmrossy, line 217: Moved to the tx feed watcher saga which can handle this stuff more elegantly

nambrot (approved):
attestation ck changes look good to me!

packages/contractkit/src/wrappers/BaseWrapper.ts
```diff
@@ -50,19 +50,22 @@ export const valueToInt = (input: BigNumber.Value) =>
...
+// Hopefull this will improve in the future, at which point we can make improvements here
+type SolidityBytes = string | number[]
+export const stringToSolidityBytes = (input: string) => ensureLeading0x(input) as SolidityBytes
+export const bufferToSolidityBytes = (input: Buffer) => stringToSolidityBytes(bufferToHex(input))
+export const solidityBytesToString = (input: SolidityBytes): string => {
+  if (input === null || input === undefined || typeof input === 'string') {
+    return input
+  } else if (Array.isArray(input)) {
+    let hex = ''
+    input.map((num) => (hex += num))
```
gastonponti: (Detail)
Although this work, you are using a `map` as an iterator, and the essence is different.
You can use `forEach` instead of `map` for the same result, or even `reduce`
```
const hex = input.reduce((accum, num) => accum + num, '')
```
With reduce it's clearer that you are transforming an array into a result

gastonponti: This is not a detail, haha
What characteristic has each number? it's always a one-digit?
Otherwise, if every number it's a two-digit, you should handle the zero padding
```
hex +=  num.toString().padStart(2, '0')
```
Otherwise [88, 1, 50] will be '88150', and probably should be '880150'

packages/mobile/src/identity/commentEncryption.test.ts
```diff
@@ -0,0 +1,259 @@
...
+const simpleCommentEnc =
+  'BNFXzyIGjZqqNyq6r35aV2HlMMqUbGnIqboReD77MwAlI5IyzqLQ99WF5B1bsZSVS1K+7trtJtKGhIdI1vbSJSsBAQEBAQEBAQEBAQEBAQEBCVYJWqi/TZNXbAR9ziyX5MJCtfdulxA1tjlvHR/xpE6WnlC/kyXAfKIMqgKGJXchBNFXzyIGjZqqNyq6r35aV2HlMMqUbGnIqboReD77MwAlI5IyzqLQ99WF5B1bsZSVS1K+7trtJtKGhIdI1vbSJSsBAQEBAQEBAQEBAQEBAQEBBhjruDecYg9fsrPNcQbI3AkcvWra1MHIeOZlcycn7Vqtx+UVNR59A3kqdIDbLuGiAQEBAQEBAQEBAQEBAQEBAQ13jJ/iheQ05aIkSedfLcnSa3Rtr4dsbNt59TAZsQqGxPClubtekbOKaHz7lE5zyI803qP65frfyyCxuQ=='
+const simpleCommentWithMetadata = 'Simple comment~+14155550000piWqRHHYWtfg9'
+const simpleCommentWithMetadataEnc =
+  'BNFXzyIGjZqqNyq6r35aV2HlMMqUbGnIqboReD77MwAlI5IyzqLQ99WF5B1bsZSVS1K+7trtJtKGhIdI1vbSJSsBAQEBAQEBAQEBAQEBAQEBCVYJWqi/TZNXbAR9ziyX5MJCtfdulxA1tjlvHR/xpE6WnlC/kyXAfKIMqgKGJXchBNFXzyIGjZqqNyq6r35aV2HlMMqUbGnIqboReD77MwAlI5IyzqLQ99WF5B1bsZSVS1K+7trtJtKGhIdI1vbSJSsBAQEBAQEBAQEBAQEBAQEBBhjruDecYg9fsrPNcQbI3AkcvWra1MHIeOZlcycn7Vqtx+UVNR59A3kqdIDbLuGiAQEBAQEBAQEBAQEBAQEBAQ13jJ/iheQ05aIkSedfLcnSa3Rtr4dsbNt59TBTlRnqnaAoz0HyOyScr6ckQczlhtxwBqKgGuXE9OnjyeH4PIJ9yutViWOO+glrLIFjpDyl+7FebiJv/p9Ay21LHpUd7b5lSV4PWPDxamBMp6MHRdA='
+const complexComment = 'long comment with emoji 😇🤠👳🏽‍♂️~ test TEST test'
```
gastonponti, line 45: Always be prepared ❤️

gastonponti (commented):
The padding with zero probably is a bug, but all the other stuff lgtm

packages/contractkit/src/wrappers/BaseWrapper.ts
gastonponti: oh, and even more, this won't handle the `10` as `A`
Probably we have to check this, but I think the real solution is:
```
hex +=  num.toString(16).padStart(2, '0')
```
And [23, 1, 10, 15] will be '17010A0F'

jmrossy: Good questions here. Unfortunately, I'm not sure what the actual format of the data is when web3 gives a `number[]`. It doesn't seem to be documented and I've just pieced together some information based on these discussions:
https://github.com/ethereum-ts/TypeChain/issues/98
https://github.com/ethereum-ts/TypeChain/pull/153
https://github.com/ethereum-ts/TypeChain/issues/123

jmrossy: And in my testing, I've only seen it return string. Maybe we should log and throw an error here until we find a situation that actually hits this case? Then we can inspect the number[] and see how this should be implemented. WDYT?

jmrossy: I'm not a big fan of `reduce` in JS, i find the interface unintuitive, but agreed in this case it would work nicely. I'll make that change when we decide what to do with this section

gastonponti: Ok, reading some other questions, I think that this case is happening when is used an array of `bytes` in solidity.
And it would probably be a `number` for every `byte`, so I'm 98% sure (haha) that this:
`hex +=  num.toString(16).padStart(2, '0')` will work, otherwise it's impossible to have a letter in the hex, and that would be really weird.

This was probably not working before. My only problem with adding something that will fail if the array was used, is that we don't know if the user will know how to handle that.

In my point of view, in these cases, I prefer to follow what could be a convention in transforming array of numbers into hex, and if we are doing something wrong, this would be an issue from some functionality using it.

gastonponti: I found something. The lib `eth-lib` that we use, has this function:
```
const fromArray = arr => {
  let hex = "0x";
  for (let i = 0, l = arr.length; i < l; ++i) {
    let b = arr[i];
    hex += (b < 16 ? "0" : "") + b.toString(16);
  }
  return hex;
}
```
that basically transform an array of integers, the same way we are saying.
If I have to strictly think of an array of numbers how it turns to a hex, this is the only way possible.

jmrossy: Thanks for investigating! I'll go ahead and use a hex conversion approach similar to your suggestion and that snippet :)

trianglesphere (commented):
The comment changes look good. I like the change from `encrypted` -> `success`.
The unit tests look good and are great to have, but (nit) if how the tx feed gets populated hasn't changed too much (from a specialized service, and not local geth), it's nice to make sure the emojis are working end to end.

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -107,7 +111,7 @@ export interface Votes {
 export type HotfixParams = Parameters<Governance['methods']['executeHotfix']>
 export const hotfixToParams = (proposal: Proposal, salt: Buffer): HotfixParams => {
   const p = proposalToParams(proposal, '') // no description URL for hotfixes
-  return [p[0], p[1], p[2], p[3], bufferToString(salt)]
+  return [p[0], p[1], p[2], p[3], bufferToHex(salt)]
```
aslawson, line 114: we loose the `ensureLeading0x` check by changing this to bufferToHex -- not a large concern, but if anyone uses this without the `0x` and it will break for them.

packages/mobile/src/config.ts
```diff
@@ -32,7 +32,7 @@ export const DEV_SETTINGS_ACTIVE_INITIALLY = stringToBoolean(
 // VALUES
 export const GAS_INFLATION_FACTOR = 1.5 // Used when estimating gas for txs
 export const GAS_PRICE_INFLATION_FACTOR = 5 // Used when getting gas price, must match what Geth does
-export const BALANCE_OUT_OF_SYNC_THRESHOLD = 5 * 60 // 5 minutes
+export const BALANCE_OUT_OF_SYNC_THRESHOLD = 1 * 60 // 1 minute
```
aslawson, line 35: is this change intentional or did you just lower it for manual testing?

packages/mobile/src/identity/commentEncryption.test.ts
aslawson: 🤠

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -0,0 +1,309 @@
...
+  return newIdentityData
+}
+
+// Verify the metadata claims (they could be spoofed)
+// Returns set of metadata objects with invalid ones filtered out
+function* verifyIdentityMetadata(data: IdentityMetadataInTx[]) {
+  // This function looks a bit convoluted because:
+  // 1. Data could have dupes in it
+  // 2. Some duped phone hashes could be true, others false
+  // 3. And a given phoneHash could legitimately have multiple addresses
```
annakaz, line 233: 👍

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -81,7 +82,7 @@ function* doImportContacts() {
   yield call(updateRecipientsCache, e164NumberToRecipients, otherRecipients)
 }
 
-// Find the user's contact among those important and save useful bits
+// Find the user's own contact among those imported and save useful bits
```
annakaz, line 85: 🎨

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -0,0 +1,309 @@
...
+    contractKit.contracts.getAccounts,
+  ])
+  const hexString: string = yield call(accountsWrapper.getDataEncryptionKey, address)
+  return !hexString ? null : hexToBuffer(hexString)
+}
+
+export function* encryptComment(
+  comment: string | null,
+  toAddress: string | null,
+  fromAddress: string | null,
```
aslawson, line 65: when are these null?

packages/mobile/src/transactions/saga.ts
```diff
@@ -1,25 +1,133 @@
...
+  }
+}
+
+// Remove standby txs from redux state when the real ones show up in the feed
+function* cleanupStandbyTransacitons(newFeedTransactions: TransactionFeedFragment[]) {
+  const standbyTxs: StandbyTransaction[] = yield select(standbyTransactionsSelector)
+  const newFeedTxHashes = new Set(newFeedTransactions.map((tx) => tx?.hash))
+  // const inQueryTxs = (tx: StandbyTransaction) =>
+  //   tx.hash && queryDataTxHashes.has(tx.hash) && tx.status !== TransactionStatus.Failed
+  // const filteredStandbyTxs = this.props.standbyTransactions.filter(inQueryTxs)
```
annakaz: Remove commented out lines or add TODO?

annakaz (approved):
LGTM on the wallet side!

The refactor of how standby transactions are removed once confirmed may also close #3624 and #2704 😁

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+export function extractPhoneNumberMetadata(commentData: string) {
+  const phoneNumMetadata = commentData.match(PHONE_METADATA_REGEX)
+  if (!phoneNumMetadata || phoneNumMetadata.length < 4) {
+    return { comment: commentData }
+  }
+
+  return {
+    comment: phoneNumMetadata[1],
```
aslawson, line 162: I'm not super knowledgable about regex, but would the comment not be index 0 here?

aslawson: also wondering how this solution would scale if we add more metadata and perhaps some of it is missing.  feels almost like we should send json or some type:value in the metadata for more foolproof parsing

packages/mobile/src/qrcode/QRScanner.tsx
aslawson: interesting -- nice catch

packages/mobile/src/transactions/saga.ts
jmrossy: Ah good catch, thanks!

packages/mobile/src/identity/commentEncryption.ts
jmrossy: No, index 0 would be the full string. 
This scheme would not work well for generalized metadata but we can cross that bridge when we come to it :)

packages/mobile/src/identity/commentEncryption.ts
jmrossy: Practically speaking, when the user didn't get verified so never registered a DEK

packages/mobile/src/config.ts
jmrossy: intentional :)

packages/contractkit/src/wrappers/Governance.ts
jmrossy: Good catch here! I guess you mean we add the `ensureLeading0x`? I'm inclined to leave it as is, since long-term it'd be good for hex values to be properly prefixed, but if it causes trouble for anyone I'm happy to revert this.

packages/mobile/src/home/saga.ts
tarikbellamine: good catch!

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+export function* encryptComment(
+  comment: string | null,
+  toAddress: string | null,
+  fromAddress: string | null,
+  includePhoneNumMetadata: boolean = false
+) {
+  Logger.debug(TAG + 'encryptComment', 'Encrypting comment')
+  if (!features.USE_COMMENT_ENCRYPTION || !comment || !toAddress || !fromAddress) {
```
tarikbellamine, line 69: I believe you've already raised this, but this is what you meant when you said comment encryption was only possible between two verified accounts, right?

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -0,0 +1,309 @@
...
+    return comment
+  }
+
+  let commentToEncrypt = comment
+  if (features.PHONE_NUM_METADATA_IN_TRANSFERS && includePhoneNumMetadata) {
+    Logger.debug(TAG + 'encryptComment', 'Including phone number metadata in comment')
+    const selfPhoneDetails: PhoneNumberHashDetails | undefined = yield call(
+      getUserSelfPhoneHashDetails
+    )
+    commentToEncrypt = embedPhoneNumberMetadata(comment, selfPhoneDetails)
```
tarikbellamine, line 95: Out of curiosity, why is embedding phone number metadata necessary?

packages/mobile/src/identity/commentEncryption.ts
jmrossy: Currently, yes that's right. We should explore options for bringing this to more users

packages/mobile/src/identity/commentEncryption.ts
jmrossy: We used to lookup a user's whole address book on-chain to see who's verified. It was slow but it worked. Now that's not possible because of PGPNP. So we include the metadata in comments, enabling the recipient to connect the dots

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// Check all transaction comments for metadata
+function findIdentityMetadataInComments(
+  transactions: TransactionFeedFragment[],
+  commentKeyPrivate: string
+) {
+  const newIdentityData: IdentityMetadataInTx[] = []
+  for (const tx of transactions) {
+    if (tx.__typename !== 'TokenTransfer' || tx.type !== TokenTransactionType.Received) {
```
tarikbellamine, line 210: Do we want to skip `TokenTransactionType.Sent`?

jmrossy: We do, no sense decrypting and processing your own identity data as you send it out :)

packages/mobile/src/send/saga.ts
```diff
@@ -154,12 +153,9 @@ function* sendPaymentOrInviteSaga({
...
     }
 
-    const ownAddress = yield select(currentAccountSelector)
-    const comment = features.USE_COMMENT_ENCRYPTION
-      ? yield call(encryptComment, reason, recipientAddress, ownAddress)
-      : reason
-
+    const ownAddress: string = yield select(currentAccountSelector)
     if (recipientAddress) {
+      const comment = yield call(encryptComment, reason, recipientAddress, ownAddress, true)
```
tarikbellamine: Nit: consider renaming the result `encryptedComment`. As someone new to the codebase, the difference between `comment` and `reason` was already confusing to me.

Definitely optional but I think it would be useful if downstream references were also renamed `encryptedComment` so it is clear when we are dealing with a raw comment vs an encrypted one

packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -209,28 +209,7 @@ function mapInvite(tx: FeedItem): FeedItem {
   return tx
 }
 
-function getTransactions(data: UserTransactionsQuery | undefined) {
-  return data?.tokenTransactions?.edges.map((edge) => edge.node).filter(isPresent) ?? []
-}
-
 export class TransactionsList extends React.PureComponent<Props> {
```
tarikbellamine, line 215: Nit: can this be refactored to be a functional component now?

packages/mobile/src/transactions/saga.ts
```diff
@@ -1,25 +1,130 @@
...
+}
+
+// Remove standby txs from redux state when the real ones show up in the feed
+function* cleanupStandbyTransacitons(newFeedTransactions: TransactionFeedFragment[]) {
+  const standbyTxs: StandbyTransaction[] = yield select(standbyTransactionsSelector)
+  const newFeedTxHashes = new Set(newFeedTransactions.map((tx) => tx?.hash))
+  for (const standbyTx of standbyTxs) {
+    if (
+      standbyTx.hash &&
+      standbyTx.status !== TransactionStatus.Failed &&
```
tarikbellamine, line 39: Is there value in keeping failed transactions? If timestamp is easily accessible, maybe we can safely cleanup any failed transactions of a certain age or older?

tarikbellamine (commented):
Nice work! Left a few minor comments/questions

packages/mobile/src/identity/commentEncryption.ts
tarikbellamine: Haha good point!

packages/mobile/src/apollo/types.ts
```diff
@@ -161,6 +161,10 @@ export enum TokenTransactionType {
   NetworkFee = 'NETWORK_FEE',
 }
 
+export function isTokenTxTypeSent(type: TokenTransactionType) {
+  return type === TokenTransactionType.Sent || type === TokenTransactionType.EscrowSent
+}
+
```
jeanregisser: This file is generated by graphql-codegen.
We should move this somewhere else.

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -113,7 +110,7 @@ export class WalletHome extends React.Component<Props> {
   }
 
   componentDidMount() {
-    this.props.resetStandbyTransactions()
+    // TODO find a better home for this, its unrelated to wallet home
```
jeanregisser, line 113: 👍

packages/mobile/src/transactions/TransactionFeed.tsx
```diff
@@ -59,7 +59,7 @@ export class TransactionFeed extends React.PureComponent<Props> {
     `,
   }
 
-  renderItem = (commentKeyBuffer: Buffer | null, kind: FeedType) => ({
+  renderItem = (commentKey: string | null, kind: FeedType) => ({
```
jeanregisser: Looks like we don't need the `renderItem` factory indirection anymore (function returning another function).
`commentKey` and `kind` can be taken from props directly.

packages/mobile/src/transactions/TransactionsList.test.tsx
```diff
@@ -294,11 +294,6 @@ it('ignores pending standby transactions that are completed in the response and
   const feed = await waitForElement(() => getByType(TransactionFeed))
   expect(feed.props.data.length).toEqual(2)
   expect(toJSON()).toMatchSnapshot()
-
-  expect(store.getActions()).toEqual([
-    { type: 'HOME/REFRESH_BALANCES' },
-    { type: 'TRANSACTIONS/REMOVE_STANDBY_TRANSACTION', idx: 'a-standby-tx-id' },
-  ])
```
jeanregisser, line 301: nit: we could keep that and check the array is empty.

packages/mobile/src/transactions/reducer.ts
```diff
@@ -85,7 +68,20 @@ export const reducer = (state: State | undefined = initialState, action: ActionT
           }
         }),
       }
+    case Actions.NEW_TRANSACTIONS_IN_FEED:
+      const newKnownFeedTransactions = { ...state.knownFeedTransactions }
+      action.transactions.map((tx) => (newKnownFeedTransactions[tx.hash] = true))
```
jeanregisser: nit: this is more a `forEach`

packages/utils/src/attestations.ts
```diff
@@ -92,6 +95,54 @@ export function extractAttestationCodeFromMessage(message: string) {
...
+  completed: number
+}
+
+interface AttestationStat {
+  completed: number
+  total: number
+}
+
+/**
+ * Returns true if an AttetationStat is considered verified using the given factors,
```
jeanregisser: typo `AttestationStat`

packages/mobile/src/transactions/saga.ts
```diff
@@ -1,25 +1,130 @@
...
+      'Error proccesing user tx query channel event',
+      error
+    )
+  } finally {
+    Logger.debug(`${TAG}/watchUserTransactionQueryChannel`, 'User tx channel terminated')
+  }
+}
+
+// Remove standby txs from redux state when the real ones show up in the feed
+function* cleanupStandbyTransacitons(newFeedTransactions: TransactionFeedFragment[]) {
```
jeanregisser: typo `Transactions`

packages/mobile/src/transactions/saga.ts
```diff
@@ -1,25 +1,130 @@
...
+            transactionFeedFragments: getTxsFromUserTxQuery(queryResult.data),
+          })
+        },
+        error: (e) => console.error(e),
+      })
+
+    return querySubscription.unsubscribe
+  })
+  yield spawn(watchUserTxQueryChannel, userTxQueryChannel)
+}
```
jeanregisser: This feels disconnected from what's happening at the component level.
It looks like it's working for both cUSD and cGLD feeds because there's a link between them.
But this is gonna create differences, and probably one extra query when the local currency is changed for instance.

What do you think of instead of trying to replicate the query here we would dispatch the new transactions from the `onComplete` function at the component level in `TransactionsList`?

packages/mobile/src/identity/commentEncryption.test.ts
```diff
@@ -0,0 +1,259 @@
...
+  'BNFXzyIGjZqqNyq6r35aV2HlMMqUbGnIqboReD77MwAlI5IyzqLQ99WF5B1bsZSVS1K+7trtJtKGhIdI1vbSJSsBAQEBAQEBAQEBAQEBAQEBCVYJWqi/TZNXbAR9ziyX5MJCtfdulxA1tjlvHR/xpE6WnlC/kyXAfKIMqgKGJXchBNFXzyIGjZqqNyq6r35aV2HlMMqUbGnIqboReD77MwAlI5IyzqLQ99WF5B1bsZSVS1K+7trtJtKGhIdI1vbSJSsBAQEBAQEBAQEBAQEBAQEBBhjruDecYg9fsrPNcQbI3AkcvWra1MHIeOZlcycn7Vqtx+UVNR59A3kqdIDbLuGiAQEBAQEBAQEBAQEBAQEBATJ3ip/hhfM0qaIiSahfI8nQa3xtrId9bJV59jBElUbqxKA8zxXyYyTGr/gkHcz1htGoMXyuwrUZoDHHFawgk1046uFz352n+l1rPoFhpHal6vE9k3lR9wF0GN/aWTe2L2g/hp8QQdlwuCgsZnjOVB7hDjSvOZ68mz7WqNk29RT57sb3nMkV'
+
+const phoneDetails: PhoneNumberHashDetails = {
+  e164Number: mockE164Number,
+  salt: mockE164NumberSalt,
+  phoneHash: 'hash',
+}
+
+describe('Encrypt Comment', () => {
+  it('Empty comment', async () => {
```
jeanregisser: nit: no big deal but we are inconsistently naming tests in this file.
From what I've seen, we usually try to use a verb after `it` in most other test files.

jeanregisser (changes requested):
Awesome work!! 💯 

I'm mostly concerned about the disconnect introduced by the Tx query watcher in the saga and would prefer we change it before merging this. See comment below.

packages/mobile/src/transactions/TransactionsList.tsx
jmrossy: Maybe but I'll avoid temptation to make this PR bigger :)

packages/mobile/src/send/saga.ts
jmrossy: Will do!

packages/mobile/src/transactions/saga.ts
jmrossy: Hmm, good question. In theory, this case never happens b.c. if the tx failed, it'll never show up in blockscout and therefore the feed. But I'm not 100% sure so I'd rather keep the logic we had for this (previously in TransactionList)

packages/mobile/src/transactions/TransactionsList.test.tsx
jmrossy: There are no expected store actions now, not sure what you mean

packages/mobile/src/transactions/TransactionFeed.tsx
jmrossy: Good idea!

packages/mobile/src/apollo/types.ts
jmrossy: Will do!

packages/mobile/src/transactions/saga.ts
jmrossy: Can you clarify what you mean by creating differences? I don't follow.

The problem with the onComplete is that
1) It seems to be less reliable than this watcher
2) It forces us to pollute the component with this business logic

As for creating a new query, this will pull data from the cache first so if the component already queried, it won't use additional bandwidth.

packages/mobile/src/identity/commentEncryption.test.ts
jmrossy: Will fix :)

packages/mobile/src/transactions/saga.ts
jmrossy: To clarify what I mean by unreliable, issues #2647, #1335,  #3624 and #2704 shouldn't be happening if the onComplete was working as expected. My suspicion in that only having it run when the component is mounted contributes to these issues.

packages/mobile/src/transactions/TransactionsList.tsx
```diff
@@ -276,14 +272,17 @@ export class TransactionsList extends React.PureComponent<Props> {
         query={TRANSACTIONS_QUERY}
         pollInterval={POLL_INTERVAL}
         variables={{ address: queryAddress, token, localCurrencyCode }}
-        onCompleted={this.txsFetched}
         children={UserTransactions}
+        onCompleted={this.onTxsFetched}
+        // Adding this option because the onCompleted doesn't work properly without it.
+        // It causes the onCompleted to trigger too often but that's okay.
+        // https://github.com/apollographql/react-apollo/issues/2293
+        notifyOnNetworkStatusChange={true}
```
jeanregisser, line 280: 👍

jeanregisser (approved):
Awesome!! 💯 🎉

ValoraQA:
Hey @jmrossy I have verified this issue using Android internal build V1.15.1 & Test flight build V1.15.1 & observed that user is able to receive cUSD balance after completion of transaction successfully.
**Verified On Devices:** Samsung Galaxy J5 Prime(8.0.0), Samsung Galaxy A5 (7.0)
**Attachment:** [AbleToReceivecUSDInUserB'sAccount.wmv](https://drive.google.com/file/d/1FzlzAdstcCIv5VhpJAORW7Kw4oYdWEZi/view?usp=sharing)

PR #3888, opened 2020-05-24
Significant refactor of the GCP Terraform code for deploying validators on mainnet

alchemydc:
### Description
* updated to support RC1/mainnet
* bootstrap.sh makes configuring a separate GCP project for Terraformed Celo infrastructure easy
* Adds Stackdriver support for logging, monitoring and alerting.
* Includes Celo/Geth specific Stackdriver Monitoring Dashboard
Screen Shot 2020-05-23 at 22 13 32

* Fixes Docker start/stop via systemd to avoid chaindata corruption
* Adds support for backup/restore of chaindata via cloud storage to avoid slow p2p sync
* Rotates geth logs (/var/log/daemon.log) to prevent boot disk from filling up
* Adds support for a public static website bucket for hosting metadata, etc.
* Fixes security issues associated with txnode exposing RPC to the Internet
* Moves secrets out of git, and into terraform.tfvars (per custom)
* Fixes to support new mainnet attestation service docker image
* Fixes stats support on proxy node (by unlocking an otherwise unused ethereum account on proxy)

### Tested

Used to successfully run infrastructure on RC1 / mainnet


jcortejoso:
@alchemydc Awesome! Thanks for contributing.

packages/terraform-modules-public/gcp/example/main.tf
```diff
@@ -115,3 +116,110 @@ module "celo_cluster" {
...
+#  }
+#}
+
+#resource "google_storage_bucket_iam_binding" "public_www_binding_read" {
+#  bucket = var.public_www_fqdn
+#  role = "roles/storage.objectViewer"
+#  members = [
+#    "allUsers"
+#  ]
+#}
```
jcortejoso, line 247: Should this be removed?

jcortejoso (approved):
Great improvement from current setup!

packages/terraform-modules-public/gcp/example/main.tf
alchemydc: No.  Validators need to host metadata on a publicly accessible web server, so I enabled this to make it easy to do so on GCS.  Agree it was confusing having it in there but commented so, so it's now enabled.

https://docs.celo.org/celo-codebase/protocol/identity/metadata

alchemydc:
@jcortejoso note that we've updated to use debian10, switched to using rsync for backup and restore of chaindata, and made some security improvements.  Hoping to get this merged sooner or later :)

PR #3892, opened 2020-05-25
Adding bootnodes tool and using it in wallet conf

### LINKED ISSUE #3655: Test full node discovery in the wallet
### Expected Behavior

Ensure the wallet can connect to full nodes

jose-compu:
### Description

1. Adding `contractkit` tool to get list of bootnodes from cloud.
2. Using tool to configure wallet `geth` to using these bootnodes according to each network name.

### Tested

I have an issue building contraction, cannot test enough. Will test with CircleCI.

### Related issues

- Fixes #3655

### Backwards compatibility

Is backward compatible you just have to turn off network discovery putting `USE_FULL_NODE_DISCOVERY` to `false` on `mobile/src/config.ts`


packages/contractkit/src/network-utils/bootnode-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+   * along with the response data from Google api.
+   * @param networkName Name of the network to fetch config for
+   */
+  static getBootnodes(networkName: string): string {
+    const bucketName = BootnodesGoogleStorageBucketName
+    const fileName = networkName
+    let ret = ''
+    GoogleStorageUtils.fetchFileFromGoogleStorage(bucketName, fileName)
+      .then((res) => (ret = res))
+      .catch(() => 'obligatory catch')
```
jmrossy: I think your promise handling here is incorrect. This will simply return the empty string.
A more typical approach would simply be to return the promise returned by `fetchFileFromGoogleStorage` and let the consumer use `async/await` syntax on the result

packages/mobile/src/geth/networkConfig.ts
```diff
@@ -36,6 +44,8 @@ const networkConfigs: { [testnet: string]: NetworkConfig } = {
     blockchainApiUrl: 'https://integration-dot-celo-testnet.appspot.com/',
     pgpnpUrl: pgpnpUrlStaging,
     pgpnpPubKey: pgpnpPubKeyStaging,
+    peerDiscovery: USE_FULL_NODE_DISCOVERY,
+    bootnodeEnodes: BootnodeUtils.getBootnodes(Testnets.integration).split(';'),
```
jmrossy: This is a config which we parse early in the app lifecycle. Given that getBootnodes is async, I don't think this is the right place for this call. Somewhere in `geth/saga.ts` or `geth/geth.ts` would be better

jmrossy (dismissed):

packages/contractkit/src/network-utils/bootnode-utils.ts
jose-compu: Okey, got it.

packages/mobile/src/geth/networkConfig.ts
jose-compu: Cool, Connor told me to move it to be similar to Static Nodes.

packages/mobile/src/geth/geth.ts
```diff
@@ -88,6 +88,13 @@ async function createNewGeth(): Promise<typeof RNGeth> {
     syncMode,
     useLightweightKDF: true,
     ipcPath: IPC_PATH,
+    noDiscovery: !USE_FULL_NODE_DISCOVERY,
+  }
+
+  if (USE_FULL_NODE_DISCOVERY) {
+    const nodes = (await BootnodeUtils.getBootnodesAsync(DEFAULT_TESTNET)).trim()
+    gethOptions.bootnodeEnodes = nodes.split(';')
+    Logger.debug('Geth@newGeth', 'bootnodes = ' + nodes.split(';'))
```
jmrossy, line 97: Nit, consider assigning a var to nodes.split(';') so we don't do it twice.
Also we can use string interpolation here instead of the manual concatenation

jmrossy (commented):
LGTM

barbaraliau:
@joigno is this PR still relevant?

PR #3893, opened 2020-05-25
Check storage layout changes for backwards compatibility

### LINKED ISSUE #3738: Smart contract backwards compatibility checking in CI

m-chrzan:
### Description

This PR introduces a script in the protocol package, `check-storage-layout.ts`, which allows one to check whether existing storage layout of smart contracts wouldn't change across updates.

The script bases on the [@openzeppelin/upgrades](https://www.npmjs.com/package/@openzeppelin/upgrades) library to ensure storage variables weren't reordered/inserted/deleted/etc.

TODOs (some of this work potentially in future PRs):

- [x] Add check to verify that struct defintions have not changed
- [ ] Add check against a network that bytecodes of the original contracts matches what's currently deployed
- [ ] Currently the tool operates on directories of build artifacts - might want to go a step up and operate on Solidity sources.
- [ ] Add to CI

### Tested

Added unit tests for the code checking backwards compatibility.

### Related issues

- Fixes #3738 

### Backwards compatibility

Will improve future backwards compatibility checking!!


nategraf:
Is the intention here genuinely to add nearly 100,000 lines to the codebase? Could this be managed as an `npm` dependency or as a Git submodule?

m-chrzan:
> Is the intention here genuinely to add nearly 100,000 lines to the codebase? Could this be managed as an `npm` dependency or as a Git submodule?

These are artifacts for unit tests. If we're really concerned about the linecount of the repo, I guess we could make them a submodule or something similar, but I'm not sure if we want to introduce this extra complexity to something as basic as unit testing.

They could potentially be slimmed down - the artifacts Truffle generates end up having a *lot* more information than we need.

packages/protocol/lib/layout.ts
```diff
@@ -0,0 +1,213 @@
...
+  errors: string[]
+}
+
+const generateErrorMessage = (operation: Operation) => {
+  let message: string
+
+  const updated = operation.updated
+  const original = operation.original
+
+  switch (operation.action) {
```
asaj, line 97: Any reason `append` isn't here?

packages/protocol/test/layout/layout.ts
```diff
@@ -0,0 +1,207 @@
+import { getBuildArtifacts } from '@openzeppelin/upgrades'
+
+/* HACK! truffle test was unable to compile this test (TypeScript would end up
+ * claiming that it couldn't find the name `assert`) without the following 5
```
asaj: Which 5 lines are the hack?

packages/protocol/test/resources/layout/contracts_appended/Migrations.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.5.3;
```
asaj: Why is this contract needed?

asaj (commented):
Nice work, I'm really looking forward to working on the contract development and release process over the coming weeks and this is a great first step!

None of the test cases cover mappings or arrays types, is there any reason to worry that those won't be handled properly?

packages/protocol/test/layout/layout.ts
m-chrzan: Ah, as typical for a code comment, it's already outdated. It's the *three* lines `require`ing a random contract artifact.

packages/protocol/lib/layout.ts
m-chrzan: `append` operations aren't considered erroneous, I am assuming that `generateErrorMessage` is only called with erroneous operations. 

Happy to add an `append` case for robustness, could then rename this function to `operationToDescription`.

packages/protocol/test/resources/layout/contracts_appended/Migrations.sol
m-chrzan: Ah, thought maybe it was necessary for `truffle compile` but doesn't appear to be so.

m-chrzan:
> None of the test cases cover mappings or arrays types, is there any reason to worry that those won't be handled properly?

Huh, this is annoying. Apparently the OpenZeppelin library I'm using for verifying storage layout erases the input type to a mapping, so it treats for example `mapping(uint256 => uint256)` as equivalent to say `mapping(string => uint256)`. This is their justification for doing so (https://github.com/OpenZeppelin/openzeppelin-sdk/blob/9726f0410de56e57afd1b7fc0d482cb2f9c1bc84/packages/lib/src/validations/Storage.ts#L245):

    
    // We ignore the keyTypeId, since it's always hashed and takes up the same amount of space; we only care about the last value type

Should we take the same approach or patch this library to be more conservative?

bowd:
I think the point here is really fair, in the sense that indeed the Storage layout doesn't strictly change in that situation. But the _intent_ and _access pattern_ clearly changes. So the way I see it it depends on how we phrase backwards compatibility:

Do we want to detect (a) backwards compatibility issues in how the _code reasons_ about the storage, or (b) specific changes to the storage layout. I think (a) makes more sense in this case, which would imply that the key type change is a problem irrespective of the underlying storage layout not actually changing.

So this might be also caused by an issue with clarity of intent in the library. This might not be a good example as I don't have a full intuition about the problem space, but I'm wondering what happens if I have a change from:

```solidity
bytes32 storage variableOne = 0x4d616b65724d4343000000000000000000000000000000000000000000000000;
bytes32 storage variableTwo = 0x4d616b65724d4344000000000000000000000000000000000000000000000000;
```

to

```solidity
bytes32 storage variableTwo = 0x4d616b65724d4343000000000000000000000000000000000000000000000000;
bytes32 storage variableOne = 0x4d616b65724d4344000000000000000000000000000000000000000000000000;
```

> I'm working under the assumption that storage locations are allocated in order of how variables are defined. But I don't know the EVM that well, so I might be screwing up my assumptions.

This would still, strictly speaking, not be a change to the underlying storage structure, but it would again be a change in how the codes maps to/reasons about the underlying structure. Would the library flag this as an issue? If yes than logically it should see the key change as an issue as well.

m-chrzan:
> I'm working under the assumption that storage locations are allocated in order of how variables are defined. But I don't know the EVM that well, so I might be screwing up my assumptions.

Yep, that's the correct assumption. If you're curious about all the details of storage layout, the [Solidity documentation](https://solidity.readthedocs.io/en/v0.5.11/miscellaneous.html#layout-of-state-variables-in-storage) has a good section on it.

> This would still, strictly speaking, not be a change to the underlying storage structure, but it would again be a change in how the codes maps to/reasons about the underlying structure. Would the library flag this as an issue? If yes than logically it should see the key change as an issue as well.

Great point. The OZ library would flag this as a "rename", which the library implemented in this PR treats as an error (specifically to deal with a situtation like in your example - renames could be reorderings of variables, which is backwards incompatible)

bowd:
Well if the lib does flag renames but not mapping keys that's clearly inconsistent logically so we should try to solve it upstream. In the meantime we could move forward with this accepting the quirks.

packages/protocol/lib/layout.ts
```diff
@@ -0,0 +1,216 @@
...
+      message = `variable ${original.label} was removed`
+      break
+    case 'rename':
+      message = `variable ${updated.label} was renamed from ${original.label}`
+      break
+    case 'append':
+      message = `variable ${updated.label} was appended`
+
+    default:
+      message = `unknown operation ${operation.action}`
```
asaj, line 117: Should we include the label here as well?

packages/protocol/lib/layout.ts
```diff
@@ -0,0 +1,216 @@
...
+    return {
+      same: false,
+      errors: [`struct ${newType.label} has changed members`]
+    }
+  }
+
+  const memberErrors = newType.members.map((newMember, i) => {
+    const oldMember = oldType.members[i]
+    if (oldMember.label !== newMember.label) {
+      return `struct ${newType.label} has new member ${newMember.label}`
```
asaj: Is this accurate? What is `newMember.label` here? Would the following be more accurate?

`struct ${newType.label} had ${oldMember.label} in slot ${i}, now has ${newMember.label}`

packages/protocol/lib/layout.ts
```diff
@@ -0,0 +1,216 @@
...
+
+const generateStructsCompatibilityReport = (oldLayout: StorageLayoutInfo, newLayout: StorageLayoutInfo) => {
+  let compatible = true
+  let errors = []
+
+  Object.keys(newLayout.types).forEach(typeName => {
+    const newType = newLayout.types[typeName]
+    const oldType = oldLayout.types[typeName]
+
+    if (newType.kind === 'struct' && oldType !== undefined) {
```
asaj, line 181: Is it always true that if newType.kind === 'struct' && oldType === undefined, it's an append?

packages/protocol/test/layout/layout.ts
```diff
@@ -0,0 +1,297 @@
...
+  typechange_in_struct: getTestArtifacts('typechange_in_struct'),
+  typechange_in_parent: getTestArtifacts('typechange_in_parent'),
+  typechange_in_library_struct: getTestArtifacts('typechange_in_library_struct'),
+  removed_from_struct: getTestArtifacts('removed_from_struct'),
+  removed_from_parent: getTestArtifacts('removed_from_parent'),
+  inserted_in_struct: getTestArtifacts('inserted_in_struct'),
+  inserted_in_library_struct: getTestArtifacts('inserted_in_library_struct'),
+  removed_from_library_struct: getTestArtifacts('removed_from_library_struct'),
+
+  original_complex: getTestArtifacts('original_complex'),
```
asaj, line 41: Why have `original` and `original_complex`?

packages/protocol/test/layout/layout.ts
```diff
@@ -0,0 +1,297 @@
...
+      const report = reportLayoutIncompatibilities(
+        testCases.original_complex,
+        testCases.dynamic_to_fixed_array
+      )
+      assertNotCompatible(report)
+      assertContractErrorsMatch(report, 'TestContract', [/had type/])
+    })
+  })
+
+  // TODO(m-chrzan): @openzeppelin/upgrades erases information about mapping key
```
asaj, line 261: I was very confused until I saw `describe.skip` 😄 

+1 to patching this behavior, as changing the key type should be considered backwards incompatible. I think it's fine to rely on PR reviews and audits to catch these changes for now.

asaj (approved):

asaj:
Super thorough test cases, nice work! 👏 👏 👏 

If we want to add/modify test cases in the future, how do we do so?

packages/protocol/test/layout/layout.ts
m-chrzan: I added tests with "complex" types (mappings and arrays) later. Rather than rewriting/recompiling all the previous tests I just created a fresh starting point for these tests.

m-chrzan:
> If we want to add/modify test cases in the future, how do we do so?

Good question. I was doing it kind of ad-hoc manually, but looks like @hbandura added some nice test tooling around this in https://github.com/celo-org/celo-monorepo/pull/4277/files#diff-0b0e0bfa67d32019ef51c73dba1e6887R45. If I understand correctly, would just need to create a new contracts directory in the test resources directory and use `getTestArtifacts` in code, which will compile the test case in the background if needed.

hbandura:
> > If we want to add/modify test cases in the future, how do we do so?
> 
> Good question. I was doing it kind of ad-hoc manually, but looks like @hbandura added some nice test tooling around this in https://github.com/celo-org/celo-monorepo/pull/4277/files#diff-0b0e0bfa67d32019ef51c73dba1e6887R45. If I understand correctly, would just need to create a new contracts directory in the test resources directory and use `getTestArtifacts` in code, which will compile the test case in the background if needed.

Yes. Just add a contract folder and refer to it, the getTestArtifacts function will compile it before returning the reference.


I tried to avoid adding the compiled solidity files for these reasons:
1) The tool has to work with the compiler we are using. The compiled files are frozen to whatever compiler was used when they were created. 
2) Feels like they are too big to be commited (personal opinion)

Of course, cons:
1) Depends on compiler version used by the client
2) Adds time to the tests run.

packages/protocol/lib/layout.ts
m-chrzan: Since it's an unknown operation, it's not clear which, if any label is defined. This shouldn't happen anyways, but one could imagine an unkown operation being "appendy" where the original label is undefined, or "deletey" where the updated label is undefined.

packages/protocol/lib/layout.ts
m-chrzan: This part of the code isn't concerned with the layout of the variables themselves in storage, only checks that a type name refers to the exact same types in both the original and new contract.

asaj: What would it mean for newType.kind === 'struct' && oldType === undefined?

m-chrzan: That `newType` is a struct type that wasn't defined in the old contract. A variable of this type could have been for example appended or typechanged - in either case that will be caught by the variable layout checker.

PR #3894, opened 2020-05-25
Switch wallet to drawer navigation

cmcewen:
### Description

Implement drawer navigation in the wallet

Screenshot_1590442249
Screenshot_1590442253

### Tested

Playing around with the app and screens


packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -80,16 +76,27 @@ class FiatExchange extends React.Component<Props, State> {
   }
 
   render() {
-    return this.state.signedUrl === '' ? (
-      <ActivityIndicator size="large" color={colors.celoGreen} />
-    ) : (
-      <WebView style={styles.exchangeWebView} source={{ uri: this.state.signedUrl }} />
+    return (
+      <SafeAreaView style={styles.background}>
+        <DrawerTopBar />
```
aaronmgdr, line 82: I notice you are removing navigationOptions often and placing the topbar directly inside of the component. 

why?

cmcewen: The drawer navigator unfortunately doesn't have a header bar, so we could either put a new stack inside each drawer, or just do this - which I think is more performant and easier

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -80,16 +76,27 @@ class FiatExchange extends React.Component<Props, State> {
...
+        )}
+      </SafeAreaView>
     )
   }
 }
 
 const styles = StyleSheet.create({
   exchangeWebView: {},
+  background: {
+    backgroundColor: 'white',
```
jmrossy: Small nit: `colors.background`

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -80,16 +76,27 @@ class FiatExchange extends React.Component<Props, State> {
...
+    flex: 1,
+    flexDirection: 'column',
```
jmrossy: Column is the default

packages/mobile/src/icons/navigator/AccountKey.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react'
+import Svg, { Circle, Path } from 'react-native-svg'
+
+function AccountKey() {
```
jmrossy: We use PureComponents for our other icons, which seems like a good idea for SVGs like these.
Can you change this to a PureComponent or React.Memo it?

packages/mobile/src/icons/navigator/AddWithdraw.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react'
+import Svg, { Path } from 'react-native-svg'
+
+function AddWithdraw() {
```
jmrossy: ditto for this and the next few files

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const symbol = useLocalCurrencySymbol()
+  const { t } = useTranslation(Namespaces.global)
+  const account = useSelector(currentAccountSelector)
+  const appVersion = deviceInfoModule.getVersion()
+
+  return (
+    <DrawerContentScrollView {...props}>
+      <View style={styles.drawerTop}>
+        <ContactCircle thumbnailPath={contactDetails.thumbnailPath} size={64} />
+        <Text style={[fontStyles.large600, styles.nameLabel]}>{displayName}</Text>
```
jmrossy: Nit: Can you fold the styles into `styles.nameLabel` via destructuring to avoid these style arrays. Not a big deal either way.

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -0,0 +1,196 @@
...
+        <PhoneNumberWithFlag
+          e164PhoneNumber={e164PhoneNumber}
+          defaultCountryCode={defaultCountryCode}
+        />
+        <View style={styles.border} />
+        <Text style={fontStyles.regular500}>{`${symbol} ${localBalance?.toFixed(2)}`}</Text>
+        <Text style={[styles.smallLabel, styles.dollarsLabel]}>{`${bigNumBalance?.toFixed(2)} ${t(
```
jmrossy, line 70: ditto

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -0,0 +1,196 @@
...
+    <DrawerContentScrollView {...props}>
+      <View style={styles.drawerTop}>
+        <ContactCircle thumbnailPath={contactDetails.thumbnailPath} size={64} />
+        <Text style={[fontStyles.large600, styles.nameLabel]}>{displayName}</Text>
+        <PhoneNumberWithFlag
+          e164PhoneNumber={e164PhoneNumber}
+          defaultCountryCode={defaultCountryCode}
+        />
+        <View style={styles.border} />
+        <Text style={fontStyles.regular500}>{`${symbol} ${localBalance?.toFixed(2)}`}</Text>
```
jmrossy, line 69: Can you use the existing money components for this instead of formatting it yourself. It'd be good to avoid the mess that @jeanregisser recently got us out of where money values are formatted 30 diff ways in diff places and don't handle local currency right.

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -0,0 +1,86 @@
+import colorsV2 from '@celo/react-components/styles/colors.v2'
+import { elevationShadowStyle } from '@celo/react-components/styles/styles'
+import { useNavigation } from '@react-navigation/native'
+import * as React from 'react'
+import { StyleSheet, TouchableOpacity, View } from 'react-native'
+import { Line, Svg } from 'react-native-svg'
+import Logo from 'src/icons/Logo.v2'
+
+const Hamburger = () => {
```
jmrossy: Should this get its own file? Also memo

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+        y2="21.75"
+        stroke="#2E3338"
+        strokeWidth="2.5"
+        strokeLinecap="round"
+      />
+    </Svg>
+  )
+}
+
+const shadow = elevationShadowStyle(1)
```
jmrossy: why does this have to be outside the styles object?

jmrossy (commented):
Really great work here! Just some small comments

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const styles = StyleSheet.create({
+  container: {
+    flexDirection: 'row',
+    height: 62,
+    backgroundColor: colorsV2.white,
+    alignItems: 'flex-end',
+    justifyContent: 'space-between',
+    // borderBottomWidth: 1,
+    // borderBottomColor: colorsV2.gray2,
```
annakaz: Nit: remove comments

packages/react-components/components/PhoneNumberWithFlag.tsx
```diff
@@ -22,8 +23,8 @@ export class PhoneNumberWithFlag extends React.PureComponent<Props> {
               )
             : getCountryEmoji(this.props.e164PhoneNumber)}
         </Text>
-        <Text style={fontStyles.telephoneHeadline}>
-          {(parsedNumber && parsedNumber.displayNumber) || ''}
+        <Text style={[fontStyles.regular, style.phoneNumber]}>
+          {(parsedNumber && `+${parsedNumber.countryCode} ${parsedNumber.displayNumber}`) || ''}
```
annakaz: Nit: I would find this more readable as `parsedNumber ? `+${parsedNumber.countryCode} ${parsedNumber.displayNumber}`) : ''`

annakaz (approved):
New menu looks great. A few nits but LGTM!

packages/mobile/src/icons/navigator/AccountKey.tsx
cmcewen: I tried doing this but react navigation doesn't like it since it's passed an option. I did `export default React.memo` though in case they are used anywhere else

packages/mobile/src/navigator/DrawerNavigator.tsx
cmcewen: I looked at doing this using `CurrencyDisplay` but it seems like it adds a lot of complexity. @jeanregisser curious if you think it's necessary here - from my understanding we will always show both the celo dollars balance and the local amount, so i am not sure what edge cases would be possible (but maybe i am missing something)

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -28,13 +28,13 @@ import BackupIntroduction from 'src/backup/BackupIntroduction'
...
 import WalletHome from 'src/home/WalletHome'
 import { Namespaces } from 'src/i18n'
-import AccountKey from 'src/icons/navigator/AccountKey'
-import AddWithdraw from 'src/icons/navigator/AddWithdraw'
-import Gold from 'src/icons/navigator/Gold'
-import Help from 'src/icons/navigator/Help'
-import Home from 'src/icons/navigator/Home'
-import Invite from 'src/icons/navigator/Invite'
-import Settings from 'src/icons/navigator/Settings'
+import { AccountKey } from 'src/icons/navigator/AccountKey'
```
aaronmgdr, line 31: why import the non memorized version?

packages/mobile/src/navigator/Hamburger.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react'
+import { Line, Svg } from 'react-native-svg'
+
+const Hamburger = () => {
```
aaronmgdr, line 4: I think @tarikbellamine also made an hamburger so eventually we can consolidate. 

also dont we usually prefer named function to arrow for standalone function?

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -1,15 +1,17 @@
 import colors from '@celo/react-components/styles/colors'
```
tarikbellamine, line 1: You may be able to just delete this? I think colors.v2 is backwards compatible but logs a warning if you are using a deprecated property

aaronmgdr: it's mostly  compatible but it is missing some old colors. 

But still tarik has a point.

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -80,16 +77,26 @@ class FiatExchange extends React.Component<Props, State> {
...
+        )}
+      </SafeAreaView>
     )
   }
 }
 
 const styles = StyleSheet.create({
   exchangeWebView: {},
+  background: {
+    backgroundColor: colorsV2.background,
```
tarikbellamine, line 96: @aaronmgdr correct me if I am wrong, but should we be using `light` instead of `background`? Does it matter?

aaronmgdr: technically background is an alias for `light` aka `#fff` now right now it logs that its deprecated but doesn't actually cause any harm. 

if we want to keep a `colors.background` we could remove the loging. I can see that being nice. 

thoughts ?

tarikbellamine: I think `background` is more intuitive but have been using `light` everywhere because I saw the deprecation notice

packages/mobile/src/navigator/DrawerNavigator.tsx
tarikbellamine: Yeah, think you may need to drop the destructuring or just not export the non-memoized functions

packages/mobile/src/icons/navigator/AccountKey.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react'
+import Svg, { Circle, Path } from 'react-native-svg'
+
+export function AccountKey() {
```
tarikbellamine, line 4: Does this need to be exported (i.e., do we ever want to use the non-memoized version)?

packages/mobile/src/navigator/Hamburger.tsx
tarikbellamine: Indeed I did

packages/utils/src/phoneNumbers.ts
```diff
@@ -37,7 +37,7 @@ export function getCountryEmoji(
   const userCountryArray = countries.filter((c: any) => c.alpha2 === regionCode)
   const country = userCountryArray.length > 0 ? userCountryArray[0] : undefined
 
-  return (country ? country.emoji : '') + ` +${countryCode}`
```
tarikbellamine, line 42: Why don't we need `countryCode` anymore?

tarikbellamine (commented):
Good stuff! Excited to get this merged

packages/mobile/src/navigator/DrawerNavigator.tsx
jmrossy: Another option is to use the formatting utils in https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/utils/formatting.ts#L7

jeanregisser: +1 for using `<CurrencyDisplay />` here too, we can adapt https://github.com/celo-org/celo-monorepo/blob/6139ec275bc261564e581168b95b112d54b87451/packages/mobile/src/home/CeloDollarsOverview.tsx#L17-L48

aaronmgdr:
looks to be a lot more in the diff than just wallet changes?

PR #3895, opened 2020-05-26
[Wallet] Redesign of "Chose recipient" and "Enter amount" screens

### LINKED ISSUE #3911: New Design for Send Part 1 (till the TX review)

i1skn:
### Description

Implement choose Recipient screen and Amount enter screen for Send and Request flows. Search input is excluded for now and will updated later

### Tested

Android emulator and iOS simulator

### Related issues

- Fixes #3911

### Backwards compatibility

Yes

### Screenshots

Screenshot_1590434295
Screenshot_1590514745
Simulator Screen Shot - iPhone 11 Pro - 2020-05-27 at 02 16 55
Simulator Screen Shot - iPhone 11 Pro - 2020-05-27 at 02 17 26


packages/mobile/src/send/SendAmount.tsx
```diff
@@ -1,254 +1,217 @@
...
-  fetchDollarBalance: typeof fetchDollarBalance
-  showMessage: typeof showMessage
-  showError: typeof showError
-  hideAlert: typeof hideAlert
-  fetchAddressesAndValidate: typeof fetchAddressesAndValidate
-}
+  React.useEffect(() => {
+    dispatch(fetchDollarBalance())
+    if (recipient.e164PhoneNumber) {
+      dispatch(fetchAddressesAndValidate(recipient.e164PhoneNumber))
```
i1skn: This leads to "Enter PIN" screen appearing right when user is shown "Enter Amount" screen. This is not the best user experience, but I do not know how to solve it. @jmrossy is that something related to PGPNP, right?

packages/mobile/src/account/DollarEducation.tsx
```diff
@@ -20,7 +20,7 @@ export class DollarEducation extends React.Component<Props> {
   goToSend = () => {
     this.props.setEducationCompleted()
     CeloAnalytics.track(CustomEventNames.send_dollar_nux)
-    navigate(Screens.Send)
+    navigate(Screens.Send, {})
```
aaronmgdr: I think if params are never needed it can be re typed to accept undefined so you dont need empty object

@cmcewen  is that right?

packages/mobile/src/navigator/Headers.v2.tsx
```diff
@@ -77,6 +89,83 @@ export const emptyHeader: StackNavigationOptions = {
...
+  route: RouteProp<StackParamList, Screens.ExchangeTradeScreen>
+}) => {
+  const { makerToken } = route.params?.makerTokenDisplay
+  const title =
+    makerToken === CURRENCY_ENUM.DOLLAR
+      ? i18n.t('exchangeFlow9:buyGold')
+      : i18n.t('exchangeFlow9:sellGold')
+  return {
+    ...headerWithCancelButton,
+    headerLeft: () => <CancelButton style={{ color: colors.dark }} />,
```
aaronmgdr: would be nice to uninline this style

packages/mobile/src/account/DollarEducation.tsx
i1skn: My bad, will add `| undefined` in `navigator/types.tsx`

packages/mobile/locales/en-US/sendFlow7.json
```diff
@@ -105,5 +105,6 @@
     "body1": "Your contact can find their account number (also known as an address) at the top of the Settings menu.",
     "body2": "Account numbers begin with “0x” and are 42 characters long. They appear directly beneath the profile photo."
   },
-  "transferAddressChanged": "This contact changed their account number since the transaction. This payment was sent to:"
+  "transferAddressChanged": "This contact changed their account number since the transaction. This payment was sent to:",
+  "review": "Review"
```
jmrossy: Consider putting one-word strings like this in `global.json` for easier use elsewhere

packages/mobile/src/navigator/Headers.v2.tsx
```diff
@@ -77,6 +92,83 @@ export const emptyHeader: StackNavigationOptions = {
...
+  route: RouteProp<StackParamList, Screens.ExchangeTradeScreen>
+}) => {
+  const { makerToken } = route.params?.makerTokenDisplay
+  const title =
+    makerToken === CURRENCY_ENUM.DOLLAR
+      ? i18n.t('exchangeFlow9:buyGold')
+      : i18n.t('exchangeFlow9:sellGold')
+  return {
+    ...headerWithCancelButton,
+    headerLeft: () => <CancelButton style={styles.cancelBtn} />,
```
jmrossy: Nit, should cancelBtn just be the default style of this component?

packages/mobile/src/navigator/Headers.v2.tsx
```diff
@@ -77,6 +92,83 @@ export const emptyHeader: StackNavigationOptions = {
...
+        testID="EditButton"
+        onPress={navigateBack}
+        titleStyle={{ color: colors.goldDark }}
+      />
+    ),
+    headerTitle: () => <HeaderTitleWithBalance title={title} token={makerToken} />,
+  }
+}
+
+export const sendOptions = ({ route }: { route: RouteProp<StackParamList, Screens.Send> }) => {
```
jmrossy: I think putting these in this file made sense when they were shared. But if they're going to be specific to one screen like these new ones, it'd be better to colocate it with the component. WDYT?

packages/mobile/src/send/Send.tsx
```diff
@@ -187,6 +183,7 @@ class Send extends React.Component<Props, State> {
 
   onSelectRecipient = (recipient: Recipient) => {
     this.props.hideAlert()
+    const isRequest = this.props.route.params?.isRequest ?? false
```
jmrossy, line 186: Nice use of the `?` operators here ;)

packages/react-components/icons/Times.tsx
```diff
@@ -0,0 +1,38 @@
...
+import * as React from 'react'
+import Svg, { Path } from 'svgs'
+
+export interface Props {
+  height?: number
+  color?: string
+  strokeWidth?: number
+}
+
+function Times({ color, height, strokeWidth }: Props) {
```
jmrossy, line 11: Same comment I left on Connor's last PR: I think icons should be PureComponents or React.Memo'ed

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -1,254 +1,217 @@
...
-)
-const CommentInput = withTextInputLabeling<TextInputProps>(TextInput)
+type RouteProps = StackScreenProps<StackParamList, Screens.SendAmount>
+type Props = RouteProps
 
-interface State {
-  amount: string
-  reason: string
-}
+const { decimalSeparator } = getNumberFormatSettings()
```
jmrossy, line 66: Does this change if a user changes their language via settings?

packages/mobile/src/send/SendAmount.tsx
jmrossy: Yes, this is unfortunately necessary atm because PGPNP uses the wallet key to sign an auth token. We can explore alternatives in the future for using the DEK instead. @nityas and I talked about this already so she should be aware. It's essentially just bumping the pin up here instead of after review.

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -1,254 +1,217 @@
...
 
-interface DispatchProps {
-  fetchDollarBalance: typeof fetchDollarBalance
-  showMessage: typeof showMessage
-  showError: typeof showError
-  hideAlert: typeof hideAlert
-  fetchAddressesAndValidate: typeof fetchAddressesAndValidate
-}
+  React.useEffect(() => {
+    dispatch(fetchDollarBalance())
```
jmrossy, line 75: do we want to do this each time?

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -1,254 +1,217 @@
...
+    if (recipient.e164PhoneNumber) {
```
jmrossy: I believe there's a bug here, possibly from @tarikbellamine's changes for SecureSend.
We used to check if the recipient type was address or qrcode and skip the lookup in those cases, but that check seems to be gone:
https://github.com/celo-org/celo-monorepo/blob/6e6a9e5105796286064c4c7a9f903137d7d674d3/packages/mobile/src/send/SendAmount.tsx#L163

jmrossy: @tarikbellamine do we handle this case somewhere else now?

jmrossy (commented):
Mostly looks good! Some questions below

packages/mobile/src/send/SendAmount.tsx
tarikbellamine: My overarching reasoning was that we should fetch the latest mapping whenever we can as it's better to be safe that sorry with SecureSend.
- The false positives will at most require 2 or 3 more clicks and an extra 10-20 seconds to go through SecureSend as they already have their QR code or address ready to verify. Also worth noting that recipients caught here will have to go through SecureSend eventually anyways, we are just pulling the process forward to a transaction that may have not needed it.
- A false negative is a loss of funds. Admittedly, I don't know enough about how we assign recipient types to say how likely I think a false negative would be.

Given the uncertainty and the relatively low cost of playing it safe, that's the direction I went in. What do you think I could be missing?

packages/mobile/src/send/SendAmount.tsx
i1skn: Yes, but the app needs to be restarted. Essentially, even if it was not restarted - it would show previous `decimalSeparator`, but would be consistent around the app.

packages/mobile/src/send/SendAmount.tsx
i1skn: No, only once, hence I set the second parameter to `[]`

packages/mobile/src/navigator/Headers.v2.tsx
i1skn: Yeah, I've been back and forth with this :) First did it in place, then decided to move them to `Headers`. But now, when you put this way I see it makes sense to have them in-place 👍

packages/react-components/icons/Times.tsx
i1skn: Thanks!

packages/mobile/src/fees/selectors.ts
```diff
@@ -29,19 +29,21 @@ export const getReclaimEscrowFeeEstimateDollars = feeEstimateDollarsSelectorFact
   getReclaimEscrowFeeEstimateInWei
 )
 
-export const getFeeEstimateDollars = (state: RootState, feeType: FeeType | null) => {
-  if (feeType === null) {
-    return undefined
-  }
+export const getFeeEstimateDollars = (feeType: FeeType | null) => {
```
aaronmgdr: while youre changing this anyway want to make it into a named function? 

Thats the guideline we have right? standalone functions named. class methods as arrows?

packages/mobile/src/navigator/types.tsx
```diff
@@ -112,9 +112,14 @@ export type StackParamList = {
   [Screens.RegulatoryTerms]: undefined
   [Screens.Security]: undefined
   [Screens.SelectLocalCurrency]: undefined
-  [Screens.Send]: undefined
+  [Screens.Send]:
+    | {
```
aaronmgdr, line 116: what does the pipe before the type do? between them its like an OR but before?

i1skn: Yeah, Prettier adds it automatically. It is convenient for reading. 
```
...:
  | a
  | b
```
instead of
```
...: a 
  | b
```

packages/mobile/src/send/SendAmount.tsx
jmrossy: Sorry, I should have been more clear. Do we want to do this every time we load the SendAmount screen? Now that I think about it more I guess it's okay.

packages/mobile/src/send/SendAmount.tsx
jmrossy: That's all reasonable but we may still want to skip for these cases.
In the `address` case, I don't think we'll have e164Number set so we're probably fine either way.
In the QRCode case, my preference is that QR codes *always* work seamlessly. If we force a lookup in that case, we're vulnerable to a whole bunch of possible issues like:
1. PGPNP being down/unavailable
2. User being out of PGPNP lookups and having to buy more first
3. Recipient has incorrect/outdated data on-chain

QR code should take precedence over the on-chain mappings IMO because the user know who they're interacting with.

tarikbellamine: I see. I wasn't aware of the costs associated with doing a lookup and agree that it is preferable to have one method (QR scanning) that always works, no matter what.

In that case, I am good with reverting back to the old conditional logic. From a process perspective, does it make more sense for me to open up a quick PR or let @i1skn revert that one line?

i1skn: or just propose a change right in this PR using 

Screenshot 2020-05-27 at 20 19 23

tarikbellamine: Thanks for the heads up on that feature. Looking at it a bit more, I think it may make more sense to submit a quick PR because it requires importing `Recipient` type and I'd feel better if I could also run the tests with the augmented logic to make sure there aren't downstream issues.

Sorry for the inconvenience!

tarikbellamine: FYI it's been [merged into master](https://github.com/celo-org/celo-monorepo/pull/3915)

packages/mobile/src/recipients/RecipientItem.tsx
```diff
@@ -24,55 +26,50 @@ class RecipientItem extends React.PureComponent<Props> {
     const { recipient } = this.props
 
     return (
-      <TouchableHighlight onPress={this.onPress} underlayColor={colors.altDarkBg}>
-        <View style={style.row}>
+      <Touchable onPress={this.onPress}>
```
jeanregisser, line 29: Nothing to do now but I think we should have a `TouchableItem` component to use in scrollable list of items that would use `TouchableHighlight` on iOS and something based on `TouchableNativeFeedback` on Android, to respect platform conventions.

`Touchable` alone currently uses `TouchableOpacity` on iOS and `TouchableNativeFeedback` on Android.

packages/react-components/icons/Times.tsx
jeanregisser, line 11: nit: Would "Close" be a more descriptive name for this icon?

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -1,254 +1,222 @@
...
 
-interface State {
-  amount: string
-  reason: string
-}
+const { decimalSeparator } = getNumberFormatSettings()
 
-type OwnProps = StackScreenProps<StackParamList, Screens.SendAmount>
-type Props = StateProps & DispatchProps & OwnProps & WithTranslation
+function SendAmount(props: Props) {
```
jeanregisser, line 68: Nice refactor into a functional component 💪

jeanregisser (approved):
Awesome!! 👍

packages/react-components/icons/Times.tsx
i1skn: Usually I see this icon under "times" name - https://fontawesome.com/icons?d=gallery&q=times

PR #3896, opened 2020-05-26
Improve BackupQuiz Account Key Fetching

aaronmgdr:
### Description

🔐 fetch account key mneumonic directly from store so it doesnt get passed thru nav params

### Tested

changes to tests are because it is now async. 

### Backwards compatibility

yes


packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -82,21 +84,32 @@ export class BackupQuiz extends React.Component<Props, State> {
     this.setBackSpace()
   }
 
-  componentDidMount() {
-    const mnemonic = this.getMnemonicFromNavProps()
+  componentDidMount = async () => {
+    const mnemonic = await this.retrieveMnemonic()
     const shuffledMnemonic = getShuffledWordSet(mnemonic)
```
jmrossy: Should you check for null/empty here since that's what the method returns on error

packages/mobile/test/values.ts
```diff
@@ -13,6 +13,7 @@ import {
   RecipientWithContact,
   RecipientWithMobileNumber,
 } from 'src/recipients/recipient'
+import { Screens } from 'src/navigator/Screens'
```
jmrossy: Why dis?

aaronmgdr: mistake :)

packages/mobile/src/backup/BackupQuiz.tsx
aaronmgdr: hmm a good question. I wasn't sure exactly how to handle the empty case. right now retrieveMnemonic returns an empty string if there is an error.

I guess it probably is better to short circuit if there isnt a mneumonic aye?

aaronmgdr: ok looking that the code more it seems very very very unlikely mnemonic will ever be blank. we already Log and throw errors in several places when it is. I don't think there is much to be gained from doing an empty string check here.

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -82,21 +84,32 @@ export class BackupQuiz extends React.Component<Props, State> {
...
+      return mnemonic
+    } catch {
+      const message = 'Failed to retrieve mnemonic'
+      Logger.error('BackupQuiz/retrieveMnemonic', message)
+      this.props.showError(ErrorMessages.FAILED_FETCH_MNEMONIC)
+      CeloAnalytics.track(CustomEventNames.backup_error, {
+        title: message,
+        component: 'BackupQuiz',
+      })
+      return ''
```
i1skn: I do not understand why we return `''` in case of an error?

aaronmgdr: basically that was just to get typescript to not complain about type being undefined. 

would love to talk about handling the error case cause at the moment feels like there are a lot of `if`s and `trys` and its getting out of hand

jmrossy (approved):
lgtm :)

PR #3901, opened 2020-05-26
[PGPNP] Split service into combiner and signer

### LINKED ISSUE #3897: Split PGPNP package into two
Create two packages, one for the combiner and another for the signer. Proposed structure: 
/packages/phone-number-privacy/combiner
Package name: ‘@celo/pgpnp-combiner’
/packages/phone-number-privacy/signer
Package name: ‘@celo/pgpnp-signer’

jmrossy:
### Description

Split up the PGPNP service because we'll need two unique services going forward: the combiner and the signer.
The combiner will orchestrate multiple signers and (for now) handles matchmaking.

*No significant logic or code changes, just copied files and a few package.json tweaks*

### Tested

Did a lint, build, test and run of the packages. Also confirmed that lerna jobs from monorepo root (like `yarn lint`) run on both these new packages.

### Related issues

- Fixes #3897 

### Backwards compatibility

Not really


packages/phone-number-privacy/combiner/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@celo/phone-number-privacy-combiner",
```
jmrossy, line 2: The only change here is the name and description, not sure why git didn't recognize it as a rename

packages/phone-number-privacy/signer/scripts/createBlsKeyPair.ts
```diff
@@ -0,0 +1,15 @@
+// tslint:disable: no-console
```
aslawson, line 1: it appears like a lot of files were duplicated in both directories.  can't they just reference some shared files? this seems like a maintainability nightmare

in some cases I don't think the combiner needs all these files, or it just needs parts of it (though it would be tedious and harder to reverse to separate logic needed from each service so I feel a shared file would be better).

aslawson (commented):
> No new logic or code changes, just copied files and a few package.json tweaks

> The combiner orchestrates multiple signers and (for now) handles matchmaking.

Where does the combiner orchestrate the multiple signers?  Or is the intent to handle that part in future PR?

packages/phone-number-privacy/signer/scripts/createBlsKeyPair.ts
jmrossy: Yep I intend to clean things up as we chisel down the functionality in each package to what it actually needs. I did a dumb copy here intentionally to keep this large PR easy to review :)

jmrossy:
> > No new logic or code changes, just copied files and a few package.json tweaks
> 
> > The combiner orchestrates multiple signers and (for now) handles matchmaking.
> 
> Where does the combiner orchestrate the multiple signers? Or is the intent to handle that part in future PR?

Future PR yes. I wanted to keep this PR simple since its big

packages/phone-number-privacy/signer/scripts/createBlsKeyPair.ts
aslawson: gotcha -- step one of many then :)

aslawson (approved):
looks fine to me as first step

PR #3904, opened 2020-05-27
[Wallet] Redesign of SendConfirmation and Secure Send screens

tarikbellamine:
### Description
Redesign for the 2nd half of the Send flow and the Secure Send screens minus the QR scanner.

### Other changes
None

### Tested
iOS device

### Related issues

- Part of #3645 

### Backwards compatibility
Yes

secure send obf

IMG_4148

IMG_4150

IMG_4154

IMG_4155

IMG_4161

IMG_4151

IMG_4152


packages/mobile/src/alert/reducer.ts
```diff
@@ -3,6 +3,7 @@ import { ErrorMessages } from 'src/app/ErrorMessages'
 import { RootState } from 'src/redux/reducers'
 
 export interface State {
+  errorType: 'show' | 'log'
```
tarikbellamine: Adding a new `errorType` so that we can log errors for use with inline error messaging without having the error banner pop up

packages/mobile/src/web3/gas.ts
```diff
@@ -32,8 +32,10 @@ export async function getGasPrice(currency: CURRENCY_ENUM = CURRENCY_ENUM.DOLLAR
 
 async function fetchGasPrice(currency: CURRENCY_ENUM) {
   const contractKit = await getContractKitOutsideGenerator()
-  const gasPriceMinimum = await contractKit.contracts.getGasPriceMinimum()
-  const address = await getCurrencyAddress(currency)
+  const [gasPriceMinimum, address] = await Promise.all([
```
tarikbellamine, line 35: random optimization opportunity i spotted

packages/mobile/src/web3/utils.ts
```diff
@@ -36,7 +36,7 @@ export function* isAccountLocked(address: string) {
   try {
     // Test account to see if it is unlocked
     const contractKit = yield call(getContractKit)
-    yield call(contractKit.eth.sign, '', address)
+    yield call(contractKit.web3.eth.sign, '', address)
```
tarikbellamine, line 39: this was a bug that was resulting in invite transaction errors

packages/mobile/src/web3/gas.ts
aaronmgdr: dope

packages/mobile/src/alert/reducer.ts
aaronmgdr: I feel like it might be better to just call Logger directly. as if we are not updating UI why pay the cost of updating redux state?

packages/mobile/src/components/CurrencyDisplay.v2.tsx
```diff
@@ -0,0 +1,227 @@
...
+    ? currency === CURRENCY_ENUM.GOLD
+      ? colors.celoGold
+      : colors.celoGreen
+    : StyleSheet.flatten(style)?.color
+
+  if (type === DisplayType.Big) {
+    // In this type the symbol is displayed as superscript
+    // the downside is we have to workaround React Native not supporting it
+    // and have to involve a View, which prevents this type to be embedded into a Text node
+    // see https://medium.com/@aaronmgdr/a-better-superscript-in-react-native-591b83db6caa
```
aaronmgdr: haha

packages/mobile/src/alert/reducer.ts
cmcewen: If I understand this correctly, you maybe want this to be `'banner' | 'inline'`? also making an enum would be nice

packages/mobile/src/alert/reducer.ts
```diff
@@ -16,10 +17,21 @@ const initialState = null
...
-        ...other,
+        errorType: 'show',
+        type: action.alertType,
+        message: action.message,
+        dismissAfter: action.dismissAfter,
+        buttonMessage: action.buttonMessage,
+        title: action.title,
+        underlyingError: action.underlyingError,
+      }
+    case Actions.LOG:
```
cmcewen: again not sure if i totally understand this correctly,  but curious if there is a different action needed if you just want to display the error differently

packages/mobile/src/components/ContactCircle.v2.tsx
```diff
@@ -0,0 +1,92 @@
+import colors from '@celo/react-components/styles/colors'
```
cmcewen: I think at this point it's probably fine just to redo the main one instead of create a dupe

packages/mobile/src/components/CurrencyDisplay.v2.tsx
```diff
@@ -0,0 +1,227 @@
+import colors from '@celo/react-components/styles/colors.v2'
```
cmcewen: What's the difference here? I believe we're already using this in the main feed, which has been redesigned, so if you need to modify it more I think it should happen in that component

packages/mobile/src/components/ErrorInline.tsx
```diff
@@ -0,0 +1,62 @@
...
+  errorContainer: {
+    height: 64,
+  },
+  errorMessage: {
+    ...fontStyles.small,
+    color: colors.warning,
+  },
+})
+
+export default connect<{}, DispatchProps, {}, RootState>(
```
cmcewen: for functional components i've been using the `useDispatch` hook, which i think is a little easier and keeps our component tree smaller since there's no HOC

packages/mobile/src/icons/MenuBurger.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react'
```
cmcewen, line 1: I created one of these too - we can consolidate after merge

packages/mobile/src/fees/CalculateFee.tsx
```diff
@@ -56,6 +56,12 @@ export type PropsWithoutChildren =
 
 type Props = InviteProps | SendProps | ExchangeProps | ReclaimEscrowProps
 
+// Max lengthed comment to fetch fee estimate before user finalizes comment
+const MAX_PLACEHOLDER_COMMENT: string = new Array(7)
```
cmcewen: not important at all, but fyi 😄 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat

packages/mobile/src/web3/utils.ts
cmcewen: nice catch

packages/mobile/src/alert/reducer.ts
tarikbellamine: That's a good naming suggestion. Will do.

@aaronmgdr that's a good point and highlights my poor naming decision. We actually do want to `show` and not `log`, we just want it to show inline rather than in banner. I think Conner's suggestion clears up that confusion.

packages/mobile/src/fees/CalculateFee.tsx
tarikbellamine: That's definitely good to know haha

packages/mobile/src/alert/reducer.ts
tarikbellamine: Good catch. It can be consolidated into one with the `errorType` parameter used to differentiate.

packages/mobile/src/components/ContactCircle.v2.tsx
tarikbellamine: Sounds good

packages/mobile/src/components/CurrencyDisplay.v2.tsx
tarikbellamine: Main difference is that we don't want the currency sign to be superscript. Wasn't aware that this had already been redone, I can just add a new parameter to disable superscript if we still want to display it like that elsewhere

packages/mobile/src/alert/reducer.ts
tarikbellamine: Settled on consolidating into one action but still having separate dispatching functions called `showError` and `showErrorInline`. Changing `showError` is in undertaking an and of itself because it's so widely used and already stacked with optional parameters. Felt cleaner to make a separate function rather than pile onto `showError`.

packages/mobile/src/components/ContactCircle.v2.tsx
tarikbellamine: Actually, took a 2nd look and I am a hesitant to replace because the original `ContactCircle` is in the higher-level `react-components` directory. The fact that we use both react directories interchangeably has been confusing for me and gives me pause with regards to making changes. Would be great if we settled on one place for mobile components and imported from web into a mobile components directory as needed.

Are you sure that `ContactCircle` and `Avatar` are exclusively used in the`mobile` directory? Looks like it's also used in `verifier` but I am not sure if that's actually mobile related.

packages/mobile/src/alert/AlertBanner.tsx
```diff
@@ -29,15 +29,19 @@ export class AlertBanner extends React.Component<Props> {
   render() {
     const { alert, hideAlert: hideAlertAction } = this.props
 
+    if (!alert || alert.errorType === 'log') {
+      return null
+    }
+
```
jeanregisser: I think by doing this it's breaking the animation because `SmartTopAlert` expects to be always mounted.

packages/mobile/src/alert/actions.ts
```diff
@@ -49,6 +57,13 @@ export const showError = (
   )
 }
 
+export const logError = (error: ErrorMessages, i18nOptions?: object): LogAlertAction => ({
```
jeanregisser: We should annotate with the right type for i18nOptions.

packages/mobile/src/components/CurrencyDisplay.v2.tsx
jeanregisser: If you don't want superscript, you should use `DisplayType.Default` and pass the text style you want using the `style` prop.

packages/mobile/src/components/ContactCircle.v2.tsx
jeanregisser: AFAIK things in `react-components` are only used by `mobile` or `verifier.
You should be fine changing them now.

packages/mobile/src/alert/actions.ts
```diff
@@ -7,6 +7,7 @@ import i18n from 'src/i18n'
 export enum Actions {
   SHOW = 'ALERT/SHOW',
   HIDE = 'ALERT/HIDE',
+  LOG = 'ALERT/LOG',
```
jeanregisser: The `alert` module was meant to manage our global modal alerts.
Though the use case now evolves to include inline messages, I'm not entirely convinced we should let it handle them too.

I know the difficulty is linking error/messages raised by sagas to the screen/components that initiated the original action.
I've thought a lot about this component/saga communication use case.
And we're not the only one asking about this: https://github.com/redux-saga/redux-saga/issues/907

I guess one option is passing a callback/promise to the action which the saga could call/fulfill when done to notify the component.

Happy to chat about this directly in our team channel.

packages/mobile/src/components/ErrorInline.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const mapDispatchToProps = {
+  hideAlert,
+}
+
+type Props = DispatchProps & OwnProps & WithTranslation
+
+// Pass null as dismissAfter value to show error indefinitely
+function ErrorMessageInline(props: Props) {
```
jeanregisser: nit: component name should match its filename to reduce confusion.

packages/mobile/src/fees/CalculateFee.tsx
jeanregisser: This should be based on `MAX_COMMENT_LENGTH`.

packages/mobile/src/icons/InfoIcon.v2.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react'
+import Svg, { Circle, Path } from 'svgs'
+
+export default function InfoIcon() {
```
jeanregisser: nit: We should use `React.memo` or `PureComponent` to prevent useless re-renders.

packages/mobile/src/icons/MenuBurgerCard.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { elevationShadowStyle } from '@celo/react-components/styles/styles'
+import * as React from 'react'
+import { StyleSheet, View } from 'react-native'
+import MenuBurger from 'src/icons/MenuBurger'
+
+interface Props {
+  length: number
+}
+
+export default function MenuBurgerCard({ length }: Props) {
```
jeanregisser: nit: ditto about `React.memo` | `PureComponent`

packages/mobile/src/navigator/types.tsx
```diff
@@ -118,6 +118,7 @@ export type StackParamList = {
   }
   [Screens.SendConfirmation]: {
     transactionData: TransactionDataInput
+    addressJustValidated?: true
```
jeanregisser: shouldn't this be `boolean`?

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.test.tsx
```diff
@@ -76,7 +75,6 @@ describe('IncomingPaymentRequestListItem', () => {
     const props = {
       id: 1,
       amount: new BigNumber(1),
-      comment: 'Something',
```
jeanregisser, line 79: Why removing this?

packages/mobile/src/send/SendConfirmation.test.tsx
```diff
@@ -34,6 +53,24 @@ describe('SendConfirmation', () => {
     mockedGetSendFee.mockClear()
   })
 
+  describe('when commenting', () => {
+    const store = createMockStore(storeData)
+    const getWrapper = () =>
+      render(
+        <Provider store={store}>
+          <SendConfirmation navigation={mockNavigation} route={mockRoute} />
```
jeanregisser: nit: I've added a `getMockStackScreenProps` in `test/utils` to simplify this.

packages/mobile/src/send/SendConfirmation.test.tsx
```diff
@@ -43,6 +80,8 @@ describe('SendConfirmation', () => {
       },
     })
 
+    mockedGetSendFee.mockImplementation(async () => TEST_FEE)
+
```
jeanregisser: Looks like a wrong merge happened, it's the same as line 74.

packages/mobile/src/send/SendConfirmation.test.tsx
```diff
@@ -73,6 +112,10 @@ describe('SendConfirmation', () => {
       },
     })
 
+    mockedGetSendFee.mockImplementation(async () => {
+      throw new Error('Calculate fee failed')
+    })
```
jeanregisser: Same here, looks like a wrong merge happened, it's the same as line 105.

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -47,6 +67,7 @@ interface StateProps {
   confirmationInput: ConfirmationInput
   addressValidationType: AddressValidationType
   validatedRecipientAddress?: string
+  addressJustValidated?: true
```
jeanregisser: Shouldn't this be `boolean`?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -105,33 +150,39 @@ export class SendConfirmation extends React.Component<Props, State> {
...
     buttonReset: false,
+    reason: '',
+    newReason: '',
   }
 
   async componentDidMount() {
+    const { addressJustValidated, t } = this.props
+    this.setState({ buttonReset: true }, () => {
+      this.setState({ buttonReset: false })
+    })
```
jeanregisser: What's going on here? 😅

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -221,65 +287,151 @@ export class SendConfirmation extends React.Component<Props, State> {
...
+            titleIcon={<FeeIcon />}
+            amount={
+              securityFeeAmount && (
+                <CurrencyDisplay amount={securityFeeAmount} formatType={FormatType.Fee} />
+              )
+            }
+            isLoading={asyncFee.loading}
+            hasError={!!asyncFee.error}
+          />
+          {/* <HorizontalLine /> */}
```
jeanregisser: nit: Remove?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -221,65 +287,151 @@ export class SendConfirmation extends React.Component<Props, State> {
...
+                      onPress={this.onEditAddressClick}
+                    >
+                      {t('edit')}
+                    </TextButton>
+                  </View>
+                )}
+              </View>
+            </View>
+            <CurrencyDisplay
+              type={DisplayType.Big}
```
jeanregisser: Use `Default` instead and adjust the font you want using `style`.

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -184,29 +183,43 @@ export class ValidateRecipientAccount extends React.Component<Props, State> {
...
-            {t('confirmAccountNumber.help', { displayName })}
-          </Text>
+          <View>
+            <Text style={styles.h2}>{t('confirmAccountNumber.title')}</Text>
+            <View>{this.renderInstructionsAndInputField()}</View>
+            <ErrorMessageInline error={error} />
+            <Button
+              style={styles.button}
+              onPress={this.onPressConfirm}
+              text={`${TEXT_SPACING}${t('confirmAccountNumber.submit')}${TEXT_SPACING}`}
```
jeanregisser: Can't we adjust the style instead?

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -184,29 +183,43 @@ export class ValidateRecipientAccount extends React.Component<Props, State> {
...
+          <Text style={styles.modalBody}>{t('helpModal.body1')}</Text>
+          <View style={styles.menuContainer}>
+            <View style={styles.menuCardContainer}>
+              <MenuBurgerCard length={30} />
             </View>
+            <Text style={styles.menuText}>Menu</Text>
+          </View>
+          <Text style={styles.modalBody}>{t('helpModal.body2')}</Text>
+          <View style={styles.addressContainer}>
+            {/* Placeholder for the address card */}
```
jeanregisser: Cool, `AccountNumber` has just been merged into master and you can use it.

jeanregisser (commented):
Awesome work, great seeing this land! 🎉 👍 
Left a couple of small comments mostly and a somewhat more important remark on how to handle inline error messages (which we'll probably handle later on).

packages/mobile/src/alert/actions.ts
aaronmgdr: object is actually the right type. its what is used by i18n itself when typing the t function. It cant be any more precise because you can litterally pass any key and value you want and use like `{{keyname}}` in the translation value string

packages/mobile/src/alert/actions.ts
aaronmgdr: Does `LOG` ever show anything to the user? if not i dont think it should go thru redux.

packages/mobile/src/components/ContactCircle.v2.tsx
aaronmgdr: verifier was an app for forwarding text messages. prior to attestation service. 

react-components is only used by mobile and verifier

packages/mobile/src/alert/reducer.ts
```diff
@@ -2,9 +2,14 @@ import { Actions, ActionTypes } from 'src/alert/actions'
 import { ErrorMessages } from 'src/app/ErrorMessages'
 import { RootState } from 'src/redux/reducers'
 
+export enum ErrorDisplayType {
+  'BANNER' = 0,
```
aaronmgdr:
```suggestion
export enum ErrorDisplayType {
BANNER,
INLINE
}
```

enum will default to enumerating 0 1 2 etc

packages/mobile/src/alert/AlertBanner.tsx
tarikbellamine: Great catch!

packages/mobile/src/alert/actions.ts
tarikbellamine: It does. I've renamed `log` to `showErrorInline` to be more clear about that.

packages/mobile/src/fees/CalculateFee.tsx
tarikbellamine: good call

packages/mobile/src/alert/reducer.ts
tarikbellamine: good to know! thanks

packages/mobile/src/send/SendConfirmation.tsx
tarikbellamine: Nice job making this so extensible. Makes it really easy to use!

packages/mobile/src/send/SendConfirmation.tsx
tarikbellamine: Could be. There's no use case for `false` but you aren't the first person to ask this so I will use `boolean` from now on to avoid confusion

packages/mobile/src/components/ErrorInline.tsx
tarikbellamine: ended up needing state here after all

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.test.tsx
tarikbellamine: comments are now written on the `SendConfirmation` screen so it doesn't make sense for it to be included in props before that point

packages/mobile/src/app/App.tsx
```diff
@@ -47,7 +47,9 @@ BigNumber.config({
 
 // Enables LayoutAnimation on Android
 if (Platform.OS === 'android') {
-  UIManager.setLayoutAnimationEnabledExperimental(true)
+  // tslint:disable-next-line:no-unused-expression
+  UIManager.setLayoutAnimationEnabledExperimental &&
```
aaronmgdr: could avoid the tslint disabled by combining the android check

`if (Platform.OS === 'android' && UIManager.setLayoutAnimationEnabledExperimental) {

packages/mobile/src/alert/actions.ts
jeanregisser: You can pass any string value indeed, but there are other options which are different. Using `object` will bypass the real types. See i18next types:
https://github.com/i18next/i18next/blob/f604b34e00d14f74faada327851a47f78603bc5a/index.d.ts#L645

packages/mobile/src/alert/actions.ts
```diff
@@ -49,6 +59,14 @@ export const showError = (
   )
 }
 
+export const showErrorInline = (error: ErrorMessages, i18nOptions?: object): ShowAlertAction => ({
+  type: Actions.SHOW,
+  alertType: AlertTypes.ERROR,
+  displayMethod: ErrorDisplayType.INLINE,
+  message: i18n.t(error, { ns: 'global', ...(i18nOptions || {}) }),
```
jeanregisser: nit: we should use `Namespaces.global` instead.

packages/mobile/src/alert/actions.ts
```diff
@@ -1,12 +1,15 @@
...
 import { DefaultEventNames } from 'src/analytics/constants'
 import { ErrorMessages } from 'src/app/ErrorMessages'
 import { ALERT_BANNER_DURATION } from 'src/config'
-import i18n from 'src/i18n'
+import i18n, { Namespaces } from 'src/i18n'
 
 export enum Actions {
   SHOW = 'ALERT/SHOW',
   HIDE = 'ALERT/HIDE',
+  LOG = 'ALERT/LOG',
```
aaronmgdr: nit: rename to inline like you did for other parts?

packages/mobile/src/alert/actions.ts
```diff
@@ -17,13 +20,21 @@ enum AlertTypes {
...
   alertType: AlertTypes
+  displayMethod: ErrorDisplayType
   message: string
   dismissAfter?: number | null
   buttonMessage?: string | null
   title?: string | null
   underlyingError?: ErrorMessages | null
 }
 
+interface LogAlertAction {
```
aaronmgdr: is this used?

packages/mobile/src/components/ErrorMessageInline.tsx
```diff
@@ -0,0 +1,74 @@
...
+function ErrorMessageInline(props: Props) {
+  const { error, displayMethod, dismissAfter, t } = props
+
+  // Keep the space empty when there isn't an inline error
+  if (!error || displayMethod !== ErrorDisplayType.INLINE) {
+    return <View style={dismissAfter !== null && styles.errorContainer} />
+  }
+
+  if (dismissAfter !== null) {
+    setTimeout(props.hideAlert, (dismissAfter || DISMISS_DEFAULT) * 1000)
```
aaronmgdr: This might need to be inside  a useFffect Hook otherwise each render creates a new setTimeout.

packages/mobile/src/alert/actions.ts
tarikbellamine: nope, forgot to delete. good catch

packages/mobile/src/components/ErrorMessageInline.tsx
aaronmgdr:
```
React.useEffect(() => {
let timer
 if (dismissAfter !== null) {
   timer = setTimeout(props.hideAlert, (dismissAfter || DISMISS_DEFAULT) * 1000)
}
// return a function to stop timmer
return () => timer && timer.disconnect()
}, [hideAlert, dismissAfter])

File: packages/mobile/src/transactions/TransferConfirmationCard.tsx [hunk 34]
```diff
@@ -119,9 +116,8 @@ function InviteReceivedContent({
         e164PhoneNumber={e164PhoneNumber}
         avatar={
           <ContactCircle
-            name={recipient?.displayName}
```
aaronmgdr, line 122: curious why this new syntax didnt work?

tarikbellamine: Needed it to return `null` not `undefined`. We prefer to use initials over the undefined user avatar so I wanted us to be explicit about when we didn't have `name` by requiring a string or null

packages/mobile/locales/en-US/global.json
```diff
@@ -160,5 +160,6 @@
   "addressValidationPartialOwnAddress": "Please enter the last 4 digits of your contact's account number, not your own. Press the link below for help",
   "helpFindAccount": "Help your contact find their account number",
   "close": "Close",
+  "dismiss": "Dismiss",
```
jmrossy, line 163: Nit, I guess we already have this here:
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/locales/en-US/walletFlow5.json#L72

But global.json is a better place for it. That old one is a dumping ground before we broke these out by major flow. If you don't mind, please remove the old one.

packages/mobile/src/alert/actions.ts
jmrossy: Still here

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
```diff
@@ -100,7 +95,8 @@ class PaymentRequestConfirmation extends React.Component<Props> {
       requesterE164Number: this.props.e164PhoneNumber,
       requesteeAddress,
       currency: currencyToShortMap[CURRENCY_ENUM.DOLLAR],
-      comment: reason,
+      // NOTE: Add this back in when redesigning this screen
+      comment: 'placeholder',
```
jmrossy, line 99: Should this be set back to reason?

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
```diff
@@ -141,7 +132,7 @@ class PaymentRequestConfirmation extends React.Component<Props> {
             recipient={recipient}
             address={requesteeAddress || ''}
             e164PhoneNumber={recipient.e164PhoneNumber}
-            comment={reason}
+            // comment={reason} NOTE: Add this back in when redesigning this screen
```
jmrossy, line 135: Ditto

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -71,9 +94,30 @@ const mapDispatchToProps = {
   completePaymentRequest,
 }
 
+interface DisplayNameProps {
+  recipient: Recipient
+  recipientAddress?: string | null
+  t: TFunction
+}
+
+function getDisplayName({ recipient, recipientAddress, t }: DisplayNameProps) {
```
jmrossy: Should this be in `recipient.ts`? Also, it looks very similar to the one in `Avatar.ts`, wondering if we can DRY those up

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -105,33 +150,36 @@ export class SendConfirmation extends React.Component<Props, State> {
   state = {
     modalVisible: false,
     buttonReset: false,
+    reason: '',
```
jmrossy: Like you said the other day, we should start calling these `comment`

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -146,6 +194,21 @@ export class SendConfirmation extends React.Component<Props, State> {
...
+  onKeyDown = (e: NativeSyntheticEvent<TextInputKeyPressEventData>) => {
+    if (e.nativeEvent.key === 'Enter') {
+      Keyboard.dismiss()
+    }
+  }
+
+  cleanInput = () => {
+    // Erase the "enter" keystroke from the comment
+    // if it was used to submit input
+    if (this.state.reason.slice(-1) === '\n') {
```
jmrossy: would `.trim()` do this?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -146,6 +194,21 @@ export class SendConfirmation extends React.Component<Props, State> {
     )
   }
 
+  onKeyDown = (e: NativeSyntheticEvent<TextInputKeyPressEventData>) => {
```
jmrossy: Would be nice if this kind of logic was baked into the textInput component. Can we add it there instead?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -221,66 +284,147 @@ export class SendConfirmation extends React.Component<Props, State> {
...
+                <Text style={styles.headerText}>Sending</Text>
+                <Text style={styles.displayName}>
+                  {getDisplayName({ recipient, recipientAddress, t })}
+                </Text>
+                {validatedRecipientAddress && (
+                  <View style={styles.editContainer}>
+                    <Text style={styles.address}>{`${validatedRecipientAddress.slice(
+                      0,
+                      6
+                    )}...${validatedRecipientAddress.slice(-4)}`}</Text>
```
jmrossy: Just wondering if the designs call for this in more than one place? Should we make a ShortenedAddress component?

packages/react-components/components/ContactCircle.tsx
```diff
@@ -25,64 +26,73 @@ const getNameInitial = (name: string) => name.charAt(0).toLocaleUpperCase()
...
+          style={[style.image, { height: iconSize, width: iconSize, borderRadius: iconSize / 2.0 }]}
           resizeMode={'cover'}
         />
       )
     }
 
-    // If children components (i.e. a default image) has been provided, use that
-    if (children) {
-      return children
+    // Mobile # is what default display name when contact isn't saved
```
jmrossy, line 48: On both iOS and Android?

jmrossy (commented):
Mostly looks great! Just a few small comments

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
tarikbellamine: Not yet, I am going to do the redesign for the payment screens in the next PR

packages/mobile/src/send/SendConfirmation.tsx
tarikbellamine: yep! will use instead

packages/react-components/components/ContactCircle.tsx
tarikbellamine: Yeah, we assign the default display name when importing contacts to standardize it

packages/mobile/src/components/ErrorMessageInline.tsx
```diff
@@ -42,15 +42,16 @@ type Props = DispatchProps & OwnProps & WithTranslation & StateProps
 function ErrorMessageInline(props: Props) {
   const { error, displayMethod, dismissAfter, t } = props
 
+  React.useEffect(() => {
+    const timer = window.setTimeout(props.hideAlert, (dismissAfter || DISMISS_DEFAULT) * 1000)
+    return () => window.clearTimeout(timer)
+  }, [error])
```
aaronmgdr, line 49: it looks like dismissAfter should be included here. 

?

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
tarikbellamine: Payment screens aren't included in this PR. That's next up :)

packages/mobile/src/send/SendConfirmation.tsx
tarikbellamine: Moved to `recipient.ts` and left a note for whoever redesigns `Avatar` that there is a function that they can import. Reluctant to put in the effort to make it extensible if I am not sure if/how it will be used once redesigned

packages/mobile/src/components/ErrorMessageInline.tsx
tarikbellamine: will leave comment for why i am using `error`

packages/mobile/src/send/SendConfirmation.tsx
tarikbellamine: Good point. Our custom `TextInput` is already pretty loaded so I've opted to create a new `CommentTextInput` component

packages/mobile/src/send/SendConfirmation.tsx
tarikbellamine: Not sure but I'll create one just in case

aaronmgdr (approved):
🥳

jmrossy (approved):
It LGTM, nice work!

PR #3906, opened 2020-05-27
Helm chart for eksportisto

nambrot:
### Description

Simple helm chart to deploy eksportisto runs


packages/celotool/src/lib/eksportisto.ts
```diff
@@ -0,0 +1,47 @@
...
+async function helmParameters(celoEnv: string) {
+  const suffix = fetchEnvOrFallback(envVar.EKSPORTISTO_SUFFIX, '1')
+  const params = [
+    `--namespace ${celoEnv}`,
+    `--set environment="${celoEnv}"`,
+    `--set imageRepository="${fetchEnv(envVar.EKSPORTISTO_DOCKER_IMAGE_REPOSITORY)}"`,
+    `--set imageTag="${fetchEnv(envVar.EKSPORTISTO_DOCKER_IMAGE_TAG)}"`,
+    `--set deploymentSuffix=${suffix}`,
+  ]
+  if (isVmBased()) {
```
jcortejoso, line 41: What's is the value of web3Provider if the network is not vmbased?

jcortejoso (approved):
Lgtm

packages/celotool/src/lib/eksportisto.ts
nambrot: You know that is a great question. I just copied it from the `transaction-metrics-exporter` equivalent

PR #3907, opened 2020-05-27
Make all possible oracle values easily configured, upgrade oracle infra

tkporter:
### Description

* Provides a way for us to easily specify a value for any configurable env var for the oracle client
* By default does not expose RPC ports to the external internet
* Uses the websocket RPC endpoint, which is needed for block based reporting
* Changes the `--useFullNodes` flag on oracle deployment commands to `--useForno`. By default, now full nodes are used

### Other changes

n/a

### Tested

Deployed successfully!

### Related issues

n/a

### Backwards compatibility

The default values for the oracle env vars are not backward compatible due to the introduction of a new aggregation method & report strategy


packages/helm-charts/oracle/values.yaml
```diff
@@ -29,6 +32,37 @@ oracle:
...
+  aggregation:
+    mid:
+      maxExchangeVolumeShare: 1
+      # askMaxPercentageDeviation
+      # bidMaxPercentageDeviation
+      # maxExchangeVolumeShare
+      # maxPercentageBidAskSpread
+      # minReportPriceChangeThreshold
+    trades:
+      period: 300000
```
codyborn: Nit: can you specify the units "periodInMs"

codyborn: Same for `apiRequestTimeout` and `removeExpiredFrequency`

codyborn (approved):
LGTM, let me know when this is checked in and I'll redeploy the oracleintegration full-nodes

PR #3909, opened 2020-05-27
[utils/cli] Generate addresses using custom mnemonics and derivationPaths

gastonponti:
# Description

Add to the command `account:new` from the `cli` two flags, one to set the mnemonic, and other (hidden for internal use) to set the derivationPath

[Requires to publish the utils package]

### Tested

Manually


packages/cli/src/commands/account/new.ts
```diff
@@ -38,13 +39,23 @@ export default class NewAccount extends LocalCommand {
       description:
         "Language for the mnemonic words. **WARNING**, some hardware wallets don't support other languages",
     }),
```
timmoreton, line 45: I would much prefer the mnemonic flag (and password) to pass a file that contains the mnemonic rather than have that exposed on the command line (hence in the ps list and in bash history)

packages/cli/src/commands/account/new.ts
```diff
@@ -38,13 +39,23 @@ export default class NewAccount extends LocalCommand {
...
+    mnemonic: flags.string({
+      description:
+        'Instead of generating a new mnemonic (seed phrases) the user can set the mnemonic to be used. It is required to set it as a string with all the words in order, separated by spaces (example: "word1 word2 word3 ... word24"). If the words are in other language than enaglish, the --language flag must be used. Should be a bip39 mnemonic',
+    }),
+    derivationPath: flags.string({
+      hidden: true,
+      description:
+        "Choose a different derivation Path (Celo's default is \"m/44'/52752'/0/0\"). This flags is hidden because is required only to help with specific problems (Example: A user that used in a transfer an Ethereum address and wants to recover its funds). NON technical users that don't understand the danger of changing it, SHOULDN'T use it",
```
timmoreton: Typo? `m/44'/52752'/0'/0/`

timmoreton: Given the idea is to make it simple for users to import from ethereum it might be nice to have flag like` —ethereumMnemonic` or special value for this flag like` —derivationPath=eth`

packages/cli/src/commands/account/new.ts
```diff
@@ -38,13 +39,23 @@ export default class NewAccount extends LocalCommand {
       description:
         "Language for the mnemonic words. **WARNING**, some hardware wallets don't support other languages",
     }),
+    mnemonic: flags.string({
+      description:
+        'Instead of generating a new mnemonic (seed phrases) the user can set the mnemonic to be used. It is required to set it as a string with all the words in order, separated by spaces (example: "word1 word2 word3 ... word24"). If the words are in other language than enaglish, the --language flag must be used. Should be a bip39 mnemonic',
+    }),
+    derivationPath: flags.string({
+      hidden: true,
```
timmoreton: I think it may as well be shown - what’s the value in hiding it from a user ? Also I think your warning is a little scary. Maybe “Recreating the same account requires knowledge of both the mnemonic and derivation path”

packages/cli/src/commands/account/new.ts
timmoreton: Also Path->path

timmoreton (commented):
Few comments- but I think this will be a really useful feature to have to add to an FAQ to help users!

packages/cli/src/commands/account/new.ts
gastonponti: Haha, yes, my idea was to be scary to avoid non technical users, but I can change it

packages/cli/src/commands/account/new.ts
gastonponti: Yes, I totally agree
In two different plain files? (mnemonic and password)
Or have a file with a configuration json?

marekolszewski: I suggest the following: one file with each word in the mnemonic separated with a space or newline. If there's a 25th word, that word would is considered the password.

packages/cli/src/commands/account/new.ts
```diff
@@ -38,13 +39,23 @@ export default class NewAccount extends LocalCommand {
       description:
         "Language for the mnemonic words. **WARNING**, some hardware wallets don't support other languages",
     }),
+    mnemonic: flags.string({
+      description:
+        'Instead of generating a new mnemonic (seed phrases) the user can set the mnemonic to be used. It is required to set it as a string with all the words in order, separated by spaces (example: "word1 word2 word3 ... word24"). If the words are in other language than enaglish, the --language flag must be used. Should be a bip39 mnemonic',
```
jmrossy:
```suggestion
        'Instead of generating a new mnemonic (seed phrase), use the user-supplied mnemonic instead. The mnemonic should contain all words separate by a space (example: "word1 word2 word3 ... word24"). If the words are a language other than English, the --language flag must be used. Only bip39 mnemonics are supported',
```

jmrossy (commented):
LGTM, just suggested a rephrasing of the description

packages/cli/src/commands/account/new.ts
marekolszewski: I like --derivationPath=eth

packages/cli/src/commands/account/new.ts
gastonponti: The only problems that I find on using one file, is that we are adding a possible error in how the user used the command
if for some reason added the password behind instead of the front, of if it's using a mnemonic of other strength, for example 21, and with a password of 3 words separated by a whitespace (unlikely, of course, but I prefer to go to the safe options of two files)

packages/cli/src/commands/account/new.ts
gastonponti: added

jmrossy (approved):

packages/cli/src/commands/account/new.ts
```diff
@@ -8,17 +8,20 @@ import {
...
+
 export default class NewAccount extends LocalCommand {
   static description =
     "Creates a new account locally using the Celo Derivation Path (m/44'/52752'/0/0/indexAddress) and print out the key information. Save this information for local transaction signing or import into a Celo node. Ledger: this command has been tested swapping mnemonics with the Ledger successfully (only supports english)"
 
   static flags = {
     ...LocalCommand.flags,
-    password: flags.string({
-      description: 'Choose a password to generate the keys',
+    passwordPath: flags.string({
```
timmoreton: I think this should really be `s/password/passphrase` -- this is what it's called in [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) and on the Ledger UI

packages/cli/src/commands/account/new.ts
```diff
@@ -8,17 +8,20 @@ import {
...
+      description: 'Path to a file that contains the password to generate the keys',
```
timmoreton:
```suggestion
      description: 'Path to a file that contains the BIP39 passphrase to combine with the mnemonic specified using the mnemonicPath flag and the index specified using the indexAddress flag. Every passphrase generates a different private key and wallet address.',
```

packages/cli/src/commands/account/new.ts
```diff
@@ -39,23 +42,22 @@ export default class NewAccount extends LocalCommand {
       description:
         "Language for the mnemonic words. **WARNING**, some hardware wallets don't support other languages",
     }),
-    mnemonic: flags.string({
+    mnemonicPath: flags.string({
       description:
-        'Instead of generating a new mnemonic (seed phrases) the user can set the mnemonic to be used. It is required to set it as a string with all the words in order, separated by spaces (example: "word1 word2 word3 ... word24"). If the words are in other language than enaglish, the --language flag must be used. Should be a bip39 mnemonic',
+        'Instead of generating a new mnemonic (seed phrase), use the user-supplied mnemonic instead. Path to a file that contains all the mnemonic words separate by a space (example: "word1 word2 word3 ... word24"). If the words are a language other than English, the --language flag must be used. Only bip39 mnemonics are supported',
```
timmoreton:
```suggestion
        'Instead of generating a new mnemonic (seed phrase), use the user-supplied mnemonic instead. Path to a file that contains all the mnemonic words separated by a space (example: "word1 word2 word3 ... word24"). If the words are a language other than English, the --language flag must be used. Only BIP39 mnemonics are supported',
```

packages/cli/src/commands/account/new.ts
```diff
@@ -39,23 +42,22 @@ export default class NewAccount extends LocalCommand {
...
     }),
     derivationPath: flags.string({
-      hidden: true,
       description:
-        "Choose a different derivation Path (Celo's default is \"m/44'/52752'/0/0\"). This flags is hidden because is required only to help with specific problems (Example: A user that used in a transfer an Ethereum address and wants to recover its funds). NON technical users that don't understand the danger of changing it, SHOULDN'T use it",
+        "Choose a different derivation Path (Celo's default is \"m/44'/52752'/0'/0\"). Allows to use \"eth\" as an alias of the Ethereum derivation path (\"m/44'/60'/0'/0/\"). Recreating the same account requires knowledge of both the mnemonic and derivation path",
```
timmoreton:
```suggestion
        "Choose a different derivation Path (Celo's default is \"m/44'/52752'/0'/0\"). Use \"eth\" as an alias of the Ethereum derivation path (\"m/44'/60'/0'/0/\"). Recreating the same account requires knowledge of the mnemonic, passphrase (if any), and the derivation path",
```

timmoreton (approved):
LGTM except minor changes below!

packages/cli/src/commands/account/new.ts
gakonst: You could also just have the mnemonic as-is in the file, and pop a silent prompt where the user inputs their password to stdin (without showing the password in the console of course). Storing passphrases on disk is poor practice and should be avoided IMO.

Imagine this from the POV of an attacker: If they have access to the file with the mnemonic, they also have access to the password. Why did you end up adding a password then? 

The password should be in a separate location (i.e. not on disk) from the mnemonic which would keep funds secure even if the mnemonic was compromised.

[This](https://github.com/jdxcode/password-prompt/blob/master/index.js) package is available on NPM and is relatively small, so you could either import it or re-write something similar

PR #3915, opened 2020-05-27
[Wallet] Allow QR scan to circumvent Secure Send check

tarikbellamine:
### Description
Given QR scans are in person and have a lower likelihood of malicious activity, we want to minimize surface area for failures of QR scans so that it's as error-proof as possible. Quick fix to allow QR scans to circumvent the attestation contract lookup and thus bypass the SecureSend logic.

A more subtle change is that we are throwing an error if the recipient isn't doing a QR can and also doesn't have a phone number. In this case, we probably don't want the transaction to continue because we can't confirm the address mapping is up-to-date.

### Tested
Yep

### Related issues
Issue was raised in #3895 

### Backwards compatibility
Yes


packages/mobile/src/send/SendAmount.tsx
```diff
@@ -157,9 +157,15 @@ export class SendAmount extends React.Component<Props, State> {
...
     const { recipient } = this.props
 
-    if (recipient.e164PhoneNumber) {
-      this.props.fetchAddressesAndValidate(recipient.e164PhoneNumber)
+    if (recipient.kind === RecipientKind.QrCode) {
+      return
+    }
+
+    if (!recipient.e164PhoneNumber) {
+      throw Error('Recipient phone number is required if not sending via QR Code')
```
jmrossy: This will throw if the recipient type is `RecipientKind.Address` right? That's what we use if the use types/pastes an address in to the send field. You can fix this by adding an `||` case for it above :)

jmrossy (approved):

Issue #3918, opened 2020-05-27
Multisig wrapper should allow for sending a second tx with the same parameters


asaj:
### Expected Behavior

Multisig should submit a new tx if all matching txs have already been confirmed by the owner
 
### Current Behavior

Multisig wrapper will try to re-confirm the tx

aaronmgdr:
closing since aslawons pr seems to address this

PR #3920, opened 2020-05-28
[contractKit] Add condition for MultiSig wrapper to resubmit a transaction if it was already executed

aslawson:
### Description

- Currently, if multisg owner wants to submit an action again, it will not resubmit it.  Instead it will just re-confirm that users action for the old event.
- This change adds a conditional check that will create a new transaction if the previous was already executed. 

### Other changes

NA

### Tested

Ran tests

### Related issues

https://github.com/celo-org/celo-monorepo/issues/3918

### Backwards compatibility

Yes


packages/contractkit/src/wrappers/MultiSig.ts
```diff
@@ -42,7 +42,8 @@ export class MultiSigWrapper extends BaseWrapper<MultiSig> {
       if (
         transaction.data === data &&
         transaction.destination === destination &&
-        transaction.value === '0'
+        transaction.value === '0' &&
```
asaj: Can you remind me why we check that `transaction.value == '0'` as opposed to checking that it matches txObject?

aslawson: because we currently only submit transactions with 0 value (line 55).  this method could be generalized even more by having it take in a `value` parameter.  

although this feature wasn't needed yet. 

and I'm not sure if it's even needed at all since the `data` field of the transaction its approving/executing will have the any `value` it needs for that transaction.  my understanding is this `value` is if there will be a transfer from multisig to the other contract while transacting

asaj: Yeah `value` is for cGLD transfers, so there would e.g. be no way right now to lock cGLD via the multisig wrapper.

Can you fix this as a driveby as well? Feels like important functionality

aslawson: done

packages/contractkit/src/wrappers/MultiSig.ts
```diff
@@ -33,7 +33,11 @@ export class MultiSigWrapper extends BaseWrapper<MultiSig> {
    * Otherwise, submits the `txObject` to the multisig and add confirmation.
    * @param index The index of the pending withdrawal to withdraw.
    */
-  async submitOrConfirmTransaction(destination: string, txObject: TransactionObject<any>) {
+  async submitOrConfirmTransaction(
+    destination: string,
+    txObject: TransactionObject<any>,
+    value: string = '0'
```
asaj, line 39: Does `TransactionObject` not have a `value` field?

asaj (approved):

packages/contractkit/src/wrappers/MultiSig.ts
aslawson: it has an `arguments` array of type "any", but no direct `value` field

PR #3926, opened 2020-05-29
Refactor/dependency graph with package names

AlexBHarley:
### Description

This PR https://github.com/celo-org/celo-monorepo/pull/3922 splits up the phone-number-privacy package into two folders and breaks the dependency graph. 

Here we make the `dependency-graph-utils.js` more robust so that it can handle packages at any depth in the tree.

### Other changes

N/A

### Tested

Locally and in CI

### Related issues

- https://github.com/celo-org/celo-monorepo/pull/3922


jmrossy (approved):
Ship it!

PR #3931, opened 2020-05-29
[Wallet] Add analytics for Gold screens + Send Part 1 redesigned screens

### LINKED ISSUE #3657: Wallet Analytics: Celo Gold
Parent Epic: #3641 
### Expected Behavior

https://docs.google.com/document/d/1OGNJBijiWp7bWc2I0p8yyJ72EMCj9pNRT1VWxJJu5JY/edit#heading=h.aq9387z75ieq

### Current Behavior

What is the current behavior?

i1skn:
### Description

Add analytics for Gold screens + Send Part 1 redesigned screens

### Other changes

Fix when header background was blocking ripple effect when button pressed on Android

### Tested

Android emulator

### Related issues

- Fixes #3657

### Backwards compatibility

Yes


packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -66,13 +70,32 @@ export class ExchangeReview extends React.Component<Props, State> {
   }
 
   onPressConfirm = () => {
-    const makerToken = this.state.makerToken
+    const { makerToken, inputToken, inputAmount } = this.state
     const makerAmount = this.getMakerAmount()
+    // BEGIN: Analytics
+    const isDollarToGold = inputToken === CURRENCY_ENUM.DOLLAR
```
jmrossy, line 76: This seems like a lot of data wrangling just for analytics. Does the saga do any of this already? If so, how about we trigger analytics from there instead?

packages/mobile/src/exchange/ExchangeReview.tsx
```diff
@@ -66,13 +70,32 @@ export class ExchangeReview extends React.Component<Props, State> {
...
+    const goldAmount = isDollarToGold ? makerAmount : inputAmount
+    const dollarsAmount = isDollarToGold ? inputAmount : makerAmount
+    const localCurrencyAmount = convertDollarsToLocalAmount(
+      dollarsAmount,
+      this.props.localCurrencyExchangeRate
+    )
+    CeloAnalytics.track(
+      isDollarToGold ? CustomEventNames.gold_buy_confirm : CustomEventNames.gold_sell_confirm,
+      {
+        localCurrencyAmount,
```
jmrossy, line 91: Are these on the properties whitelist? I think they don't get included if they're not. I didn't set up analytics so I'm not 100% sure how it works but I think props need to be whitelisted, which confused many people tbh https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/analytics/constants.ts#L207

jmrossy (commented):
Mostly looks good! Just a few small comments

packages/mobile/src/exchange/ExchangeReview.tsx
i1skn: I've updated `PROPERTY_PATH_WHITELIST` - looks a bit odd to me, so I guess we should work really soon on this `TODO` right above it :)

packages/mobile/src/exchange/ExchangeReview.tsx
i1skn: Well, it does not calculate quite, so I've decided to have it there. Essentially it does a couple `get` from state and a couple of arithmetic calculations, so should not be too heavy.

jmrossy (approved):

PR #3938, opened 2020-06-01
Add readiness probe to full nodes, use exec for long running processes in containers ran from sh/bash

tkporter:
### Description

This PR adds a readiness probe to full nodes. A node is deemed ready if the latest block has a timestamp that is less than 30 seconds ago (this can be overridden). If a node fails the readiness probe, traffic from a service won't be sent to the node and any upgrades of the statefulset will halt until the readiness probe is successful. This is especially useful for the oracle nodes to ensure the full nodes that RPCs are sent to are healthy.

This requires the http RPC endpoint to be listening, but ideally this should be moved to IPC in the future

Additionally, when running a container in k8s with a command like `/bin/sh -c "....; geth ..."` instead of just running geth directly, any signals that are sent to the pod/container will not be forwarded to geth (just using geth here as an example for a long-running process). When a pod is trying to be deleted by kubernetes, it'll first send SIGTERM to all the containers in the pod, wait a grace period, and then send SIGKILL. As a result I was seeing some issues where a pod was being deleted and geth wasn't receiving the SIGTERM so it was unable to gracefully shutdown, and chain data was being corrupted.
The fix is to run geth with exec (eg run `exec geth ...`) to have geth take over the current process, so this fixes that across the repo

### Other changes

n/a

### Tested

Deployed it

### Related issues

n/a

### Backwards compatibility

Yes


packages/helm-charts/load-test/templates/load-test.statefulset.yaml
```diff
@@ -136,7 +136,7 @@ spec:
             # Send the txs to the next load test client
             RECIPIENT_INDEX=$(( ($RID + 1) % {{ .Values.replicas }} ))
 
-            celotooljs.sh geth simulate-client \
+            exec celotooljs.sh geth simulate-client \
```
jcortejoso, line 139: Not sure if this could give some problems with `celotooljs.sh` location. If the it could use deterministic path (`/celo-monorepo/packages/celotool/bin/celotooljs.sh`)

jcortejoso (approved):
lgtm

packages/helm-charts/load-test/templates/load-test.statefulset.yaml
tkporter: I just tested it, it works fine :)

PR #3940, opened 2020-06-01
Fixing Tech debt in ReleaseGold and Election

### LINKED ISSUE #3500: Tech debt in ReleaseGold and Election
### Expected Behavior

There should be less tech  debt

### Current Behavior

In `Election:forceDecrementVotes` the last line is `return value.sub(info.remainingValue);` however `info.remainingValue` is required to be zero.

The computation of `getRemainingLockedBalance` in `ReleaseGold`. There is an `if` that the contract is an account before computing the `lockedgold` pending but not for the locked  part. This should all be in the same `if` block

37ng:
### Description

Fixing Tech debt in ReleaseGold and Election

### Tested

Passed unit tests

### Backwards compatibility

It's backwards compatible, no dependencies modified.


packages/protocol/contracts/governance/ReleaseGold.sol
```diff
@@ -426,15 +426,17 @@ contract ReleaseGold is UsingRegistry, ReentrancyGuard, IReleaseGold, Initializa
...
    * @notice Calculates remaining locked gold balance in the release schedule instance.
    *         The returned amount also includes pending withdrawals to maintain consistent releases.
+   *         Return 0 if address of caller is not an account.
    * @return The remaining locked gold of the release schedule instance.
    * @dev The returned amount may vary over time due to locked gold rewards.
    */
   function getRemainingLockedBalance() public view returns (uint256) {
     uint256 pendingWithdrawalSum = 0;
     if (getAccounts().isAccount(address(this))) {
       pendingWithdrawalSum = getLockedGold().getTotalPendingWithdrawals(address(this));
```
nategraf:
```suggestion
    if (getAccounts().isAccount(address(this))) {
      uint256 pendingWithdrawalSum = getLockedGold().getTotalPendingWithdrawals(address(this));
```
I'd suggest defining the variable only only in the if statement now.

nategraf (approved):

Issue #3899, opened 2020-05-26
Package signer as needed for partner infrastructure


jmrossy:
We’ll start requirements gathering asap from partners and then package the signer code in whatever way(s) best suits partner infrastructure. This includes the following:
- [x] Replace dependency on firebase config with something platform independent

Replace dependency on Azure Key Vault with something configurable. Possibly a set of supported key providers:
- [ ] GCP keystore
- [ ] AWS key management
- [x] Package the code for easy consumption as a docker image

PR #3941, opened 2020-06-01
[PGPNP] Remove matchmaking from signer and de-dupe test utils

jmrossy:
### Description

- Remove matchmaking from signer, only combiner needs it for now
- De-dupe some test utils and values. More can be de-duped but we'll do this gradually.
- Remove unused Model class

### Related issues

- #3899

### Backwards compatibility

yes


packages/phone-number-privacy/combiner/src/database/models/account.ts
```diff
@@ -1,20 +1,17 @@
...
-
 export const ACCOUNTS_TABLE = 'accounts'
 export enum ACCOUNTS_COLUMNS {
   address = 'address',
   createdAt = 'created_at',
   numLookups = 'num_lookups',
   didMatchmaking = 'did_matchmaking',
 }
-export class Account extends Model {
+export class Account {
```
aslawson, line 8: can this model be shared?

packages/phone-number-privacy/combiner/test/end-to-end/get-blinded-salt.test.ts
```diff
@@ -4,8 +4,8 @@ import { normalizeAddressWith0x, privateKeyToAddress } from '@celo/utils/lib/add
 import { serializeSignature, signMessage } from '@celo/utils/lib/signatureUtils'
 import 'isomorphic-fetch'
 import Web3 from 'web3'
+import { getBlindedPhoneNumber } from '../../../common/test/utils'
```
aslawson, line 7: if we pull these outside and rely on relative paths, how do we plan on packaging these separate service? will it be one package that contains both combiner+signer with the options to run either

aslawson: perhaps this will be answered when I get to your next PR

packages/phone-number-privacy/signer/src/index.ts
```diff
@@ -1,16 +1,9 @@
...
 
 // EG. curl -v "http://localhost:5000/celo-phone-number-privacy/us-central1/getBlindedSalt" -H "Authorization: 0xdaf63ea42a092e69b2001db3826bc81dc859bffa4d51ce8943fddc8ccfcf6b2b1f55d64e4612e7c028791528796f5a62c1d2865b184b664589696a08c83fc62a00" -d '{"hashedPhoneNumber":"0x5f6e88c3f724b3a09d3194c0514426494955eff7127c29654e48a361a19b4b96","blindedQueryPhoneNumber":"n/I9srniwEHm5o6t3y0tTUB5fn7xjxRrLP1F/i8ORCdqV++WWiaAzUo3GA2UNHiB","account":"0x588e4b68193001e4d10928660aB4165b813717C0"}' -H 'Content-Type: application/json'
 export const getBlindedSalt = functions.https.onRequest(async (request, response) => {
   logger.info('Begin getBlindedSalt request')
   return handleGetBlindedMessageForSalt(request, response)
 })
-
-// EG. curl -v "http://localhost:5000/celo-phone-number-privacy/us-central1/getContactMatches" -H "Authorization: <SIGNED_BODY>" -d '{"userPhoneNumber": "+99999999999", "contactPhoneNumbers": ["+5555555555", "+3333333333"], "account": "0x117ea45d497ab022b85494ba3ab6f52969bf6812"}' -H 'Content-Type: application/json'
-export const getContactMatches = functions.https.onRequest(async (request, response) => {
```
aslawson, line 13: one thing to note is that by keeping the contact matching in the combiner, the combiner still needs to maintain a db `accounts` table.  it will only use/update the `didMatchmaking` field for a record, but will have inaccurate `numlookups` info.  

this currently won't cause any issues in the performance, but it could cause confusion down the line, since it would be out of sync (not even maintaining) this field.  

In the ideal case, I wonder:
1. should the combiner have its own db table for matchmaking (without this access field)?
or
2. should the getBlindedSalt services return the number of queries they calculated for the request and then the combiner updates its table to include this field?


Also for future use of the matchmaking, the only way to give free queries will be if the matchmaking is done by the signers and then the results combined in the combiner.  By removing this now from the signer, will we have difficulty having partners update to perform this request as part of the service later?

aslawson (commented):
looks good -- few comments/concerns

packages/phone-number-privacy/combiner/test/end-to-end/get-blinded-salt.test.ts
jmrossy: We would be packaging the compiled result, not the source + tests. So essentially only the `dist` folder.

packages/phone-number-privacy/combiner/src/database/models/account.ts
jmrossy: I'm not sure what you mean, sorry

packages/phone-number-privacy/signer/src/index.ts
jmrossy: Great points here.

> didMatchmaking field for a record, but will have inaccurate numlookups info

I'm torn on this. We could just have the combiner keep its own record of lookups. This could be nice to prevent unnecessary requests to signers when a quota's already been exceeded. Or we could modify the Combiner's schema for remove those surplus fields. Let's chat about it.

> Also for future use of the matchmaking

We'd need the partners to upgrade their service anyway so I don't think removing it makes supporting it later meaningfully harder.

packages/phone-number-privacy/combiner/src/database/models/account.ts
aslawson: its duplicated across both combiner/signer

jmrossy: Ah k. It maybe could but I'm holding off on sharing code until we have a better idea what the signer will look like. Seems like there's a decent chance they'll have slightly diff models

aslawson: ya -- this seems it has become irrelevant with the latest decisions

packages/phone-number-privacy/signer/src/index.ts
aslawson: for documentation purposes, the decision made moving forward will be to update the table for combiner to not include query counts.  the combiner may be a distributed service in the future and keeping these in sync will become very difficult.  because of this and for "free matchmaking queries" the logic and table for matchmaking will likely also be moved to the signers, so combiner would not maintain state.

aslawson (approved):

Issue #3898, opened 2020-05-26
Implement proxy and combine logic in the combiner


jmrossy:
The combiner should be kept simple, as it may eventually be just a lib we include in the wallet/sdk.
For now it can have the same interface as the signers, to whom it proxies calls. It then combines the results.

PR #3943, opened 2020-06-02
[PGPNP] Added Combiner Logic for Threshold Signatures

aslawson:
### Description

- Added Combiner Logic for Threshold Signatures -- defaulted to 1/1 threshold
- The combiner makes requests to all service it has url's for.  It then verifies the partial signatures and attempts to combine and return the result.

### Other changes

- added script for generating threshold keys
- combiner also checks authentication header

### Tested

- locally ran a signers and combiner, pointed combiner to local signer for 3-out-of-4 signer test
`curl -v "http://localhost:5000/celo-phone-number-privacy/us-central1/getDistributedBlindedSalt" -H "Authorization: 0xfc2ee61c4d18b93374fdd525c9de09d01398f7d153d17340b9ae156f94a1eb3237207d9aacb42e7f2f4ee0cf2621ab6d5a0837211665a99e16e3367f5209a56b1b" -d '{"blindedQueryPhoneNumber":"+Dzuylsdcv1ZxbRcQwhQ29O0UJynTNYufjWc4jpw2Zr9FLu5gSU8bvvDJ3r/Nj+B","account":"0xdd18d08f1c2619ede729c26cc46da19af0a2aa7f", "hashedPhoneNumber":"0x8fb77f2aff2ef0343706535dc702fc99f61a5d1b8e46d7c144c80fd156826a77"}' -H 'Content-Type: application/json'`
- happy case (all work) - {"success":true,"combinedSignature":"iWPBrMxvekp/LxD9m5Sxr7PRylWeMtijzVrc8DRh3YEAxpMVTv2IJbAJZwsj0oUB"}
- one down (still works) - {"success":true,"combinedSignature":"iWPBrMxvekp/LxD9m5Sxr7PRylWeMtijzVrc8DRh3YEAxpMVTv2IJbAJZwsj0oUB"}
- two down (below threshold, fails) -- error logs "CELO_PNP_ERR_12 Not enough partial signatures 2/3"
- added unit tests to test 3/4 threshold

### Related issues

https://github.com/celo-org/celo-monorepo/issues/3898

### Backwards compatibility

NOT BACKWARDS COMPATIBLE -- It changes the signer to use `partialSignBlindedMessage` rather than `signBlindedMessage` which will cause mobile verification to fail.  We will to set the new combiner instance up and all the services separately and then transition mobile client to use new endpoint.

will wait for  https://github.com/celo-org/celo-monorepo/pull/3941 to be merged before merging


packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
```diff
@@ -1,57 +1,44 @@
...
-import config, { DEV_MODE, DEV_PRIVATE_KEY } from '../config'
+import config from '../config'
 
+export interface ServicePartialSignature {
+  url: string
+  signature: string
+}
+
+function flattenSigsArray(sigs: Uint8Array[]) {
+  return Uint8Array.from(sigs.reduce((a, b) => Array.from(a).concat(Array.from(b)), [] as any))
```
codyborn: Is the `Array.from(a)` necessary given that it's initialized to be an array? I'm thinking something like:
`sigs.reduce((a, b) => a.concat(Array.from(b)), [] as any)`

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
```diff
@@ -1,57 +1,44 @@
...
-
-      if (!signedMsg) {
-        throw new Error('Empty threshold sign result')
+  public static async combinePartialBlindedSignatures(
+    serviceResponses: ServicePartialSignature[],
+    blindedMessage: string
+  ): Promise<string> {
+    const polynomial = config.thresholdSignature.polynomial
+    const sigs: Uint8Array[] = []
+    for (const service of serviceResponses) {
```
codyborn: nitpick: maybe name the variable `serviceResponse` to make it more clear

packages/phone-number-privacy/combiner/package.json
```diff
@@ -27,6 +28,7 @@
     "firebase-admin": "^8.10.0",
     "firebase-functions": "^3.6.0",
     "knex": "^0.20.13",
+    "node-fetch": "^2.2.0",
```
jmrossy: I see we use 2.2.0 and 2.3.0 in the monorepo right now. Rather than adding this old version, it would be great to upgrade node-fetch across the monorepo to the latest (2.6.0). Totally optional though.

packages/phone-number-privacy/combiner/package.json
```diff
@@ -18,7 +18,8 @@
     "lint": "tslint --project .",
     "test": "jest --runInBand --testPathIgnorePatterns test/end-to-end",
     "db:migrate": "NODE_ENV=development ts-node scripts/runMigrations.ts",
-    "keygen": "ts-node scripts/createBlsKeyPair.ts"
+    "keygen": "ts-node scripts/createBlsKeyPair.ts",
+    "thresholdKeygen": "ts-node scripts/createThresholdBlsKeys.ts"
```
jmrossy: Nit: how about `keygen:threshold` and we change the one above to `keygen:single`?

packages/phone-number-privacy/combiner/scripts/createBlsKeyPair.ts
```diff
@@ -5,7 +5,7 @@ import crypto from 'crypto'
 console.log('Creating BLS keypair...')
 
 const seed = crypto.randomBytes(32)
-const keypair = threshold.keygen(seed)
+const keypair = threshold.threshold_bls(seed)
```
jmrossy: Why change this?

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
jmrossy: or `response`

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
```diff
@@ -1,57 +1,44 @@
...
+      const sigBuffer = Buffer.from(service.signature, 'base64')
+      try {
+        await threshold_bls.partialVerifyBlindSignature(
+          Buffer.from(polynomial, 'base64'),
+          Buffer.from(blindedMessage, 'base64'),
+          sigBuffer
+        )
+        sigs.push(sigBuffer)
+      } catch (e) {
+        logger.warn(`could not verify signature for service url ${service.url}`)
```
jmrossy: Should we `error` on this, it should never happen. Maybe a custom `ErrorMessage` for this case would be good

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -0,0 +1,68 @@
...
+      },
+      body: JSON.stringify(request.body),
+    })
+
+    // if res no ok, then it will not be added to the list of signatures
+    if (res.ok) {
+      const signResponse = (await res.json()) as SignMessageResponse
+      signatures.push({ url: service.url, signature: signResponse.signature })
+    }
+  }
```
jmrossy: Consider breaking this out into another function

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -0,0 +1,68 @@
...
+  try {
```
jmrossy, line 25: Similar to what we had before, a try/catch around the entire handler would be good to prevent leaking of information in cases of unexpected errors

packages/phone-number-privacy/signer/.firebaserc
```diff
@@ -0,0 +1,5 @@
+{
```
jmrossy, line 1: I left these out intentionally cause I suspect the signers won't be cloud functions (see https://github.com/celo-org/celo-monorepo/pull/3944). It'll depend on what we decide in today's infra meeting.

It's okay if you want to add them in for now, just an FYI

jmrossy (commented):
Some small comments but overall looks good! Great progress!

packages/phone-number-privacy/signer/test/salt-generation/bls-signature.test.ts
```diff
@@ -33,10 +33,21 @@ describe(`BLS service computes signature`, () => {
     const blindedMsg = Buffer.from(blindedMsgResult.message).toString('base64')
 
     const actual = await BLSCryptographyClient.computeBlindedSignature(blindedMsg)
-    expect(actual).toEqual('ZeYBwDBxkWe1ZNDqiViz2MNGIT6PIW2c3pemkMcmM5gM1vaaf5RieVp+2SxR83YA')
+    expect(actual).toEqual(
+      'MAAAAAAAAADJpFrx/eDNs1Qm986trWFZpMcJNRM5W/yKoI+cxk0gBul1PNAVzw1uFpEWJx5iK4EAAAAA'
+    )
+
+    expect(
```
codyborn, line 40: Should this be `expect().not.toThrow()`?

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
```diff
@@ -1,57 +1,44 @@
-import { AzureKeyVaultClient } from '@celo/contractkit/lib/utils/azure-key-vault-client'
-import threshold from 'blind-threshold-bls'
-import { ErrorMessages } from '../common/error-utils'
+import threshold_bls from 'blind-threshold-bls'
```
codyborn, line 1: Nit: Other imports are named `threshold`, if we can name this anything we can name it `threshold` for consistency.

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
codyborn: Should we log the exception?

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
```diff
@@ -1,57 +1,44 @@
...
-  private static async getPrivateKey(): Promise<string> {
-    if (DEV_MODE) {
-      return DEV_PRIVATE_KEY
     }
-
-    if (BLSCryptographyClient.privateKey) {
-      return BLSCryptographyClient.privateKey
+    const threshold = config.thresholdSignature.threshold
+    if (sigs.length < threshold) {
+      logger.error(`not enough not enough partial signatures'${sigs.length}'/'${threshold}'`)
```
codyborn: nit: "not enough" dupes

codyborn: Also, I don't think the single quote chars are necessary since you're using the `

packages/phone-number-privacy/combiner/src/index.ts
```diff
@@ -1,12 +1,13 @@
 import * as functions from 'firebase-functions'
 import logger from './common/logger'
 import { handleGetContactMatches } from './match-making/get-contact-matches'
-import { handleGetBlindedMessageForSalt } from './salt-generation/get-salt'
+import { handleGetDistributedBlindedMessageForSalt } from './salt-generation/get-threshold-salt'
 
-// EG. curl -v "http://localhost:5000/celo-phone-number-privacy/us-central1/getBlindedSalt" -H "Authorization: 0xdaf63ea42a092e69b2001db3826bc81dc859bffa4d51ce8943fddc8ccfcf6b2b1f55d64e4612e7c028791528796f5a62c1d2865b184b664589696a08c83fc62a00" -d '{"hashedPhoneNumber":"0x5f6e88c3f724b3a09d3194c0514426494955eff7127c29654e48a361a19b4b96","blindedQueryPhoneNumber":"n/I9srniwEHm5o6t3y0tTUB5fn7xjxRrLP1F/i8ORCdqV++WWiaAzUo3GA2UNHiB","account":"0x588e4b68193001e4d10928660aB4165b813717C0"}' -H 'Content-Type: application/json'
-export const getBlindedSalt = functions.https.onRequest(async (request, response) => {
```
codyborn, line 7: Thinking about the rollout.. It'd probably easiest to support backward compatibility in the server over the mobile client. We can leave both methods in here and once the mobile client is updated to use `getDistributedBlindedSalt` we can clean up the old `getBlindedSalt` method. WDYT?

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (!isValidGetSignatureInput(request.body)) {
+    respondWithError(response, 400, ErrorMessages.INVALID_INPUT)
+    return
+  }
+  if (!authenticateUser(request)) {
+    respondWithError(response, 401, ErrorMessages.UNAUTHENTICATED_USER)
+    return
+  }
+  const signatures: ServicePartialSignature[] = []
+  for (const service of config.pgpnpServices) {
```
codyborn: We should make each API call in parallel to reduce the overall latency. @tkporter has a good example here:
https://github.com/celo-org/celo-oracle/blob/2678f7046100f337dd46bdaac8ba8b969011a8bc/src/data_aggregator.ts#L269

https://github.com/celo-org/celo-oracle/blob/2678f7046100f337dd46bdaac8ba8b969011a8bc/src/utils.ts#L352

aslawson: great callout 👍

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
aslawson: ya -- copy pasta.  i'll fix it up

packages/phone-number-privacy/combiner/src/index.ts
aslawson: ya -- that's a possibility to have the mobile client call the combiner directly with the old method...🤔

i'll look into it as a good transition option

aslawson: alternative rossy suggested was just to stand it up separately and leave the current service in firebase running.  then redirect the mobile client to the new url of the combiner

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
aslawson: i used the variable `threshold` for the threshold number of signers needed and thus had to change this to be more descriptive

packages/phone-number-privacy/signer/test/salt-generation/bls-signature.test.ts
aslawson: ya -- that would be more descriptive.  the test will fail is this throws an error as is

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -0,0 +1,68 @@
...
+      method: 'POST',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        Authorization: request.headers.authorization!,
+      },
+      body: JSON.stringify(request.body),
+    })
+
+    // if res no ok, then it will not be added to the list of signatures
```
codyborn: nit: "if res not ok"

packages/phone-number-privacy/signer/.firebaserc
aslawson: Gotcha -- ya I had to add them to run it locally for testing

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
aslawson: I can try it -- I copied from this example
https://github.com/celo-org/blind-threshold-bls-wasm/commit/e1e2f8a1ab5154c2f0b1c55cb0d61fbb1d907208

packages/phone-number-privacy/combiner/scripts/createBlsKeyPair.ts
aslawson: Actually I'm not sure how this changed, but when I tried to change it back to #keygen, that function didn't exist...

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
aslawson: Sure -- I think that's a good idea for logging.  Though I don't think we should throw an Error, since the intension is for this to work if a signer fails, while a threshold of signers still succeed.

packages/phone-number-privacy/combiner/test/salt-generation/bls-signature.test.ts
```diff
@@ -32,14 +62,90 @@ describe(`BLS service computes signature`, () => {
...
+      {
+        url: 'url4',
+        signature:
+          'MAAAAAAAAAAhzTl/S+mldhE+5F5rt+2XKJQsNtELZeo+aoHjhsVVdw8Ofk1ZRr9EUZbvVKetNYADAAAA',
+      },
+    ]
+
+    const message = Buffer.from('hello world')
+    const userSeed = new Uint8Array(32)
+    for (let i = 0; i < 31; i++) {
```
codyborn: You can use: `i < userSeed.length - 1`

codyborn (dismissed):
Looks good! Added a couple of comments.

packages/phone-number-privacy/combiner/src/index.ts
codyborn: Rossy's idea sounds like less hassle 👍

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
codyborn: Got it. We can also change the other references to `threshold_bls` for consistency

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
codyborn: Agreed that we shouldn't let one signer have the ability to fail the call. However, if we don't have k/m signatures then we don't have another option but to fail.

jmrossy (approved):

PR #3944, opened 2020-06-02
[PGPNP] Convert Signer to be an express app

jmrossy:
### Description

Convert signer package to be an express app instead of a cloud function.
Update tests and configs accordingly

### Tested

Ran server locally

### Related issues

- #3899

### Backwards compatibility

No


aslawson (approved):
LGTM

PR #3945, opened 2020-06-02
Add to custodian docs

pranaymohan:
### Description

Update custodian + exchange documentation to cover the smart contracts in detail, and provide some context for the calls they need to make to track balance changing ops.


critesjosh (approved):
great additions!

yorhodes (approved):

PR #3946, opened 2020-06-02
POC for stitched blockchain-api

nambrot:
### Description

This PR creates a POC to expand blockchain-api to include a stitched schema consistenting of: - the current schema serving the wallet
- the entire schema of the blockscout API
- an additional outlet for contractKit calls


packages/blockchain-api/package.json
```diff
@@ -17,7 +17,7 @@
     "deploy": "./deploy.sh"
   },
   "dependencies": {
-    "@celo/contractkit": "0.3.6",
+    "@celo/contractkit": "0.4.6-dev",
```
jmrossy, line 20: bc-api can't use a -dev version because GAE installs the deps during deployment

packages/blockchain-api/src/cacheDataSource.ts
```diff
@@ -0,0 +1,22 @@
+import { DataSource, DataSourceConfig } from 'apollo-datasource'
+import { KeyValueCache } from 'apollo-server-caching'
+
+export default class CacheDataSource<T> extends DataSource {
+  // @ts-ignore We can't express that initialize will definitely be called
```
jmrossy, line 5: If we set it in the constructor instead, would that solve the TS error?

packages/blockchain-api/src/cacheDataSource.ts
```diff
@@ -0,0 +1,22 @@
+import { DataSource, DataSourceConfig } from 'apollo-datasource'
+import { KeyValueCache } from 'apollo-server-caching'
+
+export default class CacheDataSource<T> extends DataSource {
```
jmrossy, line 4: Nit: `CachedDataSource`?

packages/blockchain-api/src/contractKitDataSource.ts
```diff
@@ -0,0 +1,11 @@
+import { ContractKit, newKit } from '@celo/contractkit'
+import CacheDataSource from './cacheDataSource'
+
+export default class ContractKitDataSouce extends CacheDataSource<any> {
```
jmrossy, line 4: Is it potentially problematic if CK results are cached? Like a balance changes on-chain but the CK data source returns a cached value?

packages/blockchain-api/package.json
nambrot: Thanks, good reminder!

packages/blockchain-api/src/cacheDataSource.ts
nambrot: The way apollo-data sources are setup is that they receive the cache object in the `initialize` function unfortunately

packages/blockchain-api/src/contractKitDataSource.ts
nambrot: It's up to the consumer to define cache keys that appropriately consider expiration. I.e. including block or epoch numbers in the cache key should be part of it

packages/blockchain-api/src/apolloServer.ts
```diff
@@ -1,20 +1,56 @@
...
+  const schema = await introspectSchema(link)
+  const blockscoutSchema = makeRemoteExecutableSchema({
+    schema,
+    link,
+  })
+
+  const legacySchema = makeExecutableSchema({
+    typeDefs,
+    resolvers,
+  })
```
jeanregisser, line 34: nit: rename to `walletSchema`?

jeanregisser (commented):
Nice POC!
Though I understand the technical need for this, one drawback I see is the inconsistency it introduces with 3 different APIs stitched together.
I feel these could/should live together in a single project.
"One Celo API to rule them all" :D
And at the same time this might be quite difficult!

PR #3953, opened 2020-06-02
[Wallet Redesign ] Payment Request and Invite Redesign

tarikbellamine:
### Description
Redesign for the Payment Request and Invite flows

### Tested
Added tests as needed.

### Related issues
#3645 

### Backwards compatibility
Yes

IMG_4189

IMG_4190

IMG_4193

IMG_4192


packages/mobile/src/send/SendAmount.tsx
```diff
@@ -153,10 +153,7 @@ function SendAmount(props: Props) {
   const isDollarBalanceSufficient = isAmountValid && newAccountBalance.isGreaterThanOrEqualTo(0)
 
   const reviewBtnDisabled =
-    !isAmountValid ||
-    (isRequest
```
tarikbellamine, line 157: if `isRequest` and address isn't verified we will now route to an error screen instead of keeping the button disabled

packages/mobile/src/components/InviteOptionsModal.tsx
```diff
@@ -1,99 +1,57 @@
...
-        {this.state.inviteConfirmed && (
-          <View style={style.loadingIcon}>
-            <ActivityIndicator size="large" color={colors.celoGreen} />
-          </View>
-        )}
+  return (
+    <Modal isVisible={isVisible} style={styles.container}>
+      <View style={styles.buttonContainer}>
+        <SmsIcon />
+        <TextButton style={styles.buttonText} onPress={onSMS}>
```
i1skn, line 25: I'm concerned, that we are ditching here `accessibilityLabel`, but I see we do not have it all the way in our `Touchable`. So, I will create a separate issue for this.

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
```diff
@@ -110,32 +133,53 @@ class PaymentRequestConfirmation extends React.Component<Props> {
...
-  renderHeader = () => <ReviewHeader title={this.props.t('requestPayment')} />
-
   render() {
-    const { t } = this.props
-    const { amount, recipient, recipientAddress: requesteeAddress } = this.props.confirmationInput
+    const { t, confirmationInput } = this.props
+    const { recipient, recipientAddress: requesteeAddress } = confirmationInput
+    const amount = {
+      value: this.props.confirmationInput.amount,
+      currencyCode: CURRENCIES[CURRENCY_ENUM.DOLLAR].code, // Only cUSD for now
```
i1skn, line 139: Why only cUSD here?

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
```diff
@@ -110,32 +133,53 @@ class PaymentRequestConfirmation extends React.Component<Props> {
...
+    }
+
+    const renderFooter = () => (
```
i1skn: Can we implement `renderFooter` as a method instead of a lambda? Otherwise it would be recreated every time `render` in called.

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
```diff
@@ -110,32 +133,53 @@ class PaymentRequestConfirmation extends React.Component<Props> {
...
-          />
+          <View style={styles.transferContainer}>
+            <View style={styles.headerContainer}>
+              <ContactCircle
+                name={this.props.confirmationInput.recipient.displayName}
+                thumbnailPath={getRecipientThumbnail(recipient)}
+                address={requesteeAddress || ''}
+              />
+              <View style={styles.recipientInfoContainer}>
+                <Text style={styles.headerText}>Requesting</Text>
```
i1skn: I guess, we need to have `Requesting` in i18n files

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
```diff
@@ -145,8 +189,37 @@ class PaymentRequestConfirmation extends React.Component<Props> {
 const styles = StyleSheet.create({
   container: {
     flex: 1,
-    backgroundColor: colors.background,
-    paddingTop: 20,
+    backgroundColor: colors.light,
+    padding: 8,
+    flexDirection: 'column',
```
i1skn: AKAIK it's default `flexDirection` and can be avoided, the same down below a couple of times

packages/mobile/src/paymentRequest/PaymentRequestUnavailable.tsx
```diff
@@ -0,0 +1,60 @@
...
+type Props = RouteProps
+
+const PaymentRequestUnavailable = (props: Props) => {
+  const { t } = useTranslation(Namespaces.paymentRequestFlow)
+  const { recipient } = props.route.params.transactionData
+
+  return (
+    <SafeAreaView style={styles.container}>
+      <ScrollView contentContainerStyle={styles.scrollContainer}>
+        <Text style={styles.h2}>
```
i1skn: We usually call styles semantically, so you can easier identify where and for what it is used. Something like `recipientName` would fit better here. But not a big deal tbh.

packages/mobile/src/paymentRequest/PaymentRequestUnavailable.tsx
```diff
@@ -0,0 +1,60 @@
...
+  container: {
+    flex: 1,
+    backgroundColor: colors.light,
+    justifyContent: 'space-between',
+  },
+  scrollContainer: {
+    flex: 1,
+    paddingHorizontal: 24,
+    paddingVertical: 32,
+    flexDirection: 'column',
```
i1skn: ditto

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -260,6 +257,8 @@ function SendAmount(props: Props) {
         addressValidationType,
         isPaymentRequest: true,
       })
+    } else if (recipientVerificationStatus !== RecipientVerificationStatus.VERIFIED) {
+      navigate(Screens.PaymentRequestUnavailable, { transactionData })
```
i1skn, line 262: Do we want to fire `CeloAnalytics.track(CustomEventNames.request_continue, continueAnalyticsParams)` as well?

packages/mobile/src/utils/formatting.ts
```diff
@@ -126,7 +126,7 @@ export const formatDisplayName = (displayName: string) => {
   // Default displayName is "Mobile #" when no contact is saved
   // To make the copy nicer, we make the displayName "your contact" isntead
   if (displayName === 'Mobile #') {
-    return { displayName: 'your contact', displayNameCapitalized: 'Your contract' }
+    return { displayName: 'your contact', displayNameCapitalized: 'Your contact' }
```
i1skn: Ha, nice one! 👍

i1skn (commented):
Great work overall! See some minor comments I've left.

packages/mobile/src/components/InviteOptionsModal.tsx
tarikbellamine: Yeah, I noticed that as well. Thanks for opening the issue...would be great to have a general "accessibility" issue to make sure all of our buttons have it!

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
tarikbellamine: I'm not sure...this was hardcoded prior to redesign so I left it as is. I've noticed that in a couple of other places too. I think we way need a separate issue for this whenever we feel we are ready to enable other currencies

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
tarikbellamine: Good catch. Will do

packages/mobile/src/paymentRequest/PaymentRequestUnavailable.tsx
tarikbellamine: You're right, better to get into the habit. Will do!

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
tarikbellamine: Oh nice, didn't know that. Thanks!

packages/mobile/locales/en-US/paymentRequestFlow.json
```diff
@@ -6,5 +6,12 @@
...
   "paymentRequestUpdateFailed": "Cannot update payment request",
   "incomingPaymentRequestNotificationTitle": "{{name}} requested",
-  "outgoingPaymentRequestNotificationTitle": "Waiting for {{name}}"
+  "outgoingPaymentRequestNotificationTitle": "Waiting for {{name}}",
+  "requestUnavailableHeader": "{{displayName}}'s account is unverified",
+  "requestUnavailableNoDisplayNameHeader": "Your contact's account is unverified",
+  "requestUnavailableBody": "The number {{e164PhoneNumber}} doesn't have an account or hasn't verified it yet. Requests can only be sent to verified numbers.",
+  "requestUnavailableNoNumberBody": "Your contact doesn't have an account or hasn't verified it yet. Requests can only be sent to verified numbers.",
+  "requesting": "Requesting",
+  "request": "Request",
```
i1skn, line 15: There are should be `request` already in Global. Usually for one-worders we try to put in `global`

packages/mobile/src/analytics/constants.ts
```diff
@@ -43,6 +43,7 @@ export enum CustomEventNames {
   request_select_recipient = 'request_select_recipient', // when recipient was selected
   request_amount_back = 'request_amount_back', // back button pressed on amount screen
   request_continue = 'request_continue', // when next button pressed on amount enter page
+  request_unavailable = 'request_unavailable', // when request attempted to unverified user
```
i1skn, line 46: Please add it as well here https://github.com/celo-org/celo-monorepo/blob/e751e5a851563a14961089cc2e870e320ef9a307/packages/mobile/src/analytics/constants.ts#L237. Otherwise it would be filtered.

i1skn (approved):
LGTM with the exception of couple of comments I've wrote! Great work 💪 !

PR #3955, opened 2020-06-03
Adding transfers with comments to CLI

mrsmkl:
### Description

Added option to have transfer comments to `transfer:gold` and `transfer:dollars`

### Other changes

Also added a check that there are enough tokens. It doesn't include gas cost so not best possible.

### Tested

Only manual testing.


packages/cli/src/utils/checks.ts
```diff
@@ -295,6 +295,16 @@ class CheckBuilder {
     )
   }
 
+  hasEnoughUsd = (account: Address, value: BigNumber) => {
+    const valueInEth = this.kit.web3.utils.fromWei(value.toFixed(), 'ether')
+    return this.addCheck(`Account has at least ${valueInEth} cUSD`, () =>
+      this.kit.contracts
+        .getStableToken()
+        .then((gt) => gt.balanceOf(account))
```
m-chrzan: Should this be `st`? I personally would even prefer not abbreviating here or above, `stableToken` and `goldToken`.

m-chrzan (approved):
Looks good, just one comment on naming.

packages/cli/src/utils/checks.ts
mrsmkl: Yeah, decided to replace other abbreviations from this file, too.

PR #3971, opened 2020-06-03
EventKit 1 (pre work)

### LINKED ISSUE #3690: Left Nav runs in to footer in BrandKit
### Expected Behavior

Have the left nav become static before the footer appears

### Current Behavior

What is the current behavior?
image

aaronmgdr:
### Description

Fix side bar on for Experience Kits

create an experience folder for all kits `src/brand => src/experience/brand` this way there is one spot where all experience kits stuff is kept.

fix it so we can have ordered lists not just unordered. 

### Tested

yes and deployable 

### Related issues

- Fixes #3690 

### Backwards compatibility
yes


packages/web/src/experience/common/Page.tsx
```diff
@@ -29,106 +30,31 @@ export const ICONS_PATH = `${ROOT}/icons`
...
   title: string
   path: string
   metaDescription: string
+  ogImage: ImageRequireSource
 }
 
 interface State {
   routeHash: string
   isSidebarFrozen: boolean
-  distanceToTop: number
```
aaronmgdr, line 123: no longer needed since now using `postition: sticky`

packages/web/src/experience/common/TopBar.tsx
```diff
@@ -0,0 +1,94 @@
+import * as React from 'react'
```
aaronmgdr, line 1: a lot of this is actually not knew. I moved the file but then edited it some

packages/web/src/experience/common/scrollToHash.ts
```diff
@@ -0,0 +1,10 @@
+import scrollIntoView from 'scroll-into-view'
+import { HEADER_HEIGHT } from 'src/shared/Styles'
+
+export default function scrollToHash(extraDistance: number) {
+  const element = document.getElementById(window.location.hash.replace('#', ''))
+  scrollIntoView(element, {
```
jmrossy: Just curious, what happens if element doesn't exist?

jmrossy (approved):
LGTM, just one question

packages/web/src/experience/common/scrollToHash.ts
aaronmgdr: oh no 😱

PR #3976, opened 2020-06-04
Feat/eksportisto sensitive accounts

AlexBHarley:
### Description

Add a `sensitive-accounts.json` file to the eksportisto deployment. Any transfer from an account in this json file will trigger a request to a configurable URL.

### Tested

A bunch in GCP


packages/celotool/src/lib/eksportisto.ts
```diff
@@ -29,6 +29,12 @@ export async function removeHelmRelease(celoEnv: string) {
   await execCmdWithExitOnFailure(`helm del --purge ${celoEnv}-eksportisto-${suffix}`)
 }
 
+function fetchSensitiveAccounts() {
+  return JSON.stringify({
+    "0x456f41406B32c45D59E539e4BBA3D7898c3584dA": "https://example.com"
```
nambrot: Let's keep it empty as the default?

packages/helm-charts/eksportisto/templates/statefulset.yaml
```diff
@@ -35,15 +35,24 @@ spec:
...
       containers:
       - name: eksportisto-{{ .Values.deploymentSuffix }}
         image: {{ .Values.imageRepository }}:{{ .Values.imageTag }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
         args:
-          - --nodeUri
-          - $(WEB3_PROVIDER)
-          - --datadir
-          - /root/.eksportisto
+          - "-nodeUri=https://alfajores-forno.celo-testnet.org" # $(WEB3_PROVIDER)
```
nambrot: You probably don't want to check this in?

nambrot (approved):

PR #4017, opened 2020-06-05
[Wallet][ReDesign] QR Code / Scan screens

jeanregisser:
### Description

This implements the [QR code scan redesign](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=2644%3A2625).

Note: 
- The state when Camera permission is denied is not with a white background like in the design but in dark to simplify the integration with the animation.
- Avatar used is the old one.

### Other changes

- Improve `Touchable` props types to accept all props from `TouchableWithoutFeedback` (accessibility, testID, etc)

### Tested

qr-anim3

### Related issues

- Epic #3645

### Backwards compatibility

Yes.


packages/mobile/src/qrcode/QRGen.tsx
```diff
@@ -64,54 +69,40 @@ export function transformMatrixIntoPath(cellSize: number, matrix: any) {
...
-  static defaultProps = {
-    value: 'This is a QR Code.',
-    size: 100,
-    color: 'black',
-    backgroundColor: 'white',
-    ecl: 'M',
-    onError: undefined,
-  }
+function QRCode({
+  value = 'This is a QR Code.',
```
jmrossy: I see this was taken from the old defaultProps but why do we set this?

aaronmgdr:
oh wow that slide action! 🤩

packages/mobile/src/qrcode/QRScanner.tsx
```diff
@@ -1,176 +1,95 @@
...
-  // QR code, so we need to catch only the first one
-  onBarCodeDetected = memoize(
-    (qrCode: QrCode) => {
-      Logger.debug('QRScanner', 'Bar code detected')
-      const { scanIsForSecureSend, transactionData } = this.props
-      this.props.handleBarcodeDetected(qrCode, scanIsForSecureSend, transactionData)
-    },
-    (qrCode) => qrCode.data
+  const onBarCodeDetected = useCallback(
+    memoize(
```
jmrossy, line 60: Does this work? Or do we create a new memoized function on each render?

packages/mobile/src/navigator/QRNavigator.tsx
```diff
@@ -0,0 +1,136 @@
...
+  position: Animated.Value<number>
+}
+
+// Component doing our custom transition for the QR scanner
+function AnimatedScannerScene({ position, ...props }: AnimatedScannerSceneProps) {
+  const isFocused = useIsFocused()
+  const [wasFocused, setWasFocused] = useState(isFocused)
+  const [isPartiallyVisible, setIsPartiallyVisible] = useState(false)
+  const cameraPermission = useAsync(check, [
+    Platform.select({ ios: PERMISSIONS.IOS.CAMERA, default: PERMISSIONS.ANDROID.CAMERA }),
```
aaronmgdr, line 36: why default? and not android?

packages/mobile/src/qrcode/QRScanner.tsx
```diff
@@ -1,176 +1,95 @@
...
 
-class QRScanner extends React.Component<Props> {
-  static navigationOptions = () => ({
-    ...headerWithBackButton,
-    headerTitle: i18n.t('sendFlow7:scanCode'),
-  })
+export default function QRScanner({ route }: Props) {
+  const dispatch = useDispatch()
+  const { t } = useTranslation(Namespaces.sendFlow7)
+  const inset = useSafeArea()
```
jmrossy, line 55: I'm wondering if this would work better as an HOC instead of getting the inset and manually using it in margins like we do here.

packages/mobile/src/qrcode/QRTabBar.tsx
```diff
@@ -0,0 +1,102 @@
...
+    outputRange: [1, 0],
+  })
+
+  const color = interpolateColors(position, {
+    inputRange: [0.9, 1],
+    outputColorRange: [colors.dark, colors.white],
+  })
+
+  const onPressClose = () => {
+    navigation.dangerouslyGetParent()?.goBack()
```
jmrossy, line 43: Is this method... dangerous?

packages/mobile/src/navigator/QRNavigator.tsx
```diff
@@ -0,0 +1,136 @@
...
+      onChange(
+        greaterThan(position, 0),
+        call([position], ([value]) => {
+          setIsPartiallyVisible(value > 0)
+        })
+      ),
+    [position]
+  )
+
+  const animatedStyle = useMemo(() => {
```
aaronmgdr, line 59: one thing ive been thinking about is defining big hooks like this outside the component and then just calling them. 

at best its more performant (no proof) at worst it keeps the voice in my head that functions should be 10 lines or less feeling less ignored.

packages/mobile/src/send/QRCodeIcon.tsx
```diff
@@ -7,7 +7,7 @@ import { Screens } from 'src/navigator/Screens'
 
 export function QRCodeIcon() {
   const onQrCodePress = React.useCallback(() => {
-    navigate(Screens.QRScanner)
+    navigate(Screens.QRNavigator, { screen: Screens.QRScanner })
```
jmrossy, line 10: Is `screen` the default screen here? If so, maybe rename it?

packages/mobile/src/navigator/types.tsx
```diff
@@ -9,7 +9,14 @@ import { TransactionDataInput } from 'src/send/SendAmount'
 import { ReviewProps } from 'src/transactions/TransactionReview'
 import { TransferConfirmationCardProps } from 'src/transactions/TransferConfirmationCard'
 
-// tslint:disable-next-line
+// Typed nested navigator params
+type NestedNavigatorParams<ParamList> = {
```
aaronmgdr, line 13: bravo

packages/react-components/components/BorderlessButton.v2.tsx
```diff
@@ -2,16 +2,15 @@ import Touchable, { Props as TouchableProps } from '@celo/react-components/compo
...
-type Props = TouchableProps & {
+export type Props = Omit<TouchableProps, 'style'> & {
   style?: StyleProp<TextStyle>
-  testID?: string
 }
 
 // unstyled Touchable Text, good for making other Text Buttons such as TopBarButton
 export default function BoarderlessButton(props: Props) {
-  const { onPress, style, children, disabled, testID } = props
+  const { style, children, ...rest } = props
```
jmrossy: Nit: I believe `passThroughProps` is a common convention for `rest` here

packages/mobile/src/navigator/QRNavigator.tsx
jeanregisser: yes I feel you, TypeScript complains otherwise because `Platform.select` return type would also contain `undefined`. Which is not accepted by the `check` function.

patches/react-native-tab-view+2.14.4.patch
```diff
@@ -0,0 +1,36 @@
+patch-package
```
jmrossy, line 1: Is there a github issue open on the repo with this issue? If not, should we open one? Or use a diff lib we don't have to patch?

jmrossy (commented):
Looks great!! Just a few questions

packages/mobile/src/navigator/QRNavigator.tsx
aaronmgdr: ahh.

packages/mobile/src/qrcode/QRScanner.tsx
aaronmgdr: yeah im not sure i get why its memoized.

packages/mobile/src/navigator/QRNavigator.tsx
jeanregisser: yes, kinda torn with this too, I'd say it depends, sometimes it's good to have all the context in a "big" function.
Here the function is still relatively short.

packages/mobile/src/navigator/types.tsx
jeanregisser: Thanks! Had to think carefully about this 🤯 😄

packages/mobile/src/qrcode/QRGen.tsx
jeanregisser: Yes don't know why it was like this, I'll remove this.

packages/mobile/src/qrcode/QRScanner.tsx
jeanregisser: Yes not very obvious on the first look indeed.

It does create a new memoized version on each render, but only the first memoized version is returned, unless `[scanIsForSecureSend, transactionData]` change.

We could think calling `memoize` on each render seems heavy, but it's not very different from creating a new function in a standard `useCallback` call.

See https://github.com/lodash/lodash/blob/master/memoize.js#L43-L60

packages/mobile/src/qrcode/QRScanner.tsx
jeanregisser: I'm not sure I get the concern here.
I initially used `SafeAreaView`, but it was causing a visual glitch in the animation because of its use of `onLayout`.

packages/mobile/src/qrcode/QRTabBar.tsx
jeanregisser: Good question 😄 
From the docs: 
> Reason why the function is called dangerouslyGetParent is to warn developers against overusing it to eg. get parent of parent and other hard-to-follow patterns.

packages/mobile/src/send/QRCodeIcon.tsx
jeanregisser: No, default screen is `QRCode`

packages/react-components/components/BorderlessButton.v2.tsx
jeanregisser: Agreed 👍

patches/react-native-tab-view+2.14.4.patch
jeanregisser: Very good concern, cut a few corners to avoid spending too much time on this here.
I plan to open a PR on `react-native-tab-view`. 
The lib itself is very solid, I've used it in a bunch of past projects and a lot of effort has been put into it to provide native feeling for this.
It was created by @satya164 which is also one of the creator of of `react-navigation`.

jmrossy (approved):

PR #4018, opened 2020-06-05
Add tooltips on Validator explorer

Pedro-vk:
### Description

_Add tooltips on Validator explorer_

### Tested

_Tested on Chrome_

### DIFF
--- a/packages/web/src/dev/ValidatorsListStyles.tsx
+++ b/packages/web/src/dev/ValidatorsListStyles.tsx
@@ -7,6 +7,11 @@ export const styles = StyleSheet.create({
     position: 'static',
     zIndex: 'initial',
   } as any,
+  pStaticOverflow: {
+    position: 'static',
+    zIndex: 'initial',
+    overflow: 'initial',
+  } as any,
   content: {
     paddingBottom: 10,
   },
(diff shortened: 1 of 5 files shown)


packages/web/src/shared/Hoverable.tsx
```diff
@@ -25,6 +25,7 @@ export default class Hoverable extends React.Component<Props, State> {
   }
 
   checkIfHoverEnabled = () => {
+    console.log('check')
```
aaronmgdr: remove before merge

packages/web/src/dev/ValidatorsListStyles.tsx
```diff
@@ -7,6 +7,11 @@ export const styles = StyleSheet.create({
     position: 'static',
     zIndex: 'initial',
   } as any,
+  pStaticOverflow: {
+    position: 'static',
+    zIndex: 'initial',
+    overflow: 'initial',
```
aaronmgdr, line 13: curious which of these is causing it to need to be typed as any?

aaronmgdr (approved):

packages/web/src/dev/ValidatorsListStyles.tsx
Pedro-vk: The "initial" value seems not supported

PR #4019, opened 2020-06-05
[Wallet] Animated balance in the Gold screen header

i1skn:
### Description

Implement animated balance in the Gold screen header

### Other changes

* `<DrawerTopBar />` cosmetic lift up:
  * Because we need shadow only when the content is scrolled and only beneath it - I replaced it with border for now, cause have not find a good way of doing "only bottom" shadow for both platforms.
  * Replace hardcoded Logo with middleElement, cause of different screen have different elements or none in the middle of `<DrawerTopBar />` 

* Change the way how we calculate data for Gold prices chart. That was needed to show numbers in the animated header. Also, I've made the way it would be easier to integrate, when we have aggregated rates in `blockchain-api`

### Tested

Android and iOS simulators.

### Backwards compatibility

Yes

### Screenshots
Screenshot_1591356336
Screenshot_1591356335
Screenshot_1591356776
Screenshot_1591356342


packages/mobile/src/exchange/CeloGoldHistoryChart.tsx
```diff
@@ -80,6 +78,7 @@ function ChartAwareSvgText({
       onLayout={onLayout}
       fill={colors.gray4}
       fontSize="14"
+      fontFamily="Hind-Regular"
```
jeanregisser: Shouldn't this be `Inter-Regular` instead?

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,247 @@
...
+  const headerOpacity = useCallback(
+    () => ({
+      opacity: animatedValue.interpolate({
+        inputRange: [0, 100],
+        outputRange: [0, 1],
+        extrapolate: Animated.Extrapolate.CLAMP,
+      }),
+    }),
+    []
+  )
```
jeanregisser: `useMemo` would be slightly nicer.

packages/mobile/src/exchange/reducer.ts
```diff
@@ -18,36 +19,73 @@ export interface State {
   exchangeRatePair: ExchangeRatePair | null
   tobinTax: string | null
   history: {
+    // TODO this should be remove once we have aggreagation on
```
jeanregisser: typo `aggregation`

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,247 @@
...
+                    setIsScrolled(offsetY > 0)
+                  }),
+                ]),
+            },
+          },
+        },
+      ],
+      {
+        useNativeDriver: true,
+      }
```
jeanregisser: AFAIK this is unnecessary with `react-native-reanimated`.

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,247 @@
...
+    Animated.event(
+      [
+        {
+          nativeEvent: {
+            contentOffset: {
+              y: (y: Animated.Node<number>) =>
+                Animated.block([
+                  Animated.set(animatedValue, y),
+                  Animated.call([y], ([offsetY]) => {
+                    setIsScrolled(offsetY > 0)
```
jeanregisser: Maybe this is fine for now, but this is gonna trigger `setIsScrolled` every 50 ms as the user scrolls.
Would be better to use an `Animated.Value` for this, and then pass it to `DrawerTopBar`.

I've got an old Nexus 5 here where the scroll perf could be impacted very much by this.
I can try without and without this to see the difference.

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -72,28 +78,18 @@ const mapStateToProps = (state: RootState): StateProps => ({
...
-    })
-    this.onScroll = Animated.event(
-      [{ nativeEvent: { contentOffset: { y: this.animatedValue } } }],
-      {
-        useNativeDriver: true,
-      }
-    )
+    this.onScroll = (event) => {
+      this.setState({ isScrolled: event.nativeEvent.contentOffset.y > 0 })
+    }
```
jeanregisser: Again maybe this is fine, but this is gonna call `setState` every 50 ms when you scroll.
It would be more performant to use an `Animated.Value` for this.

jeanregisser (commented):
Great! 🎉 
Just a small comment about a possible issue with the scroll performance given how the logic is currently done to update `DrawerTopBar`'s look using component state.

cmcewen (changes requested):
Just wanted to flag the onScroll setting state point Jean mentioned - will be quite bad for performance, so please don't use that 😄. You can interpolate opacity (https://reactnative.dev/docs/animations#interpolation), or try reanimated - haven't used it myself but heard good things

i1skn:
@jeanregisser @cmcewen Big thanks for the feedback! I've changed the logic how `onScroll` is handled and now it uses `react-native-reanimated` instead of state manipulations. PTAL

packages/mobile/src/exchange/CeloGoldHistoryChart.tsx
```diff
@@ -300,30 +265,9 @@ const styles = StyleSheet.create({
...
-  goldPriceWentDown: {
-    ...fontStyles.regular,
-    marginBottom: 2, // vertically align with the current price
-    marginLeft: 4,
-    color: colors.warning,
-  },
   loader: {
     width: CHART_WIDTH,
-    height: CHART_HEIGHT + 130,
+    height: CHART_HEIGHT + 37.5,
```
jeanregisser, line 268: 37.5 is very precise 😄

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,235 @@
...
-interface DispatchProps {
-  fetchExchangeRate: typeof fetchExchangeRate
-}
+import { goldToDollarAmount } from 'src/utils/currencyExchange'
+import { getLocalCurrencyDisplayValue } from 'src/utils/formatting'
 
 type OwnProps = StackScreenProps<StackParamList, Screens.ExchangeHomeScreen>
 
-type Props = StateProps & DispatchProps & WithTranslation & OwnProps
+type Props = OwnProps
```
jeanregisser: nit: we can now assign `StackScreenProps<StackParamList, Screens.ExchangeHomeScreen>` directly to props.

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,235 @@
...
 
-const mapStateToProps = (state: RootState): StateProps => ({
-  exchangeRate: getRateForMakerToken(state.exchange.exchangeRatePair, CURRENCY_ENUM.DOLLAR),
-  goldBalance: state.goldToken.balance,
-  dollarBalance: state.stableToken.balance,
-})
+function ExchangeHomeScreen(_props: Props) {
+  const scrollPosition = useRef(new Animated.Value(0)).current
+  const navigation = useNavigation()
```
jeanregisser: nit: navigation is passed as a prop here, we could use it directly instead of using the hook.

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -1,46 +1,62 @@
...
+import Animated from 'react-native-reanimated'
 
-const DrawerTopBar = () => {
+interface Props {
+  middleElement?: React.ReactNode
+  scrollPosition?: Animated.Value<number>
+}
+
+// Test mocks would not work if import these together with Animated
+const { cond, greaterThan } = Animated
```
jeanregisser: Oh ok, I believe this is because we need to do this in `__mocks__/react-native-reanimated`:
```
module.exports = require('react-native-reanimated/mock')
```

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -1,46 +1,62 @@
...
+    () => ({
+      borderBottomColor: colors.gray2,
+      borderBottomWidth: cond(greaterThan(scrollPosition ?? new Animated.Value(0), 0), 1, 0),
+    }),
+    [scrollPosition]
+  )
 
   return (
-    <View style={styles.container}>
+    <Animated.View style={[styles.container, borderBottom]}>
```
jeanregisser: When using `Animated.View`, it's more performant to pass a style that doesn't change.
We could easily change `useMemo` above to output the full style.

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -159,8 +149,9 @@ export class WalletHome extends React.Component<Props> {
 
     return (
       <SafeAreaView style={styles.container}>
-        <DrawerTopBar />
+        <DrawerTopBar middleElement={<Logo />} scrollPosition={this.scrollPosition} />
         <AnimatedSectionList
+          scrollEventThrottle={50}
```
jeanregisser: Reading up on this with `react-native-reanimated`, it looks like it's required to declare a value on iOS, otherwise the animation can lag behind, but I think the actual value is ignored.
But what do you think of changing it to `scrollEventThrottle={1}` to clearly indicate we want the animation to be truly in sync with the scroll movement?

packages/mobile/src/exchange/CeloGoldHistoryChart.tsx
i1skn: Well, this is the size of the block with Gold exchange rate. It is needed, so after loader disappears - buttons below it would not flicker. I've found it using binary search, and neither 37 nor 38 were not good enough 😄

packages/mobile/src/navigator/DrawerTopBar.tsx
i1skn: Good catch, did not think of that. So used to `import/export` nowadays 😄

packages/mobile/src/home/WalletHome.tsx
i1skn: AFAIK value is in ms, most examples use 16 or 1, so I'm a bit confused here. I'm proposing to use here 16 (corresponds to 60+ fps). WDYT?

i1skn:
@jeanregisser I've addressed the feedback, PTAL.

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,231 @@
...
+
+  const headerOpacity = useMemo(
+    () => ({
+      opacity: scrollPosition.interpolate({
+        inputRange: [0, 100],
+        outputRange: [0, 1],
+        extrapolate: Animated.Extrapolate.CLAMP,
+      }),
+    }),
+    []
```
cmcewen: shouldn't scrollPosition be in this array?

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,231 @@
...
+  const dollarBalance = useSelector((state) => state.stableToken.balance)
+  const goldBalance = useSelector((state) => state.goldToken.balance)
+  const dispatch = useDispatch()
 
-type OwnProps = StackScreenProps<StackParamList, Screens.ExchangeHomeScreen>
+  // TODO: revert this back to `useLocalCurrencyCode()` when we have history data for cGDL to Local Currency.
+  const localCurrencyCode = null
+  const localExchangeRate = useSelector(getLocalCurrencyExchangeRate)
+  const currentExchangeRate = useExchangeRate()
+  const goldToDollars = useCallback((amount) => goldToDollarAmount(amount, currentExchangeRate), [
```
cmcewen: reading this again (https://kentcdodds.com/blog/usememo-and-usecallback), just wondering if the useCallback is worth it here or if it just adds unnecessary complexity

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,231 @@
...
+    Animated.event([
+      {
+        nativeEvent: {
+          contentOffset: {
+            y: scrollPosition,
+          },
+        },
+      },
+    ]),
+    []
```
cmcewen: same - scrollPosition?

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -1,126 +1,231 @@
...
 
-export class ExchangeHomeScreen extends React.Component<Props> {
-  componentDidMount() {
-    this.props.fetchExchangeRate()
-  }
+  useEffect(() => {
+    dispatch(fetchExchangeRate())
+  }, [])
+
+  const headerOpacity = useMemo(
```
cmcewen: i'm a little confused about the necessity of useMemo here

cmcewen (commented):
animation is cool 👍. i am not a hook expert, but i am wondering about useCallback/useMemo everywhere and if we should save it for calculations that we know to be expensive

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
i1skn: As @jeanregisser pointed out:
> When using Animated.View, it's more performant to pass a style that doesn't change.

My understanding, that changing `Animated.View->style` could lead to some heavy computations (re-inits and checking the current state of the animations, probably resetting it), so we are memoing the style here.

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
i1skn: I've moved unnecessary `useCallback` to the top of the component and defined them as `function() { ... }` instead of arrow function. I think, it has improved the readability. LMK WDYT?

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
i1skn: As it states here https://reactjs.org/docs/hooks-reference.html#useref, `scrollPosition` will persist for the full lifetime of the component, so it is unnecessary to add it to deps array.

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
i1skn: See above

jeanregisser (approved):
Nice!! Ship it! 🎉

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
jeanregisser: Yes exactly. It has to recreate animation nodes if they are different on each render.
You can see it by following what it does here https://github.com/software-mansion/react-native-reanimated/blob/1.9.0/src/createAnimatedComponent.js#L141-L161
Also more discussion here: https://github.com/software-mansion/react-native-reanimated/issues/267

i1skn:
@cmcewen PTAL whenever you would have time 🙂

cmcewen (approved):

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -1,46 +1,60 @@
...
     flexDirection: 'row',
     height: 62,
-    backgroundColor: colorsV2.white,
-    alignItems: 'flex-end',
+    alignItems: 'center',
+    backgroundColor: colors.background,
     justifyContent: 'space-between',
-    ...elevationShadowStyle(1),
+  },
+  withBorderBottom: {
```
tarikbellamine, line 49: It doesnt look like this styling is being applied anywhere

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -1,46 +1,60 @@
...
-        <Logo />
-      </View>
+      {middleElement}
       <View style={styles.spacer} />
-    </View>
+    </Animated.View>
   )
 }
 
+DrawerTopBar.defaultProps = {
```
tarikbellamine, line 37: This also isn't being applied and I think it should be. It would be nice if the default would be for the bar to have the wallet icon present and have a property we can pass in if we want to hide it or show a custom one.

Think that would be the easiest way to avoid importing the logo in a bunch of different places and sometimes forgetting to do it.

i1skn: I see the wallet logo in the top bar only for Home screen. I might be missing something here, are we using somewhere else?

packages/mobile/src/navigator/DrawerTopBar.tsx
i1skn: good catch, will address it as a part of my next PR 👍

packages/mobile/src/navigator/DrawerTopBar.tsx
tarikbellamine: Hmm yeah it looks like you are right. For some reason I thought we wanted it in more places but I guess not!

PR #4022, opened 2020-06-05
[Wallet][Redesign] Add funds / Cash out

i1skn:
### Description

Implement redesign for Add funds / Cash out screens https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=3754%3A506

#### Things left
- [ ] Pass amount to moonpay API when generaing web page
- [ ] Configure colors if possible for Moonpay web page

### Other changes

Fix crash on `Licenses` screen after moving from it.

### Tested

Android device

### Backwards compatibility

Yes

### Screenshots

5raSXy1nKSCVDNoQTSzzA2WLdTEg6umc
fU5Hp50iODIKTgEtac0if3ykcrDVmy1m
qlIpr4gIKaCcU8IBmPKo3rOlETRXcrzm
YVRq6xjosmbwbhZAvK0t5VdECoI2Eluw
s1L5ebS5Eu4YHHinR8498jN2eikHbUwP
9lm9ljSxqCrWEleEakkWplCBMg4mZoym


packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -1,104 +1,125 @@
...
-}
-
-class FiatExchange extends React.Component<Props, State> {
-  state: State = {
-    signedUrl: '',
+function FiatExchange(props: Props) {
+  const { t } = useTranslation()
+  const navigation = useNavigation()
+  const goToAddFunds = React.useCallback(() => {
+    navigation.navigate(Screens.FiatExchange)
```
cmcewen: I'm not sure if this would ever happen, but shouldn't navigation be in the dependencies array?

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -1,104 +1,125 @@
...
-        this.handleError()
-      }
+  let localAmount
+  if (!isUsdLocalCurrency) {
+    localAmount = {
+      value: localBalance,
+      currencyCode: localCurrencyCode,
     }
   }
+  console.log(dollarAmount)
```
cmcewen: Don't forget these

packages/mobile/src/fiatExchanges/Moonpay.tsx
```diff
@@ -0,0 +1,97 @@
...
+  return {
+    localCurrency: getLocalCurrencyCode(state),
+    account: state.web3.account,
+  }
+}
+
+type Props = StateProps & WithTranslation
+
+const celoCurrencyCode = 'CUSD'
+const signMoonpayFirebaseUrl = 'https://us-central1-celo-org-mobile.cloudfunctions.net/signMoonpay'
```
cmcewen: should this be in the config file?

packages/mobile/src/fiatExchanges/Moonpay.tsx
```diff
@@ -0,0 +1,97 @@
...
+      }
+    }
+  }
+
+  async componentDidMount() {
+    await this.updateMoonpayUrl()
+  }
+
+  handleError = () => {
+    showError(ErrorMessages.FIREBASE_FAILED)
```
cmcewen: @annakaz we should probably show a different error here, right?

packages/mobile/src/account/FiatExchange.tsx
i1skn: Yeap, it is very much in progress, not really ready to review yet :) But thanks for the comments, gives me some good ideas on how to improve and continue!

packages/mobile/src/fiatExchanges/Moonpay.tsx
annakaz: If the moonpayUrl is not updated then it would be a failure of the firebase function. Added a comment to clarify
```suggestion
    showError(ErrorMessages.FIREBASE_FAILED) // Firebase signing function failed
```

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+function FiatExchangeOptions({ route, navigation }: Props) {
+  function goToProvider(screen: keyof StackParamList) {
+    return () => navigation.navigate(screen)
+  }
+  function onDismiss() {
+    navigation.setParams({ isExplanationOpen: false })
+  }
+
+  const providers: {
```
i1skn, line 70: was not sure where to put them, so left them here

packages/mobile/src/fiatExchanges/MoonPay.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { getLocalCurrencyCode } from 'src/localCurrency/selectors'
+import { currentAccountSelector } from 'src/web3/selectors'
+
+const celoCurrencyCode = 'CUSD'
+
+function FiatExchangeWeb() {
+  const [uri, setUri] = React.useState('')
+  const account = useSelector(currentAccountSelector)
+  const localCurrencyCode = useSelector(getLocalCurrencyCode)
+  React.useEffect(() => {
```
aaronmgdr, line 34: what do you think about defining this custom hook outside of the function? 

or having a getSignedURL function outside of the hook that takes in `amount, celoCurrencyCode, account, localCurrencyCode as arguments` ? 

or both?

packages/mobile/src/config.ts
```diff
@@ -71,6 +77,7 @@ export const DEFAULT_SYNC_MODE: GethSyncMode = Config.DEFAULT_SYNC_MODE
   : GethSyncMode.Lightest
 // TODO Remove when feature is stable
 export const USE_PHONE_NUMBER_PRIVACY = true
+export const SHOW_CASH_OUT = stringToBoolean(Config.SHOW_CASH_OUT || 'false')
```
aaronmgdr, line 78: if `Config.SHOW_CASH_OUT` is a string then  `Config.SHOW_CASH_OUT || 'false'` wont the false here not ever be needed?  i guess it would for the empty string state?

packages/mobile/src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -0,0 +1,193 @@
...
+  subtotalBodyText: {
+    ...fontStyles.small,
+  },
+  currencyInput: {
+    ...fontStyles.regular,
+    marginLeft: 10,
+    flex: 1,
+    textAlign: 'right',
+    fontSize: 19,
+    lineHeight: Platform.select({ android: 27, ios: 23 }), // vertical align = center
```
aaronmgdr, line 204: have you check this on different ios and different android phones? seems like something that could be more about a specific phone size  than platform.

packages/mobile/src/fiatExchanges/MoonPay.tsx
i1skn: What would be the benefits of doing it this way?

packages/mobile/src/config.ts
i1skn: Yeah, if `Config.SHOW_CASH_OUT` is an empty string or undefined (which should be the case for environments in which we do not want this feature) this would evaluate to `false`.

packages/mobile/src/fiatExchanges/FiatExchangeAmount.tsx
i1skn: I have not checked more than two phones (real one and emulator). We have the same logic at `ExchangeTrade`, here the height was changed, therefor I've adjusted lineHeight as well. To my understanding it depends on the height of the input, which is hardcoded a couple lines later, not the screen size. But I might be wrong 🙂

packages/mobile/locales/en-US/fiatExchangeFlow.json
```diff
@@ -0,0 +1,11 @@
+{
+  "add_funds": "Add funds",
+  "cash_out": "Cash out",
```
aaronmgdr: any reason these are snake_case and the rest are Camel?

packages/mobile/src/fiatExchanges/MoonPay.tsx
aaronmgdr: the intention would be a smaller function size. 

I just think its a nice way to organize :)

aaronmgdr (approved):

packages/mobile/src/config.ts
```diff
@@ -57,6 +61,8 @@ export const FAQ_LINK = 'https://celo.org/faq'
 export const FORUM_LINK = 'https://forum.celo.org/c/support'
 export const CELO_SUPPORT_EMAIL_ADDRESS = 'support@celo.org'
 export const DEFAULT_FORNO_URL = `https://${DEFAULT_TESTNET}-forno.celo-testnet.org`
+export const SIGN_MOONPAY_FIREBASE_URL =
+  'https://us-central1-celo-org-mobile.cloudfunctions.net/signMoonpay'
```
jmrossy: Should this vary for different environments? If so, maybe move it to `network-config.ts`

packages/mobile/src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -0,0 +1,193 @@
...
+type Props = RouteProps
+
+const oneDollarAmount = {
+  value: new BigNumber('1'),
+  currencyCode: CURRENCIES[CURRENCY_ENUM.DOLLAR].code,
+}
+
+export function ExchangeTradeScreen({ navigation, route }: Props) {
+  function isNextButtonValid() {
+    const passMinAmount = dollarAmount.isGreaterThanOrEqualTo(
```
jmrossy, line 70: I find it strange that you can reference `dollarAmount` before its been initialized. JS is strange.

packages/mobile/src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -0,0 +1,193 @@
...
+    (!parsedInputAmount.isNaN() &&
+      convertLocalAmountToDollars(parsedInputAmount, localCurrencyCode ? localExchangeRate : 1)) ||
+      new BigNumber('0')
+  )
+
+  return (
+    <SafeAreaView
+      // Force inset as this screen uses auto focus and KeyboardSpacer padding is initially
+      // incorrect because of that
+      forceInset={{ top: 'never', bottom: 'always' }}
```
jmrossy: Nit: pull this object out so we don't remake it each time? Probably doesn't matter but its what we do with other styles

packages/mobile/src/fiatExchanges/ListItem.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { StyleSheet, View } from 'react-native'
+
+interface Props {
+  children: React.ReactNode
+  onPress: () => void
+  disabled?: boolean
+  testID?: string
+}
+
+export default function ListItem({ children, onPress, disabled, testID }: Props) {
```
jmrossy, line 14: Is this a generic component of just for exchanges?

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -355,6 +359,64 @@ const backupScreens = (Navigator: typeof Stack) => (
   </>
 )
 
+const fiatExchangesAmountScreenOptions = ({
```
jmrossy: Given how giant this file is becoming, I want to re-suggest moving these out of here

packages/mobile/src/stableToken/selectors.ts
```diff
@@ -0,0 +1,2 @@
+import { RootState } from 'src/redux/reducers'
+export const dollarBalanceSelector = (state: RootState) => state.stableToken.balance
```
jmrossy: Think this is a dupe, there's already one in stableToken/reducer

jmrossy (commented):
Mostly looks good to me, just some small comments

packages/mobile/locales/en-US/fiatExchangeFlow.json
i1skn: My fault, will do camelCase! Good catch!

packages/mobile/src/fiatExchanges/FiatExchangeAmount.tsx
i1skn: Yeah, quite often I think the same way and go back to https://github.com/getify/You-Dont-Know-JS/blob/1st-ed/scope%20%26%20closures/ch4.md to remind myself 🙂

packages/mobile/src/fiatExchanges/ListItem.tsx
i1skn: So far, it's only used in the Fiat Exchange screens, but I think we could find similar places later on and create a generic `ListItem` for different use cases (like with checkmark, with value in the right, etc.).

packages/mobile/src/config.ts
i1skn: Good idea!

i1skn:
@jmrossy PTAL, I've addressed all the comments

packages/mobile/src/navigator/Navigator.tsx
cmcewen: Yeah I think putting this in the screen component file, or a separate file entirely is a good idea

jmrossy (approved):
Looks like you need to update a snapshot but otherwise LGTM.
RE the growing pile in Navigator, it's np for me if we want to make a separate PR to sort that out.

PR #4034, opened 2020-06-06
[Wallet + PGPNP] Add matchmaking experience to wallet (Part 2)

jmrossy:
### Description

- Implement matchmaking in wallet saga
- Finish implementing import contacts screen, incl. loading states
- Add cancel + timeout ability to importing contacts 
- Rename VerificationSuccessScreen to OnboardingSuccessScreen
- Delete obsolete ContactSyncBanner component
- Add matchmaking saga test

### Other changes

- Organize analytics event names a bit
- Add types to PGPNP service requests

### Tested

Did full onboarding flow, incl. matchmaking, using PGPNP staging.

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/474
- Fixes https://github.com/celo-org/celo-labs/issues/475

### Backwards compatibility

Yes

Screen Shot 2020-06-04 at 18 17 43
Screen Shot 2020-06-04 at 18 17 50
Screen Shot 2020-06-04 at 18 19 18


packages/mobile/src/identity/contactMapping.ts
```diff
@@ -39,13 +43,37 @@ import { getConnectedAccount } from 'src/web3/saga'
 import { currentAccountSelector } from 'src/web3/selectors'
 
 const TAG = 'identity/contactMapping'
+export const IMPORT_CONTACTS_TIMEOUT = 2 * 60 * 1000 // 2 minutes
```
tarikbellamine: What's the thought on this length? Seems a bit on the long side to me.

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -39,13 +43,37 @@ import { getConnectedAccount } from 'src/web3/saga'
...
+}
 
-export function* doImportContactsWrapper() {
+export function* doImportContactsWrapper({ doMatchmaking }: ImportContactsAction) {
   yield call(getConnectedAccount)
   try {
     Logger.debug(TAG, 'Importing user contacts')
 
-    yield call(doImportContacts)
+    const { result, cancel, timeout } = yield race({
```
tarikbellamine, line 54: Nice implementation on this

packages/mobile/src/identity/reducer.ts
```diff
@@ -73,11 +78,12 @@ const initialState: State = {
   e164NumberToAddress: {},
   e164NumberToSalt: {},
   askedContactsPermission: false,
-  isLoadingImportContacts: false,
-  contactMappingProgress: {
+  importContactsProgress: {
+    status: 0,
```
tarikbellamine: Can you use the enum property name instead of the value?

packages/mobile/src/identity/reducer.ts
```diff
@@ -149,39 +155,38 @@ export const reducer = (
...
       }
     case Actions.DENY_IMPORT_CONTACTS:
       return {
         ...state,
         askedContactsPermission: true,
       }
+    case Actions.ADD_CONTACT_MATCHES:
+      return {
+        ...state,
+        matchedContacts: [...state.matchedContacts, ...action.matches],
```
tarikbellamine: If we search for matches across the full contact list every time, should new matches overwrite previous matches instead of adding to them? Thinking about a scenario where a user deletes contacts from their address book yet they are still showing up as a matched contact here.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -52,29 +95,83 @@ class ImportContactScreen extends React.Component<Props, State> {
     })
   }
 
+  onFinish = () => {
+    navigate(Screens.OnboardingSuccessScreen)
+  }
+
+  renderImportStatus = () => {
+    const { t, importContactsProgress, matchedContacts } = this.props
+    const { status, total } = importContactsProgress
```
tarikbellamine: Nit: You can probably go for the double destructure of `status` and `total` here if you arent using `importContactsProgress` anywhere else

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -52,29 +95,83 @@ class ImportContactScreen extends React.Component<Props, State> {
...
+        <Text style={styles.h2Status}>{t(h2Text, { total })}</Text>
+      </View>
+    )
+  }
+
   render() {
     const { isFindMeSwitchChecked } = this.state
-    const { t } = this.props
+    const { t, importContactsProgress } = this.props
+    const { status } = importContactsProgress
```
tarikbellamine: Nit: ditto on the double destructure

tarikbellamine (commented):
Good stuff! I'm excited for this feature 🥇

packages/mobile/src/identity/contactMapping.ts
jmrossy: Yeah its def a bit on the long side. The thought was to be cautious until we have better estimates of how long matchmaking will take. It'll hopefully be < 10 seconds. 

I'll drop it down to 1 minute and see how that goes.

packages/mobile/src/identity/contactMapping.ts
jmrossy: Thanks! Yeah this is a handy Saga pattern we use in a few places, like verification too

packages/mobile/src/identity/reducer.ts
jmrossy: Great question. I tried that but it leads to uninitialized values because the enum isn't defined in this file.

jmrossy: The root underlying problem is dependency cycles

packages/mobile/src/identity/reducer.ts
jmrossy: Yeah good question. We currently only allow matchmaking to be done once, so it's essentially equivalent if we went with a `SET` vs an `ADD` here. 
We want to eventually allow users to do a subsequent matchmaking session on just their new contacts, in which case an `ADD` approach will be better. So I'm inclined to leave this as it for now. WDYT?

packages/mobile/src/identity/matchmaking.ts
```diff
@@ -0,0 +1,137 @@
+import { ContractKit } from '@celo/contractkit'
```
cmcewen, line 1: I think a comment in this file, or maybe in the README just explaining the whole architecture here would be pretty useful. Or maybe there's a design doc? I just haven't been playing super close attention and feel like there is a lot going on here, but overall flow isn't super clear

packages/mobile/src/identity/matchmaking.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+  return hashes
+}
+
+// Hashes the phone number using a static salt
+// This is different than the phone + unique salt hashing that
+// we use for numbers getting verified and going on chain
+// Matchmaking doesn't support per-number salts yet
+export function obfuscateNumberForMatchmaking(e164Number: string) {
+  return crypto
```
cmcewen, line 70: maybe memoize this function? also react-native-fast-crypto?

packages/mobile/src/identity/pgpnp.ts
```diff
@@ -0,0 +1,50 @@
+// Utilities for interacting with the PGPNP service
+// PGPNP == Pretty Good Phone Number Privacy
```
cmcewen: https://celo-org.slack.com/archives/C012E7YGP34/p1591117434383000

packages/mobile/src/identity/pgpnp.ts
```diff
@@ -0,0 +1,50 @@
...
+  account: string,
+  contractKit: ContractKit,
+  body: object,
+  endpoint: string
+) {
+  Logger.debug(`${TAG}@postToPGPNP`, `Posting to ${endpoint}`)
+
+  // Sign payload using account privkey
+  const bodyString = JSON.stringify(body)
+  const authHeader = await contractKit.web3.eth.sign(bodyString, account)
```
cmcewen: @yorhodes will this require any work with your PR?

packages/mobile/src/identity/privacy.test.ts
```diff
@@ -6,7 +6,6 @@ import { updateE164PhoneNumberSalts } from 'src/identity/actions'
 import { fetchPhoneHashPrivate, getSaltFromThresholdSignature } from 'src/identity/privacy'
 import { e164NumberToSaltSelector } from 'src/identity/reducer'
 import { getConnectedUnlockedAccount } from 'src/web3/saga'
-import { currentAccountSelector } from 'src/web3/selectors'
 import { mockAccount, mockE164Number, mockE164Number2 } from 'test/values'
 
 jest.mock('react-native-blind-threshold-bls', () => ({
```
cmcewen, line 11: is it worth moving this to the mocks folder?

packages/mobile/src/identity/reducer.ts
cmcewen: maybe create a new file just for the enum then?

packages/phone-number-privacy/combiner/src/match-making/get-contact-matches.ts
```diff
@@ -4,15 +4,29 @@ import { authenticateUser, isVerified } from '../common/identity'
...
 import { getNumberPairContacts, setNumberPairContacts } from '../database/wrappers/number-pairs'
 
+interface GetContactMatchesRequest {
+  account: string
+  userPhoneNumber: string
+  contactPhoneNumbers: string[]
+  hashedPhoneNumber: string
+}
+
+interface ContactMatch {
```
cmcewen, line 21: just noting that you have a `ContactMatch` interface in the mobile code with different fields - confused me for a sec when searching

packages/mobile/src/identity/reducer.ts
cmcewen: just wondering if using an object and indexing by phone number of contact id might be useful in terms of deduplicating and also quick lookups?

packages/phone-number-privacy/combiner/src/common/input-validation.ts
```diff
@@ -18,7 +18,11 @@ export function isBodyReasonablySized(requestBody: any): boolean {
 }
 
 export function hasValidQueryPhoneNumberParam(requestBody: any): boolean {
-  return requestBody.blindedQueryPhoneNumber
+  return !!requestBody.blindedQueryPhoneNumber
+}
+
+export function hasValidPhoneNumberHash(requestBody: any): boolean {
```
cmcewen, line 24: lots of anys here - possible to add some better typings in this file?

packages/mobile/src/identity/pgpnp.ts
yorhodes: the provider will intercept this already

packages/mobile/src/identity/matchmaking.ts
jmrossy: https://docs.google.com/document/d/1SvQ7qD_ms2S7OFxWJ3wy5x3vvYeJFtMx2yU5TevEAic/edit#

packages/mobile/src/identity/matchmaking.ts
jmrossy: Yeah there's a TODO in here about moving to RN-fast-crypto for this. I'd just like to get it working e2e first

packages/mobile/src/identity/pgpnp.ts
jmrossy: Okay I'll rename some things

packages/mobile/src/identity/matchmaking.ts
```diff
@@ -0,0 +1,137 @@
...
+    const hash = obfuscateNumberForMatchmaking(e164Number)
+    hashes[hash] = e164Number
+  }
+  return hashes
+}
+
+// Hashes the phone number using a static salt
+// This is different than the phone + unique salt hashing that
+// we use for numbers getting verified and going on chain
+// Matchmaking doesn't support per-number salts yet
```
jeanregisser, line 68: I assume this suggests it will use per-number salts eventually?

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -35,15 +61,32 @@ class ImportContactScreen extends React.Component<Props, State> {
...
-  onPressConnect = () => {
-    // TODO do matchmaking if isFindMeSwitchChecked
-    this.props.importContacts()
+  componentDidUpdate(prevProps: Props) {
+    if (
+      prevProps.importContactsProgress.status !== ImportContactsStatus.Done &&
+      this.props.importContactsProgress.status === ImportContactsStatus.Done
+    ) {
+      // Set timeout to leave checkmark in place for a little while
+      setTimeout(this.onFinish, 1500)
```
jeanregisser: timer should be cleared on unmount.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,6 +1,7 @@
 import Button, { BtnSizes, BtnTypes } from '@celo/react-components/components/Button.v2'
 import Switch from '@celo/react-components/components/Switch.v2'
 import TextButton from '@celo/react-components/components/TextButton.v2'
+import Checkmark from '@celo/react-components/icons/Checkmark'
 import colors from '@celo/react-components/styles/colors'
 import colorsV2 from '@celo/react-components/styles/colors.v2'
```
jeanregisser: nit: not related to this specific PR, but we should remove old `colors`

packages/mobile/src/onboarding/success/OnboardingSuccessScreen.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { navigateHome } from 'src/navigator/NavigationService'
+
+function OnboardingSuccessScreen() {
+  const onAnimationFinish = useCallback(() => {
+    navigateHome()
+  }, [])
+
+  return (
+    <View style={styles.container}>
+      <DancingRings onAnimationFinish={onAnimationFinish} />
```
jeanregisser: nit: `navigateHome` could be passed directly.

jeanregisser (approved):
Looks great! 👍 🎉 
Just some very small comments.

packages/mobile/src/identity/reducer.ts
jmrossy: The num of matches will likely be very small but okay will do

packages/mobile/src/identity/reducer.ts
jmrossy: Okay I'll give that a try

packages/mobile/src/identity/matchmaking.ts
jmrossy: As for memoizing, I don't think we can cause the state of the created hash gets altered with digest. This is just how the node crypto lib works

packages/phone-number-privacy/combiner/src/common/input-validation.ts
jmrossy: Mostly not my code but okay I'll try to improve it

packages/phone-number-privacy/combiner/src/match-making/get-contact-matches.ts
jmrossy: Renamed the mobile one

packages/phone-number-privacy/combiner/src/common/input-validation.ts
jmrossy: These are currently shared across a few diff request types. We'd need to get creative with unions and generics to type them right. I'll get to it but rather not do it in this PR.

packages/mobile/src/identity/privacy.test.ts
jmrossy: It's only used in this file

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
jmrossy: Good catch, I'll remove the dep from this new file

packages/mobile/src/onboarding/success/OnboardingSuccessScreen.tsx
jmrossy: Good idea. We used to do other things in that callback.

packages/mobile/src/identity/matchmaking.ts
jmrossy: Or some other crypto magic that the crypto bubble is currently working on :)

packages/mobile/src/identity/matchmaking.ts
cmcewen: I was just thinking that we could cache this per e164Number in an object or something

jmrossy: Why cache it? The likelyhood of the same e164Number showing up in a contact list twice is very low and we only do matchmaking once ever.

packages/mobile/src/identity/reducer.ts
```diff
@@ -149,39 +148,38 @@ export const reducer = (
...
-    case Actions.INCREMENT_IMPORT_SYNC_PROGRESS:
+    case Actions.UPDATE_IMPORT_CONTACT_PROGRESS:
+      const curProgress = state.importContactsProgress
       return {
         ...state,
-        contactMappingProgress: {
-          current: state.contactMappingProgress.current + action.increment,
-          total: state.contactMappingProgress.total,
+        importContactsProgress: {
+          current: action.current ?? curProgress.current,
```
aslawson, line 159: what it this `??` syntex?

packages/mobile/src/identity/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  Prepping = 1,
+  GettingStatus = 2,
+  RequestingAttestations = 3,
+  RevealingNumber = 4,
+  RevealAttemptFailed = 5,
+  Done = 6,
+}
+
+export enum ImportContactsStatus {
+  Failed = -1,
```
aslawson, line 19: negative enum for failure cases, that's interesting

packages/phone-number-privacy/combiner/src/match-making/get-contact-matches.ts
```diff
@@ -4,15 +4,29 @@ import { authenticateUser, isVerified } from '../common/identity'
...
   hasValidContractPhoneNumbersParam,
+  hasValidPhoneNumberHash,
   hasValidUserPhoneNumberParam,
-  isBodyReasonablySized,
 } from '../common/input-validation'
 import logger from '../common/logger'
 import { getDidMatchmaking, setDidMatchmaking } from '../database/wrappers/account'
 import { getNumberPairContacts, setNumberPairContacts } from '../database/wrappers/number-pairs'
 
+interface GetContactMatchesRequest {
```
aslawson, line 14: nice add of the request/response object definitions

packages/phone-number-privacy/combiner/src/match-making/get-contact-matches.ts
```diff
@@ -22,35 +36,36 @@ export async function handleGetContactMatches(request: Request, response: Respon
       respondWithError(response, 401, ErrorMessages.UNAUTHENTICATED_USER)
       return
     }
-    if (!(await isVerified(request.body.account, request.body.userPhoneNumber))) {
+
+    const { account, userPhoneNumber, contactPhoneNumbers, hashedPhoneNumber } = request.body
```
aslawson, line 40: 🙌 so much better

packages/phone-number-privacy/combiner/src/match-making/get-contact-matches.ts
```diff
@@ -22,35 +36,36 @@ export async function handleGetContactMatches(request: Request, response: Respon
...
-function isValidGetContactMatchesInput(requestBody: any): boolean {
+function isValidGetContactMatchesInput(requestBody: GetContactMatchesRequest): boolean {
   return (
     hasValidAccountParam(requestBody) &&
     hasValidUserPhoneNumberParam(requestBody) &&
     hasValidContractPhoneNumbersParam(requestBody) &&
-    isBodyReasonablySized(requestBody)
+    hasValidPhoneNumberHash(requestBody)
+    // TODO find way to check content body size without RE-JSONifying it
+    // isBodyReasonablySized(requestBody)
```
aslawson, line 69: why commented out?

aslawson (commented):
lgtm -- i'll leave the shipping to the mobile-familiar devs

aslawson:
does the wallet automatically transition to the contact list  after the "you have 2 friends on Celo" message?  it feels off to have the "skip" button there since there really isn't anything left to skip

packages/phone-number-privacy/combiner/src/match-making/get-contact-matches.ts
jmrossy: A normal size address book is more than the current size limit allows. I'll follow-up in another PR with a fix for this TODO

packages/mobile/src/identity/types.ts
jmrossy: Yeah it can be helpful for things like `if (status < 0)` when there are multiple failure cases

packages/mobile/src/identity/reducer.ts
jmrossy: Use action.current unless its null or undefined, in which case use curProgress.current.
It's similar to `||` except it doesn't coalesce types. E.g. it won't treat 0 as falsy

jmrossy:
> does the wallet automatically transition to the contact list after the "you have 2 friends on Celo" message? it feels off to have the "skip" button there since there really isn't anything left to skip

It does auto transition in about 1 second yes. 

Yep great point. The UI work on that screen isn't done and still needs some polish. It's pending finalization of designs.

PR #4036, opened 2020-06-08
Mnemonic support fix

codyborn:
### Description

- Update baklava oracle variables
- Set default for all Azure variables

### Tested

Deployed using celotool


tkporter (approved):
Thank you 🙏

PR #4040, opened 2020-06-08
[PGPNP] Update combiner config to work with local signer express app, remove firebase stuff for signer

aslawson:
### Description

- Update combiner config to work with local signer express app
- remove firebase stuff for signer (it is no longer a firebase function)

### Other changes

NA

### Tested

locally ran signer at port 3000

### Related issues

NA

### Backwards compatibility

NA - dev endpoints


jmrossy (approved):
Ship it!

PR #4043, opened 2020-06-09
Oracle optional in memory signing infra, baklava deployment

tkporter:
### Description

* Allows infra to use in memory signing (ie use a private key generated from a mnemonic)
* Uses `ORACLE_PRIMARY_ADDRESSES_FROM_MNEMONIC_COUNT` or `ORACLE_SECONDARY_ADDRESSES_FROM_MNEMONIC_COUNT` to determine how many oracles to create from the mnemonic
* Includes some changes to the env variables in the oracle client

### Other changes

n/a

### Tested

Deployed baklava using in memory signing

### Related issues

n/a

### Backwards compatibility

yes, with the exception of some env vars are only valid for newer versions of the oracle client


packages/helm-charts/oracle/templates/statefulset.yaml
```diff
@@ -81,8 +81,13 @@ spec:
...
 
-            # Get the correct key vault name
-            AZURE_KEY_VAULT_NAMES={{- range $index, $identity := .Values.oracle.identities -}}{{ $identity.azure.keyVaultName }},{{- end }}
+            # Set the private key path. If Azure HSM signing is specified,
+            # it will take precedence.
+            export PRIVATE_KEY_PATH="/private-keys/private-key-$RID"
+
+            # Get the correct key vault name. If this oracle's identity is not
+            # using Azure HSM signing, the key vault name will be empty and ignored
+            AZURE_KEY_VAULT_NAMES={{- range $index, $identity := .Values.oracle.identities -}}{{- if (hasKey $identity "azure" ) -}}{{ $identity.azure.keyVaultName | default "" }}{{- end }},{{- end }}
```
codyborn, line 90: Nice!

packages/celotool/src/lib/oracle.ts
```diff
@@ -66,14 +72,19 @@ const oracleAzureContextClusterConfigEnvVars: {
...
 } = {
   [OracleAzureContext.PRIMARY]: {
-    identities: envVar.ORACLE_PRIMARY_ADDRESS_AZURE_KEY_VAULTS,
+    addressAzureKeyVaults: envVar.ORACLE_PRIMARY_ADDRESS_AZURE_KEY_VAULTS,
+    addressesFromMnemonicCount: envVar.ORACLE_PRIMARY_ADDRESSES_FROM_MNEMONIC_COUNT,
   },
   [OracleAzureContext.SECONDARY]: {
-    identities: envVar.ORACLE_SECONDARY_ADDRESS_AZURE_KEY_VAULTS,
+    addressAzureKeyVaults: envVar.ORACLE_SECONDARY_ADDRESS_AZURE_KEY_VAULTS,
+    addressesFromMnemonicCount: envVar.ORACLE_SECONDARY_ADDRESSES_FROM_MNEMONIC_COUNT,
```
codyborn, line 87: Something we can add later, but it'd be cool to be able to specify an offset for the second mnemonic addresses. Ex. if we have 5 in primary, then we skip the first 5 addresses to prevent overlap when deriving the second 5.

packages/helm-charts/oracle/templates/azure-identity-binding.yaml
```diff
@@ -1,4 +1,5 @@
 {{- range $index, $identity := .Values.oracle.identities -}}
+{{ if (hasKey $identity "azure") }}
```
codyborn, line 2: It feels like these should be indented, if not just for legibility

packages/helm-charts/oracle/values.yaml
```diff
@@ -17,14 +17,18 @@ kubectl:
 
 oracle:
   replicas: 2
-  rpcProviderUrl: https://alfajoresstaging-forno.celo-testnet.org
+  rpcProviderUrls:
```
codyborn, line 20: This will be moved to the .env file, correct?

codyborn (approved):
Looks good, thanks Trevor!

packages/celotool/src/lib/oracle.ts
tkporter: Totally agree!

packages/helm-charts/oracle/templates/azure-identity-binding.yaml
tkporter: I definitely agree it's ugly, but as far as I know helm can complain if there are indentations because yaml relies on whitespace

packages/helm-charts/oracle/values.yaml
tkporter: They are but indirectly- they're overwritten around here https://github.com/celo-org/celo-monorepo/pull/4043/files#diff-78d16d829b51505638952f432154adcaR156 using info from the .env file (like the environment name)

PR #4044, opened 2020-06-09
Baklava env vars

codyborn:
### Description

- Add HSM identity variables
- Docker image points to current master build


tkporter (approved):
Beautiful! Do you think we should merge now or wait till the gov proposal?

Issue #3957, opened 2020-06-03
Complete and publish the smart contract release process


nategraf:
Complete the writing of the WIP smart contract release process and publish it for public reference. Additionally, create any issues needed to track work needed to enact the release process.

PR #4045, opened 2020-06-09
Publish proposed release process documentation

nategraf:
### Description

This PR contains the first version of the release process that we will be implementing for the smart contracts and blockchain client.

### Other changes

None

### Related Issues

* https://github.com/celo-org/celo-monorepo/issues/3957
* https://github.com/celo-org/celo-blockchain/issues/1052


asaj (approved):
Should we flag somewhere that implementation of these processes is still WIP?

hbandura (approved):

nategraf:
@asaj Yes, that's a good callout. I'll add a warning to the top of the document.

packages/docs/community/release-process/blockchain-client.md
```diff
@@ -0,0 +1,179 @@
...
+  </tr>
+  <tr>
+    <td>T+1w</td>
+    <td>
+      <ol>
+        <li>Confirm Baklava users have upgraded without issues</li>
+        <li>If release introduces a hard fork</li>
+        <ol>
+          <li>Ensure at least a quorum of the validator set has upgraded</li>
+          <li>Submit governance proposal to increment minimum client version</li>
```
bowd, line 122: I think we might need to add some additional clarification here. I'm still a bit murky on what it means to introduce a hard fork. I might have some faulty assumptions so let me do a little brain dump.

As far as I understand it there's a `minimum client version` stored on chain that, I imagine, acts as a circuit breaker and makes the node go "I can't play with others so I'm shutting down". But in order for the network to fork you'd also need another off-chain `minimum client version` that tells the client what peers she can play with, thus an update that increments this off-chain constraint will cause the nodes that update to play among themselves. 
But at this point we have two chains so the governance proposal seams useless because it essentially happens on the chain that already is constrained by the off-chain version. And the old chain will not necessarily get the proposal and continue to function independently, albeit without any use, but without the nodes affected by the on-chain `minimum client version`. 

So I feel like I'm miss-understanding something and have some faulty assumptions, either way I think it would be valuable to expand on these scenarios better here. Explain how a release that introduces a hard fork happens vs one that doesn't.

I guess this also started from me trying to understand why incrementing happened _only_ in hard fork releases.

bowd (changes requested):
I'd like to take a stab at clarifying that section regarding updating and hard forks once I understand it better.

packages/docs/community/release-process/blockchain-client.md
```diff
@@ -0,0 +1,179 @@
...
+Artifacts produced by this build process (e.g. tags, binaries, Docker images) will be signed. Signatures are produced using any one of the core developer keys listed below.
+
+{% hint style="warning" %}
+Work in progress: Core developer keys are not yet hosted.
+{% endhint %}
+
+Public keys for these developers are hosted on celo.org and can be imported to `gpg` with the following command:
+
+```bash
+gpg --auto-key-locate wkd --locate-keys $EMAIL
```
nambrot, line 29: What do you think about using the same signature scheme on the EVM vs. gpg? I know gpg of course is the more standardized toolchain for virtually all other software, but there is something nice about having one consistent scheme in the Celo ecosystem?

nategraf: A) `gpg` supports ECC signatures, although I am not sure if it supports `secp256k1`.
B) Unless we want to built a lot of tooling ourselves, it's really best to use something standardized, and nothing else is as standardized as `gpg`. Even if we did end up building our own tooling, you'd end up with a chicken and egg problem where someone needs to verify they got our tool to verify the binary.

packages/docs/community/release-process/blockchain-client.md
bowd: Was a miss-understanding on my part which was resolved in call with @nategraf.
I'm still unsure wether we should include more information here about this or leave it as is.

bowd (approved):

packages/docs/community/release-process/blockchain-client.md
nategraf: I am personally in favor of leaving it at the current level of specificity because I think we will understand better how to handle handforks once we are used to doing non-fork upgrades. Additionally, there may be multiple processes that would be used at any time, and this PR is designed to get use to an initial implementation, and does not have to be complete or perfect.

packages/docs/community/release-process/blockchain-client.md
nategraf: Update: 
* `gpg` supports `secp256k1`. https://wiki.gnupg.org/ECC
* Yubikey 5.2.3+ supports `secp256k1` as well! https://support.yubico.com/support/solutions/articles/15000027139-yubikey-5-2-3-enhancements-to-openpgp-3-4-support

With that, I will recommend in the doc I am writing now that developer public keys be made with that curve, which will allow them to be used as on-chain addresses in the future.

PR #4049, opened 2020-06-09
fix: set default block time to 5 seconds

AlexBHarley:
### Description

The tx nodes in the testnet fail to sync if the block time is 1 second due to `########## BAD BLOCK #########` errors.


tkporter:
Do you know why making it 1 second causes bad block errors?

nambrot (approved):

nambrot:
I believe overwriting it to 1 on validators causes them to produce 1s blocks, but tx nodes will believe that the protocol was violated

PR #4053, opened 2020-06-09
[Wallet] Surface matched contacts to top of Send recipient list

jmrossy:
### Description
- Surface matched contacts to top of Send recipient list
- Improve performance of recipient searching
- Cleanup Send.tsx a bit
- Remove some redundant and PII leaking analytics
- Improve error handling on ImportContactsScreen

### Tested

Tested on Search screen and wrote some new unit tests

### Related issues

- https://github.com/celo-org/celo-labs/issues/475

### Backwards compatibility

Yes


packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -243,6 +244,9 @@ const styles = StyleSheet.create({
   statusContainer: {
     alignItems: 'center',
   },
+  invisible: {
```
jmrossy, line 247: A convenient way to hide without having the layout contents jump :)

packages/mobile/src/alert/actions.ts
```diff
@@ -60,6 +61,16 @@ export const showErrorInline = (error: ErrorMessages, i18nOptions?: TOptions): S
   underlyingError: error,
 })
 
+// Useful for showing a more specific error if its a documented one, with
+// a fallback to something more generic
+export function* showErrorMessageOrFallback(error: any, fallback: ErrorMessages) {
```
tarikbellamine: Nice, we use this pattern pretty often! Does this need to be a generator?

jmrossy: It does so it can yield put to `showError`.
An alternative would be have it just return the ErrorMessage, which we can yield put, but seemed nice to have it all together :)

tarikbellamine: Oh I see I assumed the latter.

IMO it's a little non-intuitive that we would `yield put(showError)` and `yield call(showErrorMessageOrFallback)` but that's more of a nit so up to you :)

packages/mobile/src/recipients/recipient.ts
```diff
@@ -204,30 +204,55 @@ const fuzzysortToRecipients = (
...
   if (nameA > nameB) {
-    comparison = 1
+    return 1
   } else if (nameA < nameB) {
-    comparison = -1
+    return -1
   }
-  return comparison
+  return 0
 }
```
tarikbellamine, line 217: I think this can all be simplified to `return a.displayName.toUpperCase() > b.displayName.toUpperCase()`

packages/mobile/src/send/Send.tsx
```diff
@@ -174,46 +176,26 @@ class Send extends React.Component<Props, State> {
...
+    this.props.storeLatestInRecents(recipient)
 
     CeloAnalytics.track(
-      this.props.route.params?.isRequest
+      isRequest
         ? CustomEventNames.request_select_recipient
         : CustomEventNames.send_select_recipient,
       {
-        query: this.state.searchQuery,
+        recipientKind: recipient.kind,
```
tarikbellamine, line 198: I think we should also collect the `query`. It will be informative to know if people are selecting recipients via scrolling vs. searching

packages/mobile/src/recipients/recipient.ts
jmrossy: That doesn't handle the equality case right? Also, the Array sort function is expecting a number return value

packages/mobile/src/send/Send.tsx
jmrossy: The query can have names in it, which is PII

packages/mobile/src/alert/actions.ts
jmrossy: Cool I'll change it then, it'd be good for it to match people's first intuition :)

packages/mobile/src/send/Send.tsx
tarikbellamine: Good point. How about we just pass a boolean depending on if `query.length > 0`? What specifically is entered is less important IMO anyways

packages/mobile/src/recipients/recipient.ts
tarikbellamine: You are right. Looks like I've been playing fast and loose with returning boolean!

tarikbellamine (approved):
Nice work!

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -178,8 +178,9 @@ class ImportContactScreen extends React.Component<Props, State> {
           )}
           {status > ImportContactsStatus.Stopped && this.renderImportStatus()}
         </ScrollView>
-        <View style={styles.bottomButtonContainer}>
+        <View style={[styles.bottomButtonContainer, isSkipHidden ? styles.invisible : undefined]}>
```
i1skn: I see, that `StyleProp` accepts `type Falsy = undefined | null | false;`, so could we use this one instead?
```suggestion
        <View style={[styles.bottomButtonContainer, isSkipHidden && styles.invisible]}>
```

i1skn (approved):
Overall great work! See my very little nit 🙂

packages/mobile/src/alert/actions.ts
i1skn: `yield put(showErrorMessageOrFallback)` would be more intuitive for me as well

packages/mobile/src/send/Send.tsx
jmrossy: Will do

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
jmrossy: Good idea!

PR #4055, opened 2020-06-09
[Wallet] Remove lots of empty string handling

cmcewen:
### Description

While debugging some issues today, I found that our types encourage the belief that certain fields are always set, when in fact they might not be yet. This tries to remove as many of those cases as possible.


packages/mobile/src/account/EditProfile.tsx
```diff
@@ -31,11 +31,15 @@ const mapStateToProps = (state: RootState): StateProps => {
...
+  name: string
+}
+
+export class EditProfile extends React.Component<Props, State> {
   static navigationOptions = headerWithBackButton
 
-  state = {
-    name: this.props.name,
+  state: State = {
+    name: this.props.name || '',
```
aaronmgdr, line 42: I thought you were getting rid of empty strings?

cmcewen: For text editing, it does make sense to have an empty string. This will only translate to local component state

jmrossy: IMO this the downside of these changes. A field like name will never be empty after the first onboarding screen, but now we need to handle that case everywhere its used in the app.

jmrossy (approved):
I'm torn on this. For the moment, it seems to introduce more mandatory boilerplate for fields that can never be null. But I also want to see the name + number field be made optional. So I guess I'm overall in favor of this :)

cmcewen:
@jmrossy yeah i had the same feeling about things being optional in the future. i'm okay with boilerplate for clarity's sake 😄

Issue #3988, opened 2020-06-04
Address existing crashes/errors


i1skn:

tarikbellamine:
Epic over

PR #4057, opened 2020-06-10
[Wallet] Balance check on VerificationEducationScreen and PhoneNumberLookupQuotaScreen

### LINKED ISSUE #2990: [Wallet] Users SBAT know that verification failed due to insufficient funds
### Expected Behavior

Users are alerted if they do not have a sufficient balance to verify 

### Current Behavior

General verification failed message when users don't have enough funds to verify. This can happen when someone tries to verify, pays for a transaction but does not receive all verification texts, but then no longer has enough to send another verification transaction

tarikbellamine:
### Description
Adding a check of user's balance on PhoneNumberLookupQuotaScreen and VerificationEducationScreen to prevent the user from initiating a transaction they don't have enough funds to carry out. This should prevent a frequently occurring Sentry error.

### Other changes
Added some error catching/propagation for when gas estimations failed to avoid it hitting the error boundary and creating Sentry issues.

### Tested
No new tests needed

### Related issues

- Fixes #2990 and part of #3988

### Backwards compatibility
Yes


packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -11,22 +11,38 @@ import { ScrollView, StyleSheet, Text, View } from 'react-native'
...
 import { setHasSeenVerificationNux } from 'src/identity/actions'
+import { isUserBalanceSufficient } from 'src/identity/PhoneNumberLookupQuotaScreen'
+import { INVITE_FEE } from 'src/invite/saga'
 import { nuxNavigationOptions } from 'src/navigator/Headers'
 import { navigate, navigateHome } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
+import { RootState } from 'src/redux/reducers'
+
+// Open to better suggestions for estimates
+const VERIFICATION_FEE_ESTIMATE = Number(INVITE_FEE) * 0.9
```
tarikbellamine, line 25: Open to better suggestions for how to estimate verification fee

packages/mobile/src/identity/PhoneNumberLookupQuotaScreen.tsx
```diff
@@ -5,81 +5,95 @@ import SearchUser from '@celo/react-components/icons/SearchUser'
...
+import { LOOKUP_PURCHASE_FEE } from 'src/identity/privateHashing'
 import { Screens } from 'src/navigator/Screens'
 import { StackParamList } from 'src/navigator/types'
+import { stableTokenBalanceSelector } from 'src/stableToken/reducer'
 
-type Props = WithTranslation & StackScreenProps<StackParamList, Screens.PhoneNumberLookupQuota>
+type Props = StackScreenProps<StackParamList, Screens.PhoneNumberLookupQuota>
 
-class PhoneNumberLookupQuotaScreen extends React.Component<Props> {
-  static navigationOptions = { gestureEnabled: false, header: null }
```
jmrossy, line 22: Is the `gestureEnabled: false` being lost in your conversion to hooks here?

packages/mobile/src/identity/PhoneNumberLookupQuotaScreen.tsx
```diff
@@ -5,81 +5,95 @@ import SearchUser from '@celo/react-components/icons/SearchUser'
...
-          />
-        </View>
-        <KeyboardSpacer />
-      </SafeAreaView>
-    )
-  }
+  useEffect(() => {
+    BackHandler.addEventListener('hardwareBackPress', onSkip)
+
+    return function cleanup() {
```
jmrossy: Nit, use lambdas for inline defined functions like this

packages/mobile/src/identity/PhoneNumberLookupQuotaScreen.tsx
```diff
@@ -5,81 +5,95 @@ import SearchUser from '@celo/react-components/icons/SearchUser'
...
 import { StackParamList } from 'src/navigator/types'
+import { stableTokenBalanceSelector } from 'src/stableToken/reducer'
 
-type Props = WithTranslation & StackScreenProps<StackParamList, Screens.PhoneNumberLookupQuota>
+type Props = StackScreenProps<StackParamList, Screens.PhoneNumberLookupQuota>
 
-class PhoneNumberLookupQuotaScreen extends React.Component<Props> {
-  static navigationOptions = { gestureEnabled: false, header: null }
+export function isUserBalanceSufficient(userBalance: string | null, estimatedFee: number) {
+  if (!userBalance || Number(userBalance) < estimatedFee) {
```
jmrossy: always use BigNumber for parsing num strings

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -22,6 +23,7 @@ import { currentAccountSelector } from 'src/web3/selectors'
 const TAG = 'identity/privateHashing'
 const SIGN_MESSAGE_ENDPOINT = '/getBlindedSalt'
 export const SALT_CHAR_LENGTH = 13
+export const LOOKUP_PURCHASE_FEE = 0.01 // one penny
```
jmrossy: This isn't the actual fee because the money is being sent to yourself. the fee is the transaction gas so we need to tap into gas estimates for this.

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -37,19 +39,19 @@ export function* fetchPhoneHashPrivate(e164Number: string) {
   } catch (error) {
     if (error.message === ErrorMessages.SALT_QUOTA_EXCEEDED) {
       Logger.error(
-        `${TAG}@fetchPrivatePhoneHash`,
+        `${TAG}@fetchPhoneHashPrivate`,
```
jmrossy, line 43: Thanks you sir 💯

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -37,19 +39,19 @@ export function* fetchPhoneHashPrivate(e164Number: string) {
...
         'Salt quota exceeded, navigating to quota purchase screen'
       )
-      const quotaPurchaseSucess: boolean = yield call(navigateToQuotaPurchaseScreen)
-      if (quotaPurchaseSucess) {
+      const quotaPurchaseSuccess: boolean = yield call(navigateToQuotaPurchaseScreen)
```
jmrossy, line 46: And thanks again!

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -215,10 +217,16 @@ function* navigateToQuotaPurchaseScreen() {
 
     const ownAddress: string = yield select(currentAccountSelector)
     const txId = generateStandbyTransactionId(ownAddress)
+
+    const userBalance = yield select(stableTokenBalanceSelector)
+    if (!userBalance || Number(userBalance) < 0.01) {
```
jmrossy: ditto BigNumber

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -215,10 +217,16 @@ function* navigateToQuotaPurchaseScreen() {
...
+      throw Error('Insufficient balance to purchase lookups')
```
jmrossy: Should you throw the new `ErrorMessage` type you made instead?

packages/mobile/src/identity/privateHashing.ts
jmrossy: We need to consider gas estimates here as well

packages/mobile/src/verify/VerificationEducationScreen.tsx
jmrossy: I think this is a good improvement and the next level of accuracy requires a 10x more complex approach.

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -11,22 +11,38 @@ import { ScrollView, StyleSheet, Text, View } from 'react-native'
 import Modal from 'react-native-modal'
 import SafeAreaView from 'react-native-safe-area-view'
 import { connect } from 'react-redux'
+import { ErrorMessages } from 'src/app/ErrorMessages'
+import ErrorMessageInline from 'src/components/ErrorMessageInline'
 import { Namespaces, withTranslation } from 'src/i18n'
 import { setHasSeenVerificationNux } from 'src/identity/actions'
+import { isUserBalanceSufficient } from 'src/identity/PhoneNumberLookupQuotaScreen'
```
jmrossy: If you're going to use this util in many places, please find a better home for it :)

jmrossy (commented):
Thanks for taking this on!! You're on the right track here, please see comments below

packages/mobile/src/identity/PhoneNumberLookupQuotaScreen.tsx
tarikbellamine: Good catch. Navigation options are done a little differently now and I needed to circle back to refactor this but it slipped through the cracks

packages/mobile/src/identity/PhoneNumberLookupQuotaScreen.tsx
tarikbellamine: Would that just be `new BigNumber(userBalance) < new BigNumber(estimatedFee)` or is there a specific method I should be using?

packages/mobile/src/identity/privateHashing.ts
tarikbellamine: Right, that was a poor name for it I'll pick something else. Instead of waiting for a gas estimate, would requiring a balance of 0.02 be a decent 80/20 solution?

I worry a little about introducing the complexity of a gas price fetch + the added wait time involved with the check itself.

packages/mobile/src/verify/VerificationEducationScreen.tsx
tarikbellamine: You're right, I'll create a new utils file somewhere

packages/mobile/src/identity/PhoneNumberLookupQuotaScreen.tsx
jmrossy: You can't use native operators with BigNumber. It would be `new BigNumber(userBalance).isLessThan(estimatedFee)`

packages/mobile/src/identity/privateHashing.ts
jmrossy: I'd say lets require 0.03 to be safe but yep I agree

jmrossy (approved):

PR #4064, opened 2020-06-10
[PGPNP] Create dockerfile for PGPNP signer

jmrossy:
### Description

Create a dockerfile and configure cloudbuild to generate docker images for the PGPNP signer.
Update readme with more detailed instructions.

### Other changes

Add a few safe-to-ignore packages to the .dockerignore file

### Tested

Testing here via CI and tested by downloading the docker image and running it.

### Related issues

- #3899

### Backwards compatibility

Yes


packages/phone-number-privacy/signer/package.json
```diff
@@ -30,7 +30,8 @@
     "@types/uuid": "^7.0.3",
     "@types/supertest": "^2.0.9",
     "dotenv": "^8.2.0",
-    "supertest": "^4.0.2"
+    "supertest": "^4.0.2",
+    "typescript": "^3.8.3"
```
jmrossy: Adding typescript to the dev deps was necessary to avoid copying over a significant chunk of the whole monorepo yarn workspace. It's not ideal but IMO worth the trade-off

codyborn (approved):
🐋

dockerfiles/phone-number-privacy/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+COPY packages/phone-number-privacy/signer signer
+COPY packages/phone-number-privacy/common common
+
+WORKDIR /celo-phone-number-privacy/signer
+COPY yarn.lock ./
+
+RUN yarn install --frozen-lockfile --network-timeout 100000 && yarn cache clean
+
+ENV NODE_ENV production
```
aslawson, line 12: where will we add the instructions for what to set other ENV variables here?

packages/phone-number-privacy/signer/src/common/error-utils.ts
```diff
@@ -1,4 +1,4 @@
-import { Response } from 'firebase-functions'
```
aslawson, line 1: do these work similarly enough that it can just be replaced like this.  it's interesting that this worked with the firebase type even though I ran this as the express app

packages/phone-number-privacy/signer/src/common/error-utils.ts
```diff
@@ -9,8 +9,6 @@ export enum ErrorMessages {
   INVALID_INPUT = 'CELO_PNP_ERR_04 Invalid input paramaters',
   EXCEEDED_QUOTA = 'CELO_PNP_ERR_05 Requester exceeded salt service query quota',
   SIGNATURE_COMPUTATION_FAILURE = 'CELO_PNP_ERR_06 Failed to compute BLS signature',
-  DUPLICATE_REQUEST_TO_MATCHMAKE = 'CELO_PNP_ERR_08 Attempt to request >1 matchmaking',
-  UNVERIFIED_USER_ATTEMPT_TO_MATCHMAKE = 'CELO_PNP_ERR_09 Unverified user attempting to matchmake',
```
aslawson: do you no think this should just be consolidated to common folder?  removing #8 and #9 is a little confusing otherwise.  and in the future, we discussed that signer would manage matchmaking

aslawson (approved):
looks good, nice work
🐳

packages/phone-number-privacy/signer/src/common/error-utils.ts
jmrossy: Firebase functions use express under the covers

dockerfiles/phone-number-privacy/Dockerfile
jmrossy: Added to the readme

packages/phone-number-privacy/signer/src/common/error-utils.ts
jmrossy: Yeah that's fair, I'll revert this

aslawson (approved):
changes for config look good.

thanks for updating the README!

PR #4067, opened 2020-06-11
Support OpenPGP WKD serving from @celo/web

nategraf:
### Description

OpenPGP, the most popular implementation being GnuPG or `gpg`, is a commonly used standard for asymmetric key cryptography with support for signing, encryption, and multiple methods of public key distribution. GPG is commonly used to encrypt emails sign Git commits and built binaries.

As mentioned in the release documents introduced by https://github.com/celo-org/celo-monorepo/pull/4045, build signing is a key component of the Celo release process. In order to verify signatures, users must be able to securely obtain public keys for the core developers.

This PR introduces the components needed to implement a [Web Key Directory (WKD)](https://wiki.gnupg.org/WKD), which will allow users to easily and securely retrieve public keys hosted on [celo.org](https://celo.org). 

As an example, once my keys are added to the `packages/web/openpgpkey` directory, anyone will be able to run the following command to download the public key associated with victor@celo.org

```
gpg --auto-key-locate wkd --locate-keys victor@celo.org
```

Note: Our goal is to use the @clabs.co domain, but additional work will be required. This PR is sufficient to start hosting keys associated with @celo.org user names (which do not nessasrily have to be valid email addresses)

### Tested

Deployed to dev.celo.org and verified that `gpg --auto-key-locate wkd --locate-keys victor@dev.celo.org` successfully retrieves the desired key.

### Related issues

- Related https://github.com/celo-org/celo-blockchain/issues/1052


packages/web/server/index.ts
```diff
@@ -121,6 +123,52 @@ function wwwRedirect(req: express.Request, res: express.Response, nextAction: ()
     res.redirect('https://forum.celo.org/t/the-great-celo-stake-off-the-details/136')
   })
 
+  // Redirects for OpenPGP WKD https://gnupg.org/blog/20161027-hosting-a-web-key-directory.html
```
aaronmgdr: Maybe change the comment as these are not redirects?

aaronmgdr (approved):
Cool

packages/web/server/index.ts
nategraf: Good catch!

PR #4070, opened 2020-06-11
Allowing for different distribution ratios in deploy script

lucasege:
### Description

Allows for grant configs to have different `initialDistributionRatio`s within one file.

Leaving as draft since I don't know if we want to merge this into master

### Tested

Manual tests


lucasege:
Done using, but functionality is also already present with use of `yesreally` flag

PR #4072, opened 2020-06-11
Slim down the attestation service docker image

nambrot:
### Description

h/t to @jmrossy for suggesting to slimming down the docker image by using published dependency versions


packages/attestation-service/package.json
```diff
@@ -55,7 +55,6 @@
     "yargs": "13.3.0"
   },
   "devDependencies": {
-    "@celo/protocol": "1.0.0",
```
aslawson, line 58: is protocol just not needed?

aslawson (approved):
looks good

packages/attestation-service/package.json
nambrot: Yeah, not entirely sure how it made its way in here

PR #4076, opened 2020-06-11
[Docs] Fix build issue

gastonponti:
### Description

Docs build was failing due to dependencies issues


yorhodes (approved):
thanks for doing this! I know the work required was >>> what's visible

package.json
```diff
@@ -91,7 +91,9 @@
     "randomatic": "^3.1.1",
     "logkitty": "^0.7.1",
     "minimist": "^1.2.5",
-    "npm": "^6.14.5",
+    "yargs-parser": "^13.1.2",
+    "npm": "^5.10.0",
```
jmrossy, line 95: We recently upgraded the npm version due to a security alert from github. Can you check if this reintroduces that?

gastonponti: Yes, I talked with @cmcewen (who made the security resolution), and he told me that it was ok.
I'm trying to see if there is a way to check this package from the github database, but I'm not finding any way

jmrossy: Cool thanks for checking

PR #4081, opened 2020-06-11
Correctly propagate error message when sending an RPC

tkporter:
### Description

In the oracle repo, we're trying to account for specific reasons why a transaction may fail. We found that when a transaction fails and the RPC returns a specific error, it was being logged with `DEBUG=*` set, but the error itself was not being propagated to the call site. Instead, just a generic error `Error: Invalid JSON RPC response: {"id":10,"jsonrpc":"2.0"}` is given.

Here's an example without this PR, where I've set the transaction to have a gas price that is lower than the gas price minimum. You can see the error `gasprice is less than gas price minimum` is given in the RPC response, but the useful error information is stripped.

```
rpc:response { jsonrpc: '2.0',
  rpc:response   id: 1591907174344892,
  rpc:response   error:
  rpc:response    { code: -32000,
  rpc:response      message: 'gasprice is less than gas price minimum' } } +1s
  kit:provider:response { id: 10, jsonrpc: '2.0', result: undefined } +0ms
  kit:tx:result send-error: Error: Invalid JSON RPC response: {"id":10,"jsonrpc":"2.0"} at Object.InvalidResponse (/Users/trevor/celo-oracle/node_modules/web3-core-helpers/src/errors.js:42:16) at /Users/trevor/celo-oracle/node_modules/web3-core-requestmanager/src/index.js:144:36 at decoratedCallback (/Users/trevor/celo-oracle/node_modules/@celo/contractkit/lib/providers/celo-provider.js:102:13) at /Users/trevor/celo-oracle/node_modules/@celo/contractkit/lib/utils/rpc-caller.js:50:13 at process._tickCallback (internal/process/next_tick.js:68:7) +0ms
[2020-06-11T20:26:14.408Z] ERROR: oracle/64779 on Trevors-MacBook-Pro.local: Error reporting (context=reporter, errorContext=report)
    Error: Invalid JSON RPC response: {"id":10,"jsonrpc":"2.0"}
```

With these changes, we now have more informative errors:
```
  rpc:response { jsonrpc: '2.0',
  rpc:response   id: 1591907474447892,
  rpc:response   error:
  rpc:response    { code: -32000,
  rpc:response      message: 'gasprice is less than gas price minimum' } } +2s
  kit:provider:response { id: 10,
  kit:provider:response   jsonrpc: '2.0',
  kit:provider:response   result: null,
  kit:provider:response   error:
  kit:provider:response    { message: 'gasprice is less than gas price minimum',
  kit:provider:response      code: -32000 } } +0ms
  kit:tx:result send-error: Error: gasprice is less than gas price minimum at decoratedCallback (/Users/trevor/celo-oracle/node_modules/@celo/contractkit/lib/utils/rpc-caller.js:130:21) at XMLHttpRequest.request.onreadystatechange (/Users/trevor/celo-oracle/node_modules/web3-providers-http/src/index.js:102:13) at XMLHttpRequestEventTarget.dispatchEvent (/Users/trevor/celo-oracle/node_modules/xhr2-cookies/dist/xml-http-request-event-target.js:34:22) at XMLHttpRequest._setReadyState (/Users/trevor/celo-oracle/node_modules/xhr2-cookies/dist/xml-http-request.js:208:14) at XMLHttpRequest._onHttpResponseEnd (/Users/trevor/celo-oracle/node_modules/xhr2-cookies/dist/xml-http-request.js:318:14) at IncomingMessage.<anonymous> (/Users/trevor/celo-oracle/node_modules/xhr2-cookies/dist/xml-http-request.js:289:61) at IncomingMessage.emit (events.js:203:15) at endReadableNT (_stream_readable.js:1145:12) at process._tickCallback (internal/process/next_tick.js:63:19) +0ms
[2020-06-11T20:31:14.499Z] ERROR: oracle/65005 on Trevors-MacBook-Pro.local: Error reporting (context=reporter, errorContext=report)
    Error: gasprice is less than gas price minimum
```

The reason why this doesn't work out of the box is because of web3's HttpProvider send function: https://github.com/ethereum/web3.js/blob/1.x/packages/web3-providers-http/src/index.js#L95. The callback is not called with an error if the content of the RPC response includes an error.

### Other changes

I gave higher precedence to the error message rather than the stack in order to get a clean error (eg `gasprice is less than gas price minimum`) instead of the error message along with the stack in one big ugly string.

I originally wanted to override the type in the `types` directory of contractkit, but apparently this doesn't allow you to change the existing type definition and only allows you to extend it. My solution was to patch the type def

### Tested

Tested the fix with the oracle client

### Related issues

n/a

### Backwards compatibility

I believe it should still be backward compatible


gastonponti (approved):
I'm a little worry about the patch, because it will work only for the people that applies that patch.
But basically we already have the same problem with the `web3-celo.d.ts` that extends the types and it's required to include it in the `tsconfig.json`.
I would use the same extension approach only because we are already using it (this milestone we have an epic to fix of find a better workaround for the types)

But again, it's not part of the PR scope to fix these typing problems. So your call, both are acceptable.

And great adding. Details matter the most, haha. LGTM

tkporter:
Thanks! I did try the `web3-celo.d.ts` approach, but I was having trouble. I read a bit online that it can only help with adding properties to an object's type definition but not necessarily changing the existing types of properties in an object

PR #4083, opened 2020-06-12
[Wallet] Created and placed FeeDrawer component

tarikbellamine:
### Description
Created a FeeDrawer component that shows the total fees as a topline number and when clicked drops down all the fee line items.

### Other changes
Placed it in the SendConfirmation and TransferConfirmationCard.

### Tested
Yes

### Related issues
- #3645 

### Backwards compatibility
Yes

IMG_4234
IMG_4230
IMG_4228
IMG_4227


jeanregisser (approved):
Nice work! 🎉 
My only concern is that the list of wrapped fees is a bit static. But as always we'll revisit when we need more 👍

packages/mobile/src/send/SendConfirmation.test.tsx
```diff
@@ -43,16 +43,17 @@ describe('SendConfirmation', () => {
...
       <Provider store={store}>
         <SendConfirmation {...mockScreenProps} />
       </Provider>
     )
 
     // Initial render
     expect(toJSON()).toMatchSnapshot()
-    expect(queryByText('securityFee')).not.toBeNull()
-    expect(queryByText('0.001')).toBeNull()
+    // TODO: Find out why there is no handler function found for event "press"
```
aaronmgdr: yeah the way testing-library react-native does on press seems to be to just call the prop rather than triggering an actual press event.

packages/mobile/src/transactions/TransferConfirmationCard.tsx
```diff
@@ -90,6 +94,13 @@ function InviteSentContent({
         }
       />
       <HorizontalLine />
+      <FeeDrawer
+        currency={CURRENCY_ENUM.DOLLAR}
+        inviteFee={inviteFee}
+        isInvite={true}
+        securityFee={securityFee}
+        totalFee={inviteFee}
```
aaronmgdr: nit: maybe create a variable for total fee? seems like it would be easy to forget that total doesnt just mean invite and then months later we find this and "think woah we accidentally set total just the invite fee!"

packages/mobile/src/transactions/TransferConfirmationCard.tsx
```diff
@@ -72,6 +73,9 @@ function InviteSentContent({
 }: Props) {
   const { t } = useTranslation(Namespaces.sendFlow7)
   const totalAmount = amount
+  const inviteFee = getInvitationVerificationFeeInDollars()
+  // TODO: Use real fee
```
aaronmgdr, line 77: whats the plan for this?

PR #4090, opened 2020-06-12
[PGPNP] Add support for Google Secret Manager to signer

### LINKED ISSUE #3899: Package signer as needed for partner infrastructure

jmrossy:
### Description

Add support for Google Secret Manager to signer.
Generalize keystore utils and configs to support multiple options.
Update Readme with more details

### Tested

Tested using key in staging GCP project

### Related issues

- #3899 

### Backwards compatibility

Yes


packages/phone-number-privacy/signer/README.md
```diff
@@ -6,21 +6,38 @@ A service run by distributed partners that generates unique partial signatures f
 
 You can use the following environment variables to configure the Phone Number Privacy - Signer service:
 
-- `NODE_ENV` - eg. development or production
+- `NODE_ENV` - `development` or `production`
 - `SERVER_PORT` - The port on which the express node app runs (8080 by default).
-- `BLOCKCHAIN_PROVIDER` - The blockchain node provider for on chain requests. This could be a node with RPC set up.  For development testing you can use https://alfajores-forno.celo-testnet.org
-- `DB_HOST` - The URL under which your database is accessible.  Supports Postgres, MSSQL, MySQL, MariaDB, SQLite3, Oracle.
+- `BLOCKCHAIN_PROVIDER` - The blockchain node provider for chain state access. This could be a node with RPC set up.  For development testing you can use `https://alfajores-forno.celo-testnet.org`. For Mainnet, it would be `https://rc1-forno.celo-testnet.org`
```
aslawson: do we want to encourage the user of forno.  Maybe make this a softer offer (as most individuals we chatted with seemed to think it was counter productive to use the centralized forno endpoint).

Maybe `If you chose not to run your own node with this service, you can opt to use https://rc1-forno.celo-testnet.org`

packages/phone-number-privacy/signer/src/key-management/key-provider.ts
```diff
@@ -0,0 +1,22 @@
+import config, { SupportedKeystore } from '../config'
+import { AzureKeyProvider } from './azure-key-provider'
+import { GoogleKeyProvider } from './google-key-provider'
+import { KeyProvider } from './key-provider-base'
+
+let keyProvider: KeyProvider
```
aslawson, line 7: wondering if there is value in having this instantiated immediately rather than lazily instantiated when the first request comes in -- that way a service runner would know instantly if they aren't set up correctly.  or would that be caught by the config initialization anyway since incorrect input would match the `SupportedKeystore` enum?

aslawson (approved):
looks really good.  very clean with the abstract class :)

packages/phone-number-privacy/signer/README.md
jmrossy: Good idea, I'll change it

packages/phone-number-privacy/signer/src/key-management/key-provider.ts
jmrossy: Agreed! Will change

PR #4101, opened 2020-06-15
[Wallet][ReDesign] Name and number screen

jeanregisser:
### Description

This implements the [name and number screen redesign](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=5404%3A399).

Main highlights:
- `<PhoneNumberInput />` has been refactored and is hopefully now much simpler to reason about. It's been turned into a controlled component.
- Default country is set using the device locale setting.
- Number placeholder changes according to the selected country
- Language selection is now done in a dedicated screen (displayed with the new modal card style on iOS).
- Countries can now also be searched using their 2 or 3 letters ISO code. This makes it possible to find `United States` by just entering `us`.
- Similar to the previous implementation on Android, we ask the user for permission to read the device phone number when focusing the phone number input or pressing the country button.
- Performance should be hopefully improved when typing a number quickly (no more redux dispatch and iterating through the countries when typing).

### Other changes

- Fix position of `<DevSkipButton />` on iOS devices with notch

### Tested

name-and-number

### Related issues

- Epic #3645

### Backwards compatibility

Yes


packages/mobile/src/onboarding/registration/JoinCelo.tsx
```diff
@@ -0,0 +1,252 @@
...
+  state: State = {
+    name: this.props.cachedName || '',
+    ...getPhoneNumberState(
+      this.props.cachedNumber || '',
+      this.props.cachedCountryCallingCode || '',
+      RNLocalize.getCountry()
+    ),
+  }
+
+  UNSAFE_componentWillReceiveProps(nextProps: Props) {
```
jeanregisser: Don't judge me 😅 
This will be eliminated by switching to a function component. 
I wanted to keep the class component here for now to prevent unnecessary changes from the existing code.

packages/mobile/src/navigator/Headers.v2.tsx
```diff
@@ -97,6 +100,12 @@ export const headerWithCancelButton: StackNavigationOptions = {
   headerLeft: () => <CancelButton />,
 }
 
+export const headerWithCloseButton: StackNavigationOptions = {
+  ...emptyHeader,
+  headerLeft: () => <TopBarIconButton icon={<Times />} onPress={navigateBack} />,
+  headerLeftContainerStyle: { paddingLeft: 20 },
```
aaronmgdr, line 111: my understand is that paddings are 8, 12, 16 or 24. 

also how about having this style be defined in a Stylesheet.create?

packages/mobile/src/onboarding/registration/SelectCountry.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  const onSelect = useCallback(
+    (country: LocalizedCountry) => {
+      navigation.navigate(Screens.JoinCelo, { selectedCountryCodeAlpha2: country.alpha2 })
+
+      // Wait for next frame before navigating back
+      // so the user can see the changed selection briefly
+      // requestAnimationFrame(() => {
+      //   navigateBack()
```
aaronmgdr: ??

jeanregisser: Oops, obsolete comment, removed 👍

packages/mobile/src/navigator/Headers.v2.tsx
jeanregisser: Yes it should be one of these sizes, however it felt a bit too far on the right in the iOS modal style.
We can revisit later.

packages/mobile/src/onboarding/registration/JoinCelo.tsx
jmrossy: Why not put it in componentDidUpdate? Though we waste a render, we avoid using a deprecated method

jmrossy (approved):
Looks great to me!

packages/mobile/src/onboarding/registration/JoinCelo.tsx
jeanregisser: Good point. Changing this.

packages/mobile/src/onboarding/registration/JoinCelo.tsx
```diff
@@ -0,0 +1,252 @@
...
+    const countryCallingCode = this.countries.getCountryByCodeAlpha2(countryCodeAlpha2)
+      ?.countryCallingCode
+    const { cachedName, cachedNumber, cachedCountryCallingCode } = this.props
+
+    if (
+      cachedName === name &&
+      cachedNumber === e164Number &&
+      cachedCountryCallingCode === countryCallingCode
+    ) {
+      this.goToNextScreen()
```
i1skn, line 165: Should we add `this.props.setPromptForno(true)` here as well or it's cached in redux-persist?

packages/mobile/src/onboarding/registration/JoinCelo.tsx
```diff
@@ -0,0 +1,252 @@
...
+  container: {
+    flex: 1,
+    justifyContent: 'space-between',
+    backgroundColor: colors.onboardingBackground,
+  },
+  scrollContainer: {
+    padding: 24,
+    paddingTop: 80,
+  },
+  nameInput: {},
```
i1skn: I guess, we can remove this 🙂

packages/mobile/src/onboarding/registration/SelectCountry.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const inset = useSafeArea()
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.searchInputContainer}>
+        <SearchInput
+          placeholder={t('global:search')}
+          value={searchText}
+          onChangeText={onChangeText}
```
i1skn: Can we use here `setSearchText` directly without `onChangeText` function?

packages/mobile/src/onboarding/registration/SelectCountry.tsx
```diff
@@ -0,0 +1,86 @@
...
+  function onChangeText(text: string) {
+    setSearchText(text)
+  }
+
+  const onSelect = useCallback(
+    (country: LocalizedCountry) => {
+      navigation.navigate(Screens.JoinCelo, { selectedCountryCodeAlpha2: country.alpha2 })
+    },
+    [dispatch]
+  )
```
i1skn: I think here we can do:
```suggestion
function onSelect(country: LocalizedCountry) {
   navigation.navigate(Screens.JoinCelo, { selectedCountryCodeAlpha2: country.alpha2 })
 }
```

packages/react-components/components/FormField.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <View style={[styles.container, style]}>
+      <FormLabel style={styles.label}>{label}</FormLabel>
+      {children}
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {},
```
i1skn: I guess, we can remove this

packages/mobile/src/onboarding/registration/JoinCelo.tsx
jeanregisser: It's persisted.

i1skn (commented):
Overall looks very good and thanks for such a tremendous effort!

packages/mobile/src/onboarding/registration/JoinCelo.tsx
jeanregisser: Yep, thanks.

packages/mobile/src/onboarding/registration/SelectCountry.tsx
jeanregisser: Yes, thanks 👍

packages/mobile/src/onboarding/registration/SelectCountry.tsx
jeanregisser: Indeed 👍

packages/react-components/components/FormField.tsx
jeanregisser: 👍

PR #4104, opened 2020-06-15
Add POC telemetry handler

nambrot:
### Description

Adds a POC for handling telemetry in contractkit


packages/contractkit/src/utils/rpc-caller.ts
```diff
@@ -110,3 +128,13 @@ export class DefaultRpcCaller implements RpcCaller {
...
   }
 }
+
+export interface TelemetryHandler {
+  handleEvent: (path: string[], measurements: { duration: number }, metadata: any) => void
+}
+
+export class DebugTelemetryHandler implements TelemetryHandler {
+  handleEvent(path: string[], measurements: { duration: number }, metadata: any) {
+    debugRpcTelemetry('Path: %o, Duration: %s, metadata: %o', path, measurements.duration, metadata)
```
mcortesi, line 151: where do you want to use this? You're using the debug handler, which is not a great fit if this is something you want always "on" for some cases

packages/contractkit/src/utils/rpc-caller.ts
```diff
@@ -110,3 +128,13 @@ export class DefaultRpcCaller implements RpcCaller {
     }
   }
 }
+
+export interface TelemetryHandler {
+  handleEvent: (path: string[], measurements: { duration: number }, metadata: any) => void
+}
+
+export class DebugTelemetryHandler implements TelemetryHandler {
+  handleEvent(path: string[], measurements: { duration: number }, metadata: any) {
```
mcortesi, line 150: I would add some documentation here... 
What does this is supposed to mean... ? path, measurements? metadata? 

For me; it's something no one else would know what to do with this or what is for

mcortesi (approved):
approved w/comments

packages/contractkit/src/utils/rpc-caller.ts
nambrot: Yeah I think the debug handler is a misnomer as we will probably always want to just add it to `debug`? The intention was to use this in the mobile wallet to measure eth_calls

packages/contractkit/src/utils/rpc-caller.ts
nambrot: Thanks for the comments. This wasn't mean to be an actual PR but just as a way to start the conversation. Would certainly properly type and document the actual payload. Right now it's just to measure how long calls take

aaronmgdr:
close?

barbaraliau:
@nambrot is this PR still relevant?

barbaraliau:
Closing as this PR is stale. Please reopen if it is still relevant!

PR #4106, opened 2020-06-15
[Wallet][ReDesign] Update terms screen style

jeanregisser:
### Description

This implements the [terms screen redesign](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=5404%3A375).

Note: Depends on #4101, review and merge it first!

### Other changes

N/A

### Related issues

- Epic #3645

### Backwards compatibility

Yes

PR #4120, opened 2020-06-16
[Wallet][ReDesign] Implement new pincode

jeanregisser:
### Description

This implements the [pincode redesign](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=5404%3A685).

Note: Depends on #4106, review and merge it first!

### Other changes

N/A

### Tested

**Onboarding**
pincode2

**Enter PIN**

### Related issues

- Epic #3645

### Backwards compatibility

Yes


packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -158,6 +157,23 @@ const verificationScreens = (Navigator: typeof Stack) => {
   )
 }
 
+const pincodeSetScreenOptions = ({
```
cmcewen, line 168: This is the only part I don't like of new navigation - definitely feels like this should live on the component. What do you think about making this part of the screen's file and exporting it as opposed to doing it here?

packages/mobile/src/pincode/PincodeDisplay.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const prevPinRef = useRef(pin)
+
+  useEffect(() => {
+    const prevPin = prevPinRef.current
+    prevPinRef.current = pin
+
+    // Check if pin length is smaller, so as not to reveal previous digits
+    // when deleting
+    if (pin.length < prevPin.length) {
+      LayoutAnimation.configureNext({
```
cmcewen, line 27: ✨

packages/mobile/src/pincode/PincodeEnter.tsx
```diff
@@ -57,13 +51,11 @@ class PincodeEnter extends React.Component<Props, State> {
   }
 
   onWrongPin = () => {
```
cmcewen, line 53: this looks like it will be easy to do with these changes, but just a note that i'd like to limit the number of incorrect entries before someone gets locked out for a while

cmcewen (approved):
woot!

packages/mobile/src/navigator/Navigator.tsx
jeanregisser: Yes I agree, we could easily adapt the old way and use `MyComponent.navigationOptions` (or similar) inside `Navigator.tsx`.
We can discuss quickly with the rest of the team and do a global pass on this later on.

packages/mobile/src/pincode/PincodeEnter.tsx
jeanregisser: 👍

packages/mobile/src/navigator/Navigator.tsx
i1skn: Yeah, I've noticed this as well, think that keeping options close to the screen is good idea 👍

packages/mobile/src/pincode/Pincode.tsx
```diff
@@ -2,35 +2,44 @@
...
-
+function Pincode({
+  title,
+  errorText,
+  maxLength = PIN_LENGTH,
+  pin,
+  onChangePin,
+  onCompletePin,
+}: Props) {
   const onDigitPress = useCallback(
```
i1skn: I guess, we can remove `useCallback` here

i1skn (approved):
Great work!!! See my only one small comment 🙂

packages/mobile/src/pincode/Pincode.tsx
jeanregisser: yep!

PR #4123, opened 2020-06-16
[Wallet] Fix failed escrow payment fetch right after verification

i1skn:
### Description

After successful phone number verification we want to wait for one block to mined. This is because we want the latest attestation is "mined" as well before fetching an escrow payment, otherwise we can get `This account does not have enough attestations to withdraw this payment.`

### Tested

iPhone

### Backwards compatibility

Yes


packages/mobile/src/identity/verification.ts
```diff
@@ -138,48 +138,52 @@ export function* doVerificationFlow() {
...
-
-    yield put(setVerificationStatus(VerificationStatus.RequestingAttestations))
-    const attestations: ActionableAttestation[] = yield call(
-      requestAndRetrieveAttestations,
-      attestationsWrapper,
-      phoneHash,
-      account,
-      status.numAttestationsRemaining
-    )
+    if (!status.isVerified) {
```
i1skn, line 141: This just removes duplication of:
```
yield put(setVerificationStatus(VerificationStatus.Done))
yield put(setNumberVerified(true))
return true
```

cmcewen (approved):

packages/mobile/src/identity/verification.ts
jmrossy: I preferred it as it was cause now this large code block gets further indented, but it's nbd either way

jmrossy (approved):

PR #4124, opened 2020-06-17
[Wallet] Transaction feed improvements

### LINKED ISSUE #715: “Unknown” text is shown when user getting the escrew payment or adding amount from Add and Withdraw option
**Frequency:** 100%
** Repro App version:** Celo integration build v1.8.2 (36)
**Does not repro:** app store build v1.8.0

**Repro Steps:** 
2) Make sure the user has verified the phone number. 
3) Make sure user is got the escrow payment 
4) Observe Unknown text is shown. 
5) Add amount form Add & Withdraw option. 
6) Observe Unknown text is shown. 

 **Current Behavior:**  Unknown text is shown
**Expected Behavior:** Appropriate text should be shown instead of Unknown text. 
**Attachment:**  UnknownText.png

### LINKED ISSUE #4086: “Unknown” text is shown when user getting the escrew payment or adding amount from Add and Withdraw option
**Frequency:** 100%
** Repro App version:** Celo integration build v1.8.2 (36)
**Does not repro:** app store build v1.8.0

**Repro Steps:** 
2) Make sure the user has verified the phone number. 
3) Make sure user is got the escrow payment 
4) Observe Unknown text is shown. 
5) Add amount form Add & Withdraw option. 
6) Observe Unknown text is shown. 

 **Current Behavior:**  Unknown text is shown
**Expected Behavior:** Appropriate text should be shown instead of Unknown text. 
**Attachment:**  UnknownText.png

tarikbellamine:
### Description
Did three main things to improve the tx feed:
- If the user has previously given permission, we now pull their contacts into the `recipientCache` when WalletHome mounts
- Created `recentTxRecipientCache` that stores the last 10 recipients the user interacted with and persists it (unlike `recipientCache`). We use this to populate the tx feed when the app first loads and the larger `recipientCache` hasn't finished populating so that we don't see a bunch of phone numbers instead of names
- Created a hacky workaround for the blockchain API feeding useless EscrowSent information. Matching transaction timestamps with invite timestamps to deduce who the recipient was so that we don't populate "Unknown" for all escrow transactions

Unfortunately, we don't store reclaimed escrow payments anywhere so wasn't able to build something for EscrowReceived. 

### Tested
Yes

### Related issues
- Part of #3988
- Closes #4086 

### Backwards compatibility
Yes


packages/mobile/src/home/WalletHome.tsx
```diff
@@ -90,12 +96,29 @@ export class WalletHome extends React.Component<Props> {
...
+  tryImportContacts = async () => {
+    const { numberVerified, recipientCache } = this.props
+
+    // Only import contacts if number is verified and
+    // recip cache is empty so we haven't already
+    if (!numberVerified || recipientCache.length) {
+      return
+    }
+
+    const hasGivenContactPermission = await requestContactsPermission()
```
jmrossy: I think you're looking for `checkContactsPermission` here

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -90,12 +96,29 @@ export class WalletHome extends React.Component<Props> {
...
   }
 
-  componentDidMount() {
+  componentDidMount = async () => {
     // TODO find a better home for this, its unrelated to wallet home
     this.props.initializeSentryUserContext()
     if (SHOW_TESTNET_BANNER) {
       this.showTestnetBanner()
     }
+    await this.tryImportContacts()
```
jmrossy: Consider setting a timeout and triggering this after 2 seconds, so it doesn't slow down other things like loading the feed.

packages/mobile/src/transactions/actions.ts
```diff
@@ -54,12 +57,22 @@ export interface NewTransactionsInFeedAction {
   transactions: TransactionFeedFragment[]
 }
 
+export interface FetchRecentTxRecipients {
+  type: Actions.FETCH_RECENT_TX_RECIPIENTS
+}
+export interface UpdatedRecentTxRecipientsCacheAction {
```
jmrossy, line 60: Nit: spacing

packages/mobile/src/transactions/saga.ts
```diff
@@ -98,6 +112,53 @@ export function* sendAndMonitorTransaction<T>(
   }
 }
 
+export function* fetchRecentTxRecipients() {
```
jmrossy: I tend to use `fetch` for things that go to network. How about `refresh` or `find`?

packages/mobile/src/home/saga.ts
```diff
@@ -42,6 +42,8 @@ export function* refreshBalances() {
   yield put(fetchDollarBalance())
   yield put(fetchGoldBalance())
   yield put(fetchSentEscrowPayments())
+  // TODO: Find a better home for this
+  yield put(fetchRecentTxRecipients())
```
jmrossy: Like you'd TODO says, this doesn't seem like the right trigger for this. Should it happen on feed update?

tarikbellamine: Which action do you suggest?

I originally tried `NEW_TRANSACTION_IN_FEED` but `addressToE164PhoneNumber` is needed  and doesn't get updated until the send saga is completely done. Ideally we'd pair it with an action that triggers when both a send or receive transfer is completed.

packages/mobile/src/home/WalletHome.tsx
tarikbellamine: Yup, didnt know that existed. Thanks!

packages/mobile/src/home/WalletHome.tsx
tarikbellamine: That may not be necessary. Looks like the first render fills in the tx feed pretty well. You'll be able to notice when the import is done because recipients will populate ~5 seconds later and it doesn't appear that anything is being held up

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -90,12 +96,28 @@ export class WalletHome extends React.Component<Props> {
     this.props.refreshAllBalances()
   }
 
-  componentDidMount() {
+  componentDidMount = async () => {
```
cmcewen: so this is a "fire and forget" situation where you want to trigger tryImportContacts but don't really care about the result. i'm assuming the linter complained about an unhandled promise if you just call `this.tryImportContacts()` - which you could add a catch to if you wanted to. just noting the semantics here since we don't really care about awaiting the result

packages/mobile/src/transactions/saga.ts
```diff
@@ -16,19 +19,30 @@ import {
...
+  KnownFeedTransactionsType,
+  standbyTransactionsSelector,
+} from 'src/transactions/reducer'
 import { sendTransactionPromises, wrapSendTransactionWithRetry } from 'src/transactions/send'
 import { StandbyTransaction, TransactionStatus } from 'src/transactions/types'
 import Logger from 'src/utils/Logger'
 
 const TAG = 'transactions/saga'
 
+const RECENT_TX_RECIPIENT_CACHE_LIMIT = 10
```
cmcewen, line 37: what led you to this number? just based on screen size, storage, the work being expensive?

packages/mobile/src/transactions/reducer.ts
```diff
@@ -14,15 +16,17 @@ export interface State {
   // tx feed query watcher. Necessary so we don't re-process
   // txs more than once.
   knownFeedTransactions: KnownFeedTransactionsType
+  recentTxRecipientsCache: NumberToRecipient
 }
 
 export interface KnownFeedTransactionsType {
-  [txHash: string]: boolean
+  [txHash: string]: string | boolean
```
cmcewen, line 25: is it worth just making these all strings? it looks like boolean true represents it being an exchange - is there any other type represented?

packages/mobile/src/transactions/saga.ts
```diff
@@ -98,6 +112,53 @@ export function* sendAndMonitorTransaction<T>(
...
+    !Object.keys(knownFeedTransactions).length
+  ) {
+    return
+  }
+
+  const knownFeedAddresses = Object.values(knownFeedTransactions)
+
+  let remainingCacheStorage = RECENT_TX_RECIPIENT_CACHE_LIMIT
+  const recentTxRecipientsCache: NumberToRecipient = {}
+  // Start from back of the array to get the most recent transactions
```
cmcewen, line 128: I don't know that object.values has any guarantees about ordering `The ordering of the properties is the same as that given by looping over the property values of the object manually` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values

I think if you expect them to be ordered by timestamp it's worth making sure somehow

cmcewen (approved):
looks good overall - few questions for ya

packages/mobile/src/transactions/saga.ts
tarikbellamine: It was a combination of screen size and storage considerations. 10 tx feed items is what's viewable without scrolling on a pretty large phone. We could always store more but wanted to be mindful of storage.

Do you think the number should be different?

packages/mobile/src/home/WalletHome.tsx
jmrossy: I would test it on a shitty phone before to be sure but okay sounds good

packages/mobile/src/home/saga.ts
jmrossy: I think `NEW_TRANSACTION_IN_FEED` is the closet we have to when a send or receive is complete. Can you clarify why it didn't work for this?

packages/mobile/src/transactions/reducer.ts
jmrossy: We could do this, we're not really using the boolean value for anything atm

tarikbellamine: I believe Exchange and Transfer are the only two types of transactions. We could make them all strings and set something arbitrary for exchanges but leaving it as is also makes it easy to differentiate between the two.

Is there a concern that I may not be considering?

packages/mobile/src/transactions/saga.ts
tarikbellamine: I originally tried creating a Map for that reason but don't think we can store those in Redux.

Given the logic is not visible to the user (i.e., no specific user expectation of caching recipients) and in practice [order is generally maintained for consistent property types](https://www.stefanjudis.com/today-i-learned/property-order-is-predictable-in-javascript-objects-since-es2015/), I opted to avoid the added complexity of doing a timestamp comparison.

packages/mobile/src/transactions/saga.ts
cmcewen: Sounds reasonable - was just curious

packages/mobile/src/transactions/reducer.ts
cmcewen: Not a huge concern - more making it clear what the existence of a value in this object means. I had to look at the other code to figure out tx -> string with tx.address and escrow -> boolean true

packages/mobile/src/home/saga.ts
tarikbellamine: The query that triggers `NEW_TRANSACTION_IN_FEED` has a very short polling cycle  so it picks up on a send transaction before the mapping of address to phone number has been stored in `addressToE164Number`. This is a problem because `addressToE164PhoneNumber` is  needed to do the lookup from address -> e164PhoneNumber -> recipient.

This only happens on the first transaction to a particular recipient. If this is a transaction to someone you've already transacted with (i.e., the address to phone number mapping already exists in `addressToE164PhoneNumber`), then this isn't an issue

packages/mobile/src/transactions/reducer.ts
tarikbellamine: Gotcha, will add a comment to make that distinction clear

packages/mobile/src/home/WalletHome.tsx
tarikbellamine: Opted to use a setTimeout of 0.5 secs as a compromise for now and we can always ramp that up if need be

packages/mobile/src/home/saga.ts
jmrossy: Ah I think the problem is that you need to wait for this saga to run first:
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/identity/saga.ts#L94

So how about trigger your new saga whenever UPDATE_E164_PHONE_NUMBER_ADDRESSES *or* NEW_TRANSACTION_IN_FEED happens?

jmrossy:
Removed the automerge, lets please address my comment before merging.

packages/mobile/src/home/saga.ts
tarikbellamine: I think that combo will do the trick. Will test it out now

jmrossy (approved):

PR #4142, opened 2020-06-18
[pgpnp] fix combiner index tests

aslawson:
### Description

-failed tests in index.test.ts were not indicated as test failing because the requests were asynchronous and required `await`
- fixed the failing tests
- will update the signer index.test.ts next

### Other changes

NA

### Tested

- ran tests and confirmed when they failed, it was indicated as a test failure

### Related issues

NA

### Backwards compatibility

Yes


jmrossy (approved):
LGTM, file just needs to be prettified

PR #4144, opened 2020-06-18
[Wallet] Add error related analytics

### LINKED ISSUE #4131: Devs SBAT see error messages in analytics
### Expected Behavior

Error message detailed text is included in analytics for all errors

### Current Behavior

Error messages are in local logs but only the error shown to the user (ex:  `nuxVerification2:verificationFailure`) included in analytics

annakaz:
### Description

Add error related analytics to make verification issues easier to debug

### Tested

Just adding analytics, have not tested locally

### Related issues

- Fixes #4131


packages/mobile/src/escrow/saga.ts
```diff
@@ -88,6 +90,7 @@ function* transferStableTokenToEscrow(action: EscrowTransferPaymentAction) {
     yield put(fetchSentEscrowPayments())
   } catch (e) {
     Logger.error(TAG + '@transferToEscrow', 'Error transfering to escrow', e)
+    CeloAnalytics.track(CustomEventNames.escrow_failed_to_transfer, { error: e.message })
```
jmrossy: Do we want to be using Analytics as the channel for tracking errors? It seems like Sentry is already the goto place for that. We can send events to sentry manually if we want

tarikbellamine: Analytics probably shouldn't be the only place but I do think it's useful to track here.

Sentry could potentially be configured to serve some of the same needs but I believe this gives us an easy way to contextualize errors with general app usage and group and analyze them the way we want. I imagine this won't be heavily relied upon for debugging.

packages/mobile/src/exchange/saga.ts
```diff
@@ -260,11 +270,13 @@ export function* exchangeGoldAndStableTokens(action: ExchangeTokensAction) {
     )
 
     if (!txId) {
+      CeloAnalytics.track(CustomEventNames.missing_tx_id)
       Logger.error(TAG, 'No txId. Did not exchange.')
       return
     }
     yield call(sendAndMonitorTransaction, txId, tx, account)
   } catch (error) {
+    CeloAnalytics.track(CustomEventNames.exchange_failed, { error })
```
tarikbellamine: Looks like you forgot to add this to `CustomEventNames`

packages/mobile/src/escrow/saga.ts
nambrot: I'm not working on the wallet day to day of course, but my 2cents are that indeed they should be sent via analytics. In a lot of places in the mobile wallet we are catching errors which makes using analytics data very challenging because it's virtually impossible to tell what has happened without knowing what kind of errors occurred.

That being said, what I'd say is more important is for the mobile wallet team to decide what the various outlets (device logs, sentry, segment analytics) are supposed to do. Maybe we aren't supposed to use analytics to analyze how users are moving through the app?

packages/mobile/src/escrow/saga.ts
```diff
@@ -178,6 +181,7 @@ function* withdrawFromEscrow() {
     Logger.showMessage(i18n.t('inviteFlow11:transferDollarsToAccount'))
   } catch (e) {
     Logger.error(TAG + '@withdrawFromEscrow', 'Error withdrawing payment from escrow', e)
+    CeloAnalytics.track(CustomEventNames.escrow_failed_to_withdraw, { error: e })
```
nambrot: Sometimes we pass through `e` and sometimes `e.message`, what should be the general default?

packages/mobile/src/identity/verification.ts
```diff
@@ -124,7 +124,10 @@ export function* doVerificationFlow() {
       contractKit.contracts.getAccounts,
     ])
 
-    CeloAnalytics.track(CustomEventNames.verification_setup)
+    CeloAnalytics.track(CustomEventNames.verification_setup, {
```
nambrot: No strong opinion, but IMO there should be a separate event for this information. It can be joined against verification setup via the ID?

packages/mobile/src/identity/verification.ts
```diff
@@ -391,6 +401,7 @@ function attestationCodeReceiver(
       yield put(inputAttestationCode({ code, issuer }))
     } catch (error) {
       Logger.error(TAG + '@attestationCodeReceiver', 'Error processing attestation code', error)
+      CeloAnalytics.track(CustomEventNames.verification_error, { error })
```
nambrot: I think it would be nice to get some events when we receive the code ahead as well

nambrot (commented):
Can we add more events in the send abstraction?

packages/mobile/src/escrow/saga.ts
jmrossy: I'm _okay_ with us using analytics for this but skeptical we'll made good use of that data given that we don't check it. We do check Sentry. I'd rather us use one channel well than use both half-heartedly.

annakaz: +1 to Tarik on this not being the only place to track, but a useful place for us to contextualize the errors

@jmrossy I agree that Sentry is the best place for us to first become aware of errors as we check it regularly (the errors will still be piped to Sentry), but it is also helpful to have access to them in our Analytics to see what other actions occurred leading up to the error. This has been helpful while debugging verification flakiness - our current analytics only show that a generic "Verification Failed" error message was thrown, but while looking into different types of failures in the analytics we'd like to be able to see more details on that error

packages/mobile/src/exchange/saga.ts
annakaz: Thanks for catching this

packages/mobile/src/identity/verification.ts
annakaz: Good idea, added `verification_hash_retrieved` with the data

packages/mobile/src/escrow/saga.ts
annakaz: Switched over to e.message consistently, this should be the default to avoid an unexpected large error in the analytics

annakaz:
Send events already added in Nam's PR here: https://github.com/celo-org/celo-monorepo/pull/4165

tarikbellamine (approved):

PR #4148, opened 2020-06-19
updated rc1 and baklava env with latest docker tag post oracle upgrade

jasonrudianto1:
### Description

Updated docker image tag after upgrading oracles.


tkporter (approved):
Woooo

PR #4149, opened 2020-06-19
Get service principal object id instead of its clientId/appId, account for MSI clusters

tkporter:
### Description

When I tried redeploying some oracles I ran into some issues where the cluster SP was not getting getting the correct role assignment as a managed identity operator for the oracle identity scopes. Turns out az was silently failing when we would pass in the SP's clientId/appId as an object id, so this gets the SP's objectId and uses that instead.

Also includes:
* Accounts for situations where the AKS cluster does not use a service principal and instead uses Managed Service Identity (MSI) - this is the case for the test cluster
* Adds the `oracledev` environment, which is an environment intended for testing infrastructure changes or general development

### Other changes

n/a

### Tested

Redeployed a few times on baklava, this fixed the issue

Was able to deploy oracledev

### Related issues

n/a

### Backwards compatibility

Yes


codyborn (approved):
Nice, thanks for adding this!

PR #4145, opened 2020-06-18
[Wallet][ReDesign] Invite code screen

jeanregisser:
### Description

This implements the [enter invite code redesign](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=5404%3A430).

Note: I've left the "Restore Existing Wallet" for now at the bottom until we have the new flow where this choice is made possible earlier.

### Other changes

- Made shadows closer to the designs on iOS

### Tested

invite-code2

### Related issues

- Epic #3645

### Backwards compatibility

Yes

PR #4152, opened 2020-06-19
[Wallet] Fix withTranslation HOC typing

jeanregisser:
### Description

This PR fixes the typing of our `withTranslation` HOC.

This is so we can still call hoisted static methods from the original component in a type safe way and not causing a TypeScript error.

This is useful in `Navigator` when we want to access `MyComp.navigationOptions`.

The small inconvenience of this is we have to explicitly pass the component `Props` when calling `withTranslation<Props>(...)`.
This is very similar to the `connect` HOC we already use.

I tried to lift this requirement but couldn't. Maybe if one day TypeScript supports generics of generics it could be inferred automatically. But for now I think it's good enough.

Note: this PR depends on #4145, please merge that one first.

### Tested

TypeScript passes successfully.

### Backwards compatibility

Yes.

### DIFF
--- a/packages/mobile/src/i18n.ts
+++ b/packages/mobile/src/i18n.ts
@@ -6,7 +6,6 @@ import {
   initReactI18next,
   WithTranslation,
   withTranslation as withTranslationI18Next,
-  WithTranslationProps,
 } from 'react-i18next'
 import * as RNLocalize from 'react-native-localize'
 import Logger from 'src/utils/Logger'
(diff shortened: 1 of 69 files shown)


packages/mobile/src/i18n.ts
```diff
@@ -101,10 +100,10 @@ RNLocalize.addEventListener('change', () => {
...
-export const withTranslation = (namespace: Namespaces) => <P extends WithTranslation>(
-  component: React.ComponentType<P>
-): React.ComponentType<Omit<P, keyof WithTranslation> & WithTranslationProps> =>
-  // cast as `any` here otherwise TypeScript complained
-  hoistStatics(withTranslationI18Next(namespace)(component), component as any)
+export const withTranslation = <P extends WithTranslation>(namespace: Namespaces) => <
+  C extends React.ComponentType<P>
+>(
+  component: C
+) => hoistStatics(withTranslationI18Next(namespace)(component), component)
```
jeanregisser, line 108: This is the important change in this PR.

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -204,7 +204,6 @@ const nuxScreens = (Navigator: typeof Stack) => (
     <Navigator.Screen
       name={Screens.EnterInviteCode}
       component={EnterInviteCode}
-      // @ts-ignore this is caused by withTranslation HOC
       options={EnterInviteCode.navigationOptions}
```
jeanregisser, line 214: This is the TypeScript error it fixes.

cmcewen (commented):
Niceeeee! Also some day we will have all hooks so no need to use withTranslation 😂

cmcewen (approved):

PR #4153, opened 2020-06-19
[PGPNP] Add timeout/logging for combiner request to signers

aslawson:
### Description

- Add timeout/logging for combiner request to signers

### Other changes

- Added MockSecretManager as a secret manager option

### Tested

Ran a local version of the signer and added a ` await sleep(1000000)`. confirmed that it logged the error message.

### Related issues

NA

### Backwards compatibility

Yes -- added default config value


packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -104,6 +108,21 @@ async function requestSignatures(request: Request) {
   return { successCount, signatures, majorityErrorCode }
 }
 
+function requestSigatureWithTimeout(
```
codyborn: nit typo: `requestSignatureWithTimeout`

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -104,6 +108,21 @@ async function requestSignatures(request: Request) {
...
+  service: SignerService,
+  request: Request
+): Promise<FetchResponse | boolean> {
+  return Promise.race([requestSigature(service, request), abort()])
+}
+
+async function abort(): Promise<boolean> {
```
codyborn: It looks like `node-fetch` has an option for specifying a timeout via an extension:
https://www.npmjs.com/package/node-fetch#options

aslawson: ya i saw that too, but not as clean of an example as what you posted so it has seemed overly complicated.

packages/phone-number-privacy/combiner/src/config.ts
```diff
@@ -119,6 +121,7 @@ if (DEV_MODE) {
     },
     pgpnpServices: {
       signers: functionConfig.pgpnpservices.signers,
+      timeoutMilliSeconds: functionConfig.pgpnpservices.timeoutMilliSeconds || 30 * 1000,
```
jmrossy: How about we lower this to 5s? Any delay more than that will noticeably hurt the wallet send flow UX. If we start seeing timeouts, maybe we can relax it.

jmrossy (commented):
LGTM! Just one comment about the default timeout.
I'll re-review after Cody adds his changes

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
aslawson: kk updated to use that -- it actually made the tests much easier

packages/phone-number-privacy/combiner/src/config.ts
aslawson: sounds good

packages/phone-number-privacy/signer/test/index.test.ts
```diff
@@ -66,6 +66,7 @@ describe(`POST /getBlindedMessageSignature endpoint`, () => {
           {
             success: true,
             signature: BLS_SIGNATURE,
+            version: VERSION,
```
aslawson, line 64: If you want to return this from the signer, you'll have to update `SignMessageResponse` interface that the combiner defines as the return object to include `version` field as well.

packages/phone-number-privacy/signer/src/salt-generation/get-salt.ts
```diff
@@ -9,6 +9,7 @@ import {
   phoneNumberHashIsValidIfExists,
 } from '../common/input-validation'
 import logger from '../common/logger'
+import { VERSION } from '../config'
```
aslawson, line 12: where is this defined?  I don't see VERSION added to the config.ts in the /signer directory

aslawson (commented):
@codyborn added some comments on your update (it looks like i'm commenting on my own pr 😂)

packages/phone-number-privacy/signer/src/salt-generation/get-salt.ts
codyborn: It was added in a previous PR: https://github.com/celo-org/celo-monorepo/blob/aae2b147608ef1303df18b208c6778e01a22ccfd/packages/phone-number-privacy/signer/src/config.ts#L3

packages/phone-number-privacy/signer/test/index.test.ts
codyborn: Missed this, thank you!

.env
```diff
@@ -59,6 +59,21 @@ MOCK_ORACLE_CRON_SCHEDULE="*/5 * * * *"
 MOCK_ORACLE_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/oracle"
 MOCK_ORACLE_DOCKER_IMAGE_TAG="default"
 
+ORACLE_DOCKER_IMAGE_REPOSITORY="celoprod.azurecr.io/oracle/baklava"
```
aslawson: @codyborn these look unrelated :)

packages/contractkit/src/wrappers/Governance.ts
```diff
@@ -56,6 +56,7 @@ export interface GovernanceConfig {
   queueExpiry: BigNumber
   stageDurations: ProposalStageDurations
   participationParameters: ParticipationParameters
+  approver: Address
```
aslawson: and this

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -48,63 +50,76 @@ export async function handleGetDistributedBlindedMessageForSalt(
...
+          signatures.push({ url: service.url, signature: signResponse.signature })
           successCount += 1
+          // Send response immediately once we cross threshold
+          if (!sentResult && successCount >= config.thresholdSignature.threshold) {
+            sentResult = true
+            const combinedSignature = await BLSCryptographyClient.combinePartialBlindedSignatures(
+              signatures,
+              request.body.blindedQueryPhoneNumber
+            )
+            response.json({ success: true, combinedSignature, version: VERSION })
```
jmrossy, line 91: Can you confirm that this actually sends this response immediately instead of just adjusting the properties of the response object for express to handle later?

.env
codyborn: Oops, thanks for noticing

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
codyborn: Yeah, I've tested it out to confirm. While testing it I used @aslawson's method of running two signers: one normally and the other with a 1000s sleep. I've configured the new combiner timeout to be 100 seconds and verified that it returns the salt immediately after the non-slow signer has provided its signature. 

I originally found the approach via this article:
https://www.bennadel.com/blog/3275-you-can-continue-to-process-an-express-js-request-after-the-client-response-has-been-sent.htm

tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
```
codyborn: I've added this for testing in VSCode which is now run at the root directory

codyborn (approved):
Signing off on @aslawson's changes

tsconfig.json
jmrossy: I'm worried this will cause issues or confusion with our existing 'root' tsconfig file here: https://github.com/celo-org/celo-monorepo/blob/master/packages/typescript/tsconfig.json#L1

jmrossy (commented):
Everything lgtm except the tsconfig file. can we remove that?

aslawson (commented):
@codyborn changes looks good, but i'll wait on the resolution of @jmrossy's concern to merge

PR #4155, opened 2020-06-19
[PGPNP] Add support for MySql and MSSql to Signer

jmrossy:
### Description

Add support for postgres db alternatives
Initialize and test DB connection on service start

### Tested

Ran services against MySql in GCP staging and MS SQL in Azure staging

### Backwards compatibility

Yes


packages/phone-number-privacy/signer/src/database/database.ts
```diff
@@ -1,16 +1,88 @@
...
-})
+let db: Knex
+export async function initDatabase(doTestQuery = true) {
+  logger.info('Initializing database connection')
+  const { type, host, port, user, password, database, ssl } = config.db
 
-export function getDatabase() {
+  let dbConfig: any
+  let client: string
+  if (type === SupportedDatabase.Postgres) {
```
codyborn, line 14: Should this be a case insensitive comparison? No strong preference but wanted to be explicit.

packages/phone-number-privacy/signer/src/database/database.ts
```diff
@@ -1,16 +1,88 @@
...
+  }
+
+  db = knex({
+    client,
+    connection: dbConfig,
+    debug: DEV_MODE,
+  })
+
+  if (doTestQuery) {
+    await executeTestQuery(db)
```
codyborn, line 57: nice!

codyborn (approved):
Looks good! Would be great to add some simple unit tests before checking in.

packages/phone-number-privacy/signer/src/database/database.ts
jmrossy: We lower case the value as it gets put in the config so that's covered

PR #4157, opened 2020-06-19
Replace Celo Gold with CELO and add warning about using Ethereum wallets

nategraf:
### Description

Two separate users have reported being confused by the reference to ERC-20 in this overview document, and as a result trying to use an Ethereum wallet under the impression that Celo exists on the Ethereum network. This PR adds a warning to hopefully prevent this confusion in the future.

Also, since renaming the Celo native asset to CELO, many docs need to be updated. Here is one small update.


packages/docs/overview.md
```diff
@@ -14,9 +14,13 @@ The code of the Celo Blockchain has shared ancestry with [Ethereum](https://www.
...
 
-Celo has a native unit of accounting, the cryptocurrency **Celo Gold**, equivalent to Ether in Ethereum. Celo's ledger consists of **accounts**, identified by an **address**. There are two types of account. **Externally owned accounts** have an associated Celo Gold balance and are controlled by a user holding the associated public-private keypair. **Contract accounts** contain the code and data of a single smart contract which can be called and manipulate its own stored data.
+Celo has a native unit of accounting, the cryptocurrency **CELO**, equivalent to Ether in Ethereum. Celo's ledger consists of **accounts**, identified by an **address**. There are two types of account. **Externally owned accounts** have an associated CELO balance and are controlled by a user holding the associated public-private keypair. **Contract accounts** contain the code and data of a single smart contract which can be called and manipulate its own stored data.
```
jmrossy:
```suggestion
Celo has a native unit of accounting, the cryptocurrency **CELO**, comparable to Ether on Ethereum. Celo's ledger consists of **accounts**, identified by an **address**. There are two types of account. **Externally owned accounts** have an associated CELO balance and are controlled by a user holding the associated public-private keypair. **Contract accounts** contain the code and data of a single smart contract which can be called and manipulate its own stored data.
```

packages/docs/overview.md
```diff
@@ -128,9 +132,13 @@ The Attestations contract allows a user to request attestations to their phone n
 
 Celo provides a number of enhancements to regular transactions as familiar to Ethereum developers.
 
-Celo Gold has a duality as both the native currency and is also an ERC-20 token, simplifying the work of application developers.
+The Celo native asset has a duality as both the native currency and is also an ERC-20 token, simplifying the work of application developers.
+
+{% hint style="warning" %}
+Celo assets exist on an independent blockchain, and cannot be accessed through wallets that connect to the Ethereum network. Only use wallets disigned to work with the Celo network.
```
jmrossy:
```suggestion
Celo assets exist on an independent blockchain, and cannot be accessed through wallets that connect to the Ethereum network. Only use wallets designed to work with the Celo network.
```

jmrossy (commented):
One suggestion and found a typo

timmoreton:
@nategraf Also worth updating the entry in the glossary

nategraf:
@timmoreton I've added those to the glossary in 498e49bb3cfd856f17a7aa9a899efa4385dd50a0.

nambrot:
Thank you for doing this, if you happen to have the bandwidth, it would be nice to adjust https://github.com/celo-org/celo-monorepo/blob/master/packages/docs/developer-resources/integrations/custody.md as well

packages/docs/GLOSSARY.md
```diff
@@ -48,6 +48,18 @@ An account that receives a transfer as part of epoch rewards to cover the cost o
 
 An open platform that makes financial tools accessible to anyone with a mobile phone.
 
+## CELO
+
+The ticker symbol for the [Celo native asset](#celo-native-asset), often written in capital letters to avoid confusion with references to the Celo protocol.
+
+## Celo Gold
+
+The deprecated name for the Celo native asset, which now is referred to simply as "Celo" or "CELO" when it might be ambiguous whether the reference is to the asset or the protocol.
```
jmrossy:
```suggestion
The deprecated name for the Celo native asset, which now is referred to simply as "Celo" or preferably "CELO".
```

jmrossy (approved):
LGTM, one suggestion

Issue #4158, opened 2020-06-20
User is not redirected to the “Settings” screen after updating the user name from the “Edit profile” option


Lss-Ankit:
**Frequency:** 100%
** Repro App version:** Celo Integration Wallet build v1.8.2 (39)
**Does not repro on:** App store build v1.8.0
**Repro on device:**iPhone XS max (13.1), iphone 6+ (12.4.5)

**Repro Steps:** 

1) Launch app and login with tested account
2) Tap on the hamburger menu. 
3) Tap on setting option. 
4) tap on edit profile option. 
5) tap on Edit name 
6) Update the username and tap on the return button on Keypad. 
7) Observe the user's stay on the Edit name screen. 
**Current Behavior:** User is not redirected to the “Settings” screen

**Expected Behavior:** Users should be redirected to the settings screen after updating username. 
**Impact:** Users need to put in manual efforts to see changes. 
**Attachment:** [EditNameIssue.mp4](https://drive.google.com/file/d/1Y-zGLch7ZL_AxQEhjW8vZFPcWZBQhXBc/view?usp=sharing) (Bug Observed at 14 sec)

tarikbellamine:
@i1skn please look out for this while finishing the Settings redesign

Lss-Ankit:
HI @i1skn  i have verified issue on latest Android internal build v1.1.0(1004294309) & iOS test flight build v1.1.0(21) and observed that issue is still reproduced. 
- User is not redirected to the “Settings” screen after updating the user name from the “Edit profile” option

Verified device: iPhone 7+ (13.3.1) 	, Google Pixel Xl (8.1)

Lss-Ankit:
HI @i1skn i have verified issue on latest using Test Flight build 1.2.0(25) and Android internal build Android internal build 1.2.0(1004294313) and observed that issue is still reproduced.

User is not redirected to the “Settings” screen after updating the user name from the “Edit profile” option
Verified device: iPhone 11 (14 iOS) , Google Pixel Xl (8.1)

Lss-Ankit:
Hi @tarikbellamine I have verified this issue using Android internal build v1.2.0(1004294314) on lower end devices and observed that issue is still reproduced.
**Repro on devices:** Samsung Galaxy S4(4.4.2) , Samsung Galaxy A3(5.0.2) ,  Samsung Galaxy A5(7.0)

Lss-Ankit:
[...]

Issue #2101, opened 2019-12-06
Users SBAT start geth multiple times without restarting app


annakaz:
### Expected Behavior

Geth can be started and restarted without having restart the app or returning gas fee errors

### Current Behavior

After starting geth a second time in one app session, view and state-changing functions both return a gas fee error ("0 is below minimum gasPrice"). The gasFee of 0 comes from [this line](https://github.com/celo-org/celo-blockchain/blob/master/contract_comm/gasprice_minimum/gasprice_minimum.go#L100
) as `FallbackGasPriceMinimum` , which is 0, is returned as the smart contract is not deployed. It is not clear why the smart contract is not deployed, as it gets an up to date current block. 

As a work around right now, the app requires users to restart the app if they are starting geth more than once in a session (ex: start by using Forno, then start geth, back to Forno, start geth agaun and restart. Or start using geth, turn on Forno, turn back on geth and restart)

annakaz:
Related to https://github.com/celo-org/celo-monorepo/issues/3142

PR #4161, opened 2020-06-20
[Wallet] [Redesign] Settings screen

### LINKED ISSUE #566: User is not redirected to the “Settings” screen after updating the user name from the “Edit profile” option
**Frequency:** 100%
** Repro App version:** Celo Integration Wallet build v1.8.2 (39)
**Does not repro on:** App store build v1.8.0

**Repro Steps:** 

1) Launch app and login with tested account
2) Tap on the hamburger menu. 
3) Tap on setting option. 
4) tap on edit profile option. 
5) tap on Edit name 
6) Update the username and tap on the return button on Keypad. 
7) Observe the user's stay on the Edit name screen. 
**Current Behavior:** User is not redirected to the “Settings” screen

**Expected Behavior:** Users should be redirected to the settings screen after updating username. 
**Impact:** Users need to put in manual efforts to see changes. 
**Attachment:** [EditNameIssue.mp4](https://drive.google.com/file/d/1Y-zGLch7ZL_AxQEhjW8vZFPcWZBQhXBc/view?usp=sharing) (Bug Observed at 14 sec)

### LINKED ISSUE #3956: App get crash when user taps on device back button after navigating to  "Licenses" page
**Frequency:** 100%
** Repro App version:** Play store v1.8.2
**Repro on device:** redmi note 6 pro Android (9.0), Samsung Galaxy M30s  (9.0)
**Does not repro on** Google Pixel XL (8.1), 
**Tested Account:** 4422379165 / Backup key 

**Repro Steps:** 
1) Launch app and login with tested account
2) Tap on cog icon click on Licenses option 
3) Tap on device back button once or twice and observe 
 
**Current Behavior:** App is getting crashed
**Expected Behavior:** Should get back to Celo home screen 

**Attachment:** 
1. Suddencrashinlicensesbackbutton.mp4 (Bug observe at 30 sec)
2. [celo_logs.txt](https://github.com/celo-org/celo-monorepo/files/4724101/celo_logs.2.txt)

i1skn:
### Description

This PR implements Settings screen, Edit Profile and refining Language and Currency screens (https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=3114%3A35541).

### Other changes

- Remove checkboxes and back chevron at the Language screen when app is launched.
- Reflect `h1` and `h2` from the design kit in the code

### Tested

iOS Simulator

### Related issues

- Part of https://github.com/celo-org/celo-monorepo/issues/3645
- Fixes https://github.com/celo-org/celo-monorepo/issues/3956
- Fixes https://github.com/celo-org/celo-monorepo/issues/4158

### Backwards compatibility

Yes

### Screenshots
simulator
C6NgRiEQonLfzhPnvD0e4ODXxVi5g6vW


packages/mobile/src/account/Profile.test.tsx
```diff
@@ -2,30 +2,36 @@ import * as React from 'react'
 import 'react-native'
 import { fireEvent, render } from 'react-native-testing-library'
 import { Provider } from 'react-redux'
+import { setName } from 'src/account/actions'
 import Profile from 'src/account/Profile'
-import { navigate } from 'src/navigator/NavigationService'
+// import { navigate } from 'src/navigator/NavigationService'
```
cmcewen: Remove?

packages/mobile/src/account/Settings.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+  getDevSettingsComp() {
+    const { devModeActive } = this.props
+
+    if (!devModeActive) {
+      return null
+    } else {
+      return (
+        <View style={styles.devSettings}>
+          {/* <View style={style.devSettingsItem}>
```
cmcewen: remove? or add a comment

packages/mobile/src/account/Settings.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+  toggleBackupState = () => {
+    this.props.toggleBackupState()
+  }
+
+  showDebugScreen = () => {
+    this.props.navigation.navigate(Screens.Debug)
+  }
+
+  getDevSettingsComp() {
```
cmcewen, line 150: probably worth breaking this out into a separate component, but naming it `renderDevSettings` would be a bit nicer (also i like this article https://dev.to/jcutrell/react-anti-pattern-renderthing-50nd)

packages/mobile/src/account/Settings.tsx
```diff
@@ -0,0 +1,344 @@
...
+      )
+    }
+  }
+
+  disableFornoMode = () => {
+    this.props.toggleFornoMode(false)
+    this.hideFornoSwitchOffWarning()
+  }
+
+  handleFornoToggle = (fornoMode: boolean) => {
```
cmcewen, line 201: @annakaz do you know if this is still required to prevent crashes? thought we might be okay now

packages/react-components/components/SettingsItem.tsx
```diff
@@ -1,52 +1,179 @@
...
-          </View>
+function Title({ value }: { value: string }) {
+  return <View style={[styles.left]}>{<Text style={styles.title}>{value}</Text>}</View>
+}
+
+type SettingsItemNoValueProps = {
+  title: string
+} & Omit<WrapperProps, 'children'>
+
+export function SettingsItemNoValue({ testID, title, onPress }: SettingsItemNoValueProps) {
```
cmcewen: i'm all for being explicit, but the `NoValue` in the name seems to imply that value could be an optional prop? i'm okay with this if it's truly ugly, but as a consumer it feels easier to just import one component as opposed to knowing which one i have to use

packages/mobile/src/language/Language.tsx
```diff
@@ -12,25 +12,31 @@ import { CustomEventNames } from 'src/analytics/constants'
...
-    const nextScreen = route.params?.nextScreen ?? Screens.JoinCelo
     dispatch(setLanguage(code))
-    if (nextScreen === 'GO_BACK') {
-      navigateBack()
-    } else {
-      navigate(nextScreen)
-    }
+    // Wait for next frame before navigating back
+    // so the user can see the changed selection briefly
+    requestAnimationFrame(() => {
```
cmcewen, line 42: i think this technically goes on the next tick, so i wouldn't be surprised if in the prod build it's not visible because it happens so fast. i would recommend setting an explict ms timeout if you want to be sure it's seen

packages/mobile/src/localCurrency/SelectLocalCurrency.tsx
```diff
@@ -11,51 +15,52 @@ import { navigateBack } from 'src/navigator/NavigationService'
 
 const DEFAULT_CURRENCY_CODE = LocalCurrencyCode.USD
 
-const keyExtractor = (item: LocalCurrencyCode) => item
+// const keyExtractor = (item: LocalCurrencyCode) => item
```
cmcewen: stray comment

packages/mobile/src/localCurrency/SelectLocalCurrency.tsx
```diff
@@ -11,51 +15,52 @@ import { navigateBack } from 'src/navigator/NavigationService'
...
-      />
-    ),
-    [selectedCurrencyCode]
-  )
+  function onSelect(code: string) {
+    dispatch(selectPreferredCurrency(code as LocalCurrencyCode))
 
+    // Wait for next frame before navigating back
+    // so the user can see the changed selection briefly
+    requestAnimationFrame(() => {
```
cmcewen, line 33: same comment about waiting

packages/mobile/src/language/Language.tsx
```diff
@@ -12,25 +12,31 @@ import { CustomEventNames } from 'src/analytics/constants'
...
+      if (fromSettings) {
```
cmcewen, line 43: i think the nextScreen pattern is more generic and gives more info about the behavior on selection - is there any reason to switch it?

packages/mobile/src/navigator/types.tsx
```diff
@@ -25,12 +25,10 @@ type NestedNavigatorParams<ParamList> = {
 
 // tslint:disable-next-line: interface-over-type-literal
 export type StackParamList = {
-  [Screens.Account]: undefined
-  [Screens.Analytics]: undefined
   [Screens.BackupComplete]: undefined
   [Screens.BackupIntroduction]:
     | {
-        fromAccountScreen?: boolean
+        fromSettings?: boolean
```
cmcewen, line 32: i personally prefer using `nextScreen` as opposed to switching to fromSettings

packages/react-components/components/SettingsItem.tsx
cmcewen: maybe SettingsItem has an enum type?

cmcewen (commented):
lots of changes! i added an item to our redesign to make sure to remove unused fonts, but just wanted to mention here too 😄

packages/mobile/src/account/Settings.tsx
i1skn: I'm not sure why it was commented in the first place. I just moved `dev` related things untouched. @jmrossy I see you've commented it some time ago, can we uncomment it now?

packages/mobile/src/language/Language.tsx
i1skn: Adding a special case `GO_BACK` to Screens did not feel right for me, so I've tried to replace it with `if nextScreeen is defined - use it, otherwise navigateBack`. It is rather a proposal, so if folks would like it - I'm happy to revert.

packages/mobile/src/navigator/types.tsx
i1skn: The same, as above

packages/mobile/src/language/Language.tsx
i1skn: I piggy back this approach from Currency screen, where @jeanregisser has implemented this way. I think it should be still visible on faster phones, but can be wrong. Also navigation is done via animation as well, so screen does not disappear right away. WDYT?

packages/react-components/components/SettingsItem.tsx
i1skn: Good idea, I will merge `SettingsItemNoValue` and `SettingsItemTextValue` into one component, but would keep `SettingsItemSwitch` and `SettingsItemInput` separate. WDYT?

packages/mobile/src/account/Settings.tsx
i1skn: It make sense for me, but I would do it in a separate PR, not as a part of Redesign.

aaronmgdr:
On the change name screen. what about having the keyboard's return key be labeled "done".

packages/mobile/src/account/Profile.test.tsx
```diff
@@ -2,30 +2,35 @@ import * as React from 'react'
...
-    <Provider store={createMockStore({})}>
-      <Profile {...getMockStackScreenProps(Screens.Profile)} />
-    </Provider>
-  )
-}
-
 describe('Profile', () => {
   it('renders correctly', () => {
-    const { toJSON } = render(profileFactory())
+    const store = createMockStore({})
```
jmrossy: Two nits: no need to pass in `{}` and no need to remake the store multiple times (line 23 below). The tests can share this value

packages/mobile/src/account/Settings.tsx
```diff
@@ -76,36 +91,32 @@ const mapDispatchToProps = {
...
+  fornoSwitchOffWarning: boolean
+  promptFornoModal: boolean
 }
 
 export class Account extends React.Component<Props, State> {
-  state: State = {
-    version: DeviceInfo.getVersion(),
+  componentDidMount() {
+    const promptFornoModal = this.props.route.params?.promptFornoModal
+    if (promptFornoModal) {
```
jmrossy: Nit: no need for the if right?

packages/mobile/src/account/Settings.tsx
jmrossy: It's commented because it broke a while back but this is something we'd like to re-enable

packages/mobile/src/account/Settings.tsx
```diff
@@ -227,101 +201,123 @@ export class Account extends React.Component<Props, State> {
...
+    this.props.toggleFornoMode(true)
+    navigateBack()
+  }
+
+  hidePromptModal = () => {
+    this.props.toggleFornoMode(false)
+    navigateBack()
+  }
+
+  goToTerms() {
```
jmrossy: Can we keep to a common convention for these handler names? Ideally they should begin with `on`

jmrossy: Maybe not your code though, I understand

packages/mobile/src/language/Language.tsx
jmrossy: I'm fine either way personally :)

packages/mobile/src/localCurrency/SelectLocalCurrency.tsx
```diff
@@ -11,51 +15,50 @@ import { navigateBack } from 'src/navigator/NavigationService'
...
-      <SelectionOption
-        text={code}
-        onSelect={onSelect}
-        isSelected={code === selectedCurrencyCode}
-        data={code}
-      />
-    ),
-    [selectedCurrencyCode]
-  )
+  function onSelect(code: string) {
```
jmrossy: ditto arrow

jmrossy: I thought there was a lint rule for this but maybe not

packages/mobile/src/localCurrency/SelectLocalCurrency.tsx
```diff
@@ -11,51 +15,50 @@ import { navigateBack } from 'src/navigator/NavigationService'
...
   return (
-    <FlatList
-      style={styles.container}
-      data={LOCAL_CURRENCY_CODES}
-      extraData={selectedCurrencyCode}
-      renderItem={renderItem}
-      keyExtractor={keyExtractor}
-    />
+    <ScrollView style={styles.container}>
+      <SafeAreaView>
```
jmrossy, line 52: Don't we usually put SAV outside the scrollview?

packages/react-components/components/SettingsItem.tsx
```diff
@@ -1,52 +1,170 @@
...
+} & Omit<BaseProps, 'onPress'>
+
+export function SettingsItemInput({
+  testID,
+  title,
+  onValueChange,
+  value,
+  placeholder,
+}: SettingsItemInputProps) {
+  function onFocus() {
```
jmrossy: ditto

jmrossy (commented):
Mostly looks good! Just some small comments

packages/mobile/src/localCurrency/SelectLocalCurrency.tsx
i1skn: Actually, @jeanregisser  pointed me out against it - otherwise we see padding at the bottom, when scrolling, which is not native experience.

packages/mobile/src/localCurrency/SelectLocalCurrency.tsx
```diff
@@ -11,51 +15,50 @@ import { navigateBack } from 'src/navigator/NavigationService'
...
+  }
   return (
-    <FlatList
-      style={styles.container}
-      data={LOCAL_CURRENCY_CODES}
-      extraData={selectedCurrencyCode}
-      renderItem={renderItem}
-      keyExtractor={keyExtractor}
-    />
+    <ScrollView style={styles.container}>
```
jeanregisser, line 51: Though we have few items today, why removing `FlatList`?

packages/mobile/src/language/Language.tsx
```diff
@@ -54,14 +61,21 @@ export default function Language({ route }: Props) {
...
 }
 
+export function languageScreenOptions({ route }: RouterProps) {
+  if (route.params?.fromSettings) {
+    return headerWithBackButton
+  } else {
+    return emptyHeader
+  }
+}
+
```
jeanregisser: nit: we started using `navigationOptions` attached to the component for easier handling (no import/export needed)

https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/verify/VerificationEducationScreen.tsx#L104

packages/mobile/src/account/Settings.tsx
```diff
@@ -76,36 +91,32 @@ const mapDispatchToProps = {
   setAnalyticsEnabled,
   toggleBackupState,
   devModeTriggerClicked,
+  setLockWithPinEnabled,
+  toggleFornoMode,
+}
+
+interface State {
+  fornoSwitchOffWarning: boolean
+  promptFornoModal: boolean
```
jeanregisser: nit: `promptFornoModal` state is redundant with the route params and can be removed. It will simplify `componentDidMount` too

packages/mobile/src/account/Settings.tsx
jeanregisser: Cool, that would be awesome for testing from a clean unverified state 👍

packages/react-components/components/SettingsItem.tsx
```diff
@@ -1,52 +1,170 @@
...
+    setInputColor(colors.dark)
+  }
+  function onBlur() {
+    setInputColor(colors.gray4)
   }
+
+  const [inputColor, setInputColor] = React.useState(colors.gray4)
+  return (
+    <>
```
jeanregisser: nit: Fragment is not needed here.

jeanregisser (commented):
Looks great, nice to see obsolete thing deleted! 🎉 Just some small comments.

i1skn:
@jeanregisser @jmrossy I've addressed the feedback, PTAL :)

jmrossy (approved):

packages/mobile/src/account/Settings.tsx
annakaz: I think @etuleu was looking into this issue https://github.com/celo-org/celo-monorepo/issues/2101 , but I have not checked whether it reproduces recently

PR #4162, opened 2020-06-20
[Wallet][ReDesign] Verification start screen

jeanregisser:
### Description

This implements the [verification start screen redesign](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=5404%3A450).

Note: this PR depends on #4152, please merge that one first.

### Other changes

N/A

### Tested

verification-education2

### Related issues

- Epic #3645

### Backwards compatibility

Yes.


packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -1,205 +1,156 @@
...
+type Props = ScreenProps
 
-interface State {
-  isModalVisible: boolean
-}
+function VerificationEducationScreen({ route, navigation }: Props) {
+  const showSkipDialog = route.params?.showSkipDialog || false
+  const [showLearnMoreDialog, setShowLearnMoreDialog] = useState(false)
+  const { t } = useTranslation(Namespaces.onboarding)
+  const userBalance = useSelector((state) => state.stableToken.balance)
```
cmcewen, line 38: i think we've been a little inconsistent overall about selector functions (defining inline vs sharing) - i don't have a strong preference either way but think we should be consistent

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -1,205 +1,156 @@
...
+  const balanceIsSufficient = isUserBalanceSufficient(userBalance, VERIFICATION_FEE_ESTIMATE)
```
cmcewen, line 39: i don't think this is quite expensive enough to be worth memoizing, just thought this was an interesting read (https://react-redux.js.org/next/api/hooks#using-memoizing-selectors) and we could reselect here if we wanted

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -1,205 +1,156 @@
...
-class VerificationEducationScreen extends React.Component<Props, State> {
-  static navigationOptions = nuxNavigationOptions
-
-  state: State = {
-    isModalVisible: false,
-  }
-
-  onPressLearnMore = () => {
-    navigate(Screens.VerificationLearnMoreScreen)
+  function onPressStart() {
```
cmcewen, line 44: similar question on fat arrows vs function keyword - any preference?

cmcewen (approved):
love all those lines of code deleted 😄

packages/mobile/src/verify/VerificationEducationScreen.tsx
jeanregisser: Yep. We should probably use selectors everywhere.

packages/mobile/src/verify/VerificationEducationScreen.tsx
jeanregisser: Yes thanks well aware of this 👍

packages/mobile/src/verify/VerificationEducationScreen.tsx
jeanregisser: no strong preference either, the `function` keyword makes it arguably slightly more readable, but fat arrow makes it easy to wrap with `useCallback` for instance 😄

PR #4165, opened 2020-06-22
Add more timing and error information

nambrot:
Some more

### DIFF
--- a/packages/mobile/src/transactions/send.ts
+++ b/packages/mobile/src/transactions/send.ts
@@ -31,6 +31,7 @@ const ALWAYS_FAILING_ERROR = 'always failing transaction'
 const KNOWN_TX_ERROR = 'known transaction'
 
 const getLogger = (tag: string, txId: string) => {
+  let startTime = new Date().getTime()
   return (event: SendTransactionLogEvent) => {
     switch (event.type) {
       case SendTransactionLogEventType.Confirmed:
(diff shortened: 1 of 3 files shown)


packages/mobile/src/transactions/send.ts
```diff
@@ -41,27 +42,48 @@ const getLogger = (tag: string, txId: string) => {
...
         break
       case SendTransactionLogEventType.ReceiptReceived:
         Logger.debug(
           tag,
           `Transaction id ${txId} received receipt: ${JSON.stringify(event.receipt)}`
         )
-        CeloAnalytics.track(CustomEventNames.transaction_send_gas_receipt, { txId })
+        CeloAnalytics.track(CustomEventNames.transaction_send_gas_receipt, {
+          txId,
+          duration: new Date().getTime() - startTime,
```
annakaz: 👍

annakaz (approved):
LGTM

packages/mobile/src/transactions/send.ts
```diff
@@ -31,6 +31,7 @@ const ALWAYS_FAILING_ERROR = 'always failing transaction'
 const KNOWN_TX_ERROR = 'known transaction'
 
 const getLogger = (tag: string, txId: string) => {
+  let startTime = new Date().getTime()
```
jmrossy: I think `Date.now()` is a shorter-hand version of this same thing. Avoids need to all the `new` keywords

packages/mobile/src/transactions/send.ts
```diff
@@ -41,27 +42,48 @@ const getLogger = (tag: string, txId: string) => {
...
+          txId,
+          error: event.error.message,
+          duration: new Date().getTime() - startTime,
+        })
         break
       case SendTransactionLogEventType.Exception:
         Logger.error(tag, `Transaction Exception caught ${txId}: `, event.error)
+        CeloAnalytics.track(CustomEventNames.transaction_exception, {
+          txId,
+          error: event.error.message,
```
jmrossy, line 84: Repeating the feedback I left on Anna's PR. Are we using analytics for errors now? That's what Sentry was meant for and seems better suited to it.

annakaz: We're still using Sentry for errors, but we are also including them in analytics when we want to understand the error in context of what was going on in the app when the error was thrown

Discussion here- https://github.com/celo-org/celo-monorepo/pull/4144#discussion_r443681952

packages/mobile/src/transactions/send.ts
nambrot: Woah, awesome!

PR #4168, opened 2020-06-22
Update deployment configurations for Baklava

nategraf:
### Description

This PR contains updates to the network and deployment configurations for the Baklava testnet.

### Related issues

- Related: https://github.com/celo-org/celo-blockchain/issues/1054


.env.baklavastaging
```diff
@@ -13,27 +13,27 @@ CLUSTER_DOMAIN_NAME="celo-testnet"
...
-BLOCKSCOUT_DOCKER_IMAGE_TAG="72dacea44b1779e5b15069d991453772d4e8f032"
+BLOCKSCOUT_DOCKER_IMAGE_TAG="09add19279a01388295f63df64223264d8a288ea"
 BLOCKSCOUT_WEB_REPLICAS=3
-BLOCKSCOUT_DB_SUFFIX=3
+BLOCKSCOUT_DB_SUFFIX=4
 BLOCKSCOUT_SUBNETWORK_NAME="Baklava Staging"
 
 CELOSTATS_SERVER_DOCKER_IMAGE_REPOSITORY="gcr.io/celo-testnet/celostats-server"
-CELOSTATS_SERVER_DOCKER_IMAGE_TAG="70bbe4d9bd4015ddcb9adfd5ea0e205a043e87c9"
+CELOSTATS_SERVER_DOCKER_IMAGE_TAG="master"
```
nambrot: Is there a reason why we are switching to a master tag here vs. commits? I think for predictable deploys, the latter seem preferrable?

packages/protocol/migrationsConfig.js
```diff
@@ -330,29 +331,53 @@ const NetworkConfigs = {
...
     },
   },
   baklava: {
-    blockchainParameters: {
-      minimumClientVersion: {
-        major: 0,
-        minor: 10,
-        patch: 0,
+    election: {
+      frozen: false,
```
nambrot, line 340: I know we don't want to have frozen elections, but IMO to keep baklava as close to Mainnet as possible, what do you think about keeping them all frozen and have a governance proposal to unfreeze? I know that it shouldn't practically make a different, but you never know?

packages/protocol/migrationsConfig.js
```diff
@@ -330,29 +331,53 @@ const NetworkConfigs = {
...
       approvalStageDuration: 4 * HOUR,
       referendumStageDuration: DAY,
       executionStageDuration: WEEK,
+      participationBaseline: 1 / 200, // Very low participation requirement given its a testnet.
+      participationBaselineFloor: 1 / 100,
+      participationBaselineUpdateFactor: 1 / 5,
+      participationBaselineQuorumFactor: 1,
+    },
+    governanceApproverMultiSig: {
+      signatories: [network.from],
```
nambrot: I assume we are ok with anyone who has access to the MNEMONIC to be able to effectively shut down the network?

nambrot (approved):

packages/protocol/migrationsConfig.js
nategraf: We want the elections enabled immediately so folks can start joining the network as validators. I would also like to avoid the operational overhead of running governance proposals right away if possible. Given that we've already tested the code for starting a network with elections disabled on both the latest instance of Baklava and Mainnet, it doesn't seem like the extra coverage will make enough of a difference to warrant the work required.

.env.baklavastaging
nategraf: Good point. I've changed it to be the latest commit on `master` instead of the literal string `master`

packages/protocol/migrationsConfig.js
nategraf: I've gone ahead and created a new account that is loaded into Google Cloud IAM where we can share. Additionally, I went back in git history to add back the signers from RC0, which are still active.

PR #4175, opened 2020-06-23
Rebrand gold token

kphed:
### Description

Initial smart contract changes to reflect the cGLD to CELO rebranding (limited the scope of this PR's changes to `GoldToken` so that it's manageable for the reviewer).

- Updated the token-related variable values in the `GoldToken` contract
    - `NAME`: Changed `Celo Gold` to `Celo`
    - `SYMBOL`: Changed `cGLD` to `CELO`
- Updated `GoldToken`-related file names to `CeloToken`
    - Renamed files are in [this](https://github.com/celo-org/celo-monorepo/commit/b067ddd175298c701e653b4b56a1569f746bd977) commit
- Changed instances of `(g|G)oldToken` to `(c|C)eloToken` in all files
- Change instances of `(c|C)GLD` to `(c|C)ELO` in all files

### Other changes

None at the moment. I'll continue making changes until the CI build passes and will update the above.

### Tested

Going to defer test-running to CI since the test suite is causing my machine to bug out.

### Backwards compatibility

Many of the other Celo packages may attempt to interact with the files and code in the changed contracts by the defunct name. I'll start tackling those as well.


packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -445,7 +445,7 @@ contract Reserve is IReserve, Ownable, Initializable, UsingRegistry, ReentrancyG
     ISortedOracles sortedOracles = ISortedOracles(sortedOraclesAddress);
     uint256 reserveGoldBalance = getUnfrozenReserveGoldBalance();
     uint256 stableTokensValueInGold = 0;
-    FixidityLib.Fraction memory cgldWeight = FixidityLib.wrap(assetAllocationWeights["cGLD"]);
+    FixidityLib.Fraction memory celoWeight = FixidityLib.wrap(assetAllocationWeights["CELO"]);
```
nategraf, line 448: This change will require a governance proposal to call `setAssetAllocations`.

packages/protocol/contracts/common/UsingRegistry.sol
```diff
@@ -42,7 +42,7 @@ contract UsingRegistry is Ownable {
     abi.encodePacked("FeeCurrencyWhitelist")
   );
   bytes32 constant FREEZER_REGISTRY_ID = keccak256(abi.encodePacked("Freezer"));
-  bytes32 constant GOLD_TOKEN_REGISTRY_ID = keccak256(abi.encodePacked("GoldToken"));
+  bytes32 constant GOLD_TOKEN_REGISTRY_ID = keccak256(abi.encodePacked("CeloToken"));
```
nategraf, line 45: Changing the REGISTRY_ID will require a governance proposal to additionally add the new registry ID, pointing to the existing CeloToken (i.e. GoldToken) proxy.

nategraf:
Thanks for opening this PR! Overall this looks good. The one difficulty we will need to work out is that the changes to the registry I and the reserve symbol will require governance proposal to be submitted in the same governance proposal that upgrades the contracts to make sure everything keeps working. If that is required, it can be done but will need some thought to make sure it is executed correctly. If there is a way to avoid this, that would be better.

kphed:
> Thanks for opening this PR! Overall this looks good. The one difficulty we will need to work out is that the changes to the registry I and the reserve symbol will require governance proposal to be submitted in the same governance proposal that upgrades the contracts to make sure everything keeps working. If that is required, it can be done but will need some thought to make sure it is executed correctly. If there is a way to avoid this, that would be better.

No problem Nate, thanks for making the time to go through it. I'd be happy to work with you on making the change happen. There are also other cases of `Celo Gold` and `cGLD` in the contracts but I'm inclined to hold off on those changes since it seems like we may have our hands full with what you mentioned above. What do you think?

barbaraliau:
@nategraf @kphed Is this PR still relevant?

PR #4179, opened 2020-06-23
[Wallet][ReDesign] Verification input screen

jeanregisser:
### Description

This implements the [verification input redesign](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=5404%3A1009).

### Other changes

N/A

### Tested

enter-verif-codes

### Related issues

- Epic #3645

### Backwards compatibility

Yes.


packages/mobile/src/navigator/types.tsx
```diff
@@ -186,7 +186,7 @@ export type StackParamList = {
     isFromScan?: boolean
   }
   [Screens.VerificationEducationScreen]: { showSkipDialog: boolean } | undefined
-  [Screens.VerificationInputScreen]: undefined
+  [Screens.VerificationInputScreen]: { showHelpDialog: boolean } | undefined
```
jmrossy, line 189: Why is this a nav param?

jmrossy (approved):
Looks great! Just one question

packages/mobile/src/navigator/types.tsx
jeanregisser: This was needed to communicate between the NavBar Help button and the screen itself.
In retrospect I could have embedded the Dialog in the button 🤔 
Anyway I think we should probably change those dialogs to dedicated screens. This way they would not be unmounted abruptly when navigating away from the screen that hosted it.

PR #4181, opened 2020-06-23
[DRAFT] Fix docker image vulnerability

codyborn:
### Description

Switches to use alpine base image to avoid the following vulnerabilities:
https://security-tracker.debian.org/tracker/CVE-2017-17458
https://security-tracker.debian.org/tracker/CVE-2019-19816

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


codyborn:
Unfortunately, Alpine also suffers from this vulnerability: https://security.netapp.com/advisory/ntap-20190625-0005/

PR #4182, opened 2020-06-23
Rename Celo Gold / cGLD to CELO

henry-zhang-bohan:
Rename the following to CELO:
- Celo Gold (cGLD)
- Celo Gold
- cGLD


packages/cli/src/commands/validator/requirements.ts
```diff
@@ -3,7 +3,7 @@ import { printValueMap } from '../../utils/cli'
 
 export default class ValidatorRequirements extends BaseCommand {
   static description =
-    'List the Locked Gold requirements for registering a Validator. This consists of a value, which is the amount of Celo Gold that needs to be locked in order to register, and a duration, which is the amount of time that Gold must stay locked following the deregistration of the Validator.'
+    'List the Locked Gold requirements for registering a Validator. This consists of a value, which is the amount of CELO that needs to be locked in order to register, and a duration, which is the amount of time that Gold must stay locked following the deregistration of the Validator.'
```
timmoreton:
```suggestion
    'List the Locked Gold requirements for registering a Validator. This consists of a value, which is the amount of CELO that needs to be locked in order to register, and a duration, which is the amount of time that CELO must stay locked following the deregistration of the Validator.'
```

packages/contractkit/src/wrappers/GasPriceMinimum.ts
```diff
@@ -14,7 +14,7 @@ export interface GasPriceMinimumConfig {
 export class GasPriceMinimumWrapper extends BaseWrapper<GasPriceMinimum> {
   /**
    * Query current gas price minimum in gGLD.
```
timmoreton:
```suggestion
   * Query current gas price minimum in CELO.
```

packages/docs/SUMMARY.md
```diff
@@ -19,7 +19,7 @@
 - [Validator Setup FAQ](getting-started/validator-troubleshooting-faq.md)
 - [Glossary](GLOSSARY.md)
 
-## Celo Gold Holder Guide
+## CELO Holder Guide
```
timmoreton, line 22: This is going to change the URL. We need redirects for the old URLs.

packages/docs/celo-codebase/protocol/proof-of-stake/locked-gold-rewards.md
```diff
@@ -6,7 +6,7 @@ Rewards to Locked Gold are totally independent from validator and validator grou
 
 ![](https://storage.googleapis.com/celo-website/docs/locked-gold-rewards.jpg)
 
-First, an on-target reward rate is determined. The protocol has a target for the proportion of circulating Celo Gold that is locked and used for voting, and adjusts the reward rate to increase or reduce the attractiveness of locking up additional supply. This aims to balance having sufficient liquidity for Celo Gold, while making it more challenging to buy enough Celo Gold to meaningfully influence the outcome of a validator election.
+First, an on-target reward rate is determined. The protocol has a target for the proportion of circulating CELO that is locked and used for voting, and adjusts the reward rate to increase or reduce the attractiveness of locking up additional supply. This aims to balance having sufficient liquidity for CELO, while making it more challenging to buy enough CELO to meaningfully influence the outcome of a validator election.
 
 Adjusting the on-target reward rate to account for under- or over-spending against the target schedule gives a baseline reward, essentially the percentage increase for a unit of Locked Gold voting for a group eligible for rewards.
```
timmoreton: ref to gold

packages/docs/celo-codebase/protocol/proof-of-stake/locked-gold.md
```diff
@@ -1,26 +1,26 @@
...
 
-To participate in validator elections, users must first make a transfer of Celo Gold to the `LockedGold` smart contract.
+To participate in validator elections, users must first make a transfer of CELO to the `LockedGold` smart contract.
 
 ## Concurrent Use of Locked Gold
 
-Locking up Celo Gold guarantees that the same asset is not used more than once in the same vote. However every unit of Locked Gold can be deployed in several ways at once. Using an amount for voting for a validator does not preclude that same amount also being used to vote for a governance proposal, or as a stake at the same time. Users do not need to choose whether to have to move funds from validator elections in order to vote on a governance proposal.
+Locking up CELO guarantees that the same asset is not used more than once in the same vote. However every unit of Locked Gold can be deployed in several ways at once. Using an amount for voting for a validator does not preclude that same amount also being used to vote for a governance proposal, or as a stake at the same time. Users do not need to choose whether to have to move funds from validator elections in order to vote on a governance proposal.
```
timmoreton:
```suggestion
Locking up CELO guarantees that the same asset is not used more than once in the same vote. However every unit of Locked CELO can be deployed in several ways at once. Using an amount for voting for a validator does not preclude that same amount also being used to vote for a governance proposal, or as a stake at the same time. Users do not need to choose whether to have to move funds from validator elections in order to vote on a governance proposal.
```

timmoreton (commented):
LGTM - few minor suggestions!

packages/docs/SUMMARY.md
henry-zhang-bohan: changed the links in commit `9598f9cc8df7103408ba1d38e8dda28359c711c9`

henry-zhang-bohan: But how do we add redirects?

henry-zhang-bohan: addressed!

packages/docs/GLOSSARY.md
```diff
@@ -176,7 +176,7 @@ A device or computer running the Celo Blockchain software that keeps typically o
 
 [CELO](#celo-native-asset) balances held in escrow at the Locked Gold contract for the account that deposited it there. This permits that balance to be used for voting in validator elections, governance proposals, and to meet staking requirements for registering a validator or validator group.
```
marekolszewski, line 177: Consider just calling this anchor #CELO. My understanding of the long-form name is that it's useful when just using CELO would be ambiguous. E.g. Stablecoins on the Celo platform are backed by the Celo native asset.

marekolszewski: Also, consider mentioning here that CELO was previously called Celo Gold (cGLD).

marekolszewski (commented):
Left two comments.

packages/docs/GLOSSARY.md
henry-zhang-bohan: This is now addressed in the paragraph below. There is an entry on CELO, one on Celo Gold, and another on Celo Native Asset -- I think they should be sufficient.

marekolszewski (approved):

timmoreton:
Checked the redirects. Per my understanding of how they work LGTM!

gastonponti:
the `contractkit-test` from circle ci is failing because you have to run `yarn docs` in the package/contractkit folder

PR #4184, opened 2020-06-23
Disable check for committing docs

yorhodes:
### Description

Disable circleCI check for docs updates. Will require contributor to explicitly build docs.


gastonponti:
I think that this kind of help us to maintain the docs.
I know it's kind of annoying to push and to check 20 minutes later that the only that failed were the docs, but we could consider make it a pre-push hook warning/error, and then remove it from circle.

barbaraliau:
Where did we land on this?

PR #4185, opened 2020-06-24
[Wallet Redesign] ImportContactsScreen redesign

### LINKED ISSUE #4178: User is not redirect back to "Connect Contacts" screen from "Verifying + <Phone number>" screen after tapping on < Arrow from Connect Contacts screen.
**Frequency:** 100%
**Repro App version:** Celo Integration Wallet build v1.8.2 (39)
**Repro on device:** Iphone 11(13.3)
**Repro Steps:** 
1) Launch the app 
2)  Navigate to Verify Phone number screen 
3) Copy & paste the Three verification code.  
4) Make sure once after Submitting the codes user navigates to Almost done and later to Connect Contacts screen
5) Tap on < arrow from Connect contact screen. 
6) Observe user stuck on "Verifying + <Phone number>" screen

**Current Behavior:**  User is not redirect back to "Connect Contacts" screen 
**Expected Behavior:**  User is should redirect back to "Connect Contacts" screen 
**Investigation:** User cannot navigate from connect contacts page to connect contacts page by tapping < arrow
 **Impact** User is blocked to navigate back to Connect Contacts screen
**Attachment:** [ IOS_VerificationFlowIssue.Mp4](https://drive.google.com/file/d/1eY70G82uOS7ZNvR1_tiwfY9w5EBSNnPj/view?usp=sharing) (Bug Observed at 06Sec)

tarikbellamine:
### Description
Redesign of ImportContactsScreen

### Other changes
Refactored to use hooks

### Tested
Yes

### Related issues
Part of #3645
Closes #4178

### Backwards compatibility
Yes

IMG_4273

IMG_4274

IMG_4275

IMG_4276


packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
-    const isSkipHidden = status === ImportContactsStatus.Done
-
-    return (
-      <SafeAreaView style={styles.container}>
-        <ScrollView contentContainerStyle={styles.scrollContainer}>
-          {status <= ImportContactsStatus.Stopped && (
+  return (
+    <SafeAreaView style={styles.container}>
+      <ScrollView contentContainerStyle={styles.scrollContainer}>
+        <View>
```
jeanregisser: nit: because we use fullscreen layout for onboarding screens, on iOS this makes the content collide with the navigation bar content when you overscroll,



One way to prevent this is to start the scrollview under the navigation bar like done in:
https://github.com/celo-org/celo-monorepo/blob/aae2b147608ef1303df18b208c6778e01a22ccfd/packages/mobile/src/verify/VerificationEducationScreen.tsx#L68-L69

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
-  }
-
-  onToggleFindMeSwitch = (value: boolean) => {
-    this.setState({
-      isFindMeSwitchChecked: value,
-    })
-  }
-
-  onFinish = () => {
+  function onFinish() {
```
jeanregisser: nit: I think we agreed to use fat arrow for inline handlers.

jeanregisser (approved):
Looks great thanks! 👍

packages/mobile/locales/en-US/onboarding.json
```diff
@@ -4,12 +4,9 @@
     "body": "Adding your contacts makes it easy to send and request payments with your friends. Connecting does not send invites to your contacts.",
     "findSwitch": "Allow contacts to find you",
     "syncing": {
-      "loadingHeader": "Syncing contacts...",
-      "loadingStatus0": "Importing",
-      "loadingStatus1": "Processing {{total}} contacts",
-      "loadingStatus2": "Finding matches",
```
jmrossy, line 10: Why have we removed these? For slow phones, having more steps would be nice to show progress?

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
+  headerTitle: () => (
+    <HeaderTitleWithSubtitle
+      title={i18n.t('onboarding:verificationEducation.title')}
+      subTitle={i18n.t('onboarding:step', { step: '5' })}
+    />
+  ),
+  headerRight: () => (
+    <TopBarTextButton
+      title={i18n.t('global:skip')}
+      testID="VerificationEducationSkip"
```
jmrossy: Copy paste bug

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
 
+ImportContactsScreen.navigationOptions = ({ navigation }: ScreenProps) => ({
+  ...nuxNavigationOptionsNoBackButton,
+  headerTitle: () => (
+    <HeaderTitleWithSubtitle
+      title={i18n.t('onboarding:verificationEducation.title')}
+      subTitle={i18n.t('onboarding:step', { step: '5' })}
+    />
+  ),
+  headerRight: () => (
```
jmrossy: The skip button shouldn't show when matching is done.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
+    <TopBarTextButton
+      title={i18n.t('global:skip')}
+      testID="VerificationEducationSkip"
+      // tslint:disable-next-line: jsx-no-lambda
+      onPress={() => navigate(Screens.OnboardingSuccessScreen)}
```
jmrossy: This is missing a lot of the stuff we need to do onSkip right? Like cancelling the import

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
+      titleStyle={{ color: colors.goldDark }}
```
jmrossy: inline style

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
-                {t('contacts.syncing.successHeader', { matches: matchesFound })}
-              </Text>
-              <Text style={styles.h2Status}>{t('contacts.syncing.successStatus')}</Text>
-            </>
-          ) : (
-            <Text style={styles.h1Status}>{t('contacts.syncing.successStatus')}</Text>
-          )}
-        </View>
-      )
+  // If import has already been done, move user forward
```
jmrossy: When would this happen?

jmrossy (changes requested):
It seems like this breaks some functionality that was working in my initial implementation.
I'd strongly prefer we not convert things to functional components without a good reason, especially if it increases risks of regressions

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
 
-    let h2Text: string
-    if (status === ImportContactsStatus.Processing) {
-      h2Text = 'contacts.syncing.loadingStatus1'
-    } else if (status === ImportContactsStatus.Matchmaking) {
-      h2Text = 'contacts.syncing.loadingStatus2'
-    } else {
-      h2Text = 'contacts.syncing.loadingStatus0'
+  React.useEffect(() => {
+    const prevImportContactProgress = prevImportContactProgressRef.current
```
jmrossy, line 111: Does this work when the status has multiple steps? Or does the ref *always* have the first value of `importContactsProgress`?

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,187 @@
...
+  }, [])
 
-    let h2Text: string
-    if (status === ImportContactsStatus.Processing) {
-      h2Text = 'contacts.syncing.loadingStatus1'
-    } else if (status === ImportContactsStatus.Matchmaking) {
-      h2Text = 'contacts.syncing.loadingStatus2'
-    } else {
-      h2Text = 'contacts.syncing.loadingStatus0'
+  React.useEffect(() => {
```
jmrossy, line 109: Maybe I'm just a cranky old react developer, but do people feel like this ref + useEffect pattern is handling this scenario more elegantly than the class pattern? There are some things I like about hooks but this feels worse to me

jmrossy (commented):
Looking at the screenshots, it looks like we need some more padding btwn the main elements right? Why keep them cramped but then have a giant open space beneath? @coreycelo

packages/mobile/locales/en-US/onboarding.json
tarikbellamine: Discussed with @coreycelo. I do not feel strongly about this.

coreycelo: My understanding is we can no longer show progress (ie. 30 of 134 contacts) and only total (134 contacts) so we simplified it since there was less value add.

coreycelo:
> Looking at the screenshots, it looks like we need some more padding btwn the main elements right? Why keep them cramped but then have a giant open space beneath? @coreycelo

There's a few things here including that the iPhone is particularly long and we're generally optimizing for a smaller screen. Overall just trying to keep this matching formatting throughout onboarding. Agree it's a bit cramped. A quick measure looks like the margins may be a bit different than spec'd, but assuming this matches the rest of onboarding, would rather get to that in a polish session across onboarding.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
tarikbellamine: I certainly could be wrong, but it looked to me that `denyImportContacts` wasn't actually doing anything. It triggered an action that flipped `askedContactsPermission` in redux state to `true` but as far as I can tell `askedContactsPermission` is not used anywhere. I didn't want to delete without knowing the background. Do you have insight into the context there? 

On second thought, I can see why having `cancelImportContacts` occur on skip is useful for much slower phones. I will add that back. It happens so quickly on my phone it felt useless to have but need to get better at remembering that's not likely to be the norm.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
tarikbellamine: Unclear if it's possible in the wild. It was happening pretty frequently while testing so I built it as a catch-all. If you can't think of any instance where this would happen (I can't off top of my head), I am happy to remove it

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
tarikbellamine: I believe it does. We update the ref to the latest instance on every update (line 69)

packages/mobile/locales/en-US/onboarding.json
jmrossy: The importing step is while we literally import and the finding step is while we wait for matchmaking results from PGPNP. Given that they're distinct tasks and I already did the work to dispatch fine-grained status updates, I don't know why we would cut this.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
jmrossy: Yeah askedContactsPermission was used in navigator but looks like thats gone now, probably during the big nav update. 
I think that's still useful to track this though in case we want to show different UI for users that saw this and skipped vs never saw it at all.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
jmrossy: If it happened, it seems like a bug that we didn't correctly reset the import status before coming here.
It shouldn't ever happen cause you need to get verified before we import contacts? Maybe one edge case is the invite screen? We should just cut that crappy screen IMO haha

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
jmrossy: Ah now I see. so the initial setting of prevImportContactProgressRef on line 34 is not actually the previous. Is the current.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
tarikbellamine: I like this approach, thanks!

packages/mobile/locales/en-US/onboarding.json
tarikbellamine: Designs have been updated and I've added it back in. Processing happens too quickly for me to catch it in the screenshots

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
tarikbellamine: Gotcha, I will add `denyImportContacts` back into the `onSkip` stuff with the understanding that we may make use of it sometime in the future

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
tarikbellamine: I was bumping into it when using the dev verification toggle, which we don't need to worry about in prod.

I think all routes to the Invite screen are being cut off as part of the redesign so I will go ahead and remove this catch-all then.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
tarikbellamine: Yep that's right, similar to how prev props don't really exist on first render

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
tarikbellamine: This is an interesting point. I was working under the assumption that refactoring everything to use hooks was a matter of when not if, so I was doing it every chance I got.

@cmcewen might it be worthwhile to discuss our team philosophy on this?

packages/mobile/locales/en-US/onboarding.json
jmrossy: Yeah on most phones it'll be too fast. On my old Nexus 7 with 2000 contacts it shows for a while.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,222 @@
...
+  const importStatus = importContactsProgress.status
+  const totalContacts = importContactsProgress.total
+  const prevImportContactProgressRef = React.useRef(importContactsProgress)
+  const matchedContacts = useSelector(matchedContactsSelector)
+  const matchedContactsCount = Object.keys(matchedContacts).length
+  const dispatch = useDispatch()
+  const insets = useSafeArea()
+  const headerHeight = useHeaderHeight()
+
+  function renderStatusContainer() {
```
jmrossy: Nit: per our convention, these should be lambdas. Though I would consider pulling this out entirely into its own SFC at the top of the file.

packages/mobile/src/onboarding/contacts/ImportContactsScreen.tsx
```diff
@@ -1,227 +1,222 @@
...
+        </View>
+      </View>
+    )
   }
 
-  componentWillUnmount() {
-    if (this.onFinishTimeout) {
-      clearTimeout(this.onFinishTimeout)
-    }
+  function onFinish() {
```
jmrossy: ditto, lamdba

jmrossy (approved):
Looks good now, thanks! Two nits below

Issue #3985, opened 2020-06-04
Create Release Process, gather assets needed for release, general release work


i1skn:

tarikbellamine:
Epic over

Issue #3060, opened 2020-03-11
Address Tech Debt in Firebase (coupling w/ app, project structure, etc)


nityas:
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

i1skn:
Reference to an issue with upgrading react-native firebase library to v6 https://github.com/celo-org/celo-monorepo/issues/2571

tarikbellamine:
@nityas should this still be open?

ole-nov:
closed as fixed

PR #4198, opened 2020-06-24
[Wallet] Split mobile firebase projects and preparation for mainnet app

jeanregisser:
### Description

This PR starts using new firebase projects for the mobile app and services (`blockchain-api`, `notification-service`). For now the following Firebase projects were created:
- `celo-mobile-mainnet`
- `celo-mobile-alfajores`

The logic is to have one Firebase project per network as described in https://docs.google.com/document/d/1BgorLJ3Wlww0xnnPfPkDRc_4SMwR94bUxxyEmv9Sfgs/edit#heading=h.zd8hnexmsp8w

The wallet project was updated to use them too for the following configs:
- `packages/mobile/.env.alfajores`
- `packages/mobile/.env.mainnet`

Other configs/envs remain unchanged.

### Other changes

- The service account key used by `blockchain-api` to access the Firebase db has been removed now that it's deployed in the same Firebase project and has direct access to it.
- Consolidated bundle ID and app display name into the dot env files. This way the env file remain the single source of truth for this and both platforms use the same value.
- Added new `mainnet` tasks in Fastlane config.
- Deleted unused wallet envs (`.env.production`, `.env.staging`).
- mainnet config uses a different bundle ID (`co.clabs...`)

### Tested

- `blockchain-api` and `notification-service` were deployed into the 2 new projects.

### Related issues

- #3985
- #3060

### Backwards compatibility

No.


packages/blockchain-api/app.mainnet.yaml
```diff
@@ -0,0 +1,12 @@
...
+service: blockchain-api
+env_variables:
+  NODE_ENV: 'production'
+  DEPLOY_ENV: 'mainnet'
+  EXCHANGE_RATES_API: 'https://apilayer.net/api'
+  BLOCKSCOUT_API: 'https://rc1-blockscout.celo-testnet.org/graphql'
+  FIREBASE_PROJECT_ID: 'celo-mobile-mainnet'
+  # TODO Pull addresses from the build artifacts of the network in protocol/build
+  FAUCET_ADDRESS: '0x456f41406B32c45D59E539e4BBA3D7898c3584dA'
+  VERIFICATION_REWARDS_ADDRESS: '0xb4fdaf5f3cd313654aa357299ada901b1d2dd3b5'
```
annakaz: AFAIK the `VERIFICATION_REWARDS_ADDRESS` is not in use (or updated in alfajores vs alfajoresstaging etc) - would it make sense to hardcode this to `0x00000 .... 00verification` or add a comment to make it clear this does not need to be updated

packages/blockchain-api/deploy.sh
```diff
@@ -7,18 +7,17 @@ set -euo pipefail
 # -n: Name of the network, maps to App Engine 'service' (integration, production, etc.)
```
annakaz: Update to `mainnet` or `alfajores`?

packages/blockchain-api/deploy.sh
```diff
@@ -7,18 +7,17 @@ set -euo pipefail
...
     n) NETWORK="$OPTARG" ;;
-    p) PROJECT="$OPTARG" ;;
     *) echo "Unexpected option ${flag}" ;;
   esac
 done
 
 [ -z "$NETWORK" ] && echo "Need to set the NETWORK via the -n flag" && exit 1;
-[ -z "$PROJECT" ] && echo "Need to set the PROJECT via the -p flag" && exit 1;
+
+PROJECT="celo-mobile-${NETWORK}"
```
annakaz, line 20: Is it possible to throw an error if the project `celo-mobile-alfajoresstaging` for example is not yet set up?

packages/mobile/.env.mainnet
```diff
@@ -0,0 +1,16 @@
...
+# TODO
+SMS_RETRIEVER_APP_SIGNATURE=GH+4Okn6nOW
+# If FORNO_ENABLED_INITIALLY, local geth will not run initially. 
+# If toggled on, it will use DEFAULT_SYNC_MODE. See src/geth/consts.ts for more info
+FORNO_ENABLED_INITIALLY=false
+DEFAULT_SYNC_MODE=5
+DEV_SETTINGS_ACTIVE_INITIALLY=false
+FIREBASE_ENABLED=true
+SECRETS_KEY=production
+SHOW_TESTNET_BANNER=true
```
annakaz: No more testnet banner needed 🚀

packages/mobile/src/geth/networkConfig.ts
```diff
@@ -94,6 +95,15 @@ const networkConfigs: { [testnet: string]: NetworkConfig } = {
...
     signMoonpayUrl: signMoonpayUrlStaging,
   },
+  [Testnets.mainnet]: {
+    nodeDir: `.${Testnets.mainnet}`,
+    syncMode: DEFAULT_SYNC_MODE,
+    initiallyForno: FORNO_ENABLED_INITIALLY,
+    blockchainApiUrl: 'https://blockchain-api-dot-celo-mobile-mainnet.appspot.com/',
+    pgpnpUrl: pgpnpUrlProd,
+    pgpnpPubKey: pgpnpPubKeyProd,
+    signMoonpayUrl: signMoonpayUrlStaging, // TODO
```
annakaz, line 105: 👍  the moonpay auth function is still hosted in `celo-org-mobile` so a prod version will have to be made accessible

packages/blockchain-api/app.mainnet.yaml
jmrossy: Yeah its not used anymore. We can remove it or whatever's easiest.

packages/blockchain-api/src/config.ts
```diff
@@ -41,7 +38,8 @@ export const DEPLOY_ENV = (process.env.DEPLOY_ENV as string).toLowerCase()
 export const EXCHANGE_RATES_API = (process.env.EXCHANGE_RATES_API as string).toLowerCase()
 export const { EXCHANGE_RATES_API_ACCESS_KEY } = getSecrets(DEPLOY_ENV)
 export const BLOCKSCOUT_API = (process.env.BLOCKSCOUT_API as string).toLowerCase()
-export const FIREBASE_DB = process.env.FIREBASE_DB
+export const FIREBASE_PROJECT_ID = process.env.FIREBASE_PROJECT_ID
```
jmrossy, line 41: I wonder if there's a way for App Engine to give us this value so we don't need to pass it in via a config? Have you tried logging process.env to see what's already included?
Def not a blocker

packages/mobile/android/.gitignore
```diff
@@ -1,10 +1,5 @@
 # keys
-app/google-services.json
-app/src/staging/google-services.json
-app/src/integration/google-services.json
-app/src/alfajores/google-services.json
-app/src/debug/google-services.json
-app/src/pilot/google-services.json
+**/google-services*.json
```
jmrossy, line 2: 👍

jmrossy (approved):
Very nice!
A few optional comments below and please see Anna's comments above.

packages/blockchain-api/deploy.sh
jeanregisser: Good idea! I'll add a check for supported projects!

packages/mobile/.env.mainnet
jeanregisser: Yeah 💯

packages/blockchain-api/src/config.ts
jeanregisser: Yes I looked at this and it's available as `GOOGLE_CLOUD_PROJECT` but then thought, it's better to be explicit for when we test locally.
Thoughts?

packages/blockchain-api/deploy.sh
jeanregisser: Backtracking on this, `gcloud` naturally complains if the project doesn't exist. This is probably enough for now.

Example:
```
❯ ./deploy.sh -n pilot
Starting notification service deployment.
Copying over config file
Deploying to gcloud
ERROR: (gcloud.app.deploy) Permissions error fetching application [apps/celo-mobile-pilot]. Please make sure you are using the correct project ID and that you have permission to view applications on the project.
```

packages/mobile/src/geth/networkConfig.ts
jeanregisser: Yes let's sync today to do this 👍

packages/mobile/ios/celo.xcodeproj/project.pbxproj
```diff
@@ -5,11 +5,12 @@
 	};
 	objectVersion = 46;
 	objects = {
+
 /* Begin PBXBuildFile section */
 		00E356F31AD99517003FC87E /* celoTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 00E356F21AD99517003FC87E /* celoTests.m */; };
-		0317E786245B696200AAAC2E /* Inter-Regular.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 0317E783245B696200AAAC2E /* Inter-Regular.ttf */; };
-		0317E788245B696200AAAC2E /* Inter-Medium.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 0317E784245B696200AAAC2E /* Inter-Medium.ttf */; };
-		0317E78A245B696200AAAC2E /* Inter-SemiBold.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 0317E785245B696200AAAC2E /* Inter-SemiBold.ttf */; };
+		0317E786245B696200AAAC2E /* ../fonts/Inter-Regular.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 0317E783245B696200AAAC2E /* ../fonts/Inter-Regular.ttf */; };
```
i1skn, line 11: Why to change this?

packages/mobile/ios/celo.xcodeproj/project.pbxproj
```diff
@@ -78,14 +79,14 @@
...
 		BF61F27C85F8E9CE335BA142 /* Pods-celoTests.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-celoTests.release.xcconfig"; path = "Target Support Files/Pods-celoTests/Pods-celoTests.release.xcconfig"; sourceTree = "<group>"; };
 		C5952C9D168E4172BB52364C /* Hind-SemiBold.ttf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = 9; includeInIndex = 0; lastKnownFileType = unknown; name = "Hind-SemiBold.ttf"; path = "../fonts/Hind-SemiBold.ttf"; sourceTree = "<group>"; };
 		CCB1F49C3DAE62A0F72F43DF /* Pods-celo-celoTests.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-celo-celoTests.release.xcconfig"; path = "Target Support Files/Pods-celo-celoTests/Pods-celo-celoTests.release.xcconfig"; sourceTree = "<group>"; };
+		F19E0D1442F8460181626D12 /* Jost-Book.ttf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = 9; includeInIndex = 0; lastKnownFileType = unknown; name = "Jost-Book.ttf"; path = "../fonts/Jost-Book.ttf"; sourceTree = "<group>"; };
```
i1skn, line 88: I see it has been already added here: https://github.com/celo-org/celo-monorepo/commit/16f1da7bbc428810ce47ea8b2303164e4b130260

packages/mobile/ios/celo.xcodeproj/project.pbxproj
jeanregisser: It's Xcode doing its thing 😄

packages/mobile/ios/celo.xcodeproj/project.pbxproj
jeanregisser: Yes those will probably get resolved once I sync with master.

packages/mobile/ios/celo.xcodeproj/project.pbxproj
i1skn: I've renamed them in the Xcode directly, when it tried to do this 🙂

i1skn (approved):
LGTM! Thanks for the tremendous effort to sort this out!

PR #4201, opened 2020-06-24
[Wallet] Restore Account Key screen redesign

i1skn:
### Description

Redesigning Account Key screen in onboarding https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=5404%3A440. I've straggled to make multiline input to work as expected on both platform, when `placeholder` is multiline, but no luck. So I've decided to use short placeholder for now and later come back to this problem. This means, that instead of full 24 words as a placeholder you would see only a first couple of them followed by `...` (see Screenshots)

### Tested

Android and iPhone

### Related issues

- Related to https://github.com/celo-org/celo-monorepo/issues/3645

### Backwards compatibility

Yes

### Screenshots

simulator
XPlONPN7yZEDpKct8IL2QBZpxRipauge
pp0WKxKQPjOIRLQ0BRxJqLXHhZipEAD1
vnQTSxr89UcVzJHl7IISIoZ0SpeTzQz4


i1skn:
@coreycelo please take a look on the temporary proposed design changes if that's ok, thanks!

packages/mobile/src/import/ImportWallet.tsx
```diff
@@ -100,92 +109,93 @@ export class ImportWallet extends React.Component<Props, State> {
...
-            <Text style={fontStyles.semiBold}>{t('tip')}</Text>
-            {t('backupKeyTip')}
-          </Text>
-        </KeyboardAwareScrollView>
-
-        {isImportingWallet && (
-          <View style={styles.loadingSpinnerContainer} testID="ImportWalletLoadingCircle">
-            <ActivityIndicator size="large" color={colors.celoGreen} />
-          </View>
+      <HeaderHeightContext.Consumer>
```
jmrossy, line 125: Will other onboarding screen need these wrappers HeaderHeight and SafeArea wrappers? Wondering if we should just make a single wrapper component?

jmrossy (approved):
One optional comment but looks good to me.

coreycelo:
> @coreycelo please take a look on the temporary proposed design changes if that's ok, thanks!

@i1skn Cool. This is something that will confuse some users so it'd be good to get it in eventually, but makes sense. In the second screenshot after the video, the word "medal" is underlined. Is that an Android text field feature or part of the app?

packages/mobile/src/import/ImportWallet.tsx
i1skn: Well, tbh I've just copied from what @jeanregisser has implemented. I guess we could created a wrapper out of it if need it more. @jeanregisser wdyt?

i1skn:
> > @coreycelo please take a look on the temporary proposed design changes if that's ok, thanks!
> 
> @i1skn Cool. This is something that will confuse some users so it'd be good to get it in eventually, but makes sense. In the second screenshot after the video, the word "medal" is underlined. Is that an Android text field feature or part of the app?

It's Android feature.

Issue #4010, opened 2020-06-04
Handle TX failure/tx fee edge cases


annakaz:
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

cmcewen:
Transaction reverted by the EVM: https://sentry.io/organizations/celo/issues/1526588141/?project=1250733&query=is%3Aunresolved

cmcewen:
Could not decode reason: https://sentry.io/organizations/celo/issues/1648484432/?project=1250733&query=is%3Aunresolved&statsPeriod=14d

Issue #4008, opened 2020-06-04
Major flows (e.g. sending cUSD, onboarding, etc.) analytics is easily accessible


annakaz:
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

annakaz:
Closing this as ValoraAnalytics has what we need instrumented and can be viewed through dbt/metabase

Issue #4004, opened 2020-06-04
Mobile Analytics data is easily accessible


annakaz:
### Expected Behavior

- [ ]  Ability to look up a user ID and see associated logs

### Current Behavior

What is the current behavior?

PR #4209, opened 2020-06-25
[Wallet] Add balance and contact tracking

annakaz:
### Description

The new wallet KPIs require a few analytics we weren't already tracking. This PR adds balance and contact imports tracking, which is the final addition to be able to calculate the KPIs listed [here](https://docs.google.com/document/d/1G_XGpggW6cTBENcvs4v6XpaZzkTpST2D0kh31V9OVtU/edit). 

Note that transaction size analytics will be calculated using eksportisto data

### Tested

Not tested, no logic change- just adding analytics

### Related issues

- Related to epics #4008 and #4004 
- Fee tracking to help identify the extent of #4010 

### Backwards compatibility

Yes


packages/mobile/src/tokens/saga.ts
```diff
@@ -73,7 +73,14 @@ export function tokenFetchFactory({ actionName, token, actionCreator, tag }: Tok
       const tokenContract = yield call(getTokenContract, token)
       const balanceInWei: BigNumber = yield call([tokenContract, tokenContract.balanceOf], account)
       const balance: BigNumber = yield call(convertFromContractDecimals, balanceInWei, token)
-      CeloAnalytics.track(CustomEventNames.fetch_balance)
+      CeloAnalytics.track(
+        CustomEventNames.fetch_balance,
+        token === CURRENCY_ENUM.DOLLAR
+          ? {
+              dollarBalance: balance,
```
nambrot, line 80: balance is a big number, i think we want a string or number representation?

nambrot (commented):
Just a small comment

packages/mobile/src/tokens/saga.ts
nambrot: Also it might be nice to just have a balance field and then a separate currency field. Makes empty columns less prevalent

i1skn (approved):
LGTM

PR #4210, opened 2020-06-25
Upgrade oracles on mainnet & baklava to use OKCoin, improve upgrading process

tkporter:
### Description

* Includes oracle redeploy on new baklava and mainnet to use the OKCoin adapter and a newer commit
* Correctly sets the Azure subscription
* Only registers a static IP if one is not already registered (helps w/ upgrading full nodes without extensive Azure permissions)
* Only runs `az aks get-credentials ...` when needed (helps w/ deploying in general without extensive Azure permissions)
* Only sets key vault access policies if they are not already set (helps w/ upgrading oracles without extensive Azure permissions)

### Other changes

Changed some configurations for the oracle as suggested by @rcroessmann 

### Tested

Deployed baklava & mainnet oracles, and redeployed baklava full nodes to use the new network

### Related issues

n/a

### Backwards compatibility

Yes


codyborn (approved):
LGTM! Can you create a [new release](https://github.com/celo-org/celo-oracle/releases) with this commit once it's checked in?

tkporter:
Yup Audrey created one yesterday :) https://github.com/celo-org/celo-oracle/releases/tag/v1.0.2

PR #4211, opened 2020-06-25
[Mobile] Add additional security to wallet key storage

cmcewen:
### Description

To improve security of key storage, this generates a unique client pepper and store in the keystore/keychain to provide more entropy to the blockchain. It, combined with a user pin, is used to encrypt the private key on disk. This prevents an attacker with disk access from easily getting the private key

_[Rossy edit]:_

Also blacklisted unsafe pins, sanitized logs, fixed issue with pin caching while navigating, renamed lockWithPinEnabled to requirePinOnAppOpen, and some refactoring.

### Other changes

Store the mnemonic encrypted in the keystore.

### Tested

Run locally on android and iOS

_[Rossy edit]:_

Tested with adding, removing, and changing PIN settings on Android, confirmed the keystore values were still accessible. Mnemonic decrypts correctly and account can be unlocked.

Tested changing PIN on iOS, confirmed the same. Tested with no PIN on iOS with simulator, all good.


packages/mobile/__mocks__/react-native-keychain.ts
```diff
@@ -0,0 +1,10 @@
+const keychainMock = {
+  SECURITY_LEVEL_ANY: 'MOCK_SECURITY_LEVEL_ANY',
+  SECURITY_LEVEL_SECURE_SOFTWARE: 'MOCK_SECURITY_LEVEL_SECURE_SOFTWARE',
+  SECURITY_LEVEL_SECURE_HARDWARE: 'MOCK_SECURITY_LEVEL_SECURE_HARDWARE',
+  setGenericPassword: jest.fn(),
+  getGenericPassword: jest.fn(),
+  resetGenericPassword: jest.fn(),
+}
+
+export default keychainMocka
```
jmrossy: typo?

packages/mobile/android/app/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -1,2 +1,13 @@
-connection.project.dir=..
+arguments=
```
jmrossy: Does this file just affect eclipse users?

packages/mobile/android/build.gradle
```diff
@@ -4,8 +4,8 @@ buildscript {
     ext {
         buildToolsVersion = "29.0.2"
         minSdkVersion = 16
-        compileSdkVersion = 28
-        targetSdkVersion = 28
+        compileSdkVersion = 29
```
jmrossy, line 7: why bump to 29?

packages/mobile/src/account/saga.ts
```diff
@@ -14,25 +13,13 @@ import { CustomEventNames } from 'src/analytics/constants'
...
 const TAG = 'account/saga'
 
-export function* setPincode({ pincodeType, pin }: SetPincodeAction) {
+export function* setPincode({ pincodeType }: SetPincodeAction) {
   try {
-    if (pincodeType === PincodeType.PhoneAuth) {
-      Logger.debug(TAG + '@setPincode', 'Setting pincode with using system auth')
-      pin = randomBytes(10).toString('hex') as string
-      yield call(setPinInKeystore, pin)
-    } else if (pincodeType === PincodeType.CustomPin && pin) {
```
jmrossy, line 29: with this gone, what's this function supposed to be doing now?

packages/mobile/src/pincode/PincodeSet.tsx
```diff
@@ -69,12 +69,18 @@ export class PincodeSet extends React.Component<Props, State> {
...
     if (this.isPin1Valid(pin1) && this.isPin2Valid(pin2)) {
-      this.props.setPincode(PincodeType.CustomPin, this.state.pin1)
-      navigate(Screens.EnterInviteCode)
+      try {
+        await storePasswordHash(this.state.pin1, '')
+        setPinCache(this.state.pin1)
+        this.props.setPincode(PincodeType.CustomPin)
+        this.props.navigation.navigate(Screens.EnterInviteCode)
+      } catch (err) {
+        CeloAnalytics.track(CustomEventNames.pin_store_error, { error: err.message })
```
jmrossy: Might wanna confirm the pin's not in this message

packages/mobile/src/pincode/authentication.ts
```diff
@@ -0,0 +1,99 @@
...
+const TAG = 'pincode/authentication'
+enum STORAGE_KEYS {
+  PEPPER = 'PEPPER',
+  PASSWORD_HASH = 'PASSWORD_HASH',
+}
+const PEPPER_LENGTH = 64
+export const PIN_LENGTH = 6
+
+let cachedPepper: string
+const cachedPasswordHashes: {
```
jmrossy: Wondering if this should go with the other cached stuff in your new passwordCache file?

packages/mobile/src/pincode/authentication.ts
```diff
@@ -0,0 +1,99 @@
...
+    await storeItem({ key: STORAGE_KEYS.PEPPER, value: pepper })
+    cachedPepper = pepper
+    return cachedPepper
+  }
+  cachedPepper = storedPepper
+  return cachedPepper
+}
+
+async function hashForPin(pin: string) {
+  return sha256(new Buffer(await passwordForPin(pin), 'hex')).toString('hex')
```
jmrossy: Should we use a multi-iteration hash here for more security?

packages/mobile/src/pincode/authentication.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (unlocked) {
+      await storePasswordHash(pin, account)
+      return true
+    }
+    return false
+  }
+  if (hash === passwordHash) {
+    return true
+  }
+  return false
```
jmrossy:
```suggestion
  return hash === passwordHash
```

packages/mobile/src/pincode/authentication.ts
```diff
@@ -0,0 +1,99 @@
...
+  const passwordHash = await retrievePasswordHash(account)
+  if (!passwordHash) {
+    Logger.error(`${TAG}@checkPin`, 'No password hash stored')
+    const currentAccount = currentAccountSelector(store.getState())
+    if (!currentAccount) {
+      throw new Error('No account to unlock')
+    }
+    const unlocked = await ensureCorrectPin(await passwordForPin(pin), currentAccount)
+    if (unlocked) {
+      await storePasswordHash(pin, account)
```
jmrossy: Why do we store the password hash instead of recomputing it with the pin each time?

packages/mobile/src/pincode/authentication.ts
```diff
@@ -0,0 +1,99 @@
...
+
+function pinStorageKey(account: string) {
+  return `${STORAGE_KEYS.PASSWORD_HASH}-${account}`
+}
+
+async function retrieveOrGeneratePepper() {
+  if (cachedPepper) {
+    return cachedPepper
+  }
+  const storedPepper = await retrieveStoredItem(STORAGE_KEYS.PEPPER)
```
jmrossy: Should we have error handling at some layer or is it intentional to allow them to propagate up to the component?

packages/mobile/src/pincode/authentication.ts
```diff
@@ -0,0 +1,99 @@
...
+  const hash = await hashForPin(pin)
+  await storeItem({ key: pinStorageKey(account), value: hash })
+}
+
+export async function checkPin(pin: string, account: string) {
+  const hash = await hashForPin(pin)
+  const passwordHash = await retrievePasswordHash(account)
+  if (!passwordHash) {
+    Logger.error(`${TAG}@checkPin`, 'No password hash stored')
+    const currentAccount = currentAccountSelector(store.getState())
```
jmrossy: Any reason this new code isn't in a saga? Is it just to avoid the boilerplate?

packages/mobile/src/utils/keyStore.ts
```diff
@@ -8,8 +8,9 @@ export async function setKey(key: string, value: string) {
   try {
     Logger.debug(TAG, `Setting key ${key} in keystore`)
     // TODO(Rossy + Jean): Revisit this accessible setting
-    await RNSecureKeyStore.set(key, value, {
-      accessible: ACCESSIBLE.WHEN_UNLOCKED,
+    await Keychain.setGenericPassword('CELO', value, {
```
jmrossy: We should probably folder this code into your new storage/keychain file right?

packages/mobile/src/web3/saga.ts
```diff
@@ -259,12 +263,14 @@ export function* unlockAccount(account: string) {
   }
 
   try {
-    const pincode = yield call(getPincode, true)
-    yield call([wallet, wallet.unlockAccount], account, pincode, UNLOCK_DURATION)
+    const pin = yield call(getPincode, true)
+    const password: string = yield call(passwordForPin, pin)
```
jmrossy: These two seem to be always called together. Should they be merged or wrapped?

packages/mobile/android/build.gradle
cmcewen: This was necessary for react-native-keychain to work properly

packages/mobile/src/account/saga.ts
cmcewen: I left it so that we could easily add biometrics in later

packages/mobile/src/pincode/authentication.ts
cmcewen: I just want to avoid any password handling getting accidentally logged, so trying to keep it as contained as possible

packages/mobile/src/backup/utils.ts
```diff
@@ -162,3 +170,14 @@ export function isValidBackupPhrase(phrase: string) {
 export function isValidSocialBackupPhrase(phrase: string) {
   return isValidMnemonic(phrase, 13)
 }
+
+export async function encryptMnemonic(phrase: string) {
+  const pepper = await retrieveOrGeneratePepper()
+  return CryptoJS.AES.encrypt(phrase, pepper).toString()
```
jeanregisser: Shouldn't we encrypt the mnemonic with both the pepper and passwordHash?

jeanregisser (approved):
Awesome work and effort for tightening the security of the wallet! 🥇 🎉  
Left one question.

packages/mobile/src/backup/utils.ts
jmrossy: Do we need to specify the encoding here too like line 182?

packages/mobile/src/backup/utils.ts
```diff
@@ -162,3 +170,14 @@ export function isValidBackupPhrase(phrase: string) {
 export function isValidSocialBackupPhrase(phrase: string) {
   return isValidMnemonic(phrase, 13)
 }
+
+export async function encryptMnemonic(phrase: string) {
+  const pepper = await retrieveOrGeneratePepper()
```
jmrossy: I thought we were going to use the pin to encrypt the mnemonic in case the pepper is lost before the user does the backup flow?

packages/mobile/src/pincode/authentication.ts
```diff
@@ -0,0 +1,151 @@
...
+  PASSWORD_HASH = 'PASSWORD_HASH',
+}
+const PEPPER_LENGTH = 64
+export const PIN_LENGTH = 6
+
+export function isPinValid(pin: string) {
+  return pin.length === PIN_LENGTH
+}
+
+function pinStorageKey(account: string) {
```
jmrossy: The name of this confused me. It's actually the account password hash storage key right?

packages/mobile/src/pincode/authentication.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+    if (!hash) {
+      return null
+    }
+    setCachedPasswordHash(account, hash)
+  }
+  return getCachedPasswordHash(account)
+}
+
+type PinPromise = (pin: string) => void
```
jmrossy: PinCallback?

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      .returns(expectedPassword)
+      .run()
+
+    expect(navigate).toHaveBeenCalled()
+    expect(navigateBack).toHaveBeenCalled()
+  })
+})
+
+describe(getPincode, () => {
+  // it('returns PIN from phone keystore', async () => {
```
jmrossy, line 37: Copied over from the old `getPincode` tests in the now deleted account/saga.test file

packages/mobile/src/pincode/authentication.ts
```diff
@@ -1,94 +1,109 @@
...
   PASSWORD_HASH = 'PASSWORD_HASH',
 }
+
 const PEPPER_LENGTH = 64
 export const PIN_LENGTH = 6
+// Pepper and pin not currently generalized to be per account
+// Using this value in the caches
+export const DEFAULT_CACHE_ACCOUNT = 'default'
+
+const PIN_BLACKLIST = [
```
tarikbellamine, line 45: May be worth creating/throwing a custom error message for this to surface to the user. IMO may not be obvious why these are invalid.

Also would be nice to disable this with a dev flag because we've just blacklisted all of my go-tos ;)

packages/react-components/services/ReactNativeLogger.ts
```diff
@@ -58,6 +58,20 @@ export default class ReactNativeLogger {
...
 
+  sanitizeError = (error: Error, valueToPurge?: string) => {
+    const message = this.getErrorMessage(error)
+
+    if (message.toLowerCase().includes('password') || message.toLowerCase().includes('key')) {
+      return new Error('Error message hidden for privacy')
+    }
+
+    if (valueToPurge) {
+      return new Error(message.replace(new RegExp(valueToPurge, 'g'), ''))
```
i1skn: Do we want to place something like `<purged>` instead of `''` here?

i1skn (approved):
Great work!! Only one little comment.

packages/react-components/services/ReactNativeLogger.ts
jmrossy: Will do

packages/mobile/src/backup/utils.ts
jmrossy: I'll be changing this to use the password instead, which is pin + pepepr

packages/mobile/src/backup/utils.ts
jmrossy: Using password instead.

packages/mobile/src/pincode/authentication.ts
jmrossy: Yeah that's fair but I'm going to punt that to a follow-up PR. Juggling a few things atm and I think it's important to land this asap cause its risky.

jmrossy: For the moment it shows 'Invalid Pin' to the user

PR #4255, opened 2020-06-27
[contractkit] Fix gas price backwards compatibility

gastonponti:
### Description

Fix backwards compatibility for the gasPrice call

### Backwards compatibility

Yes


yorhodes (approved):
thanks!

PR #4263, opened 2020-06-29
Show badges on Validator Explorer

Pedro-vk:
### Description

_Show badges on Validator Explorer_

### Tested

_Tested on Chrome._


packages/web/src/dev/ValidatorsListBadges.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const allBadges = [
+  { src: '/images/badges/1st-proposer.svg', list: lProposer },
+  { src: '/images/badges/1st-transaction.svg', list: lTransaction },
+  { src: '/images/badges/attestation.svg', list: lAttestation },
+  { src: '/images/badges/founder-validator.svg', list: lFounder },
+  { src: '/images/badges/genesis-validator.svg', list: lGenesis },
+  { src: '/images/badges/master-validator.svg', list: lMaster },
+  { src: '/images/badges/vote-recipient.svg', list: lVote },
+].map((_) => ({ ..._, list: (_.list as any).toLowerCase() as string }))
```
aaronmgdr: please give this variable a name. `_` should only be used in cases where the variable is not used.

aaronmgdr: Also I believe `list` can be transformed into a Set like 

`.map((addresses) => ({ ...addresses, list:  new Set(addresses.split('/n')) }))`

this way below you can do 

 `allBadges.filter(({ list }) => list.has(address.toLowerCase()))`

and avoid looping inside a loop while rendering.

packages/web/public/images/badges/vote-recipient.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aaronmgdr, line 1: I see we already have badges in another folder. can we deduplicate?

packages/web/src/dev/ValidatorsListBadges.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+interface Props {
+  address: string
+}
+
+class ValidatorsListBadges extends React.PureComponent<Props & I18nProps> {
+  state = {
```
aaronmgdr: remove this state

aaronmgdr (changes requested):
good work. just a few things to do

packages/web/src/dev/ValidatorsListBadges.tsx
Pedro-vk: `allBadges.filter(({ list }) => list.includes(address.toLowerCase()))`
This part of the code is not a double loop, is just a string matching. List is a long string with addresses (and new lines), so just checking if the string contains the address.

But anyway, I will change it.

packages/web/public/images/badges/vote-recipient.svg
Pedro-vk: They are not the same, this badges are without borders.

aaronmgdr (approved):

PR #4264, opened 2020-06-29
Use path.join() in metadata CLI requests to avoid // in URIs

lahabana:
### Description

If the metadataURL was registered with a trailling slash the verify attestation service calls would fail as it would build URIs with 2 "/".
This avoids this by using `path.join()`.


gastonponti (approved):
LGTM 👍 
Thanks!

lahabana:
Rebased. #4313 Did fix this issue but it was introducing a function which I think is unecessary @nambrot might have a different opinion though

jmrossy:
From the node documentation, it looks like the `path` package is meant for 'working with file and directory paths'.
Just did a quick test with it in a node repl and it seems like it mangles URLs:
```
> const path = require('path')
undefined
> path.join('http://google.com', 'hi')
'http:/google.com/hi'
```

@lahabana Are you sure path can be safely used for URL construction on all platforms?

jmrossy (changes requested):
See comment above

lahabana:
My bad then! Closing thanks for checking out

PR #4269, opened 2020-06-30
[Wallet] Restructuring Wallet Analytics

tarikbellamine:
### Description
Restructured and renamed the Valora wallet analytics. Changes include:
- Organized analytics events by category and introduced type checking for parameters
- Refactored all analytics parameters to be primitive data types to avoid pipeline or database errors
- Created a sessionId property and attach it and user address on each logged event
- Deleted unused or redundant events and parameters
- Consolidated all analytics into the `mobile` directory (deleted parent component in `react-components`)
- Removed subevent logging
- Renaming `CeloAnalytics` to `ValoraAnalytics`

To cover in next PR:
- Standardize event naming
- Further consolidate events
- Audit against [analytics doc](https://docs.google.com/document/d/1OGNJBijiWp7bWc2I0p8yyJ72EMCj9pNRT1VWxJJu5JY/edit?pli=1#heading=h.wip61atcgd2c)
- Introduce new events to fill gaps
- Update documentation

### Other changes
Fixed a couple of events that were broken.

### Tested
Yes

### Related issues
- Part of #4008

### Backwards compatibility
Yes


packages/mobile/src/identity/privateHashing.ts
```diff
@@ -247,10 +249,19 @@ function* navigateToQuotaPurchaseScreen() {
       throw new Error('Purchase tx failed')
     }
 
+    ValoraAnalytics.track(AnalyticsEvents.phone_number_quota_purchase_success)
     Logger.debug(`${TAG}@navigateToQuotaPurchaseScreen`, `Quota purchase successful`)
     navigateBack()
     return true
   } catch (error) {
+    // CC Rossy
```
tarikbellamine: @jmrossy does this look right to you? i don't think we were distinguishing between a skip and fail in the previous iteration

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    Analytics.setup(SEGMENT_API_KEY, SEGMENT_OPTIONS).catch((error) =>
+      Logger.debug(TAG, `Segment setup error: ${error.message}\n`, error)
+    )
+    Logger.debug(TAG, 'Segment Analytics Integration initialized!')
+
+    getDeviceInfo()
+      .then((res) => {
+        this.deviceInfo = res
+        this.sessionId = Web3Utils.soliditySha3({
```
tarikbellamine, line 73: Is there a better way to generate a session identifier?

packages/mobile/src/identity/privateHashing.ts
jmrossy: Looks okay. It might be more natural to just `reject('skipped')` and then compare the error to 'skipped' directly but this is fine if you've already tested this.

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,150 @@
...
+
+async function getDeviceInfo() {
+  return {
+    AppName: DeviceInfo.getApplicationName(),
+    Brand: DeviceInfo.getBrand(),
+    BuildNumber: DeviceInfo.getBuildNumber(),
+    BundleId: DeviceInfo.getBundleId(),
+    Carrier: await DeviceInfo.getCarrier(),
+    DeviceId: DeviceInfo.getDeviceId(),
+    DeviceName: await DeviceInfo.getDeviceName(), // NOTE(nitya) this might contain PII, monitor
```
jmrossy: ditto

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,150 @@
...
+
+const TAG = 'ValoraAnalytics'
+
+async function getDeviceInfo() {
+  return {
+    AppName: DeviceInfo.getApplicationName(),
+    Brand: DeviceInfo.getBrand(),
+    BuildNumber: DeviceInfo.getBuildNumber(),
+    BundleId: DeviceInfo.getBundleId(),
+    Carrier: await DeviceInfo.getCarrier(),
```
jmrossy, line 19: Can we remove this? I haven't had time to do the PII purge yet but this should probably go

packages/mobile/src/analytics/ValoraAnalytics.ts
jmrossy: Does segment and/or firebase make one for us already?

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.sessionId = ''
+    this.deviceInfo = {}
+    this.userAddress = 'not yet set'
+
+    if (!SEGMENT_API_KEY) {
+      Logger.debug(TAG, 'Segment API Key not present, likely due to environment. Skipping enabling')
+      return
+    }
+
+    Analytics.setup(SEGMENT_API_KEY, SEGMENT_OPTIONS).catch((error) =>
```
jmrossy: I think it's generally not advisable to due async / fail-able actions in constructors. You could instead move this to an async `initi` method

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,150 @@
...
+
+class ValoraAnalytics {
+  sessionId: string
+  userAddress: string | null | undefined
+  deviceInfo: object
+
+  constructor() {
+    this.sessionId = ''
+    this.deviceInfo = {}
+    this.userAddress = 'not yet set'
```
jmrossy: @nityas how would you feel about leaving out the user address. Given the growing scope of analytics, I'm not sure how ethical it is to tie all this activity to particular addresses

jmrossy (commented):
Mostly LGTM, just a few comments below

packages/mobile/src/analytics/ValoraAnalytics.ts
nityas: hm...on ethics I lean on the ability of any user to opt out of analytics at any time(which we mention in onboarding), which is far more than most consumer apps (and all comparables) offer. In general, we could always tie together events implicitly through anonymous IDs, so I see this as a way to make sure our data is accurate (this also sets us up better to accurately bulk delete user data if a request is made, which is a feature I could see us adding in the future)

packages/mobile/src/analytics/ValoraAnalytics.ts
tarikbellamine: I would think so but we haven't seen it come through in the data. At any rate, I think it's good to have universal identifiers that we have granular control over and aren't tied to any specific database or packages so the data remains easily portable.

packages/mobile/src/analytics/ValoraAnalytics.ts
tarikbellamine: @nityas are you good with cutting this? I agree that the probability of it containing PII is pretty high.

packages/mobile/src/analytics/ValoraAnalytics.ts
nityas: carrier is really helpful for understanding verification issues...

packages/mobile/src/analytics/ValoraAnalytics.ts
nityas: +1 to removing this

packages/mobile/src/analytics/ValoraAnalytics.ts
nityas: Feature request for analytics: ability to log encrypted fields & decrypt on-demand while querying, such that risk is minimized if analytics infrastructure is compromised

tarikbellamine: I'm certainly open to it but don't feel strongly either way.

Most countries only have a handful of carriers, which can be easily implied by country codes. I can see having the granularity of carrier be useful for analytics around verification but you are much closer to that than I am so I'd defer to your assessment of the tradeoffs.

packages/mobile/src/analytics/ValoraAnalytics.ts
tarikbellamine: Are there any specific data points you're hesitant to tie to address or is this more of a general concern?

I think no collection of PII, an easily accessible opt-out of all analytics, and clear documentation on what's being collected sets a sufficiently high bar.

packages/mobile/src/analytics/ValoraAnalytics.ts
tarikbellamine: Makes sense.

I'm not very familiar with this design pattern. Do you mind taking a look at how I've done it in my latest "PR feedback" commit?

packages/mobile/src/analytics/ValoraAnalytics.ts
annakaz: @jmrossy because we're tracking transaction hashes, we can identify the user address from the hash sent from a particular wallet. I agree that we should not log any PII (phone number, name, etc) but would argue that the address falls more in the category of on-chain data that is acceptable to track

packages/mobile/src/alert/actions.ts
```diff
@@ -65,12 +65,9 @@ export const showErrorInline = (error: ErrorMessages, i18nOptions?: TOptions): S
 export function showErrorOrFallback(error: any, fallback: ErrorMessages) {
   if (error && Object.values(ErrorMessages).includes(error.message)) {
     return showError(error.message)
-  } else {
-    CeloAnalytics.track(CustomEventNames.error_fallback, {
```
annakaz, line 69: Any reason to remove this? Error_fallback was helping to catch missing errors- for example if verification failed we could see a more detailed error message than the one showed to users

packages/mobile/src/app/ErrorBoundary.tsx
```diff
@@ -24,7 +24,7 @@ class ErrorBoundary extends React.Component<Props, State> {
 
   componentDidCatch(error: Error, info: any) {
     this.setState({ childError: error })
-    CeloAnalytics.track(DefaultEventNames.errorDisplayed, { error }, true)
+    ValoraAnalytics.track(AnalyticsEvents.error_displayed, { error: error.message })
```
annakaz, line 27: Could be useful to keep the device info for errors

packages/mobile/src/app/saga.ts
```diff
@@ -113,8 +113,9 @@ function* watchAppState() {
       const newState = yield take(appStateChannel)
       Logger.debug(`${TAG}@monitorAppState`, `App changed state: ${newState}`)
       yield put(setAppState(newState))
+      // place event here
```
annakaz: Add TODO?

packages/mobile/src/geth/geth.ts
```diff
@@ -317,13 +317,10 @@ async function uploadLogs(gethLogFilePath: string, reactNativeLogFilePath: strin
...
     const uploadPath = `${bundleId}/${DEFAULT_TESTNET}`
 
-    // Phone number might not be verified here but that does not matter for logging.
-    const phoneNumber = (await DeviceInfo.getPhoneNumber()) || 'unknown'
     const timestamp = new Date().getTime()
     const deviceId = DeviceInfo.getUniqueId()
-    const uploadId = `${timestamp}_${deviceId}`
-    const gethUploadFileName = `${phoneNumber}_${uploadId}_geth.txt`
-    const reactNativeUploadFileName = `${phoneNumber}_${uploadId}_rn.txt`
+    const gethUploadFileName = `${deviceId}_${timestamp}_geth.txt`
```
annakaz, line 322: 👍  Thanks for removing the phone number here

packages/mobile/src/geth/saga.ts
```diff
@@ -137,19 +137,22 @@ export function* initGethSaga() {
     }
   }
 
-  CeloAnalytics.track(CustomEventNames.geth_init_failure, {
-    error: result,
+  ValoraAnalytics.track(AnalyticsEvents.geth_init_failure, {
+    error: result.message,
```
annakaz: Looks like `result` is just of type GethInitOutcomes and can be passed without .message

packages/mobile/src/identity/verification.ts
```diff
@@ -63,23 +63,31 @@ export function* startVerification() {
...
   const { result, cancel, timeout } = yield race({
     result: call(doVerificationFlow),
     cancel: take(Actions.CANCEL_VERIFICATION),
     timeout: delay(VERIFICATION_TIMEOUT),
   })
 
   if (result === true) {
-    CeloAnalytics.track(CustomEventNames.verification_success)
+    ValoraAnalytics.track(AnalyticsEvents.verification_success, {
+      duration: Date.now() - startTime,
```
annakaz, line 76: Generally opposed to client side timing metrics but agree with the trade off to have these easily accessible

packages/mobile/src/identity/verification.ts
```diff
@@ -63,23 +63,31 @@ export function* startVerification() {
...
   } else if (cancel) {
-    CeloAnalytics.track(CustomEventNames.verification_cancelled)
+    ValoraAnalytics.track(AnalyticsEvents.verification_cancelled, {
+      duration: Date.now() - startTime,
+    })
     Logger.debug(TAG, 'Verification cancelled')
   } else if (timeout) {
-    CeloAnalytics.track(CustomEventNames.verification_timed_out)
+    ValoraAnalytics.track(AnalyticsEvents.verification_timed_out, {
+      duration: Date.now() - startTime,
```
annakaz, line 89: Do these durations end up passed as a string? Not sure how the pipeline handles date objects

packages/mobile/src/invite/saga.ts
```diff
@@ -314,7 +316,9 @@ export function* doRedeemInvite(inviteCode: string) {
       tempAccount
     )
     if (tempAccountBalanceWei.isLessThanOrEqualTo(0)) {
-      CeloAnalytics.track(CustomEventNames.redeem_invite_failed, { context: 'Empty invite' })
+      ValoraAnalytics.track(AnalyticsEvents.redeem_invite_failed, {
+        error: 'Escrow account has no balance',
```
annakaz: In this case the invite is empty, not the escrow account. There are two accounts used for invite codes: a temp account to pay for verification and some escrow amount sent after verification

packages/mobile/src/onboarding/registration/JoinCelo.tsx
```diff
@@ -169,7 +169,11 @@ export class JoinCelo extends React.Component<Props, State> {
     }
 
     if (!e164Number || !isValidNumber || !countryCallingCode) {
-      CeloAnalytics.track(CustomEventNames.inavlid_phone_number, { phoneNumber: e164Number })
+      // Replacing all integers except 0 with and “X” before logging so
```
annakaz: nit: typo "an"

packages/mobile/src/tokens/saga.ts
```diff
@@ -73,13 +73,13 @@ export function tokenFetchFactory({ actionName, token, actionCreator, tag }: Tok
...
       const balanceInWei: BigNumber = yield call([tokenContract, tokenContract.balanceOf], account)
       const balance: BigNumber = yield call(convertFromContractDecimals, balanceInWei, token)
-      CeloAnalytics.track(
-        CustomEventNames.fetch_balance,
+      ValoraAnalytics.track(
+        AnalyticsEvents.fetch_balance,
         token === CURRENCY_ENUM.DOLLAR
           ? {
-              dollarBalance: balance,
+              dollarBalance: balance.toString(),
```
annakaz, line 80: 👍  thanks for fixing

annakaz (approved):
LGTM overall, a few comments below. See @jmrossy suggestion on async constructors too

packages/mobile/src/identity/verification.ts
aaronmgdr: `Date.now` returns a number

tarikbellamine: Yup, it's a number type

packages/mobile/src/identity/verification.ts
tarikbellamine: Yeah, I was thinking of scrapping this because we can just calculate the difference between the `verification_start` event and everything else. Decided to hold off on doing it until we have bandwidth to focus on creating more complex data queries but hope to eliminate these timing metrics sometime soon

packages/mobile/src/invite/saga.ts
tarikbellamine: Ah, gotcha. I will change the error message.

packages/mobile/src/app/saga.ts
tarikbellamine: Good catch. Actually don't need this anymore because looks like segment gets the app state changes.

packages/mobile/src/analytics/ValoraAnalytics.ts
jmrossy: Yeah good points here. The verification debugging bit is a strong argument. I guess I'm okay with keeping it included for a while.

packages/mobile/src/app/ErrorBoundary.tsx
tarikbellamine: What we are doing now is exclusively capturing device info on app load. Device info rarely changes so the thought is we should only really log it once per session at a maximum. From there, we can tie in device info for any particular event by doing a join query on sessionId. This minimizes payload sizes, which should have positive performance + network implications.

Do you foresee any issues with this approach?

packages/mobile/src/analytics/ValoraAnalytics.ts
jmrossy: Yeah this is a tricky one. On the one hand, all on-chain activity is already publicly tied to an address. On the other, addresses are the blockchain-equivalent of user IDs so I feel like including it weakens our claim that we do analytics in an anonymous way. 

It certainly helps us make correlations across services (like identifying misbehaving PGPNP users). 

I have an inherent bias towards logging less, tracking less, but I'll defer to you three on this one :)

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,152 @@
...
+  sessionId: string
+  userAddress: string | null | undefined
+  deviceInfo: object
+
+  constructor() {
+    this.sessionId = ''
+    this.userAddress = ''
+    this.deviceInfo = {}
+
+    if (!SEGMENT_API_KEY) {
```
jmrossy: This needs to go in init now right?

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,152 @@
...
+      Logger.debug(TAG, 'Segment API Key not present, likely due to environment. Skipping enabling')
+    }
+  }
+
+  init() {
```
jmrossy: Recommend making this async in possible and then using `await` instead of hooking on `.catch` and `.then`. callbacks

packages/mobile/src/alert/actions.ts
tarikbellamine: Oh my bad, I missed this use case and will add it back. Was just trying to cut down on repetitiveness because we are already logging errors with context specific error events, on error boundary, on showError, and also here.

Thinking about if there's a better way to do this for the next PR

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    Analytics.screen(page, eventProperties).catch((err) => {
+      Logger.error(TAG, 'Error tracking page', err)
+    })
+  }
+}
+
+const ValoraAnalyticsInstance = new ValoraAnalytics()
+ValoraAnalyticsInstance.init()
```
jmrossy: How about we call this from `index.js` or `App.tsx` for greater control of when it runs. By putting it here, it'll execute whenever this file happens to be first imported, and it'll block until init returns.

tarikbellamine: Got it. Put it in `App` right before I call `startSession`. I considered folding `init` and `startSession` into one method but figured it's better to keep the concerns a little more separate.

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -0,0 +1,153 @@
...
+  },
+}
+
+class ValoraAnalytics {
+  sessionId: string
+  userAddress: string | null | undefined
+  deviceInfo: object
+
+  constructor() {
+    this.sessionId = ''
```
jmrossy: Nit, looks like we don't even need the constructor anymore at this point, you can just assign these defaults above.

jmrossy (approved):

PR #4273, opened 2020-06-30
monitor full node health

jasonrudianto1:
### Description

Set the --pprof flag for full nodes and configured prometheus to scrape the /debug/metrics/prometheus endpoint exposed on the full node's port 6060.

### Other changes

N/A

### Tested

Deployed to oracledev cluster.

### Related issues

- Fixes # celo-oracle-111


packages/helm-charts/celo-fullnode/values.yaml
```diff
@@ -82,3 +82,11 @@ genesis:
   network: baklavastaging
   useGenesisFileBase64: false
   genesisFileBase64: ""
+
+# If Prometheus is true, metrics and pprof set to true
+prometheus: true
+metrics: true
```
tkporter, line 88: Right now `metrics` isn't used, what do you think of changing the template `common.full-node-container` to conditionally add the `--metrics` flag to the additional flags? We can have it default to true to preserve backward compatibility

tkporter (commented):
Looks great :) 1 small thing

tkporter (approved):

PR #4275, opened 2020-06-30
AWS HSM wallet

AlexBHarley:
### Description

Integrates ContractKit with AWS HSMs. 

### Tested

Running an oracle client locally against an HSM I set up in AWS.


packages/contractkit/src/utils/ber-utils.ts
```diff
@@ -0,0 +1,33 @@
+import { ensureLeading0x } from '@celo/utils/lib/address'
+import * as asn1 from 'asn1js'
+
+export const toArrayBuffer = (buffer: Buffer): ArrayBuffer => {
```
codyborn: It looks like we can just use the built in [.buffer property](https://nodejs.org/docs/latest/api/buffer.html#buffer_buf_buffer) in node versions >= 4.x
https://stackoverflow.com/a/12101012/677846

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
```diff
@@ -0,0 +1,109 @@
...
+      const N = bufferToBigNumber(secp256k1Curve.curve.n)
+      if (isCanonical(S, N)) {
+        flag = false
+      }
+    }
+
+    return { S: S!, R: R! }
+  }
+
+  private async sign(buffer: Buffer): Promise<{ v: number; r: Buffer; s: Buffer }> {
```
tkporter: wdyt of creating a new type for `{ v: number; r: Buffer; s: Buffer }`? It seems to be used a lot outside of this PR too

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
```diff
@@ -0,0 +1,109 @@
...
+    const recoveryParam = getRecoveryParam(buffer, this.address, rBuff, sBuff)
+
+    return {
+      r: rBuff,
+      s: sBuff,
+      v: recoveryParam,
+    }
+  }
+
+  async signTransaction(
```
tkporter: `signTransaction` and `signPersonalMessage` here are pretty similar to the ones in the azure signer with the exception of the `sign` calls, do you think there's any way we could pull all that common code into `Signer` such that it will use the `sign` implementation that's specific to AWS/Azure?

tkporter (commented):
Looks really really good
What are the test issues you're running into?

packages/contractkit/src/utils/ber-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  return ensureLeading0x(Buffer.from(value.valueBlock.valueHex.slice(1)).toString('hex'))
+}
+
+/**
+ * AWS returns DER encoded signatures but DER is valid BER
+ */
+export function parseBERSignature(sig: Buffer): { r: Buffer; s: Buffer } {
+  const { result } = asn1.fromBER(toArrayBuffer(sig))
+
+  const part1 = (result as asn1.Sequence).valueBlock.value[0] as asn1.BitString
```
codyborn: I'm unable to find info on this `valueBlockType` constructor. Is it guaranteed to have a value array property with at least two values?

packages/contractkit/src/wallets/aws-hsm-wallet.ts
```diff
@@ -0,0 +1,46 @@
...
+      } catch (e) {
+        // todo: what does the error look like here
+        throw e
+      }
+    }
+    return addressToSigner
+  }
+
+  private generateKmsClient() {
+    return new KMS({ region: 'eu-central-1', apiVersion: '2014-11-01' })
```
codyborn: It seems like this value should be configurable via environment vars so that we can support a variety of regions.

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
```diff
@@ -0,0 +1,109 @@
...
+import * as ethUtil from 'ethereumjs-util'
+import { parseBERSignature } from '../../utils/ber-utils'
+import { bigNumberToBuffer, bufferToBigNumber, isCanonical } from '../../utils/signature-utils'
+import { getHashFromEncoded, RLPEncodedTx } from '../../utils/signing-utils'
+import { Signer } from './signer'
+
+const SigningAlgorithm = 'ECDSA_SHA_256'
+const secp256k1Curve = new EC('secp256k1')
+
+function getRecoveryParam(message: Buffer, expectedAddress: string, r: Buffer, s: Buffer): number {
```
codyborn: Can we re-use `recoverKeyIndex` from [here](https://github.com/celo-org/celo-monorepo/blob/11814bdb3b44bddb3c19a9ec0f439ad45e86b6bc/packages/contractkit/src/utils/azure-key-vault-client.ts#L147)? From what I read it's possible for this key to be in the set [0,1,2,3], although 2,3 are very rare:
https://ethereum.stackexchange.com/questions/42455/during-ecdsa-signing-how-do-i-generate-the-recovery-id

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.keyId = keyId
+    this.address = address
+  }
+
+  private async findCanonicalSignature(buffer: Buffer): Promise<{ S: BigNumber; R: BigNumber }> {
+    let S: BigNumber
+    let R: BigNumber
+
+    let flag = true
+    while (flag) {
```
codyborn: Is this signing operation non-deterministic? What causes it to become canonical?

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
codyborn: +1 There's one here, but it should probably be moved to a common signature lib:
https://github.com/celo-org/celo-monorepo/blob/11814bdb3b44bddb3c19a9ec0f439ad45e86b6bc/packages/contractkit/src/utils/azure-key-vault-client.ts#L213

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
codyborn: I think that's a good idea.

packages/contractkit/src/wallets/remote-wallet.test.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Signer } from './signers/signer'
+import { ACCOUNT_ADDRESS1, CHAIN_ID, TYPED_DATA } from './test-utils'
+
+class RemoteWalletImpl extends RemoteWallet {
+  protected async loadAccountSigners(): Promise<Map<Address, Signer>> {
+    return new Map<Address, Signer>()
+  }
+}
+
+describe('AzureHSMWallet class', () => {
```
codyborn:
```suggestion
describe('RemoteWallet class', () => {
```

packages/contractkit/src/wallets/remote-wallet.test.ts
```diff
@@ -0,0 +1,71 @@
+import { Address } from '@celo/utils/lib/address'
+import Web3 from 'web3'
+import { Tx } from 'web3-core'
+import { RemoteWallet } from './remote-wallet'
+import { Signer } from './signers/signer'
+import { ACCOUNT_ADDRESS1, CHAIN_ID, TYPED_DATA } from './test-utils'
+
+class RemoteWalletImpl extends RemoteWallet {
```
codyborn, line 8: Thanks for adding this!!

packages/contractkit/src/wallets/aws-hsm-wallet.test.ts
```diff
@@ -0,0 +1,259 @@
...
+import { recoverTransaction, verifyEIP712TypedDataSigner } from '../utils/signing-utils'
+import AwsHsmWallet from './aws-hsm-wallet'
+import {
+  ACCOUNT_ADDRESS1,
+  ACCOUNT_ADDRESS2,
+  ACCOUNT_ADDRESS_NEVER,
+  CHAIN_ID,
+  TYPED_DATA,
+} from './test-utils'
+
```
codyborn, line 20: I believe you'll have to add this line to read from env file:
`require('dotenv').config()`

codyborn (changes requested):
Looks good! Before you check-in please confirm that you've ran the tests using the non-mock versions of AWS and Azure.

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
AlexBHarley: I've really struggled getting this to work... The ecdsaRecover fn I an using just calls into the `secp256k1` but I can't figure out what is going wrong.

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
AlexBHarley: I believe signing is non-deterministic yes.

packages/contractkit/src/wallets/signers/aws-hsm-signer.ts
AlexBHarley: This is a bit blocked on getting the `recoverKeyIndex` function to work... I have struggled with the implementation that we have in this repo already.

packages/contractkit/src/utils/ber-utils.ts
AlexBHarley: Added a check just in case

packages/contractkit/src/utils/azure-key-vault-client.ts
```diff
@@ -75,7 +70,10 @@ export class AzureKeyVaultClient {
     }
     const cryptographyClient = await this.getCryptographyClient(keyName)
     // @ts-ignore-next-line (ECDSA256 is not included in the client enum but is valid)
-    const signResult = await cryptographyClient.sign(this.SIGNING_ALGORITHM, message)
+    const signResult = await cryptographyClient.sign(
+      this.SIGNING_ALGORITHM,
+      new Uint8Array(message)
```
codyborn, line 70: This line was changed to fix the `value must be of type Uint8Array` issue. I believe some change in monorepo to our dependencies caused this to start failing.

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -214,3 +223,41 @@ export function decodeSig(sig: any) {
...
+      publicKey.eq(recoveredPublicKey)
+    )
+    if (publicKey.eq(recoveredPublicKey)) {
+      return i
+    }
+  }
+  throw new Error('Unable to generate recovery key from signature.')
+}
+
+export function getAddressFromPublicKey(publicKey: BigNumber): Address {
```
codyborn, line 250: @tkporter I tried consolidating this to use the util `publicKeyToAddress` method; however it introduced more problems. `ethUtil.pubToAddress` method does additional cleansing that allows it to succeed under more conditions.

packages/contractkit/src/utils/signature-utils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const secp256k1Curve = new EC('secp256k1')
+
+/**
+ * If the signature is in the "bottom" of the curve, it is non-canonical
+ * Non-canonical signatures are illegal in Ethereum and therefore the S value
+ * must be transposed to the lower intersection
+ * https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures
+ */
+export const makeCanonical = (S: BigNumber): BigNumber => {
```
codyborn, line 14: New method to move more common logic between AWS + Azure into signing-utils

codyborn (approved):
Approving changes, will wait for @tkporter to sign off before merging.

packages/contractkit/src/utils/ber-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+import { bigNumberToBuffer, bufferToBigNumber } from './signature-utils'
+
+export const toArrayBuffer = (b: Buffer): ArrayBuffer => {
+  return b.buffer.slice(b.byteOffset, b.byteOffset + b.byteLength)
+}
+
+export function publicKeyFromAsn1(b: Buffer): BigNumber {
+  const { result } = asn1.fromBER(toArrayBuffer(b))
+  const values = (result as asn1.Sequence).valueBlock.value
+  const value = values[1] as asn1.BitString
```
tkporter, line 15: just want to make sure it's safe to assume `values.length >= 2`

packages/contractkit/src/utils/ber-utils.ts
```diff
@@ -0,0 +1,54 @@
+import * as asn1 from 'asn1js'
+import { BigNumber } from 'bignumber.js'
+import { bigNumberToBuffer, bufferToBigNumber } from './signature-utils'
+
+export const toArrayBuffer = (b: Buffer): ArrayBuffer => {
+  return b.buffer.slice(b.byteOffset, b.byteOffset + b.byteLength)
+}
+
+export function publicKeyFromAsn1(b: Buffer): BigNumber {
```
tkporter, line 9: are these functions pretty generic or specific to expected responses from Azure/AWS?

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -213,3 +222,41 @@ export function decodeSig(sig: any) {
...
+): number {
+  for (let i = 0; i < 4; i++) {
+    const compressed = false
+    // Force types to be Uint8Array
+    const signatureArray = new Uint8Array(signature)
+    const hashArray = new Uint8Array(hash)
+    const recoveredPublicKeyByteArr = ecdsaRecover(signatureArray, i, hashArray, compressed)
+    const publicKeyBuff = Buffer.from(recoveredPublicKeyByteArr)
+    const recoveredPublicKey = bufferToBigNumber(publicKeyBuff)
+    console.log(
```
tkporter: not sure if `console.log` is used in contractkit, I think instead `debug` from line 14 is used

packages/contractkit/src/utils/azure-key-vault-client.ts
```diff
@@ -88,29 +81,17 @@ export class AzureKeyVaultClient {
...
-
-    // The Azure Signature MAY not be canonical, which is illegal in Ethereum
-    // thus it must be transposed to the lower intersection.
-    // https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures
-    const N = AzureKeyVaultClient.bufferToBigNumber(this.secp256k1Curve.curve.n)
-    if (!AzureKeyVaultClient.isCanonical(S, N)) {
-      debug('Canonicalizing signature')
-      S = N.minus(S)
-    }
+    const R = bufferToBigNumber(Buffer.from(rawSignature.slice(0, 32)))
```
tkporter: should we use `thirtyTwo` and `sixtyFour` here?

tkporter (approved):
Looks great! some very small things

packages/contractkit/src/utils/ber-utils.ts
codyborn: Good point. Either way we'll fail, but it's better to fail with a more descriptive error.

packages/contractkit/src/utils/signing-utils.ts
codyborn: good catch

packages/contractkit/src/utils/ber-utils.ts
codyborn: This method takes a BER encoded key with ASN1 serialization protocol and converts it to a BigNumber format, which our libraries are using as the standard. AWS sends the key in this format so although this method is generic, I don't expect other libraries to use this.

PR #4276, opened 2020-06-30
[PGPNP] Fix aws keystore response handling and add a test

jmrossy:
### Description

AWS secrets manager results results as a JSON string, even if you store the secrets as plaintext.
This fixes that and adds a test for the AWS key provider

### Tested

Locally with an AWS key store

### Backwards compatibility

Yes


packages/phone-number-privacy/signer/test/key-management/aws-key-provider.test.ts
```diff
@@ -0,0 +1,31 @@
+import { AWSKeyProvider } from '../../src/key-management/aws-key-provider'
+
+const mockKey = '010101010101010101010101010101010101010101010101010101010101010101010101'
+const mockResponse = { SecretString: `{"mockSecretKey":"${mockKey}"}` }
```
codyborn, line 4: Looks good! Can you add a test for SecretBinary as well?

codyborn (approved):

PR #4277, opened 2020-07-01
Add backward compatibility script for compiled contracts

### LINKED ISSUE #3961: Complete automated testing for smart contract backwards compatibility
Complete the WIP backwards compatibility testing work started here https://github.com/celo-org/celo-monorepo/pull/3893

hbandura:
### Description

Add a script to check backwards compatibility and semantic versioning for compiled contract folders, located in `packages/protocol/scripts/check-backward.ts`. The script allows for creating the compatibility report, calculating what the delta semantic version should be, inferring the final version from a given one plus the calculated delta, or just checking if the old and version number are correct.

The script traverses the truffle generated AST files to check for differences. KNOWN_ISSUE: An important catch is that, the compiled bytecode seems to be dependant on the source folder path. Therefore, if spotting 'deployedBytecode' differences is important, the artifacts should be compiled in the same path.

The clasification of what constitutes a major/minor/patch change is hardcoded in the class `CategorizerChangeVisitor` in `packages/protocol/lib/backward/ast-code.ts`.

Command outline
```
check-backward.ts <command>

Commands:
  check-backward.ts report     Generates a backward compatibility report
  check-backward.ts sem_check  Check if the semantic version change provided is
                               correct (exit code 0, or 1 otherwise)
  check-backward.ts sem_infer  Infer what the semantic version change should be
  check-backward.ts sem_delta  Generates the semantic version delta

Options:
  --version            Show version number                             [boolean]
  --exclude, -e        Contract name exclusion pattern                  [string]
  --old_contracts, -o  Old contracts build artifacts folder  [string] [required]
  --new_contracts, -n  Old contracts build artifacts folder  [string] [required]
  --old_version, -v    Semantic version string for the old contracts    [string]
  --new_version, -w    Semantic version string for the new contracts    [string]
  --output_file, -f    Destination file output for the compatibility report
                                                                        [string]
  --quiet, -q          Run in quiet mode (no logs)    [boolean] [default: false]
  --help, -h           Show help                                       [boolean]
```

Sample usage: 
```
$ ts-node check-backward.ts report -o ../build/contracts_old -n ../build/contracts
Instantiating old artifacts...Done
Instantiating new artifacts...Done
Running storage report...Done
Running code report...Done
Generating backward report...Done
Writing report to /var/folders/hz/dj7jgrkd5y988lyvzw79gnr40000gn/T/tmp-4604668YKxxON0baw ...Done
```

### Tested

End to end tests for the report(layout/code) incompatibilities are supplied in this pull request. The compilation of said test contracts is done during test execution, to avoid commiting the compiled artifacts in the repository. These are recompiled if a change is detected in the test resources, or if some time has passed since the last compilation (hardcoded to 1 day currently) to avoid unnecessary compilations in developer's day-to-day use.

Review hint: `.sol` files are test resources. If those are 'ignored', this PR consists of 10 files only.

### Related issues

- Fixes #3961 (Continues work from PR: #3893 )

### Backwards compatibility

This change only adds a new script, based on all-new typescript lib files. It should not raise any backward compatibility issues.


packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,392 @@
...
+
+const VISIBILITY_PUBLIC = 'public'
+const VISIBILITY_EXTERNAL = 'external'
+const CONTRACT_TYPE_CONTRACT = 'contract'
+const STORAGE_DEFAULT = 'default'
+
+const OUT_VOID_PARAMETER_STRING = 'void'
+
+export class ASTCodeCompatibilityReport {
+  changes: Change[];
```
codyborn: The general .ts style in our repo has been to not include semi-colons at the end of a line

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,392 @@
...
+  constructor(changes: Change[]) {
+    this.changes = changes;
+  }
+  public push(...changes: Change[]) {
+    this.changes.push(...changes)
+  }
+  public include(other: ASTCodeCompatibilityReport) {
+    this.push(...other.changes)
```
codyborn:
```suggestion
    this.changes.push(...other.changes)
```

packages/protocol/lib/backward/ast-code.ts
hbandura: Yeah I think I was a little bit confused, since I used semi-colons in class definitions but not in general code. I'll fix it

packages/protocol/lib/backward/ast-code.ts
hbandura: I think this is a left-over from my scala/java days. The reasoning usually on this type of code is that you leave space for someone to actually extend & modify the 'push' function, without needing to modify the 'pushAll' (include) function.

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,392 @@
...
+}
+
+export class MethodParametersChange extends MethodValueChange {
+  type = "MethodParameters";
+  accept<T>(visitor: ChangeVisitor<T>): T {
+    return visitor.visitMethodParameters(this);
+  }
+}
+
+export class MethodReturnChange extends MethodValueChange {
```
codyborn: It'd be helpful to include a UML diagram and some docs since there's alot being handled by inheritance.

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,392 @@
...
+    return visitor.visitMethodReturn(this);
+  }
+}
+
+export interface ASTError {
+  message: string,
+  wrapped: Error
+}
+
+export const createIndexByChangeType = (changes: Change[], categorizer: ChangeVisitor<ChangeType>): Change[][] => {
```
codyborn: Can you add header comments for exported methods? If you type `/***` VSCode should automatically start a boilerplate header.

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,392 @@
...
+
+const getSignature = (method: any): string => {
+  // This is used as the ID of a method
+  return `${method.selector}`
+}
+
+
+const createMethodIndex = (methods: any[]): any[] => {
+  const asPairs = methods.map(m => ({ [`${getSignature(m)}`]: m }))
+  return Object.assign({}, ...asPairs)
```
codyborn: What does the `...` do in this context?

packages/protocol/lib/backward/ast-code.ts
codyborn: oh gotcha

packages/protocol/lib/backward/utils.ts
```diff
@@ -0,0 +1,160 @@
...
+    this.storageReports = this.storageReports.filter(r => included(r.contract))
+  }
+}
+
+export const isValidVersion = (version: string): boolean => {
+  const v = version.split(".")
+  if (v.length !== 4) {
+    return false
+  }
+  return !isNaN(Number(v[0])) && !isNaN(Number(v[1])) && !isNaN(Number(v[2])) && !isNaN(Number(v[3]))
```
codyborn:
```suggestion
  const isNumber = (versionDigit) => !isNan(Number(versionDigit))
  return Number.every(isNumber)
```

packages/protocol/lib/backward/utils.ts
```diff
@@ -0,0 +1,160 @@
...
+  if (report.storage.length > 0) {
+    return `${V_STORAGE}+1.0.0.0`
+  }
+  if (report.major.length > 0) {
+    return `${V_STORAGE}.${V_MAJOR}+1.0.0`
+  }
+  if (report.minor.length > 0) {
+    return `${V_STORAGE}.${V_MAJOR}.${V_MINOR}+1.0`
+  }
+  if (report.minor.length > 0) {
```
codyborn:
```suggestion
  if (report.patch.length > 0) {
```

codyborn (changes requested):
Overall looks really good. Just added some small comments.

packages/protocol/lib/backward/utils.ts
hbandura: woops!

packages/protocol/lib/backward/ast-code.ts
hbandura: `...` expands an array. 
```
const arr = [e1, e2, e3]
myFunc(e1, e2, e3) == myFunc(...arr)
```
In this particular place, I'm creating an object given it's properties as an array of objects. `[{k1: v1}, {k2: v2}]` (joining all objects in one)

packages/protocol/lib/backward/ast-code.ts
hbandura: What's the best way to do that? Would ASCII doc suffice?

packages/protocol/lib/backward/ast-code.ts
hbandura: Will do

packages/protocol/lib/backward/utils.ts
hbandura: nice, love it

packages/protocol/test/backward/common.ts
```diff
@@ -0,0 +1,66 @@
...
+const recompileThresholdTime = 1000 * 60 * 60 * 24 // One day
+
+const ROOT_DIR = path.normalize(path.join(__dirname, '../../'))
+
+function exec(cmd: string) {
+  return execSync(cmd, { cwd: ROOT_DIR, stdio: 'inherit' })
+}
+
+// Return the latest modification time of a file in the given directory
+const getLatestUpdateTime = (dir) => {
```
mcortesi: i believe that our style guide favor the use of `function` for root level functions. Although, I personally like lambda style, for small or curryfied function

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,323 @@
...
+const CONTRACT_TYPE_CONTRACT = 'contract'
+const STORAGE_DEFAULT = 'default'
+
+const OUT_VOID_PARAMETER_STRING = 'void'
+
+export class ASTCodeCompatibilityReport {
+  changes: Change[]
+  constructor(changes: Change[]) {
+    this.changes = changes
+  }
```
mcortesi: You can also do:
```suggestion
  constructor(public readonly changes: Change[]) {
  }
```

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,323 @@
...
+  public push(...changes: Change[]) {
```
mcortesi: no need to specify `public` that's the default.

packages/protocol/lib/backward/ast-code.ts
mcortesi: +1 to keep the original `this.push()`

packages/protocol/lib/backward/ast-code.ts
mcortesi: better with `private` i guess

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,323 @@
...
+  onMethodAdded(change: MethodAddedChange): T
+  onMethodRemoved(change: MethodRemovedChange): T
+  onContractType(change: ContractTypeChange): T
+  onNewContract(change: NewContractChange): T
+  onDeployedBytecode(change: DeployedBytecodeChange): T
+}
+
+abstract class ContractChange implements Change {
+  type: string
+  contract: string
```
mcortesi: make them private?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,323 @@
...
+  constructor(contract: string) {
+    this.contract = contract
+  }
```
mcortesi:
```suggestion
  constructor(private readonly contract: string) {
  }
```

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,323 @@
...
+  getContract() {
+    return this.contract
+  }
+
+  abstract accept<T>(visitor: ChangeVisitor<T>): T
+}
+
+export class NewContractChange extends ContractChange {
+  type = "NewContract"
```
mcortesi: this looks strange.... type here is an instance variable. Not sure

packages/protocol/lib/backward/ast-code.ts
hbandura: ah, nice, didn't know about this, i'll replace it

packages/protocol/lib/backward/ast-code.ts
hbandura: I needed a discriminator to describe what the change type is in the serialized json. TBH I don't like it either but I'm not sure how to accomplish this in a different way

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
+// tslint:disable: max-classes-per-file
```
m-chrzan: Package naming: consider `compatibility/` or `backwards-compatibility/`.

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+ */
+export class DeployedBytecodeChange extends ContractChange {
+  type = "DeployedBytecode"
+  accept<T>(visitor: ChangeVisitor<T>): T {
+    return visitor.onDeployedBytecode(this)
+  }
+}
+
+/**
+ * The Kind / type of a contract changed. Kind / type examples are
```
m-chrzan: I would stick with one of "kind" or "type". Personally prefer "kind", as "type" can refer to variable types, and [Solidity's documentation](https://solidity.readthedocs.io/en/v0.5.11/structure-of-a-contract.html) does, in passing, refer to these as "kinds":

> There are also special kinds of contracts called libraries and interfaces.

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+    return visitor.onContractType(this)
+  }
+}
+
+/**
+ * Abstract implementation for the {@link Change} interface for
+ * method changes.
+ */
+abstract class MethodChange extends ContractChange {
+  constructor(contract: string, private readonly signature: string) {
```
m-chrzan: Is the parameter the new or the old signature? Or does it depend on the change? If it's always the same, would it be worth including this infomration in the variable name?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+ * The return parameters of a method changed.
+ */
+export class MethodReturnChange extends MethodValueChange {
+  type = "MethodReturn"
+  accept<T>(visitor: ChangeVisitor<T>): T {
+    return visitor.onMethodReturn(this)
+  }
+}
+
+const getSignature = (method: any): string => {
```
m-chrzan: Not sure myself, but does a typing for methods exist somewhere in web3?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+  // This is used as the ID of a method
+  return `${method.selector}`
```
m-chrzan: Does this need to be wrapped in `` `${...}` ``?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+}
+
+
```
m-chrzan: Extra newline

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+const createMethodIndex = (methods: any[]): any[] => {
```
m-chrzan: Do you think a comment explaining what a "method index" is be useful?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+  const asPairs = methods.map(m => ({ [`${getSignature(m)}`]: m }))
```
m-chrzan: Is `` `${...}` `` necessary if we know this function returns a string?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+
+const mergeReports = (reports: ASTCodeCompatibilityReport[]): ASTCodeCompatibilityReport => {
+  const report = new ASTCodeCompatibilityReport([])
+  reports.forEach((r: ASTCodeCompatibilityReport): void => {
+    report.include(r)
+  })
+  return report
+}
+
+const parametersSignature = (parameters: any[]): string => {
```
m-chrzan: Could this be typed better?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+  const state1 = m1.stateMutability
+  const state2 = m2.stateMutability
+  if (state1 !== state2) {
+    report.push(new MethodMutabilityChange(contract, signature, state1, state2))
+  }
+  return report
+}
+
+const getCheckableMethodsFromAST = (contract: ContractAST, id: string): any[] => {
+  const checkableMethods = (method: any) => method.visibility === VISIBILITY_EXTERNAL || method.visibility === VISIBILITY_PUBLIC
```
m-chrzan: Naming: should this be `methodCheckable`? (checks if a given method is checkable)

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+  // Need to manually use ContractAST since its internal use in ZContract
+  // does not pass the artifacts parameter to the constructor, therefore
+  // forcing a reloading of BuildArtifacts.
+  const newAST = new ContractAST(newContract, newArtifacts)
+  const newKind = newAST.getContractNode().contractKind
+  if (oldContract === null) {
+    if (newKind === CONTRACT_TYPE_CONTRACT) {
+      return new ASTCodeCompatibilityReport([new NewContractChange(contractName)])
+    } else {
+      // New contract added of a non-contract type (library/interface)
```
m-chrzan: type -> kind

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+    }
+  }
+
+  // Name sanity check
+  if (oldContract.schema.contractName !== contractName) {
+    throw new Error(`Contract names should be equal: ${oldContract.schema.contractName} !== ${contractName}`)
+  }
+
+  const oldAST = new ContractAST(oldContract, oldArtifacts)
+  const kind = oldAST.getContractNode().contractKind
```
m-chrzan: Naming nit: for consistency, consider `oldKind`.

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+import { makeZContract } from '@celo/protocol/lib/backward/internal'
+import {
+  BuildArtifacts,
+  Contract as ZContract
+} from '@openzeppelin/upgrades'
+import ContractAST from '@openzeppelin/upgrades/lib/utils/ContractAST'
+
+export const VISIBILITY_PUBLIC = 'public'
+export const VISIBILITY_EXTERNAL = 'external'
+const CONTRACT_TYPE_CONTRACT = 'contract'
```
m-chrzan: TYPE -> KIND?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+ * Visitor pattern implementation for the {@link Change} hierarchy.
+ */
+export interface ChangeVisitor<T> {
+  onMethodMutability(change: MethodMutabilityChange): T
+  onMethodParameters(change: MethodParametersChange): T
+  onMethodReturn(change: MethodReturnChange): T
+  onMethodVisibility(change: MethodVisibilityChange): T
+  onMethodAdded(change: MethodAddedChange): T
+  onMethodRemoved(change: MethodRemovedChange): T
+  onContractType(change: ContractTypeChange): T
```
m-chrzan: Type -> Kind (both method and class)?

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  const oldAST = new ContractAST(oldContract, oldArtifacts)
+  const kind = oldAST.getContractNode().contractKind
+  if (kind !== newKind) {
+    // different contract kind (library/interface/contract)
+    return new ASTCodeCompatibilityReport([new ContractTypeChange(contractName, kind, newKind)])
+  }
+
+  const report = doASTCompatibilityReport(contractName, oldAST, newAST)
+  // Check deployed byte code change
```
m-chrzan: Does this deal with differing metadata? Potentially check out `scripts/truffle/upgrade.ts#stripMetadata` for how I've done this before, although we might want a more robust check here.

m-chrzan (commented):
Looked at ast-code so far.

packages/protocol/lib/backward/ast-code.ts
hbandura: it's the "same" since the signature is what I use as the ID of the method. I'll add a comment here explaining that too

packages/protocol/lib/backward/ast-code.ts
hbandura: Not sure either, I'll check

packages/protocol/lib/backward/ast-code.ts
hbandura: This is a leftover, I was firstly doing contractName + selector. I'll change it

packages/protocol/lib/backward/ast-code.ts
hbandura: I'll check that out, what that does strip metadata do? The only issue I had so far with metadata is that the deployedBytecode seems to depend on the source folder path, which was annoying to make the test resources compile properly

m-chrzan: `stripMetadata` erases the part of the bytecode where the metadata hash is stored, so by using it on both bytecodes, we can ensure that bytecode is exactly the same modulo the metadata we don't care about. 

More info here: https://solidity.readthedocs.io/en/v0.5.3/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode
 
Also, just noticed that newer versions of Solidity store this differently. For example 0.5.11 has a compiler version also appended in the metadata, and in e.g. 0.6.10, it's an IPFS hash instead of a swarm hash that's stored. So make sure to take that into account if you end up doing something with this.

packages/protocol/lib/backward/ast-code.ts
hbandura: Found one but it's definitely not the same structure as the one returned by the library

hbandura: I added my own defined types for the parts of the method I use

packages/protocol/lib/backward/ast-code.ts
hbandura: This seems like it could probably need its own research. What do you think if we leave it for a follow up issue?

packages/protocol/lib/backward/ast-code.ts
hbandura: the second one is too long IMO. And I don't have any strong opinions about either `backward` or `compatibility`. Since one is already implemented I'd leave that, unless there is any strong argument for the other. What do you think?

packages/protocol/lib/backward/ast-code.ts
hbandura: I added docs for each class, and I think the implementation is a bit simpler now. Do you think that a UML diagram is still necessary?

packages/protocol/lib/backward/ast-code.ts
m-chrzan: Without context, I don't think I would have much of an idea of what I'm looking at when I see `lib/backward`, whereas `lib/compatibility` is much clearer in the library's intent.

packages/protocol/lib/backward/ast-code.ts
```diff
@@ -0,0 +1,425 @@
...
+
+const OUT_VOID_PARAMETER_STRING = 'void'
+
+// Exported classes
+
+/**
+ * A compatibility report with all the detected changes from two compiled
+ * contract folders.
+ */
+export class ASTCodeCompatibilityReport {
```
m-chrzan: Another naming question: would it be more precise to replace `AST` with `ABI` or `API` in names related to this library?

packages/protocol/lib/backward/report.ts
```diff
@@ -0,0 +1,176 @@
...
+    public readonly patch: Change[]) {}
+
+
+
+  /**
+   * @returns a mapping {contract name => {@link CategorizedChanges}}
+   */
+  explode = (): CategorizedChangesIndex => {
+    const builders: {[k: string]: CategorizedChangesBuilder} = {}
+    const builder = (k: string) => {
```
m-chrzan: Can `k` have a more meaningful name throughout?

packages/protocol/lib/backward/report.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  constructor(
+    public readonly storage: ASTStorageCompatibilityReport[],
+    public readonly major: Change[],
+    public readonly minor: Change[],
+    public readonly patch: Change[]) {}
+
+
+
```
m-chrzan: Extra newlines

packages/protocol/lib/backward/report.ts
```diff
@@ -0,0 +1,176 @@
...
+  /**
+   * @returns a mapping {contract name => {@link CategorizedChanges}}
+   */
+  explode = (): CategorizedChangesIndex => {
```
m-chrzan: Not sure about this name, not clear from it what this function does, but don't have any suggestions off the top of my head.

m-chrzan: Maybe `makeIndex`?

packages/protocol/lib/backward/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+    logFunction("Done\n")
+    // Run reports
+    logFunction("Running storage report...")
+    const storage = reportLayoutIncompatibilities(artifacts1, artifacts2)
+    logFunction("Done\n")
+    logFunction("Running code report...")
+    const code = reportASTIncompatibilities(artifacts1, artifacts2)
+    logFunction("Done\n")
+  
+    const excludeRegexp: RegExp = exclude? new RegExp(exclude) : null
```
m-chrzan: Style nit: by convention I think space before `?`?

packages/protocol/lib/backward/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+    const fullReports = new ASTReports(code, storage).excluding(excludeRegexp)
+    
+    logFunction("Generating backward report...")
```
m-chrzan: Wording: consider "Generating backwards compatibility report".

packages/protocol/lib/backward/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+    const artifacts1 = instantiateArtifacts(oldArtifactsFolder)
+    logFunction("Done\n")
+    logFunction("Instantiating new artifacts...")
+    const artifacts2 = instantiateArtifacts(newArtifactsFolder)
+    logFunction("Done\n")
+    // Run reports
+    logFunction("Running storage report...")
+    const storage = reportLayoutIncompatibilities(artifacts1, artifacts2)
+    logFunction("Done\n")
+    logFunction("Running code report...")
```
m-chrzan: Wording nit: maybe "Running X *analysis*"?

packages/protocol/lib/backward/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+export function instantiateArtifacts(buildDirectory: string): BuildArtifacts {
+  // Check if all jsons in the buildDirectory are valid artifacts,
+  // otherwise getBuildArtifacts fail with the enigmatic
+  // "Cannot read property 'absolutePath' of undefined"
+  ensureValidArtifacts(Contracts.listBuildArtifacts(buildDirectory))
+  try {
+    return getBuildArtifacts(buildDirectory)
+  } catch (error) {
+    console.error(`ERROR: could not create BuildArtifacts on directory '${buildDirectory}`)
+    process.exit(10002)
```
m-chrzan: Why this error code?

packages/protocol/lib/backward/version.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+}
+
+export enum Delta {
+  None = "=",
+  Increment = "+1",
+  Reset = "0"
+}
+
+export class DeltaUtil {
```
m-chrzan: Does this need to be a class?

packages/protocol/lib/backward/version.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export class ContractVersionDelta {
+
+  static fromChanges = (
+    storageChanged: boolean,
+    majorChanged: boolean,
+    minorChanged: boolean,
+    patchChanged: boolean): ContractVersionDelta => {
+    const r = Delta.Reset
+    const n = Delta.None
```
m-chrzan: Style opinion: personally would prefer just explicitly writing out `Delta.Reset`/`Delta.None` throughout over creating a  local alias like this.

packages/protocol/test/backward/ast-code.ts
```diff
@@ -0,0 +1,131 @@
...
+  MethodAddedChange,
+  MethodMutabilityChange,
+  MethodRemovedChange,
+  MethodReturnChange,
+  MethodVisibilityChange,
+  NewContractChange,
+  reportASTIncompatibilities,
+} from '@celo/protocol/lib/backward/ast-code'
+import { getTestArtifacts } from '@celo/protocol/test/backward/common'
+import { assert } from 'chai'
```
m-chrzan: Does this need to be explicitly imported?

packages/protocol/test/backward/common.ts
```diff
@@ -0,0 +1,66 @@
...
+    // Build is too old
+    return true
+  }
+  return false
+}
+
+export function getTestArtifacts(caseName: string) {
+  const back = './test/resources/backward'
+  const srcDirectory = `${back}/contracts_${caseName}`
+  const buildDirectory = `${back}/build/b_${caseName}`
```
m-chrzan: Is this the correct path? (`build/b_...`)

packages/protocol/test/backward/version.ts
```diff
@@ -0,0 +1,158 @@
+import {
+  ContractVersion,
+  ContractVersionDelta,
+  Delta,
+  DeltaUtil,
+} from '@celo/protocol/lib/backward/version'
+import { assert } from 'chai'
+
+describe('#version()', () => {
```
m-chrzan: Is there a `version()` function we're testing? Maybe just have top-level `describe(<class>)` for the different classes you're testing from this library.

packages/protocol/test/backward/version.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  describe('ContractVersionDelta.fromChanges', () => {
+    it('sets properly the storage Delta', () => {
+      const cvd = ContractVersionDelta.fromChanges(true, false, true, true)
+      assert.equal(cvd.storage, Delta.Increment)
+      assert.equal(cvd.major, Delta.Reset)
+      assert.equal(cvd.minor, Delta.Reset)
+      assert.equal(cvd.patch, Delta.Reset)
+
+      const cvd2 = ContractVersionDelta.fromChanges(false, false, true, true)
```
m-chrzan: This feels like a new `it` test case since it doesn't use any of the previous setup.

packages/protocol/test/backward/version.ts
```diff
@@ -0,0 +1,158 @@
...
+      assert.notEqual(cvd2.major, Delta.Reset)
+    })
+
+    it('sets properly the major Delta', () => {
+      const cvd = ContractVersionDelta.fromChanges(false, true, true, true)
+      assert.equal(cvd.major, Delta.Increment)
+      assert.equal(cvd.minor, Delta.Reset)
+      assert.equal(cvd.patch, Delta.Reset)
+
+      const cvd2 = ContractVersionDelta.fromChanges(false, false, true, true)
```
m-chrzan: Ditto throughout.

packages/protocol/test/backward/version.ts
```diff
@@ -0,0 +1,158 @@
...
+describe('#version()', () => {
+  describe('ContractVersion', () => {
+    it('isValid fails on invalid string', () => {
+      assert.isFalse(ContractVersion.isValid('1.2.3'))
+      assert.isFalse(ContractVersion.isValid('1.2'))
+      assert.isFalse(ContractVersion.isValid('1.2.3.5.'))
+      assert.isFalse(ContractVersion.isValid('test'))
+      assert.isFalse(ContractVersion.isValid('1.a.3.4'))
+    })
+    it('isValid true on valid string', () => {
```
m-chrzan: Style consistency: ewline before `it`.

packages/protocol/test/backward/version.ts
```diff
@@ -0,0 +1,158 @@
...
+      assert.isTrue(ContractVersion.isValid('1.2.3.4'))
+      assert.isTrue(ContractVersion.isValid('-1.0.9.787'))
+      assert.isTrue(ContractVersion.isValid('10.0.224.9'))
+    })
+  })
+  describe('with a new ContractVersion instance', () => {
```
m-chrzan: Newline before `describe` (and similar throughout; may have also missed the same stuff in previous test file?)

packages/protocol/test/backward/version.ts
```diff
@@ -0,0 +1,158 @@
...
+  describe('DeltaUtil', () => {
+    it('applyToNumber() works properly', () => {
+      assert.equal(DeltaUtil.applyToNumber(Delta.Reset, 7), 0)
+      assert.equal(DeltaUtil.applyToNumber(Delta.Increment, 9), 10)
+      assert.equal(DeltaUtil.applyToNumber(Delta.None, 11), 11)
+    })
+  })
+
+  describe('with a new ContractVersionDelta instance', () => {
+    it('properties are well set', () => {
```
m-chrzan: Wording (here and in other cases): try to make these read as a sentence, like "it *does something*". So here maybe something like `it('sets version numbers correctly')`.

m-chrzan (commented):
Looks really good! Mostly style comments.

Also, making note of the change we talked about offline: versions need to be verified per contract rather than one for the whole directory.

packages/protocol/lib/backward/ast-code.ts
hbandura: renamed to compatibility

packages/protocol/lib/backward/ast-code.ts
hbandura: I did consider this too. As a usage perspective, we are checking the API. But we are extracting the information from the AST. I'm guessing that, if we were to model this in more detail, I'd have an interface to be "CodeComp..Report" or "APICodeComp...Report", but this actual concrete class be "ASTCode...Report". Since this abstraction is not needed, I just left it as "AST...".

packages/protocol/lib/backward/report.ts
hbandura: I didn't mind much about it since it's just an internal variable for the method, but if it's not clear enough I'll rename it to 'contract' (it's actually the contract name, but I think contract will suffice)

packages/protocol/lib/backward/report.ts
hbandura: Well, it's not just making an index. It's a big CategorizedChanges which gets split/exploded into different CategorizedChanges for each contract, with an Index to find them. So it's both a makeIndex but also for NEW CategorizedChanges objects.

I'll change it to `byContract`. The syntax maybe will be more clear:

```
myContractChanges = categorizedChanges.byContract()[myContract]
```

packages/protocol/lib/backward/ast-code.ts
codyborn: Thanks! I think the comments really help.

codyborn (approved):
Looks good! Excited to see this in practice

packages/protocol/lib/compatibility/ast-code.ts
```diff
@@ -0,0 +1,425 @@
...
+// Types used by the Method interface
+// OpenZep uses 'any' for methods so we
+// define our own here
+
+interface TypeDescriptions {
+  typeString: string
+}
+
+interface Parameter {
+  storageLocation: string
```
asaj: Should this be an enum?

packages/protocol/lib/compatibility/ast-code.ts
```diff
@@ -0,0 +1,425 @@
...
+  typeDescriptions: TypeDescriptions
+}
+
+interface Parameters {
+  parameters: Parameter[]
+}
+interface Method {
+  selector: string
+  visibility: string
```
asaj: Ditto

packages/protocol/lib/compatibility/ast-code.ts
```diff
@@ -0,0 +1,425 @@
...
+  stateMutability: string
+  parameters: Parameters
+  returnParameters: Parameters
+}
+
+// Implementation
+
+function getSignature(method: Method): string {
```
asaj, line 266: nit: Why signature and not selector? I'm assuming this is the 4 bytes function selector?

asaj: Following up on this after reading `checkMethodCompatibility`, I would now assume this *isn't* the 4 byte function selector, as `checkMethodCompatibility` expects the signature to not have changed before checking parameter and return type compatibility. So I'm not clear on what `getSignature` is supposed to return.

packages/protocol/lib/compatibility/ast-code.ts
```diff
@@ -0,0 +1,425 @@
...
+  // Need to manually use ContractAST since its internal use in ZContract
+  // does not pass the artifacts parameter to the constructor, therefore
+  // forcing a reloading of BuildArtifacts.
+  const newAST = new ContractAST(newContract, newArtifacts)
+  const newKind = newAST.getContractNode().contractKind
+  if (oldContract === null) {
+    if (newKind === CONTRACT_KIND_CONTRACT) {
+      return new ASTCodeCompatibilityReport([new NewContractChange(contractName)])
+    } else {
+      // New contract added of a non-contract kind (library/interface)
```
asaj, line 405: How does this tool handle changes to linked (rather than inlined) libraries?

packages/protocol/lib/compatibility/categorizer.ts
```diff
@@ -0,0 +1,59 @@
...
+  Change,
+  ChangeVisitor,
+  ContractKindChange, DeployedBytecodeChange, MethodAddedChange,
+  MethodMutabilityChange, MethodParametersChange, MethodRemovedChange,
+  MethodReturnChange, MethodVisibilityChange, NewContractChange,
+  VISIBILITY_EXTERNAL, VISIBILITY_PUBLIC
+} from '@celo/protocol/lib/compatibility/ast-code'
+
+/**
+ * Change type categories according to semantic versioning standards
```
asaj, line 11: nit: Consider adding a comment as to why the storage version type isn't relevant here

packages/protocol/lib/compatibility/categorizer.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+  changes.map(c => byCategory[c.accept(categorizer)].push(c))
+  return byCategory
+}
+
+/**
+ * Default implementation of {@link Categorizer}, where:
+ *  Major:
+ *    Mutability, Params, Return Params, Method Removed, Contract type,
+ *    Visibilty changes
```
asaj: nit: typo

asaj (commented):
This tool is really incredible, amazing work @hbandura and @m-chrzan  👏 👏 👏 

Minor nits, only comment blocking approval is the question RE linked libraries

packages/protocol/lib/compatibility/categorizer.ts
hbandura: The `categorizer` file contains standard semantic versioning assumptions, where storage is not one of them. Maybe what needs more commenting is why the storage is there in `version`, but here I think it could be weird

packages/protocol/lib/compatibility/ast-code.ts
hbandura: Actually it is the selector, but in its full string form. As in "${methodName}(${methodArgs})". In an earlier draft of the implementation I was prepending "${contractName}" to the 'signature'. But overall I just keep using signature as a synonim for selector. Maybe this ends up being confusing?

packages/protocol/lib/compatibility/ast-code.ts
hbandura: Great question, can you point me to an example of using a linked library?

packages/protocol/lib/compatibility/ast-code.ts
hbandura: Probably, yes. I just added the interface to make the typing a little bit better. It's a rabbit hole how deep one can go with it. The reason why I didn't add an enum was that I didn't want the script to fail if there's any value not considered in the enum. I'll add it

packages/protocol/lib/compatibility/ast-code.ts
hbandura: Ok so since linked libraries are only a name placeholder, if the linked library name doesn't change, the tool can't spot any change, since, the bytecode is exactly the same, and the change would only be a deployment change.

For our specific needs in cLabs, we may want to do some extra work here to manage versioning for contracts that depend on modified linked libraries.

packages/protocol/test/compatibility/version.ts
```diff
@@ -0,0 +1,165 @@
...
+      assert.isFalse(ContractVersion.isValid('1.2.3'))
+      assert.isFalse(ContractVersion.isValid('1.2'))
+      assert.isFalse(ContractVersion.isValid('1.2.3.5.'))
+      assert.isFalse(ContractVersion.isValid('test'))
+      assert.isFalse(ContractVersion.isValid('1.a.3.4'))
+    })
+
+    it('returns true on valid string for isValid', () => {
+      assert.isTrue(ContractVersion.isValid('1.2.3.4'))
+      assert.isTrue(ContractVersion.isValid('-1.0.9.787'))
```
m-chrzan: Should this be a valid version number?

packages/protocol/test/compatibility/version.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  describe('ContractVersionDelta.fromChanges', () => {
+    it('sets properly the storage Delta', () => {
+      const cvd = ContractVersionDelta.fromChanges(true, false, true, true)
+      assert.equal(cvd.storage, Delta.Increment)
+      assert.equal(cvd.major, Delta.Reset)
+      assert.equal(cvd.minor, Delta.Reset)
+      assert.equal(cvd.patch, Delta.Reset)
+    })
+    it('sets properly the storage Delta', () => {
```
m-chrzan: Style nit: newline between `it`s.

packages/protocol/test/compatibility/version.ts
```diff
@@ -0,0 +1,165 @@
...
+      assert.notEqual(cvd2.major, Delta.Reset)
+    })
+
+    it('sets properly the major Delta', () => {
+      const cvd = ContractVersionDelta.fromChanges(false, true, true, true)
+      assert.equal(cvd.major, Delta.Increment)
+      assert.equal(cvd.minor, Delta.Reset)
+      assert.equal(cvd.patch, Delta.Reset)
+    })
+    it('sets properly the major Delta', () => {
```
m-chrzan: Ditto here and throughout

packages/protocol/test/compatibility/version.ts
m-chrzan: Language nit: "properly sets ..." (here and throughout the file)

packages/protocol/test/compatibility/version.ts
hbandura: I think, according to semver, no. But I wasn't sure if I should restrict it. What do you think?

PR #4278, opened 2020-07-01
Create common PGPNP lib

codyborn:
### Description

Creating a common library for the combiner and signer. This is important for having a shared interface that is kept in-sync across both services. Additionally, it allows for code-reuse and reduces duplication.

### Tested

Tested by integrating with combiner + signer (next PR).

### Backwards compatibility

The API object, SignMessageResponse, only adds fields:
  `performedQueryCount?: number`
  `totalQuota?: number`
and makes `signature` optional


jmrossy (commented):
LGTM so far! I think you need to update the monorepo dependency graph whenever you add a new package (see `scripts/update_dependency_graph.ts` and the others scripts in there)

jmrossy (approved):

PR #4279, opened 2020-07-01
[Wallet] Currency representation consistency and fix Gold Education

i1skn:
### Description

* Currency representation consistency in Drawer and Cash In/Out screen
* Add hamburger icon to the Gold Education, when user lands on it from Drawer, so it is possible to go away without finishing the whole tutorial.

### Tested

iOS and Android

### Backwards compatibility

Yes

### Screenshots
RhwQWEtRtShdMmL5PXMbBfsUac6bf5nD

### DIFF
--- a/packages/mobile/src/components/CurrencyDisplay.tsx
+++ b/packages/mobile/src/components/CurrencyDisplay.tsx
@@ -6,6 +6,7 @@ import { StyleProp, StyleSheet, Text, TextStyle, View } from 'react-native'
 import { MoneyAmount } from 'src/apollo/types'
 import { useExchangeRate as useGoldToDollarRate } from 'src/exchange/hooks'
 import { CURRENCIES, CURRENCY_ENUM } from 'src/geth/consts'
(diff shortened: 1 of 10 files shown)


packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -46,6 +47,7 @@ interface Props {
   showLocalAmount?: boolean
   showExplicitPositiveSign: boolean // shows '+' for a positive amount when true (default is false)
   formatType: FormatType
+  hideFullCurrencyName: boolean
```
jmrossy, line 50: There's lots of props already, I wonder if we should assign some default values to some like this one?

jmrossy (approved):
One suggestions but otherwise LGTM

packages/mobile/src/components/CurrencyDisplay.tsx
cmcewen: check below 😄

PR #4280, opened 2020-07-01
[Wallet] Add retry logic to verification reveal failure

annakaz:
### Description

Add retry logic to verification reveal failure when `No incomplete attestation found` error returned by attestation service

Currently, when a reveal fails, we assume that either it was already revealed in the past (Verification Failure modal shown with option to enter codes) or it is an irrecoverable failure (Verification Failure modal with no option to enter codes). However @nambrot pointed out that a reveal can fail if an attestation service has not yet received the block where it is made responsible for the attestation, in which case the reveal should retry, as time will have passed for the attestation service to receive the relevant block. 

More context: https://celo-org.slack.com/archives/CJXJ1EV0D/p1593590635047500

### Other changes

Add reveal error message to analytics

### Tested

Not yet tested

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/identity/verification.ts
+++ b/packages/mobile/src/identity/verification.ts
@@ -512,18 +512,43 @@ function* tryRevealPhoneNumber(
     )
     if (!response.ok) {
       const body = yield response.json()
-      Logger.error(TAG + '@tryRevealPhoneNumber', `Reveal response not okay: ${body.error}`)
-      ValoraAnalytics.track(AnalyticsEvents.verification_reveal_error, {
-        issuer,
-        statusCode: response.status,
-      })
-      throw new Error(
-        `Error revealing to issuer ${attestation.attestationServiceURL}. Status code: ${response.status}`
-      )
+
+      if (body.error.includes('No incomplete attestation found')) {
+        // Retry as attestation service might not yet have received the block where it was made responsible for an attestation
+        yield call(
+          [attestationsWrapper, attestationsWrapper.revealPhoneNumberToIssuer],
+          phoneHashDetails.e164Number,
+          account,
+          attestation.issuer,
+          attestation.attestationServiceURL,
+          phoneHashDetails.salt,
+          smsAppSig
+        )
+        Logger.debug(
+          TAG + '@tryRevealPhoneNumber',
+          `Revealing for issuer ${issuer} successful with retry`
+        )
+        ValoraAnalytics.track(AnalyticsEvents.verification_revealed_attestation, {
+          retryRequired: true,
+          issuer,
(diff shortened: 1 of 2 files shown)


packages/mobile/src/identity/verification.ts
```diff
@@ -512,18 +512,43 @@ function* tryRevealPhoneNumber(
...
-        issuer,
-        statusCode: response.status,
-      })
-      throw new Error(
-        `Error revealing to issuer ${attestation.attestationServiceURL}. Status code: ${response.status}`
-      )
+
+      if (body.error.includes('No incomplete attestation found')) {
+        // Retry as attestation service might not yet have received the block where it was made responsible for an attestation
+        yield call(
```
nambrot, line 518: I would add a sleep here as it might take a second or two for the service to catch up (since we suspect it to be behind)

packages/mobile/src/identity/verification.ts
```diff
@@ -512,18 +512,43 @@ function* tryRevealPhoneNumber(
...
+          `Revealing for issuer ${issuer} successful with retry`
+        )
+        ValoraAnalytics.track(AnalyticsEvents.verification_revealed_attestation, {
+          retryRequired: true,
+          issuer,
+          duration: Date.now() - startTime,
+        })
+      } else {
+        Logger.error(TAG + '@tryRevealPhoneNumber', `Reveal response not okay: ${body.error}`)
+        ValoraAnalytics.track(AnalyticsEvents.verification_reveal_error, {
```
nambrot, line 538: This branch only checks if we receive a valid response with a 400+ status code. As far as I can see, we also need to track in the catch below for other exceptions such as lower-level networking?

packages/mobile/src/identity/verification.ts
jmrossy: +1, a `yield delay(3000)` here makes sense I think

packages/mobile/src/identity/verification.ts
```diff
@@ -512,18 +512,43 @@ function* tryRevealPhoneNumber(
...
+          phoneHashDetails.e164Number,
+          account,
+          attestation.issuer,
+          attestation.attestationServiceURL,
+          phoneHashDetails.salt,
+          smsAppSig
+        )
+        Logger.debug(
+          TAG + '@tryRevealPhoneNumber',
+          `Revealing for issuer ${issuer} successful with retry`
```
jmrossy, line 529: We need to check the response status again right? We can't assume it worked just because there was no error thrown

annakaz:
Closing as these changes will go in with @jmrossy 's proxy setup for verification

PR #4298, opened 2020-07-02
Allow signer to invoke Attestations withdraw function for fee collection

### LINKED ISSUE #4296: Attestation fees cannot be withdrawn by issuers using ReleaseGold contract
### Expected Behavior

Attestation fees paid to an issuer should be accessible by that issuer.

### Current Behavior

If the the issuer is the address of a ReleaseGold contract, as is the case for most validators, then there is no way to withdraw the accumulated fees.

Two solutions are available for this issue:
* Allow the attestation signer to withdraw fees on behalf of the issuer.
* Update all deployed ReleaseGold contracts to add a wrapper function for withdrawing fees from the attestation contract

yorhodes:
### Description

Allows the attestations signer to invoke `Attestations.withdraw` on behalf of the issuer for Validators using `ReleaseGold` as the account. Funds are still transferred to the `issuer`.

### Other changes

### Tested

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/4296

### Backwards compatibility


packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -302,6 +302,15 @@ contract Attestations is
...
 
+  function withdrawToSigner(address token) external {
+    address issuer = getAccounts().attestationSignerToAccount(msg.sender);
+    uint256 value = pendingWithdrawals[token][issuer];
+    require(value > 0, "value was negative/zero");
+    pendingWithdrawals[token][issuer] = 0;
+    require(IERC20(token).transfer(msg.sender, value), "token transfer failed");
+    emit Withdrawal(msg.sender, token, value);
+  }
+
```
nategraf: I think we should add logic to the `withdraw` function to allow the attestation signer to withdraw instead of adding a new function. But I can also see the argument for doing with with a new function instead

nategraf (commented):
💯 for jumping on this

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -302,6 +302,15 @@ contract Attestations is
     emit Withdrawal(msg.sender, token, value);
   }
 
+  function withdrawToSigner(address token) external {
+    address issuer = getAccounts().attestationSignerToAccount(msg.sender);
+    uint256 value = pendingWithdrawals[token][issuer];
+    require(value > 0, "value was negative/zero");
+    pendingWithdrawals[token][issuer] = 0;
+    require(IERC20(token).transfer(msg.sender, value), "token transfer failed");
```
nambrot: @nategraf Suggested that we should move the funds to the `issuer` to maintain our attempt at making the account  key the one that can move balances, and not the signers

asaj (approved):

asaj:
FYI you'll probably need to add version number 1.1.0.1, see https://github.com/celo-org/celo-monorepo/pull/4523 for examples

PR #4299, opened 2020-07-02
Specify a context when deploying oracles, add 3rd mainnet cluster

tkporter:
### Description

This does away with the "primary" and "secondary" terminology when referring to an oracle cluster, and instead introduces "contexts" for oracle deployments. An oracle context can really have any value, but in practice it should be: `<cloud provider>-<region>`, eg `azure-westus2` or `azure-westeurope`. This is what should be passed into celotool, for example:
```
celotooljs deploy initial oracle -e oracledev --context azure-eastus
```

Behind the scenes, this context is coerced into being all caps and using underscores instead of dashes. Env variables for a an oracle context are expected to be present and have a prefix of the form `ORACLE_<context in caps and underscores>_.*` (eg `ORACLE_AZURE_EASTUS_AZURE_SUBSCRIPTION_ID`). To do this, `DynamicEnvVar`s are introduced, which are a way of dynamically generating an environment variable name. For example:
```
ORACLE_AZURE_SUBSCRIPTION_ID = 'ORACLE_{{ oracleContext }}_AZURE_SUBSCRIPTION_ID',
``` 

In this case, `{{ oracleContext }}` is replaced at runtime with the value of whatever the oracleContext is.

This also moves oracle full node env vars to be context-specific, so that one context can have more/less full nodes or more/less storage per node.

### Other changes

* Adds a 3rd cluster in Azure that has full nodes constantly running as an emergency backup.
* Adds the ability to optionally specify which resource group a key vault is located in: `<address>:<key vault name>:<resource group (optional)>`

### Tested

Deployed!

### Related issues

n/a

### Backwards compatibility

Env variables must be updated


packages/celotool/src/lib/oracle.ts
```diff
@@ -341,45 +332,83 @@ function getOracleAzureIdentityName(keyVaultName: string, address: string) {
...
+  {})
 }
 
 /**
  * Switches to the AKS cluster associated with the given context
  */
-export function switchToAzureContextCluster(celoEnv: string, context: OracleAzureContext) {
-  const azureClusterConfig = getAzureClusterConfig(context)
+export function switchToAzureContextCluster(celoEnv: string, oracleContext: string) {
+  const validOracleContexts = fetchEnv(envVar.ORACLE_CONTEXTS)
```
jasonrudianto1: I like having this check! I can see this being used for a switchToAwsContextCluster function. Perhaps lines 381-386 can be made into a new "isValidOracleContext" function for reusability and possibly better readability?

jasonrudianto1 (approved):
Looks great to me!

packages/celotool/src/lib/env-utils.ts
```diff
@@ -216,6 +218,16 @@ export function isValidCeloEnv(celoEnv: string) {
   return new RegExp('^[a-z][a-z0-9]*$').test(celoEnv)
 }
 
+/**
+ * Replaces a dynamic env var's template strings with values from an object.
+ * Returns the name of the env var.
+ */
+export function getDynamicEnvVarName(dynamicEnvVar: DynamicEnvVar, templateValues: any) {
+  return Object.keys(templateValues).reduce((agg: string, templateKey: string) => {
```
codyborn, line 234: It took me a min to figure out what was going on here 😄 Looks good! Maybe you can include usage patterns in the comments for examples.

packages/celotool/src/lib/oracle.ts
```diff
@@ -47,40 +42,24 @@ interface OracleConfig {
...
-    clusterName: envVar.ORACLE_SECONDARY_AZURE_KUBERNETES_CLUSTER_NAME,
-  },
+const oracleContextAzureClusterConfigDynamicEnvVars: { [k in keyof AzureClusterConfig]: DynamicEnvVar } = {
+  subscriptionId: DynamicEnvVar.ORACLE_AZURE_SUBSCRIPTION_ID,
+  tenantId: DynamicEnvVar.ORACLE_AZURE_TENANT_ID,
+  resourceGroup: DynamicEnvVar.ORACLE_AZURE_KUBERNETES_RESOURCE_GROUP,
+  clusterName: DynamicEnvVar.ORACLE_KUBERNETES_CLUSTER_NAME,
+}
+
+interface OracleIdentityConfig {
```
codyborn: Given that only one of these will be defined, perhaps it's cleaner to define two separate objects.
```
interface KeyVaultIdentityConfig {
  addressAzureKeyVaults: string
}
interface MnemonicIdentityConfig {
  addressesFromMnemonicCount: string
}
type OracleIdentityConfig = KeyVaultIdentityConfig | MnemonicIdentityConfig
```

codyborn (approved):
Looks good!

PR #4301, opened 2020-07-02
Rely upon SSL config value

codyborn:
### Description

Rely upon the value passed into the firebase config for non-dev environment.

### Backwards compatibility

Yes

### DIFF
--- a/packages/phone-number-privacy/combiner/src/config.ts
+++ b/packages/phone-number-privacy/combiner/src/config.ts
@@ -68,7 +68,8 @@ if (DEV_MODE) {
       password: functionConfig.db.pass,
       database: functionConfig.db.name,
       host: `/cloudsql/${functionConfig.db.host}`,
-      ssl: functionConfig.db.ssl,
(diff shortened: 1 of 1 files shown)


jmrossy (dismissed):

packages/phone-number-privacy/combiner/src/config.ts
```diff
@@ -68,7 +68,7 @@ if (DEV_MODE) {
       password: functionConfig.db.pass,
       database: functionConfig.db.name,
       host: `/cloudsql/${functionConfig.db.host}`,
-      ssl: !DEV_MODE,
+      ssl: functionConfig.db.ssl,
```
jmrossy: Do we need to parse this to boolean like we do for the signer config?

codyborn: good point, tested in staging with/out bool parsing and confirmed that it's needed.

jmrossy (approved):

PR #4302, opened 2020-07-02
Use update RN-bip39 lib to fix mnemonic derivations

jmrossy:
### Description

Use update RN-bip39 lib to fix mnemonic derivations. 
See https://github.com/celo-org/react-native-bip39/pull/4

### Other changes
Increase delay before retrying reveals during verification and add logging.

### Tested

On iOS and Android, confirmed account derivation matches CLI's output.

### Backwards compatibility

**NO, this creates different keys for mnemonics. All current Valora users must migrate**


annakaz (approved):
LGTM

PR #4303, opened 2020-07-02
Flakey Test Tracking

alecps:
### Description

Adds a tool for tracking flakey tests. 100% of tests (except the certora tests) will now be retried a configurable number of times before failing. Tests that pass on a retry are deemed flakey and are tracked by issues created via the GitHub API. Flakey tests corresponding to open issues will be skipped automatically unless specified otherwise. The FlakeTracker app also provides rich feedback about flakey tests via GitHub checks. See README for more details. 

### Other changes

Jest has been upgraded to use [jest-circus](https://www.npmjs.com/package/jest-circus).
Mocha version has been standardized across test suites.
Various dependency version upgrades (notably Detox, Jest, and Mocha)
Job added to CI workflow to provide summary of flakey tests via GitHub Checks after all tests suites have completed. 

### Tested

Tested in the course of integration with our existing test suites, and by adding fake flakiness (now removed) to different test suites to ensure everything is working properly. 

### Backwards compatibility

Changes are backwards compatible as they are confined to testing infrastructure.


.circleci/config.yml
```diff
@@ -470,8 +470,7 @@ jobs:
             ./scripts/ci_check_if_test_should_run_v2.sh @celo/protocol
       - run:
           name: test
-          # Flaky tests - run them twice
-          command: yarn --cwd packages/protocol test || yarn --cwd packages/protocol test
+          command: yarn --cwd packages/protocol test
```
yorhodes, line 482: 🙏

.vscode/settings.json
```diff
@@ -31,7 +31,7 @@
   "[typescript]": {
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": {
-      "source.organizeImports": true
+      "source.organizeImports": false
```
yorhodes: remove this?

.circleci/config.yml
```diff
@@ -485,8 +484,7 @@ jobs:
             ./scripts/ci_check_if_test_should_run_v2.sh @celo/protocol
       - run:
           name: test
-          # Flaky tests - run them twice
-          command: yarn --cwd packages/protocol test:coverage || yarn --cwd packages/protocol test:coverage
+          command: yarn --cwd packages/protocol test:coverage
```
yorhodes, line 496: 🙏

flakey-test-tracking/src/config.js
```diff
@@ -0,0 +1,32 @@
...
+// shouldSkipKnownFlakes => flakey test issues are fetched from github and corresponding tests are skipped
+const shouldSkipKnownFlakes =
+  shouldTrackFlakes && process.env.CIRCLECI && process.env.SKIP_KNOWN_FLAKES !== 'false'
+
+// shouldAddCheckToPR => GitHub Check added to PR
+const shouldAddCheckToPR = shouldTrackFlakes && process.env.CIRCLECI
+
+// shouldCreateIssues => GitHub Issues created for new flakey tests
+const shouldCreateIssues = true
+// const shouldCreateIssues =
```
yorhodes: remove this or revert the comment?

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
+const { Octokit } = require('@octokit/rest')
+const { App } = require('@octokit/app')
+const { retry } = require('@octokit/plugin-retry')
+const Client = Octokit.plugin(retry)
```
yorhodes: this is cool!

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
...
+    }
+    this.app = app
+    this.rest = rest
+  }
+
+  static async build() {
+    const app = new App({
+      id: 71131,
+      privateKey: process.env.FLAKE_TRACKER_SECRET.replace(/\\n/gm, '\n'),
+      //privateKey: privateKey,
```
yorhodes: remove?

flakey-test-tracking/src/jest/utils.js
```diff
@@ -0,0 +1,52 @@
...
+  return fmtTestTitles(getTestTitles(test))
+}
+
+const getTestTitles = (test) => {
+  const titles = []
+  let parent = test
+
+  do {
+    titles.unshift(parent.name)
+  } while ((parent = parent.parent))
```
yorhodes: `==`?

flakey-test-tracking/src/db.js
```diff
@@ -0,0 +1,114 @@
+const { fmtFlakeIssue, getTestSuiteDir } = require('./utils')
```
yorhodes: seems like some of this was probably already implemented in a well-known/tested npm package but I have no issue with what you have done

flakey-test-tracking/src/jest/environments/detox.js
```diff
@@ -0,0 +1,35 @@
+const {
+  DetoxCircusEnvironment,
```
yorhodes: amazing that you are tackling mobile e2e too!
cc @etuleu @cmcewen

flakey-test-tracking/src/utils.js
```diff
@@ -0,0 +1,176 @@
...
+  return title.replace(flakeTitlePrefix, '').trim()
+}
+
+// You can set verbosity to either 0, 1, 2, 3, or 4
+//  4 => header + list of skipped tests + errors from every retry for each new flakey test
+//  3 => header + list of skipped tests + 1 errror for each new flakey test
+//  2 => list skipped tests + list new flakey tests
+//  1 => count of skipped tests + count of new flakey tests
+//  0 => short one line status
+function fmtSummary(flakes, skippedTests, verbosity) {
```
yorhodes: I love this hahaha ❄️😌

packages/celotool/ci_test_attestations.sh
```diff
@@ -13,9 +13,9 @@ if [ "${1}" == "checkout" ]; then
     # Test master by default.
     BRANCH_TO_TEST=${2:-"master"}
     echo "Checking out geth at branch ${BRANCH_TO_TEST}..."
-    ../../node_modules/.bin/mocha -r ts-node/register src/e2e-tests/attestations_tests.ts --branch ${BRANCH_TO_TEST}
```
yorhodes, line 16: why did all these paths change?

packages/mobile/src/account/Invite.test.tsx
```diff
@@ -7,7 +7,8 @@ import { Screens } from 'src/navigator/Screens'
 import { createMockStore, getMockStackScreenProps } from 'test/utils'
 import { mockE164NumberToInvitableRecipient } from 'test/values'
 
-describe('Invite', () => {
```
yorhodes, line 10: +1

packages/mobile/src/backup/__snapshots__/BackupPhraseContainer.test.tsx.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renders correctly for input backup phrase 1`] = `
+exports[` renders correctly for input backup phrase 1`] = `
```
yorhodes, line 4: remove ?

packages/protocol/test/common/accounts.ts
```diff
@@ -81,6 +81,24 @@ contract('Accounts', (accounts: string[]) => {
   })
 
   describe('#createAccount', () => {
+    it('should fail sometimes', async () => {
```
yorhodes: skip these or remove?

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
...
+    if (typeof app === 'undefined' || typeof rest === 'undefined') {
+      throw new Error('GitHub constructor should not be called directly. Please use GitHub.build()')
+    }
+    this.app = app
+    this.rest = rest
+  }
+
+  static async build() {
+    const app = new App({
+      id: 71131,
```
yorhodes: where does this come from?

yorhodes (approved):
Super exciting
Any reason why you went with JS instead of TS? IMO it makes it slightly less maintainable but no reason to switch now
Could we remove some config up to the root of mono so we have less awkward relative paths?
I mostly trust your judgement here and this is pretty hard to review given how large it is

yorhodes:
could you also fill in the PR template to help out reviewers?

.circleci/config.yml
gastonponti: I hate this line, hahaha (the old one)

flakey-test-tracking/src/config.js
```diff
@@ -0,0 +1,32 @@
+// shouldTrackFlakes => tests are retried `numRetries` times and flakey results are logged w/ test output
+const shouldTrackFlakes =
+  (process.env.CIRCLECI && process.env.FLAKEY !== 'false') || process.env.FLAKEY === 'true'
```
gastonponti: so, an empty `process.env.FLAKEY` is considered a `true`? or do we have more than one option?

If Flakey is only true or false, this could be just the `process.env.FLAKEY === 'true'`

And if empty is a valid option and it's true, this could be just `process.env.FLAKEY !== 'false'`

flakey-test-tracking/src/config.js
```diff
@@ -0,0 +1,32 @@
...
+
+// numRetries === times test is run after the initial failure
+const numRetries = process.env.NUM_RETRIES ? Number(process.env.NUM_RETRIES) : 25
```
gastonponti: haha we should discuss that 25, hahahaha.

flakey-test-tracking/src/config.js
```diff
@@ -0,0 +1,32 @@
...
+
+// shouldSkipKnownFlakes => flakey test issues are fetched from github and corresponding tests are skipped
+const shouldSkipKnownFlakes =
+  shouldTrackFlakes && process.env.CIRCLECI && process.env.SKIP_KNOWN_FLAKES !== 'false'
```
gastonponti: Track the flakiness and skip them, could be separate things. I think that with that initial flag, you are adding a limit but maybe I could want a "fail fast without flaky". For example when you start with a migration, and you know that probably every test will fail, but I don't want to run it X times each one, but I want to already avoid the flaky ones. WDYT?

flakey-test-tracking/src/config.js
gastonponti: Are we going to manage the WIP PRs as `FLAKEY=
false'`?

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
...
+}
+
+// Authenticate as an installation of the Flake Tracker GitHub App
+const auth = async (app) => {
+  const rest = new Client({
+    auth: app.getSignedJsonWebToken(),
+  })
+
+  try {
+    const installationId = (
```
gastonponti: This would probably be the same id for all the run, right?
You could calculate it, only if you have an `undefined` variable, and save it as an instance var

gastonponti: And if this is only required to be run once, you could also avoid to generate the "first client" (line 237) again

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
...
+      await rest.apps.getRepoInstallation({
+        ...defaults,
+      })
+    ).data.id
+
+    const installationAccessToken = await app.getInstallationAccessToken({
```
gastonponti: This is probably the one that rotates. And I have a question with this.

If I have the promise A, and the promise B, and both have to renew the token.

This call, it doesn't invalidate the previous token, right?

Otherwise, if two promises want to renew, one could be invalidating the token for the other (just checking, but it that's the case, this should change)

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
...
+    }
+    this.rest = await auth(this.app)
+  }
+
+  async report(flakes, skippedTests) {
+    const promises = []
+
+    if (shouldCreateIssues) {
+      // Check list of ALL flakey issues to ensure no duplicates
+      const knownFlakes = await this.fetchKnownFlakes()
```
gastonponti: Ok, between the fetch, and the actual creation of the issue, we could have a concurrency problem from another instance of circle running the batch of tests.
Don't really know how possible it could be, so we could have this with a note, if we start to see duplicated issues.
And, as you can't "lock" issues, one possible solution, is to make that the same promise, that is creating the issue, to retrieve that same name after it's creation, and if for some reason, there are two instances, delete the one with older ID. That will generate an "eventual consistency" because one of the two instances will end up erasing the duplicated one.

But again, don't know how possible it will be for that to happen

gastonponti: Well, it also dependes on how fast github replicates their info, and if they lock that table in its creation... haha

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
...
+      promises.push(this.endCheck(flakes, skippedTests))
+    }
+
+    console.log('\nSending flake tracker results to GitHub...\n')
+
+    return Promise.all(promises)
+  }
+
+  async issues(flakes) {
+    return Promise.all(flakes.map((f) => this.issue(f)))
```
gastonponti: You should handle the possibility of an error here. And decide (and leave a comment somewhere) on the decision that we are making.
If one issue couldn't be created, are we still want the other issues that were created? or do we want to remove them?

I prefer to keep them, because it's not necessary to have an atomic thing here... we are just finding flakiness. But you should add that, for the others developers to know that you took that decision and that it's not something that you forgot to add and just happens

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
...
+      // Check list of ALL flakey issues to ensure no duplicates
+      const knownFlakes = await this.fetchKnownFlakes()
+      const newFlakes = flakes.filter((flake) => !knownFlakes.includes(flake.title))
+      if (newFlakes.length) {
+        promises.push(this.issues(newFlakes))
+      }
+    }
+
+    if (shouldAddCheckToPR) {
+      promises.push(this.endCheck(flakes, skippedTests))
```
gastonponti: And beware...
if for some reason one issue was not created, and this will assume that everything was created

Maybe you could make this to happen, after knowing that the other issue creation succeeded
`issues.then( v => this.endCheck(flakes, skippedTests) )`
or just wait for them

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,261 @@
...
+    console.log('\nFetching mandatory tests for PR ' + prNumber + '...\n')
+
+    const prBody = (await this.safeExec(fn, errMsg)).data.body
+
+    return parseMandatoryTestIssuesFromPullBody(prBody)
+  }
+
+  async fetchFlakeIssues() {
+    const fn = () => {
+      return this.rest.paginate(this.rest.issues.listForRepo, {
```
gastonponti: do we know the page size? won't we require multiple calls to this?

flakey-test-tracking/src/jest/setup.js
```diff
@@ -0,0 +1,2 @@
+const { numRetries } = require('./config')
+jest.retryTimes(numRetries)
```
gastonponti: mmmm, this retryTimes is for every test, right?

flakey-test-tracking/src/jest/utils.js
gastonponti: No, he basically iterates all the ancestors and add them to the array list
```
describe('bla'
   describe('ble'
     it('blu'
```
starts from 'blu'
=> ['bla', 'ble', 'blu']
It would be weird if it doesn't exist a util function for this...

being ULTRA into details, it's "better" to do
`titles.push(...)`
...
`titles.reverse()`

otherwise every unshift shift the array (and that depends on the implementation)
But it's a detail, your choice

flakey-test-tracking/src/jest/utils.js
```diff
@@ -0,0 +1,52 @@
...
+          child.status = 'flakey'
+        }
+        break
+      }
+    }
+  }
+  return describeBlock
+}
+
+const getTestID = (test) => {
```
gastonponti: Ok, just to check.
This gets all the "descriptions" + "it" AND the path of the test, right?

flakey-test-tracking/src/utils.js
```diff
@@ -0,0 +1,176 @@
+// Emojis
```
gastonponti: 🤣

packages/utils/src/task.test.ts
```diff
@@ -239,4 +239,12 @@ describe('tryObtainValueWithRetries()', () => {
...
     expect(task.isRunning()).toBeFalsy()
   })
+
+  test('should fail sometimes', async () => {
+    const x = Math.floor(Math.random() * Math.floor(4))
+    expect(x).toBeGreaterThan(1)
+
+    const y = Math.floor(Math.random() * Math.floor(5))
+    expect(y).toBeGreaterThan(2)
+  })
```
gastonponti: LOVE THE WARNING ❤️

gastonponti (commented):
Great work! This is beautiful!!! Amazing job! 😄 

I'm just a little concern on two things

1) the ID is the path and titles, so we should warn the user if for some reason copy-pasted the code and it didn't changed something... It's weird I know, that's why it's not important, but maybe using the same logic for generating those IDs, we could add a linter
2) if someone changes a description, for some reason (a refactor, for example) we should make the user to update the issues titles, right? Even if the user moves a folder to a subfolder

I think that for the 2) one thing that we could do, is to make the user to manually tag those issues, or to close them, and the same flaky cycle could generate again the issues for the new names.
And one way to know that that is happening, is to check if there all the open flaky issues are listed in tests for that PR, and request an action from the user. So if the user renamed, change, remove some of the issues, this way at least it will have that warning. But this would also have a lot of cases...
For example having a PR that is not updated from master, and in one of the new PRs on master, someone add a new flaky test... in that case, the user should update the branch 

Another thing, could be to make a issue cleaner...
Something like, it this commit exist in master, and it doesn't have some of the flaky issues listed, to close it.

So, let's say that we had the test "folder1 -> Test1" that was flaky. And we have already the issue "folder1 -> Test1" open and tag as flaky.
Then I make another PR, and rename the "folder1" to "folder2".
When I'm merging my PR, it will "probably" (could be in another run, but it will do that) the issue "folder2 -> Test1".
But, if we have a flaky cleaner in master (could it be the same that created the issues). You could have like a check, of which "flaky issues" live in our repo that are still open, but our last commit in master don't have that specific test. And in that case, close the issue, tag it with a new tag and add a comment like "The test is not named like that anymore, or it was deleted. If it was renamed, add the reference of the new issue".
So, in this example, the same merge, that will create the "folder2 -> test1" issue, will close and tag the "folder1 -> test1"

What do you think? Do you have something in mind?

packages/celotool/ci_test_attestations.sh
alecps: For some reason we were using a version of mocha from the root dir node-modules when we were importing a newer version of mocha (and not using it) in the celotool package. I needed the newer version of mocha so I just changed the paths

flakey-test-tracking/src/jest/setup.js
alecps: Yep!

flakey-test-tracking/src/github.js
alecps: Great point. I think my intuition here though was that the odds of us getting a duplicate issue are really small so we won't have to worry about getting cluttered with lots of issues. I think in the interest of simplicity it might be better to leave this one and if we do see duplicate issues we can always revisit

alecps: I'll definitely add a note though

flakey-test-tracking/src/config.js
alecps: Haha totally. There are a few flakey tests that don't seem to respond to retries so I set it high to try and capture those. I was thinking 10-15 would be more reasonable. What do you think?

flakey-test-tracking/src/config.js
alecps: So I've set it up this way so that the default when running locally is to not use the flake tracker (you have to set FLAKEY = true to use it) and the default when running in CI is to use the flake tracker (you have to set FLAKEY=false to not use it)

packages/mobile/src/backup/__snapshots__/BackupPhraseContainer.test.tsx.snap
alecps: This was a weird error I was seeing with the mobile tests. I'll try removing it but there's a chance the mobile tests fail again. I chatted with Rossy about it and he said it was weird but probably not a big deal. Beyond me why jest would change this for just one snapshot

flakey-test-tracking/src/jest/utils.js
alecps: Cool! That's an easy change to make. I'm by no means a js expert so I always appreciate tips like that

flakey-test-tracking/src/github.js
alecps: Great question. I don't think it invalidates previous tokens based on the fact that concurrent jobs in CI seem able to authenticate without any issues

flakey-test-tracking/src/jest/utils.js
alecps: So the testID actually ends up being 'jobName -> packageName -> rootDescribeTitle -> ... -> testTitle'

packages/mobile/src/backup/__snapshots__/BackupPhraseContainer.test.tsx.snap
gastonponti: Could you add a comment there? To avoid someone changing it in the future by mistake

flakey-test-tracking/src/config.js
gastonponti: so, empty in circle-ci means true, and in the other environments, false. I think that it could be misleading at the beginning. Try to document that somewhere, haha

flakey-test-tracking/src/config.js
alecps: Hmm okay yea I like that. I'll add it

flakey-test-tracking/src/github.js
alecps: 'paginate' returns all results across all pages so we don't need to worry about page size here

flakey-test-tracking/src/config.js
alecps: Right, I've added it to the README and left a comment in the config file

flakey-test-tracking/src/config.js
alecps: Actually, I think the way to do this would be to just set NUM_RETRIES=0 . Decoupling the shouldTrackFlakes logic and the shouldSkipTests logic would be non-trivial and I'm hesitant to make a change like that at this stage. What do you think about adding some documentation in the README explaining that if you want a 'fail fast without flaky' you should set NUM_RETRIES=0?

gastonponti: Ok, if it's documented, should be fine.
Another way could be to add a new flag, and if that flag is set, you could assume the NUM_RETRIES=0. But both seems good to me, your choice.

flakey-test-tracking/src/config.js
gastonponti: I think that it doesn't really matter if we catch the flakiness in the first try, the idea is that eventually will happen.
I'm afraid of adding this, and having a lot of PRs which tests fail because are in the beginning of the feature, and having circle taking a long time for them (we have an amount of instances in circle to run everything).
If this just run on master, we could leave a 10. If this could potentially run on every PR, I would start with a small number, and just change it in the future if we see that nothing is catch and it doesn't take too much time.

alecps: Good thought. I'll change this to be 5 for PRs and 10 when running on master

flakey-test-tracking/src/github.js
alecps: Sounds good. The safeExec function that all the API calls use handles errors with console.error as I didn't want the workflow to fail just because the flake tracker couldn't create an issue. I'll add some comments to clarify

flakey-test-tracking/src/github.js
alecps: I've updated the way this works so that we don't have to worry about ending the check anymore

flakey-test-tracking/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+To ensure that a specific known flakey test is run for your PR, simply include the link to the flakey test's issue anywhere in the body of your PR.
+
+## Still Seeing Flakey Tests?
+
+Sometimes a flakey test will fail all retries. You can try bumping up the number of retries that are attempted by setting the `NUM_RETRIES` env variable.
+
+Some flakey tests are not uncovered by retries. That is, if they fail the first time then every retry will also fail. If you encounter tests like this, please create an issue to track it. If you do so correctly, the flakey test will be disabled until the issue is closed.
+To manually create a flakey test issue, mimic the format of issues created by the FlakeTracker bot. Specifically, make sure to add the `FLAKEY` label as well as labels for the package name and the ci job the test is run in. Also, include the `testID` of the test in the issue title.
+The `testID` can be derived as follows: `jobName -> packageName -> rootDescribeBlockTitle -> ... -> testTitle`. The `testID` is printed each time the test is retried and can be found easily in the logs. Please also include the test error in the issue body.
```
gastonponti: One little add that we could do, is to add some sort of flag for running tests AND print the testID in the console, and maybe use the same wrapper that checks the retry, to just print the TestID.

So for example, if you run `yarn test --withTestID ../my_test`
will run every test in that file, with their TestIDs (using your same util)

If adding that is easy, we could add it, otherwise we could add it as a nice-to-have card for the future

flakey-test-tracking/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+It is important to note that some flakiness might exist in setup/teardown steps like `before` and `after` hooks. FlakeTracker does not currently address these cases, but you should still create issues to track them!
```
gastonponti: Mmmmm, this brings me a question, haha.
If we have 2 flaky test, and it is because of a setup, maybe fixing one of both, will "resolve" the other.
Don't think that is a big problem right now, but maybe we could run all the test batch that lived with a flaky one that was fixed...
For the moment is not that important, but it would be good to add it as a future improvement

flakey-test-tracking/src/github.js
```diff
@@ -0,0 +1,367 @@
+const { Octokit } = require('@octokit/rest')
+const { App } = require('@octokit/app')
+const { retry } = require('@octokit/plugin-retry')
+const Client = Octokit.plugin(retry)
+const stripAnsi = require('strip-ansi')
+const config = require('./config')
+const utils = require('./utils')
+
+// This is the FlakeTracker GitHub App ID. Can be found at github.com -> celo-org -> app settings
+const flakeTrackerID = 71131
```
gastonponti: It's not a problem to have the Id here, right?
Maybe we could add it in the config file

gastonponti (approved):
I've added 3 possible details.
But I don't think that are necessary to be fixed before merging. It's up to you, haha.

Amazing work! 🎉 🎉

flakey-test-tracking/README.md
alecps: The testID is already printed each time a retry happens, is that what you're suggesting? If a test fails, it will have been retried so the testID will already be there in the logs.

flakey-test-tracking/README.md
alecps: Yea that's a great point. I think that would be a lot of work to implement right now though because the tool doesn't actually have a way of detecting that a test is fixed (the developer that fixes the test is just expected to close the issue so the test is re-enabled). I think a high-impact feature we could add in the future would be checking that a test is fixed. That is, we could have a flag for retrying tests on _success_ (rather than just on failure) and then a test that passes each retry is marked "fixed". This might be non-trivial though, since Mocha and Jest only have native support for retrying _failed_ tests, so we'd have to hack together some other way to retry successful tests. I do think that feature would be really nice to have though when checking whether you've actually fixed a flakey test or not. From there, we could work it into regular test execution like you describe above.

flakey-test-tracking/src/github.js
alecps: No not a problem at all, it's publicly accessible via the API. It would be better to have it in the config file though.

flakey-test-tracking/README.md
gastonponti: Maybe not thinking only in the fail case. Just to make it easier for the user to create an issue for a flaky one that was not detected. But if we have that for fails, it will also work, so don't worry

PR #4305, opened 2020-07-03
[Wallet] Move signal handling to main

etuleu:
### Description

Moving it earlier because I still see a few crashes coming in which seem to happen in the app delegate initialization code.

### Tested

It's difficult to trigger the actual crash that we see in production, which I suspect happens from one of the libraries that we use (one stack trace contained segment-firebase code). But I locally tested with the pkill command.

### Backwards compatibility


packages/mobile/ios/celo/main.m
```diff
@@ -4,6 +4,10 @@
 
 int main(int argc, char * argv[]) {
   @autoreleasepool {
+
+    // Ignore SIGPIPE signals
+    signal(SIGPIPE, SIG_IGN);
```
jmrossy, line 12: no import needed here?

packages/mobile/ios/celo/main.m
```diff
@@ -4,6 +4,10 @@
 
 int main(int argc, char * argv[]) {
   @autoreleasepool {
+
+    // Ignore SIGPIPE signals
```
jeanregisser, line 8: Nit: could you add a little more to explain why we do this?

jeanregisser (approved):

packages/mobile/ios/celo/main.m
etuleu: Yes that surprised me as well but Xcode did not complain. I think one of the core ios headers imports it already.

PR #4311, opened 2020-07-03
[Wallet] Analytics instrumentation and standardization across Valora

tarikbellamine:
### Description
Added events throughout the app and standardized naming and placement. Also added copyable sessionId to the drawer menu and including it in support emails to aid in debugging.

### Other changes
Using specific analytics naming to make all related events easily searchable. For example, search "SendEvents" to view all analytics events related to the Send flow.

### Tested
Yes

### Related issues
- Part of #4008

### Backwards compatibility
Yes


packages/mobile/src/account/AccountKeyEducation.tsx
```diff
@@ -41,30 +47,10 @@ function useSteps() {
...
-        {
-          image: null,
-          cancelEvent: AnalyticsEvents.backup_educate_4_cancel,
-          progressEvent: AnalyticsEvents.backup_educate_4_next,
-          screenName: 'AccountKeyEducation',
-        },
+        { image: null, topic: EducationTopic.backup },
+        { image: null, topic: EducationTopic.backup },
+        { image: null, topic: EducationTopic.backup },
+        { image: null, topic: EducationTopic.backup },
```
jmrossy: What is this change meant to do?

packages/mobile/src/account/Education.tsx
```diff
@@ -45,15 +49,26 @@ export default class Education extends React.Component<Props, State> {
...
-      })
-    }
-    if (this.state.step === 0) {
+    const { step } = this.state
+    const { topic } = this.props.stepInfo[this.state.step]
+    if (step === 0) {
+      const analyticsEvent =
+        topic === EducationTopic.backup
+          ? OnboardingEvents.backup_education_cancel
+          : OnboardingEvents.celo_education_cancel
```
jmrossy: Rather than assuming the education screen will be just one of the other, can this be generalized with a nav prop or something like that?

packages/mobile/src/account/GoldEducation.tsx
```diff
@@ -51,27 +55,19 @@ function useStep() {
     return [
       {
         image: shinyGold,
-        cancelEvent: AnalyticsEvents.gold_cancel1,
-        progressEvent: AnalyticsEvents.gold_educate_1_next,
-        screenName: 'Gold_Nux_1',
+        topic: EducationTopic.celo,
```
jmrossy: ditto, I must be missing something but how is it all 4 topics are the same value now?

packages/mobile/src/analytics/Events.tsx
```diff
@@ -8,186 +10,202 @@ export enum AppEvents {
   fetch_balance = 'fetch_balance',
 }
 
-export enum SettingsEvents {
-  edit_profile = 'edit_profile', // to count the # of taps on the edit profile button in Account_Home
-  edit_name = 'edit_name', // to count the # of taps on the “Edit Name” nav element in Account_Edit
-  language_select = 'language_select',
+export enum HomeEvents {
```
jmrossy, line 13: These are def looking better organized now :)

packages/mobile/src/app/reducers.ts
```diff
@@ -119,6 +121,11 @@ export const appReducer = (
         ...state,
         locked: false,
       }
+    case Actions.SET_SESSION_ID:
+      return {
+        ...state,
+        sessionId: action.sessionId,
```
jmrossy, line 128: should this be reset to '' in the rehydrate reducer above?

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -157,9 +175,19 @@ function CustomDrawerContent(props: DrawerContentComponentProps<DrawerContentOpt
...
+        {devModeActive && (
+          <>
+            <Text style={fontStyles.label}>Session ID</Text>
+            <View style={styles.accountOuterContainer}>
+              <View style={styles.accountInnerContainer}>
+                <SessionId sessionId={storedSessionId || ''} />
+              </View>
+            </View>
+          </>
+        )}
```
jmrossy: Feels like this should go in Settings with the other devModeActive-only settings and values

jmrossy (commented):
Mostly looks great, just a few comments

packages/mobile/src/account/AccountKeyEducation.tsx
tarikbellamine: This is meant to simplify and reduce the data we are passing through to the `Education` screen. All we need to know is where the data is coming from and in the `Education` methods themselves we have the logic for determining which event to log.

I think this helps keep analytics logic a little more contained.

packages/mobile/src/account/Education.tsx
tarikbellamine: We could. We have an `EducationTopic` enum defined in this file where the two types thus far are `backup` and `celo`. Do you think we should split this up into if statements now or wait until we add more `EducationTopic`s?

packages/mobile/src/account/GoldEducation.tsx
tarikbellamine: Logic is just moved to the `Education` screen where `topic` is the identifier

packages/mobile/src/app/reducers.ts
tarikbellamine: Yeah good call

packages/mobile/src/account/Invite.tsx
```diff
@@ -85,10 +83,7 @@ class Invite extends React.Component<Props, State> {
 
     const hasGivenContactPermission = await requestContactsPermission()
     this.setState({ hasGivenContactPermission })
-
-    if (hasGivenContactPermission) {
```
annakaz, line 89: Why is checking contact permissions no longer necessary?

packages/mobile/src/account/Settings.tsx
```diff
@@ -94,15 +100,23 @@ const mapDispatchToProps = {
...
 }
 
 interface State {
   fornoSwitchOffWarning: boolean
 }
 
 export class Account extends React.Component<Props, State> {
+  componentDidMount = () => {
+    const sessionId = ValoraAnalytics.getSessionId()
+    if (sessionId !== this.props.sessionId) {
```
annakaz, line 113: When would ValoraAnalytics.getSessionId() be out of sync with what's in redux?

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -72,7 +72,7 @@ class ValoraAnalytics {
         this.deviceInfo = deviceInfo
         this.sessionId = Web3Utils.soliditySha3({
```
annakaz: Throughout the codebase we tend to use `sha256` from ethereumjs-util when we just need a generic hash function 
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/transactions/actions.ts#L74
and in `authentication.ts` and the web3 soliditySha3 for the verification hash identifier

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -72,7 +72,7 @@ class ValoraAnalytics {
...
           type: 'string',
-          value: deviceInfo.SerialNumber || deviceInfo.DeviceId,
+          value: Math.random().toString(),
```
annakaz: Hadn't noticed this before but could DeviceInfo.getUniqueId() work as this identifier?

packages/mobile/src/components/AccountNumber.tsx
```diff
@@ -4,22 +4,30 @@ import { getAddressChunks } from '@celo/utils/src/address'
...
+export default function AccountNumber({ address, touchDisabled, location }: Props) {
   const { t } = useTranslation(Namespaces.accountScreen10)
   const onPressAddress = () => {
     if (!address.length) {
       return
     }
     Clipboard.setString(address)
     Logger.showMessage(t('addressCopied'))
+
+    if (location === Screens.DrawerNavigator) {
```
annakaz, line 28: Would be helpful to track copying other users' addresses as well

packages/mobile/src/exchange/saga.ts
```diff
@@ -139,8 +140,11 @@ export function* doFetchExchangeRate(action: FetchExchangeRateAction) {
         dollarMaker: dollarMakerExchangeRate.toString(),
       })
     )
+    ValoraAnalytics.track(CeloExchangeEvents.celo_fetch_exchange_rate_complete)
```
annakaz: If not already tracked it would be helpful to track the exchange rate that was fetched and the amount it was estimated with to provide us with data in addressing the UX issues around the amount-dependent exchange rate

packages/mobile/src/exchange/saga.ts
```diff
@@ -258,7 +261,9 @@ export function* exchangeGoldAndStableTokens(action: ExchangeTokensAction) {
         convertedMakerAmount.toString()
       )
     } else {
-      ValoraAnalytics.track(AnalyticsEvents.unexpected_maker_token, { makerToken })
+      ValoraAnalytics.track(CeloExchangeEvents.celo_exchange_error, {
+        error: `Unexpected maker token: ${makerToken}`,
+      })
```
annakaz, line 270: Perhaps outside the scope of this PR but in Exchange, we see transactions fail if a user only has Celo, as the wallet pays fees in cUSD only. Would be helpful to track when people attempt an exchange when they only have celo

packages/mobile/src/home/NotificationBox.tsx
```diff
@@ -275,7 +336,7 @@ export class NotificationBox extends React.Component<Props, State> {
       ...this.generalNotifications(),
     ]
 
-    if (!notifications || !notifications.length) {
+    if (!notifications.length) {
```
annakaz, line 339: If notifications is null will this throw an error as property .length won't exist?

packages/mobile/src/account/Settings.tsx
tarikbellamine: Should only happen at the very start of a session.

packages/mobile/src/account/Invite.tsx
tarikbellamine: We are already checking for permissions in the `doImportContacts` generator within `contactMapping.ts`. Beyond not needing the redundant check on the frontend, it's easier to only have one check on the backend so we only have to place the analytics event in one place.

packages/mobile/src/analytics/ValoraAnalytics.ts
tarikbellamine: I'll go with the same format as the example you gave me above except use `UniqueID` instead of `recipientAddress`

https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/transactions/actions.ts#L74

packages/mobile/src/home/NotificationBox.tsx
tarikbellamine: It's hidden in the diff but right above this notifications is defined as `[...this.incomingPaymentRequestsNotification(), all the others]` so it should always be defined as an array

packages/mobile/src/exchange/saga.ts
tarikbellamine: We have the below check. Does this work?

`    // Ensure the user gets makerAmount at least as good as displayed (rounded to EXCHANGE_DIFFERENCE_TOLERATED)
    const minimumTakerAmount = BigNumber.maximum(
      getTakerAmount(makerAmount, exchangeRate).minus(EXCHANGE_DIFFERENCE_TOLERATED),
      0
    )
    const updatedTakerAmount = getTakerAmount(makerAmount, updatedExchangeRate)
    if (minimumTakerAmount.isGreaterThan(updatedTakerAmount)) {
      ValoraAnalytics.track(CeloExchangeEvents.celo_exchange_error, {
        error: `Not receiving enough ${makerToken}. Expected ${minimumTakerAmount} but received ${updatedTakerAmount.toString()}`,
      })
      Logger.error(
        TAG,
        `Not receiving enough ${makerToken} due to change in exchange rate. Exchange failed.`
      )
      yield put(showError(ErrorMessages.EXCHANGE_RATE_CHANGE))
      return
    }`

packages/mobile/src/exchange/saga.ts
tarikbellamine: What net new information would we need to log for that? Wouldn't `makerToken` be `CELO` in that instance?

packages/mobile/src/account/Settings.tsx
annakaz: Ok it's not clear to me why the check is needed here but not elsewhere

packages/mobile/src/exchange/saga.ts
annakaz: We would want to log the celo balance in the case of an error- sorry this comment is separate from the line I commented on

annakaz (approved):
Thanks for getting these in! LGTM overall

packages/mobile/src/account/Settings.tsx
tarikbellamine: SessionId is set in the `ValoraAnalytics` class, which isn't connected to the redux store. What I am doing here is checking if the sessionId has been defined by `ValoraAnalytics` but not yet put in the store.

We really only need to do this once per session and we are doing it on the `Account` screen because that's the only place sessionId is displayed (but only when in dev mode).

PR #4315, opened 2020-07-03
[Wallet] Add more branded images and clean up existing ones

jeanregisser:
### Description

This PR adds more branded images and cleans up existing ones.

See assets diff: https://github.com/clabs-co/valora-app-branding/compare/eefeea1...d5249be

### Other changes

- Branded assets are now being cleaned up in their final destination folder when switching branding. This is to avoid confusion having assets from the previously synced branding laying around unused.
- Also now switching to the default celo branding when running test

### Tested

Yes ;-)

### Related issues

- #3645 

### Backwards compatibility

Yes.


jmrossy (approved):
Looks good!

PR #4316, opened 2020-07-03
[Wallet] Various small fixes

jmrossy:
### Description

1. Update SMS retriever hash for Valora
2. Disable experimental layout feature on android which was causing blank screens
3. Add better error message for Invalid pin


tarikbellamine (approved):

jmrossy:
I believe the tests are broken until #https://github.com/celo-org/celo-monorepo/pull/4315 merges

packages/mobile/locales/en-US/onboarding.json
```diff
@@ -20,7 +20,7 @@
     "create": "Create a PIN",
     "verify": "Enter PIN again",
     "pinsDontMatch": "The PINs didn't match",
-    "invalidPin": "Invalid PIN"
+    "invalidPin": "Please use a stronger PIN"
```
jmrossy, line 23: This is only shown if the pin was blacklisted, so we can assume the reason is weak pin

PR #4325, opened 2020-07-05
[Wallet] Fix Celo/cUSD rates

i1skn:
### Description

Stop to use Celo/cUSD rates stored in firebase, because they are calculated
to be used for historical rates representation only and can not be used
for actual transactions. The reason for this is that exchange rate is depend
on exchnage value and all the rates in firebase are calculated using the
same exchange value. Instead we calculate the rate based on the transfers
in each transaction individually.

Based on this proposal https://docs.google.com/document/d/1-XF9Q9pZfZ--6XKqMzlsGBxZhfSagwHGPbN1ZUHgCG0/edit?userstoinvite=jean@celo.org&ts=5eff56a8&actionButton=1#heading=h.vclsg6twr41l

### Other changes

Move "safe area" inside `<ScrollView />` on CELO screen.

### Tested

Locally running `blockchain-api`

### Backwards compatibility

Yes


packages/blockchain-api/src/blockscout.ts
```diff
@@ -296,11 +309,13 @@ export class BlockscoutAPI extends RESTDataSource {
       makerAmount: {
         value: new BigNumber(inTransfer.value).dividedBy(WEI_PER_GOLD).toString(),
         currencyCode: inTransfer.token,
+        knownExchangeRates: getKnownExchangeRates(inTransfer, outTransfer),
         timestamp,
       },
       takerAmount: {
         value: new BigNumber(outTransfer.value).dividedBy(WEI_PER_GOLD).toString(),
         currencyCode: outTransfer.token,
+        knownExchangeRates: getKnownExchangeRates(outTransfer, inTransfer),
```
jmrossy: To clarify, one of these is undefined and the other is the rate. Is that correct?

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -74,14 +77,18 @@ export default class CurrencyConversionAPI<TContext = any> extends DataSource {
...
+    knownExchangeRates?: { [pair: string]: string }
   ): Promise<BigNumber> {
     const pair = `${fromCode}/${toCode}`
 
     if (pair === 'cUSD/USD' || pair === 'USD/cUSD') {
       // TODO: use real rates once we have the data
       return Promise.resolve(new BigNumber(1))
     } else if (pair === 'cGLD/cUSD' || pair === 'cUSD/cGLD') {
+      if (knownExchangeRates && knownExchangeRates[pair]) {
+        return Promise.resolve(new BigNumber(knownExchangeRates[pair]))
```
jmrossy: Why do we return a promise here? Same question for 3 lines above.

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -160,53 +160,55 @@ function ExchangeHomeScreen({ navigation }: Props) {
...
-            <Text style={styles.goldPriceCurrentValue}>
-              {currentGoldRateInLocalCurrency
-                ? displayLocalCurrency(currentGoldRateInLocalCurrency)
-                : '-'}
-            </Text>
-
-            {rateChangeInPercentage && (
-              <Text style={rateWentUp ? styles.goldPriceWentUp : styles.goldPriceWentDown}>
-                {rateWentUp ? '▴' : '▾'} {rateChangeInPercentage.toFormat(2)}%
+        <SafeAreaView style={styles.background} edges={['bottom']}>
```
jmrossy, line 163: Why do we have a safeAreaView inside a safeAreaView?

jmrossy (commented):
Mostly looks good, just some small questions

packages/blockchain-api/src/blockscout.ts
```diff
@@ -27,6 +28,18 @@ export interface BlockscoutCeloTransfer {
   value: string
 }
 
+// Calculate cGLD to cUSD rate based on the actual transfers
+function getKnownExchangeRates(
```
annakaz: nit: I would find a function name like `impliedRate` easier to understand

packages/blockchain-api/src/blockscout.ts
```diff
@@ -27,6 +28,18 @@ export interface BlockscoutCeloTransfer {
...
+  current: BlockscoutCeloTransfer,
+  comparable: BlockscoutCeloTransfer
+) {
+  return current.token === CGLD && comparable.token === CUSD
```
annakaz: If current must be cGLD and comparable must be cUSD, could this be enforced in the function signature, or throw an error rather than returning null?

packages/blockchain-api/src/schema.ts
```diff
@@ -61,11 +61,14 @@ export interface CurrencyConversionArgs {
...
   currencyCode: string
   timestamp?: number
+  knownExchangeRates?: { [pair: string]: string }
 }
 
 export interface MoneyAmount {
   value: BigNumber.Value
   currencyCode: string
+  // to override exchange rates from the APIs
+  knownExchangeRates?: { [pair: string]: string }
```
annakaz: Are multiple exchange rates needed? Also, could the rate be a number/BigNumber.value instead of a string, in order to be consistent with other types in this interface?

packages/blockchain-api/test/blockscout.test.ts
```diff
@@ -70,11 +70,15 @@ describe('Blockscout', () => {
           "hash": "0xba620de2d812f299d987155eb5dca7abcfeaf154f5cfd99cb1773452a7df3d7a",
           "makerAmount": Object {
             "currencyCode": "cUSD",
+            "knownExchangeRates": undefined,
```
annakaz: It is confusing to me that the expected behavior here is to have an undefined exchange rate. Would it make sense to store the exchange rate for both transactions, even if it is redundant information? Or find some other way to avoid this undefined?

packages/blockchain-api/src/schema.ts
```diff
@@ -61,11 +61,14 @@ export interface CurrencyConversionArgs {
   sourceCurrencyCode?: string
   currencyCode: string
   timestamp?: number
+  knownExchangeRates?: { [pair: string]: string }
```
annakaz: Same as comment below- are multiple exchange rates needed or is this always a single rate?

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -74,14 +77,18 @@ export default class CurrencyConversionAPI<TContext = any> extends DataSource {
...
+    timestamp?: number,
+    knownExchangeRates?: { [pair: string]: string }
   ): Promise<BigNumber> {
     const pair = `${fromCode}/${toCode}`
 
     if (pair === 'cUSD/USD' || pair === 'USD/cUSD') {
       // TODO: use real rates once we have the data
       return Promise.resolve(new BigNumber(1))
     } else if (pair === 'cGLD/cUSD' || pair === 'cUSD/cGLD') {
+      if (knownExchangeRates && knownExchangeRates[pair]) {
```
annakaz: What is the expected behavior if it is a cGLD/cUSD or cUSD/cGLD transfer but the exchange rate is missing? To throw an error?

annakaz (commented):
Thanks for working on this! Left some comments below

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
i1skn: Well, this is my workaround, when we have to use one part (top) of `SafeAreaView] outside of `ScrollView` and another (bottom) inside `ScrollView`. It does look odd, I agree, do you know how we can solve it better?

i1skn:
@annakaz @jmrossy Thanks a lot for the review! I've addressed all the feedback - PTAL

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
i1skn: We only want to use implied exchange rate when there was an exchange, otherwise we still ask the API.

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
```diff
@@ -74,14 +77,25 @@ export default class CurrencyConversionAPI<TContext = any> extends DataSource {
...
+      // If this is an exchange
+      if (impliedCeloToCUSDExchangeRate) {
+        return new BigNumber(impliedCeloToCUSDExchangeRate)
+      }
+      return this.goldExchangeRateAPI.getExchangeRate({
+        sourceCurrencyCode: fromCode,
+        currencyCode: toCode,
+        timestamp,
+      })
+    } else if (pair === 'cUSD/cGLD') {
```
jeanregisser: Why aren't we doing the same as above for this pair?

packages/blockchain-api/src/blockscout.ts
```diff
@@ -27,6 +28,19 @@ export interface BlockscoutCeloTransfer {
...
+  comparable: BlockscoutCeloTransfer
+): Pick<MoneyAmount, 'impliedCeloToCUSDExchangeRate'> {
+  return {
+    ...(current.token === CGLD &&
+      comparable.token === CUSD && {
+        impliedCeloToCUSDExchangeRate: new BigNumber(comparable.value).dividedBy(current.value),
+      }),
+  }
+}
+
```
jeanregisser: nit: syntax seems a bit confusing at first, why not just returning the value (or undefined)?
I guess you wanted to avoid having the key in the final object, right?

jeanregisser (commented):
Thanks for diving into this!
Left one question.

packages/blockchain-api/src/blockscout.ts
i1skn: Yeap, to avoid having `undefined` in the final object.

packages/blockchain-api/src/currencyConversion/CurrencyConversionAPI.ts
i1skn: We are only using `impliedCeloToCUSDExchangeRate` when converting from `cGLD` to `localCurrency` (see here: https://github.com/celo-org/celo-monorepo/blob/c99339c7a9bfafe681fa861f393325cf79af3e4a/packages/blockchain-api/src/schema.ts#L313-L318).

jeanregisser (approved):
Awesome!
Nice job refactoring the new additions! 👍

packages/blockchain-api/src/schema.ts
```diff
@@ -61,11 +61,15 @@ export interface CurrencyConversionArgs {
...
   currencyCode: string
   timestamp?: number
+  impliedExchangeRates?: MoneyAmount['impliedExchangeRates']
 }
 
 export interface MoneyAmount {
   value: BigNumber.Value
   currencyCode: string
+  // to override exchange rate from firebase, cause
+  // it can not be used with varying exchange value
```
annakaz:
```suggestion
  // Implied exchange rate (based on exact amount exchanged) which overwrites 
  // the estimate in firebase (based on a constant exchange amount)
```

annakaz (approved):
LGTM!

jmrossy (approved):

PR #4327, opened 2020-07-06
Catch all reveal errors

nambrot:
### Description

If there is a lower-level error in the networking, we currently do not capture it. This PR ensures that we have at least an event for it.


jeanregisser (approved):
Thanks 👍

PR #4334, opened 2020-07-06
Signer DB Discrepancy Detection

codyborn:
### Description

We'd like to be alerted when there is a discrepancy with the quota-related state in the databases of the signers. This change leverages the shared interface between the signer and combiner to pass the DB state information (performedQueryCount + totalQuota). 

### Other changes

Move shared constants and utils to the common library.

### Tested

Local unit tests.

### Backwards compatibility

Yes, the combiner shouldn't fail if the signer doesn't pass these new values.


packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
```diff
@@ -12,38 +12,53 @@ function flattenSigsArray(sigs: Uint8Array[]) {
   return Uint8Array.from(sigs.reduce((a, b) => a.concat(Array.from(b)), [] as any))
 }
 export class BLSCryptographyClient {
-  /*
-   * Computes the BLS signature for the blinded phone number.
-   */
-  public static async combinePartialBlindedSignatures(
-    serviceResponses: ServicePartialSignature[],
+  private verifiedSignatures: Uint8Array[] = []
```
aslawson, line 15: why is this being pulled out as a global variable?  it doesn't appear to get reinitialized to an empty array between requests, so wouldn't it just accumulate signatures from all requests?

aslawson: how about the caller initializes and pass in the array to `addSignature`/`combinePartialBlindedSignatures` methods

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -74,18 +76,19 @@ async function requestSignatures(request: Request, response: Response) {
...
-              request.body.blindedQueryPhoneNumber
-            )
-            logger.debug('Responding with combined signature')
-            response.json({ success: true, combinedSignature, version: VERSION })
+          if (!sentResult && blsCryptoClient.sufficientVerifiedSignatures()) {
+            const combinedSignature = await blsCryptoClient.combinePartialBlindedSignatures()
+            if (!sentResult) {
+              response.json({ success: true, combinedSignature, version: VERSION })
+              sentResult = true
+            }
```
aslawson: feels like this logic should be extracted to another method

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -105,12 +108,45 @@ async function requestSignatures(request: Request, response: Response) {
...
+  if (!blsCryptoClient.sufficientVerifiedSignatures()) {
+    handleMissingSignatures(majorityErrorCode, response)
+  }
+}
+
+function logQuotaDiscrepancies(responses: SignMsgRespWithStatus[]) {
+  // Only compare responses which have values for the quota fields
+  const successfulResponses = responses.filter(
+    (response) =>
+      response.status === 200 &&
```
aslawson: do we only want to log discrepancies for 200? if a signers is reporting an error, it could be valuable to see if the quota numbers differ (eg. if they report a 403 quota exceeded or if their signature generation is failing so they are not updating their counts).

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -105,12 +108,45 @@ async function requestSignatures(request: Request, response: Response) {
...
+    ) {
+      const values = successfulResponses.map((response) => {
+        return {
+          signer: response.url,
+          performedQueryCount: response.signMessageResponse?.performedQueryCount,
+          totalQuota: response.signMessageResponse?.totalQuota,
+        }
+      })
+      logger.error(`Discepancy found in signers' measured quotas ${values}`)
+      return
```
aslawson, line 201: does this `return` break the `forEach` iteration in javascript?  otherwise, it seems if the first response is the odd on out, we'd be logging the error message with all values `n` times.

packages/phone-number-privacy/signer/src/salt-generation/query-quota.ts
```diff
@@ -7,14 +7,13 @@ import { getPerformedQueryCount } from '../database/wrappers/account'
...
- * calculated query quota and the number of queries already performed.
+ * Returns the number of queries already performed and the calculated query quota.
  */
 export async function getRemainingQueryCount(account: string, hashedPhoneNumber?: string) {
   logger.debug('Retrieving remaining query count')
-  const queryQuota = await getQueryQuota(account, hashedPhoneNumber)
+  const totalQuota = await getQueryQuota(account, hashedPhoneNumber)
   const performedQueryCount = await getPerformedQueryCount(account)
-  return queryQuota - performedQueryCount
+  return [performedQueryCount, totalQuota]
```
aslawson: i feel this should be returned as an object rather than array.  then the values will be tied to a descriptive key
`{performedQueryCount, totalQuota}`

and you can still call it similarly -- something like
`const {performedQueryCount, totalQuota} = await getRemainingQueryCount(
      account,
      hashedPhoneNumber
    )` (i've seen rossy do it :P)

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -105,12 +108,45 @@ async function requestSignatures(request: Request, response: Response) {
...
+      resp.signMessageResponse?.totalQuota !== expectedTotalQuota
+    ) {
+      const values = successfulResponses.map((response) => {
+        return {
+          signer: response.url,
+          performedQueryCount: response.signMessageResponse?.performedQueryCount,
+          totalQuota: response.signMessageResponse?.totalQuota,
+        }
+      })
+      logger.error(`Discepancy found in signers' measured quotas ${values}`)
```
aslawson: nit -- `Discepancy`  --> `Discrepancy`

.env
```diff
@@ -177,4 +177,5 @@ VOTING_BOT_SCORE_SENSITIVITY=5
 VOTING_BOT_CRON_SCHEDULE="*/15 * * * *"
 
 # PGPNP Service Tests
-PHONE_NUM_PRIVACY_SERVICE_URL=http://localhost:5000/celo-phone-number-privacy/us-central1
\ No newline at end of file
+PHONE_NUM_PRIVACY_SERVICE_URL=https://staging-pgpnp-signer0.azurefd.net
```
jmrossy: Should we rename this?

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
```diff
@@ -12,38 +12,53 @@ function flattenSigsArray(sigs: Uint8Array[]) {
...
+        Failed to verify signature for ${serviceResponse.url}`,
+        e
+      )
     }
+  }
+
+  /**
+   * Returns true if the number of valid signatures is enough to perform a combination
+   */
+  public sufficientVerifiedSignatures(): boolean {
```
jmrossy: suggest `hasSufficientVerifiedSignatures`

packages/phone-number-privacy/combiner/src/config.ts
```diff
@@ -1,3 +1,4 @@
+import { toBool } from '@celo/phone-number-privacy-common/lib/utils/config-utils'
```
jmrossy: Can you set have the common's `index.ts` export things so we don't need to reach into `lib/...` like this. 
Here's an example:
https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/index.ts#L1

packages/phone-number-privacy/combiner/src/database/wrappers/number-pairs.ts
```diff
@@ -1,7 +1,7 @@
-import { ErrorMessage } from '../../common/error-utils'
+import { ErrorMessage } from '@celo/phone-number-privacy-common'
 import logger from '../../common/logger'
 import { getDatabase } from '../database'
-import { NUMBER_PAIRS_COLUMN, NUMBER_PAIRS_TABLE, NumberPair } from '../models/numberPair'
+import { NumberPair, NUMBER_PAIRS_COLUMN, NUMBER_PAIRS_TABLE } from '../models/numberPair'
```
jmrossy: I suspect this will fail the linter :)
VS Code's ordering differs from prettier. I don't know of a good way to fix this so I usually reorder them manually then to `command+shift+p` then choose 'Save without formatting'

yarn.lock
```diff
@@ -656,9 +656,9 @@
     adal-node "^0.1.28"
 
 "@babel/cli@^7.10.1":
-  version "7.10.3"
-  resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.10.3.tgz#4ea83bd997d2a41c78d07263ada3ec466fb3764b"
-  integrity sha512-lWB3yH5/fWY8pi2Kj5/fA+17guJ9feSBw5DNjTju3/nRi9sXnl1JPh7aKQOSvdNbiDbkzzoGYtsr46M8dGmXDQ==
+  version "7.10.4"
```
jmrossy: The large diff here tells me you ran `yarn` in a folder other than monorepo root. That messes things up unfortunately. I'd recommend adding dependencies manually.
Re-run yarn from root to fix this lock file :)

jmrossy (commented):
Mostly looks great! Please see comments below

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
codyborn: Previously this class was a stateless collection of static methods. Now, `verifiedSignatures` is a private property of this class that keeps track of verifiedSignatures that are passed to it. So the caller would first create a new `BLSCryptographyClient` (which initializes this array to be empty). Then the caller can call `client.addSignature()` to tell this class to collect the signature if it's valid. Let me know if I didn't address your question.

packages/phone-number-privacy/combiner/src/database/wrappers/number-pairs.ts
codyborn: Ah I was wondering why it kept complaining with the changes it was auto-making.. thanks!

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
codyborn: Great point!

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
codyborn: Yes, thanks for checking.

packages/phone-number-privacy/combiner/src/config.ts
codyborn: Thanks! Missed the export for this file

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -105,12 +105,100 @@ async function requestSignatures(request: Request, response: Response) {
...
+  sentResult: { sent: boolean },
+  responses: SignMsgRespWithStatus[],
+  serviceUrl: string,
+  blsCryptoClient: BLSCryptographyClient,
+  blindedQueryPhoneNumber: string
+) {
+  const signResponse = (await res.json()) as SignMessageResponse
+  if (!signResponse.success) {
+    // Continue on failure as long as signature is present to unblock user
+    const signResponseFailure = signResponse as SignMessageResponseFailure
```
bowd: So this is exactly what I was trying to avoid, the pesky use of `x as y` which one would call "the death of passion" (direct translation from Romanian 😂).

I have to bring up @mcortesi 's point of erring towards union types with discriminator fields that help the compiler figure out object structure and allows us to avoid casting.

Here's an example of what I think the type definitions could look like: https://codesandbox.io/s/intelligent-banzai-ciqr1

This now takes into account your point that the actual body of the response could contain all fields in either the success or error case so there's a base interface that both cases extend, and both specific cases set the discriminator as the `success` value, and the failure case adds the `error` attribute. If there are other attributes that can be split they should be removed from the base interface.

Then the crux of the solution is the fact that we introduce a union type for the two interfaces. This is important because the compiler can then infer which branch of the union you are on based on the code. So once you've checked `if (!signResponse.success)` the compiler will automatically pick the right interface in each branch of the `if` statement and it will just flow naturally.

While granted this isn't super useful on first glance mostly because of the shared structure of the Failure and Success cases, it still provides a more solid starting point to better specify these structures. For example, if "error" is more like a constant strings than a human readable error message it could act as a further discriminator allowing the type definition to be split between different error types that each specify their fields. I included an imagined scenario of this in the `more-errors.ts` of the sandbox, just to showcase multi-level discriminators.

bowd: There's also the one above `(await res.json()) as SignMessageResponse`. That can only be avoided by the runtime types discussed, but we can live with it for now because of the overhead, but we should seriously think about enforcing that boundary as well.

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
bowd: Wait, sorry to bring this up again but @aslawson's right, unless the intent here is something else.
https://codesandbox.io/s/jovial-architecture-gh3sx?file=/src/index.ts

You're returning from the inside of the arrow function and that won't short-circuit the `forEach`, on way of doing that is using the `some` iterator instead of `forEach` that will stop as soon as you return `true` from the arrow-function.

packages/phone-number-privacy/combiner/src/database/wrappers/number-pairs.ts
bowd: I haven't used VSCode in a while but I'm certain there's a `prittier` plugin that can be used for autoformatting so that it abides by the same rules as the linter.

packages/phone-number-privacy/combiner/src/bls/bls-cryptography-client.ts
aslawson: oh i see now :)

packages/phone-number-privacy/signer/src/salt-generation/get-salt.ts
```diff
@@ -52,7 +52,9 @@ export async function handleGetBlindedMessageForSalt(
     // In the case of a DB or blockchain connection failure, don't block user
     // but set the error status accordingly
     try {
-      ;[performedQueryCount, totalQuota] = await getRemainingQueryCount(account, hashedPhoneNumber)
+      const queryCount = await getRemainingQueryCount(account, hashedPhoneNumber)
+      performedQueryCount = queryCount.performedQueryCount
+      totalQuota = queryCount.totalQuota
     } catch (error) {
```
aslawson: i think you can set it similar to how you had it before `const { performedQueryCount, totalQuota } = await getRemainingQueryCount(account, hashedPhoneNumber)`

like https://github.com/celo-org/celo-monorepo/blob/amyslawson/combiner_second_region/packages/phone-number-privacy/signer/src/salt-generation/get-salt.ts#L38

aslawson (approved):

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
jmrossy: Let's use a for loop. It's 2-10x faster anyway ;)

aslawson: what makes it faster?

bowd: I don't know if there's something else to it, but iterating with forEach means creating closures, calling function, call stack manipulation etc. for loops are just jumps. 

But then again this also seams like a perfect target for some special case optimisations in V8 🎃

codyborn: oh wow, just when you think you know Javascript 😅 thanks again!

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
codyborn: Discrepancy in my spelling, thanks!

packages/phone-number-privacy/signer/src/salt-generation/get-salt.ts
codyborn: Just tested it out.. If I do that, I'll still need to assign them to the variables defined before the try clause. This assignment treats them as new variables instead of assigning the value to the previously defined ones. So in the end, I'll need the same number of lines for the assignment.

packages/phone-number-privacy/combiner/src/salt-generation/get-threshold-salt.ts
```diff
@@ -105,12 +108,99 @@ async function requestSignatures(request: Request, response: Response) {
...
+      response.signMessageResponse.totalQuota &&
+      response.signMessageResponse.blockNumber
   )
 
-  return { successCount, majorityErrorCode }
+  if (successfulResponses.length === 0) {
+    return
+  }
+  // Compare the first response to the rest of the responses
+  const expectedQueryCount = successfulResponses[0].signMessageResponse!.performedQueryCount!
```
jmrossy: This highlights the problem with using `filter`, `map`, or `reduce`. Though the functional style is nice, the TS compiler can't figure out that these shouldn't be necessary now. It's 'old school' in style but a for loop with a new array would have fixed the type checking

jmrossy (approved):
Nice work here!

PR #4335, opened 2020-07-07
[Wallet] Fix ContractKit init issues causing Pincode entry errors

jmrossy:
### Description

- Improve error handling in ContractKit RPC Signer to actually return unlock status
- Refactor ContractKit and GethWallet initiation code to prevent waiting for sync and simplify
- Remove superfluous ContractKitReady state
- Remove promisify generator util as it incorrectly creates impression Redux Sagas can be called from outside sagas
- Properly reset geth init status on geth saga cancel
- Wait for rehydrate before starting sagas to prevent them from running on incorrect state
- Add `getWeb3()` helpers
- Add more helpful error message on account unlock failure
- Fix bug with invite flow that @i1skn diagnosed

### Tested

Did onboarding, sent tx, enabled forno, and enabled pin-on-startup on both iOS and Android
Tested forcing an IPC not ready case, the retry logic worked.

### Backwards compatibility

Yes


packages/mobile/src/web3/providers.ts
```diff
@@ -0,0 +1,55 @@
+import * as net from 'react-native-tcp'
```
jmrossy, line 1: No new code here, just moved these functions from `web3/contracts` for better organization.

packages/mobile/src/localCurrency/saga.ts
```diff
@@ -59,7 +58,7 @@ export function* watchFetchCurrentRate() {
 
 export function* watchSelectPreferredCurrency() {
   while (true) {
-    yield take([Actions.SELECT_PREFERRED_CURRENCY, REHYDRATE, AccountActions.SET_PHONE_NUMBER])
+    yield take([Actions.SELECT_PREFERRED_CURRENCY, AccountActions.SET_PHONE_NUMBER])
```
cmcewen, line 62: why remove this here? this will ensure we fetch the local currency on app startup

packages/contractkit/src/wallets/signers/rpc-signer.ts
```diff
@@ -103,11 +105,20 @@ export class RpcSigner implements Signer {
...
+    let unlocked = false
+    try {
+      unlocked = await this.callAndCheckResponse(RpcSignerEndpoint.UnlockAccount, [
+        this.account,
+        passphrase,
+        duration,
+      ])
+    } catch (error) {
+      // Only throw errors other than expected unlock error
+      if (!error?.message?.toLowerCase().includes(UNLOCK_ACCOUNT_EXPECTED_ERROR)) {
```
cmcewen: why is this error expected? and should we not set the `unlockTime` below if it's not successful?

packages/mobile/src/geth/geth.ts
```diff
@@ -108,12 +108,12 @@ async function createNewGeth(sync: boolean = true): Promise<typeof RNGeth> {
   return new RNGeth(gethOptions)
 }
 
-async function initGeth(sync: boolean = true) {
+export async function initGeth(sync: boolean = true): Promise<typeof gethInstance> {
```
cmcewen, line 113: 😄

packages/mobile/src/app/saga.ts
```diff
@@ -34,18 +32,9 @@ const TAG = 'app/saga'
...
-  }
-  yield take(REHYDRATE)
-  return
-}
-
-export function* watchRehydrate() {
-  yield take(REHYDRATE)
-
+// Work that's done before other sagas are initalized
+// Be mindful to not but long blocking tasks here
```
tarikbellamine: nit: typo

packages/mobile/src/redux/sagas.ts
```diff
@@ -65,25 +65,31 @@ function* loggerSaga() {
 }
 
 export function* rootSaga() {
+  // Delay all sagas until rehydrate is done
+  // This prevents them from running with missing state
+  yield call(waitForRehydrate)
```
cmcewen, line 70: nice 🙌

packages/mobile/src/geth/saga.ts
```diff
@@ -65,7 +78,7 @@ function* waitForGethInstance() {
   try {
     const fornoMode = yield select(fornoSelector)
     // get geth without syncing if fornoMode
-    const gethInstance = yield call(getGeth, !fornoMode)
+    const gethInstance = yield call(initGeth, !fornoMode)
     if (gethInstance == null) {
       throw new Error('geth instance is null')
```
tarikbellamine: Should this error message change given you're now returning `null` if there's already a geth creation in progress?

packages/mobile/src/web3/contracts.ts
```diff
@@ -1,128 +1,112 @@
...
-  }
-  return contractKit
+  const fornoMode = yield select(fornoSelector)
+  Logger.info(`${TAG}@initContractKit`, `Initializing contractkit, forno mode: ${fornoMode}`)
+
+  ipcProvider = getIpcProvider()
+  const web3 = new Web3(fornoMode ? getHttpProvider(DEFAULT_FORNO_URL) : ipcProvider)
+
+  Logger.debug(`${TAG}@initContractKit`, `Initializing wallet`)
+  gethWallet = new RpcWallet(ipcProvider)
```
cmcewen: do we need a gethWallet if we aren't in forno mode? cc @yorhodes

cmcewen (commented):
I'm a little nervous about a lot of new code here right now. Also bummed about the generator/promise interface although it doesn't seem like there is an easy solution right now. Do you think it's worth doing a release build to test this outside dev mode?

packages/mobile/src/invite/saga.ts
```diff
@@ -217,7 +216,7 @@ export function* sendInvite(
 }
 
 function* initiateEscrowTransfer(temporaryAddress: string, e164Number: string, amount: BigNumber) {
-  const escrowTxId = generateStandbyTransactionId(temporaryAddress + '-escrow')
```
tarikbellamine, line 220: Do you happen to know why this was needed in the first place? Just want to make sure it's not an identifier for any downstream logic

packages/mobile/src/web3/contracts.ts
yorhodes: interesting point
we do not technically need it but would need to refactor to work without it

packages/mobile/src/import/saga.ts
```diff
@@ -45,10 +44,7 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
 
     if (!useEmptyWallet) {
       Logger.debug(TAG + '@importBackupPhraseSaga', 'Checking account balance')
-      const contractKit = yield call(getContractKit)
-      const backupAccount = contractKit.web3.eth.accounts.privateKeyToAccount(
-        ensureLeading0x(privateKey)
-      ).address
+      const backupAccount = privateKeyToAddress(privateKey)
```
yorhodes, line 47: 👏

packages/contractkit/src/wallets/signers/rpc-signer.ts
yorhodes: +1
can this try catch be raised above this interface?

packages/mobile/src/utils/generators.ts
```diff
@@ -1,19 +0,0 @@
-// adapted from https://www.promisejs.org/generators/
-export function promisifyGenerator<T>(generator: Generator<any, T, unknown>) {
```
yorhodes, line 2: anywhere I can read about why this doesn't work with redux sagas?
mostly just curious

packages/mobile/src/web3/utils.ts
```diff
@@ -27,21 +27,23 @@ export async function estimateGas(txObj: TransactionObject<any>, txParams: Tx) {
...
+  const web3 = await getWeb3Async()
+  return web3.eth.getBlock('latest')
 }
 
 export async function getLatestBlockNumber() {
   Logger.debug(TAG, 'Getting latest block number')
-  const contractKit = await getContractKitAsync()
-  return contractKit.web3.eth.getBlockNumber()
+  const web3 = await getWeb3Async()
+  return web3.eth.getBlockNumber()
```
yorhodes, line 37: love these changes

packages/mobile/src/web3/contracts.ts
annakaz: Is all private key storage handled in the RPC wallet? It could be worth keeping if we decide to switch to clef down the line

packages/mobile/src/geth/saga.ts
jmrossy: Good question. I believe the checking for specifically null instead of all falsy values was just an oversight. I'll fix this.

packages/mobile/src/localCurrency/saga.ts
jmrossy: My mistake. It's removed cause the watcher won't be running when REHYDRATE is dispatched. Easily fixable now by just fetching the rate once when this watcher starts.

packages/contractkit/src/wallets/signers/rpc-signer.ts
jmrossy: It's because the response from the RPC call includes an error, so in `callAndCheckResponse` below an error is thrown. That method is used in several places so I figured here was a reasonable place to intercept the error. Open to other suggestions

packages/mobile/src/web3/contracts.ts
jmrossy: Is there a downside to using it? Splitting the code paths would mean more code to test.

packages/mobile/src/invite/saga.ts
jmrossy: Like @yorhodes said, this was likely to avoid possible conflict with the invite tx but they would need to be created at the same millisecond. Given that web3 has been quietly ripping out the '-escrow' for us and this has never been an issue, it's unlikely this will cause issues.

packages/mobile/src/utils/generators.ts
jmrossy: There's no single place that explains this unfortunately, just small clues about how Saga actually works sprinkled throughout the docs and the github issues
https://redux-saga.js.org/

jmrossy:
> I'm a little nervous about a lot of new code here right now. Also bummed about the generator/promise interface although it doesn't seem like there is an easy solution right now. Do you think it's worth doing a release build to test this outside dev mode?

@cmcewen I'm confident these changes are more stable than what's live but yep we can do a release build if you prefer.

packages/mobile/src/localCurrency/saga.ts
jmrossy: Fixed

packages/contractkit/src/wallets/signers/rpc-signer.ts
aaronmgdr: I think you need an extra  `?.` so includes isnt called on null

packages/mobile/src/web3/contracts.ts
```diff
@@ -1,128 +1,147 @@
...
-    contractKit = newKitFromWeb3(web3, wallet)
+      const fornoMode = yield select(fornoSelector)
+      Logger.info(`${TAG}@initContractKit`, `Initializing contractkit, forno mode: ${fornoMode}`)
+
+      ipcProvider = getIpcProvider()
+      const web3 = new Web3(fornoMode ? getHttpProvider(DEFAULT_FORNO_URL) : ipcProvider)
+
+      Logger.info(`${TAG}@initContractKit`, 'Initializing wallet')
+      gethWallet = new RpcWallet(ipcProvider)
+      yield call([gethWallet, gethWallet.init])
```
yorhodes, line 51: it seems like we lose the progress of previous retries with this pattern

packages/mobile/src/web3/contracts.ts
```diff
@@ -1,128 +1,147 @@
...
+    yield call(initContractKit)
   }
-  return wallet
+  return contractKit
 }
 
-export const getContractKitAsync = async () => promisifyGenerator(getContractKit())
+// Used for cases where CK must be access outside of a saga
+export async function getContractKitAsync() {
+  let retries = 10
```
yorhodes, line 100: I think it would be nice to abstract the retry logic so that we can swap it out with something nicer in future

packages/mobile/src/geth/geth.ts
```diff
@@ -20,6 +20,8 @@ export const FailedToFetchGenesisBlockError = new Error(
   'Failed to fetch genesis block from Google storage'
 )
 
+export const PROVIDER_CONNECTION_ERROR = "connection error: couldn't connect to node"
```
yorhodes, line 23: this is different from the previous `BLOCK_CHAIN_CORRUPTION_ERROR`?

yorhodes (approved):

packages/mobile/src/geth/geth.ts
jmrossy: Just a shorter but still unique subset in case the error message gets formatted slightly differently

packages/mobile/src/web3/contracts.ts
jmrossy: Yeah, there may be opportunities to DRY things up but it tends to be we have quite unique error handling + flow management in the major retry blocks (like web3 syncing)

packages/mobile/src/web3/contracts.ts
jmrossy: Correct. Seemed like the safer thing to do. Do you think it'll leave behind state that'll be problematic?

packages/mobile/src/web3/contracts.ts
yorhodes: I guess just memory as well as potentially more round trips over the RPC (local)

packages/contractkit/src/wallets/signers/rpc-signer.ts
yorhodes: Connor's original point still stands that if we suppress this error, we need to not set the `unlockTime` and `unlockDuration`
My main qualm is for other consumers of this interface (of which there are none right now 🙄) this behavior could be confusing

jmrossy: Thanks for the eyes on this guys! I'll fix the `unlockDuration` and error safe accessing now.

RE interface, I'm open to changing this back but I'd argue this is more intuitive for consumers. Before, this could never return false. Only true or it throws. So why return anything?

jmrossy: And the problem with it never returning false is that consumers have not easy way to differentiate a wrong password from an error, other than catching and inspecting the error for this message `could not decrypt key with given password`.

cmcewen (approved):

PR #4336, opened 2020-07-07
[Wallet] Payment limit bug fix

tarikbellamine:
### Description
Fixed bug that resulted in false errors regarding users being over the daily transaction limit. This was caused for two reasons:
- Amount was being incorrectly stored in redux as a non-primitive data type (BigNumber) yet the created interface was defining it as a number, which resulted in some faulty math
- Payments were being stored on transaction initiation, not success, which would mean failed transactions would incorrectly count against the daily limit

### Other changes
Minor cleanup

### Tested
Yes

### Related issues

- Part of #3988 

### Backwards compatibility
Created redux store migration.

### DIFF
--- a/packages/mobile/src/fiatExchanges/FiatExchangeAmount.test.tsx
+++ b/packages/mobile/src/fiatExchanges/FiatExchangeAmount.test.tsx
@@ -42,7 +42,7 @@ describe('FiatExchangeAmount', () => {
     fireEvent.changeText(getByTestId('FiatExchangeInput'), '750')
     expect(getByTestId('FiatExchangeNextButton').props.disabled).toBe(false)
     fireEvent.press(getByTestId('FiatExchangeNextButton'))
-    expect(store.getActions()).toContainEqual(showError(ErrorMessages.PAYMENT_LIMIT_REACHED, null))
+    expect(store.getActions()).toContainEqual(showError(ErrorMessages.PAYMENT_LIMIT_REACHED, 5000))
   })
 
   it('validates the amount when adding funds', () => {
@@ -59,6 +59,6 @@ describe('FiatExchangeAmount', () => {
     fireEvent.changeText(getByTestId('FiatExchangeInput'), '750')
     expect(getByTestId('FiatExchangeNextButton').props.disabled).toBe(false)
     fireEvent.press(getByTestId('FiatExchangeNextButton'))
-    expect(store.getActions()).toContainEqual(showError(ErrorMessages.PAYMENT_LIMIT_REACHED, null))
+    expect(store.getActions()).toContainEqual(showError(ErrorMessages.PAYMENT_LIMIT_REACHED, 5000))
   })
 })
--- a/packages/mobile/src/fiatExchanges/FiatExchangeAmount.test.tsx
+++ b/packages/mobile/src/fiatExchanges/FiatExchangeAmount.test.tsx
@@ -3,6 +3,7 @@ import { fireEvent, render } from 'react-native-testing-library'
 import { Provider } from 'react-redux'
 import { showError } from 'src/alert/actions'
 import { ErrorMessages } from 'src/app/ErrorMessages'
+import { ALERT_BANNER_DURATION } from 'src/config'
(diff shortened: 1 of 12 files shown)


packages/mobile/src/send/reducers.ts
```diff
@@ -42,8 +40,22 @@ export const sendReducer = (
...
+        isSending: true,
+      }
     case Actions.SEND_PAYMENT_OR_INVITE_SUCCESS:
+      const now = Date.now()
+      // Keep only the last 24 hours
+      const paymentsLast24Hours = state.recentPayments.filter(
+        (p: PaymentInfo) => timeDeltaInHours(now, p.timestamp) < 24
+      )
+      const latestPayment: PaymentInfo = { amount: action.amount.toNumber(), timestamp: now }
+      return {
```
cmcewen, line 54: Do you need to set `isSending` to false here?

packages/mobile/src/send/utils.ts
```diff
@@ -101,5 +101,6 @@ export function showLimitReachedError(
     dailyRemainingcUSD,
     dailyLimitcUSD: DAILY_PAYMENT_LIMIT_CUSD,
   }
-  return showError(ErrorMessages.PAYMENT_LIMIT_REACHED, null, translationParams)
+  const dismissAfterInMilliSecs = 5000
```
cmcewen: you can define this outside the function as a constant

cmcewen (approved):
small comments but otherwise looks good!

etuleu:
I don't have strong feelings on this one but I believe the initial reasoning for appending transactions on initiation instead of success was that an "abuser" could in theory initiate multiple transactions while on airplane mode that could add up to more than the limit. I don't know how big of a deal that is. Maybe we could just get @nityas attention on this just to be sure.

i1skn:
> I don't have strong feelings on this one but I believe the initial reasoning for appending transactions on initiation instead of success was that an "abuser" could in theory initiate multiple transactions while on airplane mode that could add up to more than the limit. I don't know how big of a deal that is. Maybe we could just get @nityas attention on this just to be sure.

Considering that this limit is only imposed in the Valora app I would guess that real "abuser" would use other tools, cause this limit is not enforced on the blockchain level.

LGTM after adressing @cmcewen comments 👍 Great work on catching this!

i1skn (approved):

packages/mobile/src/send/reducers.ts
tarikbellamine: I would think so but that's not how it was setup before. Shouldn't hurt to add that in so I'll go ahead and do that.

packages/mobile/src/fiatExchanges/FiatExchangeAmount.test.tsx
```diff
@@ -42,7 +42,7 @@ describe('FiatExchangeAmount', () => {
     fireEvent.changeText(getByTestId('FiatExchangeInput'), '750')
     expect(getByTestId('FiatExchangeNextButton').props.disabled).toBe(false)
     fireEvent.press(getByTestId('FiatExchangeNextButton'))
-    expect(store.getActions()).toContainEqual(showError(ErrorMessages.PAYMENT_LIMIT_REACHED, null))
+    expect(store.getActions()).toContainEqual(showError(ErrorMessages.PAYMENT_LIMIT_REACHED, 5000))
```
jmrossy: Consider importing `ALERT_BANNER_DURATION` instead of hardcoding these

packages/mobile/src/send/utils.ts
```diff
@@ -14,6 +14,8 @@ import { PaymentInfo } from 'src/send/reducers'
 import { TransactionDataInput } from 'src/send/SendAmount'
 import { timeDeltaInHours } from 'src/utils/time'
 
+const DISMISS_SPEND_LIMIT_BANNER_TIMER = 5000
```
jmrossy: Dupe of https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/config.ts#L38

jmrossy (commented):
Looks good, just two minor comments. Removing automerge so you will see them before this merges.

PR #4337, opened 2020-07-07
Add blockNumber to signer response

codyborn:
### Description

Add blockNumber to Signer response interface and bump version.

### Backwards compatibility

Yes, adding an optional property.


packages/phone-number-privacy/common/src/interfaces/SignMessageResponse.ts
```diff
@@ -5,4 +5,5 @@ export interface SignMessageResponse {
   error?: string
   performedQueryCount?: number
   totalQuota?: number
+  blockNumber?: number
```
bowd, line 7: I know this isn't really the time or place but I'm curios what other's people's thoughts are on this. I'm not a big fan of having interfaces that have predominantly optional attributes, I think it subtracts from the benefits of Typescript, and is actually code smell for a design problem.

I'm assuming that the situation here is caused by this response having two possible shapes based on if `success` is true or false, if `false`, it only has the `error` attribute, if `true` it has all the rest except `error`. There are better ways of handling this, and I've setup an example here: https://codesandbox.io/s/jolly-yalow-k026k

The example features two possible implementations:
- `pure.ts` a pure typescript one that enforces types at the compiler level but doesn't do runtime validation on data incoming from external sources (which is the case here as well I would imagine)
- `runtypes.ts` which includes the popular [runtypes](https://github.com/pelotom/runtypes) library that extends the enforcement in the previous example and ales includes runtime "contracts" that play nicely with the TS compiler and type inference. 

I like both options in favour of the current way of doing it and `runtypes` in favour of `pure` because it's an elegant solution for the boundary between trust in typescript compiler and trust in the incoming data.

bowd (approved):
:+1:

packages/phone-number-privacy/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celo/phone-number-privacy-common",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
jmrossy: Can you rev the signer + combiner versions too and keep them in sync?

jmrossy (approved):

packages/phone-number-privacy/common/src/interfaces/SignMessageResponse.ts
codyborn: Hey @bowd! Thanks for starting this discussion. As someone with a background in strongly typed languages I definitely appreciate the benefits of runtypes.  Part of the reason these are optional is as you mentioned: we need to handle differences in error/success responses. The other two reasons are:
1. Backwards compatibility - signers (9 of them) are operated by us and third parties and we have little control over exactly when they're upgraded. Thus, we must support the case where many of these fields aren't present.
2. Failing gracefully - this service is on the critical path for user onboarding and there are service failure scenarios where we still want to return a partial response to ensure users are not blocked by an outage. Therefore, performedQueryCount, totalQuota, and blockNumber may be unset if there are DB/blockchain connection failures. This also means that signature value may or may not be specified in the case of `success === false`.

So although it won't completely solve the optional attributes problem, I still think it's useful to be explicit at compile time to define the properties we expect to be present and absent. Would like to also here thoughts from @jmrossy and @aslawson on this.

packages/phone-number-privacy/common/package.json
codyborn: I was thinking more about this and feel that it's fine if the versions diverge across all three libs. We may frequently role out combiner updates going fwd and will not want to constantly bump the signer version with no changes. Currently signer is on v1.0.3 and the combiner is on v1.0.0. Perhaps it makes most sense to keep the major versions in sync but not necessarily the minor/patch. WDYT?

jmrossy: Ok works

packages/phone-number-privacy/common/src/interfaces/SignMessageResponse.ts
aslawson: i agree on both ends --  i am also quite new to javascript and have a background in strongly typed languages, so it sets off flags to overload an object with optional fields.

i'm not sure how the unions resolve as a javascript object -- depending if the fields are just appended or if the specific fields are nested in another object, it may or may not be doable in a backwards capable manner.

for this use case even if we have the union, many of these fields would be still be optional fields for the success use case so we wouldn't gain much from the refactor (what cody mentioned in 2)

aslawson (approved):
looks fine

what's the reason we wanted to add this?

PR #4338, opened 2020-07-07
Make adjustments to the attestation bot to pass a salt

nambrot:
Make adjustments to the attestation bot to pass a salt


gastonponti (commented):
I have some doubts on the `export const SALT = 'hard_coded_salt'` thing.
I'm not that into this particular command with the celotool, but usually the salt is used to change an output, and it's something unique for every instance, but if it doesn't changes, it doesn't seems to be helpful to me.
The name `pepper` is used for secrets that won't change between every instance

So, if it is something fixed, first i would recommend to make it configurable, instead of hardcoded, and chance the name at least to pepper.

If it's salt, and this hardcode it's just for testing o some base case that I don't know, I would recommend to at least have the other functions easy configurable for salt rotation
In this case, for example
```
async function verifyCmd(argv: VerifyArgv) {
```
change it to something that it's clear that it's possible to rotate the salt
```
async function verifyCmd(argv: VerifyArgv, salt: string = SALT) {
```
or maybe receive it as a key in `argv`, and if it's undefined, use the hardcoded one

Again, I don't really know the problem, so feel free to dismiss this comment, haha.

PR #4339, opened 2020-07-07
Migrate from old bip39 address to new one

cmcewen:
### Description

This tries to make the migration process for existing users seamless. Still needs logic around only migrating once and checking to see whether it is necessary


packages/mobile/src/app/saga.ts
```diff
@@ -134,9 +148,66 @@ export function* handleSetAppState(action: SetAppState) {
...
 }
 
+function* migrateAccountToProperBip39() {
+  const account: string | null = yield select(currentAccountSelector)
+  const e164Number: string | null = yield select(e164NumberSelector)
+  if (!account) {
+    throw new Error('account not set')
+  }
+  if (!e164Number) {
+    throw new Error('e164 numbrer not sset')
```
annakaz:
```suggestion
    throw new Error('e164 number not set')
```

packages/mobile/src/invite/saga.ts
```diff
@@ -381,11 +384,13 @@ function* addTempAccountToWallet(inviteCode: string) {
...
   tempAccountBalanceWei: BigNumber,
-  newAccount: string
+  newAccount: string,
+  currency?: CURRENCY_ENUM
 ) {
   Logger.debug(TAG + '@withdrawFundsFromTempAccount', 'Unlocking temporary account')
   const wallet: RpcWallet = yield call(getConnectedWallet)
-  yield call([wallet, wallet.unlockAccount], tempAccount, TEMP_PW, 600)
+  const password: string = yield call(getPasswordSaga, tempAccount, false, true)
+  yield call([wallet, wallet.unlockAccount], tempAccount, password, 600)
```
annakaz: For other authentication we use an unlock duration of 300- https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/geth/consts.ts#L2
Perhaps worth switching to as the tempAccount could have significantly more funds than the usual invite?

packages/mobile/src/invite/saga.ts
```diff
@@ -412,7 +417,7 @@ export function* withdrawFundsFromTempAccount(
 
   const tx: CeloTransactionObject<boolean> = yield call(
     createTokenTransferTransaction,
-    CURRENCY_ENUM.DOLLAR,
+    currency ? currency : CURRENCY_ENUM.DOLLAR,
     {
       recipientAddress: newAccount,
       amount: netSendAmount,
```
annakaz: Should update `SENTINEL_INVITE_COMMENT` here and in fee estimation above to some migration comment

packages/mobile/src/identity/revoke.ts
```diff
@@ -0,0 +1,51 @@
...
+import { setNumberVerified } from 'src/app/actions'
+import { resetVerification } from 'src/identity/actions'
+import { fetchPhoneHashPrivate } from 'src/identity/privateHashing'
+import { getAttestationsStatus } from 'src/identity/verification'
+import Logger from 'src/utils/Logger'
+import { getContractKit } from 'src/web3/contracts'
+
+const TAG = 'identity/revoke'
+
+export function* revokePhoneMapping(e164Number: string, account: string) {
```
annakaz, line 15: 👨‍🎨   can use this if people want to change their phone numbers as well

annakaz (commented):
Looks great, good call on using the invite logic

packages/mobile/android/gradle.properties
```diff
@@ -23,7 +23,7 @@ CELO_RELEASE_KEY_ALIAS=celo-key-alias
 
 # Setting this manually based on version number until we have this deploying via Cloud Build
 # Example: v1.5.1 deployment number 1 = 100500101 (1 005 001 01)
-VERSION_CODE=100800202
+VERSION_CODE=100900002
```
jmrossy: If we're rev-ing this please run the pre-deploy script to also change the version in the package.json and ios build files

packages/mobile/src/app/saga.ts
```diff
@@ -134,9 +148,66 @@ export function* handleSetAppState(action: SetAppState) {
...
+      account,
+      stableTokenBalance,
+      newAccount,
+      CURRENCY_ENUM.DOLLAR
+    )
+  }
+}
+
+export function* watchMigrateAccountToProperBip39() {
+  yield takeLatest(Actions.MIGRATE_ACCOUNT_BIP39, migrateAccountToProperBip39)
```
jmrossy: Nit: This could probably just be `take`, so users only do it once.

packages/mobile/src/firebase/firebase.ts
```diff
@@ -69,7 +69,7 @@ export const initializeAuth = async (app: ReactNativeFirebase.Module, address: s
     } else if (userData.address !== undefined && userData.address !== address) {
       // This shouldn't happen! If this is thrown it means the firebase user is reused
       // with different addresses (which we don't want) or the db was incorrectly changed remotely!
-      throw new Error("User address in the db doesn't match persisted address")
+      // throw new Error("User address in the db doesn't match persisted address")
```
jmrossy: Do we need to do a firebase 'reset' as part of the migration? I believe there should be a way to get FB to make another anon id for the user.

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -109,6 +117,12 @@ export class WalletHome extends React.Component<Props> {
     // Waiting 1/2 sec before triggering to allow
     // rest of feed to load unencumbered
     setTimeout(this.tryImportContacts, 500)
+
+    this.checkMigration()
+  }
+
+  async checkMigration() {
+    this.setState({ showMigration: true })
   }
```
jmrossy: Why is this async, will you be adding to it?

packages/mobile/src/invite/saga.ts
jmrossy: +1 use UNLOCK_DURATION const

packages/mobile/src/invite/saga.ts
```diff
@@ -412,7 +417,7 @@ export function* withdrawFundsFromTempAccount(
 
   const tx: CeloTransactionObject<boolean> = yield call(
     createTokenTransferTransaction,
-    CURRENCY_ENUM.DOLLAR,
+    currency ? currency : CURRENCY_ENUM.DOLLAR,
```
jmrossy: Nit: `currency ?? CURRENCY_ENUM.DOLLAR`

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -191,6 +208,15 @@ export class WalletHome extends React.Component<Props> {
           keyExtractor={this.keyExtractor}
         />
         <SendOrRequestBar />
+        <Dialog
+          title={'Migrate to new Address'}
+          children={
+            'Due to a developer configuration error, you will need to migrate your account to a new address, which will include verifying again. You can keep your seed phrase. Do not close the app during migration. Please post on slack if you have questions.'
+          }
```
jeanregisser, line 215: 👍

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -147,6 +159,11 @@ export class WalletHome extends React.Component<Props> {
     )
   }
 
+  migrateAccount = () => {
+    this.setState({ isMigrating: true })
```
jeanregisser, line 163: nit: Though there's no bug now, we never bring this component state back to false, which could be confusing in the future. This should probably be handled by the redux store.

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -28,8 +28,8 @@ describe(getPasswordSaga, () => {
       .returns(expectedPassword)
       .run()
 
-    expect(navigate).toHaveBeenCalled()
-    expect(navigateBack).toHaveBeenCalled()
+    // expect(navigate).toHaveBeenCalled()
+    // expect(navigateBack).toHaveBeenCalled()
```
jeanregisser, line 32: Is this intentional?

jeanregisser (approved):
Awesome work! 🥇 

This is probably fine, but if I'm not mistaken, this won't migrate people who restore from their seed which was created before this change. Right?

PR #4340, opened 2020-07-07
Remove react-devtool package

jmrossy:
### Description

It includes electron, which is a big package that we don't currently need.
Was causing this security alert: https://github.com/celo-org/celo-monorepo/network/alert/yarn.lock/electron/open


i1skn:
LGTM, but please confirm before merging that `Debug`, `Show Inspector` and `Flipper` would still be working after removing this package.

cmcewen (approved):

PR #4344, opened 2020-07-07
[Wallet] Add onboarding education

jeanregisser:
### Description

This PR adds the [onboarding education screen](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=5404%3A1099).

We now also skip the language screen when we can directly detect a supported locale.

See assets diff: https://github.com/clabs-co/valora-app-branding/compare/d5249be...8a74a2c

### Tested

Updated jest and e2e tests.

### Related issues

- #3645 

### Backwards compatibility

Yes


packages/mobile/src/account/Education.tsx
```diff
@@ -68,6 +94,11 @@ export default class Education extends React.Component<Props, State> {
           currentStep: step,
           direction: ScrollDirection.previous,
         })
+      } else if (topic === EducationTopic.onboarding) {
+        ValoraAnalytics.track(OnboardingEvents.onboarding_education_scroll, {
+          currentStep: step,
+          direction: ScrollDirection.previous,
+        })
```
jeanregisser, line 101: This looks like it could be refactored to avoid repetition.
I didn't want to disrupt more code than necessary at this stage though.

packages/mobile/src/account/Education.tsx
```diff
@@ -28,19 +38,35 @@ interface EducationStep {
   image: ImageSourcePropType | null
   topic: EducationTopic
   title: string
-  text: string
+  // If set to true, title is displayed at the top
+  isTopTitle?: boolean
```
jmrossy, line 42: `useTopTitle`? or `showTopTitle`?

packages/mobile/src/i18n.ts
```diff
@@ -95,11 +97,14 @@ i18n
   })
   .catch((reason: any) => Logger.error(TAG, 'Failed init i18n', reason))
 
-RNLocalize.addEventListener('change', () => {
-  i18n
-    .changeLanguage(getLanguage())
-    .catch((reason: any) => Logger.error(TAG, 'Failed to change i18n language', reason))
-})
+// Disabling this for now as we have our own language selection within the app
```
jmrossy, line 100: Can we remove it entirely? I'm unclear on when we'd add this back

jmrossy (approved):
Looks great!

packages/mobile/scripts/sync_branding.sh
```diff
@@ -21,7 +21,7 @@ echo $mobile_root
 cd "$mobile_root"
 
 # Please update the sha when valora branding updates are needed
-valora_branding_sha=d5249be
+valora_branding_sha=8a74a2c
```
tarikbellamine, line 24: For future reference, how do we get this hash?

packages/mobile/src/account/Education.tsx
tarikbellamine: Yeah, we could just set a variable like `analyticsEvent` within the if statements then just call `ValoraAnalytics` once at the end. Don't think that would be an issue

packages/mobile/src/language/Language.tsx
```diff
@@ -32,18 +32,14 @@ function keyExtractor(item: Language) {
...
-    // Wait for next frame before navigating back
+    // Wait for next frame before navigating
     // so the user can see the changed selection briefly
     requestAnimationFrame(() => {
-      if (fromSettings) {
-        navigateBack()
-      } else {
-        navigate(Screens.JoinCelo)
-      }
+      navigate(nextScreen || Screens.OnboardingEducationScreen)
```
tarikbellamine, line 42: Can you please confirm that this works as expected when accessing language through Settings? When we were previously pushing to the stack rather than popping off the stack, it was causing undesirable results (e.g., you'd need to hit back button twice or top bar navigator would become unresponsive)

packages/mobile/src/account/Education.tsx
aaronmgdr: `is*` is pretty common way to name booleans. I dont follow the suggestion?

jmrossy: Yes but given that's its a prop, it implies *this* component is a top title. But it's just a flag for where to place to title

packages/mobile/scripts/sync_branding.sh
cmcewen: I think it's just from the corresponding repo under the clabs org

packages/mobile/e2e/Transfer.spec.js
```diff
@@ -41,8 +41,16 @@ describe('Transfer Works', () => {
     await bannerDismiss(by.id('SmartTopAlertButton'))
   })
 
-  it('Language', async () => {
-    await element(by.id('ChooseLanguage/en-US')).tap()
+  // Language is auto selected if it matches one of the available locale
+  // it('Language', async () => {
```
cmcewen, line 45: When would we uncomment this?

cmcewen (approved):
I'm going to merge this for tomorrow's build, but I think things to address:
- Removing unused comments
- Ensuring that the settings behavior is okay like @tarikbellamine mentioned

PR #4345, opened 2020-07-07
Connecting to AWS EKS cluster

jasonrudianto1:
### Description

Integrated the ability to connect to an AWS EKS cluster in celotool and confirming it is able to install tiller and the metric dependencies

Also generalized some function names when being called independent of which cloud provider, and specified some function names for cloud-specific functions. 

### Tested

Was able to connect to a newly spun up AWS EKS cluster and was able to install tiller and prometheus dependencies.


packages/celotool/src/lib/aws.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Basic info for an EKS cluster
+ */
+export interface AwsClusterConfig {
+  clusterRegion: string
+  clusterName: string 
+}
+
+// switchToAwsCluster configures kubectl to connect to the EKS cluster
+export async function switchToAwsCluster(
```
tkporter, line 12: There's a lot of common code between this fn and the one for Azure, could you try to reuse as much as possible?

packages/celotool/src/lib/aws.ts
```diff
@@ -0,0 +1,62 @@
...
+  celoEnv: string,
+  clusterConfig: AwsClusterConfig,
+  checkOrPromptIfStagingOrProduction = true
+) {
+  if (checkOrPromptIfStagingOrProduction) {
+    await doCheckOrPromptIfStagingOrProduction()
+  }
+
+  // TODO Look into switching subscription between testing and production
```
tkporter, line 17: Any leads with this?

packages/celotool/src/lib/oracle.ts
```diff
@@ -49,6 +50,16 @@ const oracleContextAzureClusterConfigDynamicEnvVars: { [k in keyof AzureClusterC
   clusterName: DynamicEnvVar.ORACLE_KUBERNETES_CLUSTER_NAME,
 }
 
+/**
+ * Env vars corresponding to each value for the AwslusterConfig for a particular context
+ */
+const oracleContextAwsClusterConfigDynamicEnvVars: { [k in keyof AwsClusterConfig]: DynamicEnvVar } = {
+  clusterName: DynamicEnvVar.ORACLE_KUBERNETES_CLUSTER_NAME,
+  // Given that the context has region, it is possible to not include this environment variable
```
tkporter: I like being explicit so I agree with using this env var

packages/celotool/src/lib/prometheus.ts
```diff
@@ -115,8 +118,10 @@ async function createPrometheusGcloudServiceAccountforAKS(serviceAccountName: st
   }
 }
 
-function getServiceAccountNameforAKS(kubeClusterName: string) {
+function getServiceAccountName(clusterConfig: AzureClusterConfig | AwsClusterConfig) {
+  // Checks if resourceGroup, an Azure specific property, is present in config
+  const prefix = ('resourceGroup' in clusterConfig) ? 'aks' : 'aws'
```
tkporter: What do you think about creating a generic cluster config and having the Azure/AWS ones inherit from the generic one? Like

```
interface ClusterConfig {
  clusterName: string
  cloudProviderName: string
}

export interface AwsClusterConfig extends ClusterConfig {
  clusterRegion: string
}

export interface AzureClusterConfig extends ClusterConfig {
  // stuff
}
```

and then set `cloudProviderName` to `azure` or `aws` as necessary, and then you can just have `clusterConfig: ClusterConfig` and use `clusterConfig.cloudProviderName` instead of prefix here

We want it to be backward compatible so maybe we should try to do `aks` and `eks` or something

packages/celotool/src/lib/cloud-provider.ts
```diff
@@ -0,0 +1,44 @@
...
+import { execCmd, execCmdWithExitOnFailure } from 'src/lib/cmd-utils'
+import { installAndEnableMetricsDeps, redeployTiller } from 'src/lib/helm_deploy'
+
+export interface ClusterConfig {
+  clusterName: string
+  cloudProviderName: string
+}
+
+// Configures kubectl to connect to a cluster named clusterConfig.clusterName
+export async function isContextSet(
```
tkporter: Because this function could actually change the context and only returns false if the cluster credentials aren't present, I think this name is a little misleading

Similarly-- on AKS, the name of the kubectl context is simply the name of the cluster. Is that the same on AWS?

Edit: ah I see the `--alias` flag, just wanna confirm though

packages/celotool/src/lib/env-utils.ts
```diff
@@ -140,6 +140,7 @@ export enum DynamicEnvVar {
   ORACLE_AWS_CLUSTER_REGION = 'ORACLE_{{ oracleContext }}_AWS_KUBERNETES_CLUSTER_REGION',
   ORACLE_AZURE_SUBSCRIPTION_ID = 'ORACLE_{{ oracleContext }}_AZURE_SUBSCRIPTION_ID',
   ORACLE_AZURE_TENANT_ID = 'ORACLE_{{ oracleContext }}_AZURE_TENANT_ID',
+  ORACLE_CLOUD_PROVIDER = '', // Will be automatically set in oracle
```
tkporter: This feels a little dirty, was there an issue with the `Omit` approach?

packages/celotool/src/lib/oracle.ts
```diff
@@ -362,7 +363,10 @@ function getOracleAzureIdentityName(keyVaultName: string, address: string) {
  * @return an AzureClusterConfig for the context
  */
 export function getAzureClusterConfig(oracleContext: string): AzureClusterConfig {
-  return getOracleContextDynamicEnvVarValues(oracleContextAzureClusterConfigDynamicEnvVars, oracleContext)
+  const cloudProviderNameIsAzure: { [k in keyof Pick<AzureClusterConfig, 'cloudProviderName'>]: string } = {
+    cloudProviderName: "azure"
```
tkporter: teeny nit: single quotes are preferred

packages/celotool/src/lib/oracle.ts
```diff
@@ -371,7 +375,10 @@ export function getAzureClusterConfig(oracleContext: string): AzureClusterConfig
  * @return an AwsClusterConfig for the context
  */
 export function getAwsClusterConfig(oracleContext: string): AwsClusterConfig {
-  return getOracleContextDynamicEnvVarValues(oracleContextAwsClusterConfigDynamicEnvVars, oracleContext)
+  const cloudProviderNameIsAWS: { [k in keyof Pick<AzureClusterConfig, 'cloudProviderName'>]: string } = {
+    cloudProviderName: "aws"
```
tkporter: ditto on the nit

packages/celotool/src/lib/oracle.ts
```diff
@@ -398,7 +405,7 @@ export function getOracleContextDynamicEnvVarValues<T>(
       const dynamicEnvVarName = getDynamicEnvVarName(dynamicEnvVar, {
         oracleContext
       })
-      const defaultValue = defaultValues ? defaultValues[key] : undefined
+      const defaultValue = (defaultValues && defaultValues[key]) ? defaultValues[key] : undefined
```
tkporter: `defaultValues[key]` will be false if the value is falsy (like `false`, `0`, or `''`), instead we probably wanna do `defaultValues[key] === undefined`

packages/celotool/src/lib/oracle.ts
```diff
@@ -414,10 +421,13 @@ export function getOracleContextDynamicEnvVarValues<T>(
  * Reads the context and swithces to the appropriate Azure or AWS Cluster
  * Switches to the AKS cluster associated with the given context
```
tkporter: nit: update this comment -- I wouldn't be surprised if there were similar comments mentioning only Azure/AKS throughout here

packages/celotool/src/lib/oracle.ts
```diff
@@ -429,7 +439,7 @@ export function switchToContextCluster(celoEnv: string, oracleContext: string) {
 /**
  * Switches to the AKS cluster associated with the given context
  */
-export function switchToAzureContextCluster(celoEnv: string, oracleContext: string) {
+export async function switchToAzureContextCluster(celoEnv: string, oracleContext: string) {
```
tkporter: nit: doesn't necessarily need to be `async` because there's no `await`ing going on here

packages/celotool/src/lib/prometheus.ts
```diff
@@ -118,10 +117,10 @@ async function createPrometheusGcloudServiceAccount(serviceAccountName: string)
   }
 }
 
-function getServiceAccountName(clusterConfig: AzureClusterConfig | AwsClusterConfig) {
+function getServiceAccountName(clusterConfig: ClusterConfig) {
   // Checks if resourceGroup, an Azure specific property, is present in config
-  const prefix = ('resourceGroup' in clusterConfig) ? 'aks' : 'aws'
+  const prefix = (clusterConfig.cloudProviderName === "azure") ? 'aks' : 'aws'
```
tkporter: single quotes preferred also

packages/celotool/src/lib/env-utils.ts
jasonrudianto1: Was having troubles with it earlier, but just figured it out! Next commit will show the new implementation.

packages/celotool/src/lib/cloud-provider.ts
jasonrudianto1: Yup, the --alias flag will set the context name to be the cluster name. Will also rename this function to be less misleading

nambrot:
Closing for now, feel free to re-open if still applicable

PR #4346, opened 2020-07-07
Adds link to reserve from celo home page

aaronmgdr:
### Description

Now that celoreserve.org is out, link to it when introducing CELO

### Tested
Screen Shot 2020-07-07 at 3 00 52 PM


jmrossy (approved):
🚢

PR #4348, opened 2020-07-07
[Wallet] Verification Balance Check Bug

tarikbellamine:
### Description
Moved balance check from the frontend screen to the backend verification saga. This gives us the ability to be more granular with the balance we require users to have before starting the flow, given verification cost is dependent on if they've already made some progress on verification in a previous attempt.

### Other changes
Created different modal text to show when balance is insufficient.

### Tested
Yes

### Related issues
- Part of #3988 

### Backwards compatibility
Yes


packages/mobile/src/identity/verification.ts
```diff
@@ -150,6 +152,14 @@ export function* doVerificationFlow() {
     })
 
     if (!status.isVerified) {
+      const balanceIsSufficient = yield call(
+        balanceSufficientForAttestations,
+        status.numAttestationsRemaining
+      )
+      if (!balanceIsSufficient) {
+        throw Error(ErrorMessages.INSUFFICIENT_BALANCE)
```
jmrossy:
```suggestion
        throw new Error(ErrorMessages.INSUFFICIENT_BALANCE)
```

packages/mobile/src/identity/verification.ts
```diff
@@ -211,13 +221,22 @@ export function* doVerificationFlow() {
     return true
   } catch (error) {
     Logger.error(TAG, 'Error occured during verification flow', error)
+    error.message === ErrorMessages.INSUFFICIENT_BALANCE
+      ? yield put(setVerificationStatus(VerificationStatus.InsufficientBalance))
+      : yield put(setVerificationStatus(VerificationStatus.Failed))
+
+    // Note: Do we need this given we have a VerificationFailureModal?
```
jmrossy: Maybe not, but don't see the harm

packages/mobile/src/identity/verification.ts
```diff
@@ -211,13 +221,22 @@ export function* doVerificationFlow() {
...
     yield put(showErrorOrFallback(error, ErrorMessages.VERIFICATION_FAILURE))
-    yield put(setVerificationStatus(VerificationStatus.Failed))
     return error.message
   }
 }
 
-// Requests if necessary additional attestations and returns all revealable attetations
+export function* balanceSufficientForAttestations(attestationsRemaining: number) {
+  const userBalance = yield select(stableTokenBalanceSelector)
+  return userBalance > attestationsRemaining * ESTIMATED_COST_PER_ATTESTATION
```
jmrossy: I think you're comparing a string to a number here

packages/mobile/src/identity/verification.ts
```diff
@@ -211,13 +221,21 @@ export function* doVerificationFlow() {
...
     yield put(showErrorOrFallback(error, ErrorMessages.VERIFICATION_FAILURE))
-    yield put(setVerificationStatus(VerificationStatus.Failed))
     return error.message
   }
 }
 
-// Requests if necessary additional attestations and returns all revealable attetations
+export function* balanceSufficientForAttestations(attestationsRemaining: number) {
+  const userBalance = yield select(stableTokenBalanceSelector)
+  return Number(userBalance) > attestationsRemaining * ESTIMATED_COST_PER_ATTESTATION
```
jmrossy: Please always use BigNumber when parsing numbers

packages/mobile/src/identity/verification.ts
```diff
@@ -211,13 +222,23 @@ export function* doVerificationFlow() {
     return true
   } catch (error) {
     Logger.error(TAG, 'Error occured during verification flow', error)
+    error.message === ErrorMessages.INSUFFICIENT_BALANCE
```
cmcewen: not to be too pedantic, but it feels a little weird to have this logic in the error handling - it's not really an "error" in the sense that we know exactly why the user is unable to complete verification. i wonder if just moving the put to your `if (!balanceIsSufficient)` block and returning there would be sufficient instead of throwing

packages/mobile/src/identity/verification.ts
```diff
@@ -211,13 +222,23 @@ export function* doVerificationFlow() {
...
+
     yield put(showErrorOrFallback(error, ErrorMessages.VERIFICATION_FAILURE))
-    yield put(setVerificationStatus(VerificationStatus.Failed))
     return error.message
   }
 }
 
-// Requests if necessary additional attestations and returns all revealable attetations
+export function* balanceSufficientForAttestations(attestationsRemaining: number) {
+  const userBalance = yield select(stableTokenBalanceSelector)
```
cmcewen, line 234: it doesn't matter here, but if you wanted to make this function sync so you could use it in the UI, you can always do `stableTokenBalanceSelector(store.getState())`

cmcewen (approved):
down to merge but also think the error handling could be a little cleaner

PR #4351, opened 2020-07-08
[Wallet] Allow to pay fee in Celo

i1skn:
### Description

Allow to pay fee in Celo. 
This a temporary solution for now. Later we would need show to a user which token is used to pay fee and how much would it be.

### Other changes

Fix white bottom border of `<DrawerTopBar />` when its background is not white.

### Tested

* Account has only Celo - it's possible **now** to buy cUSD
* Account has only cUSD - it's possible to buy Celo
* Account has only cUSD - it's possible to send cUSD to another account

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/transactions/send.ts
+++ b/packages/mobile/src/transactions/send.ts
@@ -5,7 +5,7 @@ import { TransactionEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import { ErrorMessages } from 'src/app/ErrorMessages'
 import { DEFAULT_FORNO_URL } from 'src/config'
-import { getCurrencyAddress } from 'src/tokens/saga'
+import { getCurrencyAddress, getTokenContract } from 'src/tokens/saga'
 import {
   sendTransactionAsync,
   SendTransactionLogEvent,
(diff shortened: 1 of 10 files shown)


jeanregisser (approved):
Awesome!! 🎉 

This has implications in the UI that we'll need to address, but it's a cool short term way to enable this!

Also if I remember correctly this makes these transactions slightly more expensive for the user (the more difference in the cUSD -> CELO price, the more expensive it is).

packages/mobile/src/transactions/send.ts
```diff
@@ -92,8 +92,11 @@ export function* sendTransactionPromises(
   staticGas?: number
 ) {
   Logger.debug(`${TAG}@sendTransactionPromises`, `Going to send a transaction with id ${txId}`)
-  // Use stabletoken to pay for gas by default
+
   const stableTokenAddress: string = yield call(getCurrencyAddress, CURRENCY_ENUM.DOLLAR)
```
jmrossy: Nit you don't need to do this line now if balance is 0

packages/mobile/src/transactions/send.ts
```diff
@@ -115,7 +118,10 @@ export function* sendTransactionPromises(
     sendTransactionAsync,
     tx,
     account,
-    stableTokenAddress,
+    // Use stableToken to pay fee, unless its balance is Zero
+    // then use Celo (goldToken) to pay fee (pass undefined)
+    // TODO: make it transparent for a user
+    stableTokenBalance.isGreaterThan(0) ? stableTokenAddress : undefined,
```
jmrossy: Should we check if the balance is greater than the expected amount + fee instead of just non-zero?

i1skn: well, this case right now targets only the case, when a user has just created a wallet, transferred Celo from somewhere and want to buy some Celo. Later we would want to do that, when we would also make it transparent to the user and give an ability to choose.

jmrossy: Ok

jmrossy (approved):

PR #4354, opened 2020-07-08
[PGPNP] Add second combiner region and status endpoint

aslawson:
### Description

- Add second combiner region for firebase function so that we have live two functions
- Using Azure Front Door to load balance between the two regions to increase SLA (firebase SLA is 99.5% whereas AFD 99.99%)
- Note they share the same DB still with SLA 99.95%
- Added a status function since Azure Front Door requires it for knowing whether to route traffic away from a service.  This isn't foolproof since it is a separate function but it will indicate if that region is experiencing down time.

### Other changes

NA

### Tested

Set up staging environment.

### Related issues

NA

### Backwards compatibility

Yes -- the old endpoints still remain.  Wallet will just switch to the azure front door endpoint when ready.


jmrossy (approved):
Nice and simple :)

PR #4363, opened 2020-07-10
Oracle Pod Security Policies

tkporter:
### Description

Introduces [pod security policies](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) to oracle infrastructure.

#### Pod Security Policy Overview

PSPs are policies that the k8s API server will enforce to ensure that no pods are created that do not meet the criteria outlined in the PSP. For example, a PSP can say that no pod whose containers main process is running as root should be created. Because different pods have different security requirements, PSPs are used on a per-service account level. This means that a deployment/statefulset/etc should be associated with a service account that has a role binding indicating what PSP the service account should be using.

For example, let's say we have a PSP called `psp-no-root` that will force pods to not have containers that run as root. We want pods created by the Deployment `foo-deployment` to abide by this PSP. We then create a new ServiceAccount called `foo-service-account`, a Role called `psp-foo-role` that allows an account to use the PSP `psp-no-root`, and a RoleBinding called `foo-role-binding` that assigns the Role `psp-foo-role` to the ServiceAccount `foo-service-account`. Finally, we configure the Deployment `foo-deployment` to use the ServiceAccount `foo-service-account`. Now when we deploy `foo-deployment`, pods will only be successfully created if the containers all do not run as root. If a ServiceAccount is not explicitly given a role to use a PSP, the k8s api server will complain and refuse to schedule the pods.

#### What this PR does

* Adds a new helm chart, `pod-security-policies`, that has some general PSPs in it and corresponding roles, which is installed/upgraded as a part of cluster setup
* Adds explicit ServiceAccounts associated PSPs to all oracle-related helm charts: `aad-pod-identity`, `oracle`, `celo-fullnode`, `prometheus-stackdriver` (this one is used in non-oracle circumstances as well, so it only supports PSPs optionally)

#### Some things to note

* Originally I tried making `pod-security-policies` a dependency of the charts that rely upon it using the `requirements.yaml`. This worked when installing charts, but I found that if chart A and B both have `pod-security-policies` as a dependency and one of chart A or B is removed, then all of the resources in `pod-security-policies` are removed as well. The solution here was just to make `pod-security-policies` an implied dependency by adding it to the cluster setup process
* We'll have to be super conscious when upgrading existing clusters to use this. PSP usage has to be enabled for the entire cluster and it's not enabled by default. It's recommended to upgrade all the K8s resources to use PSPs first, and then enable the enforcement of the PSPs on the cluster. I've been testing this on the test/dev cluster, and we should test first on baklava and go from there
* All service accounts in the `kube-system` namespace have privileged permissions so that no critical system pods aren't blocked from being scheduled

### Other changes

* Stopped mounting service account tokens for pods in which they are not used
* Added `installOrUpgradeGenericHelmChart` and started using that as a part of cluster setup for aad-pod-identity and pod-security-policies
* Azure CLI started rejecting commands that added roles using the principal type `MSI`, the fix was to go back to using principal type `ServicePrincipal`. No idea what changed here
* There are 3 PSPs that I created: a privileged one (can do pretty much whatever), a restricted one (best practice & largely restricted), and a restricted one that allows containers to run as root (some pods, including the full nodes, run as root and this is the short term fix-- it's easy to change the user but changing the user can result in file system permission ugliness)

### Tested

Deployed to the test/dev cluster, enabled PSP enforcement, and deleted every pod in the cluster to ensure that each pod will not be blocked by a PSP when being recreated. I plan to upgrade baklava clusters once this is merged and test doing this in a zero-downtime way

### Related issues

n/a

### Backwards compatibility

Once things are redeployed, this will be backward compatible


packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -652,11 +650,25 @@ export async function upgradeGenericHelmChart(
...
+  namespace: string,
+  releaseName: string,
+  chartDir: string,
+  parameters: string[]
+) {
+  const releaseExists = await outputIncludes(
+    `helm list`,
+    releaseName,
+  )
+  const fn = releaseExists ? upgradeGenericHelmChart : installGenericHelmChart
```
jcortejoso: Is there any reason why we do not use `helm install --upgrade...`? For me it is right using `helm install` and `helm upgrade`, but I wonder if someone tried before with the other way and found some issue with that pattern.

packages/helm-charts/oracle/templates/statefulset.yaml
```diff
@@ -71,6 +74,11 @@ spec:
               key: token
               name: {{ $e }}
         {{ end }}
+        - name: CERTIFICATE_AUTHORITY
+          valueFrom:
+            secretKeyRef:
+              key: ca.crt
+              name: {{ (index .Values.kube.serviceAccountSecretNames 0) }}
```
jcortejoso, line 81: I think that the cluster CA is always the same (for any service account of the cluster) and always mounted automatically into `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` ([link](https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod))

jcortejoso (approved):
Nice PR! Great security improvement for the cluster :raised_hands:

packages/helm-charts/oracle/templates/statefulset.yaml
tkporter: I added `automountServiceAccountToken: false` to prevent the service account token from being accessible within the oracle container, but unfortunately that property is on a pod-level and not container-level

I feel like it's worth getting the cert like this so that the oracle container is more locked down, what do you think?

packages/celotool/src/lib/helm_deploy.ts
tkporter: Great idea! moved over to using it instead of `installOrUpgradeGenericHelmChart`

packages/helm-charts/oracle/templates/statefulset.yaml
jcortejoso: Oh I see. I did not realize on the ´automountServiceAccountToken´ impact :ok_hand:

nambrot:
@tkporter Is this still open?

PR #4364, opened 2020-07-10
[Wallet] Secure Send Payment Request Bugs

### LINKED ISSUE #4127: Security send flow is not trigger when user proceed for requested payment
**Frequency:** 100%
** Repro App version:** Celo integration build v1.8.2 (37)
**Pre-condition:** 
Sender should have a payment request from user A who has two accounts with the same phone number. 

**Repro Steps:** 
2) Make sure user should have Received dollars request
3) Tap on send button and observe security flow is not shown 

**Current Behavior:** Security flow is not shown when user proceed payments from received request
**Expected Behavior:** Security flow should show if user make payments from received request 
**Impact:** User will not get send secure flow for requested payment. 

**Attachment:** 
1. [SendSecurityFlowIssue.mp4](https://drive.google.com/file/d/1BmOMH--1wSmJVW-BEawpfJzvGkxy1Mzj/view?usp=sharing) (Bug observed at  06sec)

2. [TwoaccountCreatewithOneNumber](https://user-images.githubusercontent.com/55572027/84906699-e8486980-b0cf-11ea-9b95-1e517a858a66.PNG)

### LINKED ISSUE #4226: Submit button is not functioning on “Confirm Account Number” screen when send secure flow trigger again after tapping on edit button from sending screen
**Frequency:** 100%
**Repro App version:** Celo Integration Wallet build v1.8.2 (40)
**Repro on device:** iphone 11(13.3)

**Pre-condition:** 
- Make sure user B should created two account using one phone number  
- Make sure user has completed send secure flow once for this user

**Repro Steps:**
1) Launch the app and login to test account 
2)  Tap on Send button. 
3) Select user B who has two account with one number. 
4) Enter an amount and tap on Review button. 
5) tap on Edit button shown next to phone number on sending screen. 
6) Enter last 4 digit account number of User B. 
7) Tap on Submit button and observed. 

**Current Behavior:** Submit button is not functioning. 
**Expected Behavior:** Submit button should functioning. 
 **Impact:** User is not able to proceed for send secure flow again by tapping on edit button. 
**Investigation:** Submit button is function when send secure flow is trigger first time  
**Attachment:**[ EditButtonIshown&SubmitbuttonIssue.MP4](https://drive.google.com/file/d/1ws7u3V1m1diCTQWkM3wXgcKF5cwwWbSo/view?usp=sharing)(Bug observed at 43 sec)

tarikbellamine:
### Description
Fixed two bugs:
- Secure Send would inconsistently trigger for payment requests
- Validating a user's address for a payment request would route to Send Confirmation instead of Payment Confirmation

Solution is to have a short waiting period represented by an inline spinner on the Send button when "send" is pressed from the payment request notification. Once address mappings are fetched, the user is either routed to the Send Confirmation or the Secure Send flow.

### Other changes
Minor refactoring and formatting cleanup

### Tested
Wrote new tests

### Related issues
- Closes #4127

### Backwards compatibility
Yes


packages/mobile/src/identity/actions.ts
```diff
@@ -31,6 +31,7 @@ export enum Actions {
   VALIDATE_RECIPIENT_ADDRESS = 'SEND/VALIDATE_RECIPIENT_ADDRESS',
   VALIDATE_RECIPIENT_ADDRESS_SUCCESS = 'SEND/VALIDATE_RECIPIENT_ADDRESS_SUCCESS',
   REQUIRE_SECURE_SEND = 'SEND/REQUIRE_SECURE_SEND',
+  DONE_FETCHING_ADDRESSES = 'DONE_FETCHING_ADDRESSES',
```
jmrossy: An `END_FETCHING_ADDRESSES` or a `SET_FETCH_ADDRESS_STATUS` would be more consistent with the other status-tracking actions we use here and elsewhere

jmrossy: And it would be nice to get success/fail status included in it so the UI can take that into account

packages/mobile/src/navigator/types.tsx
```diff
@@ -142,6 +142,9 @@ export type StackParamList = {
     reclaimPaymentInput: EscrowedPayment
   }
   [Screens.RegulatoryTerms]: undefined
+  [Screens.AddressFetchLoading]: {
```
jmrossy: My instinct would have been to make this a dialog / modal that can pop up anywhere (like a hidden one on the root App component on something) instead of a screen that we briefly navigate to and then away from. Can you clarify why a screen was used?

packages/mobile/src/qrcode/utils.ts
```diff
@@ -141,9 +142,17 @@ export function* handleBarcode(
...
+      addressJustValidated: true,
+    })
+  } else if (secureSendTxData) {
+    replace(Screens.SendConfirmation, {
+      transactionData: secureSendTxData,
+      addressJustValidated: true,
+    })
   } else {
-    replace(Screens.SendAmount, { recipient })
+    replace(Screens.SendAmount, { recipient, isFromScan: true })
```
jmrossy, line 156: Just wondering if the other two `replace` calls need this isFromScan prop too?

packages/mobile/src/send/AddressFetchLoading.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    )
+
+    jest.runAllTimers()
+
+    expect(replace).toHaveBeenCalledWith(Screens.ValidateRecipientIntro, {
+      transactionData: mockTransactionData,
+      addressValidationType: AddressValidationType.PARTIAL,
+    })
+  })
+})
```
jmrossy: Excellent test coverage here

packages/mobile/test/schemas.ts
```diff
@@ -342,6 +342,14 @@ export const v2Schema = {
...
 }
 
+export const v3Schema = {
+  ...v2Schema,
+  identity: {
+    ...v2Schema.identity,
+    isFetchingAddresses: false,
+  },
+}
+
```
jmrossy: You don't need a new schema version for just adding props.

packages/mobile/src/identity/reducer.ts
```diff
@@ -214,6 +216,16 @@ export const reducer = (
...
       }
+    case Actions.FETCH_ADDRESSES_AND_VALIDATION_STATUS:
+      return {
+        ...state,
+        isFetchingAddresses: true,
+      }
+    case Actions.DONE_FETCHING_ADDRESSES:
+      return {
+        ...state,
+        isFetchingAddresses: false,
```
jmrossy, line 228: Reset this to false on REHYDRATE or it could get suck there

jmrossy (commented):
This is overall a nice improvement. Confirming the address as its selected is a more intuitive validation approach

packages/mobile/src/identity/actions.ts
tarikbellamine: Yeah, I think I'll change the naming to `END_FETCHING_ADDRESSES`. My intention was for this to purely signal the end of a fetch (whether successful or not), not the status.

Any failure in the flow is currently being caught and shown to the user as an `ADDRESS_LOOKUP_FAILURE`. Do you think there is some other success/fail status that may be useful?

packages/mobile/src/navigator/types.tsx
tarikbellamine: To be honest, I didn't think of that but I think either would work. @nityas @coreycelo do you have a preference for a loading screen or loading modal in this case?

Context is that this will happen every time a user presses "send" on a payment request notification and should last ~1 second.

packages/mobile/src/qrcode/utils.ts
tarikbellamine: Not needed. `isFromScan` is purely to indicate whether or not the transaction was initiated via QR scan for analytics logging. A transaction initiated by a QR scan can never go through secure send (not to be confused with validating an address during secure send via scan) because QR codes are assumed to always be "correct"

packages/mobile/test/schemas.ts
tarikbellamine: Gotcha, will add that prop to v2Schema

packages/mobile/src/navigator/types.tsx
coreycelo: In order of preference:
1. Some form of "inline" spinner on the current screen—in or near the initiating button
2. Fullscreen spinner
3. Modal spinner

The pref for 2 I can see depending on appearance/performance. Ideally, "navigating" to a white screen with a spinner after the Review screen (which itself is on a white bg) won't feel like navigating at all—can we use either no transition or a fade/dissolve? It could feel simply like the Review screen is processing.

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.test.tsx
```diff
@@ -10,30 +10,32 @@ import { Screens } from 'src/navigator/Screens'
...
-const mockTransactionData2 = {
-  ...mockTransactionData,
-  firebasePendingRequestUid: 1,
+const props = {
+  id: '1',
+  amount: '24',
+  comment: 'Hey thanks for the loan, Ill pay you back ASAP. LOVE YOU',
+  requester: {
+    kind: RecipientKind.MobileNumber,
+    e164PhoneNumber: '+15126608970',
```
aaronmgdr: lets change this to 555 555 0000

packages/mobile/src/navigator/types.tsx
tarikbellamine: Changed to inline spinner

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -1,115 +1,148 @@
...
 
-  render() {
-    const { requester, id, comment, t } = this.props
-    const name = requester.displayName
-    const amount = {
-      value: this.props.amount,
-      currencyCode: CURRENCIES[CURRENCY_ENUM.DOLLAR].code,
+  React.useEffect(() => {
+    const prevIsFetchingAddresses = prevIsFetchingAddressesRef.current
+    prevIsFetchingAddressesRef.current = isFetchingAddresses
```
aaronmgdr, line 94: i might be wrong but will want to put extra attention here due to JS using references not values.

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -1,115 +1,148 @@
...
+
+    if (prevIsFetchingAddresses === true && isFetchingAddresses === false) {
+      setIsLoading(false)
+      if (!error) {
```
aaronmgdr, line 98: to be safe i think this should be added to the dependencies array

tarikbellamine: you're right, good catch!

packages/mobile/src/identity/actions.ts
jmrossy: Possibly, but if you don't have an immediate need for it we probably don't need to block this PR for it :)

jmrossy (approved):

PR #4384, opened 2020-07-11
fix language in faucet page.

aaronmgdr:
### Description

- Change Gold to CELO
- Add iOS to device types
- fix the language to match how to actually get the address. 

### Tested

Screen Shot 2020-07-10 at 5 42 24 PM


tarikbellamine (approved):
Looks good!

PR #4394, opened 2020-07-13
Don't expose debug RPC API on celo-fullnode chart

tkporter:
### Description

Even though we are not exposing the RPC port for `celo-fullnode` nodes outside of the cluster, there is no need to provide the `debug` API. This stops the `debug` API from being exposed in any RPC form on the `celo-fullnode` chart.

### Other changes

n/a

### Tested

Deployed

### Related issues

n/a

### Backwards compatibility

Yes


jasonrudianto1 (approved):
Looks good to me!

PR #4405, opened 2020-07-14
Support for interactive population of templates in RG deployment

nategraf:
### Description

I've recently been handling Baklava faucet requests and here are some changes I have made to alleviate my biggest pain points.

Most notably, this PR adds support for using a template `ReleaseGold` config without all the details (e.g. beneficiary address, or grant amount) filled in and having the script prompt for those options instead.

### Tested

I've been using this script for my `ReleaseGold` deployments for about 2 weeks now.


lucasege (approved):
This looks great, thanks for adding some type checks.
I think if we want to merge this to master we will want to test this with a ledger or otherwise be 100% sure this won't interfere with the deployment flow, since we are still using this script for important deploys.

PR #4416, opened 2020-07-14
Adding script to revoke set of contracts

lucasege:
### Description

Adds bash script that consumes a list of contracts to revoke, withdraw, and refund/destroy the contracts.

Script also checks for things like: that the contract exists, it has created an account, the contract has unlocked all of its gold, and that the contract's gold has passed the unlocking period.

Example contract file:
`contract_file.txt`:
```
0x34BCf674AD932711AE791BD17371395281B79C32
0x6bB206ffdDe4b7f7744B01B1437F57E621E8426c
```

### Tested

Tested on local testnets and ganache

### DIFF
--- a/packages/protocol/scripts/bash/revoke_contracts.sh
+++ b/packages/protocol/scripts/bash/revoke_contracts.sh
@@ -43,16 +43,16 @@ for contract in $(<"$1"); do
   fi
 
   contractExists=1
-  celocli releasegold:revoke --contract "$contract" --yesreally > /dev/null || contractExists=0
+  celocli releasegold:revoke --contract "$contract" --yesreally --useLedger > /dev/null || contractExists=0
   if [ "$contractExists" -eq 0 ]; then
     echo "Contract doesn't exist, rejecting $contract"
     rejects+=( $contract )
     continue
   fi
   if [ "$accountCreated" -ne 0 ]; then
-    celocli releasegold:locked-gold --contract "$contract" --action withdraw --value "$total" --yes 
+    celocli releasegold:locked-gold --contract "$contract" --action withdraw --value "$total" --yes --useLedger > /dev/null
   fi
-  celocli releasegold:refund-and-finalize --contract "$contract" > /dev/null
+  celocli releasegold:refund-and-finalize --contract "$contract" --useLedger > /dev/null
 
 done
 
--- a/packages/protocol/scripts/bash/revoke_contracts.sh
+++ b/packages/protocol/scripts/bash/revoke_contracts.sh
@@ -1,0 +1,68 @@
+#! /usr/bin/env bash
+set -euo pipefail
+rejects=()
+
+if [ "$#" -ne 1 ]; then
+  echo "Expected usage: ./revoke_contracts.sh contracts.txt"
+  echo "contracts.txt should include contracts you'd like to revoke and refund, one address per line"
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


packages/protocol/scripts/bash/revoke_contracts.sh
```diff
@@ -0,0 +1,68 @@
...
+fi
+
+for contract in $(<"$1"); do
+  echo "Contract $contract"
+
+  # Check that account has been created
+  accountExists=1
+  celocli lockedgold:show "$contract" > /dev/null 2>&1 || accountExists=0
+  if [ $accountExists -eq 0 ]; then
+    echo "Account not created, rejecting $contract"
```
asaj: Do we actually need to reject here? If they didn't create an account that just means that they never locked any CELO, and so we *should* be able to skip straight to revoking, right?

packages/protocol/scripts/bash/revoke_contracts.sh
```diff
@@ -0,0 +1,68 @@
...
+    rejects+=( $contract )
+    continue
+  fi
+
+  # Check that CELO has been unlocked
+  lockedgold=$(celocli lockedgold:show "$contract")
+  total=$(echo "$lockedgold" | grep -o "total: [0-9]*" | cut -f2 -d : | tr -d '[:space:]')
```
asaj: Assuming this doesn't include pending withdrawals

packages/protocol/scripts/bash/revoke_contracts.sh
```diff
@@ -0,0 +1,68 @@
...
+  if [ "$total" != "0" ]; then
+    echo "Reject $contract"
```
asaj: nit: `echo "Reject $contract: locked CELO > 0"`

packages/protocol/scripts/bash/revoke_contracts.sh
```diff
@@ -0,0 +1,68 @@
...
+  fi
+
+  contractExists=1
+  celocli releasegold:revoke --contract "$contract" --yesreally > /dev/null || contractExists=0
+  if [ "$contractExists" -eq 0 ]; then
+    echo "Contract doesn't exist, rejecting $contract"
+    rejects+=( $contract )
+    continue
+  fi
+  celocli releasegold:locked-gold --contract "$contract" --action withdraw --value "$total" --yes
```
asaj: What happens here if there are no pending withdrawals?

asaj (commented):
Just some minor comments and questions

packages/protocol/scripts/bash/revoke_contracts.sh
lucasege: It's a no-op if no pending withdrawals so this should be ok

packages/protocol/scripts/bash/revoke_contracts.sh
lucasege: Yep this is just to verify that there is no gold locked.

While the $total here is passed in to `releasegold:locked-gold.withdraw`, the `withdraw` fn actually just withdrawals all available pending withdrawals, regardless of value. FWIW, this is exactly how `lockedgold:withdraw` works, but I guess I forgot to make `value` optional for the `action.withdraw` on `releasegold:locked-gold`

This has no real impact on functionality, but is unclear, so should be more clear. I'll create a tech debt issue

marekolszewski (approved):
Approved

Issue #4406, opened 2020-07-14
[Wallet] Cursory dig into performance data 


annakaz:
### Expected Behavior

A cursory analysis of performance metrics, with some additional performance metrics instrumented. 

Note possible overlap with Light Client Epic #50: 'Analyse existing metrics on sync time, transaction performance, memory and data usage to identify key areas of improvement' (link issue once created)

### Current Behavior

Limited performance metrics instrumented, and the data that are there have not been analyzed

PR #4417, opened 2020-07-15
[Wallet] Instrument geth and contractkit initialization analytics 

annakaz:
### Description

Adding events throughout geth and contractkit so that we can use timestamps to understand how long these processes are taking, particularly on low end devices

### Tested

I have not run these changes locally, but am simply adding events.  

### Related issues

- Partially addresses #4406 

### Backwards compatibility

Yes


jmrossy (approved):
lgtm

PR #4418, opened 2020-07-15
[Wallet] Bugs when receiving payment request from unverified phone number

tarikbellamine:
### Description
Fixed two bugs related to accepting a payment request:
- If requester is unverified, the user would sometimes be forced to invite the requester
- If requester is unverified and has unrevoked verifications, user would not be allowed to send to the unverified address

### Other changes
Moved `isFetchingAddresses` into the `secureSendPhoneNumberMapping` cache so that we avoid unnecessary rerenders when there are multiple payment requests.

### Tested
Yes

### Backwards compatibility
Yes


packages/mobile/src/identity/contactMapping.ts
```diff
@@ -182,9 +186,15 @@ export function* fetchAddressesAndValidateSaga({ e164Number }: FetchAddressesAnd
 
     const userAddress = yield select(currentAccountSelector)
     const secureSendPhoneNumberMapping = yield select(secureSendPhoneNumberMappingSelector)
+    const possibleAddresses = addresses || []
```
jmrossy: Maybe we could have `getAddresses` return an empty array instead of defined another var for this here. Just because it's a bit confusing to have `oldAddresses`, `addresses`, and now `possibleAddresses` in the same function

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -273,10 +283,18 @@ const isValidNon0Address = (address: string) =>
...
-  secureSendPhoneNumberMapping: SecureSendPhoneNumberMapping
+  secureSendPhoneNumberMapping: SecureSendPhoneNumberMapping,
+  type: TokenTransactionType,
+  requesterAddress?: string
 ): string | null | undefined {
   const addresses = e164NumberToAddress[e164Number]
 
+  // If there are no verified addresses for the number and it is a payment request,
+  // use the requester's given address
+  if (!addresses && type === TokenTransactionType.PayRequest && requesterAddress) {
```
jmrossy: Can we remove the `type` param and just return requesterAddress if `!address && requesterAddress`?
My concern is the unnecessary coupling this change introduces. Currently the contact mapping code is unaware of the concept of token types and it's desirable to keep these sub-systems self-contained.

packages/mobile/src/identity/reducer.ts
```diff
@@ -34,10 +34,13 @@ export enum AddressValidationType {
...
-    address: string | undefined
-    addressValidationType: AddressValidationType
-  }
+  [e164Number: string]: SecureSendDetails
+}
+
+export interface SecureSendDetails {
+  address: string | undefined
+  addressValidationType: AddressValidationType
+  isFetchingAddresses: boolean
```
jmrossy, line 43: Moving this here seems fine, just curious was it because there are times when we're fetching two addresses at once?

packages/mobile/src/identity/secureSend.ts
```diff
@@ -41,24 +41,23 @@ function accidentallyBypassedValidation(
 
 export function checkIfValidationRequired(
   oldAddresses: string[],
-  newAddresses: string[] | null,
+  possibleAddresses: string[],
```
jmrossy, line 44: 👍

jmrossy (commented):
Mostly looks good, just a few small comments

packages/mobile/src/identity/reducer.ts
tarikbellamine: I was using `isFetchingAddressesSelector` in every payment request item so whenever an action in one item would result in it changing, it would cause a rerender of every payment request item. Realized I needed it to be tied to a specific address so that I could create a more granular selector.

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -165,7 +168,7 @@ export function* fetchAddressesAndValidateSaga({ e164Number }: FetchAddressesAnd
     // Clear existing entries for those numbers so our mapping consumers know new status is pending.
     yield put(updateE164PhoneNumberAddresses({ [e164Number]: undefined }, {}))
 
-    const addresses: string[] | null = yield call(getAddresses, e164Number)
+    const addresses: string[] = yield call(getAddresses, e164Number)
```
jmrossy, line 171: Thanks for addressing this but now you also need to adjust line 176 below 
`if (!address...`

tarikbellamine: Yikes, can't believe I missed that. Good catch...thanks!

jmrossy (approved):

PR #4429, opened 2020-07-15
[Wallet] Implement explanations of the fees and use fee drawer on exchanges

i1skn:
### Description

Implement Dialog with fee explanation when use clicks on "?" icon next to a fee.

### Other changes

* Implement `<FeeDrawer />` for Exchanges

### Tested

iOS Simulator

### Backwards compatibility

Yes

### Screenshots
Simulator Screen Shot - iPhone 11 Pro - 2020-07-15 at 16 40 57
Simulator Screen Shot - iPhone 11 Pro - 2020-07-15 at 16 41 00
Simulator Screen Shot - iPhone 11 Pro - 2020-07-15 at 17 31 00


packages/mobile/src/utils/formatting.ts
```diff
@@ -39,7 +39,8 @@ export const getExchangeRateDisplayValue = (value: BigNumber.Value): string => {
 
 export const getFeeDisplayValue = (value: BigNumber.Value | null | undefined): string => {
   return value
```
annakaz, line 41: Does value evaluate to false for any non-zero values? Wondering whether the case below is ever used for non-zero values

packages/mobile/src/transactions/__snapshots__/TransferConfirmationCard.test.tsx.snap
```diff
@@ -1672,7 +1672,7 @@ exports[`TransferConfirmationCard renders correctly for sent escrow transaction
                 >
                   
                   $
-                  0.001
+                  0
```
annakaz, line 1675: Confirming that this change is expected as we now only round up to 0.001 if it is non zero?

annakaz (commented):
Looks good overall! Comment on how very small fees are displayed

packages/mobile/src/transactions/__snapshots__/TransferConfirmationCard.test.tsx.snap
i1skn: Well, this is a tough question. When fee is precisely zero, do we want still to show `0.001`? cc: @cmcewen @jmrossy @jeanregisser

packages/mobile/src/account/InviteReview.tsx
```diff
@@ -14,7 +14,7 @@ import { connect } from 'react-redux'
 import { hideAlert, showError } from 'src/alert/actions'
 import Avatar from 'src/components/Avatar'
 import CurrencyDisplay, { DisplayType, FormatType } from 'src/components/CurrencyDisplay'
-import FeeIcon from 'src/components/FeeIcon'
+import { SecurityFeeIcon } from 'src/components/FeeIcon'
```
annakaz, line 17:
```suggestion
import { SecurityfeeIcon } from 'src/components/FeeIcon'
```

packages/mobile/src/utils/formatting.ts
i1skn: When value is a `BigNumber` it would evaluate to true independently of the value in it.

packages/mobile/src/transactions/__snapshots__/TransferConfirmationCard.test.tsx.snap
i1skn: After talking with @cmcewen I'm proposing the following: fees should not be zero in the long run, but now we have tobin tax set to zero and transaction fee is not present in the blockchain api for exchanges. @nityas also has agreed to if the value is exactly `0` - show zero. In the future these fees should be non-zero. Let me know what do you think? cc: @annakaz

annakaz: Sgtm!

annakaz (approved):
LGTM!

PR #4430, opened 2020-07-15
[celocli] Add missed-blocks command

trianglesphere:
### Description

This command lists all blocks that were missed by each validator during a range. The range is specified by looking back X blocks from the current or a specified block. This command also optionally takes a genesis block for a network to determine missed blocks when contracts are not available.

This command was built to test multi-proxy, and shows blocks missed be every validator.

### Other changes

Adds a path option in the celocli command flag parser.
Adds block number option to contract kit 'getRegisteredValidatorGroups'

### Tested

Tested locally. Work with and without the genesis block; work on main net + testnets; works on latest block & older blocks (only if the node is an archive node or hasn't pruned that history for validator group information).

### Related issues

- Fixes N/A

### Backwards compatibility

Yes


packages/cli/src/commands/validator/missed-blocks.ts
```diff
@@ -0,0 +1,149 @@
...
+        this.error('If contracts are not deployed, must provide the genesis block')
+      }
+      haveContracts = false
+    }
+
+    // Validators added from extra data in genesis
+    const missedBlocks = new Map<string, number[]>()
+
+    // Get blocks
+    const blocks = await concurrentMap(10, [...Array(res.flags.lookback).keys()], (i) =>
```
mcortesi, line 109: don't we have a `range()` function, to avoid this js magic? (probably implemented with this js magic). I would check utils for it

packages/cli/src/commands/validator/missed-blocks.ts
```diff
@@ -0,0 +1,149 @@
...
+    let validators, accounts
+    let haveContracts
+    let electedSigners: string[]
+    const validatorToGroupName = new Map<Address, string>()
+    // Try to get contracts, if not try to fall back to genesis block
+    try {
+      const election = await this.kit.contracts.getElection()
+      accounts = await this.kit.contracts.getAccounts()
+      validators = await this.kit.contracts.getValidators()
+      const vgroups = await validators.getRegisteredValidatorGroups()
```
mcortesi: No entirely sure. But i guess one can 'deregister'  groups. Meaning that you need to get the registered group "as were registered in that point in time". Specially if you go deep back in time

packages/cli/src/commands/validator/missed-blocks.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    // Get blocks
+    const blocks = await concurrentMap(10, [...Array(res.flags.lookback).keys()], (i) =>
+      this.web3.eth.getBlock(latest - res.flags.lookback! + i + 1)
+    )
+    // Calculate uptime
+    for (const block of blocks) {
+      // Grab new validators on new epoch
+      if (haveContracts) {
+        // @ts-ignore
```
mcortesi: why the ignore?

i guess because of the `haveContracts` flag. I would definitely change this and make it type safe

packages/cli/src/commands/validator/missed-blocks.ts
```diff
@@ -0,0 +1,149 @@
...
+        electedSigners = await electionCache.electedSigners(block.number)
+      }
+      for (let i = 0; i < electedSigners.length; i++) {
+        if (!missedBlocks.has(electedSigners[i])) {
+          missedBlocks.set(electedSigners[i], [])
+        }
+        const bitmap = parseBlockExtraData(block.extraData).parentAggregatedSeal.bitmap
+        if (!bitIsSet(bitmap, i)) {
+          // @ts-ignore
```
mcortesi: why the ignore?

packages/cli/src/commands/validator/missed-blocks.ts
trianglesphere: Fixed

packages/cli/src/commands/validator/missed-blocks.ts
trianglesphere: Typescript can't realize that I check if the element is present, and set a value for it if not just above. None of the common workarounds here work without making the code much more complex.

packages/cli/src/commands/validator/missed-blocks.ts
trianglesphere: Should be fixed, had to make some minor modifications to contract kit

packages/cli/src/commands/validator/missed-blocks.ts
trianglesphere: I couldn't easily find one. Also this is js magic, but also the way we do it in other CLI commands, and what appears to be the standard way to do this.

mcortesi: no worries, it was a nit anyway

PR #4431, opened 2020-07-15
[Wallet] Upgrade sentry from 1.3.7 to 1.6.2

etuleu:
### Description

Upgraded Sentry to the latest version (1.6.2). The main reason for the upgrade is to take advantage of the Android ndk crash handling which was introduced recently. I will follow up with a second PR that removes the other third party library we used on android for that purpose (jndcrash).

### Other changes

None.

### Tested

I caused crashes in Javascript as well as the native layer and made sure they end up in sentry.

### Backwards compatibility

Yes.


jmrossy (approved):
LGTM

etuleu:
Closing this one since all the changes got merged in through PR #4451

PR #4453, opened 2020-07-16
[Wallet] QR Request Bug + Android Activity Monitor Bug

### LINKED ISSUE #4059: “Send” screen is open even though scanning the QR code from Request screen
**Frequency:** 100%
** Repro App version:** Celo integration build v1.8.2 (35)
**Repro on device:** iphone 6+ (12.4.5)
**Repro Steps:** 
2) Tap on Request button. 
3) Tap on QR code icon. 
4) Tap on Scan code button. 
5) Scan the other user’s QR code. 
6) Observe 
 
**Current Behavior:**Send Screen is open instead Request screen 
**Expected Behavior:** Request screen should be shown 
**Attachment:**[RequestscreenIssue..mp4 ](https://drive.google.com/file/d/1KWKqMK8SXKsNXyxAien_Y7cvfY2cg7gC/view?usp=sharing)

tarikbellamine:
### Description
Fixed bug where clicking the QR icon from the request flow would send you into the send flow after a successful scan.

Also fixed two bugs regarding incoming payment requests:
- Android can't render non-text elements within a text element
- The lifecycle method on `IncomingPaymentRequestItem` would get inadvertently triggered when address fetches were done elsewhere in the app

### Other changes
No

### Tested
Yes

### Related issues
- Closes #4059 

### Backwards compatibility
Yes

### DIFF
--- a/packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
+++ b/packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
@@ -92,7 +92,7 @@ export default function IncomingPaymentRequestListItem({ id, amount, comment, re
   }
 
   React.useEffect(() => {
-    // Need this to make sure it's only triggered on clickyyyyy
+    // Need this to make sure it's only triggered on click
     if (!isLoading) {
       return
     }
--- a/packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
+++ b/packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
@@ -92,6 +92,11 @@ export default function IncomingPaymentRequestListItem({ id, amount, comment, re
   }
 
   React.useEffect(() => {
+    // Need this to make sure it's only triggered on clickyyyyy
+    if (!isLoading) {
+      return
+    }
+
     const prevSecureSendDetails: SecureSendDetails | undefined = prevSecureSendDetailsRef.current
     prevSecureSendDetailsRef.current = secureSendDetails
     const wasFetchingAddresses = prevSecureSendDetails?.isFetchingAddresses
--- a/packages/mobile/src/send/Send.tsx
+++ b/packages/mobile/src/send/Send.tsx
@@ -100,8 +100,17 @@ export const sendScreenNavOptions = ({
 }: {
   route: RouteProp<StackParamList, Screens.Send>
 }) => {
-  const goQr = () => navigate(Screens.QRNavigator)
-  const title = route.params?.isRequest
+  const isOutgoingPaymentRequest = route.params?.isOutgoingPaymentRequest
+
+  const gotoQRScanner = () =>
+    navigate(Screens.QRNavigator, {
+      screen: Screens.QRScanner,
+      params: {
+        isOutgoingPaymentRequest,
+      },
+    })
+
+  const title = isOutgoingPaymentRequest
     ? i18n.t('paymentRequestFlow:request')
     : i18n.t('sendFlow7:send')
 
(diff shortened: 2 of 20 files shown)


packages/react-components/components/CallToActionsBar.tsx
```diff
@@ -15,16 +15,25 @@ interface Props {
...
-        <TextButton
-          testID={`${testID}/${cta.text}/Button`}
-          key={i}
-          style={styles.action}
-          onPress={cta.onPress}
-        >
-          {cta.text}
-        </TextButton>
-      ))}
+      {callToActions.map((cta, i) => {
```
tarikbellamine, line 18: New logic because Android can't display non-text inside of text elements

packages/react-components/components/CallToActionsBar.tsx
```diff
@@ -37,5 +46,7 @@ const styles = StyleSheet.create({
     fontSize: 14,
     lineHeight: 16,
     marginRight: 24,
+    minWidth: 48,
```
tarikbellamine, line 49: Needed to fix width/height so shift from text to spinner didnt cause components to move around

packages/mobile/src/send/Send.tsx
```diff
@@ -100,8 +100,17 @@ export const sendScreenNavOptions = ({
 }: {
   route: RouteProp<StackParamList, Screens.Send>
 }) => {
-  const goQr = () => navigate(Screens.QRNavigator)
-  const title = route.params?.isRequest
+  const isOutgoingPaymentRequest = route.params?.isOutgoingPaymentRequest
+
+  const goToQRScanner = () =>
```
tarikbellamine, line 105: Navigating directly to the scanner with a param that lets us know if it's for a payment request

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -225,7 +224,11 @@ function SendAmount(props: Props) {
 
     dispatch(hideAlert())
 
-    if (addressValidationType !== AddressValidationType.NONE) {
+    if (
```
tarikbellamine, line 227: Extra logic to ensure that secure send is not triggered for QR code or direct address transactions

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -92,6 +92,11 @@ export default function IncomingPaymentRequestListItem({ id, amount, comment, re
   }
 
   React.useEffect(() => {
+    // Need this to make sure it's only triggered on click
+    if (!isLoading) {
```
tarikbellamine, line 96: `isLoading` is true whenever the "send" button is clicked. This was sometimes triggering when lookups were being done as part of other flows so adding a condition that the button must have been clicked

packages/react-components/components/CallToActionsBar.tsx
```diff
@@ -15,16 +15,25 @@ interface Props {
...
+              key={i}
+              style={styles.action}
+              onPress={cta.onPress}
+            >
+              {cta.text}
+            </TextButton>
+          )
+        }
+        return (
+          <View key={i} style={styles.action}>
```
aaronmgdr, line 32: this doesnt need to have onPress?

tarikbellamine: Nope it's a JSX element (activity spinner)

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
aaronmgdr: wouldnt you want to add this to the deps array?

cmcewen (approved):

packages/mobile/src/paymentRequest/IncomingPaymentRequestListItem.tsx
aaronmgdr: it also looks like is oading is only true when there is an e164 number.

tarikbellamine: I didn't add to the dependency array because I don't actually want that effect to run when `isLoading` changes, only when there are is an update to `secureSendDetail` or `error`. `isLoading` will be `true` whenever those variables change in the vast majority of cases so that check is just to cover edge cases

tarikbellamine: And when there isn't an `e164PhoneNumber` then the lifecycle function is irrelevant because we will immediately navigate to the next screen

PR #4459, opened 2020-07-16
Add safeguard for ECDSA-only authorization of validator signing keys

BisonDee:
### Description

The intent of the PR is to ensure that in order to authorize only an ECDSA key, without a BLS key, the caller must explicitly request this, otherwise both keys are necessary.

The current behaviors allows ECDSA key authorization _only_ with no warning which can result in a "partial key rotation".

### Tested

Tests have been revised to accomidate for the change. 

### Backwards compatibility

This will change authorization calls that intend to only authorize ECDSA keys to append a `--ecdsaOnly` flag


viktorbunin:
@timmoreton any thoughts on this?

PR #4461, opened 2020-07-16
Add wiggle room to locked gold requirements

### LINKED ISSUE #3959: Ensure rounding error does not lead to non-distribution of epoch rewards
As it stands, https://github.com/celo-org/celo-monorepo/issues/3843 can result in validator groups not distributing rewards if they originally locked up exactly the required balance. Either the rounding error should be eliminated, or if that is not possible, a mitigation of this effect should be added.

asaj:
### Description

To fix https://github.com/celo-org/celo-monorepo/issues/3959

### Other changes

None
### Tested

No
### Related issues

- Fixes #https://github.com/celo-org/celo-monorepo/issues/3959

### Backwards compatibility

Yes, should result in a patch version change.


nategraf (approved):
👍

PR #4462, opened 2020-07-16
Update developer key doc to only suggest using clabs.co domain

nategraf:
### Description

Originally the documentation on OpenPGP key publishing suggested created a key in the `celo.org` domain, but since the correct DNS records have been configured, that is no longer needed.

This PR removes the suggestion to use `celo.org` domains in the user identifier.


developer_key_publishing.md
```diff
@@ -6,7 +6,7 @@ In support of these uses, [OpenPGP](https://www.openpgp.org/) public keys can be
 
 If you want to read more about OpenPGP keys, including their structure and metadata, check out [Anatomy of a GPG Key by Dave Steele](https://davesteele.github.io/gpg/2014/09/20/anatomy-of-a-gpg-key/).
 
-> Note: This guide assumes you have an @clabs.co email address, but if you do not, simply change the email domain to your primary developer email (e.g. @example.com for alice@example.com)
+> Note: This guide assumes you have an @clabs.co email address, but if you do not, simply change the email domain to your primary developer email (e.g. @example.com for alice@example.com) Some additional setup will be required in the DNS records for your domain to configure [OpenPGP WKD]((https://gnupg.org/blog/20161027-hosting-a-web-key-directory.html)).
```
m-chrzan, line 9: Typo: period before "Some".

m-chrzan (approved):

PR #4473, opened 2020-07-16
[Wallet] Visual refinement 2

aaronmgdr:
### Description

1 set an explicit weight of normal on our regular font as without one in places where a third part component set a default weight it would not be overridden. example the drawer item component. 

2. use small not regular for the `PhoneNumbeWithEmoji` (Cory confirmed this was desired in both places it is used)

3. fix extra space at top of screens caused by safe area view. also

4. place exchange confirmation view's scroll view instead of the safe area view instead of other way to match how we have other screens set. 

5. hamburger in top bar was not centered vertically and was too far to the right. 

### Tested

Screen Shot 2020-07-16 at 1 41 45 PM
Screen Shot 2020-07-16 at 3 02 25 PM
Screen Shot 2020-07-16 at 3 02 33 PM
Screen Shot 2020-07-16 at 3 02 38 PM
Screen Shot 2020-07-16 at 3 02 46 PM
Screen Shot 2020-07-16 at 3 02 52 PM
Screen Shot 2020-07-16 at 3 51 44 PM

### Related issues

- Fixes a few bugs on the wallet design bug list


packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -185,7 +185,7 @@ export default function DrawerNavigator() {
       drawerContent={drawerContent}
       backBehavior={'initialRoute'}
       drawerContentOptions={{
-        labelStyle: [fontStyles.regular, { marginLeft: -20 }],
```
aaronmgdr, line 188: without setting weight to normal we get 500 set by react-navigation.

note the margin -20 was here before

aaronmgdr: i also considered setting theis globaly, but the snapshot diff was crazy big so i figured lets be conservative

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
```diff
@@ -101,8 +101,8 @@ export default function ExchangeConfirmationCard(props: Props) {
             <TotalLineItem amount={totalAmount} />
           </View>
         </View>
-      </SafeAreaView>
-    </ScrollView>
+      </ScrollView>
+    </SafeAreaView>
```
jeanregisser: `SafeAreaView` should generally be inside `ScrollView` so it displays nicely on devices with a notch.
Any reason not to do that here?

packages/react-components/styles/fonts.v2.tsx
```diff
@@ -79,7 +79,7 @@ const fontStyles = StyleSheet.create({
     color: colors.dark,
   },
   large: standards.large,
-  regular: standards.regular,
+  regular: { ...standards.regular },
```
jeanregisser: Why is this needed?

jeanregisser (commented):
Thanks for these!! 👍 🎉 
Left some very small questions.

packages/react-components/styles/fonts.v2.tsx
aaronmgdr: ah leftover from when i tried adding font weight normal here.

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
aaronmgdr: ah ha I got confused and thought it was the other way

aaronmgdr: Looking at the code I see 2 places where SafeAreaView is inside the scroll view and  about 10+ where it is outside.

jeanregisser: Yes when I initially added support for iOS, I put all of them outside as a shortcut.
In the long term most of them are meant to be inside, except when there's a fixed button area at the bottom of the screen.

jeanregisser (approved):
Great 👍

Issue #4401, opened 2020-07-14
[Wallet] Update onboarding changes


annakaz:
### Expected Behavior

Finish onboarding changes, order of the screens, and new screen to choose between sign up / restore

### Current Behavior

What is the current behavior?

annakaz:
cc @jeanregisser I think you have most context?

PR #4483, opened 2020-07-17
[Wallet][ReDesign] Verification progress screen

jeanregisser:
### Description

This PR implements the [verification progress screen redesign](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=6470%3A78990).

Outstanding tasks:
- [x] Decide if we should try to wait for SMS to be auto-read on Android
  - For now, we decided NOT to wait
- [x] Agree on the initial countdown duration (right now it's set to 2 mins)
  - Analytics data we have sofar seems to indicate it's an appropriate default
- [x] Decide what to do when countdown reaches 0
  - For now, we decided to switch to the verification input screen. We'll let the Verification bubble revisit this decision.

See assets diff: https://github.com/clabs-co/valora-app-branding/compare/4f414a0...a3f5867

### Other changes

- Add new `Pagination` component.
- Removed `VerificationInterstitialScreen` which is not needed anymore.
- Removed now unused dependencies (`react-native-snap-carousel`, `react-native-shadow`).

### Tested

Yes. See https://celo-org.slack.com/archives/C0134SBQTQR/p1594909944001600

### Related issues

- #4401

### Backwards compatibility

Yes


packages/mobile/src/components/Carousel.tsx
```diff
@@ -1,138 +1,104 @@
...
+  View,
+  ViewStyle,
+} from 'react-native'
 
-interface OwnProps {
-  containerStyle: ViewStyle
-  items: CarouselItem[]
-}
+const ITEM_WIDTH = variables.width - Spacing.Thick24 * 2
+const ITEM_HEIGHT = ITEM_WIDTH * (211 / 292)
```
i1skn, line 22: Is it important to keep the ratio here or we can replace it with float?

packages/mobile/src/verify/VerificationCountdown.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  const m = Math.floor(seconds / 60)
+  const s = Math.floor(seconds % 60)
+
+  return m + ':' + String(s).padStart(2, '0')
+}
+
+export default function VerificationCountdown() {
+  const endTime = useRef(Date.now() + TOTAL_TIME).current
```
i1skn, line 29: Why `useRef` here?

i1skn (commented):
Great work 💪 ! Left a couple of small comments.

packages/mobile/src/verify/VerificationCountdown.tsx
jeanregisser: This is needed for `endTime` to remain constant since the component was first mounted.

packages/mobile/src/components/Carousel.tsx
jeanregisser: The ratio could just be a float indeed, 
I took those values from the designs in Figma, so it gives slightly more context about the reason for those.
I guess I could mention this in a comment.

packages/mobile/package.json
```diff
@@ -117,11 +117,9 @@
     "react-native-screens": "^2.7.0",
     "react-native-secure-randombytes": "^3.0.2",
     "react-native-send-intent": "git+https://github.com/celo-org/react-native-send-intent#a0f4b00",
-    "react-native-shadow": "^1.2.2",
```
jmrossy, line 120: We used to use this to get more consistent shadow behavior on iOS and android. Can you confirm it's no longer needed? 
Always glad to see a dep removed, just want to confirm.

packages/mobile/src/components/Carousel.tsx
```diff
@@ -1,138 +1,104 @@
...
 }
 
-function Carousel(props: OwnProps) {
-  const ref = React.useRef(null)
-  const [activeItem, setActiveItem] = React.useState(0)
+function Carousel({ items, style }: Props) {
+  const [activeIndex, setActiveIndex] = useState(0)
+
+  const onScroll = (event: NativeSyntheticEvent<NativeScrollEvent>) => {
+    const newnActiveIndex = Math.round(event.nativeEvent.contentOffset.x / ITEM_SNAP_INTERVAL)
```
jmrossy: typo in `newn`?

packages/mobile/src/verify/VerificationCountdown.tsx
```diff
@@ -0,0 +1,97 @@
...
+    return () => clearInterval(interval)
+  }, [secondsLeft <= 0])
+
+  // Effect to call onFinish when appropriate
+  useEffect(() => {
+    if (secondsLeft <= 0 && !hasCalledOnFinishedRef.current) {
+      hasCalledOnFinishedRef.current = true
+      onFinish()
+    }
+  }, [secondsLeft <= 0, onFinish])
```
jmrossy, line 70: clever use of a predicate in the effect dep list here :)
Didn't know you could do this

packages/mobile/src/verify/VerificationCountdown.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <View style={styles.content}>
+        <Text style={styles.text}>{formatTimeRemaining(secondsLeft)}</Text>
+      </View>
+    </Animated.View>
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {},
+  image: {},
```
jmrossy: empty styles

jmrossy (approved):
Looks great! Just some minor issues found, please see comments :)

packages/mobile/package.json
jeanregisser: Yes in the few places where we use a shadow, using `elevation` was good enough on Android to mimic the more controllable shadow supported on iOS, though indeed they don't look exactly the same.

packages/mobile/src/components/Carousel.tsx
jeanregisser: Good catch 👍

Issue #4426, opened 2020-07-15
Finalize the automation of the smart contract upgrade process


nambrot:
- [x] Finalize #3893 and #4277 .
- [x] Script to verify a contracts directory has contracts that are deployed on-chain (deployed bytecode, modulo metadata, matches) (#4996)
- [x] Read version numbers from contracts (#4523)
- [x] Use the above in CI to verify changes against `master`. (#4523)
- [x] Create and document script that is easily usable externally for due diligence of governance proposals (#5038)
- [x] Script to generate governance proposal data and deployment transactions from diff of contracts' directories. (#4706)

asaj:
Thinking about sequencing here, the following things need to happen before we can kick off the first contract upgrade:

- Finalize automated tooling
    - Check build directory matches deployed contracts (@m-chrzan)
    - Check implied version deltas match actual (@asaj)
    - CI integration (???)
    - Automate governance proposal creation (???)
- Finalize desired contract changes
    - DowntimeSlasher (@gastonponti)
    - Exchange (https://github.com/celo-org/celo-proposals/pull/26) (???)
    - Version numbers (@asaj)
- Complete third party audit(s)

Anything else?

PR #4484, opened 2020-07-17
Updating the contract upgrade script

mrsmkl:
### Description

This script had gotten old and it didn't work anymore.
Now it seems to work, it deploys the changed contracts and then outputs a governance proposal to add them.

Usage:
```
yarn run truffle exec ./scripts/truffle/upgrade.js \
--network $NETWORK --build_directory $PWD/build/$NETWORK \
--contract Governance
```

### Related issues

- Related to https://github.com/celo-org/celo-monorepo/issues/4426


packages/protocol/scripts/truffle/upgrade.ts
```diff
@@ -36,14 +37,43 @@ function getProxiedContracts() {
...
+  } else {
+    names = fs
+      .readdirSync(contractsDir)
+      .filter((filename: string) => /\w+Proxy.json$/.test(filename))
+      .map(contractNameFromProxyFilename)
+  }
+
+  return names.map((contractName: string) => {
+    // tslint:disable-next-line:no-console
+    console.log('Reading artifact', contractName)
```
m-chrzan: Suggestions: "Reading artifacts for"

packages/protocol/scripts/truffle/upgrade.ts
```diff
@@ -36,14 +37,43 @@ function getProxiedContracts() {
...
       artifacts.require(contractName) as Truffle.Contract<any>,
       artifacts.require(contractName + 'Proxy') as Truffle.Contract<ProxyInstance>,
-    ])
+    ]
+  })
+}
+
+function fill(a: string) {
+  let res = '__' + a
+  while (res.length < 40) {
```
m-chrzan: Looks like there's a `.padEnd` function? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd

packages/protocol/scripts/truffle/upgrade.ts
```diff
@@ -95,7 +115,11 @@ function stripMetadata(bytecode: string) {
 async function needsUpgrade(Contract: Truffle.Contract<any>, proxy: ProxyInstance) {
   const implementationBytecode = stripMetadata(await getImplementationBytecode(proxy))
   const compiledBytecode = stripMetadata(await getCompiledBytecode(Contract))
-  return implementationBytecode !== compiledBytecode
+  const res = implementationBytecode !== compiledBytecode
+  if (!res) {
+    console.info('Not changed', Contract.contractName)
```
m-chrzan: Suggestion: "<contract> hasn't changed"

packages/protocol/scripts/truffle/upgrade.ts
```diff
@@ -161,14 +209,20 @@ module.exports = async (callback: (error?: any) => number) => {
...
       }
 
-      contractsToUpgrade.forEach(
-        async ([Contract, Proxy]: [Truffle.Contract<any>, Truffle.Contract<ProxyInstance>]) => {
-          // tslint:disable-next-line:no-console
-          console.log('Updating', Contract.contractName)
-          const proxy = await Proxy.deployed()
-          await upgradeContract(Contract, proxy)
-        }
+      const lst = await Promise.all(
```
m-chrzan: More descriptive name please. What is this a list of?

packages/protocol/scripts/truffle/upgrade.ts
```diff
@@ -161,14 +209,20 @@ module.exports = async (callback: (error?: any) => number) => {
...
+          async ([Contract, Proxy]: [Truffle.Contract<any>, Truffle.Contract<ProxyInstance>]) => {
+            // tslint:disable-next-line:no-console
+            console.log('Updating', Contract.contractName)
+            const proxy = await Proxy.deployed()
+            return upgradeContract(Contract, proxy)
+          }
+        )
       )
+      // tslint:disable-next-line:no-console
+      console.log('Generated proposal')
```
m-chrzan: What exactly is the structure of this proposal? Is it bytecode that can immediately be signed and submitted as a transaction? Does it have a human readable component?

m-chrzan (commented):
Good work overall! 

Wondering what exactly is the behavior we want from this script, and how it should interact with the backwards compatibility/versioning work. I envisioned this also generating or running transactions for deploying a fresh Proxy for when a storage upgrade is necessary (if I'm reading correctly, this already performs deployments of implementation contracts and generates a transaction for repointing a Proxy?) and in the generated governance proposal repointing the Registry.

I think one way to do this is change the assumptions of this script. Previously, this was the script that decided what needs updating. Instead, we could feed it the output of @hbandura's version checker and have it trust that rather than checking bytecodes here.

While reading this also realized we probably want to think about artifacts management. We want the old artifacts directory to be preserved in case the proposal doesn't pass, but we also want a fresh package of artifacts to be available as soon as the governance does pass.

packages/protocol/scripts/truffle/upgrade.ts
mrsmkl: It should be a json file that can be submitted using the CLI tool.

packages/protocol/scripts/truffle/upgrade.ts
mrsmkl: Thanks for looking it up!

mrsmkl:
Probably this script should be split into two:
1. one that checks the compiled bytecode against deployed
2. one that deploys the contracts and generates the proposal

About contracts with proxies, probably they have to be added here too. I don't think the usual migration scripts work for them because they have to update registry.

m-chrzan:
> Probably this script should be split into two:

I think there could even be three components:

1. Check if bytecode is same/different
2. Generate transactions (contract deployments and governance proposal), but only output them, don't actually execute
3. Take output of step 3 and deploy it to chain.

Additionally, I think it's best to organize these as components in `lib/` that are then used to put together the final scripts.

mrsmkl:
> 1. Check if bytecode is same/different
> 2. Generate transactions (contract deployments and governance proposal), but only output them, don't actually execute
> 3. Take output of step 3 and deploy it to chain.

This won't work because to generate the proposals we would need to get the addresses of deployed contracts.

My idea of the flow is this:
1. determine which contracts to upgrade (and make sure tha the upgrades are safe)
2. deploy the new contracts and generate proposal (I don't see any reason for generating the deployment transaction separately. If they are needed for some reason, they can just be fetched from the chain after they've been executed)
3. post the proposal using `celocli governance:propose`

About deploying the new proxy contracts, they'll need their own scripts that will deploy and initialize them. Then their owner should be changed to be the governance address. Finally the script should generate a proposal to add the proxy contract to registry. Basically this would be a modified version of the wrapper that's used by migrations.

One thing that would be useful is to have a modified version of script 1 which takes the proposal and checks that it matches with the compiled code.

m-chrzan:
> This won't work because to generate the proposals we would need to get the addresses of deployed contracts.

Ah, you're right, doing it at the same time is easier.

asaj:
+1 to everything @mrsmkl said. I think we could do everything we want if we have the following scripts:

1. `check-versions` which takes two branches, compiles the contracts, generates a change report, and checks that contract versions have been set correctly
2. `check-deployed`, which takes a single branch and a network name, compiles the contracts, and checks that compiled contracts match what's deployed on that network
3. `make-release`, which takes two branches, compiles the contracts, generates a change report, runs whatever contract/proxy deployments are necessary, and outputs proposal JSON to be consumed by `governance:propose`
4. `verify-release`, which takes two branches and a proposalID, compiles the contracts, generates a change report, and verifies that the proposal contents are as expected

We already have (1) out for review, thanks to @hbandura. 
I *think* @m-chrzan was going to pick up (2), so @mrsmkl maybe you could tackle (3)?

yorhodes:
closing in favor of https://github.com/celo-org/celo-monorepo/pull/5111

PR #4485, opened 2020-07-17
[Verification] Verification flow improvements

i1skn:
### Description

This PR aims to improve UX in cases of errors occur during the verification process
* Do not offer to enable Forno mode if Salt Quota is used up. Also create a separate Modal for Salt Quota Used Up error.
* Fix request validation in `attestation-service`, so we return `Attestation already sent` if SMS was already sent (general error was returned before).
* When user starts verification for the second time - app would not generate an error if SMSs for chosen issuers were already sent, but would simply navigate to the screen with codes.
* Recreates `unselectedRequest` when the previous one is expired.
* Rename "Go back" to "Enter codes" when an error occurred, but user might still have all the codes.
### Other changes
* Fix UI issues on `Reclaim Payments` until we have a new redesigned version.

### Tested

* Wallet - iOS simulator
* `attestation-service` a ran locally against the wallet.

### Backwards compatibility

Yes


packages/mobile/src/identity/verification.ts
```diff
@@ -224,8 +224,12 @@ export function* doVerificationFlow() {
     return true
   } catch (error) {
     Logger.error(TAG, 'Error occured during verification flow', error)
-    yield put(setVerificationStatus(VerificationStatus.Failed))
-    yield put(showErrorOrFallback(error, ErrorMessages.VERIFICATION_FAILURE))
+    if (error.message === ErrorMessages.SALT_QUOTA_EXCEEDED) {
```
i1skn, line 228: Probably, better to be done as a switch... But I thought it's not worth it for only two cases now.

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -143,11 +143,16 @@ class AttestationRequestHandler {
...
 
   async validateAttestationRequest() {
-    const { phoneNumber, account, issuer } = this.attestationRequest
+    const { account, issuer } = this.attestationRequest
 
-    const attestationRecord = await existingAttestationRequestRecord(phoneNumber, account, issuer, {
-      logging: this.sequelizeLogger,
-    })
+    const attestationRecord = await existingAttestationRequestRecord(
+      this.identifier,
```
jmrossy, line 149: Hmm, so before we were not checking for the right identifier in the db? Nice find, surprised we didn't catch this earlier. We'll need to coordinate an attestation service release to get the fix out.

packages/mobile/src/identity/verification.ts
jmrossy: I like `if` better personally :)

packages/mobile/src/identity/verification.ts
```diff
@@ -560,6 +564,11 @@ function* tryRevealPhoneNumber(
       return
     }
 
+    if (body.error && body.error.includes('Attestation already sent')) {
```
jmrossy, line 575: could case for optional chaining

packages/mobile/src/identity/verification.ts
```diff
@@ -560,6 +564,11 @@ function* tryRevealPhoneNumber(
...
+      Logger.debug(TAG + '@tryRevealPhoneNumber', `Ignore already sent SMS for issuer: ${issuer}`)
```
jmrossy:
```suggestion
      Logger.warn(TAG + '@tryRevealPhoneNumber', `Ignore already sent SMS for issuer: ${issuer}`)
```

packages/mobile/src/verify/VerificationFailedModal.tsx
```diff
@@ -44,56 +44,88 @@ export function VerificationFailedModal(props: Props) {
...
+  if (saltQuotaExceeded) {
+    // Show saltQuotaExceeded message and skip verification
+    return (
+      <WarningModal
+        isVisible={isVisible}
+        header={t('failModal.header')}
+        body1={t('failModal.body1SaltQuotaExceeded')}
+        body2={t('failModal.body2SaltQuotaExceeded')}
+        continueTitle={t('education.skip')}
+        onContinue={onSkip}
```
jmrossy, line 109: This is a recoverable error, we just need to send out a tx like the current Quota screen does. Is it intentional to skip here?

packages/attestation-service/src/requestHandlers/attestation.ts
tarikbellamine: So was this causing validators to send out repeat attestations?

packages/mobile/src/verify/VerificationFailedModal.tsx
tarikbellamine: I am wondering what the purpose of this modal is. My understanding is that when the quota is exceeded we automatically navigate to the quota purchase screen, which would mean this modal would only be visible if a users dismisses that screen or there's a purchase failure. Not sure there is a ton of value in a bespoke modal in that case.

Maybe we should consider eliminating the auto navigation and make it the CTA for this modal?

packages/mobile/src/verify/VerificationFailedModal.tsx
```diff
@@ -44,56 +44,88 @@ export function VerificationFailedModal(props: Props) {
...
+    return (
+      <WarningModal
+        isVisible={isVisible}
+        header={t('failModal.header')}
+        body1={t('failModal.body1')}
+        body2={t('failModal.enterCodesBody')}
+        continueTitle={t('education.skip')}
+        cancelTitle={t('global:goBack')}
+        onCancel={onDismiss}
+        onContinue={onSkip}
```
tarikbellamine, line 85: Doesn't `onSkip` cancel verification? Will this take users to the the code screen somehow?

packages/attestation-service/src/requestHandlers/attestation.ts
aslawson: nice catch. @jmrossy i'm guessing it flew under the radar because the only impact is that it wouldn't error if an sms was already sent.

nambrot: Great catch @i1skn. If I remember correctly though Validators have reported seeing the "Already sent error", but judging from the bug, i'm not entirely sure how. Were  you able to repro this?

i1skn: @nambrot yeah, this error would appear in the logs. But service was returning General Error 500 instead of 422 Already sent, hence it was not possible to figure on the client side what happened. Essentially it was skipping the validation and only trigger error on the later stage, when we are trying to send an SMS. I was able to reproduce the bug and also test the fix.

packages/mobile/src/verify/VerificationFailedModal.tsx
i1skn: @tarikbellamine you are right, this modal would only be visible if user has not funds to buy lookups or there was a failure during the purchase. But I think it makes sense to have, cause we want to explain why exactly user is not able to continue verification. Wdyt?

packages/attestation-service/src/requestHandlers/attestation.ts
nambrot: I believe Validators are showing this error message though, how would we explain that?

packages/mobile/src/verify/VerificationFailedModal.tsx
tarikbellamine: I agree there is no harm in having this in but I think it may be a little out of place from a UX perspective.

The order right now is:
1. Try to verify but exceeded lookup quota
2. Get automatically routed to lookup purchase screen
3. Either dismiss the screen because you don't know what it is or encounter an error while purchasing, which you'd see in the error banner
4. Get routed back into verification flow and see the lookup quota error modal. Only option is to exit the flow

It just feels weird to have step 4 end with a modal telling the user to purchase more lookups but we haven't given them any idea how to go do that. That's why I think switching step 2 and 4 could be good - we can use the error modal as an opportunity to inform them that they need to buy lookups to complete verification then the CTA on that model can navigate them to the lookup purchase screen, which has more information.

Understand if it's a bit out of scope for this PR but what do you think?

packages/attestation-service/src/requestHandlers/attestation.ts
jmrossy: @nambrot I'm also confused because I definitely remember testing and using the 'Enter codes anyway' feature we added to verification that relied on 422 responses from the Att service. Maybe what we remember working was *after* Amy removed phone numbers from these DB checks but *before* I added PGPNP support to Att service? That doesn't totally line up with my memory but I can't explain it otherwise.

packages/mobile/src/verify/VerificationFailedModal.tsx
i1skn: `onSkip` will move user to Home screen and `onDismiss` would dismiss the modal and they would see the Enter Codes screes (I've change "Go Back" to "Enter Codes" for dismiss action as well)

packages/mobile/src/verify/VerificationFailedModal.tsx
i1skn: I think it make sense, but I would involve @nityas and @coreycelo in this. For this PR it's not critical, because it rather aims to fix broken flows. Great thinking on this one!

i1skn:
@jmrossy @tarikbellamine I've addressed the feedback. PTAL

packages/mobile/src/verify/VerificationFailedModal.tsx
tarikbellamine: Makes sense, I misunderstood what `onDismiss` does in this case

packages/mobile/locales/en-US/nuxVerification2.json
```diff
@@ -50,6 +50,8 @@
     "body2": "You can skip verification for now and try again later.",
     "body1InsufficientBalance": "Your current balance is not high enough to complete the verification process.",
     "body2InsufficientBalance": "Please try again when you have added funds to your wallet.",
+    "body1SaltQuotaExceeded": "Your phone number lookup quota is used up",
```
tarikbellamine, line 49: Looks like you need these in spanish as well

i1skn: Added them!

packages/mobile/src/verify/VerificationFailedModal.tsx
```diff
@@ -42,56 +42,84 @@ export function VerificationFailedModal({ verificationStatus, retryWithForno, fo
...
     (verificationStatus === VerificationStatus.Failed ||
       verificationStatus === VerificationStatus.RevealAttemptFailed ||
-      userBalanceInsufficient) &&
+      userBalanceInsufficient ||
+      saltQuotaExceeded) &&
     !isDismissed
 
   const allowEnterCodes = verificationStatus === VerificationStatus.RevealAttemptFailed
   // Only prompt forno switch if not already in forno mode and failure
   // wasn't due to insuffuicient balance
```
jeanregisser: nit: we might wanna update the comment about the quota too?

packages/mobile/src/escrow/ReclaimPaymentConfirmationCard.tsx
```diff
@@ -47,7 +47,7 @@ export default function ReclaimPaymentConfirmationCard({
       <Avatar recipient={recipientContact} e164Number={recipientPhone} />
       <CurrencyDisplay type={DisplayType.Big} amount={amount} />
       <HorizontalLine />
-      <LineItemRow title={t('amount')} amount={<CurrencyDisplay amount={amount} />} />
+      <LineItemRow title={t('global:amount')} amount={<CurrencyDisplay amount={amount} />} />
```
jeanregisser, line 50: 👍

jeanregisser (commented):
Awesome! :+1:
Note for later, since this is a critical part in the wallet, we should probably cover this with more specific tests. 🤠

packages/mobile/src/identity/verification.ts
```diff
@@ -333,8 +338,20 @@ function* requestAttestations(
     phoneHash,
     account
   )
+  let isUnselectedRequestValid = unselectedRequest.blockNumber !== 0
+  if (isUnselectedRequestValid) {
+    isUnselectedRequestValid = !(yield call(
+      [attestationsWrapper, attestationsWrapper.isAttestationExpired],
+      unselectedRequest.blockNumber
+    ))
+  }
```
aslawson, line 348: nit -- this could be simplified
```
isUnselectedRequestValid = unselectedRequest.blockNumber !== 0 
|| !(yield call(
      [attestationsWrapper, attestationsWrapper.isAttestationExpired],
      unselectedRequest.blockNumber
    ))
```
 my understanding is that it shouldn't perform the second part of the OR if the first satisfies to true.

aslawson (commented):
looks good to me 👍 -- i'll let a mobile savvy person give the approval

packages/mobile/src/identity/verification.ts
i1skn: It's actually the opposite, it should check the second part if the first one is TRUE. `&&` would do, but I'm afraid it makes it a bit harder to understand from the first glance and have decided to split it into two statements. WDYT?

annakaz (approved):

jmrossy (approved):
LGTM

packages/mobile/src/identity/verification.ts
aslawson: I see -- in general this is solved by extracting the complexity to a method
```
unselectedRequest.blockNumber !== 0 && !isAttestationExpired(...)
```

no big -- since already merged

packages/attestation-service/src/requestHandlers/attestation.ts
aslawson: maybe the ones seeing this `Already sent` error are not running the version after I removed phone numbers from the DB

PR #4489, opened 2020-07-17
[Wallet] Secure Send Navigation Bug Fix

### LINKED ISSUE #4226: Submit button is not functioning on “Confirm Account Number” screen when send secure flow trigger again after tapping on edit button from sending screen

tarikbellamine:
### Description
Fixes a bug where repeatedly cycling through "edit" for secure send would cause navigation to stall. See tagged issue for context.

### Other changes
Minor refactoring to further consolidate secure send properties into the mapping object.

### Tested
Wrote new tests.

### Related issues
- Closes #4226

### Backwards compatibility
Yes


packages/mobile/src/navigator/Screens.tsx
```diff
@@ -62,7 +62,7 @@ export enum Screens {
   TransactionReview = 'TransactionReview',
   UpgradeScreen = 'UpgradeScreen',
   ValidateRecipientIntro = 'ValidateRecipientIntro',
-  ValidateRecipientAccount = 'ValidatemRecipientAccount',
+  ValidateRecipientAccount = 'ValidateRecipientAccount',
```
jeanregisser, line 63: 👍

packages/mobile/src/send/ValidateRecipientAccount.tsx
```diff
@@ -88,11 +95,22 @@ export class ValidateRecipientAccount extends React.Component<Props, State> {
...
+    const { e164PhoneNumber } = this.props.recipient
+    if (e164PhoneNumber) {
+      this.props.validateRecipientAddressReset(e164PhoneNumber)
+    }
+  }
+
+  componentDidUpdate = () => {
+    const { validationSuccessful, isOutgoingPaymentRequest, transactionData } = this.props
+
+    if (validationSuccessful) {
```
jeanregisser: We should check `validationSuccessful` actually changed, otherwise we risk triggering this path unnecessarily as the code evolves and other updates are triggered.

packages/mobile/test/schemas.ts
```diff
@@ -312,7 +312,6 @@ export const v1Schema = {
       total: 0,
     },
     matchedContacts: {},
-    isValidRecipient: false,
```
jeanregisser, line 315: Looks like we need a migration to remove `isValidRecipient` and add `validationSuccessful` to existing `secureSendPhoneNumberMapping`.

jeanregisser (changes requested):
Looking almost good, see comments below.

packages/mobile/src/send/ValidateRecipientAccount.tsx
tarikbellamine: You're right, better to be safe on this.

packages/mobile/test/schemas.ts
tarikbellamine: The issue is that we now have a `validationSuccessful` property for each phone number entry within the `secureSendPhoneNumberMapping` but we only had one universal `isValidRecipient`. In this case, what kind of migration is appropriate?

I wasn't sure so I just deleted `isValidRecipient` from the test schema and added `validationSuccessful` to the `SecureSendDetails` type. I don't think this would break any compatibility with older builds.

jmrossy (commented):
Looks okay to me, please see Jean's comments

packages/mobile/test/schemas.ts
jeanregisser: Yes this isn't going to break existing users.
If you decide not to add a migration, the type for `validationSuccessful` should be changed to `boolean | undefined` for correctness.

jeanregisser (approved):
Awesome 🎉

ValoraQA:
Hey @tarikbellamine   I have verified this issue using latest  Android Internal Build V1.11.0(1004294337) & Test Flight Build V1.11.0(47) & observed that issue is still reproduced. User not able to proceed by submitting last 4 digit codes again from Send secure flow screen.
**Verified on devices:** Samsung Galaxy A5(7.0),iPhone SE (13.4.1)
**Attachment:** [NotProceedBySubmitting4CodesAgain.mp4](https://drive.google.com/file/d/1nltd6ClMJxf6yErUi9gtenRAoELIw5L2/view?usp=sharing)

jeanregisser:
Thanks, could you open a new bug for this?

ValoraQA:
Sure @jeanregisser we will raise a new issue for this, Thanks!

ValoraQA:
Hi @jeanregisser We have raised new issue https://github.com/celo-org/wallet/issues/106.

PR #4495, opened 2020-07-20
[Wallet] Fix last Celo rate in the chart

i1skn:
### Description

Amounts to estimate Celo exchange rates should be the same in `firebase` and in wallet. This fixes the last rate in the app, which we get on-fly and not from `firebase`. Once this PR is merged and new version of `@celo/utils` is published, we should move to use value from `@celo/utils` to avoid discrepancy in the future.

Related to #4472 

### Backwards compatibility

Yes


annakaz (approved):
🚢  ✨  Thanks for fixing this!

packages/utils/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export { AccountUtils } from './account'
 export { AttestationUtils } from './attestations'
+export * from './celoHistory'
```
jmrossy: It's preferable to not export * from files to avoid polluting the namespace. I realize we already do this below but something we should avoid doing again going forward.

packages/mobile/src/exchange/saga.ts
```diff
@@ -43,8 +43,11 @@ const TAG = 'exchange/saga'
 // Amounts to estimate the exchange rate, as the rate varies based on transaction size
 // These values needs to be in-sync with the ones here:
 // https://github.com/celo-org/celo-monorepo/blob/master/packages/notification-service/src/exchange/exchangeQuery.ts#L9-L10
-const LARGE_DOLLARS_SELL_AMOUNT_IN_WEI = new BigNumber(10000 * 1000000000000000000) //  100 dollars
-const LARGE_GOLD_SELL_AMOUNT_IN_WEI = new BigNumber(10 * 1000000000000000000) // 10 gold
+// A small amount returns a rate closer to the median rate
+// TODO: Use CELO_AMOUNT_FOR_ESTIMATE and DOLLAR_AMOUNT_FOR_ESTIMATE from `@celo/utils`
+const WEI_PER_UNIT = 1000000000000000000
+const LARGE_DOLLARS_SELL_AMOUNT_IN_WEI = new BigNumber(0.01 * WEI_PER_UNIT) // 0.01 dollar
+const LARGE_GOLD_SELL_AMOUNT_IN_WEI = new BigNumber(0.01 * WEI_PER_UNIT) // 0.01 celo
```
jmrossy: The mobile project doesn't rely on a published utils version, it can use the latest source directly so may as well do that here.

jmrossy (changes requested):

i1skn:
@jmrossy great comments! I've addressed them, PTAL.

packages/utils/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export { AccountUtils } from './account'
 export { AttestationUtils } from './attestations'
+export { DOLLAR_AMOUNT_FOR_ESTIMATE, CELO_AMOUNT_FOR_ESTIMATE } from './celoHistory'
```
jmrossy: Sorry I wasn't clear, I meant putting these behind a wrapper object like for `AttestationUtils` above it. This update is equivalent to what was there before.

packages/mobile/src/exchange/saga.ts
```diff
@@ -3,6 +3,7 @@ import { ExchangeWrapper } from '@celo/contractkit/lib/wrappers/Exchange'
 import { GoldTokenWrapper } from '@celo/contractkit/lib/wrappers/GoldTokenWrapper'
 import { ReserveWrapper } from '@celo/contractkit/lib/wrappers/Reserve'
 import { StableTokenWrapper } from '@celo/contractkit/lib/wrappers/StableTokenWrapper'
+import { CELO_AMOUNT_FOR_ESTIMATE, DOLLAR_AMOUNT_FOR_ESTIMATE } from '@celo/utils'
```
jmrossy: This will work but in most cases we access the src directly. Search for '@celo/utils' in the mobile src package and you'll see what I mean

packages/utils/src/index.ts
i1skn: I've decided to remove this export line because there are no methods we expose and we are importing the constants directly from `src/celoHistory`.

jeanregisser (approved):
Great 👍

jmrossy (approved):

Issue #4415, opened 2020-07-14
[Wallet] Continue fixing bugs identified in internal launch


annakaz:
### Expected Behavior

All identified bugs blocking the Valora beta launch are resolved

### Current Behavior

The bugs in [this list](https://docs.google.com/document/d/1kwAgGsvYSGdMeMIXCHp95Dd-cYqB5KuuMTChMDzCotk/edit?usp=sharing) represent ongoing issues blocking the Valora beta launch

PR #4504, opened 2020-07-21
[Wallet] Fix DAppKit

### LINKED ISSUE #3729: Celo Wallet freezes when opening on request from dappkit
### Expected Behavior

When I make a login request from my Celo dapp to the celo wallet, I expect the Celo wallet to open and then ask if I would like to log in to the dapp from which I am making the request.

### Current Behavior

If the Celo Wallet is not already open, the wallet app tries to open, but freezes and doesn't open, so I can't log in. If the celo wallet is open in the background and I request to log in from my dapp, it works as expected. 

[Here is a screen recording](https://drive.google.com/file/d/10OWI5SjnkpXQEjBtVIeAKfeuw8CuO5tt/view?usp=sharing) of the current behavior. at the beginning of the recording, I am making the request with the wallet not running. On the second login attempt, the wallet is running in the background and it works.

jeanregisser:
### Description

This PR fixes various regressions with DAppKit support in the wallet:
- DAppKit actions not working because our HOC handling `dispatchAfterNavigate` was removed when we migrated to `react-navigation` v5.
- DAppKit not working on iOS (`Linking.canOpenURL` can't be used without having the queried scheme defined in `LSApplicationQueriesSchemes`)
- Deep links not handled when app is launched from a deep link. We'll probably need to revisit the launch flow with our sagas though.

### Other changes

- Upgraded react-navigation to fix edge case with type safe params, see react-navigation/react-navigation#8389
- Removed unnecessary state in DAppKit screens

### Tested

- Tested with [Savings Circle DApp](https://github.com/celo-org/savings-circle-demo) on iOS.

### Related issues

- Part of #4415
- Fixes #3729

### Backwards compatibility

Yes


packages/mobile/src/app/saga.ts
```diff
@@ -50,15 +51,18 @@ export function* appInit() {
...
   if (!inSync) {
     navigate(Screens.SetClock)
     return
   }
 
   if (deepLink) {
-    handleDeepLink(deepLink)
+    // TODO: this should dispatch (put) but since this appInit
+    // is called before the listener is set, we do it this way.
+    // This is fragile, change me :D
```
jmrossy, line 64: Why should this put instead of call?

packages/mobile/src/dappkit/DappKitSignTxScreen.tsx
```diff
@@ -50,22 +44,22 @@ class DappKitSignTxScreen extends React.Component<Props, State> {
...
+  }
+
   linkBack = () => {
-    if (!this.state.request) {
-      return
-    }
+    const request = this.getRequest()
 
-    navigateHome({ dispatchAfterNavigate: requestTxSignature(this.state.request!) })
+    navigateHome({ onAfterNavigate: () => this.props.requestTxSignature(request) })
```
jmrossy, line 54: Once we navigate home, this screen will be dismounted. So won't this reference to `this.props` be invalid? Or at least prevent it from getting garbage collected?

packages/mobile/src/app/saga.ts
jeanregisser: Ah yes my thought was that the `openDeepLink` action should be dispatched to follow the same path as handling deep links when the app is already open.
This is to avoid surprises and have a consistent behavior (maybe a reducer does something with the action).

packages/mobile/src/dappkit/DappKitSignTxScreen.tsx
jeanregisser: It will keep the reference from being garbage collected until `onAfterNavigate` is called by `requestAnimationFrame`. Which is fine since it's done in less than a second.

jmrossy (approved):

Issue #4359, opened 2020-07-09
[ContractKit] add removeAccount function


H34D:
### Expected Behavior

When i am able to add an account it would be good to be able to remove the account as well. This enables cycling of accounts. 

https://github.com/celo-org/celo-monorepo/blob/b4632d2eb40b5f6fb9cf35463b5c08ec338d418c/packages/contractkit/src/wallets/local-wallet.ts#L10
https://github.com/celo-org/celo-monorepo/blob/b4632d2eb40b5f6fb9cf35463b5c08ec338d418c/packages/contractkit/src/kit.ts#L193

### Current Behavior

It is not possible to remove an account once added without recreating web3 and the whole contract kit

mcortesi:
This could be added.

Something we consider when thinking about this is that on web3.js you can use `acccount indexes`. So for example, if you do `eth.accounts()` and the sign a tx and specify `from: 0` it will use account index `0`.

By doing add/removes we mess with that semantics. That's one of the reasons for ledger wallet it's a constructor parameter.

On hindsight, probably we don't need to be this cautious. Since this can already happen when you connect to a node and accounts are being added to the node.

PR #4509, opened 2020-07-22
[Wallet] Implement remove account feature

gnardini:
### Description

Added a button to allow removing account from the device. The user is asked for their Account Key before being allowed to do it.
Some screenshots of the flow:
Screen Shot 2020-07-22 at 1 25 48 AM
Screen Shot 2020-07-22 at 1 25 54 AM
Screen Shot 2020-07-22 at 1 26 20 AM
Screen Shot 2020-07-22 at 1 27 26 AM
Screen Shot 2020-07-22 at 1 27 35 AM

### Other changes

There were some dialogs that were using a `WarningModal` which had an old design. Replaced it with `Dialog` components that have the latest UI changes.

### Tested

Tested manually to remove the account and created a new one to verify that my account data was deleted.


packages/mobile/src/account/Settings.tsx
```diff
@@ -258,9 +263,36 @@ export class Account extends React.Component<Props, State> {
...
+    this.setState({ showAccountKeyModal: true })
+  }
+
+  hideRemoveAccountModal = () => {
+    this.setState({ showAccountKeyModal: false })
+  }
+
+  onPressContinueWithAccountRemoval = () => {
+    this.setState({ showAccountKeyModal: false })
+    this.props.navigation.navigate(Screens.BackupPhrase, { fromSettings: true })
```
gnardini: We're passing this `fromSettings` param through all the backup screens and the final one (`BackupComplete`) redirects to this `Settings` screen with the param `promptConfirmRemovalModal` as `true`.

Another option would be to store this on the redux store instead, but I thought it would be more error prone since we must make sure the flag is always reset if the user exits the flow at any point and it would be easy to forget/not realize we should do that when/if we make changes to this flow since it's not intuitive.

packages/mobile/src/identity/reducer.ts
```diff
@@ -236,6 +236,18 @@ export const reducer = (
           false
         ),
       }
+    case Actions.DELETE_USER_DATA:
+      return {
```
gnardini, line 260: Are these correct? I'm not sure what they all do so some of these may be unnecessary and I might be missing some other important ones, please take a close look :)

The idea here is to save stuff that is not directly linked to the account being deleted and would save time/computation/bandwidth when a new account is registered on the device.

packages/mobile/src/pincode/authentication.ts
```diff
@@ -230,3 +231,12 @@ export async function ensureCorrectPassword(
...
   }
 }
+
+export async function removeAccountLocally() {
+  clearPasswordCaches()
+  const removed = await removeStoredItem(STORAGE_KEYS.PEPPER)
+  if (!removed) {
+    // This shouldn't happen, should we do something here?
+  }
+  // TODO: Remove the account from geth.
```
gnardini: - Is there anything else we should do when deleting the account? 
- Right now we are not deleting the account from geth because there is no method in contractkit to do so. I'm not sure it's the same thing, but there's an open issue that might be related to this: https://github.com/celo-org/celo-monorepo/issues/4359

packages/mobile/src/redux/reducers.ts
```diff
@@ -40,7 +42,23 @@ export default combineReducers({
...
+
+const rootReducer = (state: RootState | undefined, action: Action): RootState => {
+  if (action.type === Actions.CLEAR_STORED_ACCOUNT && state) {
+    // Generate an initial state but keep the information not specific to the account
+    // that we want to save.
+    const initialState = appReducer(undefined, action)
+    return {
+      ...initialState,
+      localCurrency: state.localCurrency,
+      identity: identity(state.identity, deleteUserData()),
```
gnardini: Is there anything else important to keep?

gnardini:
One more question, are there any events in this process that we would want to track with `ValoraAnalytics`? Maybe the steps? Not sure if this is a flow we want to monitor.

packages/mobile/src/identity/reducer.ts
jeanregisser: Not sure why we'd want to keep these.
To me `Reset Valora` should clear everything, except selected language ;)

packages/mobile/src/pincode/authentication.ts
jeanregisser: Looks like we should also remove the `passwordHash`, see https://github.com/celo-org/celo-monorepo/blob/1ca65c95495f91138e1538618bd24865ece6737e/packages/mobile/src/pincode/authentication.ts#L105-L106

packages/mobile/src/redux/reducers.ts
jeanregisser: Like I said above, I feel we shouldn't keep anything inside `identity`.

packages/mobile/src/pincode/authentication.ts
jeanregisser: We should also reset the analytics session (flush current events and reset for a new identity).

jeanregisser (commented):
Awesome work! 👍 💯 
Left some comments.

packages/mobile/src/identity/reducer.ts
gnardini: I'm not 100% clear on this, but I understand that some of these mappings involve making requests that use (a lot of?) bandwidth and if we already did that for people on the phone's contact list it might make sense to keep them since we'll likely need to do them again anyways if the user logs in again.
@jmrossy is that correct?

packages/mobile/src/pincode/authentication.ts
gnardini: Good catch, thanks!

jeanregisser: Just thought about the anonymous firebase session we have should be closed too.

packages/mobile/src/account/Settings.tsx
```diff
@@ -258,9 +263,36 @@ export class Account extends React.Component<Props, State> {
...
+  }
+
+  hideConfirmRemovalModal = () => {
+    this.props.navigation.setParams({ promptConfirmRemovalModal: false })
+  }
+
+  confirmAccountRemoval = async () => {
+    await this.props.clearStoredAccount(this.props.account || '')
```
jmrossy: I don't think this does what you expect it do. This just synchronously dispatches the action to the store which, if you've set up the watcher - will have Redux Saga trigger the saga generator. Nothing happens if you await it.

packages/mobile/src/account/Settings.tsx
```diff
@@ -258,9 +263,36 @@ export class Account extends React.Component<Props, State> {
     ValoraAnalytics.track(SettingsEvents.tos_view)
   }
 
+  onRemoveAccountPress = () => {
+    this.setState({ showAccountKeyModal: true })
+  }
```
jmrossy, line 268: The way we have to define these handlers for each dialog makes me think we should just have the dialog manage its own visibility. Not something you have to fix, just thinking out loud.

packages/mobile/src/account/saga.ts
```diff
@@ -97,6 +99,10 @@ function* migrateAccountToProperBip39() {
   }
 }
 
+function* clearStoredAccount({ account }: ClearStoredAccountAction) {
```
jmrossy: Consider calling this `clearStoredAccountWrapper` or `clearStoredAccountSaga` to differentiate from the action and I'd recommend adding some error handling here. That's a typical pattern we have for our sagas. It's better to have them catch errors and show the error banner instead of letting those errors propagate, potentially killing the whole app.
Here's an example:
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/send/saga.ts#L171

packages/mobile/src/backup/BackupPhrase.tsx
```diff
@@ -94,12 +99,17 @@ class BackupPhrase extends React.Component<Props, State> {
 
   onPressContinue = () => {
     ValoraAnalytics.track(OnboardingEvents.backup_continue)
-    navigate(Screens.BackupQuiz)
+    navigate(Screens.BackupQuiz, { fromSettings: this.fromSettings() })
+  }
+
+  fromSettings = () => {
```
jmrossy: Would `navigatedFromSettings` be more clear?

packages/mobile/src/identity/reducer.ts
jmrossy: There's some value in saving them but the potential edge cases and buts they raise may not be worth the value in keeping them. The most valuable would be `e164NumberToSalt`. Can you start a slack thread for us to chat this.
We definitely don't need `importContactsProgress` and `askedContactsPermission` but the others _may_ be worth saving.

packages/mobile/src/identity/reducer.ts
```diff
@@ -238,6 +238,18 @@ export const reducer = (
           false
         ),
       }
+    case Actions.DELETE_USER_DATA:
```
jmrossy: Calling this action `DELETE_USER_DATA` is a bit counter-intuitive, given that it's actually *saving* things.

packages/mobile/src/pincode/authentication.ts
```diff
@@ -230,3 +231,17 @@ export async function ensureCorrectPassword(
     return false
   }
 }
+
+export async function removeAccountLocally(account: string) {
+  try {
+    ValoraAnalytics.reset().catch((error) => {
+      Logger.error(TAG, 'Error attempting to reset analytics', error, true)
+    })
```
jmrossy: This doesn't really belong here.

packages/mobile/src/pincode/authentication.ts
```diff
@@ -230,3 +231,17 @@ export async function ensureCorrectPassword(
...
+    clearPasswordCaches()
+    await removeStoredItem(STORAGE_KEYS.PEPPER)
+    await removeStoredItem(passwordHashStorageKey(account))
+    // TODO: Remove the account from geth.
```
jmrossy: I think you need to remove the mnemonic, but this doesn't feel like the right place to to it.
I'd recommend keeping the work in this file to *just* authentication related pieces and hoist other work to the saga that calls this.

packages/mobile/src/redux/reducers.ts
jmrossy: instead of a new deleteUserData action, could the identity reducer just handle the `CLEAR_STORED_ACCOUNT` differently? Not crucial, just an idea.

jmrossy: It's worth discussing, let's chat about it on slack?

packages/mobile/src/verify/VerificationFailedModal.tsx
```diff
@@ -60,66 +60,76 @@ export function VerificationFailedModal({ verificationStatus, retryWithForno, fo
   if (promptRetryWithForno) {
     // Retry verification with forno with option to skip verificaion
     return (
-      <WarningModal
+      <Dialog
```
jmrossy, line 64: Nice job standardizing on Dialog :)

packages/mobile/src/account/saga.ts
```diff
@@ -97,6 +99,10 @@ function* migrateAccountToProperBip39() {
...
+  yield call(removeAccountLocally, account)
```
jmrossy: Do we also need to restart the app to send user back to onboarding or something like that?

jmrossy (commented):
Good progress on this! Please see comments below.

packages/mobile/src/account/saga.ts
gnardini: I did the error handling that way and added a few more things, but I'm not sure about the banner thing. Most of the things we're doing here are just removing values which have their own error handling, so I think it's unlikely that we get an error here, but if we do get an error it would be either in `ValoraAnalytics` or when signing out of Firebase. When those run the account is already deleted so I think it's fine to ignore them. Right now, the `Settings` screen fires this saga and just navigates to the onboarding screen, ignoring any possible error. I'm not sure that's the best behavior, but if analytics or firebase fail the consequences don't seem very bad compared to the alternative.
What do you think?

packages/mobile/src/account/Settings.tsx
gnardini: I can see how the dialogs could handle disappearing automatically when a button is pressed, but we would still need to tell the dialog somehow when to appear, right? I can't think of a good way of doing that without a handler like this.

packages/mobile/src/identity/reducer.ts
gnardini: We're using `AccountActions.CLEAR_STORED_ACCOUNT` now, which could still be counter-intuitive but I think it's more expressive, so one could guess why we are saving data when clearing an account, what do you think?

packages/mobile/src/redux/reducers.ts
gnardini: Yes, makes sense, it was a bit lazy of me not to do that initially to be honest.

packages/mobile/src/account/saga.ts
gnardini: I am doing that on the `Settings` screen, navigating to `Screens.OnboardingEducationScreen`. The problem with that is that I'm not handling possible errors, this is related to what I wrote above.

jeanregisser (commented):
Very close now, I see one more problem with firebase and possibly other sagas spawn at startup and waiting for the `SET_ACCOUNT` action only once.

Given the way we initialize it today, it won't initialize again if we create/restore an account after having used the `Clear Account` button and without restarting the app.

See https://github.com/celo-org/celo-monorepo/blob/866e7b109df15aed84817309b42aa4e20a4883f4/packages/mobile/src/firebase/saga.ts#L290

gnardini:
@jmrossy Updated to restart the app on removal of the account. Took me a bit because it wasn't working and I didn't realize it was because I had to flush the `redux-persist` state to disk 😅

packages/mobile/locales/en-US/accountScreen10.json
```diff
@@ -24,6 +26,16 @@
     "body": "We’ve noticed you’re having some trouble connecting. We recommend enabling Data Saver mode to allow you to keep using {{appName}} with intermittent connection.",
     "switchToDataSaver": "Switch To Data Saver"
   },
+  "accountKeyModal": {
+    "header": "Without your Account Key you will lose access to your funds forever",
+    "body1": "In order to reset Valora, you will need to confirm you've written your Account Key.",
+    "body2": "Nobody, not event Valora, can recover your account without your key."
```
jmrossy: typo: not EVEN Valora

packages/mobile/src/account/actions.ts
```diff
@@ -26,6 +26,7 @@ export enum Actions {
   SET_RETRY_VERIFICATION_WITH_FORNO = 'ACCOUNT/SET_RETRY_VERIFICATION_WITH_FORNO',
   ACCEPT_TERMS = 'ACCOUNT/ACCEPT_TERMS',
   MIGRATE_ACCOUNT_BIP39 = 'MIGRATE_ACCOUNT_BIP39',
+  CLEAR_STORED_ACCOUNT = 'CLEAR_STORED_ACCOUNT',
```
jmrossy: This should be `ACCOUNT/CLEAR_STORED_ACCOUNT`
Ignore the missing `ACCOUNT/` for MIGRATE above it, that's just @cmcewen setting a bad example ;)

packages/mobile/src/account/saga.ts
```diff
@@ -97,6 +103,21 @@ function* migrateAccountToProperBip39() {
   }
 }
 
+function* clearStoredAccountSaga({ account }: ClearStoredAccountAction) {
+  try {
+    yield call(removeAccountLocally, account)
+    yield call(clearStoredMnemonic)
+    yield call(ValoraAnalytics.reset)
+    yield call(firebaseSignOut, firebase.app())
+    // TODO: Remove the account from geth.
```
jmrossy: Until we do this, what happens if a user removes the account then restores it using their backup key?

packages/mobile/src/account/saga.ts
```diff
@@ -106,7 +127,12 @@ export function* watchSetPincode() {
   yield takeLeading(Actions.SET_PINCODE, setPincode)
 }
 
+export function* watchClearStoredAccount() {
+  yield takeLeading(Actions.CLEAR_STORED_ACCOUNT, clearStoredAccountSaga)
```
jmrossy: This can be just `take` because we can't have multiple dispatched per app session right?

packages/mobile/src/account/saga.ts
```diff
@@ -97,6 +103,21 @@ function* migrateAccountToProperBip39() {
...
+    yield call(removeAccountLocally, account)
+    yield call(clearStoredMnemonic)
+    yield call(ValoraAnalytics.reset)
+    yield call(firebaseSignOut, firebase.app())
+    // TODO: Remove the account from geth.
+
+    yield call(persistor.flush)
+    yield call(restartApp)
+  } catch (error) {
+    Logger.info(TAG + '@clearStoredAccount', 'Error while removing account', error)
```
jmrossy: This must be logger.error

jmrossy: We should surface this error to the user somehow. I'd recommend using `yield put(showError(...` like we do in many sagas.

packages/mobile/src/pincode/authentication.ts
```diff
@@ -230,3 +231,9 @@ export async function ensureCorrectPassword(
     return false
   }
 }
+
+export async function removeAccountLocally(account: string) {
+  clearPasswordCaches()
+  await removeStoredItem(STORAGE_KEYS.PEPPER)
+  await removeStoredItem(passwordHashStorageKey(account))
```
jmrossy: Seems like a good candidate for `Promise.all`

packages/mobile/src/storage/keychain.tsx
```diff
@@ -40,3 +40,14 @@ export async function retrieveStoredItem(key: string) {
     throw new Error(ErrorMessages.KEYCHAIN_STORAGE_ERROR)
   }
 }
+
+export async function removeStoredItem(key: string) {
+  try {
+    return await Keychain.resetGenericPassword({
```
jmrossy: In general, I'd recommend against `return await`. You may as well just return the promise directly and have the caller await instead of making this function async

packages/mobile/src/storage/keychain.tsx
```diff
@@ -40,3 +40,14 @@ export async function retrieveStoredItem(key: string) {
...
+      service: key,
+    })
+  } catch (error) {
+    Logger.error(TAG, 'Error clearing item', error, true)
+    return false
```
jmrossy: Should we let the propagate like the other keychain utils in this file?

jmrossy (commented):
Almost there, just a few small things to still address!

packages/mobile/src/account/saga.ts
gnardini: I addressed the rest of the comments, but couldn't fix this one yet. Without removing the account from geth one cannot restore the same account in the same phone, it throws an error. To remove the account, I think we need to add a endpoint in the RPC listener on the blockchain that allows deleting the account.

packages/mobile/src/redux/reducers.ts
```diff
@@ -42,7 +43,23 @@ export default combineReducers({
...
+
+const rootReducer = (state: RootState | undefined, action: Action): RootState => {
+  if (action.type === Actions.CLEAR_STORED_ACCOUNT && state) {
+    // Generate an initial state but keep the information not specific to the account
+    // that we want to save.
+    const initialState = appReducer(undefined, action)
+    return {
+      ...initialState,
+      localCurrency: state.localCurrency,
+      identity: identity(state.identity, action),
```
annakaz, line 59: What identity gets persisted here? Would be helpful to have a comment explaining why this persists after removing the account

packages/mobile/src/backup/BackupQuiz.tsx
```diff
@@ -175,7 +175,8 @@ export class BackupQuiz extends React.Component<Props, State> {
     if (lengthsMatch && contentMatches(userChosenWords, mnemonic)) {
       Logger.debug(TAG, 'Backup quiz passed')
       this.props.setBackupCompleted()
-      navigate(Screens.BackupComplete)
+      const navigatedFromSettings = this.props.route.params?.navigatedFromSettings ?? false
```
annakaz, line 191: If navigating from the remove account flow, if a user clicks cancel, they still see the "Finish setting up your Account" modal that warns that they will be blocked from sending any transactions until they are finished setting up the account key. Instead can this cancel just cancel out of the remove account flow?

annakaz (commented):
Looks good overall! A few comments

packages/mobile/src/backup/BackupQuiz.tsx
gnardini: Yes! It also takes you to the home screen instead of the settings one, thanks, I thought about the cancel button when building it but then forgot :(

gnardini: Updated the buttons :)

annakaz (approved):
🚀

PR #4512, opened 2020-07-22
Change DEK derivation scheme and register DEK for all wallet users

### LINKED ISSUE #4464: Enable comment encryption for all wallet users
### Expected Behavior

It should be encouraged for all users to register a delegated encryption key (DEK) to be able to encrypt comments for privacy.

### Current Behavior

Currently, a DEK is created during the verification process. Wallets that choose not to verify will not register a DEK and therefore be unable to have comment encryption.

jmrossy:
### Description

- [x] Change DEK derivation from HKDF of private key to instead use Bip44 with Celo's derivation path
- [x] Add 'change index' option to CLI account creation command
- [x] Register a DEK during invite redemption and/or first outgoing payment
- [x] Rename 'comment key' references in the wallet to DEK
- [x] Add new UI and logic to register DEK during first outgoing payment
- [x] Add fallback for comment decryption to avoid showing ciphertext to users
- [x] Include a migration hack for existing Valora users
- [x] Clean up dead code around key generation
- [x] Prevent PII leak in redux saga logger

### Tested

- [x] Do full onboarding, confirm DEK registered during invite
- [x] Skip invite, confirm DEK registered during first payment
- [x] Confirm new UI only shows when sending to user with DEK registered
- [x] Test new UI on both Android and iOS
- [x] Test migration for existing Valora users

### Related issues

- Fixes #4464

### Backwards compatibility

No, comments will be different

Screen Shot 2020-07-23 at 19 16 17
Screen Shot 2020-07-23 at 19 17 02
Screen Shot 2020-07-23 at 19 26 29


packages/utils/src/dataEncryptionKey.ts
```diff
@@ -0,0 +1,53 @@
+import { ec as EC } from 'elliptic'
```
jmrossy, line 1: Mostly just moving @trianglesphere's code from commentEncryption to this new file.
A few notable changes in the `derive` function below

packages/mobile/src/web3/actions.ts
```diff
@@ -1,7 +1,8 @@
 export enum Actions {
   SET_ACCOUNT = 'WEB3/SET_ACCOUNT',
   SET_ACCOUNT_IN_WEB3_KEYSTORE = 'WEB3/SET_ACCOUNT_IN_WEB3_KEYSTORE',
-  SET_COMMENT_KEY = 'WEB3/SET_COMMENT_KEY',
+  SET_DATA_ENCRYPTION_KEY = 'WEB3/SET_DATA_ENCRYPTION_KEY',
```
annakaz, line 4: Not sure there is a better place to move this, but I find that /web3 is not where I would expect to find the data encryption key-related sagas. Realize this is nice to have in the same place as where we're storing a user's account

annakaz (commented):
Current changes LGTM!

packages/mobile/src/web3/actions.ts
jmrossy: Yeah I agree. In fact neither should be in web3.

jmrossy: I've moved most of the DEK stuff to a new web3/dataEncryptionKey.ts file which keeps it in /web3 but avoids cluttering the other web3 files. That's probably a good place to stop for this PR I think

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -118,28 +97,34 @@ export const decryptComment = memoize(_decryptComment, (...args) => values(args)
...
-    Logger.error(TAG + 'decryptComment', 'Decrypting comment failed, returning raw comment')
+  } else if (comment.length <= MAX_COMMENT_LENGTH) {
+    Logger.warn(TAG + 'decryptComment', 'Decrypting comment failed, returning raw comment')
     return { comment }
+  } else {
+    // Since we've changed the DEK derivation strategy, comment decryption would fail
+    // for old comments and/or mismatch between DEK types btwn sender + receiver
+    // To cover this case, the comment is hidden instead of showing garbage
+    Logger.warn(TAG + 'decryptComment', 'Comment appears to be ciphertext, hiding comment')
+    return { comment: i18n.t('global:commentUnavailable') }
```
jeanregisser, line 127: Just to rephrase this, once this PR lands, all past comments will be unreadable and have the placeholder shown, right?

packages/mobile/src/redux/sagas.ts
```diff
@@ -33,10 +33,6 @@ const loggerBlacklist = [
   'SEND/SET_RECIPIENT_CACHE',
   'IMPORT/IMPORT_BACKUP_PHRASE',
   'WEB3/SET_COMMENT_KEY',
-  'IDENTITY/UPDATE_E164_PHONE_NUMBER_ADDRESSES',
-  'IDENTITY/UPDATE_E164_PHONE_NUMBER_SALT',
-  'IDENTITY/FETCH_PHONE_ADDRESSES',
-  'IDENTITY/ADD_CONTACT_MATCHES',
```
jeanregisser, line 39: nit: Looks like we should use the Action types here so it doesn't become obsolete (checked by TypeScript).

packages/mobile/src/redux/sagas.ts
```diff
@@ -50,9 +46,13 @@ function* loggerSaga() {
...
   yield takeEvery('*', (action: AnyAction) => {
-    if (action && action.type && loggerBlacklist.includes(action.type)) {
+    if (
+      action?.type &&
+      (action.type.includes('IDENTITY/') || loggerBlacklist.includes(action.type))
+    ) {
       // Log only action type, but not the payload as it can have
-      // sensitive information.
+      // sensitive information. Excluding all IDENTITY/ actions because high likelyhood
+      // they contain PII and the blacklist may get out of date.
```
jeanregisser, line 56: 👍

packages/mobile/src/send/saga.ts
```diff
@@ -56,10 +57,16 @@ export async function getSendTxGas(
...
+  includeDekFee: boolean = false
 ) {
   try {
-    const gas = await getSendTxGas(account, currency, params)
+    let gas = await getSendTxGas(account, currency, params)
+    if (includeDekFee) {
+      const dekGas = await getRegisterDekTxGas(account, currency)
+      gas = gas.plus(dekGas)
+    }
+
```
jeanregisser, line 69: Wondering, how long does it take to call `getRegisterDekTxGas`?

packages/docs/command-line-interface/account.md
```diff
@@ -357,21 +357,24 @@ _See code: [packages/cli/src/commands/account/lock.ts](https://github.com/celo-o
...
 
-Creates a new account locally using the Celo Derivation Path (m/44'/52752'/0/0/indexAddress) and print out the key information. Save this information for local transaction signing or import into a Celo node. Ledger: this command has been tested swapping mnemonics with the Ledger successfully (only supports english)
+Creates a new account locally using the Celo Derivation Path (m/44'/52752'/0/changeIndex/addressIndex) and print out the key information. Save this information for local transaction signing or import into a Celo node. Ledger: this command has been tested swapping mnemonics with the Ledger successfully (only supports english)
 
 ```
 USAGE
   $ celocli account:new
 
 OPTIONS
+  --addressIndex=addressIndex
```
trianglesphere, line 367: Flagging this to make sure we update the celo holders guide as well.

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (!privateDataKey) {
+      throw new Error('No data key in store. Should never happen.')
+    }
+
+    /**
+     * BEGIN MIGRATION HACK
+     * This code can be safely removed once existing Valora users have all run it
+     * It's needed because we need to regenerate their DEKs now that the scheme has changed
+     * If it's still here by 2020/08/23 please remove it.
+     */
```
jeanregisser, line 84: 👍 🧹

packages/utils/src/dataEncryptionKey.test.ts
```diff
@@ -0,0 +1,33 @@
+import { randomBytes } from 'crypto'
+import { ec as EC } from 'elliptic'
+import { decompressPublicKey, deriveDEK } from './dataEncryptionKey'
+
+const ec = new EC('secp256k1')
+
+describe('deriveCEK', () => {
```
jeanregisser: nit: Should be `deriveDEK`

trianglesphere: Nit: `deriveCEK` -> `deriveDEK` (I left too many refs to the CEK)

packages/mobile/src/components/FeeDrawer.tsx
```diff
@@ -20,6 +20,8 @@ interface Props {
   isExchange?: boolean
   securityFee?: BigNumber
   exchangeFee?: BigNumber
+  dekFee?: BigNumber
+  showDekfee?: boolean
```
jeanregisser, line 24: nit: I'm always wondering about the appropriate wiring convention for acronyms in pascalCase, looks like this one follows what JavaScript does in general (`XmlHttpRequest`, etc).
But we also have `deriveDEK` later on.
Really not important, just dumping my thoughts.

jeanregisser (approved):
This is looking great! 💯 
To be honest, I don't have enough understanding of the whole architecture around DEK, but what I'm seeing here looks sane 👍 😅

trianglesphere (commented):
Nice work. I only have a couple small comments about docs/names, otherwise it looks good to me.

packages/mobile/src/identity/commentEncryption.ts
trianglesphere: Yes, all past encrypted comments will be unreadable. You could later recover the comments by rederiving the old DEK, but the cost of keeping the old DEK around adds up.

jeanregisser:
We need a migration for `isDEKRegistered`, or change its type to `boolean | undefined` for correctness.

packages/mobile/src/send/saga.ts
jmrossy: It seems to be quick, no noticeable change in fee calculation time. The `getRegisterDekTxGas` was modeled closely after our other gas estimation methods.

packages/docs/command-line-interface/account.md
jmrossy: There were a few refs in the CLI code/docs I missed so thanks for the callout but this doesn't appear in any docs. Are you confusing it with the `ledgerCustomAddresses` flag?

packages/mobile/src/identity/commentEncryption.ts
jmrossy: Correct. I confirmed with Nitya and Corey and this is the compromise we landed on for existing Valora users

packages/mobile/src/components/FeeDrawer.tsx
jmrossy: Yeah that's a good point, I guess I've been inconsistent! I'll rename deriveDEK to deriveDek and some others functions with the same problem

packages/mobile/src/redux/sagas.ts
jmrossy: I _think_ including a bunch of action files caused some dep cycles which caused problems back when Volpe first set this up, but I can't remember

PR #4516, opened 2020-07-22
docs: update docker imgs for 1.0.1 release

mcortesi:
Docs change for 1.0.1 release


nategraf (approved):
Find and replace for the win! I also found some other references in non-doc areas of the monorepo, and I think we should update them. You can merge https://github.com/celo-org/celo-monorepo/pull/4517 into this PR if you agree.

PR #4518, opened 2020-07-23
CIP8: Add Storage claims

nambrot:
### Description

This PR adds a CIP8 MVP proof-of-concept. It includes a simple `NameAccessor` to read and write a name to this offchain storage. You can try it out with the cli `./bin/run account:offchain-read --from 0x639324002E14818b45B7cB7414F533c865094057` against Alfajores.

The consuming API looks as following:

```typescript
const provider = new OffchainDataWrapper(res.flags.from, this.kit)
const nameApplication = new NameAccessor(provider)
const data = await nameApplication.read(res.flags.from)
```

In the background, it will:
- Look up `0x639324002E14818b45B7cB7414F533c865094057`'s metadata on-chain
- Look for storage roots in the metadata (in this case the base it is `https://raw.githubusercontent.com/nambrot/cip8test/master`)
- Download the data from the expected `dataPath` (`/account/name`) from the storage root
- Verify the signature (which was signed by a different key that is authorized under `/account/authorizedSigners/${signer}`)
- It will validate the shape of the data with the given schema at runtime and returned typed data

The PR includes the following:
- The addition of storage claims (which eventually we will want to move to `Accounts.sol`)
- A very simple `LocalStorageWriter` that just writes the data to a folder + a `GitStorageWriter` which subclasses and tries to push the change to `origin`.
- 2 cli commands which I expect to remove after the API stabilizes (`account:offchain-read` and `account:offchain-write`)
- Very basic ability to delegate the signing of payloads on offchain-storage to other keys

It does not include the following which is expected to be done in follow-on work:
- Any kind of error handling
- A story for how this data can be encrypted
- Good names for these primitives
- Proper imports via `src` instead of `lib`, as some reshuffling is still happening
- Tests to ensure regressions to not occur. Will be done once the API and structure solidifies.

I would appreciate feedback on the following:
- The general approach and API
  - I reckon that the [CIP8 spec itself](https://github.com/celo-org/celo-proposals/pull/11) was maybe not the clearest, but now that a proof-of-concept exist that makes the idea behind it probably more clear, I'd love to get feedback on it.
- The use of `io-ts`
  - Since there are is no native schema-enforcement of storage mechanism, it seemed important to me to ensure that we have runtime validation of the data. While I'm partial to `io-ts`, I do feel that I was able to hide the the complexities of said library, while getting the benefits of runtime validation. A consumer of OffchainData should virtually never encounter `io-ts` all it gets are well-defined interfaces that are also enforced at runtime.


AlexBHarley (approved):

nambrot:
Planning on merging this for now to unblock further work, but will address all comments in a follow-on PR when mentioned here

PR #4520, opened 2020-07-23
[Wallet] Fix bug because of untyped yield take action

jeanregisser:
### Description

There was a typo in `requesterAddrress` in the code handling QR codes in secure send mode.
Not sure exactly of the exact problem it would cause, I'll let @tarikbellamine comment.

Long term fix is to find a solution for fully typed `redux-saga`.
As of today, there's no clear solution to this, see more discussion https://github.com/redux-saga/redux-saga/issues/1286

### Other changes

- Added manual types for `yield take(Actions...)`.

### Tested

TypeScript check succeeds.

### Backwards compatibility

Yes.

### DIFF
--- a/packages/mobile/src/qrcode/utils.ts
+++ b/packages/mobile/src/qrcode/utils.ts
@@ -61,22 +61,22 @@ function* handleSecureSend(
 
   const userScannedAddress = data.address.toLowerCase()
   const { e164PhoneNumber } = secureSendTxData.recipient
-  const possibleRecievingAddresses = e164NumberToAddress[e164PhoneNumber]
+  const possibleReceivingAddresses = e164NumberToAddress[e164PhoneNumber]
   // This should never happen. Secure Send is triggered when there are
-  // multiple addrresses for a given phone number
-  if (!possibleRecievingAddresses) {
+  // multiple addresses for a given phone number
+  if (!possibleReceivingAddresses) {
     throw Error("No addresses associated with recipient's phone number")
   }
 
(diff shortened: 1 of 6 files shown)


packages/mobile/src/send/saga.ts
```diff
@@ -77,7 +79,7 @@ export function* watchQrCodeDetections() {
 
     if (action.scanIsForSecureSend) {
       secureSendTxData = action.transactionData
-      requesterAddress = action.requesterAddrress
+      requesterAddress = action.requesterAddress
```
jeanregisser, line 82: This is the bug I initially spotted while checking our QR code handling.

tarikbellamine: Good catch. This bug would result in secure send validation failing for a payment request from an unverified account that was being validated via QR code.

packages/mobile/src/qrcode/utils.ts
```diff
@@ -61,22 +61,22 @@ function* handleSecureSend(
 
   const userScannedAddress = data.address.toLowerCase()
   const { e164PhoneNumber } = secureSendTxData.recipient
-  const possibleRecievingAddresses = e164NumberToAddress[e164PhoneNumber]
+  const possibleReceivingAddresses = e164NumberToAddress[e164PhoneNumber]
```
tarikbellamine, line 64: Thanks!

tarikbellamine (approved):
Good catches!

PR #4523, opened 2020-07-23
Check Smart Contract version numbers in CI

### LINKED ISSUE #4424: Add Smart Contract Version numbers

asaj:
### Description

This PR adds a CI check that smart contract versions have been updated correctly with respect to the latest smart contract release.

It also updates contract versions for those that needed updating.

### Other changes

- Reverts a no-op change to `Accounts.sol` to avoid having to increment the version number, which would imply a new version should be released.
- Updates `DowntimeSlasher.sol` to use the new `ICeloVersionedContract` interface.
- Updates the current backwards compatibility checker to ignore metadata so that comment changes are not interpreted as bytecode changes which need to be released.
- Updates the current backwards compatibility checker to register visibility changes (public <-> external) as patch version changes rather than major version changes.
- Updates the current backwards compatibility checker to include an empty storage compatibility report for contracts that are added.

### Tested
Manual experimentation. You can check for compatibility against the latest release by running:
`yarn check-versions -o release/contracts/0 -n $CURRENT_BRANCH`

### Related issues

- Fixes #4424
- Part of broader work towards https://github.com/celo-org/celo-monorepo/issues/4426

### Backwards compatibility

Yes


packages/protocol/contracts/common/interfaces/ICeloVersionedContract.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity ^0.5.3;
+
+interface ICeloVersionedContract {
```
hbandura, line 3: Maybe add some docs explaining what these 4 ints mean

hbandura: (probably including an explanation for the storage number, since the other 3 follow the semver standard)

packages/protocol/lib/compatibility/ast-code.ts
```diff
@@ -348,6 +348,21 @@ const getCheckableMethodsFromAST = (contract: ContractAST, id: string): any[] =>
...
 
+/*
+ * The Solidity compiler appends a Swarm Hash of compilation metadata to the end
+ * of bytecode. We find this hash based on the specification here:
+ * https://solidity.readthedocs.io/en/develop/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode
+ */
+const stripMetadataIfPresent = (bytecode: string): string => {
+  try {
+    // TODO: use proper CBOR parser
+    const [, bytes] = bytecode.match(/^(.*)a165627a7a72305820.*0029$/i)
```
hbandura: extract the regexp as a constant so it's already precompiled instead of recompiling it on every function call

packages/protocol/lib/compatibility/ast-code.ts
```diff
@@ -422,7 +437,7 @@ function generateASTCompatibilityReport(oldContract: ZContract, oldArtifacts: Bu
 
   const report = doASTCompatibilityReport(contractName, oldAST, newAST)
   // Check deployed byte code change
-  if (oldContract.schema.deployedBytecode !== newContract.schema.deployedBytecode) {
+  if (stripMetadataIfPresent(oldContract.schema.deployedBytecode) !== stripMetadataIfPresent(newContract.schema.deployedBytecode)) {
```
hbandura, line 438: Update line number 130 on this file, since we are now stripping the metadata

hbandura: (DeployedByteCodeChange doc)

packages/protocol/lib/compatibility/ast-layout.ts
```diff
@@ -211,6 +211,9 @@ export const reportLayoutIncompatibilities = (oldArtifacts: BuildArtifacts, newA
     const oldArtifact = oldArtifacts.getArtifactByName(newArtifact.contractName)
     if (oldArtifact !== undefined) {
       return generateCompatibilityReport(oldArtifact, oldArtifacts, newArtifact, newArtifacts)
+    } else {
+      // Generate an empty report for new contracts.
+      return generateCompatibilityReport(newArtifact, newArtifacts, newArtifact, newArtifacts)
```
hbandura: We could just return a literal structure here 
```
{
  contract: ...
  compatible: true
  errors: []
}
```
(double check, I'm writing it from memory)

packages/protocol/lib/compatibility/ast-version.ts
```diff
@@ -0,0 +1,76 @@
...
+import { ContractVersion, ContractVersionDelta, ContractVersionDeltaIndex, ContractVersions } from '@celo/protocol/lib/compatibility/version'
+import {
+  BuildArtifacts,
+} from '@openzeppelin/upgrades'
+
+/**
+ * A version report for a specific contract.
+ */
+export class ASTContractVersionReport {
+  constructor(public readonly oldVersion: ContractVersion, public readonly newVersion: ContractVersion, public readonly expectedDelta: ContractVersionDelta) {}
```
hbandura: nit: limit line width size a bit, otherwise it's tough to read on some displays. I usually go for 80-100 as a personal preference.

hbandura: (ditto throughout this file)

packages/protocol/lib/compatibility/ast-version.ts
```diff
@@ -0,0 +1,76 @@
+// tslint:disable: max-classes-per-file
+import { ContractVersion, ContractVersionDelta, ContractVersionDeltaIndex, ContractVersions } from '@celo/protocol/lib/compatibility/version'
+import {
+  BuildArtifacts,
+} from '@openzeppelin/upgrades'
+
+/**
+ * A version report for a specific contract.
+ */
+export class ASTContractVersionReport {
```
hbandura: Not sure if this is really a report, but maybe a "Checker" or "Validator" ?

Also, could this type be in `version.ts` ?

packages/protocol/lib/compatibility/ast-version.ts
```diff
@@ -0,0 +1,76 @@
...
+      const versionDelta = expectedVersionDeltas[contract] === undefined ? ContractVersionDelta.fromChanges(false, false, false, false) : expectedVersionDeltas[contract]
+      const oldVersion = oldVersions.contracts[contract] === undefined ? ContractVersion.fromString('0.0.0.0') : oldVersions.contracts[contract]
+      contracts[contract] = new ASTContractVersionReport(oldVersion, newVersions.contracts[contract], versionDelta)
+    })
+    return new ASTContractVersionsReport(contracts)
+  }
+  constructor(public readonly contracts: ASTContractVersionReportIndex) {}
+
+  /**
+   * @return a new {@link ASTContractVersionsResport} with the same version
```
hbandura: typo in type name

packages/protocol/lib/compatibility/version.ts
```diff
@@ -105,3 +124,57 @@ export class ContractVersionDelta {
...
+ * A mapping {contract name => {@link ContractVersion}}.
+ */
+export interface ContractVersionIndex {
+  [contract: string]: ContractVersion;
+}
+
+/**
+ * A mapping {contract name => {@link ContractVersion}}.
+ */
+export class ContractVersions {
```
hbandura: Regarding the file usage, we could have `version.ts` with all independent-from-openzeppeling types (contract version, delta, version checker), and then have `ast-version.ts` with things that use both version and artifacts stuff. If you agree, then move this type to `ast-version.ts` so we can remove the `ast-code.ts` dependency.

packages/protocol/lib/compatibility/version.ts
```diff
@@ -26,7 +34,18 @@ export class ContractVersion {
       Number(v[3])
     )
   }
-  
+
+  static fromBuffer = (version: Buffer): ContractVersion => {
```
hbandura: Maybe add some docs explaining this. Without context I'm not sure what kind of buffer is this about.

packages/protocol/lib/compatibility/version.ts
```diff
@@ -105,3 +124,57 @@ export class ContractVersionDelta {
...
+ */
+export interface ContractVersionIndex {
+  [contract: string]: ContractVersion;
+}
+
+/**
+ * A mapping {contract name => {@link ContractVersion}}.
+ */
+export class ContractVersions {
+  static fromArtifacts = async (artifacts: BuildArtifacts): Promise<ContractVersions>=> {
```
hbandura: Add a doc to the method explaining that it will call the method `getContractVersion() (uint, uint, ... )` to retrieve the information, and what the default version will be if the method is not available

packages/protocol/lib/compatibility/version.ts
```diff
@@ -105,3 +124,57 @@ export class ContractVersionDelta {
...
+    static: true,
+    data: Buffer.from(data.slice(2), 'hex')
+  })
+  if (result.execResult.exceptionError === undefined) {
+    const value = result.execResult.returnValue
+    if (value.length === 4 * 32) {
+      return ContractVersion.fromBuffer(value)
+    }
+  }
+  // If we can't fetch the version number, assume version 1.0.0.0.
```
hbandura: I would manage the error in a different way here:

if the method is not present, if it doesn't exist, then we could assume a default value. (the only thing that worries me about that and the default value, is if someone adds the method with a typo or different signature and doesn't realize it)

if the method exists, but there is an error in calling it, then I would just throw a big exception. It's ok to break there. What do you think?

packages/protocol/lib/compatibility/ast-version.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export class ASTContractVersionsReport {
+  static create = async (oldArtifacts: BuildArtifacts, newArtifacts: BuildArtifacts, expectedVersionDeltas: ContractVersionDeltaIndex): Promise<ASTContractVersionsReport> => {
+    const oldVersions = await ContractVersions.fromArtifacts(oldArtifacts)
+    const newVersions = await ContractVersions.fromArtifacts(newArtifacts)
+    const contracts = {}
+    Object.keys(newVersions.contracts).map((contract:string) => {
+      const versionDelta = expectedVersionDeltas[contract] === undefined ? ContractVersionDelta.fromChanges(false, false, false, false) : expectedVersionDeltas[contract]
+      const oldVersion = oldVersions.contracts[contract] === undefined ? ContractVersion.fromString('0.0.0.0') : oldVersions.contracts[contract]
```
hbandura: Does this imply that when adding a new contract that has no storage, its first version number would be 0.0.1.0 ? (Nothing wrong about that, just checking)

packages/protocol/scripts/check-backward.ts
```diff
@@ -104,44 +69,39 @@ try {
...
-        `${argv.old_version} + ${backward.report.global.versionDelta} == ${argv.new_version}`,
-        true
-      )
-      out(`\n`)
-      process.exit(0)
+    const doVersionCheck = async () => {
+      const excludeRegexp: RegExp = exclude ? new RegExp(exclude) : null
+      const versionReport = (
+        await ASTContractVersionsReport.create(
+          instantiateArtifacts(oldArtifactsFolder),
```
hbandura: I'm not sure what the concurrency model is in typescript, but since you are already using promises, maybe we could concurrently run both instantiateArtifacts calls? I bet they are probably 80%[citation needed] of the run time of the script

hbandura (changes requested):
Added some comments.

Also, could you add tests to it? namely: a comparison of the removed metadata test in `test/compatibility/ast-code.ts`, and a few tests for the actual version retrieval from a compiled contract by calling the getContractVersion.

packages/protocol/lib/compatibility/ast-version.ts
asaj: Ah yeah prettier didn't catch this because it specifically excludes `lib`

packages/protocol/lib/compatibility/version.ts
asaj: The `ICeloVersionedContract` interface should protect us from this

packages/protocol/contracts/common/interfaces/ICeloVersionedContract.sol
asaj: Included the natspec used in the contracts in which the interface is implemented

packages/protocol/lib/compatibility/ast-version.ts
asaj: Wdyt of having a follow up PR that just runs `prettify`? It will have a *lot* of changes and I'd prefer to avoid polluting the diff.

packages/protocol/lib/compatibility/version.ts
asaj: Great suggestion!

packages/protocol/lib/compatibility/version.ts
asaj: Added docs to `getContractVersion`

packages/protocol/lib/compatibility/ast-version.ts
asaj: Yeah great question, I think this should actually be `1.0.0.0`, as the storage/code compatibility report should come back clean, which implies no version delta, which implies the actual version should be the same as what we have here. Clarified in a comment.

packages/protocol/scripts/check-backward.ts
asaj: It's reasonably performant already, perhaps we save for a future optimization?

packages/protocol/lib/compatibility/ast-version.ts
m-chrzan: Is it useful to have this be a class rather than a `verifyDeltaApplies(oldVersion, delta, newVersion)` (or similar) function?

packages/protocol/lib/compatibility/ast-version.ts
```diff
@@ -0,0 +1,76 @@
...
+    return new ASTContractVersionsReport(contracts)
+  }
+  constructor(public readonly contracts: ASTContractVersionReportIndex) {}
+
+  /**
+   * @return a new {@link ASTContractVersionsResport} with the same version 
+   * reports, excluding all contract names that match the {@param exclude}
+   * parameter.
+   */
+  excluding = (exclude: RegExp): ASTContractVersionsReport => {
```
m-chrzan: Is there a way to deduplicate this and similar functionality in lib/compatibility/report.ts?

packages/protocol/lib/compatibility/version.ts
```diff
@@ -26,7 +34,18 @@ export class ContractVersion {
...
+
+  static fromBuffer = (version: Buffer): ContractVersion => {
+    if (version.length !== 4 * 32) {
+      throw new Error(`Invalid version buffer: ${version}`)
+    }
+    const storage = version.slice(0, 32)
+    const major = version.slice(32, 64)
+    const minor = version.slice(64, 96)
+    const patch = version.slice(96, 128)
+    return ContractVersion.fromString(`${storage.toString('hex')}.${major.toString('hex')}.${minor.toString('hex')}.${patch.toString('hex')}`)
```
m-chrzan: Suggestion: might look cleaner if you put the `.toString`s when defining `storage`, `major`, etc. above.

packages/protocol/lib/compatibility/version.ts
```diff
@@ -105,3 +124,57 @@ export class ContractVersionDelta {
...
+
+/**
+ * A mapping {contract name => {@link ContractVersion}}.
+ */
+export class ContractVersions {
+  static fromArtifacts = async (artifacts: BuildArtifacts): Promise<ContractVersions>=> {
+    const contracts = {}
+
+    await Promise.all(artifacts.listArtifacts().map(async (artifact) => {
+      contracts[artifact.contractName] = await getContractVersion(makeZContract(artifact))
```
m-chrzan: Is an OpenZeppelin Contract object necessary here, rather than just a plain build artifact?

packages/protocol/lib/compatibility/version.ts
```diff
@@ -105,3 +124,57 @@ export class ContractVersionDelta {
...
+    data: Buffer.from(data.slice(2), 'hex')
+  })
+  if (result.execResult.exceptionError === undefined) {
+    const value = result.execResult.returnValue
+    if (value.length === 4 * 32) {
+      return ContractVersion.fromBuffer(value)
+    }
+  }
+  // If we can't fetch the version number, assume version 1.0.0.0.
+  return ContractVersion.fromString('1.0.0.0')
```
m-chrzan: Would putting this in an `else` to the above `if` make the codepaths clearer?

packages/protocol/scripts/bash/check-versions.sh
```diff
@@ -0,0 +1,39 @@
...
+rm -rf build/contracts
+yarn build
+rm -rf $BUILD_DIR_1 && mkdir -p $BUILD_DIR_1
+mv build/contracts $BUILD_DIR_1
+
+
+BUILD_DIR_2=$(echo build/$(echo $BRANCH_2 | sed -e 's/\//_/g'))
+git checkout $BRANCH_2
+rm -rf build/contracts
+yarn build:sol
```
m-chrzan, line 37: Above you had `yarn build`, here you have only `yarn build:sol`. Is this intended?

packages/protocol/scripts/bash/check-versions.sh
```diff
@@ -0,0 +1,39 @@
...
+rm -rf $BUILD_DIR_2 && mkdir -p $BUILD_DIR_2
+mv build/contracts $BUILD_DIR_2
+yarn ts-node scripts/check-backward.ts sem_check -o $BUILD_DIR_1/contracts -n $BUILD_DIR_2/contracts -e ".*Test.*|I[A-Z].*|ReleaseGold"
```
m-chrzan: To make this more readable, maybe name the regex?

asaj:
> Also, could you add tests to it? namely: a comparison of the removed metadata test in test/compatibility/ast-code.ts, and a few tests for the actual version retrieval from a compiled contract by calling the getContractVersion.

Are these tests being run in CI?

m-chrzan (commented):
Looks great. Mostly style comments. +1 to some unit tests to run in CI.

packages/protocol/scripts/bash/check-versions.sh
asaj: Yes, `yarn build:sol` did not exist in the previous branch, will update with a comment.

packages/protocol/lib/compatibility/version.ts
asaj: +1, cleaned this up in a slightly different way

packages/protocol/lib/compatibility/version.ts
asaj: It is not!

packages/protocol/lib/compatibility/version.ts
asaj: No, IMO

packages/protocol/lib/compatibility/ast-version.ts
asaj: Probably. I tried filtering the artifacts at the get-go and passing the filtered artifacts into the various constructors but got the following error that I didn't have any immediate leads on:
```
Running storage report...No AST nodes with id 6458 found
```

packages/protocol/lib/compatibility/ast-version.ts
asaj: I think it could go either way, I don't feel strongly if you do

packages/protocol/lib/compatibility/version.ts
hbandura: True. Still it may make sense to differentiate between a 'error calling the function' and 'function doesn't exist'.

packages/protocol/scripts/check-backward.ts
hbandura: Sure, not really needed right now

packages/protocol/lib/compatibility/ast-version.ts
hbandura: Sure as long asa we don't forget to do it xP

packages/protocol/lib/compatibility/ast-version.ts
hbandura: If you filter an artifact that's used from another one (say, you filter ParentContract which ChildContract uses), then you'll get that error when openzeppellin tries to import all the parent stuff when inspecting ChildContract. That's just the same error we get when we do an incremental build and node ids get stale.

hbandura (approved):

m-chrzan (approved):
Looks good!

packages/protocol/lib/compatibility/ast-version.ts
m-chrzan: Feels like that would remove a lot of boilerplate that otherwise noises up the code here and where this is used.

PR #4525, opened 2020-07-24
[Wallet] Add transitions to some screens on iOS

gnardini:
### Description

Added some transitions for screens that make elements appear vertically instead of horizontally.
The `PincodeEnter` screen is the exception, which has a fade in/out transition.

### Tested

Manually on simulator.

ezgif com-video-to-gif


packages/mobile/src/account/Education.tsx
```diff
@@ -180,6 +180,7 @@ export default class Education extends React.Component<Props, State> {
             loop={false}
             dotStyle={dotStyle}
             activeDotStyle={activeDotStyle}
+            removeClippedSubviews={false}
```
gnardini, line 183: Having this as `true` is a performance improvement for big lists, but this element has only 4 small images and without this the first image is blank until the user scrolls a little. It was the suggested fix on an issue from the library: https://github.com/leecade/react-native-swiper/issues/1199

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -112,7 +112,14 @@ const commonScreens = (Navigator: typeof Stack) => {
       <Navigator.Screen
         name={Screens.PincodeEnter}
         component={PincodeEnter}
-        options={headerWithBackButton}
+        options={{
+          ...noHeader,
```
gnardini: The fade animation was not working with the header. There is a `headerStyleInterpolator` property but it never faded no matter what I did with it (it just horizontally or vertically swiped). The header animations behave different on all transitions but on this one it was very noticeable and didn't look good.

What I did was remove the back button from the header and add it to the screen itself. This way it has the same fade in/out animation than the rest of the screen.

packages/mobile/src/pincode/PincodeEnter.tsx
```diff
@@ -31,8 +31,6 @@ interface StateProps {
 type Props = StateProps & WithTranslation & StackScreenProps<StackParamList, Screens.PincodeEnter>
 
 class PincodeEnter extends React.Component<Props, State> {
-  static navigationOptions = { gestureEnabled: false, ...nuxNavigationOptions }
-
```
jeanregisser: it's important that the user can't swipe to go back, the gesture part should remain.

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -241,7 +251,14 @@ const nuxScreens = (Navigator: typeof Stack) => (
 const sendScreens = (Navigator: typeof Stack) => (
   <>
     <Navigator.Screen name={Screens.Send} component={Send} options={sendScreenNavOptions} />
-    <Navigator.Screen name={Screens.QRNavigator} component={QRNavigator} options={noHeader} />
+    <Navigator.Screen
+      name={Screens.QRNavigator}
+      component={QRNavigator}
+      options={{
+        ...noHeader,
+        ...TransitionPresets.ModalSlideFromBottomIOS,
```
jeanregisser: Should be `ModalTransition` which under the hood uses `ModalSlideFromBottomIOS` on iOS and `RevealFromBottomAndroid` / `FadeFromBottomAndroid` on Android.

jeanregisser (changes requested):
Great initiative!
I think most the screens touched in this PR should be moved to `RootStackScreen` and use the iOS modal style there.
As is I see the following issues:
- for the send/request screen and language screen the nav bar animates from the top while the screen comes from the bottom, which doesn't look very standard and then, when closing them, the transition is left to right for the nav bar.

Other screen transitions (QR Code, CELO education) look good because they don't make use of the nav bar.

packages/mobile/src/pincode/PincodeEnter.tsx
gnardini: I don't think this was being used anymore, this was used in a previous version of `react-navigation`, now the options have to be set wherever the navigation is defined (`Navigator` file in our case). In fact, I'm able to swipe back on the Testflight build. I added the `gestureEnabled: false` in the options of this screen so as to disallow that :)

gnardini:
Updated the gif to show how the transitions look like now :)

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -527,18 +499,81 @@ const modalScreenOptions: ScreenOptions = Platform.select({
   }),
 })
 
+const animatedScreens = (Navigator: typeof Stack) => (
```
jmrossy: Is there an approach that doesn't require us to group together these unrelated screens this way? I'm worried that this grouping could lead to navigation bugs. I don't have a particular case in mind but something like a back button not working as expected for screens in a Stack Navigator

packages/mobile/src/pincode/PincodeEnter.tsx
jeanregisser: Yes I see what you mean. 
The code is currently in a transition phase for screen options as we decided to relocate them to each screen file using the `navigationOptions` static prop on the component, like it was with `react-navigation` v4.
We decided this to avoid cluttering the `Navigator` as we felt it belonged better with the component.

packages/mobile/src/navigator/Navigator.tsx
jeanregisser: I feel what you say, but unfortunately, I don't see another way for now.
`react-navigation` mentions we should avoid unnecessary nesting and only use them to get the desired UI (tab bar, modal, etc)

> Think of nesting navigators as a way to achieve the UI you want rather than a way to organize your code.

See https://reactnavigation.org/docs/nesting-navigators/#best-practices-when-nesting

jeanregisser: nit: how about we rename them to `modalAnimatedScreens`? And maybe move `SelectCountry` to it too.

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -527,18 +499,81 @@ const modalScreenOptions: ScreenOptions = Platform.select({
...
+      options={{
+        ...headerWithBackButton,
+        gestureEnabled: false,
+        cardStyleInterpolator: ({ current }) => ({
+          containerStyle: {
+            opacity: current.progress,
+          },
+        }),
+      }}
+    />
```
jeanregisser: Screen options should be moved to the component `navigationOptions` field.

jeanregisser (commented):
Looking almost good! 👍 
If you can move navigation options to components we can then merge it.

packages/mobile/src/navigator/Navigator.tsx
gnardini: Moved the options to the respective files :)

jeanregisser (approved):
Awesome!! ship it 👍 🚢

PR #4530, opened 2020-07-24
[Wallet] Fix default background color for screens

jeanregisser:
### Description

- This fixes the incorrect gray color in the CELO education screen
- Also stop redefining the background color for screens when unnecessary

### Tested

**before/after**

### Related issues

- #4415

### Backwards compatibility

Yes


packages/mobile/src/navigator/NavigatorWrapper.tsx
```diff
@@ -32,6 +33,15 @@ const getActiveRouteName = (state: NavigationState) => {
...
 
+// Global app them used by react-navigation
+const AppTheme = {
+  ...DefaultTheme,
+  colors: {
+    ...DefaultTheme.colors,
+    background: colors.light,
+  },
+}
+
```
jeanregisser, line 44: This is where we define the default background color for screens.

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -171,7 +171,7 @@ export class WalletHome extends React.Component<Props, State> {
       <RefreshControl
         refreshing={this.props.loading}
         onRefresh={this.onRefresh}
-        colors={[colors.celoGreen]}
+        colors={[colors.greenUI]}
```
aaronmgdr, line 174: nice

packages/mobile/src/navigator/NavigatorWrapper.tsx
aaronmgdr: cool.

also just noticed you can comment on multiple lines. nice

packages/mobile/src/account/Profile.tsx
```diff
@@ -69,18 +71,6 @@ const styles = StyleSheet.create({
     flexDirection: 'column',
     alignItems: 'center',
   },
-  underlinedBox: {
-    borderTopWidth: 1,
-    borderColor: '#EEEEEE',
-  },
-  scrollView: {
```
aaronmgdr, line 76: were these unused?

jeanregisser: Yep

aaronmgdr (approved):
Assuming that this works like you say it does it looks great! 

(ps i think you have some react-components package tests to fix)

PR #4531, opened 2020-07-24
Update setup guide

trianglesphere:
Fixes issues with docs

### Description

Gitbook is borked on a fresh install. Note, I've personally be able to fix this is by running `npm install` in `~/.gitbook/versions/3.2.3`. Not sure if that's a recommended fix.

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


tkporter (approved):
🙏

Issue #4214, opened 2020-06-25
[Wallet] Users SBAT use production Moonpay Cash In 


annakaz:
### Expected Behavior

Production environment is used by updating API keys in the /moonpay-auth cloud function and repointing wallet to that function. If cUSD is not yet live on Moonpay, will also require switch to CELO in Cash In flow

### Current Behavior

Staging environment is used with cUSD

annakaz:
If cUSD is not yet live on Moonpay, will also require switch to CELO in Cash In flow

PR #4533, opened 2020-07-24
[Wallet] Update Cash In for CELO purchases

annakaz:
### Description

Update Cash In flow for CELO purchases, as cUSD is not yet listed on any fiat exchange providers used in the wallet (only provider is Moonpay right now). 

Note that the minimum purchase amount for Moonpay is $20 (or a similar threshold in local currency, for example 2500 Kenyan Shillings), which gets prefilled if the user inputs less than $20. There is no indication to the user of this minimum in the Valora amount flow but it is out of scope of this PR. 

### Other changes

- Grey out "Add Funds" until Cash Out is enabled
- UI changes to FiatExchangeAmount (okayed with @coreycelo )
- All exchange rate fetching logic has been moved to FiatExchangeOptions. This is to make it easier to list prices when other providers are added to Valora

### Tested

Went through Moonpay flow with USD and with Kenyan shillings. 

### Related issues

- Related to #4214 

### Backwards compatibility

Yes, backwards compatible


annakaz:
Screenshots: 
Screen Shot 2020-07-24 at 11 25 27 AM
Screen Shot 2020-07-24 at 8 13 32 AM
Screen Shot 2020-07-24 at 8 13 46 AM
Screen Shot 2020-07-24 at 8 14 08 AM

packages/mobile/src/config.ts
```diff
@@ -75,9 +75,13 @@ export const DEFAULT_SYNC_MODE: GethSyncMode = Config.DEFAULT_SYNC_MODE
   : GethSyncMode.Lightest
 // TODO Remove when feature is stable
 export const USE_PHONE_NUMBER_PRIVACY = true
-export const SHOW_CASH_OUT = stringToBoolean(Config.SHOW_CASH_OUT || 'false')
 
 // SECRETS
 export const SEGMENT_API_KEY = keyOrUndefined(secretsFile, Config.SECRETS_KEY, 'SEGMENT_API_KEY')
 export const FIREBASE_WEB_KEY = keyOrUndefined(secretsFile, Config.SECRETS_KEY, 'FIREBASE_WEB_KEY')
 export const SENTRY_URL = keyOrUndefined(secretsFile, Config.SECRETS_KEY, 'SENTRY_URL')
+export const MOONPAY_PUBLIC_KEY = keyOrUndefined(
```
annakaz, line 83: The public key for hitting Moonpay's API to get the CELO price. Different from the secret key used in `/moonpay-auth` to sign requests.

packages/mobile/src/fiatExchanges/MoonPay.tsx
```diff
@@ -34,11 +32,9 @@ type Props = RouteProps
 
 function FiatExchangeWeb({ route }: Props) {
   const [uri, setUri] = React.useState('')
-  const { amount } = route.params
+  const { localAmount, currencyCode } = route.params
   const account = useSelector(currentAccountSelector)
-  const localCurrencyCode = useSelector(getLocalCurrencyCode)
```
annakaz, line 39: Moved currency conversion logic to FiatExchangeOptions as we may want to display the different providers' different fiat, CELO and cUSD rates down the line

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -4,20 +4,34 @@ import fontStyles from '@celo/react-components/styles/fonts.v2'
...
 import { moonpayLogo } from 'src/images/Images'
+import { LocalCurrencyCode } from 'src/localCurrency/consts'
+import { getLocalCurrencyCode } from 'src/localCurrency/selectors'
 import { emptyHeader, HeaderTitleWithSubtitle } from 'src/navigator/Headers.v2'
 import { Screens } from 'src/navigator/Screens'
 import { TopBarIconButton } from 'src/navigator/TopBarButton.v2'
 import { StackParamList } from 'src/navigator/types'
+import Logger from 'src/utils/Logger'
+
+const MOONPAY_RATE_API = `https://api.moonpay.io/v3/currencies/celo/price?apiKey=${MOONPAY_PUBLIC_KEY}`
```
jmrossy: Nit, could go in Config.ts

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -59,14 +78,55 @@ export const fiatExchangesOptionsScreenOptions = ({
...
+          : providerLocalAmount || new BigNumber(0),
+        currencyCode: providerlocalCurrency || FALLBACK_CURRENCY,
+      })
   }
   const onDismiss = () => {
     navigation.setParams({ isExplanationOpen: false })
   }
 
+  const dispatch = useDispatch()
+  React.useEffect(() => {
```
jmrossy: Nit, if you import `useEffect` directly then you can use it without going through React, similar to how you're using useDispatch in live above.

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -59,14 +78,55 @@ export const fiatExchangesOptionsScreenOptions = ({
...
+        setProviderLocalAmount(moonpayRates[FALLBACK_CURRENCY])
+      } else {
+        throw new Error('Unable to fetch Moonpay rates')
+      }
+    }
+    // TODO: Get rates from other providers when they are added
+    getMoonpayRates().catch((error) => {
+      Logger.error(
+        TAG,
+        `Failed to fetch Moonpay rate for ${localCurrency} at ${amount}, error: ${error}`
```
jmrossy: pass the error in as the 3rd param to Logger.error instead of interpolating it into the message.

jmrossy (approved):
Mostly lgtm, see minor comments below

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -59,14 +78,55 @@ export const fiatExchangesOptionsScreenOptions = ({
...
+        currencyCode: providerlocalCurrency || FALLBACK_CURRENCY,
+      })
   }
   const onDismiss = () => {
     navigation.setParams({ isExplanationOpen: false })
   }
 
+  const dispatch = useDispatch()
+  React.useEffect(() => {
+    async function getMoonpayRates() {
```
jmrossy, line 105: I wonder if this structure here could be simplified by using one of the hook utils we already have for doing async requests. For example see Jean's use of `useAsync` here: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/fees/CalculateFee.tsx#L68

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
annakaz: Good call

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
annakaz: 👍 updated for useEffect and useState

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
annakaz: 👌

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
annakaz: Oh sweet, hadn't seen that library. Simplified the function a bit but will wait to use the `react-async-hook` until we start using the local rates in the display

PR #4536, opened 2020-07-27
Add payment processing options to QRs and deep links

yorhodes:
### Description

Adds additional optional parameters to the QR code scheme in the app and formalizes encoding/decoding:
1. adds `currencyCode` for locale specific PoS systems
2. adds `amount` for prefilling transaction value in `SendConfirmation` component
3. adds `comment` for use with `transferWithComment`

### Other changes

- Adopts `io-ts` for parsing qr data contents
- Sets `preferredCurrencyCode` state to qr contents to minimize code churn

### Tested

 celo://wallet/pay?address=0xf7f551752A78Ce650385B58364225e5ec18D96cB&displayName=Super+8&currencyCode=PHP&amount=500&comment=92a53156-c0f2-11ea-b3de-0242ac13000

corresponding QR generated from deeplink:
beam-and-go (1)

Currently app does not gracefully handle insufficient user balances on the `SendConfirmation` screen

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


packages/mobile/src/navigator/types.tsx
```diff
@@ -154,6 +154,8 @@ export type StackParamList = {
     recipient: Recipient
     isOutgoingPaymentRequest?: true
     isFromScan?: boolean
+    currencyCode?: keyof typeof LocalCurrencyCode
```
annakaz: Why not directly use LocalCurrencyCode?

packages/mobile/src/qrcode/scheme.ts
```diff
@@ -0,0 +1,16 @@
+import { AddressType, E164PhoneNumberType } from '@celo/utils/lib/io'
+import * as t from 'io-ts'
+import { LocalCurrencyCode } from 'src/localCurrency/consts'
+
+export const QrDataType = t.type({
```
annakaz: Could you separate QrDataType into different types that depend on each other being non-null, one with just `address`, and another with the other requirements? It would be helpful to know for example whether currencyCode is required if amount is provided

packages/mobile/src/qrcode/utils.ts
```diff
@@ -100,26 +101,25 @@ export function* handleBarcode(
   isOutgoingPaymentRequest?: true,
   requesterAddress?: string
 ) {
-  let data: { address: string; e164PhoneNumber: string; displayName: string } | undefined
-
+  let data: object
```
annakaz: If this can be two different types can they both be listed? Anything more specific than `object` possible?

packages/mobile/src/qrcode/utils.ts
```diff
@@ -100,26 +101,25 @@ export function* handleBarcode(
...
-  }
-  if (typeof data !== 'object' || isEmpty(data.address)) {
-    yield put(showMessage(ErrorMessages.QR_FAILED_NO_ADDRESS))
     return
   }
-  if (!isValidAddress(data.address)) {
-    yield put(showMessage(ErrorMessages.QR_FAILED_INVALID_ADDRESS))
+
+  const either = qrDataFromJson(data)
+  if (isLeft(either)) {
```
annakaz: Is the library `fp-ts/lib/Either` used just to differentiate qrDataTypes? Could this by done just by checking the type directly/whether a particular exists?

annakaz (commented):
Excited to see this new data included in QR codes! Left some comments

Has there been any testing of whether lower end Android devices have cameras good enough for the new QR codes with more data?

packages/mobile/src/qrcode/QRCode.tsx
```diff
@@ -15,26 +16,16 @@ interface Props {
...
-const mapStateToProps = (state: RootState) => {
-  return {
-    name: state.account.name,
-    account: currentAccountSelector(state),
-    e164Number: state.account.e164PhoneNumber,
-  }
-}
+const mapStateToProps = (state: RootState): Partial<QrData> => ({
+  address: currentAccountSelector(state)!,
+  displayName: state.account.name === null ? undefined : state.account.name,
```
tarikbellamine: Could simplify to `state.account.name || undefined`

packages/mobile/src/qrcode/utils.ts
```diff
@@ -100,26 +101,25 @@ export function* handleBarcode(
...
+    yield put(showMessage(PathReporter.report(either)[0]))
     return
   }
+  const parsed = either.right
```
tarikbellamine: Nit: would be nice to have a more descriptive variable name here

packages/mobile/src/qrcode/scheme.ts
tarikbellamine: +1 

Are the two relevant QR code types send vs. request?

tarikbellamine (commented):
How would a user go about encoding an amount and currency code into a QR code? If those two items are included, wouldn't we also want to include a transaction comment? I guess the question becomes what the performance hit is to lower end cameras.

I also think when a QR with that data included is scanned, it should route to `SendConfirmation` rather than `SendAmount`.

packages/mobile/src/qrcode/utils.ts
yorhodes: its not possible
the qr could contain anything
the `qrDataFromJson` returns a strongly typed object

packages/mobile/src/qrcode/utils.ts
yorhodes: this emits specific parsing failures, like an invalid country code in the phone number field
I think this is a significant improvement over the legacy parsing

packages/mobile/src/qrcode/scheme.ts
```diff
@@ -0,0 +1,16 @@
+import { AddressType, E164PhoneNumberType } from '@celo/utils/lib/io'
```
jmrossy: Did you mean to call this file `schema`?

packages/mobile/src/qrcode/QRCode.tsx
```diff
@@ -15,26 +16,15 @@ interface Props {
...
-      JSON.stringify({
-        address: account,
-        e164PhoneNumber: e164Number,
-        displayName: name,
-      }),
-    [name, account, e164Number]
-  )
-
+  const data = useSelector(mapStateToProps, shallowEqual)
+  const qrContent = useMemo(() => JSON.stringify(data), Object.values(data))
```
cmcewen: Do you think we should switch this to use the new URL format? For better interop?

packages/mobile/src/qrcode/schema.ts
```diff
@@ -0,0 +1,17 @@
+import { AddressType, E164PhoneNumberType } from '@celo/utils/lib/io'
+import * as t from 'io-ts'
```
cmcewen: Is this required to import *? Also could you make this an explicit dependency?

packages/mobile/src/send/utils.ts
```diff
@@ -1,18 +1,34 @@
 import BigNumber from 'bignumber.js'
-import { showError } from 'src/alert/actions'
+import { isLeft } from 'fp-ts/lib/Either'
```
cmcewen: Do you know if this is tree shaken? https://bundlephobia.com/result?p=fp-ts@2.7.1

packages/mobile/src/qrcode/QRCode.tsx
yorhodes: probably a good idea

packages/mobile/src/send/utils.ts
yorhodes: this is used by utils and the attestations contract wrapper
doesn't seem like this is a significant regression

packages/mobile/src/send/utils.ts
```diff
@@ -105,3 +118,48 @@ export function showLimitReachedError(
...
+    const exchangeRate = yield select(getLocalCurrencyExchangeRate)
+    const amount = convertLocalAmountToDollars(data.amount, exchangeRate)
+    if (!amount) {
+      Logger.warn('handleSendPaymentData', 'null amount')
+      return
+    }
+    const transactionData: TransactionDataInput = {
+      recipient,
+      amount,
+      reason: data.comment,
```
annakaz, line 166: Does this reason get set as a comment here? https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/send/SendConfirmation.tsx#L147

annakaz (approved):
LGTM after adding amount check in the TODO

packages/mobile/src/qrcode/utils.ts
```diff
@@ -104,72 +99,44 @@ export function* handleBarcode(
...
   } catch (e) {
-    Logger.warn(TAG, 'QR code read failed with ' + e)
-  }
-  if (typeof data !== 'object' || isEmpty(data.address)) {
-    yield put(showMessage(ErrorMessages.QR_FAILED_NO_ADDRESS))
-    return
-  }
-  if (!isValidAddress(data.address)) {
-    yield put(showMessage(ErrorMessages.QR_FAILED_INVALID_ADDRESS))
+    Logger.warn(TAG, 'qr scan failed with ' + e)
```
jmrossy: Use Logger.error so you can pass the e in as the 3rd param

packages/mobile/src/send/utils.ts
```diff
@@ -105,3 +118,49 @@ export function showLimitReachedError(
...
+    displayId: data.e164PhoneNumber,
+    displayName: data.displayName || cachedRecipient?.displayName || 'anonymous',
+    phoneNumberLabel: cachedRecipient?.phoneNumberLabel,
+    thumbnailPath: cachedRecipient?.thumbnailPath,
+    contactId: cachedRecipient?.contactId,
+  }
+  yield put(storeLatestInRecents(recipient))
+
+  if (data.currencyCode) {
+    yield put(selectPreferredCurrency(data.currencyCode as LocalCurrencyCode))
```
jmrossy, line 143: Would this permanently set the user's currency to this new currencyCode?

jmrossy, line 143: Can UriData be typed correctly to prevent the need for casting here?

packages/mobile/src/send/utils.ts
```diff
@@ -105,3 +118,49 @@ export function showLimitReachedError(
...
+    yield take([Actions.FETCH_CURRENT_RATE_SUCCESS, Actions.FETCH_CURRENT_RATE_FAILURE])
```
jmrossy: What happens if it's FAILURE?

packages/utils/package.json
```diff
@@ -31,6 +31,8 @@
     "elliptic": "^6.5.3",
     "ethereumjs-util": "^5.2.0",
     "google-libphonenumber": "^3.2.4",
+    "io-ts": "2.0.1",
+    "fp-ts": "2.1.1",
```
jmrossy, line 35: Good catch 👍

packages/mobile/package.json
```diff
@@ -77,6 +77,8 @@
     "graphql": "^14.1.1",
     "hermes-engine": "0.4.2-rc1",
     "i18next": "^19.4.3",
+    "io-ts": "2.0.1",
+    "fp-ts": "^2.8.1",
```
jmrossy: Can you use 2.1.1 to stay consistent with the other packages? Or upgrade the others.

jmrossy (commented):
Mostly looks good! Just some small comments

packages/mobile/src/send/utils.ts
yorhodes: yes, this is unavoidable without a much larger refactor to the `SendConfirmation` component
I figured it's okay for the pilot purposes but we should find a workaround

yorhodes: enums in typescript are weird
it's typed as the value strings of the enum which aren't recognized by typescript as the enum type

jmrossy: Ok

jmrossy (approved):

jmrossy:
@yorhodes you'll need to update the yarn.lock by running yarn

PR #4542, opened 2020-07-27
AWS Terraform: Cloudwatch, EIPs, image updates, and bugfixes

bgeihsgt:
### Description

1. Separate examples into a testnet and mainnet example
2. Add support for cloudwatch logging
3. Use EIPs for attestation and proxy so IP is stable
4. Update images to latest
5. Some bug fixes if some configs aren't set
6. Don't replace nodes if AMI or user_data changes so people don't accidentally replace validators.

### Tested

I use these modules for my own mainnet infrastructure

### Related issues

None

### Backwards compatibility

These changes are not backward compatible with previous changes, but the `ignore_changes` changes should make an update up to the user.


tkporter (approved):
LGTM! @jcortejoso could you take a look too?

bgeihsgt:
@jcortejoso - wanting to ping on this review. Would love to get this merged this week.

Thanks.

packages/terraform-modules-public/aws/testnet/modules/proxy/main.tf
```diff
@@ -5,37 +5,54 @@ module "ami" {
 resource "aws_instance" "celo_proxy" {
```
jcortejoso, line 5: What do you think about creating and attaching a data volume to each instance to store there the celo chain folder? Full syncs are starting to take some time and having a volume outside the instance lifecycle could help if somebody needs to update some parameters of the instance.

packages/terraform-modules-public/aws/testnet/modules/attestation-service/main.tf
```diff
@@ -5,41 +5,60 @@ module "ami" {
 resource "aws_instance" "attestation_service" {
```
jcortejoso, line 5: What do you think about creating and attaching a data volume to each instance to store there the celo chain folder? Full syncs are starting to take some time and having a volume outside the instance lifecycle could help if somebody needs to update some parameters of the instance.

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/run-attestation-service.sh
```diff
@@ -9,13 +9,23 @@ mkdir $NODE_DIRECTORY
...
+ATTESTATION_NODE_CLOUDWATCH_LOG_GROUP_NAME=${cloudwatch_attestation_node_log_group_name}
+ATTESTATION_NODE_CLOUDWATCH_LOG_STREAM_NAME=${cloudwatch_attestation_node_log_stream_name}
+
+if [[ -z $ATTESTATION_NODE_CLOUDWATCH_LOG_GROUP_NAME || -z $ATTESTATION_NODE_CLOUDWATCH_LOG_STREAM_NAME ]]; then
+  DOCKER_LOGGING_PARAMS=''
+else
+  DOCKER_LOGGING_PARAMS="--log-driver=awslogs --log-opt awslogs-group=$ATTESTATION_NODE_CLOUDWATCH_LOG_GROUP_NAME --log-opt awslogs-stream=$ATTESTATION_NODE_CLOUDWATCH_LOG_STREAM_NAME"
+fi
+
+docker run -d --name celo-attestations $DOCKER_LOGGING_PARAMS --restart always -p 127.0.0.1:8545:8545 -v $PWD:/root/.celo $CELO_IMAGE --verbosity 3 --networkid $NETWORK_ID --syncmode full --nousb --rpc --rpcaddr 0.0.0.0 --rpcapi eth,net,web3,debug,admin --unlock 0x$CELO_ATTESTATION_SIGNER_ADDRESS --password /root/.celo/.password --bootnodes $BOOTNODE_ENODES --allow-insecure-unlock
```
jcortejoso, line 27: Sometimes verbosity `3` could be too verbose, but it could help with debugging. What is your opinion having this as a variable? For having only a few instances I think 3 is ok, but when you deploy more or with testnets sometimes makes sense to have in 1 to save some dollars on monitoring.

packages/terraform-modules-public/aws/testnet/variables.tf
```diff
@@ -35,6 +35,37 @@ variable instance_types {
   }
 }
 
+variable iam_instance_profiles {
+  description = "Optional IAM instances profile names for each component. For example, a profile that gives CloudWatch agent permissions"
```
jcortejoso: Makes sense to add here the link to https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/create-iam-roles-for-cloudwatch-agent.html as some context/example for configuring this?

jcortejoso (approved):
Looks great to me. Some comments only to check your opinion, but PR is good to me as it is.

packages/terraform-modules-public/aws/testnet/modules/proxy/main.tf
bgeihsgt: A cool idea. I was planning to solve this problem by backing up chaindata to s3 and providing an optional way to restore it. A data volume could also be a good option too. I'll probably start with the s3 option first (new PR) since that provides some more flexibility.

packages/terraform-modules-public/aws/testnet/modules/attestation-service/main.tf
bgeihsgt: See comment on other spot.

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/run-attestation-service.sh
bgeihsgt: I like the idea of making this a parameter. Will consider this in a future PR.

Re: dollars - in my experience the monitoring $$ is dwarfed by compute costs, so probably not a huge cost savings here.

packages/terraform-modules-public/aws/testnet/variables.tf
bgeihsgt: Good idea. https://github.com/celo-org/celo-monorepo/pull/4542/commits/efa5916b83fca15ad05b2d909eda6fb65e6753ea

bgeihsgt:
Thanks @jcortejoso ! Just responded to your comments.

packages/terraform-modules-public/aws/testnet/modules/proxy/main.tf
jcortejoso: Yeah, using S3 sounds great to me also for scaling perspective :+1:

packages/terraform-modules-public/aws/testnet/modules/startup-scripts/run-attestation-service.sh
jcortejoso: :+1: Haha, in my experience I saw several times the monitoring bill was the highest service used :), but probably caused by too verbose services.

jcortejoso:
@bgeihsgt Cool! Everything great from my side.

PR #4548, opened 2020-07-28
PNP Auth Update

codyborn:
### Description

- Uses DEK for auth (when available)
- Moves PNP requests to ContractKit (for standardization)
- Signer supports old and new auth approaches

These changes improve the wallet user experience since users won't have to enter their PIN when authenticating with PNP service. This also makes it easier for 3rd party apps to integrate with the Celo identity protocol.

### Tested

- [x] Unit tests
- [x] E2E on alfajores staging using wallet key auth (no invite code)
- [x] E2E on alfajores staging using DEK key auth
- [x] Old mobile client (curr master) works with updated combiner/signers

### Backwards compatibility

Signer is backwards compatible, but wallet changes are not compatible with old Signer. Therefore we'll need to update all Signers before rolling out the app update.


packages/contractkit/package.json
```diff
@@ -34,9 +34,10 @@
     "@ledgerhq/hw-app-eth": "^5.11.0",
     "@ledgerhq/hw-transport": "^5.11.0",
     "bignumber.js": "^9.0.0",
+    "blind-threshold-bls": "https://github.com/celo-org/blind-threshold-bls-wasm#e1e2f8a",
     "cross-fetch": "3.0.4",
     "debug": "^4.1.1",
-    "elliptic": "6.5.2",
+    "elliptic": "^6.5.3",
```
jmrossy: It seems we're using 3 different versions of this across the monorepo (protocol, utils, here).
Differing versions can cause unexpected behavior. It would be good to update the others as well

packages/contractkit/src/index.ts
```diff
@@ -6,6 +6,10 @@ export * from './kit'
 export { GenesisBlockUtils } from './network-utils/genesis-block-utils'
 export { StaticNodeUtils } from './network-utils/static-node-utils'
 export { CeloProvider } from './providers/celo-provider'
+export * from './utils/phone-number-lookup/bls-blinding-client'
+export * from './utils/phone-number-lookup/matchmaking'
+export * from './utils/phone-number-lookup/phone-number-identifier'
+export * from './utils/phone-number-lookup/phone-number-lookup'
```
jmrossy: Consider grouping these under a shared object, like `GenesisBlockUtils` above, to avoid polluting the global namespace with all these exports

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
```diff
@@ -0,0 +1,48 @@
+const crypto = require('crypto')
```
jmrossy: Whenever possible, ES6 imports are preferred (import foo from bar)

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
```diff
@@ -0,0 +1,48 @@
...
+    const unblindMessage = await this.thresholdBls.unblind(
+      blindedSignature,
+      this.blindedValue.blindingFactor
+    )
+    // this throws on error
+    await this.thresholdBls.verify(this.pgpnpPubKey, this.rawMessage, unblindMessage)
+    return unblindMessage
+  }
+
+  private reactNativeEnvironment(): boolean {
```
jmrossy: nit: `isReactNativeEnvironment` ?

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
```diff
@@ -0,0 +1,48 @@
...
+    return typeof navigator !== 'undefined' && navigator.product === 'ReactNative'
```
jmrossy:
```suggestion
    return navigator && navigator.product === 'ReactNative'
```

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  constructor(pgpnpPubKey: string) {
+    this.pgpnpPubKey = new Uint8Array(Buffer.from(pgpnpPubKey, 'base64'))
+    // Dynamically load the Wasm library
+    if (!this.reactNativeEnvironment()) {
+      this.thresholdBls = require('blind-threshold-bls')
+    } else {
+      // When using react instead rely upon this library instead
+      // https://github.com/celo-org/react-native-blind-threshold-bls#cc36392
+      throw Error('Cannot use WasmBlsBlindingClient in a react app')
```
jmrossy:
```suggestion
      throw new Error('Cannot use WasmBlsBlindingClient in a React Native app')
```

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export class WasmBlsBlindingClient implements BlsBlindingClient {
+  private thresholdBls: any
+  private blindedValue: any
+  private pgpnpPubKey: Uint8Array
+  private rawMessage: Buffer | undefined
+
+  constructor(pgpnpPubKey: string) {
+    this.pgpnpPubKey = new Uint8Array(Buffer.from(pgpnpPubKey, 'base64'))
```
jmrossy: I believe Buffer is a subclass of Uint8Array, so there's no need to wrap it here.
https://stackoverflow.com/questions/54814215/node-js-buffer-vs-unint8array#:~:text=Uint8Array%20is%20a%20general%2Dpurpose,manipulating%20binary%20(byte)%20data.

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
```diff
@@ -0,0 +1,48 @@
...
+  async blindMessage(base64PhoneNumber: string): Promise<string> {
+    const userSeed = crypto.randomBytes(32)
+    this.rawMessage = Buffer.from(base64PhoneNumber)
+    this.blindedValue = await this.thresholdBls.blind(this.rawMessage, userSeed)
+    const blindedMessage = this.blindedValue.message
+    return Buffer.from(blindedMessage).toString('base64')
+  }
+
+  async unblindAndVerifyMessage(base64BlindSig: string): Promise<string> {
+    const blindedSignature = new Uint8Array(Buffer.from(base64BlindSig, 'base64'))
```
jmrossy: ditto

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
```diff
@@ -0,0 +1,101 @@
...
+const TAG = 'contractkit/utils/phone-number-lookup/matchmaking'
+
+// Uses the phone number privacy service to find mutual matches between Celo users
+export async function getContactMatches(
+  e164NumberCaller: E164Number,
+  e164NumberMatches: E164Number[],
+  account: string,
+  signer: AuthSigner,
+  context: ServiceContext,
+  logger: Logger,
```
jmrossy: Are there other instances of Logger being passed into CK? 
I think it mostly uses a different approach for logging:
https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/web3-contract-cache.ts#L147

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
```diff
@@ -0,0 +1,101 @@
...
+  failureCallback: (response: Response) => void,
+  walletVersion?: string
```
jmrossy: Should this function assume it only serves wallets?

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
```diff
@@ -0,0 +1,101 @@
...
+
+// Uses the phone number privacy service to find mutual matches between Celo users
+export async function getContactMatches(
+  e164NumberCaller: E164Number,
+  e164NumberMatches: E164Number[],
+  account: string,
+  signer: AuthSigner,
+  context: ServiceContext,
+  logger: Logger,
+  failureCallback: (response: Response) => void,
```
jmrossy: Can we throw instead of using callbacks?

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
```diff
@@ -0,0 +1,101 @@
...
+const MATCHMAKING_ENDPOINT = '/getContactMatches'
+// Eventually, the matchmaking process will use blinded numbers same as salt lookups
+// But for now numbers are simply hashed using this static salt
+const SALT = '__celo__'
+const TAG = 'contractkit/utils/phone-number-lookup/matchmaking'
+
+// Uses the phone number privacy service to find mutual matches between Celo users
+export async function getContactMatches(
+  e164NumberCaller: E164Number,
+  e164NumberMatches: E164Number[],
```
jmrossy: Are these the contact numbers of the matches?

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
```diff
@@ -0,0 +1,101 @@
...
+  const response = await postToPhoneNumPrivacyService<MatchmakingResponse>(
+    signer,
+    body,
+    context,
+    MATCHMAKING_ENDPOINT,
+    logger,
+    failureCallback
+  )
+
+  const matchHashes: string[] = response.matchedContacts.map((match) => match.phoneNumber)
```
jmrossy, line 47: Nit why the explicit type here? Are we missing a type on `response` that makes the inference not work?

packages/contractkit/src/utils/phone-number-lookup/phone-number-identifier.ts
```diff
@@ -0,0 +1,88 @@
...
+import {
+  AuthSigner,
+  Logger,
+  postToPhoneNumPrivacyService,
+  ServiceContext,
+  SignMessageRequest,
+  SignMessageResponse,
+} from './phone-number-lookup'
+
+// const crypto = require('crypto')
```
jmrossy: Remove?

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
```diff
@@ -0,0 +1,103 @@
...
+  pgpnpPubKey: string
+}
+
+export async function postToPhoneNumPrivacyService<ResponseType>(
+  signer: AuthSigner,
+  body: PhoneNumberPrivacyRequest,
+  context: ServiceContext,
+  endpoint: string,
+  logger: Logger,
+  handleFailure: (response: Response) => void
```
jmrossy: Ditto feedback as above

packages/contractkit/src/utils/phone-number-lookup/phone-number-identifier.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export async function getPhoneNumberIdentifier(
+  e164Number: string,
+  account: string,
+  signer: AuthSigner,
+  context: ServiceContext,
+  logger: Logger,
+  failureCallback: (response: Response) => void,
+  selfPhoneHash?: string,
+  walletVersion?: string,
```
jmrossy: Ditto feedback as above

packages/mobile/src/identity/bls-blinding-client.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  constructor(pgpnpPubKey: string) {
+    this.pgpnpPubKey = pgpnpPubKey
+  }
+
+  async blindMessage(base64PhoneNumber: string): Promise<string> {
+    return (await BlindThresholdBls.blindMessage(base64PhoneNumber)).trim()
+  }
+
+  async unblindAndVerifyMessage(base64BlindSig: string): Promise<string> {
```
jmrossy: No need for this to be async

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -292,3 +295,39 @@ function* updatePhoneNumberMappings(newIdentityData: IdentityMetadataInTx[]) {
   yield put(updateE164PhoneNumberSalts(e164NumberToSaltUpdates))
   yield put(updateE164PhoneNumberAddresses(e164NumberToAddressUpdates, addressToE164NumberUpdates))
 }
+
+export function* getAuthSignerForAccount(account: string) {
```
jmrossy: Recommend moving this to src/web3/dataEncryptionKey

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -292,3 +295,39 @@ function* updatePhoneNumberMappings(newIdentityData: IdentityMetadataInTx[]) {
...
+    const encryptionKeyPrivate: string | null = yield select(dataEncryptionKeySelector)
+    if (!encryptionKeyPrivate) {
+      Logger.error(TAG + 'getAuthSignerForAccount', 'Missing comment key, should never happen.')
+    } else {
+      const encryptionKeyPublic = hexToBuffer(compressedPubKey(hexToBuffer(encryptionKeyPrivate)))
+      if (encryptionKeyPublic !== registeredEncryptionKey) {
+        Logger.warn(TAG + 'getAuthSignerForAccount', 'DEK mismatch.')
+      } else {
+        Logger.info(TAG + 'getAuthSignerForAccount', 'Using DEK for authentication')
+        authSigner = {
```
jmrossy: Nit: How about just returning here instead of re-assigning the var and then checking it below?

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -292,3 +295,39 @@ function* updatePhoneNumberMappings(newIdentityData: IdentityMetadataInTx[]) {
...
+  const registeredEncryptionKey: Buffer | null = yield call(doFetchDataEncryptionKey, account)
+  let authSigner: AuthSigner | undefined
+  if (registeredEncryptionKey) {
+    const encryptionKeyPrivate: string | null = yield select(dataEncryptionKeySelector)
+    if (!encryptionKeyPrivate) {
+      Logger.error(TAG + 'getAuthSignerForAccount', 'Missing comment key, should never happen.')
+    } else {
+      const encryptionKeyPublic = hexToBuffer(compressedPubKey(hexToBuffer(encryptionKeyPrivate)))
+      if (encryptionKeyPublic !== registeredEncryptionKey) {
+        Logger.warn(TAG + 'getAuthSignerForAccount', 'DEK mismatch.')
```
jmrossy:
```suggestion
        Logger.error(TAG + 'getAuthSignerForAccount', 'DEK mismatch.')
```

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -292,3 +295,39 @@ function* updatePhoneNumberMappings(newIdentityData: IdentityMetadataInTx[]) {
...
+  // Use the DEK for authentication if the current DEK is registered with this account
+  const registeredEncryptionKey: Buffer | null = yield call(doFetchDataEncryptionKey, account)
+  let authSigner: AuthSigner | undefined
+  if (registeredEncryptionKey) {
+    const encryptionKeyPrivate: string | null = yield select(dataEncryptionKeySelector)
+    if (!encryptionKeyPrivate) {
+      Logger.error(TAG + 'getAuthSignerForAccount', 'Missing comment key, should never happen.')
+    } else {
+      const encryptionKeyPublic = hexToBuffer(compressedPubKey(hexToBuffer(encryptionKeyPrivate)))
+      if (encryptionKeyPublic !== registeredEncryptionKey) {
```
jmrossy: Consider using `eqAddress` here. Reading on-chain data can mess with casing

packages/mobile/src/identity/phoneNumPrivacyService.ts
```diff
@@ -1,23 +1,26 @@
...
-import { ContractKit } from '@celo/contractkit'
+import { ec as EC } from 'elliptic'
 import { ErrorMessages } from 'src/app/ErrorMessages'
 import networkConfig from 'src/geth/networkConfig'
 import Logger from 'src/utils/Logger'
+const ec = new EC('secp256k1')
 
 const TAG = 'identity/phoneNumPrivacyService'
 
 export async function postToPhoneNumPrivacyService<ResponseType>(
```
jmrossy: Are we planning on moving this too?

packages/phone-number-privacy/signer/src/common/identity.ts
```diff
@@ -1,21 +1,46 @@
...
+  if (!registeredEncryptionKey) {
+    logger.info(`Account ${signer} does not have registered encryption key`)
+  } else {
+    const validSignature: boolean = verifySignature(
+      message,
+      messageSignature,
+      registeredEncryptionKey
+    )
+    if (validSignature) {
+      return true
```
jmrossy: return validSignature?

packages/phone-number-privacy/signer/src/common/identity.ts
```diff
@@ -1,21 +1,46 @@
...
   logger.debug('Authenticating user')
 
   // https://tools.ietf.org/html/rfc7235#section-4.2
   const messageSignature = request.get('Authorization')
+  const message = JSON.stringify(request.body)
+  if (!messageSignature || !message) {
+    return false
+  }
   const signer = request.body.account
   if (!messageSignature || !signer) {
```
jmrossy, line 21: Nit, looks like you just checked this above

jmrossy (commented):
This would be a good time for us to decide on PGPNP's future name before we file all this under 'phone-number-lookup'

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
codyborn: No, this is just the pre-filtered list of contacts. Updated to `e164NumberContacts`

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
codyborn: I was trying to be consistent with the existing mobile logging. I can update it here to stick with the CK approach and we can revisit in the future to have a consistent story across modules.

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
codyborn: I like to add explicit type for legibility to easily see what type the object is and it also adds a sanity check if the type isn't as expected. Happy to remove if it doesn't make sense. WDYT?

packages/mobile/src/identity/phoneNumPrivacyService.ts
codyborn: Ah this was already moved (see `phone-number-lookup.ts`). Just forgot to remove it from here.. thanks!

packages/phone-number-privacy/signer/src/common/identity.ts
codyborn: If it's false, I don't want to return it since I want to be able to fallback to the wallet approach. I don't expect Valora to fall into this case, but 3rd parties might call only using the wallet key as an example.

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
codyborn: Good point.. I wanted a way to differentiate between client and server version.. will rename to `clientVersion`

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
jmrossy: Totally fine, was just wondering if it was an override of some kind.

packages/contractkit/src/index.ts
```diff
@@ -6,6 +6,7 @@ export * from './kit'
 export { GenesisBlockUtils } from './network-utils/genesis-block-utils'
 export { StaticNodeUtils } from './network-utils/static-node-utils'
 export { CeloProvider } from './providers/celo-provider'
+export * from './utils/phone-number-lookup/phone-number-lookup-utils'
```
jmrossy: Sorry, I should have been more clear. This approach still exports everything from under /phone-number-lookup/ into the global namespace. What I was suggesting was wrapping the exports into an object. Something like:
```
import foo from bar

const PhoneNumberUtils = {
  foo,
}
```

Then export `PhoneNumberUtils` here

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
```diff
@@ -0,0 +1,48 @@
+import { randomBytes } from 'crypto'
+
+export interface BlsBlindingClient {
+  blindMessage: (base64PhoneNumber: string) => Promise<string>
+  unblindAndVerifyMessage: (blindedMessage: string) => Promise<string>
+}
+
+export class WasmBlsBlindingClient implements BlsBlindingClient {
+  private thresholdBls: any
```
jmrossy, line 9: We could make an interface for this, something that has a `blind` and `unblind` method.

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
```diff
@@ -0,0 +1,99 @@
...
+import debugFactory from 'debug'
+import {
+  AuthSigner,
+  MatchmakingRequest,
+  MatchmakingResponse,
+  postToPhoneNumPrivacyService,
+  ServiceContext,
+} from './phone-number-lookup'
+
+const debug = debugFactory('kit:registry')
```
jmrossy: should be kit:phone-number-lookup:matchmaking

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const MATCHMAKING_ENDPOINT = '/getContactMatches'
+// Eventually, the matchmaking process will use blinded numbers same as salt lookups
+// But for now numbers are simply hashed using this static salt
+const SALT = '__celo__'
+const TAG = 'contractkit/utils/phone-number-lookup/matchmaking'
```
jmrossy: Do we still need the tag if we use the debugFactory?

packages/contractkit/src/utils/phone-number-lookup/phone-number-identifier.ts
```diff
@@ -0,0 +1,88 @@
...
+import { BlsBlindingClient, WasmBlsBlindingClient } from './bls-blinding-client'
+import {
+  AuthSigner,
+  postToPhoneNumPrivacyService,
+  ServiceContext,
+  SignMessageRequest,
+  SignMessageResponse,
+} from './phone-number-lookup'
+
+const debug = debugFactory('kit:registry')
```
jmrossy: ditto

packages/contractkit/src/utils/phone-number-lookup/phone-number-identifier.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const SALT_CHAR_LENGTH = 13
+const SIGN_MESSAGE_ENDPOINT = '/getDistributedBlindedSalt'
+const TAG = 'contractkit/utils/phone-number-lookup/phone-number-identifier'
```
jmrossy: ditto

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup-utils.ts
```diff
@@ -0,0 +1,4 @@
+export * from './bls-blinding-client'
```
jmrossy: Consider renaming this to `index.ts`

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
```diff
@@ -0,0 +1,124 @@
+// Utilities for interacting with the Phone Number Privacy Service service (aka PGPNP)
+
+import { hexToBuffer, trimLeading0x } from '@celo/utils/lib/address'
+import debugFactory from 'debug'
+import { ec as EC } from 'elliptic'
+import { ContractKit } from '../../kit'
+
+const TAG = 'contractkit/utils/phone-number-lookup/phone-number-lookup'
+const debug = debugFactory('kit:registry')
```
jmrossy: ditto

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -92,107 +89,42 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
...
+      networkConfig,
+      selfPhoneHash,
+      DeviceInfo.getVersion(),
+      blsBlindingClient
     )
-    return response.combinedSignature
   } catch (error) {
     if (error.message === ErrorMessages.PGPNP_QUOTA_ERROR) {
-      throw new Error(ErrorMessages.SALT_QUOTA_EXCEEDED)
+      throw new Error(ErrorMessages.MATCHMAKING_QUOTA_EXCEEDED)
```
jmrossy: This file is for getting salts, not matchmaking. Was this change intentional?

packages/mobile/src/identity/verification.ts
```diff
@@ -160,6 +160,11 @@ export function* doVerificationFlow() {
 
       yield put(setVerificationStatus(VerificationStatus.RequestingAttestations))
 
+      // Account needs to be unlocked to submit attestations
+      const unlocked: boolean = yield call(unlockAccount, account)
+      if (!unlocked) {
+        throw new Error(ErrorMessages.INCORRECT_PIN)
+      }
```
jmrossy: This addition isn't necessary because we call `getConnectedUnlockedAccount` above so the account is already unlocked.

packages/mobile/src/identity/verification.ts
```diff
@@ -189,6 +194,11 @@ export function* doVerificationFlow() {
 
       yield put(setVerificationStatus(VerificationStatus.RevealingNumber))
       ValoraAnalytics.track(VerificationEvents.verification_reveal_all_attestations_start)
+      // Unlock account to sign attestations txs
+      const success: boolean = yield call(unlockAccount, account)
+      if (!success) {
+        throw new Error(ErrorMessages.INCORRECT_PIN)
+      }
```
jmrossy: Ditto

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -132,6 +133,7 @@ async function isAccountUpToDate(
     accountsWrapper.getWalletAddress(address),
     accountsWrapper.getDataEncryptionKey(address),
   ])
+  Logger.debug(`${TAG}/isAccountUpToDate`, `Found match for DEK ${currentDEK}`)
```
jmrossy: We shouldn't be logging the DEK and also we don't know if we found a match. That's determined on the next line which checks equality

packages/phone-number-privacy/combiner/package.json
```diff
@@ -27,6 +27,7 @@
     "@celo/phone-number-privacy-common": "1.0.5",
     "@celo/utils": "0.1.13",
     "blind-threshold-bls": "https://github.com/celo-org/blind-threshold-bls-wasm#e1e2f8a",
+    "elliptic": "6.5.2",
```
jmrossy: Should this be 6.5.3 now?

packages/phone-number-privacy/signer/package.json
```diff
@@ -21,9 +21,11 @@
     "@celo/phone-number-privacy-common": "1.0.5",
     "@celo/utils": "0.1.13",
     "@google-cloud/secret-manager": "3.0.0",
+    "@types/elliptic": "^6.4.12",
     "aws-sdk": "^2.696.0",
     "blind-threshold-bls": "https://github.com/celo-org/blind-threshold-bls-wasm#e1e2f8a",
     "dotenv": "^8.2.0",
+    "elliptic": "6.5.2",
```
jmrossy: ditto

jmrossy (commented):
Almost there, just needs a bit more tweaking!

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
codyborn: That's what `BlsBlindingClient` is. Did you have something different in mind?

packages/contractkit/src/utils/phone-number-lookup/matchmaking.ts
codyborn: Good point, I see that debugFactory handles this in the constructor.

packages/mobile/src/identity/verification.ts
codyborn: Should we change the above to `getConnectedAccount`, and only unlock if `!status.isVerified`? It seems that if the user is using DEK and the account is already verified, then it's unnecessary to request account unlock

packages/mobile/src/web3/dataEncryptionKey.ts
codyborn: Changed to `DEK associated with account ${currentDEK}`. Why shouldn't we log the DEK? From my understanding this is the public key pulled from the Accounts contract.

packages/phone-number-privacy/combiner/package.json
codyborn: Yes, thanks

packages/contractkit/src/utils/phone-number-lookup/bls-blinding-client.ts
jmrossy: Fair enough, resolving.

packages/mobile/src/web3/dataEncryptionKey.ts
jmrossy: That's true. I guess logging any kind of keys just makes me a bit nervous because these logs gets uploaded to Firebase. Though there's in theory nothing to be concerned about here.

packages/mobile/src/identity/verification.ts
jmrossy: In most cases, the account is already unlocked when a user gets to verification because the account is unlocked during invite redemption or account import. It's even less common for them to reach here but be already verified. Given how complex the verification code is already, it's better not to add more code unless it's necessary. Were you hitting this case?

codyborn: I did hit it but only when I intentionally waited 5 minutes.. I'll remove the changes - less risk and we can re-add if it's an issue.

packages/mobile/src/identity/privateHashing.ts
codyborn: Good eye, thank you

packages/mobile/src/identity/matchmaking.test.ts
```diff
@@ -63,19 +74,27 @@ describe('Fetch contact matches', () => {
...
       ])
+      .withState(state)
       .put(addContactsMatches(expectedMatches))
       .run()
   })
 })
 
-describe(obfuscateNumberForMatchmaking, () => {
+describe(PNPUtils.Matchmaking.obfuscateNumberForMatchmaking, () => {
   it('Hashes sigs correctly', () => {
```
jmrossy, line 96: Would be good to move this test into contract kit. Seems odd to test it from the outside like this.

packages/mobile/src/identity/privateHashing.test.ts
```diff
@@ -54,9 +71,12 @@ describe('Fetch phone hash details', () => {
   })
 })
 
-describe(getSaltFromThresholdSignature, () => {
+describe(PNPUtils.PhoneNumberIdentifier.getSaltFromThresholdSignature, () => {
```
jmrossy, line 74: Ditto

jmrossy (approved):
Ship it!

packages/mobile/src/identity/matchmaking.test.ts
codyborn: Since we need to get this out soon, I'm going to postpone this to a future PR. I've created this item to track it here:
https://github.com/celo-org/celo-monorepo/issues/4592
Thanks!

PR #4549, opened 2020-07-28
update lockedgold spec for new Certora prover

nd-certora:
Due to a new version of the Certora Prover with many fixes, among which fixes to the modelling of the behavior when unknown external calls are performed. (e.g., as triggered by msg.sender.transfer(...)
In the `noChangeByOther `rule, we attempt to check that the native balance of an address a that is not invoking a function f (in our case, f=withdraw() is not changed by the operation f. But this is not true because underlying withdraw() there is a call to a transfer() that relinquishes control and may indeed, for example, perform a transfer to the address a and increase its balance.


packages/protocol/specs/lockedGold.spec
```diff
@@ -101,7 +101,7 @@ rule noChangeByOther(address a, address b, method f) {
 	uint256 ercBalance_ = sinvoke ercBalanceOf(a); 
 	uint256 accountNonVoting_ = sinvoke getAccountNonvotingLockedGold(a);
 	uint256 accountTotalPendingWithdrawals_ =  sinvoke getTotalPendingWithdrawals(a);
-	assert(_ercBalance == ercBalance_, "Unexpected change to erc tokens");
+	assert(f.selector != withdraw(uint256).selector => _ercBalance == ercBalance_, "Unexpected change to erc tokens");
```
yorhodes, line 104: should this include the `slash` selector too?

nd-certora: slash does not satisfy the assertions in the rule and is explicitly excluded ( a few lines up)

yorhodes (approved):
LGTM. Thanks for the fix!

PR #4550, opened 2020-07-28
[Wallet] Some minor cleanup

jmrossy:
### Description

- Remove dead code related to Verifier app
- Move PGPNP flag to flags.ts
- Address some TODOs

### Backwards compatibility

Yes


packages/mobile/locales/en-US/walletFlow5.json
```diff
@@ -23,9 +23,6 @@
   "exchange": "Exchange",
   "viewAll": "View All",
   "thanksForVerifying": "Thanks for verifying",
-  "earnCeloRewards": "Earn Celo Rewards for verifying others",
```
aaronmgdr, line 26: Nice

aaronmgdr (approved):

packages/mobile/src/account/reducer.ts
```diff
@@ -61,7 +60,6 @@ export const initialState = {
   backupCompleted: false,
   backupDelayedTime: 0,
   socialBackupCompleted: false,
-  dismissedEarnRewards: false,
```
tarikbellamine, line 64: Should this also be removed from the `schema` file?

tarikbellamine (approved):
Minor comment about schema but otherwise looks good!

jeanregisser (approved):
Great 👍

packages/mobile/src/account/reducer.ts
jmrossy: Good catch!

PR #4552, opened 2020-07-28
[Wallet] Minor fixes before build shipping 7/28

annakaz:
### Description

Minor fixes for build shipping 7/28

- Verification modal navigation
One line fix from @i1skn 's PR here: https://github.com/celo-org/celo-monorepo/commit/1952b259d5e81a01f630525824cb52980993f9f5#diff-ca8a9b866182b4dffdeff0db064e5f47R30

- Fixes from Connor's Valora branch from last week: spawn instead of call `appInit` saga, add DisconnectBanner to WalletImport screen


packages/mobile/src/redux/sagas.ts
```diff
@@ -69,7 +69,7 @@ export function* rootSaga() {
   // Delay all sagas until rehydrate is done
   // This prevents them from running with missing state
   yield call(waitForRehydrate)
-  yield call(appInit)
+  yield spawn(appInit)
```
jmrossy, line 72: Was the call causing issues? This was originally `spawn` but I changed it when I fixed a few saga related issues. The intention was to have this be the work we *need* to run first before all other sagas start firing.

annakaz: I see- this change built on @cmcewen 's try catch which was shipped in the last build: https://github.com/celo-org/celo-monorepo/compare/cmcewen/valora?expand=1#diff-f3c49d6759b745014b7db88fa58c2b51R38 and was what @cmcewen suggested - while testing I did not encounter issues from the call, but wanted to avoid a regression from the last build

Moving forward I think it's important that what we deploy matches what we have on master as closely as possible to avoid confusion like this

i1skn:
My change is already merged, so we can remove it from this PR 🙂

PR #4545, opened 2020-07-28
[Wallet] Add new onboarding Welcome screen

jeanregisser:
### Description

This PR implements the new [onboarding welcome screen](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=6184%3A72673).

Highlights:
- This new screen is pushed after the onboarding education screen.
- Whether the user chooses `Create Account` or `Restore Account` the next few screens will be the same
  - `RegulatoryTerms`
  - `NameAndNumber`
  - `PincodeSet`
- Then the next screen will be either `EnterInviteCode` or `ImportWallet` based on the initial choice.
- Going back from those 2 screens will directly go back to the `Welcome` screen (skipping the intermediary screens.
- Making a new choice on the Welcome screen will again go through the same screens listed above, but with some info pre-filled (name and number). The PIN will need to be created again. This was decided to avoid confusion.
- The user can interrupt this onboarding process anytime. It will pick up at the appropriate place when the app is resumed/relaunched. Note that if the user didn't reach `EnterInviteCode` or `ImportWallet`, it will resume at the onboarding education.
- Add new analytics events.

See assets diff: https://github.com/clabs-co/valora-app-branding/compare/a3f5867...d58569f

### Other changes

- Renamed `JoinCelo` -> `NameAndNumber`
- Removed restore account button from `EnterInviteCode` screen, since we now have the welcome screen for that choice.
- Don't hide Android emulator window during e2e test when NOT running on the CI

### Related issues

- Fixes #4401

### Backwards compatibility

Yes

PR #4557, opened 2020-07-29
[Wallet] fix attestation codes being wrongly set as completed

### LINKED ISSUE #4477: Error processing attestation code :: No issuer found for attestion code
### Expected Behavior

Users able to complete verification

### Current Behavior

One of the three text messages resulted in this error.

`Info [2020-07-16T17:14:11.355Z] identity/verification@attestationCodeReceiver :: Error processing attestation code :: No issuer found for attestion code in index.android.bundle:1884:11497`

Can provide more specific information if needed

i1skn:
### Description

Fix a bug, described in details here - https://github.com/celo-org/celo-monorepo/issues/4477#issuecomment-665275466

### Other changes

* Remove error toast, when verification fails, cause we show Modal instead.
* Fix a bug, when user presses `Enter Codes` on the error modal but nothing happens until timer has elapsed.

### Tested

iOS simulator

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/4477

### Backwards compatibility

Yes


packages/mobile/src/identity/actions.ts
```diff
@@ -69,14 +70,19 @@ export interface ReceiveAttestationMessageAction {
   inputType: CodeInputType
 }
 
+export interface SetCompletedCodes {
```
jmrossy: Should be `SetCompletedCodesAction` or `SetNumCompletedCodesAction`

packages/mobile/src/redux/migrations.ts
```diff
@@ -54,4 +54,13 @@ export const migrations = {
       },
     }
   },
+  4: (state: any) => {
```
jmrossy, line 57: Please make a test for this new migration in `migrations.test.ts`

jmrossy (commented):
LGTM, see two minor comments below

packages/mobile/src/redux/migrations.ts
```diff
@@ -54,4 +54,13 @@ export const migrations = {
...
+    return {
+      ...state,
+      identity: {
+        ...state.identity,
+        acceptedAttestationCodes: [],
+      },
+    }
+  },
```
jeanregisser, line 65: Not super important but I think we should also update the test schema to avoid confusion.
https://github.com/celo-org/celo-monorepo/blob/52d5de63a990d3bb5dbf8333b59722e489f1588a/packages/mobile/test/schemas.ts#L332
I saw `v3Schema` was missing and added it in https://github.com/celo-org/celo-monorepo/pull/4545/files#diff-2d8d77f9d884fa463eafd5c95a7cb7d0

What do you think?

jeanregisser (commented):
Looks great to me! 👍 
I'll let people with a deeper understanding of the verification do the actual approval 👀

packages/mobile/src/redux/migrations.ts
i1skn: I think, if we treat migration as something, which could update only the data and keep the schema untouched, we should only update the schema when it was really changed. So migration number and schema are disconnected. wdyt?

jmrossy (approved):

packages/mobile/src/redux/migrations.ts
jeanregisser: Yep!

aslawson (commented):
nice work! 🙌

PR #4568, opened 2020-07-29
Add transaction confirmed event

nategraf:
### Description

Currently, `transaction_start` is emitted when a transaction confirmation is returned from `web3`.
This causes two `transaction_start` events to appear in the analytics database for each transaction, in addition to not being the intended semantics event.

This PR adds a `transaction_confirmed` event which is emitted when a transaction confirmation is received from `web3`


tarikbellamine (approved):
Looks good to me

jmrossy (approved):

PR #4572, opened 2020-07-30
Headless CMS thru contentful for Experience Kits

aaronmgdr:
### Description

Adds ability to dynamically create/manage/update kit content via Contentful CMS

This is especially dynamic as it allows pages at celo.org/experience/*/* to be create completely in Contentful and our Kit Layout will apply. Inner components are rendered via Contentful's node mapping. The result is flexibility to choose what renders on the cms side but how it looks is determined in code. 

This should reduce the amount of code needed to run the website and frequency of updating it.

### Tested
built a merchant kit using this. see at dev.celo.org/experience/merchant


packages/web/pages/experience/[kit]/[kitPage].tsx
```diff
@@ -0,0 +1,6 @@
+import Kit from 'src/experience/contentful/ContentfulKit'
```
Comment (aaronmgdr) [inline packages/web/pages/experience/[kit]/[kitPage].tsx:1]: the `[` `]` in filename create a dynamic route

packages/web/src/experience/common/Showcase.tsx
```diff
@@ -81,9 +82,7 @@ const styles = StyleSheet.create({
   },
   text: { flex: 1 },
   pullStart: { paddingLeft: 0 },
-  container: {
-    justifyContent: 'space-between',
-  },
+  container: {},
```
jmrossy: Empty style

jmrossy (approved):

PR #4573, opened 2020-07-30
Update dependencies due to security vulnerabilities

jeanregisser:
### Description

Update `elliptic` and `dot-prop` dependencies due to security vulnerabilities
See https://www.npmjs.com/advisories/1547
See https://www.npmjs.com/advisories/1213

As a temporary solution until we can republish our packages, I've set custom resolutions in yarn.

TODO to completely eliminate the vulnerable versions of `elliptic <= 6.5.2`:
- [ ] republish `contractkit` and `utils` which depends on fixed version `elliptic@6.5.2`
- [ ] update `ethers` which also which depends on fixed version `elliptic@6.3.3`

### Other changes

None

### Tested

Waiting for test to pass on CircleCI.

### Related issues

Discussed on Slack

### Backwards compatibility

Should be 👀


packages/protocol/package.json
```diff
@@ -43,7 +43,6 @@
     "@0x/sol-profiler": "^3.0.0",
     "@0x/sol-trace": "^2.0.16",
     "@0x/subproviders": "^5.0.0",
-    "@openzeppelin/upgrades": "^2.8.0",
```
gastonponti, line 46: why this one? (it doesn't seem to be part of the same issue)

jeanregisser: This line was duplicated (see below) and when I used `yarn upgrade-interactive` it rewrote it (and others).

gastonponti: Ah, you are right!

gastonponti (approved):

codyborn (approved):

jmrossy (approved):

jmrossy:
RE ethers, we don't have any direct dependencies on it so we may be able to solve that transitive dep issue with a yarn resolution config

PR #4575, opened 2020-07-30
[Wallet] Migrate all colors to v2

i1skn:
### Description

This PR migrate usage of old `colors` to `colors.v2` and then renames `colors.v2` to just `colors`, so v2 now is the only version and a default one.

All colors, which were not in the `colors.v2` are replaced with similar ones.

### Tested

`yarn build` and `yarn test`

### Backwards compatibility

Yes


packages/react-components/styles/colors.tsx
```diff
@@ -1,38 +1,79 @@
...
-  sectionBorder: '#C4C4C4',
-  errorRed: '#FD785B',
-  messageBlue: '#3C9BF4',
-  purple: '#D39CFE',
-  teal: '#7AD6FE',
-  orange: '#FEB45E',
-  disconnectBannerGrey: 'rgba(46, 51, 56, 0.4)',
-  disconnectBannerRed: '#FF7F6D',
+function deprecationNotice(color: string) {
+  console.warn(`deprecated: trying colors.${color}`)
```
aaronmgdr: at this point maybe we dont show a warning?

aaronmgdr (approved):
super duper

packages/react-components/styles/colors.tsx
i1skn: Fixed this!

jmrossy (approved):

PR #4577, opened 2020-07-30
Support decryption in ContractKit's wallet/signer interface

nambrot:
### Description

This PR adds a `decrypt` function to the wallet/signer interface in ContractKit. It looks to be the primary interface upon which applications can use encryption.

### Tested

- Small unit test


gastonponti (approved):
LGTM!

PR #4578, opened 2020-07-30
Fix certora rules broken by #4523

asaj:
### Description

#4523 broke certora rules by making an internal mapping private, to prevent the compatibility tool detecting a version change. However, now that bytecode metadata is being stripped, no change is being detected and this mapping can be made internal with no versioning implications

### Other changes

None

### Tested

No
### Related issues

None
### Backwards compatibility

Yes


yorhodes (approved):
lgtm

yorhodes:
already in master

PR #4580, opened 2020-07-30
Add sell & buy functions to Exchange.sol, deprecate exchange function

tkporter:
### Description

* Adds `function buy(uint256 buyAmount, uint256 maxSellAmount, bool buyGold)`
* Adds `function sell(uint256 sellAmount, uint256 minBuyAmount, bool sellGold)`
* Deprecates `function exchange(uint256 sellAmount, uint256 minBuyAmount, bool sellGold)` which has the same functionality as `sell`
* Runs same suite of tests for `sell` and `exchange`
* Adds tests for `buy`

### Other changes

n/a

### Tested

Ran tests

### Related issues

Related CIP: https://github.com/celo-org/celo-proposals/pull/26

### Backwards compatibility

Fully compatible


packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -297,6 +373,27 @@ contract Exchange is
...
   }
 
+  /**
+   * @dev Returns the amount of sellToken a user would need to exchange to receive buyAmount of
+   * buyToken.
+   * @param buyAmount The amount of buyToken the user would like to purchase.
+   * @param sellGold `true` if gold is the sell token
+   * @return The corresponding sellToken amount.
+   */
+  function _getSellTokenAmount(uint256 buyAmount, bool sellGold) private view returns (uint256) {
```
tkporter: This is a copy of `getSellTokenAmount`, I copied just because that is what was done for `_getBuyTokenAmount`/`getBuyTokenAmount`. Is there a reason for copying rather than having `getBuyTokenAmount` call `_getBuyTokenAmount`?

nvtaveras:
Do we need to update the version of this contract?
https://github.com/celo-org/celo-monorepo/blob/d0d8e1d6849c8a620ea714ac634e97721cd800c9/packages/protocol/contracts/stability/Exchange.sol#L66-L68

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -107,17 +107,95 @@ contract Exchange is
...
+   * @dev Deprecated. Exchanges sellAmount of sellToken in exchange for at
+   * least minBuyAmount of buyToken.
+   * Requires the sellAmount to have been approved to the exchange
+   * @param sellAmount The amount of sellToken the user is selling to the exchange
+   * @param minBuyAmount The minimum amount of buyToken the user has to receive for this
+   * transaction to succeed
+   * @param sellGold `true` if gold is the sell token
+   * @return The amount of buyToken that was transfered
+   * @dev This function can be frozen using the Freezable interface.
+   */
```
nvtaveras: nit: how about saying "Deprecated. See comment in sell" instead of duplicating the comment? 

Similar to https://github.com/celo-org/celo-monorepo/pull/4580/files#diff-b3cbc9c000b2a3d8f2f8360587d70368R394

packages/protocol/contracts/stability/Exchange.sol
nvtaveras: I would like to get more context on this too. I did noticed that `_getBuyTokenAmount` is calling `_getBuyAndSellBuckets` instead of `getBuyAndSellBuckets` as in your function below.

I think the logic there is that `_getBuyAndSellBuckets` doesn't try to update the buckets because that is already done in `sell`/`buy` with the `updateBucketsIfNecessary` modifier.

yerdua: I also have this question.

packages/protocol/test/common/integration.ts
```diff
@@ -414,7 +414,7 @@ contract('Integration: Exchange', (accounts: string[]) => {
       originalGold = await goldToken.balanceOf(accounts[0])
       originalReserve = await goldToken.balanceOf(reserve.address)
       await goldToken.approve(exchange.address, sellAmount)
-      await exchange.exchange(sellAmount, minBuyAmount, true)
+      await exchange.sell(sellAmount, minBuyAmount, true)
```
yerdua: Could `exchange.buy` also be tested here?

packages/protocol/test/stability/exchange.ts
```diff
@@ -548,30 +1095,30 @@ contract('Exchange', (accounts: string[]) => {
...
             updatedGoldBucket,
             updatedStableBucket
           )
 
-          it('the exchange should succeed', async () => {
-            await exchange.exchange(
-              goldTokenAmount,
-              expectedStableAmount.integerValue(BigNumber.ROUND_FLOOR),
-              true,
+          it(`the exchange should succeed`, async () => {
```
yerdua: nit: was the change of quote style intentional? Nothing's being interpolated into this string...

yerdua (approved):
Looks good to me, but someone with more solidity experience should also sign-off.

packages/protocol/contracts/stability/Exchange.sol
rcroessmann: +1 and maybe @m-chrzan has some more background on this?

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -107,17 +107,95 @@ contract Exchange is
...
+   * @dev Exchanges buyAmount of buyToken in exchange for at most maxSellAmount of sellToken
+   * Requires the sellAmount to have been approved to the exchange
+   * @param buyAmount The amount of buyToken the user is buying from the exchange
+   * @param maxSellAmount The maximum amount of sellToken the user can sell for this
+   * transaction to succeed
+   * @param buyGold `true` if gold is the buy token
+   * @return The amount of sellToken that was sold
+   * @dev This function can be frozen using the Freezable interface.
+   */
+  function buy(uint256 buyAmount, uint256 maxSellAmount, bool buyGold)
```
rcroessmann, line 154: The way `buy` and `sell` work solves everything discussed in https://github.com/celo-org/celo-proposals/pull/26. To increase usability even further and mimic how the trading interface a centralized exchanges usually looks like, we could think about alternatively (or additionally) offering functions

`buyCELO(uint256 buyAmount, uint256 maxPrice)`

and 

`sellCELO(uint256 sellAmount, uint256 minPrice)`

where the price parameter is always the number of cUSD (paid/received) per CELO. Allowing folks to specify the price directly avoids that they need to calculate the minBuyAmount or maxSellAmount themselves from the price they target. WDYT?

rcroessmann: Thought about this again - I think it's totally fine to just stay with `buy` and `sell` for now and add the above functions later if we at some point feel like we need to improve usability further.

rcroessmann (approved):

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -107,17 +107,95 @@ contract Exchange is
...
     onlyWhenNotFrozen
     updateBucketsIfNecessary
     nonReentrant
     returns (uint256)
+  {
+    return _sell(sellAmount, minBuyAmount, sellGold);
+  }
+
+  /**
+   * @dev Exchanges sellAmount of sellToken in exchange for at least minBuyAmount of buyToken
```
asaj: This is an @notice, not an @dev. Ditto everywhere else

packages/protocol/contracts/stability/Exchange.sol
asaj: The above interface can (and should) be implemented at the contractKit level to save gas

packages/protocol/contracts/stability/Exchange.sol
tkporter: I'm sort of wary of doing that because the `@...` comments are used for automatically creating documentation for functions (https://solidity.readthedocs.io/en/develop/natspec-format.html). Any input from @m-chrzan as the resident smart contract expert would be helpful

packages/protocol/contracts/stability/Exchange.sol
tkporter: Ah I didn't realize that @nvtaveras, that makes sense to me. I'll change right now to use `_getBuyAndSellBuckets`. Thanks!

Context from familiar folks would still be helpful

packages/protocol/test/common/integration.ts
tkporter: good idea, will do that

asaj:
In retrospect I don't think the version number needs bumping since it's already been bumped to 1.1.1.0 since the last release

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -98,7 +98,28 @@ contract Exchange is
   }
 
   /**
-   * @dev Exchanges sellAmount of sellToken in exchange for at least minBuyAmount of buyToken
+   * @notice Exchanges sellAmount of sellToken in exchange for at least minBuyAmount of buyToken
```
asaj:
```
@notice Exchanges a specific amount of one token for an unspecified amount of another.
@param sellAmount The number of tokens to send to the exchange.
@param minBuyAmount The minimum number of tokens for the exchange to send in return.
@param sellGold True if the caller is sending CELO to the exchange, false otherwise.
@return The number of tokens sent by the exchange.
@dev The caller must first have approved `sellAmount` to the exchange.
@dev This function can be frozen via the Freezable interface.
```

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -98,7 +98,28 @@ contract Exchange is
...
+    onlyWhenNotFrozen
+    updateBucketsIfNecessary
+    nonReentrant
+    returns (uint256)
+  {
+    return _sell(sellAmount, minBuyAmount, sellGold);
+  }
+
+  /**
+   * @notice Deprecated. Exchanges sellAmount of sellToken in exchange for at
```
asaj:
```
@dev DEPRECATED - Use `buy` or `sell`
@notice Exchanges a specific amount of one token for an unspecified amount of another.
@param sellAmount The number of tokens to send to the exchange.
@param minBuyAmount The minimum number of tokens for the exchange to send in return.
@param sellGold True if the caller is sending CELO to the exchange, false otherwise.
@return The number of tokens sent by the exchange.
@dev The caller must first have approved `sellAmount` to the exchange.
@dev This function can be frozen via the Freezable interface.```

File: packages/protocol/contracts/stability/Exchange.sol [hunk 9]
```diff
@@ -113,11 +134,68 @@ contract Exchange is
     updateBucketsIfNecessary
     nonReentrant
     returns (uint256)
+  {
+    return _sell(sellAmount, minBuyAmount, sellGold);
+  }
+
+  /**
+   * @notice Exchanges sellAmount of sellToken in exchange for at least minBuyAmount of buyToken
```
asaj:
```
@notice Exchanges a specific amount of one token for an unspecified amount of another.
@param sellAmount The number of tokens to send to the exchange.
@param minBuyAmount The minimum number of tokens for the exchange to send in return.
@param sellGold True if the caller is sending CELO to the exchange, false otherwise.
@return The number of tokens sent by the exchange.
@dev The caller must first have approved `sellAmount` to the exchange.
```

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -113,11 +134,68 @@ contract Exchange is
...
     uint256 buyAmount = _getBuyTokenAmount(sellAmount, sellGold);
 
     require(buyAmount >= minBuyAmount, "Calculated buyAmount was less than specified minBuyAmount");
 
+    _exchangeAmounts(sellAmount, buyAmount, sellGold);
+    return buyAmount;
+  }
+
+  /**
+   * @notice Exchanges buyAmount of buyToken in exchange for at most maxSellAmount of sellToken
```
asaj:
```
@notice Exchanges an unspecified amount of one token for a specific amount of another.
@param buyAmount The number of tokens for the exchange to send in return.
@param maxSellAmount The maximum number of tokens to send to the exchange.
@param buyGold True if the exchange is sending CELO to the caller, false otherwise.
@return The number of tokens sent to the exchange.
@dev The caller must first have approved `maxSellAmount` to the exchange.
@dev This function can be frozen via the Freezable interface.
```

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -113,11 +134,68 @@ contract Exchange is
...
+      sellAmount <= maxSellAmount,
+      "Calculated sellAmount was greater than specified maxSellAmount"
+    );
+
+    _exchangeAmounts(sellAmount, buyAmount, sellGold);
+    return sellAmount;
+  }
+
+  /**
+   * @notice Exchanges sellAmount of sellToken in exchange for buyAmount of buyToken
```
asaj:
```
@notice Exchanges a specific amount of one token for a specific amount of another.
@param sellAmount The number of tokens to send to the exchange.
@param buyAmount The number of tokens for the exchange to send in return.
@param sellGold True if the msg.sender is sending CELO to the exchange, false otherwise.
```

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -113,11 +134,68 @@ contract Exchange is
...
+   * Requires the sellAmount to have been approved to the exchange
+   * @param sellAmount The amount of sellToken the user is selling to the exchange
+   * @param buyAmount The amount of buyToken the user is buying from the exchange
+   * @param sellGold `true` if gold is the sell token
+   */
+  function _exchangeAmounts(uint256 sellAmount, uint256 buyAmount, bool sellGold) private {
```
asaj: consider `_exchange`

packages/protocol/contracts/stability/Exchange.sol
asaj: Yeah @nvtaveras is right, but I think we can do better here. Wdyt of the following example? We would do the same for the `buy` versions of these functions as well
```
function getSellTokenAmount(uint256 buyAmount, bool sellGold) external view returns (uint256) {
    (uint256 buyTokenBucket, uint256 sellTokenBucket) = getBuyAndSellBuckets(sellGold);
    return _getSellTokenAmount(buyTokenBucket, sellTokenBucket, buyAmount);
}

function _getSellTokenAmount(uint256 buyTokenBucket, uint256 sellTokeBucket, buyAmount) external view returns (uint256) {
  // Shared code in the current versions of _getSellTokenAmount and getSellTokenAmount goes here
}

function buy(uint256 buyAmount, uint256 maxSellAmount, bool buyGold) {
  (uint256 buyTokenBucket, uint256 sellTokenBucket) = _getBuyAndSellBuckets(!buyGold);
   uint256 sellAmount = _getSellTokenAmount(buyTokenBucket, sellTokenBucket, buyAmount);
   ...
}
```

asaj (commented):
Minor nits and a small refactor suggestion to clean up some tech debt

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -98,14 +98,33 @@ contract Exchange is
...
   /**
-   * @dev Exchanges sellAmount of sellToken in exchange for at least minBuyAmount of buyToken
-   * Requires the sellAmount to have been approved to the exchange
-   * @param sellAmount The amount of sellToken the user is selling to the exchange
-   * @param minBuyAmount The minimum amount of buyToken the user has to receive for this
-   * transaction to succeed
-   * @param sellGold `true` if gold is the sell token
-   * @return The amount of buyToken that was transfered
-   * @dev This function can be frozen using the Freezable interface.
+   * @notice Exchanges a specific amount of one token for an unspecified amount of another.
```
m-chrzan: What do you think about leaving the "for at least..." language from before?

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -98,14 +98,33 @@ contract Exchange is
...
+   * @param sellAmount The number of tokens to send to the exchange.
+   * @param minBuyAmount The minimum number of tokens for the exchange to send in return.
+   * @param sellGold True if the caller is sending CELO to the exchange, false otherwise.
```
m-chrzan, line 105: Should we change this to sellCelo? Guess it's not obvious since that contract will remain named GoldToken, but maybe the strategy is to do renames whenever we upgrade a contract.

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -98,14 +98,33 @@ contract Exchange is
...
+    updateBucketsIfNecessary
+    nonReentrant
+    returns (uint256)
+  {
+    return _sell(sellAmount, minBuyAmount, sellGold);
+  }
+
+  /**
+   * @dev DEPRECATED - Use `buy` or `sell`
+   * @notice Exchanges a specific amount of one token for an unspecified amount of another.
```
m-chrzan: If you take suggestions for comment above, copy here.

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -114,10 +133,66 @@ contract Exchange is
     nonReentrant
     returns (uint256)
   {
-    uint256 buyAmount = _getBuyTokenAmount(sellAmount, sellGold);
+    return _sell(sellAmount, minBuyAmount, sellGold);
+  }
+
+  /**
+   * @notice Exchanges a specific amount of one token for an unspecified amount of another.
```
m-chrzan: Ditto.

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -114,10 +133,66 @@ contract Exchange is
...
+   * @param sellAmount The number of tokens to send to the exchange.
+   * @param minBuyAmount The minimum number of tokens for the exchange to send in return.
+   * @param sellGold True if the caller is sending CELO to the exchange, false otherwise.
+   * @return The number of tokens sent by the exchange.
+   * @dev The caller must first have approved `sellAmount` to the exchange.
+   */
+  function _sell(uint256 sellAmount, uint256 minBuyAmount, bool sellGold)
```
m-chrzan: Actually, any reason to have this as a separate function rather than put this logic in `sell` and have `exchange` call `sell`?

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -114,10 +133,66 @@ contract Exchange is
...
+    uint256 buyAmount = _getBuyTokenAmount(buyTokenBucket, sellTokenBucket, sellAmount);
 
     require(buyAmount >= minBuyAmount, "Calculated buyAmount was less than specified minBuyAmount");
 
+    _exchange(sellAmount, buyAmount, sellGold);
+    return buyAmount;
+  }
+
+  /**
+   * @notice Exchanges an unspecified amount of one token for a specific amount of another.
```
m-chrzan: Again, might want to add some language around "unspecified" being constrained, maybe "unspecified (up to a threshold) amount..."

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -141,56 +216,33 @@ contract Exchange is
...
+    return _getBuyTokenAmount(buyTokenBucket, sellTokenBucket, sellAmount);
   }
 
   /**
-   * @dev Returns the amount of sellToken a user would need to exchange to receive buyAmount of
-   * buyToken.
+   * @notice Returns the amount of sell tokens a user would need to exchange to receive buyAmount of
+   * buy tokens.
    * @param buyAmount The amount of buyToken the user would like to purchase.
    * @param sellGold `true` if gold is the sell token
```
m-chrzan: gold -> CELO

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -293,7 +346,32 @@ contract Exchange is
...
+    // than maxFixedDivisor.
+    // Fortunately, we expect an integer result, so integer division gives us as
+    // much precision as we could hope for.
+    return numerator.unwrap().div(denominator.unwrap());
+  }
+
+  /**
+   * @notice Returns the amount of sell tokens a user would need to exchange to receive buyAmount of
+   * buy tokens.
+   * @param buyTokenBucket The buy token bucket size
```
m-chrzan: Mega nit: periods at ends of sentences in natspecs.

packages/protocol/test/common/integration.ts
```diff
@@ -408,61 +406,129 @@ contract('Integration: Exchange', (accounts: string[]) => {
...
-      await exchange.exchange(sellAmount, minBuyAmount, true)
-      finalStable = await stableToken.balanceOf(accounts[0])
-      finalGold = await goldToken.balanceOf(accounts[0])
-      finalReserve = await goldToken.balanceOf(reserve.address)
-    })
+  describe('Selling', () => {
+    const sellAmount = new BigNumber('1000000000000000000')
+    const minBuyAmount = 1
+
+    describe('When selling gold', () => {
```
m-chrzan, line 413: gold -> CELO?

packages/protocol/test/common/integration.ts
```diff
@@ -408,61 +406,129 @@ contract('Integration: Exchange', (accounts: string[]) => {
...
 
-    it(`should reduce user's stable`, async () => {
-      assert.isTrue(finalStable.lt(originalStable))
-    })
+      it(`should increase user's stable`, async () => {
+        assert.isTrue(finalStable.gt(originalStable))
+      })
+
+      it(`should reduce user's gold`, async () => {
+        const block = await web3.eth.getBlock('latest')
```
m-chrzan: Could this miner-checking code be pulled out to a function to deduplicate? Potentially even out to test-utils.ts, since I think this is used in other tests as well.

packages/protocol/test/stability/exchange.ts
```diff
@@ -376,13 +387,549 @@ contract('Exchange', (accounts: string[]) => {
     })
   })
 
-  describe('#exchange', () => {
+  describe('#getSellTokenAmount', () => {
+    it('should return the correct amount of sellToken', async () => {
+      const amount = 10
```
m-chrzan: Consider `buyAmount`

packages/protocol/test/stability/exchange.ts
```diff
@@ -376,13 +387,549 @@ contract('Exchange', (accounts: string[]) => {
...
+        new BigNumber(amount),
+        initialGoldBucket,
+        initialStableBucket
+      )
+
+      assert.equal(sellAmount.toString(), expectedSellAmount.toString())
+    })
+  })
+
+  // Run the following test for both these functions
```
m-chrzan: Maybe add a note that `exchange` is deprecated?

packages/protocol/test/stability/exchange.ts
```diff
@@ -376,13 +387,549 @@ contract('Exchange', (accounts: string[]) => {
...
+  const sellFunctionNames = ['sell', 'exchange']
+
+  for (const fnName of sellFunctionNames) {
+    describe(`#${fnName}`, () => {
+      const user = accounts[1]
+
+      // This test is run for both the `sell` and `exchange` functions
+      let fn: ExchangeInstance['sell'] | ExchangeInstance['exchange']
```
m-chrzan: Meaningful variable name, please. E.g. `sellFunction`.

packages/protocol/test/stability/exchange.ts
```diff
@@ -376,13 +387,549 @@ contract('Exchange', (accounts: string[]) => {
...
+
+      assert.equal(sellAmount.toString(), expectedSellAmount.toString())
+    })
+  })
+
+  // Run the following test for both these functions
+  const sellFunctionNames = ['sell', 'exchange']
+
+  for (const fnName of sellFunctionNames) {
+    describe(`#${fnName}`, () => {
```
m-chrzan: Style nit: not as important here, but still wold prefer `fnName` -> `sellFunctionName`.

packages/protocol/test/stability/exchange.ts
```diff
@@ -376,13 +387,549 @@ contract('Exchange', (accounts: string[]) => {
...
+            true,
+            {
+              from: user,
+            }
+          )
+          const newTotalSupply = await stableToken.totalSupply()
+          assert.isTrue(newTotalSupply.eq(oldTotalSupply.plus(expectedStableBalance)))
+        })
+
+        it('should affect token supplies', async () => {
```
m-chrzan: Maybe supplies -> buckets, since supply can be confused for global floating supply?

m-chrzan (approved):
Looks great, just a bunch of nits. Biggest code question is if we should remove the `sell` -> `_sell` indirection.

packages/protocol/contracts/stability/Exchange.sol
asaj: +1, we can make `sell` public and remove the modifiers from `exchange`

packages/protocol/contracts/stability/Exchange.sol
```diff
@@ -141,56 +216,33 @@ contract Exchange is
...
-    FixidityLib.Fraction memory denominator = FixidityLib.newFixed(sellTokenBucket).add(
-      reducedSellAmount
-    );
-
-    // Can't use FixidityLib.divide because denominator can easily be greater
-    // than maxFixedDivisor.
-    // Fortunately, we expect an integer result, so integer division gives us as
-    // much precision as we could hope for.
-    return numerator.unwrap().div(denominator.unwrap());
+    (uint256 buyTokenBucket, uint256 sellTokenBucket) = getBuyAndSellBuckets(sellGold);
```
asaj, line 213: Should we put the short circuit above (return 0) in the internal version of the function?

tkporter:
Thanks @asaj and @m-chrzan! I'm holding off on the gold -> CELO rename for now, I just made a thread in #eng-contracts to discuss it a little bit

PR #4583, opened 2020-07-30
Add transferAttestationMapping to Attestations.sol

aslawson:
### Description

As part of the initiative for [Feeless Attestations with Meta Transactions](https://docs.google.com/document/d/1wBFX6c1Tu4Dyxrx9JdKhmXPXXaFmVpHmP5bAAzmX6EU/edit#heading=h.79symh8gkwep), we need to be able to migrate attestations between Valora externally owned account and the smart wallet (or vice versa if Valora chooses not to be backed by the smart wallet and we only use the smart wallet as a mean for covering attestations fees)

NOTE: I followed the `revoke` function's pattern of requiring the index of the address as a parameter.  I'm assuming this is a pattern to reduce transaction cost as the caller can make a view call to `lookupAccountsForIdentifier` and determine the index themselves.

### Other changes

NA

### Tested

Added Test

### Related issues

[Feeless Attestations with Meta Transactions](https://docs.google.com/document/d/1wBFX6c1Tu4Dyxrx9JdKhmXPXXaFmVpHmP5bAAzmX6EU/edit#heading=h.79symh8gkwep)

### Backwards compatibility

Requires smart contract release -- adds new external function.  Does not impact current storage or existing functions.


packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -649,6 +654,34 @@ contract Attestations is
...
+    external
+  {
+    uint256 numAccounts = identifiers[identifier].accounts.length;
+    require(index < numAccounts, "Index is invalid");
+    require(
+      msg.sender == identifiers[identifier].accounts[index],
+      "Index does not match msg.sender"
+    );
+    identifiers[identifier].accounts[index] = replacementAddress;
+    AttestedAddress memory attestedAddress = identifiers[identifier].attestations[msg.sender];
```
asaj: nit: Do we need to keep these in memory? Or could we do something like:
```
identifiers[identifier].attestations[replacementAddress] = identifiers[identifier].attestations[msg.sender];
delete identifiers[identifier].attestations[msg.sender];
```

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -649,6 +654,34 @@ contract Attestations is
...
 
+  /**
+   * @notice Transfer an attestation identifier mapping from the sender address to a replacement address.
+   * @param identifier The hash of the identifier for this attestation.
+   * @param index The index of the account in the accounts array.
+   * @param replacementAddress The address to replace the sender address in the indifier mapping.
+   * @dev Throws if index is out of bound for account array.
+   & @dev Throws if msg.sender is not in the account array at the given index.
+   */
+  function transferAttestationMapping(bytes32 identifier, uint256 index, address replacementAddress)
```
asaj: nit: Consider just `transfer`

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -649,6 +654,34 @@ contract Attestations is
...
+    );
+    identifiers[identifier].accounts[index] = replacementAddress;
+    AttestedAddress memory attestedAddress = identifiers[identifier].attestations[msg.sender];
+    UnselectedRequest memory unselectedRequests = identifiers[identifier].unselectedRequests[msg
+      .sender];
+    delete identifiers[identifier].attestations[msg.sender];
+    delete identifiers[identifier].unselectedRequests[msg.sender];
+    identifiers[identifier].attestations[replacementAddress] = attestedAddress;
+    identifiers[identifier].unselectedRequests[replacementAddress] = unselectedRequests;
+    emit AttestationMappingTransferred(identifier, msg.sender, replacementAddress);
```
asaj: Consider `AttestationsTransferred`

asaj (commented):
Only looked at the contract, but those changes LGTM, just some minor nits.

Will leave to the folks in Europe to think more deeply on the implications of making attestations transferrable.

asaj:
> NOTE: I followed the revoke function's pattern of requiring the index of the address as a parameter. I'm assuming this is a pattern to reduce transaction cost as the caller can make a view call to lookupAccountsForIdentifier and determine the index themselves.

Exactly right 😄

asaj:
Oh, and you're going to need to add a version number! CI will complain otherwise 😄 

You can run `yarn check-versions -o release/contracts/0 -n YOUR_BRANCH` to get more details

packages/protocol/contracts/identity/Attestations.sol
aslawson: ya that seems doable

packages/protocol/contracts/identity/Attestations.sol
aslawson: sure

packages/protocol/contracts/identity/Attestations.sol
aslawson: kk

nambrot:
So one thing to think about is that we should not let transfers happen where the replacement account is already on the identifier, otherwise you can "overwrite" the attestations of the replacement account which is not ideal.

nambrot:
Also wondering if we want to require a signature from the `replacementAddress`?

bowd:
Based on the chat we had I updated the contract to with an approval mapping and functionality.
I've also added a requirement that the `replacementAddress` doesn't have exiting attestations. I'm having trouble writing the test for that though, and it's currently failing. I left a `TODO` comment there.

I'm also a bit unsure about how I'm verifying that invariant:
```solidity
identifiers[identifier].attestations[replacementAddress].requested == 0
```
Left a `TODO` comment there as well.

I'm gonna take a break now and will check in a bit later.

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -680,6 +698,17 @@ contract Attestations is
       msg.sender == identifiers[identifier].accounts[index],
       "Index does not match msg.sender"
     );
+    require(
+      transferAllowed[replacementAddress][msg.sender] == true,
```
aslawson: @bohan-zhang is the implication of having the `transferAllowed` that the `msg.sender` could now be the Onboarding contract?  In which case, we should change all these `msg.sender` in this `transfer` method to an `address originalAddress` parameter

asaj: cc @bowd ?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -649,6 +668,57 @@ contract Attestations is
     }
   }
 
+  /**
+   * @notice Update the approval status of allowing an attestation identifier 
+   * mapping to be transfered from the specified address to msg.sender
+   * @param source The current attesation address to be transfered
+   * @param status The approval status
+   */
+  function approveTransfer(address source, bool status) external returns (bool) {
```
asaj: Posting this comment for posterity:
The current design requires that the transferee approve before the transferer transfers. Wdyt of the following?
```
mapping(address => mapping(bytes32 => bool)) transferApprovals;
function approveTransfer(bytes32 identifier, address from, address to, bool status) {
  bytes32 key = keccak(identifier, from, to);
  transferApprovals[msg.sender][key] = status;
  if (status && other party approved) {
     _transfer(identifier, from, to);
    clear approvals
  } 
}
```

packages/protocol/contracts/identity/Attestations.sol
aslawson: lol whoop

bowd (approved):
🎉 Great job! The only thing that I had to think about a bit is the naming of the "isOtherAddressApproved" internal function. I feel like there's something to improve in clarity there, but I don't have any good suggestions.

bowd:
Ah one more question: do we want to add the new method to the contract interface?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -90,6 +92,9 @@ contract Attestations is
   // Maps a token and attestation issuer to the amount that they're owed.
   mapping(address => mapping(address => uint256)) public pendingWithdrawals;
 
+  // Maps approval of attesation transfer
+  mapping(address => mapping(bytes32 => bool)) public transferApprovals;
```
asaj, line 97: It's not clear what the second key is here, consider e.g.:
```
// Attestation transfer approvals, keyed by user and keccak(identifier, from, to)
```

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -649,6 +673,70 @@ contract Attestations is
...
+   * is executed.
+   * @param identifier The identifier for this attestation.
+   * @param from The current attestation address to which the identifier is mapped.
+   * @param to The new address to map to identifier.
+   * @param status The approval status
+   */
+  function approveTransfer(bytes32 identifier, uint256 index, address from, address to, bool status)
+    external
+  {
+    require(from != address(0) && to != address(0), "reserved address 0x0 cannot have approval");
```
asaj: I think it's okay to drop this check, as `0x0` will never be able to approve

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -649,6 +673,70 @@ contract Attestations is
...
+    bytes32 key = keccak256(abi.encodePacked(identifier, from, to));
```
asaj, line 692: Minor suggestion that would allow you to remove `isOtherAddressApproved`
```
    require(msg.sender == from || msg.sender == to);
    bytes32 key = keccak256(abi.encodePacked(identifier, from, to));
    address other = msg.sender == from ? to : from;
    if (status && transferApprovals[other][key]) {
      transferApprovals[other][key] = false;
      _transfer(identifier, index, from, to);
    } else {
      transferApprovals[msg.sender][key] = status;
      emit TransferApproval(msg.sender, from, to, status);
    }
```

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -649,6 +673,70 @@ contract Attestations is
...
+    if (
+      status && (isOtherAddressApproved(from, to, key) || isOtherAddressApproved(to, from, key))
+    ) {
+      _transfer(identifier, index, from, to);
+      transferApprovals[to][key] = false;
+      transferApprovals[from][key] = false;
+    } else {
+      transferApprovals[msg.sender][key] = status;
+      emit TransferApproval(msg.sender, from, to, status);
```
asaj: Should we include the identifier here?

packages/protocol/contracts/identity/Attestations.sol
```diff
@@ -649,6 +673,70 @@ contract Attestations is
...
+   * @dev Throws if `from` is not in the account array at the given index.
+   * @dev Throws if `to` already has attestations
+   */
+  function _transfer(bytes32 identifier, uint256 index, address from, address to) internal {
+    uint256 numAccounts = identifiers[identifier].accounts.length;
+    require(index < numAccounts, "Index is invalid");
+    require(from == identifiers[identifier].accounts[index], "Index does not match from address");
+    require(
+      identifiers[identifier].attestations[to].requested == 0,
+      "address tranferring to has already requested attestations"
```
asaj: nit: For consistency, capitalize

asaj (approved):

packages/protocol/contracts/identity/Attestations.sol
aslawson: ya -- that's a good way to do it.  i'll update

packages/protocol/contracts/identity/Attestations.sol
aslawson: good idea

packages/protocol/contracts/identity/Attestations.sol
asaj: I don't think this code works without the first `require`

PR #4585, opened 2020-07-31
[attestations] Delivery reports, per-country provider ordering, retry between providers, metrics

timmoreton:
### Description

A lot of changes to attestation service to better support multiple providers, add delivery receipts, and provide insight into failures.

* Nexmo is now in the default config
* Provider configs can be specified per country code like `SMS_PROVIDERS_US=twilio,nexmo`
* Interface for providers to support delivery receipts, and support added to receive them for Twilio and Nexmo 
* Better support for using multiple providers
* SMS send retry logic made generic for all providers:
** use country specific provider order, then filtered by unsupported regions
** retry send for each provider up to `` times with backoff (then moving on to next provider)
** if a delivery report comes back with a failure, that causes a retry to happen
* Renamed `BLACKLIST` config options params to `UNSUPPORTED_REGIONS` though existing configs are backwards compatible and will work unchanged 
* If `VERIFY_CONFIG_ON_STARTUP` option set, service startup now verifies account and signer key match, that metadata is available, and that attestation claim is made, to reduce chance of misconfiguration 
* Off-by-default `NEXMO_ACCOUNT_BALANCE_METRIC` option to expose Nexmo balance as a metric so it can be alerted on to prevent validators running out of funds
* New metrics:
** Counts by provider, country code, and delivery status: `attestation_attempts_delivery_status`
** Counts by provider, country code, and error code: `attestation_attempts_delivery_error_codes`
** Eventual delivery success (or timeout without a delivery report): `attestation_requests_believed_delivered_sms`
** Eventual delivery report of failure: `attestation_requests_failed_to_deliver_sms`
** `attestation_requests_unable_to_serve` now broken down by country code

### Tested

With phone and some test accounts. 

### Backwards compatibility

I think so!


packages/attestation-service/config/.env.development
```diff
@@ -1,18 +1,26 @@
 DATABASE_URL=sqlite://db/dev.db
 CELO_PROVIDER=http://localhost:8545
-CELO_VALIDATOR_ADDRESS=0xE6e53b5fc2e18F51781f14a3ce5E7FD468247a15
+CELO_VALIDATOR_ADDRESS=x
 ATTESTATION_SIGNER_ADDRESS=x
 APP_SIGNATURE=x
-SMS_PROVIDERS=twilio
+SMS_PROVIDERS=nexmo,twilio
+SMS_PROVIDERS_US=twilio,nexmo
+SMS_PROVIDERS_MX=nexmo
```
aslawson: where are these to be used?

packages/attestation-service/config/.env.development
```diff
@@ -1,18 +1,26 @@
...
+EXTERNAL_CALLBACK_HOSTPORT=http://host:3000
```
aslawson: can you update the README instructions to add these new configs and what they are for?

packages/attestation-service/config/.env.development
```diff
@@ -1,18 +1,26 @@
...
-TWILIO_BLACKLIST=
+TWILIO_UNSUPPORTED_REGIONS=
+MAX_PROVIDER_RETRIES=3
+TIMEOUT_CLEANUP_NO_RECEIPT_MIN=10
 # Options: json (default), human, stackdriver
 LOG_FORMAT= human
 # Options: fatal, error, warn, info, debug, trace
-LOG_LEVEL=
\ No newline at end of file
+LOG_LEVEL=
```
aslawson: I see it was already like this, but what does this default to?  Or does this require the server to update this?

packages/attestation-service/src/index.ts
```diff
@@ -3,26 +3,34 @@ import express from 'express'
...
   await initializeDB()
   await initializeKit()
-  // TODO: Validate that the attestation signer has been authorized by the account
-  getAttestationSignerAddress()
-  getAccountAddress()
-  await initializeSmsProviders()
+
+  // Verify configuration.
+  const verifyConfig = fetchEnvOrDefault('VERIFY_CONFIG_ON_STARTUP', '1').toLowerCase()
+  if (verifyConfig === '1' || verifyConfig === 'y') {
```
aslawson: do we expect someone might put 'y' instead?  what if they put 'true'?

aslawson (commented):
In progress-- but some initial questions before you leave

packages/attestation-service/src/db.ts
```diff
@@ -52,6 +54,41 @@ export async function isAttestationSignerUnlocked() {
   }
 }
 
+// Verify a signer and validator address are provided, are valid, match the on-chain signer,
+// that signer account is unlocked, that metadata is accessible and valid and that the
+// attestationServiceURL claim is present in the metadata (external name/port may be
+// different to instance, so we cannot validate its details)
```
aslawson, line 61: hmm -- perhaps we can still alert if external name/port differs

I think we'd more often run across instances where they forget to update their attestationServiceURL in the meta data rather than not add it at all.  event if these differ, at least it would serve as a "hey, these differ.  did you mean for them to?"

packages/attestation-service/src/db.ts
```diff
@@ -52,6 +54,41 @@ export async function isAttestationSignerUnlocked() {
   }
 }
 
+// Verify a signer and validator address are provided, are valid, match the on-chain signer,
```
aslawson, line 58: nice!

packages/attestation-service/src/db.ts
```diff
@@ -52,6 +54,41 @@ export async function isAttestationSignerUnlocked() {
...
+  const signerOnChain = await accounts.getAttestationSigner(validator)
+  if (!eqAddress(signerOnChain, signer)) {
+    throw Error(
+      `Different attestation signer in config ${signer} than on-chain ${signerOnChain} for ${validator}!`
+    )
+  }
+
+  // if (!(await isAttestationSignerUnlocked())) {
+  //   throw Error(`Need to unlock attestation signer account ${signer}`)
+  // }
```
aslawson: why commented out?

packages/attestation-service/config/.env.development
aslawson: nvm I see where the country code is parsed

packages/attestation-service/src/sms/twilio.ts
```diff
@@ -1,48 +1,88 @@
...
+import { receivedDeliveryReport } from './index'
 
 export class TwilioSmsProvider extends SmsProvider {
   static fromEnv() {
     return new TwilioSmsProvider(
       fetchEnv('TWILIO_ACCOUNT_SID'),
       fetchEnv('TWILIO_MESSAGING_SERVICE_SID'),
       fetchEnv('TWILIO_AUTH_TOKEN'),
-      readBlacklistFromEnv('TWILIO_BLACKLIST')
+      readUnsupportedRegionsFromEnv('TWILIO_UNSUPPORTED_REGIONS', 'TWILIO_BLACKLIST')
```
aslawson, line 20: Is TWILIO_BLACKLIST left here for for backwards compatability?

packages/attestation-service/src/index.ts
```diff
@@ -40,12 +48,23 @@ async function init() {
...
     createValidatedHandler(AttestationServiceTestRequestType, handleTestAttestationRequest)
   )
+
+  for (const p of smsProvidersWithDeliveryStatus()) {
+    const path = `/delivery_status_${p.type}`
+    rootLogger.info(
+      { url: deliveryStatusURLForProviderType(p.type) },
+      'Registered delivery status handler'
+    )
+    app.post(path, ...p.deliveryStatusHandlers(), handleAttestationDeliveryStatus(p.type))
```
aslawson, line 70: i'm trying to understand the flow to get this report data.  do these express post handlers occur sequentially?  so the flows would be something like this?:
nexmo -- we parse the JSON from the request and then act on delivery report via `receivedDeliveryReport`
twilio -- we parse the url sent, interpret it with twilio webhook, with result act on delivery report via `receivedDeliveryReport`

packages/attestation-service/src/index.ts
```diff
@@ -40,12 +48,23 @@ async function init() {
...
     createValidatedHandler(AttestationRequestType, handleAttestationRequest)
   )
   app.post(
     '/test_attestations',
+    express.json(),
     createValidatedHandler(AttestationServiceTestRequestType, handleTestAttestationRequest)
   )
+
+  for (const p of smsProvidersWithDeliveryStatus()) {
+    const path = `/delivery_status_${p.type}`
```
aslawson, line 65: it looks like we pass this url as a callback to Twilio, but how does it get triggered for the Nexmo case?

packages/attestation-service/src/sms/index.ts
```diff
@@ -17,31 +26,242 @@ export async function initializeSmsProviders() {
...
+
+// Maximum delivery attempts with any one provider
+const maxProviderAttempts = parseInt(fetchEnvOrDefault('MAX_PROVIDER_RETRIES', '3'), 10)
+
+const ongoingDeliveries: any = {}
+
+async function attemptToSendSms(delivery: SmsDelivery): Promise<SmsProviderType> {
+  if (delivery.deliveryId) {
+    ongoingDeliveries[delivery.deliveryId!] = delivery
+    delivery.deliveryId = null
```
aslawson: why reset to null? is this so for recursive retry (by means of error or triggered by receipt), we don't execute the `finallyBelievedDeliveredCallback` clean up timeout?  we are essentially relying on the clearling of `deliveryId` to determine if its been delivered or reattempted?  

can we be more explicit with `delivery.awaitingAttempt`?  set to false on method start, set to true after calling provider.sendsms(), reset to false if successful deliver receipt or recursive call that starts a new attempt

packages/attestation-service/src/sms/index.ts
```diff
@@ -17,31 +26,242 @@ export async function initializeSmsProviders() {
...
+      ongoingDeliveries.deliveryId = null
+      if (delivery.finallyBelievedDeliveredCallback) {
+        delivery.finallyBelievedDeliveredCallback()
+      }
+    } else if (deliveryStatus === DeliveryStatus.Failed) {
+      delete ongoingDeliveries[deliveryId]
+      ongoingDeliveries.deliveryId = null
+      setTimeout(() => {
+        void attemptToSendSms(delivery)
+      }, Math.pow(2, delivery.attemptsForThisProvider) * 3000)
```
aslawson, line 293: something to note in a comment maybe, that since this is exponential, it will surpass  the wait time for the `finallyBelievedDeliveredCallback` timeout at around the 8th attempt (since the attempt amount is configurable)

packages/attestation-service/src/index.ts
aslawson: If Nexmo is not suppose to have this then `supportsDeliveryStatus` should be false.  not sure why we would wait for delivery status changes if they cannot be updated

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -222,31 +245,28 @@ class AttestationRequestHandler {
...
+            void this.persistFinalAttesationResult(AttestationStatus.FAILED)
+          },
+          () => {
+            void this.persistFinalAttesationResult(AttestationStatus.COMPLETE)
+          }
+        )
         Counters.attestationRequestsSentSms.inc()
         await attestationRecord.update(
-          { status: AttestationStatus.SENT, smsProvider: provider.type },
+          { status: AttestationStatus.SENT, smsProvider: sentVia },
```
aslawson, line 236: it looks like this `AttestationStatus.SENT` would overwrite the persisted `AttestationStatus.FAILED` or `AttestationStatus.COMPLETE` from the callbacks above.

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -222,31 +245,28 @@ class AttestationRequestHandler {
...
-        await provider.sendSms(this.attestationRequest.phoneNumber, textMessage)
-        this.logger.info('Sent sms')
+        const sentVia = await startSendSms(
+          this.attestationRequest.phoneNumber,
+          textMessage,
+          () => {
+            void this.persistFinalAttesationResult(AttestationStatus.FAILED)
+          },
+          () => {
+            void this.persistFinalAttesationResult(AttestationStatus.COMPLETE)
```
aslawson: What is the different of AttestationStatus.COMPLETE/SENT? I don't think this service has any awareness of completion onchain now, but I'd assume that's what the implication for that state is.

Should this be SENT?

packages/attestation-service/src/requestHandlers/liveness.ts
```diff
@@ -9,10 +9,14 @@ export async function handleLivenessRequest(_req: express.Request, res: express.
       return
     }
 
+    if (await isNodeSyncing()) {
+      respondWithError(res, 504, ErrorMessages.NODE_IS_SYNCING)
+    }
+
     try {
       await isDBOnline()
```
aslawson, line 22: is this where we wanted to handle misconfigured db?  eg. check attestation table exists

packages/attestation-service/src/sms/index.ts
aslawson: or rename to `currentDeliveryAttemptId` 

or use methods:
isWaitingForDeliveryAttempt() {
return deliveryId != null;
}

just something a little bit more clear what is the role purpose for switching this between the id and null

aslawson (commented):
Really great.  I'm excited to see these changes in action.

packages/attestation-service/config/.env.development
nambrot: + Update the docs

packages/attestation-service/config/.env.development
jmrossy: Wonder if it'd be more convenient to pass in a JSON string with a country code to provider map instead of per-country overrides which could stack up

packages/attestation-service/src/index.ts
jmrossy: or yes

jmrossy (commented):
Mostly looks great! Please see @aslawson's comments

packages/attestation-service/config/.env.development
```diff
@@ -1,18 +1,26 @@
...
+PORT=3000
 NEXMO_KEY=x
 NEXMO_SECRET=x
-NEXMO_BLACKLIST=
+NEXMO_UNSUPPORTED_REGIONS=
+NEXMO_ACCOUNT_BALANCE_METRIC=0
 TWILIO_ACCOUNT_SID=x
 TWILIO_MESSAGING_SERVICE_SID=x
 TWILIO_AUTH_TOKEN=x
 TWILIO_ACCOUNT_SID=x
```
aslawson: duplicate to 3 lines up

packages/attestation-service/src/db.ts
timmoreton: It's not possible to do this, in general. The running process does not have visibility as to the public DNS name of the instance or of any load balancer in front of the instance

packages/attestation-service/src/sms/twilio.ts
timmoreton: yes

packages/attestation-service/config/.env.development
timmoreton: I considered that but these are env vars and the escaping etc would get messy --- short of changing the structure significantly

packages/attestation-service/src/db.ts
timmoreton: oops!

packages/attestation-service/src/sms/index.ts
timmoreton: They're two separate timeouts... the `finallyBelievedDeliveredCallback` only kicks in after a successful send, when we give up waiting for a delivery receipt reporting success and assume it's delivered ok anyway.  This timeout only kicks in after an attempt sending the SMS fails, before we retry. At that point no SMS is in flight.

packages/attestation-service/src/sms/index.ts
timmoreton: There was a bug here (line 120 should be `delete ...` to remove the mapping) which probably confused things, but I've also renamed it and added comments.

packages/attestation-service/src/requestHandlers/attestation.ts
timmoreton: `COMPLETE` was a pre-existing value, but I'm using it to indicate we got a "success" delivery notification

packages/attestation-service/src/requestHandlers/liveness.ts
timmoreton: I may do this in a separate PR

packages/attestation-service/config/.env.development
timmoreton: not sure what you mean?

packages/attestation-service/src/index.ts
timmoreton: You have to set it via the Nexmo web site. I'll talk about this in the docs

packages/attestation-service/src/index.ts
timmoreton: The flow is the same for both. Twilio and Nexmo will asynchronously callback to us via a POST to this endpoint. Then the handler parses that response, and calls `receivedDeliveryReport` which looks for an entry with a matching `deliveryId` in the `ongoingDeliveries` map. If it's a failure, this may cause us to retry delivery with the same provider (up to the retry limit) or after that with the next provider, until all retries are exhausted.

packages/attestation-service/src/requestHandlers/attestation.ts
timmoreton: I don't think this could happen because the record is locked and so the `FAILED` or `COMPLETE` could never get written before the `SENT` does?   But I don't love recording this stuff to the database that much. It's not really used for anything other than preventing replays that cause you to spend $$ sending the same SMS over and over -- yet we store a bunch of status in there too.

packages/attestation-service/config/.env.development
aslawson: oh this didn't post just the single line...and i didn't provide good explanation...

we have two `TWILIO_ACCOUNT_SID` parameters... its an old issues,  but I noted when trying to run and thought it could be remove with this pr

packages/attestation-service/src/index.ts
aslawson: I see 👍

packages/attestation-service/src/index.ts
aslawson: got it -- i guess the confusion is that with Twilio the handlers are `[bodyParser.urlencoded({ extended: false }), twilio.webhook()]` vs Nexmo `[bodyParser.json()]`.  

what exactly is happening with the Twilio handlers?  it suggests that Twilio will call this endpoint with an encoded url that the twilio.webhook() handler uses to retrieve the receipt.  is that a correct understanding?

packages/attestation-service/src/requestHandlers/attestation.ts
aslawson: but we are doing `await startSendSms()` before we update to `SENT`.  And `startSendSms()` doesn't return without triggering one of these callbacks.  

so I don't see how they wouldn't be set to `COMPLETE`/`FAILURE` and then overwritten as `SENT`

packages/attestation-service/src/sms/index.ts
aslawson: got it 👍

aslawson: thanks for explaining -- overall these function may benefit from breaking down each step into a separate function for future eng readability.

aslawson (approved):

packages/attestation-service/config/.env.development
timmoreton: oh yeah -- I'll remove one!

packages/attestation-service/src/index.ts
timmoreton: oh I see. `twilio.webhook()` is actually something totally separate - it allows you to auth the signature in the incoming requests

packages/attestation-service/src/requestHandlers/attestation.ts
timmoreton: because the callbacks need to wait to acquire the lock on the row

aslawson: oh i see -- earlier in this method we already acquire the lock.  my understanding is: if there is another attempt to update the row like the `COMPLETE`/`FAILURE` callbacks, it would just fail rather than wait for the unlock.  and even if it waited, then you'd be stuck waiting at `await startSendSms()` before being able to run ` await attestationRecord.update(SENT)`

the sequelize [documentation](https://sequelize.org/master/class/lib/transaction.js~Transaction.html) isn't explicit about this, but i've seen the error case with locked rows in the other lightweight ORM used for PGPNP.  

what are you seeing in practice?

packages/attestation-service/src/index.ts
aslawson: what happens after the incoming request gets parsed as a url?

packages/attestation-service/src/requestHandlers/attestation.ts
timmoreton: removed for now per discussion

packages/docs/operations-manual/attestation-service.md
```diff
@@ -0,0 +1,311 @@
+# Attestation Service
```
aslawson, line 1: we'll also need to update the README for the attestation-service to either point to this or contain this new information.  it currently includes information about old configuration https://github.com/celo-org/celo-monorepo/tree/master/packages/attestation-service

packages/docs/operations-manual/attestation-service.md
```diff
@@ -0,0 +1,311 @@
...
+
+Celo Validators are strongly encouraged to operate an [Attestation Service](https://github.com/celo-org/celo-monorepo/tree/master/packages/attestation-service) instance.
+
+The Attestation Service is part of the [Celo identity protocol](/celo-codebase/protocol/identity). It sends SMS on behalf of users to allow them to attest to having access to a phone number and to map that securely and privately to a Celo account. This is shown as Steps 3 and 4 in this diagram:
```
aslawson: broken link

aslawson: nit -- "and to  securely and privately map that phone number to a Celo account"

packages/docs/operations-manual/attestation-service.md
```diff
@@ -0,0 +1,311 @@
...
+
+To actually be able to send SMS, you need to create a messaging service under [Programmable SMS > SMS](https://www.twilio.com/console/sms/services). Provide the resulting `SID` in the `TWILIO_MESSAGING_SERVICE_SID` configuration variable.
+
+Now that you have provisioned your messaging service, you need to buy at least 1 phone number to send SMS from. You can do so under the `Numbers` option of the messaging service page. It is strongly recommended that you purchase at least a US (`+1`) number which seem to provide high delivery success rates.  If you purchase numbers in other locales, Twilio will intelligently select the best number to send each SMS.
+
+### Nexmo
+
+Click the balance in the top-left to go to [Billing and Payments](https://dashboard.nexmo.com/billing-and-payments), where you can add funds. It is strongly recommended that you enable auto-reload to avoid failing to deliver SMS when your funds are exhausted.
```
aslawson: lets also encourage they "Switch on low balance alerts"

packages/docs/operations-manual/attestation-service.md
```diff
@@ -0,0 +1,311 @@
...
+| `ATTESTATION_SIGNER_ADDRESS`     | Address of the Validator's attestation signer key  |
+| `SMS_PROVIDERS`                  | Comma-separated list of all enabled SMS providers, by order of preference. Can include `twilio`, `nexmo` |
+
+Optional environment variables:
+
+| Variable                       | Explanation    |
+|--------------------------------|-------------------------------------------------------------------------------------------------|
+| `PORT`                           | Port to listen on. Default `3000`. |
+| `SMS_PROVIDERS_<country>`        | Override to set SMS providers and order for a specific country code (e.g `SMS_PROVIDERS_MX=nexmo,twilio`) |
+| `MAX_PROVIDER_RETRIES`           | Number of retries (after first) when sending SMS before considering next provider Default `2`.  |
```
aslawson: seems like default is 3

packages/docs/operations-manual/attestation-service.md
```diff
@@ -0,0 +1,311 @@
...
+| `EXTERNAL_CALLBACK_HOSTPORT`     | Provide the external URL at which providers can attempt callbacks with delivery receipts. If not provided, defaults to the value retrieved from the validator metadata. |
+| `TIMEOUT_CLEANUP_NO_RECEIPT_MIN` | If a delivery report appears to be supported but is not received within this number of minutes, assume delivery success                                               |
```
aslawson, line 148: Default 5

packages/docs/operations-manual/attestation-service.md
```diff
@@ -0,0 +1,311 @@
...
+| `LOG_FORMAT`                     | One of `json`, `human`, `stackdriver`  |
+
+Twilio configuration options:
+
+| Variable                       | Explanation                                                     |
+| ------------------------------ | --------------------------------------------------------------- |
+| `TWILIO_ACCOUNT_SID`           | The Twilio account ID                                           |
+| `TWILIO_MESSAGING_SERVICE_SID` | The Twilio Message Service ID. Starts with `MG`                 |
+| `TWILIO_AUTH_TOKEN`            | The API authentication token                                    |
+| `TWILIO_UNSUPPORTED_REGIONS`   | Optional. A comma-separated list of country codes to not serve  |
```
aslawson: example?  is this suppose to be alpha country codes or phone number country codes?
"US" or "+1"

packages/docs/operations-manual/attestation-service.md
```diff
@@ -0,0 +1,311 @@
...
+celocli account:get-metadata $CELO_VALIDATOR_RG_ADDRESS
+```
+
+### Delivery Receipts
+
+Attestation Services supports Twilio and Nexmo delivery receipts so that these services can callback to provide delivery information. This triggers retries as needed, even between providers, and enables delivery success metrics to be logged.
+
+Nexmo requires manual configuration to enable this. Go to [Settings](https://dashboard.nexmo.com/settings), and under `Delivery Receipts`, enter the external URL of your Attestation Service appended by `/delivery_status_nexmo` -- for example `http://1.2.3.4:80/delivery_status_nexmo`. This should correspond to the URL printed when Attestation Service is started.
+
+There is no configuration necessary to enable Twilio delivery receipts. The Attestation Service uses the URL in the validator metadata (can be overridden with the `EXTERNAL_CALLBACK_HOSTPORT` configuration option) appended by `/delivery_status_twilio`, and supplies that to Twilio through its API.
```
aslawson, line 232: seems like it has a default setting of `http://host:3000`

externalURL = fetchEnvOrDefault('EXTERNAL_CALLBACK_HOSTPORT', claimURL)
^ they will need to actively be instructed to set `EXTERNAL_CALLBACK_HOSTPORT` to empty to be able to allow the `claimURL` to be used, or we should keep it empty by default

packages/docs/operations-manual/attestation-service.md
```diff
@@ -0,0 +1,311 @@
...
+
+  - `Upstream`: The SMS was passed to an upstream carrier.
+
+  - `Delivered`: A final delivery receipt was received indicating the SMS was succesfully delivered.
+
+- `attestation_attempts_delivery_error_codes`: Counter for delivery attempts made. Label `country` breaks down the count by country code. Label `provider` identifies the provider. Label `code` identifies the provider-specific error codes: see [Twilio error codes](https://www.twilio.com/docs/api/errors#3-anchor) and [Nexmo error codes](https://developer.nexmo.com/messaging/sms/guides/delivery-receipts#dlr-error-codes) for more details.
+
+Administrative metrics:
+
+- The `attestation_provider_balance` tracks the value of the balance of accounts at supported providers.  Label `provider` identifies the provider. This is currently only supported for Nexmo, and is off by default but can be enabled by setting `NEXMO_ACCOUNT_BALANCE_METRIC`. The metric is populated as a value in the account currency, e.g USD, and only once a successful SMS has been delivered by that provider.
```
aslawson, line 312: I'm still very unclear what `NEXMO_ACCOUNT_BALANCE_METRIC` even does...

packages/docs/operations-manual/attestation-service.md
timmoreton: 2 in the code. will update the config too

packages/docs/operations-manual/attestation-service.md
timmoreton: ditto

packages/docs/operations-manual/attestation-service.md
aslawson: also we should be clear that it needs to be added as a string 
"http://host:3000" <-- without the quotations it didn't work for me

packages/docs/operations-manual/attestation-service.md
timmoreton: clear now?

aslawson: yes!

PR #4587, opened 2020-07-31
Add MetaTransactionWallet contract

m-chrzan:
### Description

Adds a `MetaTransactionWallet` contract to enable a [subsidized attestation onboarding flow](https://docs.google.com/document/d/1wBFX6c1Tu4Dyxrx9JdKhmXPXXaFmVpHmP5bAAzmX6EU/edit#).

### Other changes
- Upgrades to solc 0.5.13 to allow for `chainId` opcode.
- Bumps SortedOracles and Reserve version numbers due the solc upgrade which changed the bytecode of those contracts

### Tested

Unit tests

### Related issues

None

### Backwards compatibility

Yes, but the solc upgrade causes minor bytecode changes to `LinkedList` libraries, `SortedOracles`, and `Reserve`


packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -0,0 +1,179 @@
...
+    // If the signer was the owner, onlyOwner functions would not be callable via
+    // meta-transactions.
+    _transferOwnership(address(this));
+  }
+
+  /**
+   * @notice Transfers control of the wallet to a new signer.
+   * @param _signer The address authorized to execute transactions via this wallet.
+   */
+  function setSigner(address _signer) public onlyOwner {
```
nambrot, line 72: Couldn't the signer just execute this function directly as well?

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -0,0 +1,205 @@
...
+   * @param value The CELO value to be sent with the transaction.
+   * @param data The data to be sent with the transaction.
+   * @return The return value of the transaction execution.
+   */
+  function _executeTransaction(address destination, uint256 value, bytes memory data)
+    private
+    returns (bytes memory)
+  {
+    nonce = nonce.add(1);
+    if (data.length > 0) require(Address.isContract(destination), "Invalid contract address");
```
aslawson: why is this requirement exclusive to the case where data > 0?

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -0,0 +1,205 @@
...
+  function executeMetaTransaction(
+    address destination,
+    uint256 value,
+    bytes calldata data,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external returns (bytes memory) {
+    address _signer = getMetaTransactionSigner(destination, value, data, nonce, v, r, s);
+    require(_signer == signer, "Invalid meta-transaction signer");
```
aslawson, line 171: I'm pretty confused how the signing works, it looks like we get a signer but don't use it.

aslawson (commented):
not sure if this is still wip (though you mentioned it was ready for review in slack), but i'm pretty lost in how the signer comes into play here.

asaj:
Not WIP, thanks for reviewing!

packages/protocol/contracts/common/MetaTransactionWallet.sol
asaj: The signer is the address that is authorized to call `executeTransaction` or sign meta transactions that can be executed by anyone via `executeMetaTransaction`

packages/protocol/contracts/common/MetaTransactionWallet.sol
asaj: Transaction data encodes smart contract function IDs and parameters. It doesn't make sense for data to be present if the destination address is not a smart contract

packages/protocol/contracts/common/MetaTransactionWallet.sol
asaj: Essentially you can think of it as the wallet contract's owner. The actual `owner` is the contract itself, so that `onlyOwner` functions can be executed by the signer via `executeTransaction` or `executeMetaTransaction`

packages/protocol/contracts/common/MetaTransactionWallet.sol
asaj: For posterity: They could, but per the comment above `_transferOwnership(this)` that would prevent `setSigner` from being callable via meta-transactions.

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -0,0 +1,256 @@
...
+   * @param destination The address to which the transaction is to be sent.
+   * @param value The CELO value to be sent with the transaction.
+   * @param data The data to be sent with the transaction.
+   * @return The return value of the transaction execution.
+   */
+  function executeTransaction(address destination, uint256 value, bytes memory data)
+    public
+    returns (bytes memory)
+  {
+    require(msg.sender == signer || msg.sender == owner(), "Invalid transaction sender");
```
asaj, line 191:
```suggestion
    // Allowing the owner to call execute transaction allows, when the contract is self-owned,
    // for the signer to sign and execute a batch of transactions via a meta-transaction.
    require(msg.sender == signer || msg.sender == owner(), "Invalid transaction sender");
```

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  /**
+   * @notice Executes a transaction on behalf of the signer.`
+   * @param destination The address to which the transaction is to be sent.
+   * @param value The CELO value to be sent with the transaction.
+   * @param data The data to be sent with the transaction.
+   * @return The return value of the transaction execution.
+   */
+  function _executeTransaction(address destination, uint256 value, bytes memory data)
```
asaj: Share this code with `MultiSig` and (eventually) `Proposals` by breaking out into a library.

asaj (approved):

PR #4593, opened 2020-07-31
Forbid LinkedList contract changes

asaj:
### Description

The version checking script does not currently handle linked libraries well. Changes will be detected for the library, but not for any contract that links that library. This PR ensures no one tries to make changes to (a subset of the) linked libraries until this is fixed.

This has the added benefit of allowing us to upgrade to solidity 0.5.12, which introduces some minor bytecode differences in the various linked list libraries.

Long term, we might consider ignoring bytecode changes for linked libraries, and only triggering upgrades when minor or greater version changes are made.

## Other changes

None

### Tested

No

### Related issues

None

### Backwards compatibility

Yes


packages/protocol/scripts/bash/check-versions.sh
```diff
@@ -38,5 +38,5 @@ yarn build:sol
 rm -rf $BUILD_DIR_2 && mkdir -p $BUILD_DIR_2
 mv build/contracts $BUILD_DIR_2
 
-CONTRACT_EXCLUSION_REGEX=".*Test.*|I[A-Z].*|ReleaseGold"
+CONTRACT_EXCLUSION_REGEX=".*Test.*|I[A-Z].*|ReleaseGold|.*LinkedList.*"
```
hbandura, line 41: Why exclude them from the report list? Is it necessary?

asaj: This will allow us to upgrade to solidity 0.5.12 so that we can use the `chainId` instruction for the MetaTransactionWallet. Unfortunately, the upgrade to 0.5.12 results in some very minor bytecode changes to these libraries, that we're happy to ignore so long as the source code doesn't change.

m-chrzan (approved):

PR #4598, opened 2020-07-31
[Wallet] Don't reset pin on onboarding

gnardini:
### Description

When the user goes back from the `EnterInviteCode` or `ImportWallet` screens, we are resetting the PIN so that the user has to set it again when he goes through the onboarding process again. What's happening is that this PIN reset happens even if the user just completes the process normally and doesn't actually go back. In that case, `navigateHome` is called, which does a `reset` of the navigation and that apparently triggers the `beforeRemove` listener

### Tested

Locally on the simulator.


jeanregisser:
Oh interesting, I had this specific case in mind when I did this. The onFocusEffect returns a clean up function which should be called when the screen looses focus. This way the beforeRemove event should never be called if the screen is not focused.
With what you said, it looks like this assumption is wrong 🤔

packages/mobile/src/onboarding/UseBackToWelcomeScreen.tsx
```diff
@@ -52,9 +52,12 @@ export function useBackToWelcomeScreen({ backAnalyticsEvent }: Props) {
     // Check if this screen is removed while focused (hardware back button, swipe, or simply navigate back)
     // so we can reset some state
     // See https://reactnavigation.org/docs/preventing-going-back
-    const cancelBeforeRemove = navigation.addListener('beforeRemove', (e) => {
+    const cancelBeforeRemove = navigation.addListener('beforeRemove', (event) => {
       ValoraAnalytics.track(backAnalyticsEvent)
```
jeanregisser: This one shouldn't be called if the action is not dispatched.

gnardini: Yup, my bad!

gnardini:
@jeanregisser The clean up function is being called, but after the `beforeRemove` listener. It makes sense to me, because if the user presses on the `continue without` button the screen is still focused when the `navigateHome` method is called.

cmcewen (approved):
nice!

PR #4599, opened 2020-08-01
Add network events to mobile Analytics package

nategraf:
### Description

In order to get a better picture for how long users are waiting to become synced to the network, this PR added a number of events around the connection and sync status of the mobile app to the Celo network.

New events included here are:

  * `network_connected`
  * `network_disconnected`
  * `network_sync_waiting`
  * `network_sync_start`
  * `network_sync_finish`
  * `network_sync_error`
  * `network_sync_lost`
  * `network_sync_restored`

Additionally, this PR changes syncing logic to consider any block older than 60 seconds to be stale, and wait to catch up. This is important because it is possible to be behind the head of the chain, but not syncing. This matches the logic used in the `celocli` https://github.com/celo-org/celo-monorepo/blob/17c47cf494b072fc96cad2e530c8650292f27444/packages/cli/src/utils/helpers.ts#L21

It is important to catch this edge case because if a transaction is submitted while not syncing, it will not get a receipt and will wait 5 minutes then report the transaction failed. Or if the client starts to sync while waiting for a receipt, and the chain advances more than 50 blocks, the error `Transaction was not mined within 50 blocks` will be reported. It's possible this is related to https://github.com/celo-org/celo-blockchain/issues/1111, but the only evidence here is that _if_ this is happening, the error would be the same.

In order to track sync status, known chain head information is added to the redux state. This information could potentially be leveraged for other uses, but at the moment is only used to report events to the analytics platform.

### Other changes

* Add a `create_geth_error` event to track when the app fails to create a Geth instance.
* Add extra data fields to various transaction events to enable better tracking

### Tested

Unit test for the syncing logic change. Otherwise untested.

### Backwards compatibility

100%


packages/mobile/src/web3/saga.ts
```diff
@@ -44,12 +44,25 @@ const MNEMONIC_BIT_LENGTH = MnemonicStrength.s256_24words
 export const SYNC_TIMEOUT = 2 * 60 * 1000 // 2 minutes
 const SWITCH_TO_FORNO_TIMEOUT = 15000 // if syncing takes >15 secs, suggest switch to forno
 const WEB3_MONITOR_DELAY = 100
+const BLOCK_AGE_LIMIT = 5 * 60 // if the latest block is older than 5 minutes, the node is not synced.
```
annakaz: Why 5 minutes? I would have expected this to be the block time of a single block (5 seconds)

packages/mobile/src/web3/saga.ts
```diff
@@ -44,12 +44,25 @@ const MNEMONIC_BIT_LENGTH = MnemonicStrength.s256_24words
...
+}
+
+enum SyncStatus {
+  UNKNOWN,
+  WAITING,
+  SYNCING,
+}
 
 // checks if web3 claims it is currently syncing and attempts to wait for it to complete
 export function* checkWeb3SyncProgress() {
```
annakaz, line 55: One thing to note about `checkWeb3SyncProgress()` is that analytics here may not capture all syncing events, as it is only called when we want to wait for the node to initially sync in onboarding or when the app is first opened (specifically, importing a backup phrase, redeeming an invite, or when the web3Saga is first spawned). It would not capture syncing as someone is using the app after these initializations

For most actions that require the node to be synced, `getConnectedAccount()` is used to wait for geth connectivity, which ensures that geth is connected based on monitoring blocks here: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/geth/saga.ts#L189

annakaz (commented):
Thanks for adding these events! Will be great to have data on this. 

Added a comment on where best to instrument block sync times- I think that where you've put them in web3 will capture the most helpful data (how long syncing takes when app is first started), but if you are expecting all sync data I believe events will have to be tracked in monitorGeth() as well

If the priority is to get this in to get data as soon as possible, open to having this merged earlier and adding more detailed geth monitoring in a future PR

packages/mobile/src/web3/saga.ts
```diff
@@ -62,28 +75,43 @@ export function* checkWeb3SyncProgress() {
...
       } else {
         throw new Error('Invalid syncProgress type')
       }
       yield delay(WEB3_MONITOR_DELAY) // wait 100ms while web3 syncs then check again
       syncLoops += 1
+
+      // If sync has been in progrees for too long, prompt the user to try forno instead.
       if (syncLoops * WEB3_MONITOR_DELAY > SWITCH_TO_FORNO_TIMEOUT) {
         if (yield select(promptFornoIfNeededSelector) && features.DATA_SAVER) {
+          ValoraAnalytics.track(NetworkEvents.network_sync_error, { context: 'sync timeout' })
```
tarikbellamine: You can certainly leave this as is but might be worth considering using only an `error` property instead of `context` + `error`. This will make things more simple when you eventually want to aggregate and analyze network sync errors

packages/mobile/src/web3/saga.ts
nategraf: I agree this is an overly flexible limit. I've changed this to 30 seconds, which is plenty of margin that I can't imagine any effects of network delay or slow processing on the phone, but restrictive enough to prevent the potential for transaction not mined errors.

packages/mobile/src/web3/saga.ts
nategraf: I've gone ahead and added some state around the chain head and analytics for events of losing or regaining sync. We could also use this to change from waiting for a connected account to send a transaction, to waiting for a synced account, which may provide better transaction reliability by eliminating cases where we are sending transaction while connected, but not synced. Since we have no idea how common this is, I didn't go that far. I think the analytics will be useful to decide if there is anything to be done here.

annakaz (approved):
LGTM!

PR #4600, opened 2020-08-01
[WIP] AttestationSubsidy.sol

bowd:
### Description

The `AttestationSubsidy.sol` is part of the [subsidies attestation fee proposal](https://docs.google.com/document/d/1wBFX6c1Tu4Dyxrx9JdKhmXPXXaFmVpHmP5bAAzmX6EU/edit), where it is referred to as the Onboarding contract. I choose a more explicit name.

The purpose of the contract is to batch 3 actions that need to happen as part of a single transaction:
1. The MetaTransactionWallet sets a cUSD allowance for the Attestation contract (executed as a meta-tx)
2. Transfer cUSD from the AttestationSubsidy to the MetaTransactionWallet.
3. Request attestations as the MetaTransactionWallet (executed as a meta-tx)

In order to achieve this the contract needs the calldata and signatures for the two meta-transactions (1) and (2). In order to achieve this we have two options:

a) Pass the respective calldatas as arguments to the function and forward them directly
b) Pass in the required variables and construct the calldata dynamically.

I initially explored (a) but realised that there's additional logic required to enforce that the meta-transactions that we're forwarding are what we expect them to be. We would need therefor to extract method ids and parameters and validated that (1) is indeed an `approve` and (3) an `request`. We would also need to deconstruct the calldata to extract `requestedAttestations` in order to compute the totalFee that needs to be subsidised.

In the end I chose (b) which is a bit more explicit and limiting, but it's safer because we're building calldata and setting destinations on-chain, and only get the signatures from off-chain, and these need to match the meta-txs we're constructing on-chain.

### Tested

In progress - tests currently incomplete/failing.

### Related issues

The branch and PR are currently based on #4587 because I wanted to have the MetaTransactionWallet interface in scope. And I opened the PR relative to that branch (and not master) so that the diff is easier to read. When we merge #4587 in master, I'll rebase this branch on master and change the PR destination to be `master`.

### DIFF
--- a/packages/protocol/contracts/identity/AttestationSubsidy.sol
+++ b/packages/protocol/contracts/identity/AttestationSubsidy.sol
@@ -51,8 +51,8 @@ contract AttestationSubsidy is
   }
 
   /**
-   * @notice Updates 'maxSubsidiesAttestations'.
-   * @param _maxSubsidiesAttestations Maximum number of attestations that can be subsidies in one request
+   * @notice Updates 'maxSubsidisedAttestations'.
+   * @param _maxSubsidisedAttestations Maximum number of attestations that can be subsidised in one request
    */
   function setMaxSubsidisedAttestations(uint256 _maxSubsidisedAttestations) public onlyOwner {
     require(_maxSubsidisedAttestations > 0, "maxSubsidisedAttestations has to be greater than 0");
--- a/packages/protocol/contracts/identity/AttestationSubsidy.sol
+++ b/packages/protocol/contracts/identity/AttestationSubsidy.sol
@@ -31,29 +31,50 @@ contract AttestationSubsidy is
   using SafeMath for uint256;
   using SafeCast for uint256;
 
-  event AttestationSubsidised(address indexed account, uint256 value);
+  // Limit the maximum number of subsidised attestations that can be requested
+  uint256 public maxSubsidisedAttestations;
 
+  event AttestationSubsidised(address indexed account, uint256 value);
+  event MaxSubsidiesAttestationsSet(uint256 value);
   /**
-     * @notice Used in place of the constructor to allow the contract to be upgradable via proxy.
-     * @param registryAddress The address of the registry core smart contract.
-     */
-  function initialize(address registryAddress) external initializer {
-    setRegistry(registryAddress);
+    * @notice Used in place of the constructor to allow the contract to be upgradable via proxy.
+    * @param registryAddress The address of the registry core smart contract.
+    * @param maxSubsidiesAttestations Maximum number of attestations that can be subsidies in one request
+    */
+  function initialize(address registryAddress, uint256 maxSubsidisedAttestations)
+    external
+    initializer
+  {
     _transferOwnership(msg.sender);
+    setRegistry(registryAddress);
+    setMaxSubsidisedAttestations(maxSubsidisedAttestations);
   }
 
   /**
-     * @notice Used to batch three operations:
-     * - Approve cUSD transfer between the beneficiary meta-wallet and the attestations contract
-     * - Transfer the cUSD subsidy to the meta-wallet
-     * - Execute the attestation request to spend the cUSD subsidy
-     * @param beneficiaryMetaWallet - the address of the beneficiary meta-wallet
-     * @param identifier - the identifier to request attestations for (see Attestations.sol)
-     * @param attestationsRequested - the number of attestations requested (see Attestations.sol)
-     * @param v,r,s - array of signatures:
-     *    -- [0] = signatures for cUSD.approve(Attestations.sol, attestationsRequested*fee)
-     *    -- [1] = signatures for Attestations.request(identifier, attestationsRequested, cUSD)
-     */
+   * @notice Updates 'maxSubsidiesAttestations'.
+   * @param _maxSubsidiesAttestations Maximum number of attestations that can be subsidies in one request
+   */
+  function setMaxSubsidisedAttestations(uint256 _maxSubsidisedAttestations) public onlyOwner {
+    require(_maxSubsidisedAttestations > 0, "maxSubsidisedAttestations has to be greater than 0");
+    maxSubsidisedAttestations = _maxSubsidisedAttestations;
+    emit MaxSubsidiesAttestationsSet(_maxSubsidisedAttestations);
+  }
+
+  /**
+    * @notice Used to batch three operations:
+    * - Approve cUSD transfer between the beneficiary meta-wallet and the attestations contract
+    * - Transfer the cUSD subsidy to the meta-wallet
+    * - Execute the attestation request to spend the cUSD subsidy
+    * @param beneficiaryMetaWallet The address of the beneficiary meta-wallet
+    * @param identifier The identifier to request attestations for (see Attestations.sol)
+    * @param attestationsRequested The number of attestations requested (see Attestations.sol)
+    * @param v Array of signature components `v` for the meta-txs
+    * @param r Array of signature components `r` for the meta-txs
+    * @param s Array of signature components `s` for the meta-txs
+    * @dev The signature component arrays should have 2 items:
+    *   (v,r,s)[0] = signature for cUSD.approve(Attestations.sol, attestationsRequested*fee)
+    *   (v,r,s)[1] = signature for Attestations.request(identifier, attestationsRequested, cUSD)
+    */
   function requestAttestationsWithSubsidy(
     address beneficiaryMetaWallet,
     bytes32 identifier,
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/identity/interfaces/IAttestationSubsidy.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity ^0.5.3;
+
+interface IAttestationSubsidy {
+  function requestAttestationsWithSubsidy(
```
bowd, line 4: Forgot to change the interface from a previous version 🤦

packages/protocol/contracts/identity/AttestationSubsidy.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+     * @notice Used to batch three operations:
+     * - Approve cUSD transfer between the beneficiary meta-wallet and the attestations contract
+     * - Transfer the cUSD subsidy to the meta-wallet
+     * - Execute the attestation request to spend the cUSD subsidy
+     * @param beneficiaryMetaWallet - the address of the beneficiary meta-wallet
+     * @param identifier - the identifier to request attestations for (see Attestations.sol)
+     * @param attestationsRequested - the number of attestations requested (see Attestations.sol)
+     * @param v,r,s - array of signatures:
```
m-chrzan: Consider "arrays of signature components" to be more accurate.

packages/protocol/contracts/identity/AttestationSubsidy.sol
```diff
@@ -0,0 +1,135 @@
...
+     *    -- [0] = signatures for cUSD.approve(Attestations.sol, attestationsRequested*fee)
```
m-chrzan: Similarly, signatures -> signature here and below.

packages/protocol/contracts/identity/AttestationSubsidy.sol
```diff
@@ -0,0 +1,135 @@
...
+    emit AttestationSubsidised(beneficiaryMetaWallet, totalFee);
+  }
+
+  /**
+     * @notice executes the cUSD approval from the metaWallet to the Attestations contract
+     * @param metaWallet - the wallet in question
+     * @param totalFee - the total amount of cUSD to be approved
+     * @param v,r,s - the signature for the meta-transaction
+     */
+  function _metaApproveCUSD(
```
m-chrzan: I don't think this and other functions need to be `_`-prefixed. The convention as I see it is to use `_` in cases where there is a public wrapper around an internal function where both could have the same name.

packages/protocol/contracts/identity/AttestationSubsidy.sol
```diff
@@ -0,0 +1,135 @@
...
+    uint256 attestationsRequested,
+    uint8[] calldata v,
+    bytes32[] calldata r,
+    bytes32[] calldata s
+  ) external onlyOwner nonReentrant {
+    require(v.length == 2, "two signatures are required (approve,request)");
+    require(r.length == 2, "two signatures are required (approve,request)");
+    require(s.length == 2, "two signatures are required (approve,request)");
+
+    uint256 totalFee = _calculateTotalFee(attestationsRequested);
```
m-chrzan: Thoughts about limiting the max `attestationsRequested`? Ideally that will happen in the onboarding service as well, but wonder if a sanity check would also be good.

packages/protocol/contracts/identity/AttestationSubsidy.sol
```diff
@@ -0,0 +1,135 @@
...
+    IMetaTransactionWallet metaWallet = IMetaTransactionWallet(beneficiaryMetaWallet);
+    IERC20 cUSD = IERC20(address(getStableToken()));
+
+    _metaApproveCUSD(metaWallet, totalFee, v[0], r[0], s[0]);
+    cUSD.transfer(beneficiaryMetaWallet, totalFee);
+    _metaRequestAttestations(metaWallet, identifier, attestationsRequested, v[1], r[1], s[1]);
```
m-chrzan: Just noting that the onboarding service *needs* to be sure that the meta wallet address passed in here is in fact a deployed MetaTransactionWallet.

packages/protocol/contracts/identity/AttestationSubsidy.sol
```diff
@@ -0,0 +1,135 @@
...
+     * @param v,r,s - the signature for the meta-transaction
+     */
+  function _metaApproveCUSD(
+    IMetaTransactionWallet metaWallet,
+    uint256 totalFee,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) internal {
+    bytes memory callData = abi.encodeWithSignature(
```
m-chrzan, line 168: Hmm, so now I'm wondering. Initially I had the same intuition, that we need to build calldata here. But this contract can only be called by `owner` anyways, and we're already forced to do some off-chain checking. By the same token, maybe it's fine to submit the already constructed data here, and have the onboarding service construct it off-chain?

packages/protocol/contracts/identity/AttestationSubsidy.sol
```diff
@@ -0,0 +1,135 @@
...
+    setRegistry(registryAddress);
+    _transferOwnership(msg.sender);
+  }
+
+  /**
+     * @notice Used to batch three operations:
+     * - Approve cUSD transfer between the beneficiary meta-wallet and the attestations contract
+     * - Transfer the cUSD subsidy to the meta-wallet
+     * - Execute the attestation request to spend the cUSD subsidy
+     * @param beneficiaryMetaWallet - the address of the beneficiary meta-wallet
```
m-chrzan: No dash after @param argument, here and throughout.

packages/protocol/contracts/identity/AttestationSubsidy.sol
m-chrzan: Also, not sure how/if NatSpec handles multiple names in a @param like this. This is definitely nicer from a readability standpoint.

packages/protocol/contracts/identity/AttestationSubsidy.sol
```diff
@@ -0,0 +1,135 @@
...
+  ReentrancyGuard,
+  UsingPrecompiles
+{
+  using SafeMath for uint256;
+  using SafeCast for uint256;
+
+  event AttestationSubsidised(address indexed account, uint256 value);
+
+  /**
+     * @notice Used in place of the constructor to allow the contract to be upgradable via proxy.
```
m-chrzan: Nit: the `*` here should be aligned with the first `*` above, here and throughout.

packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -8,7 +8,7 @@ const MetaTransactionWallet: MetaTransactionWalletContract = artifacts.require(
   'MetaTransactionWallet'
 )
 
-interface MetaTransaction {
+export interface MetaTransaction {
```
m-chrzan, line 11: Probably shouldn't export from tests file - put in a new utils file in `lib/` instead, for both tests to use. For this and the exports below.

packages/protocol/test/identity/attestationsubsidy.ts
```diff
@@ -0,0 +1,307 @@
...
+  let mockStableToken: MockStableTokenInstance
+  let otherMockStableToken: MockStableTokenInstance
+  let random: MockRandomInstance
+  let mockElection: MockElectionInstance
+  let mockLockedGold: MockLockedGoldInstance
+  let registry: RegistryInstance
+  let callerMetaWallet: MetaTransactionWalletInstance
+
+  const provider = new Web3Class.providers.HttpProvider('http://localhost:8545')
+  const web3: Web3 = new Web3Class(provider)
```
m-chrzan, line 79: Truffle should already make a `web3` instance available, should be able to just remove these two lines.

packages/protocol/test/identity/attestationsubsidy.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  const phoneHash: string = getPhoneHash(phoneNumber)
+
+  const attestationsRequested = 3
+  const attestationExpiryBlocks = (60 * 60) / 5
+  const selectIssuersWaitBlocks = 4
+  const maxAttestations = 20
+  const attestationFee = new BigNumber(web3.utils.toWei('.05', 'ether').toString())
+
+  // async function getVerificationCodeSignature(
```
m-chrzan, line 105: Remember to remove commented out code.

packages/protocol/test/identity/attestationsubsidy.ts
```diff
@@ -0,0 +1,307 @@
...
+          selectIssuersWaitBlocks,
+          maxAttestations,
+          [mockStableToken.address],
+          [attestationFee]
+        )
+      )
+    })
+  })
+
+  describe('#requestAttestationsWithSubsidy', () => {
```
m-chrzan, line 232: Would be also nice to see some negative cases with badly formed signature/inputs/etc.

packages/protocol/contracts/identity/AttestationSubsidy.sol
bowd: oh yeah, I don't know what I was thinking. I'll revert to something machine readable 🤖

packages/protocol/contracts/identity/AttestationSubsidy.sol
bowd: I was considering that this already happens [here](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/identity/Attestations.sol#L193). Which would be executed in the last meta-tx. I guess checking that earlier would save in gas cost if a revert happens due to it, and ofc sanity checking that on the onboarding service side would be ideal. Do you see any other reason why I should check that in the contract here as well?

packages/protocol/contracts/identity/AttestationSubsidy.sol
bowd: Yes, I forgot to mention this in the description but I also realised that this would be exploitable unless we know that the beneficiaryMetaWallet address is indeed a MetaTransactionWallet. This is mitigated by the fact that our onboarding service will also be responsible for deploying these wallets in the current design. But we should somehow highlight this. And make that invariant explicit.

One idea I got when I was at the market just now is to check that the number of requested attestations for that identifier actually increases as a result of the meta-tx call and revert if not. I think that's an useful extra invariant which means that we can care less of the `metaWallet` is behaving as we expect it to as long as it has the expected outcome: more attestations requested.

packages/protocol/contracts/identity/AttestationSubsidy.sol
m-chrzan: So that max is 100 on mainnet right now, I believe (number of validators, since you can't get more unique attestations than that). A lower constraint here would be to limit downside in case of any dev errors/exploits. But in general don't feel too strongly about this.

packages/protocol/test/identity/attestationsubsidy.ts
bowd: Yeah tests are WIP didn't get any to pass yet and I copied the attestations test setup as a starting point and I'm not yet sure what I will actually need from that and what I won't, still wrapping my head around that test setup.

packages/protocol/test/identity/attestationsubsidy.ts
bowd: Yes I want to add more, tests are very much WIP.

packages/protocol/test/identity/attestationsubsidy.ts
bowd: Interesting. That's how it's setup in the `attestations.ts` tests file. I'll try removing it from there as well.

packages/protocol/contracts/identity/AttestationSubsidy.sol
m-chrzan: > One idea I got when I was at the market just now is to check that the number of requested attestations for that identifier actually increases as a result of the meta-tx call and revert if not. I think that's an useful extra invariant which means that we can care less of the `metaWallet` is behaving as we expect it to as long as it has the expected outcome: more attestations requested.

Yep, something like that could work as well!

packages/protocol/contracts/identity/AttestationSubsidy.sol
bowd: Yeah we could save on gas with this we construct the calldata off-chain, but then we have two options:
1. Also provide the `attesatationRequests` & `identifier` as an argument in order to use them for extra checks
2. Extract `attesatationRequests` & `identifier` from the calldata encoded off-chain 

If we do (1) we have two bits of data which we _think_ should be connected but aren't necessarily. We would treat the calldata as if it has the same `attestationRequests` value encoded in it as the one passed as arguments, but it would be an implicit assumption.

If we do (2) it will probably end up costing the same to extract variables from the encoded calldata. Well actually the assembly `mload` thing If probably cheaper the constructing on-chain.

asaj (commented):
This feels a bit heavyweight. I wonder if it might be more useful (and more general) to add 
 `executeTransactions` and `executeMetaTransactions` to `MetaTransactionWallet`?

Then, the on-boarding service could call `executeTransactions([cUSD transfer, execute user's approval meta tx, execute user's request issuers tx])`

asaj:
For posterity, @bowd and I discovered an attack on the above suggestion, in which the user upgrades their `MetaTransactionWallet` to treat meta-transactions sent from the on-boarding service as no-ops. This will cause the cUSD transfer to succeed but the approval and attestations requests to be ignored.

The solution we settled on was adding an `Attestations.revertUnlessNAttestationsRequested` function, which could be called as the last transaction in the batch to confirm that the cUSD was used for attestations. It could even be added as a first transaction in the batch as well, to be extra safe.

PR #4607, opened 2020-08-03
Do not let a ReleaseGold contract self-destruct with a cUSD balance

### LINKED ISSUE #4571: celocli should prevent withdrawing all Celo from ReleaseGold if cUSD balance != 0
### Expected Behavior

### Current Behavior

What is the current behavior?

nambrot:
### Description

This PR prevents CLI users from withdrawing their remaining balance if they have a cUSD balance to avoid loss of funds when the contract self-destructs

### Tested

- Unit test

### Related issues

- Fixes #4571

### DIFF
--- a/packages/cli/src/commands/releasegold/withdraw.test.ts
+++ b/packages/cli/src/commands/releasegold/withdraw.test.ts
@@ -5,6 +5,7 @@ import { getContractFromEvent, testWithGanache, timeTravel } from '@celo/dev-uti
 import Web3 from 'web3'
 import CreateAccount from './create-account'
 import SetLiquidityProvision from './set-liquidity-provision'
+import RGTransferDollars from './transfer-dollars'
 import Withdraw from './withdraw'
 
 process.env.NO_SYNCCHECK = 'true'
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/releasegold/withdraw.test.ts
```diff
@@ -35,4 +36,40 @@ testWithGanache('releasegold:withdraw cmd', (web3: Web3) => {
     const balanceAfter = await kit.getTotalBalance(beneficiary)
     await expect(balanceBefore.gold.toNumber()).toBeLessThan(balanceAfter.gold.toNumber())
   })
+
+  test('cant withdraw the whole balance if there is a cUSD balance', async () => {
```
m-chrzan: Typo: *can't*

packages/cli/src/commands/releasegold/withdraw.ts
```diff
@@ -39,6 +39,20 @@ export default class Withdraw extends ReleaseGoldCommand {
       .addCheck('Contract has met liquidity provision if applicable', () =>
         this.releaseGoldWrapper.getLiquidityProvisionMet()
       )
+      .addCheck(
+        'Contract does not self-distruct with cUSD left when withdrawing the whole balance',
```
m-chrzan: What exactly is the string here, will it be presented as an error to the user? Should this be reworded to something like "Contract would self-destruct"...

Also typo, distruct -> destruct.

gastonponti (approved):

PR #4608, opened 2020-08-03
[contractKit] Balance rename gold to celo

gastonponti:
### Description

Rename the balance response of the kit
gold -> CELO
usd -> cUSD
lockedGold -> lockedCELO

### Backwards compatibility

The function's response is not the same. The type changed to have those keys. It's not backwards compatible (only that function response)


mcortesi (approved):
I guess this imply a major version :P 
non-backward compatible change...

PR #4615, opened 2020-08-03
[Wallet] Account key UI changes for better visibility on small screens

### LINKED ISSUE #4565: Backup Key Screen Contrast Improvements
### Expected Behavior

based on dry run feedback, low readability:
* increase contrast between account key section & instructions
* explicitly add a label to the account and/or instruction sections
* consider formatting the account key in a similar way to the quiz
* increase size of account key words

### Current Behavior

What is the current behavior?

annakaz:
### Description

Update copy and UI in backup phrase flow to be easier to see on a small screen. This is an improvement that came from feedback from call center onboarding sessions for the Grameen pilot.  

Following mocks here https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=2995%3A33668

Screen Shot 2020-08-03 at 3 47 02 PM
Screen Shot 2020-08-03 at 3 44 14 PM

### Tested

On an emulator with a small screen

### Related issues

- Fixes #4565 

### Backwards compatibility

Yes, just copy and UI changes


aaronmgdr (approved):
Looks Legit

tarikbellamine (commented):
Changes look great! If you're up for it, another small feature I believe we want to add on this screen is the ability to copy the account key. The designs don't currently reflect this but I think we want a copy icon near the account key, though best to check with @coreycelo about it.

annakaz:
Thanks for mentioning that @tarikbellamine ! Will leave it out of this PR for now as we may want to look more into the security implications if we decide to allow users to copy the full seed phrase to clipboard

PR #4617, opened 2020-08-04
[ODIS] Trim BLS Key

codyborn:
### Description

If the BLS key share was uploaded with whitespace, then the 72 char length check will fail. Remove any whitespace prior to use.

### Tested

Unit tests

### Backwards compatibility

Yes


jmrossy (approved):

packages/phone-number-privacy/signer/src/key-management/key-provider-base.ts
```diff
@@ -19,6 +19,7 @@ export abstract class KeyProviderBase implements KeyProvider {
   public abstract async fetchPrivateKeyFromStore(): Promise<void>
 
   protected setPrivateKey(key: string) {
+    key = key.trim()
```
jmrossy: It would be good to check that key isn't null before calling trim

jmrossy (approved):

PR #4619, opened 2020-08-04
Set contractkit/underscore version to 1.9.1

hbandura:
Fix underscore version to the same version web3 uses, to reduce dependency bloating


cmcewen:
hi @hbandura! are you looking at the monorepo version of contractkit or the published version? i don't believe we import underscore in contractkit, though i do see a single usage of it here: https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/lib/signing-utils.ts. i think it's possible we could replace that file with this (https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/utils/signing-utils.ts) and then we can completely remove underscore from our dependency tree

hbandura:
@cmcewen This change may or may not be included in contractkit, I created the PR to see if it passed all tests correctly. Both the usage you mention and web3 use underscore. If you create a new npm package that depends on contractkit, and then bundle it, it will end up having around ten copies of the underscore library, since at the flat level it will use 1.10 (or something like that, it actually requests >1.4), but all the uses from web3 packages use 1.9. By forcing it to be 1.9, then only one copy of underscore gets bundled.

This is actually something that could be part of the instructions for users of the contractkit package. As in "If you use contractkit@x.y.z, you must also depend on underscore@u.v.w to reduce bundle sizes", but we might as well force it ourselves.

I'm new to the js and npm world, so I may be approaching this from the wrong angle. Let me know what you think.

hbandura:
It was roughly a 800kb decrease in bundle size by forcing underscore:1.9 in the client package

gastonponti:
Yes, what @cmcewen says, is that even that code is already dup, and it probably could live, in the utils, without having to use underscore (as we have it in the contractkit). But I think that these are different problems, remove underscore from our monorepo, and the other is to force web3 to use a smaller version of it.

Which part of web3 is using underscore?

hbandura:
> Which part of web3 is using underscore?

I think all of them:

```
$ find . -name package.json -exec grep underscore {} + | grep web3
./node_modules/web3-core-method/package.json:    "underscore": "1.9.1",
./node_modules/web3-providers-ws/package.json:    "underscore": "1.9.1",
./node_modules/web3-utils/package.json:    "underscore": "1.9.1",
./node_modules/web3-providers-ipc/package.json:    "underscore": "1.9.1",
./node_modules/web3-eth-abi/package.json:    "underscore": "1.9.1",
./node_modules/web3-eth-accounts/package.json:    "underscore": "1.9.1",
./node_modules/web3-core-subscriptions/package.json:    "underscore": "1.9.1",
./node_modules/web3-core-helpers/package.json:    "underscore": "1.9.1",
./node_modules/web3-core-requestmanager/package.json:    "underscore": "1.9.1",
./node_modules/web3-eth/package.json:    "underscore": "1.9.1",
./node_modules/web3-eth-ens/package.json:    "underscore": "1.9.1",
./node_modules/web3-bzz/package.json:    "underscore": "1.9.1"
./node_modules/web3-eth-contract/package.json:    "underscore": "1.9.1",
```

AlexBHarley:
I'd be in favour of closing this, quite a lot has changed in our setup since this PR. I just did some quick tests with the `resolutions` property:

```
  "resolutions": {
    "underscore": "1.9.1"
  }
```

in both the root `package.json` and the `packages/sdk/contractkit/package.json` and didn't see any differences in resulting bundle size.

Feel free to re open if you feel differently.

PR #4621, opened 2020-08-04
Different error handling type semantics

nambrot:


nambrot:
Here are my thoughts

1. [Handle errors like golang with a [value, error] tuple](https://github.com/celo-org/celo-monorepo/pull/4621/commits/a9bb6dc2fc71e5ac85808d9c7660bad04c00b05f)

Pretty poor, as the lack of compiler help when `error === null` still requires consumers to do a null check against the value

2. [Handle errors by making the error part of the return value by using discriminated union types](https://github.com/celo-org/celo-monorepo/pull/4621/commits/b76c2106d009353ab632ae86610088d7db413399)

Anyone with a fairly decent grasp of Typescript should be able to understand the ability to discriminate the values as success or error. Even most developers without typescript often model HTTP requests by having a "status" field.

3. [Handle errors by using a dedicated library types like Either](https://github.com/celo-org/celo-monorepo/pull/4621/commits/fc82fd61ec5bfbb69bb794226dc0b61931873e9b)

Most foreign looking to most developers likely and the benefit over 2. seems marginal. It's more concise but requires a rather significant change in thinking.

4. [Classic throwable errors](https://github.com/celo-org/celo-monorepo/pull/4621/commits/f58f63a15fa6cde1ab4309bf82f47002f6fe650e)

Most understood by anyone, though also the hardest to actually react to errors without string matching error messages.

mcortesi:
> Here are my thoughts
> 
> 1. [Handle errors like golang with a [value, error] tuple](https://github.com/celo-org/celo-monorepo/pull/4621/commits/a9bb6dc2fc71e5ac85808d9c7660bad04c00b05f)
> 
> Pretty poor, as the lack of compiler help when `error === null` still requires consumers to do a null check against the value

Agree, not a good option for me

> 1. [Handle errors by making the error part of the return value by using discriminated union types](https://github.com/celo-org/celo-monorepo/pull/4621/commits/b76c2106d009353ab632ae86610088d7db413399)
> 
> Anyone with a fairly decent grasp of Typescript should be able to understand the ability to discriminate the values as success or error. Even most developers without typescript often model HTTP requests by having a "status" field.
> 
> 1. [Handle errors by using a dedicated library types like Either](https://github.com/celo-org/celo-monorepo/pull/4621/commits/fc82fd61ec5bfbb69bb794226dc0b61931873e9b)
> 
> Most foreign looking to most developers likely and the benefit over 2. seems marginal. It's more concise but requires a rather significant change in thinking.

I think the difference between this too is blurry. What makes `Either` more complex, is because the code then starts to have a lot of functional combinators, etc.

I can imagine something like

```ts
interface Task<A> {
  result?: A
  error?: Error
}
```

or even better (mind my ts, i'm rusty and probably doesnt' compile)
```ts
type Task<A> = OkTask<A> | FailedTask<A>

interface OKTask<A> { result: A, error?: null }
interface FailedTask<A> { result?: A, error: Error }
```

and then have several utility functions so you can go from classic to task and viceversa:
```ts
func newTask(fn () => A): Task<A>
func fromTask(t: Task<A>): A. // will throw if error
```

That i like, still.. i don't think any of this really solve the problem. Since i understand the problem is not about how to learn about the errors (a catch will get you the error).... but who to handle different type of errors.

And for that, maybe it's more a question of defining an Error Type that is good for our needs. 

To rephrase the question.... the critique to classic thinking is "the hardest to actually react to errors without string matching error messages" but neither of the examples actually show how to react to different type of errors, they only show how the error "travels" to the api user.

nambrot:
Yes I agree that I haven't actually written the code in any of the examples that show `how to react to different type of errors`, I just figured it was implicit by having errors typed that that makes it more possible (via discriminated unions basically)

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -6,14 +6,21 @@ import { ContractKit } from '../kit'
...
 import { IdentityMetadataWrapper } from './metadata'
 import { StorageWriter } from './offchain/storage-writers'
+
 const debug = debugFactory('offchaindata')
 
+class FetchError extends Error {}
+class InvalidSignature extends Error {}
+class NoStorageRootProvidedData extends Error {}
+export type OffchainErrors = FetchError | InvalidSignature | NoStorageRootProvidedData
+export type ReadResponse = { status: 'ok'; data: any } | { status: 'error'; error: OffchainErrors }
```
jmrossy: How about an enum for these status values instead of using raw strings here?

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -52,35 +69,34 @@ class StorageRoot {
...
     // TODO: Only if the signer is authorized with an on-chain key
-    const [, err] = await this.read(`/account/authorizedSigners/${toChecksumAddress(signer)}`)
-
-    if (err) {
-      return [null, err]
+    try {
+      await this.read(`/account/authorizedSigners/${toChecksumAddress(signer)}`)
+      return body
+    } catch (error) {
+      throw error
```
jmrossy: How do 'unchecked' exception (to borrow a Java term) fit in to the whole scheme here?

jmrossy:
Of the 4 options you laid out, I prefer #2 and like Mariano's suggestion of something like a Task type to facilitate the structure.
I find this raises two interesting questions:
1. Should Tasks still throw or should we guarantee to catch all Error?
2. Should we use Error subclasses or an ErrorType enum? I prefer enums because anytime I fall back to reflection like `typeof` it feels like a hack. I've also heard there are perf concerns.

mcortesi:
> Of the 4 options you laid out, I prefer #2 and like Mariano's suggestion of something like a Task type to facilitate the structure.
> I find this raises two interesting questions:
> 
> 1. Should Tasks still throw or should we guarantee to catch all Error?

catch all for sure. I would be weird to need to manage *both* error styles like:

```ts
try {
  const task = doSomethingThatReturnsATask()
  if isError(task) {
    // ...
  }
catch (otherErrors Error) {
  ...
}
```

> 2. Should we use Error subclasses or an ErrorType enum? I prefer enums because anytime I fall back to reflection like `typeof` it feels like a hack. I've also heard there are perf concerns.

+1 to not use `typeof` or `instanceof`. I had problems with that. In particular, when using nominal types (classes) it can happen that you have let's say 2 versions of contractKit so doing `instanceof` won't work if you somehow mix imports.

bowd:
💯 Thanks @nambrot for taking the time to do this and moving the conversation further!

I agree with most things said here but I want to point out that what we're referring to as (2) + a `Task` type is actually (3) in disguise, and that would be a good middleground imo, it would be intuitive to use and it would hide the actual FP primitives:

```typescript
// task.ts:
import { Either, isLeft, isRight, left, right, Right, Left } from "fp-ts/lib/Either";

export type Task<TResult, TError> = Either<TResult, TError>;
export const isError = isRight;
export const isOk = isLeft;
export const error = right;
export const ok = left;
// These next two convenience methods if we really want to hide the
// usage of task.left & task.right which are a bit more esoteric
export const getError = <T>(task: Right<T>): T => task.right;
export const getOk = <T>(task: Left<T>): T => task.left;


// in the wild:
import { Task, isError, error, ok, getError, getOk } from "task";

// ... + error definitions with discriminant 
type MyError = UserError | ServerError;

function thatCanFail(): Promise<Task<{response: string}, MyError>> {
  if (true) {
    return ok({response: "Ok"})
  } else {
    return error(UserError);
  } 
}

function thatConsumes() {
  const task = thatCanFail();
  if (isError(task)) {
    const err = getError(task) // or task.right
  } else {
    const res = getOk(task) // or task.left
  }
}

**Comment by bowd** [comment]
The risk with that approach is that consuming it as an Either would result in concise yet specialised code like in [Nam's usage of pipe here.](https://github.com/celo-org/celo-monorepo/pull/4621/commits/fc82fd61ec5bfbb69bb794226dc0b61931873e9b#diff-1c25d16c4c5fe6d0c2823bad182a9b25L35-R41)

File: packages/contractkit/src/identity/offchain-data-wrapper.ts (hunk 1 above)
Comment (nambrot) [reply packages/contractkit/src/identity/offchain-data-wrapper.ts]: Great point, added that to the Task type

**Comment by nambrot** [comment]
Thanks for everybody's input. I tried to synthesize the (2)+ that seemed to resonate

Some observations:
- I tried to keep the interface to non-typescript folks as "common" as possible. That meant to not reuse `Either` as `left/right` would probably confuse people. I also tried to make the Error type conform here so that we can reuse typical information of an error like the stack trace. However the latter was a bit awkward to get the compiler to discriminate on. Pardon the poor naming of some of the errors.
- I also changed the test to show how one were to actually change behavior depending on what kind of error is returned which hopefully clarifies things.
- @bowd rightly points out that the use of "special functions" like `pipe` should be avoided. That being said, I do wonder whether it would be helpful to provide some basic functions like `and_then` onto the `Task` type to allow for transformation of the task to avoid having to do `if(isError(result))` after every computation. But that concern is certainly secondary

**Comment by jeanregisser** [comment]
Great initiative!

The newly defined error types look great to me. 

Now, having those, do we really want/need to introduce something different than classic throwable errors?
There will still be boundaries where standard throwable errors will need to be handled (other errors thrown by dependencies).

File: packages/contractkit/src/identity/task.ts [hunk 3]
```diff
@@ -0,0 +1,42 @@
+export enum TaskStatus {
```
jmrossy: I suggest moving this out of `identity` to somewhere else like maybe `utils`

packages/contractkit/src/identity/task.ts
```diff
@@ -0,0 +1,42 @@
...
+  Ok = 'ok',
```
jmrossy: Nit: casing

packages/contractkit/src/identity/task.ts
```diff
@@ -0,0 +1,42 @@
...
+  errorType: T
+}
+
+export class RootError<T> extends Error implements BaseError<T> {
+  constructor(readonly errorType: T) {
+    super()
+    // @ts-ignore
+    if (errorType.toString) {
+      // @ts-ignore
+      this.name = errorType.toString()
```
jmrossy: If errorType is just going to be used for name, should we call it name?

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -19,23 +42,39 @@ export class SingleSchema<T> {
...
-  return parseResult.right
+  try {
+    const asJson = JSON.parse(resp.result)
+    const parseResult = type.decode(asJson)
+    if (isRight(parseResult)) {
+      return Ok(parseResult.right)
+    }
+    return Err(new RootError<SchemaErrorTypes.InvalidDataError>(SchemaErrorTypes.InvalidDataError))
+  } catch (error) {
+    return Err(new RootError<SchemaErrorTypes.InvalidDataError>(SchemaErrorTypes.InvalidDataError))
```
jmrossy: I wonder if we can make things more concise, like removing the need for `new` and the duplication of the type and param.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -19,23 +42,39 @@ export class SingleSchema<T> {
   }
 }
 
+function foo(err: SchemaErrors) {
```
jmrossy: did you mean to include this?

jmrossy (commented):
The approach lgtm, just some minor comments

packages/contractkit/src/identity/task.ts
nambrot: I tried to use it the other way round. I.e. I set name to the ErrorType in the absence of a subclass setting the name.

That being said, we need `errorType` to be of type `T` to be able to pattern match on it I believe

packages/contractkit/src/identity/offchain/schema-utils.ts
nambrot: I agree that this is quite verbose, but I was unable to find a way to make it more concise. To my understanding:

- Constructing the Error with `new` is necessary so that the Error object will have the stacktrace
- The type annotation seemed necessary as the compiler was unable to deduce the proper type from a union type
- The param seemed necessary as the Error object itself needs the value for consumers to be able to pattern match on

I would love for someone with more typescript-fu to improve this

nambrot: Looks like extending the error interfaces with `Error` I was able to remove the type annotation!

packages/contractkit/src/identity/task.ts
jmrossy: ok 👍

jmrossy (approved):
LGTM! Not sure why all the docs changed but they seem like trivial changes. You may need another docs gen though, the tests seem to be failing for that.

packages/base/src/result.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export function isError<TResult, TError extends Error>(
+  result: Result<TResult, TError>
+): result is ErrorResult<TError> {
+  return result.status === ResultStatus.Error
+}
+export function isOk<TResult, TError extends Error>(
+  task: Result<TResult, TError>
+): task is OkResult<TResult> {
+  return task.status === ResultStatus.Ok
```
bowd: s/task/result/

packages/contractkit/src/identity/offchain/schemas.ts
```diff
@@ -26,15 +27,18 @@ export class AuthorizedSignerAccessor {
   basePath = '/account/authorizedSigners'
   constructor(readonly wrapper: OffchainDataWrapper) {}
 
-  async read(account: Address, signer: Address) {
-    return readWithSchema(
+  async readAsResult(account: Address, signer: Address) {
+    console.log(this)
```
bowd: Probably remove this

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -24,20 +47,24 @@ export default class OffchainDataWrapper {
       .filterClaims(ClaimTypes.STORAGE)
       .map((_) => new StorageRoot(account, _.address))
     debug({ account, storageRoots })
-    const data = (await Promise.all(storageRoots.map((_) => _.read(dataPath)))).find((_) => _)
 
-    if (data === undefined) {
-      return undefined
+    if (storageRoots.length === 0) {
+      return Err(new RootError(OffchainErrorTypes.FetchError))
```
bowd: This should probably not be a `FetchError`, maybe `NoStorageRootProvidedData`?

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -24,20 +47,24 @@ export default class OffchainDataWrapper {
...
 
-    const [actualData, error] = data
-    if (error) {
-      return undefined
+    const item = await (await Promise.all(storageRoots.map((_) => _.readAsResult(dataPath)))).find(
+      isOk
+    )
+
+    if (item === undefined) {
+      return Err(new RootError(OffchainErrorTypes.FetchError))
```
bowd: This should probably not be a `FetchError`, maybe `NoStorageRootProvidedData`?

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -53,35 +80,31 @@ class StorageRoot {
...
     }
 
     const body = await data.text()
 
     const signature = await fetch(this.address + dataPath + '.signature')
 
     if (!signature.ok) {
-      return [null, 'Signature could not be fetched']
+      return Err(new RootError(OffchainErrorTypes.FetchError))
```
bowd: This should be `InvalidSignature`?

bowd (approved):
Overall I like it, added some minor comments around but I also have [this proposal](https://github.com/celo-org/celo-monorepo/pull/4745) about eliminating the `interface` pattern for error type definition and always use classes, to keep things a bit simpler to understand. But I'm ok even without it.

packages/base/src/result.ts
```diff
@@ -0,0 +1,92 @@
+export enum ResultStatus {
+  Ok = 'Ok',
+  Error = 'Error',
+}
+export interface OkResult<TResult> {
+  status: ResultStatus.Ok
```
mcortesi: why not a boolen?

packages/base/src/result.ts
```diff
@@ -0,0 +1,92 @@
...
+    const result = await f(...args)
+    return throwIfError(result, errorModifier)
+  }
+}
+
+export interface BaseError<T> {
+  errorType: T
+}
+
+export class RootError<T> extends Error implements BaseError<T> {
```
mcortesi, line 66: check about extending error in javascript... is not that simple!

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
```diff
@@ -120,9 +147,10 @@ testWithGanache('Offchain Data', (web3) => {
       const nameAccessor = new NameAccessor(wrapper)
       await nameAccessor.write({ name: testname })
 
-      const receivedName = await nameAccessor.read(writer)
-      expect(receivedName).toBeDefined()
-      expect(receivedName!.name).toEqual(testname)
+      const resp = await nameAccessor.readAsResult(writer)
+      if (isOk(resp)) {
```
mcortesi: I would suggest just have `Result` have `ok:boolen` field instead of status... which make reading this easier and less "Abstract" which i think is a good thing. 

So you would write:
```
if (resp.ok) {
```
which is also what you already do with the `fetch()` interface.

So basically *ditch* the `isOk` and `isError` which just creates an unnecessary abstraction

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -24,20 +55,24 @@ export default class OffchainDataWrapper {
...
+    const item = await (await Promise.all(storageRoots.map((_) => _.readAsResult(dataPath)))).find(
+      isOk
+    )
+
+    if (item === undefined) {
+      return Err(new NoStorageRootProvidedData())
     }
 
-    return actualData
+    return item
```
mcortesi, line 72: Is `item` of `Result` type? shouldn't this be `ok(item)`... 
and if this is an error, then i would more worried about the compiler not seeing this

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
```diff
@@ -111,7 +138,7 @@ testWithGanache('Offchain Data', (web3) => {
 
     it('can read the authorization', async () => {
       const authorizedSignerAccessor = new AuthorizedSignerAccessor(wrapper)
-      const authorization = await authorizedSignerAccessor.read(writer, signer)
+      const authorization = await authorizedSignerAccessor.readAsResult(writer, signer)
```
mcortesi, line 139: if we go with this... i would invert the namins scheme, making `read` return a `result` and some `readXX` be the throwable one

packages/base/src/result.ts
mcortesi: https://stackoverflow.com/questions/1382107/whats-a-good-way-to-extend-error-in-javascript

packages/base/src/result.ts
```diff
@@ -0,0 +1,92 @@
...
+export interface OkResult<TResult> {
+  status: ResultStatus.Ok
+  result: TResult
+}
+export interface ErrorResult<TError extends Error> {
+  status: ResultStatus.Error
+  error: TError
+}
+
+export type Result<TResult, TError extends Error> = OkResult<TResult> | ErrorResult<TError>
```
mcortesi, line 10: we are going to write this ALL the time... maybe `Task` to save a few keystrokes :P

packages/base/src/result.ts
```diff
@@ -0,0 +1,92 @@
...
+  TError extends Error,
+  TModifiedError extends Error
+>(
+  f: (...args: TArgs) => Result<TResult, TError>,
+  errorModifier?: (error: TError) => TModifiedError
+) {
+  return (...args: TArgs) => {
+    return throwIfError(f(...args), errorModifier)
+  }
+}
```
mcortesi:
```suggestion
export function makeThrowable<
  TArgs extends any[],
  TResult,
  TError extends Error,
  TModifiedError extends Error
>(
  f: (...args: TArgs) => Result<TResult, TError>,
  errorModifier?: (error: TError) => TModifiedError
) {
  return (...args: TArgs) => throwIfError(f(...args), errorModifier)
}
```

packages/base/src/result.ts
nambrot: Are you saying that it currently doesn't work?

The pattern matching on `errorType` should work?

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
nambrot: I personally wanted to do it this way to make this error handling opt-in vs. opt out to minimize the disruption to more convention consumption

packages/base/src/result.ts
nambrot: Someone already defined a `Task` in `base` :P

mcortesi: mmmm.... surely a really nice guy! :P

packages/contractkit/src/identity/offchain-data-wrapper.ts
nambrot: Item is a result type indeed

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
mcortesi: maybe use a convention like `readR` (`R` suffix to all result reutningn funciotns). My worry is that people won't see much benefit of using it..  and they see it as  just longer and more tedious...

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -24,20 +55,24 @@ export default class OffchainDataWrapper {
...
-    if (data === undefined) {
-      return undefined
+    if (storageRoots.length === 0) {
+      return Err(new NoStorageRootProvidedData())
     }
 
-    const [actualData, error] = data
-    if (error) {
-      return undefined
+    const item = await (await Promise.all(storageRoots.map((_) => _.readAsResult(dataPath)))).find(
```
mcortesi: why the double await?

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -10,34 +30,43 @@ export class SingleSchema<T> {
...
-  const data = await wrapper.readDataFrom(account, dataPath)
-  if (data === undefined) {
-    return
-  }
-  const asJson = JSON.parse(data)
-  const parseResult = type.decode(asJson)
-  if (isLeft(parseResult)) {
-    return undefined
+): Promise<Result<T, SchemaErrors>> => {
+  const resp = await wrapper.readDataFromAsResult(account, dataPath)
```
mcortesi, line 50: something challenging about the overall solution is that `Promises` can fail... and that's a different error path to Result...

So if you're really defensive about your code, you should `try/catch` your awaits.... and have 2 error checking  mechanism.

packages/contractkit/src/identity/offchain-data-wrapper.ts
nambrot: Good question, not sure

packages/contractkit/src/identity/offchain/schema-utils.ts
nambrot: Promise errors vs. result errors is an interesting question but also not one I think we can generally give recommendations about

I.e. at the lowest levels, code calling IO or external code should indeed catch pretty much everywhere. How much a caller calling a result can rely on that is a more difficult of course. In this example, the readAsResult method certainly needs to catch lower exceptions

PR #4624, opened 2020-08-04
Reenable Hermes

cmcewen:
### Description

We had to disable hermes due to the rn-firebase use of Proxy, but there is a rc1 build which supports it (https://github.com/facebook/hermes/issues/33)

### Tested

Did an apk build and installed it/played around with it

### Backwards compatibility

Expected to work


annakaz (approved):
Sweet! LGTM, will help a lot with performance in Grameen pilot

PR #4625, opened 2020-08-04
Replace 'lodash.range' references to remove lodash dependency from contractkit

hbandura:
Replaced two references of lodash.range with the same inline solution present in ledger-wallet.ts: `Array.from(Array(to).keys())`

This reduces bundle sizes for contractkit clients in roughly 600-700kb


packages/contractkit/src/utils/tx-uri.ts
```diff
@@ -1,11 +1,12 @@
 import { trimLeading0x } from '@celo/utils/lib/address'
 import BN from 'bn.js'
-import { range } from 'lodash'
 import qrcode from 'qrcode'
 import querystring from 'querystring'
 import { Tx } from 'web3-core'
 import abi from 'web3-eth-abi'
 
+const range = (to: number) => Array.from(Array(to).keys())
```
mcortesi: can we have a `range()` exported in some utitily pkg, inside contractkit at least?

hbandura: Done!

mcortesi (approved):
Thanks

PR #4628, opened 2020-08-04
Remove await from log upload on Geth creation

tarikbellamine:
### Description
Removed an unnecessary promise `await` of geth log uploads during geth creation. The await was unnecessarily blocking geth creation.

### Other changes
Minor naming

### Tested
Yes

### Related issues
N/A

### Backwards compatibility
Yes


packages/mobile/src/utils/LogUploader.ts
```diff
@@ -49,7 +49,7 @@ export default class FirebaseLogUploader {
     await RNFS.moveFile(logFilePath, tmpFilePath)
     // Don't "await" here. While everything else in this async method is relatively fast, data upload can be slow
     // and we don't want to block the user from using the app.
-    return FirebaseLogUploader.uploadLogsToFirebaseStorage(tmpFilePath, uploadPath, uploadFileName)
```
tarikbellamine, line 52: This `return` was functionally an `await`. Given the comment above, this feels like the right one to get rid of.

tarikbellamine (commented):
Could also see a case for removing the `await` [here](https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/geth/geth.ts#L105) and [here](https://github.com/celo-org/celo-monorepo/blob/2ffa7b5299835d9992c6e328a4d5e6d291acee95/packages/mobile/src/geth/geth.ts#L352).

Decided to leave the above as-is for now given there was a specific comment about what not to await (which we accidentally were).

tarikbellamine (commented):
Regarding the concern about not logging without user permission, we are already doing a check [here](https://github.com/celo-org/celo-monorepo/blob/2ffa7b5299835d9992c6e328a4d5e6d291acee95/packages/mobile/src/utils/LogUploader.ts#L23) for if the user has analytics enabled (among other checks for data considerations)

tkporter (approved):
🙏 didn't realize we were already checking for analytics being enabled, good find!

packages/mobile/src/utils/LogUploader.ts
jmrossy: So should we remove the comment?

packages/mobile/src/utils/LogUploader.ts
```diff
@@ -49,7 +49,7 @@ export default class FirebaseLogUploader {
     await RNFS.moveFile(logFilePath, tmpFilePath)
```
jmrossy, line 49: Do we ever remove these tmp files or do they just pile up?

tarikbellamine: I think the temp file gets overwritten every time there is an upload

PR #4630, opened 2020-08-04
Changing libphonenumber to libphonenumber-js

hbandura:
Since google libphonenumber weights ~ 700kb, in this PR we attempt to change it to libphonenumber-js, which is supposedly much smaller

### DIFF
--- a/packages/utils/src/phoneNumbers.ts
+++ b/packages/utils/src/phoneNumbers.ts
@@ -2,6 +2,7 @@ import CountryData from 'country-data'
 import {
   CountryCode,
   getCountries,
+  getCountryCallingCode,
   getExampleNumber as libGetExampleNumber,
   parsePhoneNumber as libParsePhoneNumber,
   parsePhoneNumberFromString,
@@ -85,13 +86,19 @@ export function getRegionCode(e164PhoneNumber: string) {
   }
 }
 
-function _getRegionCodeFromCountryCode(countryCode: string | undefined): CountryCode | null {
+function _getRegionCodeFromCountryCode(countryCode: string | undefined): CountryCode | undefined {
   if (!countryCode) {
-    return null
+    return undefined
   }
   const countries = getCountries()
-  const regionCode = countries[parseInt(countryCode, 10)]
-  return regionCode
+  const asNumber = parseInt(countryCode, 10)
+  // TODO: precalculate a dictionary country calling code => country iso code
+  // to avoid this iteration
+  return countries.find((country) => {
+    const cc = getCountryCallingCode(country) as string
+    const ccAsNumber = parseInt(cc, 10)
+    return ccAsNumber === asNumber
+  })
 }
 
 export function getRegionCodeFromCountryCode(countryCode: string) {
--- a/packages/utils/src/phoneNumbers.ts
+++ b/packages/utils/src/phoneNumbers.ts
@@ -1,10 +1,19 @@
 import CountryData from 'country-data'
+// import {
+//   PhoneNumber,
+//   PhoneNumberFormat,
+//   PhoneNumberType,
+//   PhoneNumberUtil,
+// } from 'google-libphonenumber'
 import {
+  CountryCode,
+  getCountries,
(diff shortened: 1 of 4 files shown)


packages/utils/src/phoneNumbers.ts
```diff
@@ -56,9 +59,13 @@ export function getCountryCode(e164PhoneNumber: string) {
     return null
   }
   try {
-    return phoneUtil.parse(e164PhoneNumber).getCountryCode()
+    // type CountryCallingCode is an extension of String, but
+    // type checks fail when attempting to pass it as a parameter for parseInt,
+    // so here we force the string representation, and then parse it to a number
+    const callingCode = parsePhoneNumberFromString(e164PhoneNumber)?.countryCallingCode
+    return parseInt(callingCode as string, 10)
```
jmrossy, line 66: We generally recommend using BigNumber for all string to number parsing.

packages/utils/src/phoneNumbers.ts
```diff
@@ -56,9 +59,13 @@ export function getCountryCode(e164PhoneNumber: string) {
...
   } catch (error) {
-    console.debug(`getCountryCode, number: ${e164PhoneNumber}, error: ${error}`)
+    console.debug(`countryCallingCode, number: ${e164PhoneNumber}, error: ${error}`)
```
jmrossy, line 68: Why change this? We use this convention of starting with the method name (or tag and method name) very often.

packages/utils/src/phoneNumbers.ts
```diff
@@ -144,12 +174,13 @@ export function parsePhoneNumber(
...
-      ? getRegionCodeFromCountryCode(defaultCountryCode)
-      : null
-    const parsedNumberUnfixed = phoneUtil.parse(phoneNumberRaw, defaultRegionCode || undefined)
-    const parsedCountryCode = parsedNumberUnfixed.getCountryCode()
-    const parsedRegionCode = phoneUtil.getRegionCodeForNumber(parsedNumberUnfixed)
+    const defaultRegionCode = _getRegionCodeFromCountryCode(defaultCountryCode)
+    const parsedNumberUnfixed = libParsePhoneNumber(phoneNumberRaw, defaultRegionCode || undefined)
+    const strParsedCountryCode = parsedNumberUnfixed?.countryCallingCode
+    const parsedCountryCode = strParsedCountryCode
+      ? parseInt(strParsedCountryCode as string, 10)
```
jmrossy, line 181: ditto

packages/utils/src/phoneNumbers.ts
```diff
@@ -159,21 +190,16 @@ export function parsePhoneNumber(
...
-          countryCode: parsedCountryCode,
-          regionCode: parsedRegionCode,
-        }
-      : null
+    if (!parsedNumber.isValid()) {
+      return null
+    }
+    return {
+      e164Number: parsedNumber.format('E.164'),
+      displayNumber: handleSpecialCasesForDisplay(parsedNumber, parsedCountryCode),
```
jmrossy, line 198: Have you confirmed these special cases are still needed for this new lib?

yarn.lock
```diff
@@ -22348,6 +22348,14 @@ libphonenumber-js@^1.7.22:
     minimist "^1.2.0"
     xml2js "^0.4.17"
 
+libphonenumber-js@^1.7.50:
```
jmrossy, line 22351: Since there are no removals in yarn.lock, some other packages must still be relying on google-phone-lib. Would be good to double check that

jmrossy (commented):
Thanks for having a look at this! This is potentially a big change, a lot of packages rely on these phone number utils. Any behavior inconsistent with the last lib will cause users to lose their verification status, fail to look each other up, etc. We'll need to test this with numbers from a bunch of different countries, particularly our target market countries

barbaraliau:
@hbandura is this PR still relevant?

hbandura:
> @hbandura is this PR still relevant?

I guess not. It was mostly an experiment to see how big of a change would this be. I'll close it

PR #4632, opened 2020-08-05
[Wallet] Explicitly passing country name along with country code for analytics

tarikbellamine:
### Description
Explicitly passing country name along with country code for analytics

### Other changes
Setting address to lowercase before logging to avoid duplicates being created with different cases

### Tested
Yes

### Related issues
N/A

### Backwards compatibility
Yes


annakaz (approved):
👑 📈

PR #4635, opened 2020-08-05
Signer replay attack prevention

### LINKED ISSUE #4465: Protect against a replay attack
### Expected Behavior

It should not be possible for a PGPNP signer to negatively impact the onboarding user experience. This is important for security and decentralization.

### Current Behavior

There's a potential vulnerability where a malicious PGPNP signer (of the 9 that we have) could replay the request it received from a given user to all other signers and exhaust the user's quota.

alecps:
### Description

Fixes potential vulnerability where a malicious signer can repeatedly forward the same request to other signers and exhaust the requesting user’s quota. Signers will now store a table of all requests they have seen, and check incoming requests against this set before processing. Timestamps have been added to the request body to help the signer differentiate between requests.

### Other changes

Adds /getQuota endpoint to signer so that an account can query its quota without effecting it.

### Tested

Using the signer e2e and unit tests. 

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/4465

### Backwards compatibility

Current changes are backwards compatible, but a future update will make timestamps a required part of the request body and will not be backwards compatible. The plan is to make this update after timestamps have been added to the message sent by the client.


packages/phone-number-privacy/combiner/src/common/constants.ts
```diff
@@ -1,3 +1,4 @@
 export const REASONABLE_BODY_CHAR_LIMIT: number = 2000
 export const DB_TIMEOUT = 10000
 export const MAX_BLOCK_DISCREPANCY_THRESHOLD = 3
+export const REQUEST_EXPIRY_WINDOW = 180000 // 3 minutes
```
codyborn: To make it more explicit, you can change it to:
`REQUEST_EXPIRY_WINDOW_MS = 3 * 60 * 1000`

packages/phone-number-privacy/combiner/src/common/input-validation.ts
```diff
@@ -25,6 +25,11 @@ export function hasValidPhoneNumberHash(requestBody: any): boolean {
   return requestBody.hashedPhoneNumber && isByte32(requestBody.hashedPhoneNumber)
 }
 
+export function hasValidTimestamp(requestBody: any): boolean {
```
codyborn, line 28: You don't have to do this in this PR, but we should move these common validation checks to our Common library so that we don't have to duplicate it across the combiner/signer:
https://github.com/celo-org/celo-monorepo/tree/master/packages/phone-number-privacy/common/src/utils

packages/phone-number-privacy/combiner/src/common/input-validation.ts
```diff
@@ -25,6 +25,11 @@ export function hasValidPhoneNumberHash(requestBody: any): boolean {
...
+  // TODO(Alec): make timestamp required
+  return !requestBody.timestamp || requestBody.timestamp > Date.now() - REQUEST_EXPIRY_WINDOW
```
codyborn: Might also be helpful to validate the type before comparison:
`typeof requestBody.timestamp === 'number'`

packages/phone-number-privacy/signer/src/common/constants.ts
```diff
@@ -2,3 +2,4 @@ export const REASONABLE_BODY_CHAR_LIMIT: number = 2000
 export const DB_TIMEOUT = 10000
 export const RETRY_COUNT = 5
 export const RETRY_DELAY_IN_MS = 100
+export const REQUEST_EXPIRY_WINDOW = 180000 // 3 minutes
```
codyborn: We should put this in the Common lib also

packages/phone-number-privacy/signer/src/database/wrappers/request.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function getRequestExists(request: GetBlindedMessageForSaltRequest): Promise<boolean> {
+  if (!request.timestamp) {
+    return false // TODO(Alec) make timestamps required
+  }
+  logger.debug('Checking if request exists')
+  try {
+    const existingRequest = await requests()
+      .where(REQUESTS_COLUMNS.timestamp, new Date(request.timestamp as number)) // indexed
+      .andWhere(REQUESTS_COLUMNS.address, request.account)
+      .andWhere(REQUESTS_COLUMNS.hashedPhoneNumber, request.hashedPhoneNumber as string) // TODO(Alec) why is this not always defined?
```
codyborn: This is the caller's phone number identifier, which may not be added if the caller is querying their own pepper for the first time.

packages/phone-number-privacy/signer/src/database/models/request.ts
```diff
@@ -0,0 +1,22 @@
+import { GetBlindedMessageForSaltRequest } from '../../salt-generation/get-salt'
+
+export const REQUESTS_TABLE = 'requests'
+export enum REQUESTS_COLUMNS {
+  address = 'address',
```
codyborn: nit: maybe we can call this "caller_address" to make it clear who's address this is

packages/phone-number-privacy/signer/src/salt-generation/get-salt.ts
```diff
@@ -33,7 +36,7 @@ export async function handleGetBlindedMessageForSalt(
 ) {
   logger.info('Begin getBlindedSalt request')
   try {
-    if (!isValidGetSignatureInput(request.body)) {
+    if (!isValidGetSignatureInput(request.body) || (await getRequestExists(request.body))) {
       respondWithError(response, 400, WarningMessage.INVALID_INPUT)
```
codyborn: I think we should let this request go through and just not consume quota. If the response got dropped on the network, the client or combiner should retry with the same request timestamp in order to not consume quota unnecessarily.

codyborn (changes requested):
Nice work! Before checking this in, would be great to get sizes of partner DBs to understand whether we need to prioritize a "cleanup" process.

alecps:
Note that the gcp build will fail until the updated `@celo/phone-number-privacy-common` package is published

packages/phone-number-privacy/combiner/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "@celo/contractkit": "0.4.5",
-    "@celo/phone-number-privacy-common": "1.0.5",
+    "@celo/phone-number-privacy-common": "1.0.6-dev",
```
jmrossy: The combiner is a cloud function so it can't rely on any packages that aren't publicly available. If you need to use common updates, you'll need to publish that to npm first.

packages/phone-number-privacy/combiner/test/end-to-end/get-blinded-salt.test.ts
```diff
@@ -102,10 +104,10 @@ describe('Running against a deployed service', () => {
     expect(response.status).toBe(403)
   })
 
-  xdescribe('With funded account', () => {
-    // TODO: Figure out a common way to prefund account to provide quota
+  xdescribe('With enough quota', () => {
```
jmrossy, line 108: Does this work now? Wondering why it's still has the x to cause it to be skipped.

packages/phone-number-privacy/signer/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "dependencies": {
     "@celo/contractkit": "0.4.5",
-    "@celo/phone-number-privacy-common": "1.0.5",
+    "@celo/phone-number-privacy-common": "1.0.6-dev",
```
jmrossy: This one should also keep to the published version but for a different reason. If we use the neighbor package from the monorepo, we'd have to bloat the dockerfile with a bunch of other monorepo stuff that we don't currently include. I'm not quite as sure about this one as I am about combiner but given that you'll have to deploy common anyway, I think it makes sense to use it here too.

packages/phone-number-privacy/signer/src/database/models/request.ts
```diff
@@ -0,0 +1,19 @@
+import { GetBlindedMessageForSaltRequest } from '../../salt-generation/get-salt'
+
+export const REQUESTS_TABLE = 'requests'
+export enum REQUESTS_COLUMNS {
+  address = 'caller_address',
+  timestamp = 'timestamp',
+  blindedQueryPhoneNumber = 'blinded_query_phone_number',
```
jmrossy: In my PR #4649 I'm gonna be renaming a bunch of stuff to make the service not so coupled to the notion of phone numbers specifically. WDYT about calling this just `blindedQuery` to future proof it a bit?

packages/phone-number-privacy/signer/src/database/wrappers/request.ts
```diff
@@ -0,0 +1,47 @@
...
+    logger.error(ErrorMessage.DATABASE_UPDATE_FAILURE, e)
+    return null
+  }
+}
+
+async function insertRecord(data: Request) {
+  await requests()
+    .insert(data)
+    .timeout(DB_TIMEOUT)
+  return true
```
jmrossy: why return a value if nothing's checking it?

jmrossy: And if we remove it, no need for this function to be async. In fact we may not need this helper at all, but I'm nitpicking.

packages/phone-number-privacy/signer/test/end-to-end/get-blinded-salt.test.ts
```diff
@@ -22,88 +23,182 @@ const BLINDING_FACTOR = new Buffer('0IsBvRfkBrkKCIW6HV0/T1zrzjQSe8wRyU3PKojCnww=
...
+): Promise<number> {
+  const body = JSON.stringify({
+    account,
+    hashedPhoneNumber,
+  })
+  if (!authHeader) {
+    const web3 = new Web3(new Web3.providers.HttpProvider(DEFAULT_FORNO_URL))
+    const contractKit = newKitFromWeb3(web3)
+    contractKit.addAccount(privateKey)
+    authHeader = await contractKit.web3.eth.sign(body, account)
```
jmrossy: I think you could just make a Web3 instance instead of instantiating and connecting contract kit, seems like overkill for to do this sign call.

jmrossy (commented):
Mostly looks great!

packages/phone-number-privacy/combiner/package.json
alecps: Right. I figured I should get these changes approved and checked in first and then I would publish the common package  and update the version right afterward in a second PR. Does that sound reasonable?

packages/phone-number-privacy/combiner/test/end-to-end/get-blinded-salt.test.ts
alecps: It doesn't work. My understanding is that the combiner e2e tests run against deployed signer services. Is there a way to run them against a local signer service? They should work if run against this version of the signer.

packages/phone-number-privacy/signer/package.json
alecps: 👍

packages/phone-number-privacy/combiner/test/end-to-end/get-blinded-salt.test.ts
codyborn: This will run against whatever endpoint is set for the variable `PHONE_NUM_PRIVACY_COMBINER_SERVICE_URL` in the .env file, which can point to the localhost. Great idea for the `replenishQuota` call! We may want to add a comment to this test case that if it's failing, it may just be that the address needs to be fauceted:
`celotooljs account faucet --account <ADDRESS> --dollar 1 --gold 1 -e <ENV> --verbose`

codyborn (approved):
Looks good!

Issue #3045, opened 2020-03-07
Offload tx creation logic in dappkit code to walletkit or contractkit


jmrossy:
### Expected Behavior

We shouldn't be hand crafting txs in mobile/src/dappkit/dappkit.ts 

### Current Behavior

https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/dappkit/dappkit.ts#L63

PR #4640, opened 2020-08-06
Implement a wallet that wraps RNGeth to sign tx in Valora without starting geth

### LINKED ISSUE #4436: Support flexible usage of the Celo blockchain client as a light client or signer-only
### Description

Using the mobile app in data-saver / forno mode requires running a celo-blockchain node to continue running and syncing to the network. It should be possible to use the `celo-blockchain` signer functions without continually syncing to the network and consuming unnecessary resources.

bowd:
### Description

In order to  [improving data-saver mode](https://github.com/celo-org/celo-monorepo/issues/4436) we have been looking at how to enable Valora to function without the need to actually start the light client when it's operating in `Data Saver` mode (aka forno mode). 

This is achievable by leveraging the [KeyStore](https://github.com/celo-org/celo-blockchain/blob/master/mobile/accounts.go#L89) exposed in the `mobile` package that's accessible from the [native bridges](https://github.com/celo-org/react-native-geth/pull/27).

Thus by [extending the native bridge implementations](https://github.com/celo-org/react-native-geth/pull/27) we can expose all the functions needed to create a `RNGethWallet` and a `RNGethSigner` that's compatible with `contractkit`, meaning that we only need to actually start the light node when we want to use it as the `web3` target.

Valora usually relies on the `wallet` and `contractkit` to sign transactions, except for [one place](https://github.com/celo-org/celo-monorepo/pull/4640/files#diff-112dfb19f72115075ccdd6fb47f91210) which is also mentioned in [this issue](https://github.com/celo-org/celo-monorepo/issues/3045)

One important change that I need feedback on is how `geth` gets initialised. I modified the current flow in order to only run `geth.start` when we're not in forno mode, but apart from that configure everything as usual. The `RNGethWallet` needs a configured `geth` instance to operate but in the past the instance was set as a global and not exposed. I ended up creating a getter for it but it's not the nicest pattern because it doesn't necessarily mean it's initialised. There should be a cleaner way to do this.

### Other changes

- I made some changes to the typing around Wallets in order to enforce a stricter relationship between the Wallet and the Signer while maintaining the constraints. This way we no longer need to typecast signers returned by `getSigner`.
- Because this depends on the new version of `react-native-geth` we get some typed niceties here and there. 

### Tested

Tested in the iOS simulator manually
Haven't touched the tests yet because I'm not sure how much of this will need refactoring.

### Related issues

- Fixes #4436


packages/contractkit/src/wallets/rpc-wallet.ts
```diff
@@ -51,7 +51,7 @@ export class RpcWallet extends RemoteWallet {
   }
 
   async unlockAccount(address: string, passphrase: string, duration: number) {
-    const signer = this.getSigner(address) as RpcSigner
+    const signer = this.getSigner(address)
```
mcortesi, line 51: there are few more `as RpcSigner` thatyou can remove in `isAccountUnlocked` and `signTransaction`

packages/mobile/src/geth/RNGethSigner.ts
```diff
@@ -0,0 +1,84 @@
...
+  constructor(
+    protected geth: RNGeth,
+    protected account: string,
+    protected unlockBufferSeconds = 5,
+    protected unlockTime = -1,
+    protected unlockDuration = -1
+  ) {}
+
+  async init(privateKey: string, passphrase: string) {
+    return await this.geth.addAccount(privateKey, passphrase)
```
mcortesi: no need to do the await here, it's alreadya async

packages/mobile/src/geth/RNGethSigner.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  async signRawTransaction(tx: Tx) {
+    if (normalizeAddressWith0x(tx.from! as string) !== this.account) {
+      throw new Error(`RNGethSigner(${this.account}) cannot sign tx with 'from' ${tx.from}`)
+    }
+    const encoded = rlpEncodedTx(tx)
+    const signedTxRLP = await this.geth.signTransaction(encoded.rlpEncode, this.account)
+    return await encodeTransaction(encoded, extractSignature(signedTxRLP))
```
mcortesi: same

mcortesi (commented):
I don't approve since i don't know much about Wallet code. But the rest looks great! Thanks

gastonponti:
Adding @yorhodes that is in the middle of a refactor for the wallets

packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -151,6 +151,22 @@ export async function encodeTransaction(
   return result
 }
 
+export function extractSignature(rawTx: string): { v: number; r: Buffer; s: Buffer } {
+  const rawValues = RLP.decode(rawTx)
+  let r = rawValues[10]
+  let s = rawValues[11]
+  // Account.recover cannot handle canonicalized signatures
```
yorhodes, line 166: does this mean we can't use a similar function defined in the `utils` package?

packages/mobile/src/dappkit/dappkit.ts
```diff
@@ -87,7 +88,7 @@ function* produceTxSignature(action: RequestTxSignatureAction) {
         params.to = tx.to
       }
       Logger.debug(TAG, 'Signing tx with params', JSON.stringify(params))
-      const signedTx = await web3.eth.signTransaction(params)
```
yorhodes: by changing this, we are preventing the `CeloProvider` from populating missing params before requesting the signature from the rpc

packages/mobile/src/geth/RNGethSigner.ts
```diff
@@ -0,0 +1,91 @@
...
+import RNGeth from 'react-native-geth'
+import * as ethUtil from 'ethereumjs-util'
+
+const INCORRECT_PASSWORD_ERROR = 'could not decrypt key with given password'
+const currentTimeInSeconds = () => Math.round(Date.now() / 1000)
+
+/**
+ * Implements the signer interface on top of a RNGeth instance
+ */
+export class RNGethSigner implements Signer {
```
yorhodes: could we merge `RNGethSigner` and `RNGethWallet` with `RpcWallet` and `RpcSigner` somehow?
it seems like the `RNGeth` interface is similar
there is also no other usage of `RpcWallet` and `RpcSigner` besides the wallet atm

packages/mobile/src/geth/RNGethSigner.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (normalizeAddressWith0x(tx.from! as string) !== this.account) {
+      throw new Error(`RNGethSigner(${this.account}) cannot sign tx with 'from' ${tx.from}`)
+    }
+    const encodedTx = rlpEncodedTx(tx)
+    const signature = await this.signTransaction(0, encodedTx)
+    return encodeTransaction(encodedTx, signature)
+  }
+
+  async signTransaction(
+    addToV: number,
```
annakaz: Can you add the comment below on addToV here to provide context on the parameter

packages/mobile/src/geth/RNGethSigner.ts
```diff
@@ -0,0 +1,91 @@
...
+    return extractSignature(await this.geth.signTransaction(encodedTx.rlpEncode, this.account))
+  }
+
+  async signPersonalMessage(data: string): Promise<{ v: number; r: Buffer; s: Buffer }> {
+    const hash = ethUtil.hashPersonalMessage(Buffer.from(data, 'hex'))
+    const signatureHex = await this.geth.signHash(hash.toString('hex'), this.account)
+    return ethUtil.fromRpcSig(signatureHex)
+  }
+
+  getNativeKey = () => this.account
```
annakaz: How would this be used? And is there any reason to refer to the account/address as `nativeKey` rather than account?

packages/mobile/src/geth/geth.ts
```diff
@@ -123,71 +126,71 @@ export async function initGeth(sync: boolean = true): Promise<typeof gethInstanc
...
 }
 
-async function ensureStaticNodesInitialized(sync: boolean = true): Promise<boolean> {
+async function ensureStaticNodesInitialized(): Promise<boolean> {
   const { nodeDir } = networkConfig
   Logger.debug('Geth@ensureStaticNodesInitialized', 'initializing static nodes')
   let enodes: string | null = null
   try {
-    enodes = sync ? await StaticNodeUtils.getStaticNodesAsync(DEFAULT_TESTNET) : '[]'
+    enodes = await StaticNodeUtils.getStaticNodesAsync(DEFAULT_TESTNET)
```
yorhodes, line 186: I think there is also a delete of this file somewhere which could be removed

packages/mobile/src/geth/geth.ts
```diff
@@ -279,15 +282,13 @@ async function writeStaticNodes(nodeDir: string, enodes: string) {
...
 }
 
-async function attemptGethCorruptionFix(geth: any, sync: boolean = true) {
+async function attemptGethCorruptionFix(geth: any, shouldStartNodec: boolean = true) {
   const deleteChainDataResult = await deleteChainData()
   const deleteGethLockResult = await deleteGethLockFile()
   if (deleteChainDataResult && deleteGethLockResult) {
     await geth.start()
     gethInstance = geth
-    if (sync) {
```
yorhodes, line 295: why remove this?

packages/mobile/src/index.d.ts
```diff
@@ -6,7 +6,6 @@ declare module '*.json' {
 declare module '*.png'
 declare module '*.jpg'
 
-declare module 'react-native-geth'
```
yorhodes, line 9: ?

packages/mobile/src/invite/saga.ts
```diff
@@ -374,7 +375,7 @@ function* addTempAccountToWallet(inviteCode: string) {
   Logger.debug(TAG + '@addTempAccountToWallet', 'Attempting to add temp wallet')
   try {
     // Import account into the local geth node
-    const wallet: RpcWallet = yield call(getWallet)
+    const wallet: Wallet = yield call(getWallet)
```
yorhodes: this should be `WriteableWallet` if it's using `addAccount` a few lines down

packages/mobile/src/web3/contracts.ts
```diff
@@ -4,30 +4,49 @@
...
 const TAG = 'web3/contracts'
 const KIT_INIT_RETRY_DELAY = 2000
 const CONTRACT_KIT_RETRIES = 3
 
-let ipcProvider: IpcProvider | undefined
-let gethWallet: RpcWallet | undefined
+let gethWallet: RNGethWallet | undefined
 let contractKit: ContractKit | undefined
 
+function* initWallet() {
```
yorhodes: these changes seem well-intentioned and positive but I have struggled a lot in the past in this file
have you tested these changes yet?

packages/mobile/src/geth/RNGethWallet.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  async addAccount(privateKey: string, passphrase: string): Promise<string> {
+    Logger.info(`${TAG}@addAccount`, `Adding a new account`)
+    const address = normalizeAddressWith0x(privateKeyToAddress(ensureLeading0x(privateKey)))
+    if (this.hasAccount(address)) {
+      throw new Error(RNGethWalletErrors.AccountAlreadyExists)
+    }
+    const signer = new RNGethSigner(this.geth, address)
+    const resultantAddress = await signer.init(privateKey, passphrase)
```
annakaz: Can throw error if `resultantAddress`is not equal to `address`?

packages/mobile/src/geth/geth.ts
```diff
@@ -1,9 +1,10 @@
 import { GenesisBlockUtils, StaticNodeUtils } from '@celo/contractkit'
+// import { call } from 'redux-saga/effects'
```
annakaz: Nit: remove if unused

packages/mobile/src/geth/saga.ts
```diff
@@ -114,7 +114,7 @@ export function* waitForNextBlock() {
 function* waitForGethInstance() {
   try {
     const fornoMode = yield select(fornoSelector)
-    // get geth without syncing if fornoMode
+    // get geth without starting if fornoMode
```
annakaz: With the change, we will no longer need to `waitForGethInstance` in forno mode

packages/mobile/src/geth/RNGethSigner.ts
```diff
@@ -0,0 +1,91 @@
...
+const currentTimeInSeconds = () => Math.round(Date.now() / 1000)
+
+/**
+ * Implements the signer interface on top of a RNGeth instance
+ */
+export class RNGethSigner implements Signer {
+  /**
+   * Construct a new instance of the RPC signer
+   *
+   * @param geth the RNGeth instance
```
annakaz: It seems like this RNGethSigner still requires the geth light client to be initialized. Is the plan to implement another signer that does not use geth?

packages/mobile/src/geth/geth.ts
```diff
@@ -123,71 +126,71 @@ export async function initGeth(sync: boolean = true): Promise<typeof gethInstanc
...
   } finally {
     gethLock = false
   }
 }
 
+// TODO: This needs a bigger change to the design
+// I'm exposing this now in order to build a proof
+// of concept but I need input from the wallet team
+// regarding what's the best way to expose the instance.
+export function getGethInstance() {
```
annakaz: Can the signer work without the geth client running at all in Data Saver mode? If so, I think this would be cleaner if all geth-related logic were skipped / not initialized when in Data Saver mode

packages/mobile/src/web3/contracts.ts
```diff
@@ -4,30 +4,49 @@
...
+  ValoraAnalytics.track(ContractKitEvents.init_contractkit_get_wallet_start)
+  const wallet = new RNGethWallet(yield call(getGethInstance))
```
annakaz: The wallet does not need a gethInstance in DataSaver mode, right?

yorhodes (commented):
left a review on the `RNGeth` changes too
this is mostly compatible with the modularization I was doing for the wallets in `ContractKit` so no big concern there

annakaz (commented):
Great work on this! Excited for Valora to be able to sign transactions without starting geth as it will improve Data Saver performance. Overall architecture makes sense, I appreciate the signer abstraction. 

I left some comments- still have some open questions around how signing is done in Data Saver mode. Ideally geth would remain entirely untouched (and wouldn't have to be initialized) in Data Saver mode but it's not clear to me how keys are stored in this case

@bowd Sent you a calendar invite for tomorrow am PST to chat about how this could be done

packages/contractkit/src/utils/signing-utils.ts
bowd: Actually this is copied from another function in the same file, but I needed a method that extracts the signature in order to be able to use the other `encodeTransaction` method that needs the signature as an argument.

The comment is also copied from that function: https://github.com/celo-org/celo-monorepo/blob/94af34bb8a68354da088395fb5d4a9fdffc15049/packages/contractkit/src/utils/signing-utils.ts#L190-L195

packages/mobile/src/dappkit/dappkit.ts
bowd: Can you expand on this? Where does this actually happen, is the `signTransaction` method overloaded somehow? As far as I can tell it's just a `web3` instance that can either point to the local geth IPC file or to forno, and the `eth.signTransaction` is just a wrapper for the RPC call that signs the transaction. Where does the CeloProvider come into play?

packages/mobile/src/geth/RNGethSigner.ts
bowd: I don't see why they should be merged tbh, one is using a web3 instance to do RPC calls and one is calling a native module via a bridge. We could potentially get rid of `RpcWallet` and only have the `RNGethWallet` which btw I'd rename to `GethBridgeWallet` or something a bit more clear.

packages/mobile/src/geth/geth.ts
bowd: This method only gets called when an error happens when trying to start geth. Before we were starting geth both in normal mode and data-saver mode. But with the new changes we don't start geth at all when in `data-saver` mode so this method shouldn't ever be called in `data-saver` mode therefor it doesn't need to split behaviour anymore based on the mode. If we ever reach this code we've had to start geth therefore we are syncing.

packages/mobile/src/index.d.ts
bowd: We were declaring an empty module here because we didn't have typing info for the `react-native-geth` so this was just a way to make typescript shut up. But this branch is based on my branch in `react-native-geth` that now exports types which we get on import and no longer need this.

packages/mobile/src/web3/contracts.ts
bowd: Not yet because I had some uncertainty around how to change some of the geth related init sagas to better fit new needs and I didn't want to spend too much time on testing before I got the chance to talk with some people about this to see if the direction is ok. 

But I had a productive chat with @annakaz today, so I'll be doing some refactoring here an will update+add tests as well for when I'm asking for the final PR.

Maybe I didn't make this clear but I didn't consider this final in any way, I shared the PRs because I wanted feedback before polishing the solution.

packages/mobile/src/web3/contracts.ts
bowd: Resolving based on our chat

packages/mobile/src/geth/saga.ts
bowd: Resolving based on our chat

packages/mobile/src/geth/geth.ts
bowd: Resolving based on our chat

packages/mobile/src/geth/RNGethSigner.ts
bowd: Not sure about the intent here, I added it to be in line with the `RpcSigner` implementation

packages/mobile/src/geth/RNGethSigner.ts
bowd: Resolving based on our chat

bowd:
@annakaz Hey, I'm in a pretty good state with this. Since we talked I implemented what we discussed vis-a-vis how to wait for initialising the geth bridge and also locking on the contractkit initialiser side, so I refactor things a bit there. I've also done some renaming (it's longer but I think it's more descriptive this way). 

It would be great if you could take another look and see if anything pops up. I'd like to tie up the loose ends early next week. What I still want to do is add some more tests. I've also started adding some for `src/geth/saga` - currently just some high level ones but I setup what mocking needs to be done for those to work.

packages/mobile/src/dappkit/dappkit.ts
yorhodes: the `CeloProvider` intercepts `web3.eth.signTransaction` and checks to see if the `from` is a "local account", or defined in the `kit.wallet.accounts`, meaning it can be signed without forwarding an RPC call to the node.
This is especially useful for the `sendTransaction` endpoint s.t. we can sign the transaction locally and then use `sendRawTransaction` so the node does not need to have the corresponding account's key (this is relevant for the forno use case). 
This pattern is by no means a perfect solution and I personally think it's quite ugly but we are using "web3 as an interface".

packages/mobile/src/geth/RNGethSigner.ts
yorhodes: agreed
we would need to use the `GethBridgeWallet` in conjunction with `forno` then
does that seem to work according to your changes?

packages/mobile/src/dappkit/dappkit.ts
bowd: @annakaz this is a point that I'm _worried_ about at the moment - am I missing something here?

packages/mobile/src/geth/GethNativeBridgeSigner.ts
```diff
@@ -0,0 +1,107 @@
...
+} from '@celo/contractkit/lib/utils/signing-utils'
+import { Signer } from '@celo/contractkit/lib/wallets/signers/signer'
+import { ensureLeading0x } from '@celo/utils/lib/address'
+import { normalizeAddressWith0x } from '@celo/utils/src/address'
+import * as ethUtil from 'ethereumjs-util'
+import { GethNativeModule } from 'react-native-geth'
+import { Tx } from 'web3-core'
+
+const INCORRECT_PASSWORD_ERROR = 'could not decrypt key with given password'
+const currentTimeInSeconds = () => Math.round(Date.now() / 1000)
```
jeanregisser: nit: this will round to future time, if ms part >= 500. Shouldn't we use floor instead?

packages/mobile/src/geth/saga.ts
```diff
@@ -111,13 +111,12 @@ export function* waitForNextBlock() {
   }
 }
 
-function* waitForGethInstance() {
+export function* waitForGethInit() {
```
jeanregisser: Is this being exported just for testing?
We have a rule to export it with an underscore to clearly indicate it's not meant to be used by other modules.
https://github.com/celo-org/celo-monorepo/blob/master/packages/docs/celo-codebase/style-guide/typescript.md#exporting-variables-only-for-testing

yorhodes (commented):
sorry, had pending comments but didn't submit review!

packages/mobile/src/invite/saga.ts
```diff
@@ -1,5 +1,5 @@
 import { CeloTransactionObject } from '@celo/contractkit'
-import { RpcWallet } from '@celo/contractkit/lib/wallets/rpc-wallet'
+import { WritableWallet } from '@celo/contractkit/lib/wallets/wallet'
```
jeanregisser: Would there be a better name for this?
From the outside I don't really see what it does/can do.

packages/mobile/src/web3/contracts.ts
```diff
@@ -155,7 +184,7 @@ export function* getWeb3() {
 }
 
 // Used for cases where the kit's web3 must be accessed outside a saga
-export async function getWeb3Async() {
+export async function getWeb3Async(): Promise<Web3> {
   const kit = await getContractKitAsync()
-  return kit.web3
+  return kit?.web3!
```
jeanregisser: Seems dangerous! Are we sure it can never be `undefined`?

jeanregisser (commented):
Great work!! 💯 👍 
I don't have the most context on this and it's a critical part of the app so let's get enough eyes on this.

packages/mobile/src/invite/saga.ts
bowd: Yeah I was hoping somebody comes up with a better name for this during the review :D so the interfaces look like this:
```typescript
export interface Wallet {
    getAccounts: () => Address[];
    hasAccount: (address?: Address) => boolean;
    signTransaction: (txParams: Tx) => Promise<EncodedTransaction>;
    signTypedData: (address: Address, typedData: EIP712TypedData) => Promise<string>;
    signPersonalMessage: (address: Address, data: string) => Promise<string>;
    decrypt: (address: Address, ciphertext: Buffer) => Promise<Buffer>;
}
export interface WritableWallet extends Wallet {
    isAccountUnlocked: (address: string) => boolean;
    unlockAccount: (address: string, passphrase: string, duration: number) => Promise<boolean>;
    addAccount: (privateKey: string, passphrase: string) => Promise<string>;
}
```

The writable wallet supports adding accounts to it and unlocking accounts. Essentially it supports modifying it's state (writing) while the base Wallet is read-only. Do you have any suggestions name wise?

packages/mobile/src/geth/GethNativeBridgeSigner.ts
bowd: Good point! Yeah I blindly copied that over from the `RpcSigner`, I'll fix in both places.

packages/mobile/src/web3/contracts.ts
bowd: You good point and it is dangerous. The danger was there before as well but after I refactored some stuff I helped typescript see the danger as well. I'll massage things upstream and add some more checks to remove the need for this.

packages/mobile/src/dappkit/dappkit.ts
bowd: I ended up reverting this change after diving deeper into the `CeloProvider`, my current understanding is that the wallet will still be used by `CeloProvider` as long as `wallet.hasAccount(address)` (local account) is true. So the functionality will be as expected here.

packages/mobile/src/geth/GethNativeBridgeSigner.ts
```diff
@@ -0,0 +1,110 @@
...
+   * latency and timing inconsistencies on the node
+   * @param unlockTime Timestamp in seconds when the signer was last unlocked
+   * @param unlockDuration Number of seconds that the signer was last unlocked for
+   */
+  constructor(
+    protected geth: GethNativeModule,
+    protected account: string,
+    protected unlockBufferSeconds = 5,
+    protected unlockTime = -1,
+    protected unlockDuration = -1
```
jeanregisser: These are all in seconds, but are they required to be integral? or can you pass numbers with a decimal part?

packages/mobile/src/web3/contracts.ts
```diff
@@ -98,54 +114,88 @@ export function* initContractKit() {
...
+      throw new Error('Contract kit initialisation timeout')
+    }
+  }
+  return contractKit
+}
+
+export function* getContractKit() {
+  if (!contractKit) {
+    if (initContractKitLock) {
+      yield call(waitForContractKit, 10)
```
jeanregisser: nit: `10` could be a constant to better explain what it is.

packages/mobile/src/web3/contracts.ts
```diff
@@ -98,54 +114,88 @@ export function* initContractKit() {
...
+  }
   return contractKit
 }
 
 export function* getWallet() {
-  if (!gethWallet) {
-    yield call(initContractKit)
+  if (!wallet) {
+    if (initContractKitLock) {
+      yield call(waitForContractKit, 10)
```
jeanregisser: nit: same here about `10` being a magic number.

packages/mobile/src/invite/saga.ts
jeanregisser: Naming is hard 😄 
I don't have a perfect name for now, one that comes to mind is `MutableWallet`, but still very similar.

jeanregisser (approved):
Looking great! 👍 

Seeing these changes, the geth/contract kit initialization feels relatively complex with timeouts which could be prone to breaking in the future.
I wonder how we could simplify this more, but I'm certainly not expecting to do this now of course 😸

packages/mobile/__mocks__/@celo/contractkit/index.ts
```diff
@@ -1,3 +1,4 @@
+import { GoogleStorageUtils } from '@celo/contractkit/lib/network-utils/google-storage-utils'
```
jmrossy: Why import this? It doesn't seem to be used in this file

packages/mobile/src/geth/GethNativeBridgeSigner.ts
```diff
@@ -0,0 +1,110 @@
...
+   * @param unlockBufferSeconds Number of seconds to shrink the unlocked duration by to account for
+   * latency and timing inconsistencies on the node
+   * @param unlockTime Timestamp in seconds when the signer was last unlocked
+   * @param unlockDuration Number of seconds that the signer was last unlocked for
+   */
+  constructor(
+    protected geth: GethNativeModule,
+    protected account: string,
+    protected unlockBufferSeconds = 5,
+    protected unlockTime = -1,
```
jmrossy: Not a big deal but if possible consider using undefined or null instead of -1 to represent these not yet being set

packages/mobile/src/geth/GethNativeBridgeSigner.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  async signPersonalMessage(data: string): Promise<{ v: number; r: Buffer; s: Buffer }> {
+    Logger.info(`${TAG}@signPersonalMessage`, `Signing ${data}`)
+    const hash = ethUtil.hashPersonalMessage(Buffer.from(data.replace('0x', ''), 'hex'))
+    const signatureBase64 = await this.geth.signHash(hash.toString('base64'), this.account)
+    return ethUtil.fromRpcSig(this.base64ToHex(signatureBase64))
+  }
+
+  getNativeKey = () => this.account
```
jmrossy, line 74: what's this for? Stands out cause its a lamdba among all these class methods

packages/mobile/src/geth/GethNativeBridgeSigner.ts
```diff
@@ -0,0 +1,110 @@
...
+    return Promise.reject(new Error('Decryption operation is not supported on this signer'))
+  }
+
+  hexToBase64(hex: string) {
+    return Buffer.from(hex.replace('0x', ''), 'hex').toString('base64')
+  }
+
+  base64ToHex(base64: string) {
+    return ensureLeading0x(Buffer.from(base64, 'base64').toString('hex'))
+  }
```
jmrossy, line 112: We already have utils for these in the kit I believe

packages/mobile/src/geth/GethNativeBridgeWallet.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  async loadAccountSigners(): Promise<Map<string, GethNativeBridgeSigner>> {
+    let accounts: string[]
+    const addressToSigner = new Map<string, GethNativeBridgeSigner>()
+
+    try {
+      accounts = await this.geth.listAccounts()
+    } catch (e) {
+      Logger.error(`${TAG}@loadAccountSigners`, e.toString())
```
jmrossy: Logger.error accepts the error as the third param and it'll handle formatting it. Try this instead:
`Logger.error(`${TAG}@loadAccountSigners`, 'Error listing accounts', e)`

packages/mobile/src/geth/GethNativeBridgeWallet.ts
```diff
@@ -0,0 +1,95 @@
...
+import { GethNativeModule } from 'react-native-geth'
+import { GethNativeBridgeSigner } from 'src/geth/GethNativeBridgeSigner'
+import Logger from 'src/utils/Logger'
+import { Tx } from 'web3-core'
+
+export enum GethNativeBridgeWalletErrors {
+  FetchAccounts = 'GethNativeBridgeWallet: failed to fetch accounts from geth wrapper',
+  AccountAlreadyExists = 'GethNativeBridgeWallet: account already exists',
+  UnexpectedAddressOnAdd = 'GethNativeBridgeWallet: unexpected address returned from addAccount',
+}
```
jmrossy: Instead of defining a new Error enum here, consider adding these to the existing `ErrorMessages` enum in mobile.

packages/mobile/src/geth/saga.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    MockGethBridge.startNode.mockClear()
+  })
+
+  it('initializes the bridge and starts the node', async () => {
+    const state = createMockStore({}).getState()
+    await expectSaga(_waitForGethInit)
+      .withState(state)
+      .returns(GethInitOutcomes.SUCCESS)
+      .run()
+    expect(state.geth.initialized).toEqual(InitializationState.INITIALIZED)
```
jmrossy, line 28: Checking for the saga to `put` certain actions is more inline with how expectSaga is meant to be used but if this works then it's fine too :)

packages/mobile/src/geth/saga.ts
```diff
@@ -127,19 +126,22 @@ function* waitForGethInstance() {
       case FailedToFetchGenesisBlockError:
         return GethInitOutcomes.NETWORK_ERROR_FETCHING_GENESIS_BLOCK
       default: {
+        throw error
```
jmrossy: Did you mean to throw here? Makes the net two lines unreachable

packages/mobile/src/web3/contracts.ts
```diff
@@ -1,33 +1,54 @@
...
 const TAG = 'web3/contracts'
 const KIT_INIT_RETRY_DELAY = 2000
 const CONTRACT_KIT_RETRIES = 3
 
-let ipcProvider: IpcProvider | undefined
-let gethWallet: RpcWallet | undefined
+let wallet: WritableWallet | undefined
 let contractKit: ContractKit | undefined
 
+function* initWallet() {
```
jmrossy: Nit: Seems like this doesn't need to be a saga (a generator)

packages/mobile/src/web3/contracts.ts
```diff
@@ -98,54 +114,88 @@ export function* initContractKit() {
...
-  ipcProvider = undefined
+  wallet = undefined
 }
 
-export function* getContractKit() {
-  if (!contractKit) {
-    yield call(initContractKit)
+let initContractKitLock = false
+
+function* waitForContractKit(tries: number) {
```
jmrossy: Why create a generator for this? Seems like it just dupes the existing waitForContractKitAsync. Note, async functions can be called from generators.

jmrossy (commented):
Mostly looks good, just some minor issues

packages/mobile/src/geth/GethNativeBridgeSigner.ts
bowd: This functionality is pretty much copied from what's happening in the `RpcSigner`, and by looking at the behaviour I feel like using sub-second values here would result in unexpected behaviour because when verifying duration the current time is `floored` to the second.

packages/mobile/src/invite/saga.ts
bowd: Something else I was considering was maybe renaming the base and have:
- `Wallet extends ReadOnlyWallet` or
- `Wallet extends StatelessWallet` or
- `Wallet extends ImmutableWallet`

What do you think?

packages/mobile/src/geth/GethNativeBridgeSigner.ts
bowd: Most of these are copied from the `RpcSigner` because I wanted to mimic its functionality, but I agree it's better to be explicit so I'll fix these here and in the `RpcSinger` as well. It might make sense to actually abstract all of this in a super class that both signers extend to DRY things up a bit, but it might also intermingle things more, I have a rule of thumb to only start DRYing things after the 3rd repetition :D

packages/mobile/src/geth/GethNativeBridgeSigner.ts
bowd: Yeah again this is the same case, it's also in the `RpcSigner` written the same and I wanted to keep the interface identical. Not 100% sure about where it's used and I didn't question the lambda format - though that could be a style guide for oneliners.

packages/mobile/src/geth/GethNativeBridgeSigner.ts
bowd: Can you point me in the right direction - I looked around but couldn't find them :(

packages/mobile/src/geth/GethNativeBridgeWallet.ts
bowd: I followed the same pattern in the `RpcWallet`, but I'll move them, the only thing is that the `ErrorMessages` doesn't look like it has full blown messages it's more of a const enum.

packages/mobile/src/geth/saga.ts
bowd: Good catch - debugging leftover

packages/mobile/src/web3/contracts.ts
bowd: Yup you're right I had a bit of a confusion there and was under the impression that there's an `async generator` thing that needs to happen but you can just `yield call(asyncFn)` 🤦 - will fix

packages/mobile/src/geth/GethNativeBridgeWallet.ts
jmrossy: The error messages should correspond to i18n'ed strings in the localized json files if they ever get shown to users.

packages/mobile/src/geth/GethNativeBridgeSigner.ts
jmrossy: Nevermind I was mistaken, sorry

jmrossy (approved):
Nice work!

packages/mobile/src/invite/saga.ts
jeanregisser: Good idea! To me looks like `ReadOnlyWallet` could be the better option.

jeanregisser (approved):
Thanks for taking the extra time to make things right :100:

Lss-Ankit:
Hi @jeanregisser I have verified task using Test flight build v1.1.0 (21) and Android play store build v1.1.0 (1004294309). 

**Tested Scenarios:**
- Send payment and Request payment after enabling the data saver toggle. 
-  Buy/Sell CELO with enabling the data saver toggle. 
- Withdraw CELO with enabling the data saver toggle. 

Please let me know if i need to verify any thing else. 
Thanks!

PR #4642, opened 2020-08-06
[cli] Recover 32byte Valora accounts

gastonponti:
### Description

Recover the private key of an account created with a seed of 32bytes.
This helps the users that had balance in old Valora accounts, to recover their tokens.


gastonponti:
Please re-check the wording. This won't be for public use, but I tend to invent words, hahaha

jmrossy (approved):
This LGTM but can you include more info about exactly when this account is needed and also a TODO comment to clean this up in a month or two.

PR #4643, opened 2020-08-06
Parallelizes protocol tests

alecps:
### Description

Breaks protocol tests into seperate CI jobs to speed up the workflow and mitigate effects of flakiness. Because the governance protocol tests take by far the longest (~50 minutes) they are also broken down into separate jobs. 

### Other changes

Adds support for running protocol tests by subfolder (ex. `yarn test identity/` runs all tests under identity subfolder) 

### Tested

Testing infrastructure change so passing the existing tests is the best test we have. 

### Related issues

- NA

### Backwards compatibility

Not applicable


.circleci/config.yml
```diff
@@ -8,11 +8,11 @@ reference:
   android_config: &android_config
     working_directory: *workspace
     macos:
-      xcode: "11.5.0"
+      xcode: '11.5.0'
```
m-chrzan: Were these changes necessary?

m-chrzan (commented):
Just looking at the GitHub UI, doesn't look like these checks are required to pass for a PR to merge, which I believe they should?

.circleci/config.yml
alecps: No just automatic formatting by my editor. I'll get rid of them

packages/protocol/runTests.js
```diff
@@ -72,7 +72,9 @@ async function test() {
 
     const testGlob =
       argv._.length > 0
-        ? argv._.map((testName) => `test/\*\*/${testName}.ts`).join(' ')
+        ? argv._.map((testName) =>
+            testName.endsWith('/') ? `test/${testName}*.ts` : `test/\*\*/${testName}.ts`
```
gastonponti: Mmm, check this line.
Probably is `test/${testName}**/*.ts`
Otherwise you won't be running subfolders

.circleci/config.yml
```diff
@@ -872,7 +927,19 @@ workflows:
             - lint-checks
       - end-to-end-mobile-test-android
       - end-to-end-mobile-test-ios
-      - protocol-test:
+      - protocol-test-common:
+          requires:
```
gastonponti, line 959: Just a question, but this will run the same lint in every parallelised batch run, or it's just running once?
(even if it runs multiple times it better than what we have, hahaha, so great!)

alecps: This just means that the job won't run until the 'lint-checks' job passes. So we're only running the linter once

packages/protocol/runTests.js
alecps: > Just looking at the GitHub UI, doesn't look like these checks are required to pass for a PR to merge, which I believe they should?

Great catch. I'll look into how that's configured

alecps:
> Just looking at the GitHub UI, doesn't look like these checks are required to pass for a PR to merge, which I believe they should?

Great catch. I'll look into how that's configured

gastonponti (approved):

PR #4646, opened 2020-08-07
[Wallet] Add CELO withdrawal flow

### LINKED ISSUE #4408: [Wallet] Allow transfers of CELO to other addresses
### Expected Behavior

Can send CELO in wallet

### Current Behavior

Can only send cUSD or exchange cUSD for CELO in wallet

gnardini:
### Description

Adds a button to withdraw CELO from Valora. It can also be used to transfer CELO to other users.

There are many file changes, but most are quite small. The important part is in the main screens of the flow `WithdrawCeloScreen` and `WithdrawCeloReviewScreen` and in the `exchange/saga`. There's also a new `CeloTransferFeedItem` for the activity feed of the CELO screen.

### Tested

Screen Shot 2020-08-07 at 05 49 55
Screen Shot 2020-08-07 at 05 50 21
Screen Shot 2020-08-07 at 05 50 28
Screen Shot 2020-08-07 at 05 50 45

celo-withdrawal

### Related issues

- Fixes #4408


packages/mobile/src/exchange/actions.ts
```diff
@@ -10,6 +10,7 @@ export enum Actions {
   EXCHANGE_TOKENS = 'EXCHANGE/EXCHANGE_TOKENS',
   FETCH_TOBIN_TAX = 'EXCHANGE/FETCH_TOBIN_TAX',
   SET_TOBIN_TAX = 'EXCHANGE/SET_TOBIN_TAX',
+  WITHDRAW_CELO = 'EXCHANGE/WITHDRAW_CELO',
```
gnardini, line 13: Should this `WITHDRAW_CELO` action and it's saga go here or on the `transactions` folder?

packages/mobile/src/exchange/WithdrawCeloScreen.test.tsx
```diff
@@ -0,0 +1,100 @@
+import * as React from 'react'
```
gnardini, line 1: I wrote in Slack about this and people seem to have different opinions on it, but I wasn't sure if I should use `Celo` or `Gold` to refer to the currency. I used Celo when I started working, so I just kept using that, but I have no strong opinion on this and can change it if anyone has.

packages/mobile/src/components/ClipboardAwarePasteIcon.tsx
```diff
@@ -0,0 +1,77 @@
+// VIEW Paste icon that disappears when the |currentValue| passed matches the content
```
aaronmgdr, line 1: Isnt there another PR right now to do with Paste Aware icons? 

any chance  things can be combined or reused?

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
```diff
@@ -0,0 +1,183 @@
...
+    navigation.navigate(Screens.WithdrawCeloReviewScreen, {
+      amount: celoAmount,
+      recipientAddress: accountAddress,
+    })
+  }, [accountAddress, celoToTransfer])
+
+  const setMaxAmount = useCallback(() => {
+    if (goldBalance) {
+      setCeloToTransfer(goldBalance)
+    }
```
aaronmgdr: could some of this logic be moved into a custom hook? 

call be old fashion but I think that its easier to follow function when they are fewer lines

packages/mobile/locales/en-US/sendFlow7.json
```diff
@@ -15,10 +15,12 @@
   "celoDollarAmount_plural": "<0></0> {{CeloDollars}}",
   "addDescription": "Add description",
   "sentTo": "Sent To",
+  "withdrawnTo": "Withdrawn to",
```
i1skn, line 18: Can we use `withdrawnTo` from `exchangeFlow9.json` here or vice versa?

packages/mobile/src/components/ClipboardAwarePasteIcon.tsx
i1skn: Yeap, I've recently landed that PR. PTAL here https://github.com/celo-org/celo-monorepo/blob/4773390237b68dc71c14f85d60a4bca197b33468/packages/react-components/components/WithTextInputPasteAware.tsx on how we can reuse Paste functionality.

packages/mobile/src/exchange/WithdrawCeloQrScannerScreen.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return JSON.parse(data).address
+  } catch (error) {
+    return null
+  }
+}
+
+function WithdrawCeloQrScannerScreen({ route }: Props) {
+  const { onAddressScanned } = route.params
+
+  const onBarCodeDetected = useCallback(
```
i1skn: I know that it has been this way initially in `<QRScanner />`, but I think we can remove `useCallback` here, no?

packages/react-components/icons/Paste.v2.tsx
```diff
@@ -0,0 +1,58 @@
+import colors from '@celo/react-components/styles/colors'
```
i1skn: You can use https://github.com/celo-org/celo-monorepo/blob/master/packages/react-components/icons/Paste.tsx - it is the same right now (I've just removed `Mask`, cause it seems to have no effect on the UI).

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const { t } = useTranslation(Namespaces.exchangeFlow9)
+
+  const celoToTransferNumber = new BigNumber(celoToTransfer)
+  const readyToReview =
+    accountAddress.startsWith('0x') &&
+    accountAddress.length === 42 &&
+    celoToTransferNumber.isGreaterThan(0) &&
+    celoToTransferNumber.isLessThanOrEqualTo(goldBalanceNumber)
+
+  const onConfirm = useCallback(() => {
```
i1skn: I think here and in the next three functions we can remove `useCallback`

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
i1skn: These are very small functions, I doubt, that moving the logic from them somewhere else would benefit readability.

packages/mobile/src/exchange/actions.ts
i1skn: I think it would feel more natural in `transactions`, cause it's essentially a transfer and not an exchange, but having said that I do not have a strong opinion on this.

packages/mobile/src/exchange/saga.ts
```diff
@@ -338,6 +339,68 @@ function* createStandbyTx(
...
+        value: amount.toString(),
+        symbol: CURRENCY_ENUM.GOLD,
+        timestamp: Math.floor(Date.now() / 1000),
+        address: recipientAddress,
+      })
+    )
+
+    setTimeout(() => {
+      navigateToExchangeScreenIfReviewing()
+    }, 3000)
```
i1skn: Why 3 seconds and why do we call `navigateToExchangeScreenIfReviewing` again after some timeout?

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -88,6 +88,31 @@ export function navigate<RouteName extends keyof StackParamList>(
     })
 }
 
+export function navigateIf<RouteName extends keyof StackParamList>(
```
i1skn: Do I understand correct, that the reason to have this method is that we do not want to call `ensureNavigator` somewhere else in the code if we need to check current navigation state?

packages/mobile/src/navigator/QRNavigator.tsx
```diff
@@ -88,10 +92,33 @@ function AnimatedScannerScene({ position, ...props }: AnimatedScannerSceneProps)
   // react-native-camera.
   const enableCamera = isFocused || (isPartiallyVisible && (hasAskedCameraPermission || wasFocused))
 
+  const dispatch = useDispatch()
+  const { scanIsForSecureSend, isOutgoingPaymentRequest, transactionData, requesterAddress } =
+    route.params || {}
+
+  const onBarCodeDetected = useCallback(
```
i1skn: The same as above

packages/mobile/src/transactions/CeloTransferFeedItem.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const isWithdrawal = new BigNumber(amount.value).isNegative()
+
+  return (
+    <Touchable onPress={onPress}>
+      <View style={styles.container}>
+        <View style={styles.firstRow}>
+          <View style={styles.desc}>
+            <Text style={styles.txMode}>
+              {t(isWithdrawal ? 'feedItemGoldWithdrawal' : 'feedItemGoldReceived', {
+                address: `${address.slice(0, 6)}...${address.slice(-4)}`,
```
i1skn: Do you think we can use https://github.com/celo-org/celo-monorepo/blob/4773390237b68dc71c14f85d60a4bca197b33468/packages/mobile/src/components/ShortenedAddress.tsx#L9-L11 here?

packages/react-components/components/SettingsItem.tsx
```diff
@@ -53,6 +53,21 @@ export function SettingsItemTextValue({
   )
 }
 
+type SettingsItemTextProps = BaseProps
```
i1skn: I think we can refactor `<SettingsItemTextProps>` above and add something like `showChevron` there instead of creating a new component? Wdyt?

i1skn (commented):
Absolutely great job on implementing this feature! See my minor comments/questions down below.

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
gnardini: I moved the inputs and input logic to custom components, I think it looks better now :)

packages/mobile/locales/en-US/sendFlow7.json
gnardini: Yes, the one from `exchangeFlow` wasn't actually being used, forgot to delete it, thanks!

packages/mobile/src/exchange/saga.ts
gnardini: I talked to Corey about this, after this screen the `ExchangeHomeScreen` shows an item as 'Confirming' until the transaction is confirmed, we hoped to give it a few seconds to confirm before going back in hopes that in many cases the user doesn't see the 'Confirming' text, but on second thought I don't think it makes much sense because it won't cover all cases and it adds unnecessary complexity to the code imo, I'll talk to him to see if it's ok to remove this.

packages/mobile/src/navigator/NavigationService.ts
gnardini: Yes, it was. It was related to the `navigateToExchangeScreenIfReviewing` above, but if we remove that we can remove this too.

packages/mobile/src/transactions/CeloTransferFeedItem.tsx
gnardini: Problem is that since this is an object if we pass it to the `t` function is shows as `[Object object]`. I did however extract the formatting function and reused it so that we don't have the formatting code repeated :)

packages/mobile/src/components/ClipboardAwarePasteIcon.tsx
gnardini: Refactored to use `withPasteAware`, thanks for pointing it out 😄

packages/mobile/src/components/CeloAmountInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+      inputStyle={inputStyle}
+      placeholder={'0'}
+      placeholderTextColor={colors.gray3}
+      keyboardType={'decimal-pad'}
+      onChangeText={onCeloChanged}
+      value={celo}
+      testID={'CeloAmount'}
+    >
+      <TouchableOpacity testID={'MaxAmount'} onPress={setMaxAmount}>
+        <Text style={[styles.maxAmount, { color }]}>{'Max'}</Text>
```
i1skn: Is `Max` language agnostic text?

gnardini: Good question, I assumed so but I guess not, I'll move to translation files.

packages/mobile/src/exchange/WithdrawCeloReviewScreen.tsx
```diff
@@ -0,0 +1,125 @@
...
+          textStyle={fontStyles.regular}
+          amount={<ShortenedAddress style={styles.withdrawAddress} address={recipientAddress} />}
+        />
+        <HorizontalLine />
+        <WithdrawCeloSummary amount={amount} recipientAddress={recipientAddress} />
+      </View>
+      <Button
+        onPress={onConfirmWithdraw}
+        text={t(`global:withdraw`)}
+        disabled={false}
```
i1skn: nit: `disabled` can be removed here

i1skn (approved):
Great work! :shipit:

packages/mobile/e2e/Transfer.spec.js
```diff
@@ -195,5 +196,39 @@ describe('Transfer Works', () => {
     // make sure it was successful
   })
 
+  it('Wallet Home -> Exchange CELO', async () => {
```
annakaz, line 199: ✨

packages/mobile/src/exchange/WithdrawCeloScreen.test.tsx
annakaz: 👍  I agree with using Celo, it will encourage a move away from Gold

annakaz (approved):
LGTM. Great work adding comprehensive testing and analytics.

Issue #4609, opened 2020-08-03
Rename PGPNP to ODIS


jmrossy:
### Expected Behavior

ODIS -> Oblivious Decentralized Identifier Service

### Current Behavior

PGPNP -> Pretty Good Phone Number Privacy

PR #4649, opened 2020-08-07
PGPNP to ODIS rename Part 2 [Signer and Combiner]

jmrossy:
### Description

Rename PGPNP and salt references in combiner and signer to ODIS and pepper

### Related issues

- #4609 

### Backwards compatibility

Yes


packages/phone-number-privacy/combiner/README.md
```diff
@@ -1,3 +1,3 @@
-## Phone Number Privacy
+## ODIS Combiner
 
-A firebase function that generates unique salts for phone numbers.
+A firebase function that generates orchestrates distributed BLS threshold signing with the set of ODIS signers.
```
alecps:
```suggestion
A firebase function that orchestrates distributed BLS threshold signing with the set of ODIS signers.
```

alecps (approved):

packages/docs/celo-codebase/protocol/identity/phone-number-privacy.md
```diff
@@ -1,57 +1,61 @@
...
...mised or a new ODIS operator is added, it will be necessary to perform a key rotation. Before going over the key rotation process, let's take a look at the implications of a compromised key. If the number of keys compromised at a time is less than the threshold _k_, the attacker will not be able to reach a sufficient threshold to compute the pepper for all phone numbers. Similarly, as long as _k_ other keys remain uncompromised, good users will still be able to perform the pepper lookup as part of the PNP service. Therefore, in the case that a single key is compromised, user data will remain private and the service operational; however, it's important that we can detect and perform a key rotation before the number of keys compromised exceeds _k_ or _m - k + 1_ (whichever is lower). For example, if there are ten PNP operators and the required threshold is three, then if three of them are compromised an attacker may compute the pepper for all phone numbers. If eight are compromised then an attacker may prevent the rest of the nodes (two in this case) from generating the pepper for users. Note that "compromised" entities in these examples could also be malicious or simply unavailable.
```
codyborn:
```suggestion
In the case that a key is compromised or a new ODIS operator is added, it will be necessary to perform a key rotation. Before going over the key rotation process, let's take a look at the implications of a compromised key. If the number of keys compromised at a time is less than the threshold _k_, the attacker will not be able to reach a sufficient threshold to compute the pepper for all phone numbers. Similarly, as long as _k_ other keys remain uncompromised, good users will still be able to perform the pepper lookup as part of the ODIS. Therefore, in the case that a single key is compromised, user data will remain private and the service operational; however, it's important that we can detect and perform a key rotation before the number of keys compromised exceeds _k_ or _m - k + 1_ (whichever is lower). For example, if there are ten ODIS operators and the required threshold is three, then if three of them are compromised an attacker may compute the pepper for all phone numbers. If eight are compromised then an attacker may prevent the rest of the nodes (two in this case) from generating the pepper for users. Note that "compromised" entities in these examples could also be malicious or simply unavailable.
```

codyborn: Changing `PNP` to `ODIS`

codyborn (approved):
Looks good, thanks Rossy!

PR #4650, opened 2020-08-07
Add Celo base package (no dependencies utils)

hbandura:
In an effort to reduce dependency bloating for `celo/utils` clients, this PR splits it into two parts. The `celo/base` package is added, which holds all utils code without any external dependencies. The premise is that it will always be safe to add the `celo/base` package as a dependency, since it will never depend from anything else (this is even enforced by a unit test checking the package.json file).

To maintain backwards compatibility with current `celo/utils` clients, the parts migrated to base were imported and exported in utils, as in:

```
import { SomeSymbol1, SomeSymbol2, ... } from '@celo/base/lib/___'
export { 
 SomeSymbol1,
 ...
} from '@celo/base/lib/___'
```

Additionally, some references to `celo/utils` from `celo/contractkit` were redirected to the `celo/base` package. This prevents `celo/contractkit` clients from bundling unnecessary dependencies that were brought in by unrelated `celo/utils` code (e.g Bip39)

Relevant unit tests for the migrated parts were copied to the `celo/base` package, but still maintained in `celo/utils` to ensure backwards compatibility is not an issue.


packages/base/jest.config.js
```diff
@@ -0,0 +1,7 @@
+const { nodeFlakeTracking } = require('@celo/flake-tracker/src/jest/config.js')
```
gastonponti, line 1: Mmmm, I think that this file should be also in the `.npmignore`, right? (maybe is wrong in all the other packages too)

packages/utils/src/account.ts
```diff
@@ -1,40 +1,23 @@
...
-}
+// Exports moved to @celo/base, forwarding them
+// here for backwards compatibility
+export {
+  Bip39,
+  CELO_DERIVATION_PATH_BASE,
+  MnemonicLanguages,
+  MnemonicStrength,
+  RandomNumberGenerator,
+} from '@celo/base/lib/account'
```
gastonponti, line 20: I'm adding it here, but it's more of an `base/account` vs `utils/account thing`
I thought that the idea, was going to add all the handling of a library in the base, and have all the utils with just populating these "base/functions"

For example, in these case:
this function (I couldn't add a comment in the line, sorry 😞 ):
```
export async function generateKeys(
  mnemonic: string,
  password?: string,
  changeIndex: number = 0,
  addressIndex: number = 0,
  bip39ToUse: Bip39 = bip39Wrapper,
  derivationPath: string = CELO_DERIVATION_PATH_BASE
): Promise<{ privateKey: string; publicKey: string }> {
  const seed = await bip39ToUse.mnemonicToSeed(mnemonic, password)
  const node = bip32.fromSeed(seed)
  const newNode = node.derivePath(`${derivationPath}/${changeIndex}/${addressIndex}`)
  if (!newNode.privateKey) {
    // As we are generating the node from a seed, the node will always have a private key and this would never happened
    throw new Error('utils-accounts@generateKeys: invalid node to derivate')
  }
  return {
    privateKey: newNode.privateKey.toString('hex'),
    publicKey: newNode.publicKey.toString('hex'),
  }
}
```

I though that the idea was to have something like this in base:
```
export async function generateKeys(
  mnemonic: string,
  password?: string,
  changeIndex: number = 0,
  addressIndex: number = 0,
  bip39: Bip39,  // we already have this Bip39 interface defined in the base
  bip32: Bip32,  // we will need to define this one (I think that only need the `fromSeed` method, that returns a node that should have the `derivePath` method
  derivationPath: string = CELO_DERIVATION_PATH_BASE
): Promise<{ privateKey: string; publicKey: string }> {
  const seed = await bip39.mnemonicToSeed(mnemonic, password)
  const node = bip32.fromSeed(seed)
  const newNode = node.derivePath(`${derivationPath}/${changeIndex}/${addressIndex}`)
  if (!newNode.privateKey) {
    // As we are generating the node from a seed, the node will always have a private key and this would never happened
    throw new Error('utils-accounts@generateKeys: invalid node to derivate')
  }
  return {
    privateKey: newNode.privateKey.toString('hex'),
    publicKey: newNode.publicKey.toString('hex'),
  }
}
```
So basically that function is requiring 2 libs that have those functions, but it's not downloading any dependency for that

and then having something like this in the `utils`:
```
import * as bip32 from 'bip32'
import * as bip39 from 'bip39'
import { generateKeys as utilsGenerateKeys } from '@celo/base/lib/accounts'

export async function generateKeys(
  mnemonic: string,
  password?: string,
  changeIndex: number,
  addressIndex: number,
  bip39ToUse: Bip39 = bip39Wrapper, // if we want to keep this bip39Wrapper thing
  derivationPath?: string
): Promise<{ privateKey: string; publicKey: string }> {
  return utilsGenerateKeys(mnemonic, password, changeIndex, bip39ToUse, bip32, derivationPath) // the bip32 should also be extracted, but this will make the same response of the "old utils"
}
```

So how I see it, we make the `base` as how we are going to use external libraries, and the utils basically populates all those libraries (for the lazy one, haha)

What do you think?

packages/contractkit/src/wrappers/LockedGold.ts
```diff
@@ -13,6 +18,10 @@ import {
   valueToString,
 } from '../wrappers/BaseWrapper'
 
+type AddressListItem = ALI<BigNumber>
+// BigNumber comparator
+const bnc: Comparator<BigNumber> = (a: BigNumber, b: BigNumber) => a.lt(b)
```
gastonponti: I would add a more readable name, or you could define here a wrapper function of the `linkedListChanges` that already uses this comparator, and use that instead

packages/base/src/types.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'web3-utils'
+declare module '@umpirsky/country-list/data/es_AR/country.json' {
```
gastonponti: 👀

packages/utils/src/address.ts
```diff
@@ -7,30 +8,25 @@ import {
...
-  trimLeading0x(input).match(/.{1,4}/g) || []
-
-export const isHexString = (input: string) => HEX_REGEX.test(input)
-
-export const hexToBuffer = (input: string) => Buffer.from(trimLeading0x(input), 'hex')
-
-export const bufferToHex = (buf: Buffer) => ensureLeading0x(buf.toString('hex'))
+// Exports moved to @celo/base, forwarding them
+// here for backwards compatibility
+export {
```
gastonponti, line 13: you can make an `export * from @celo/base/lib/address` to export everything
Same with the others. This could probably make us avoid some problems in the future, when someone adds something in the base lib, but not in the utils

packages/utils/src/collections.ts
```diff
@@ -1,99 +1,24 @@
...
-
-// Warning: sortedList is modified
-function _linkedListChange(sortedList: AddressListItem[], change: AddressListItem) {
-  const idx = upsert(sortedList, change)
-  const greater = idx === 0 ? NULL_ADDRESS : sortedList[idx - 1].address
-  const lesser = idx === sortedList.length - 1 ? NULL_ADDRESS : sortedList[idx + 1].address
-  return { lesser, greater }
-}
+// BigNumber comparator
+const bnc: base.Comparator<BigNumber> = (a: BigNumber, b: BigNumber) => a.lt(b)
```
gastonponti: same as a comment from above

gastonponti (commented):
Check the comment of the use of libraries as parameters.
I'm also curious about how many utils function are still in contractKit that we could make this sort of, "just add the lib and consume from base", that will only pick what is necessary

packages/base/src/types.d.ts
hbandura: Hmmm copy paste gone wrong

packages/utils/src/account.ts
hbandura: Eventually, I'd like to have something like this for most of the utils package. But doing all of that in this PR would be an enourmous amount of work both for coding and reviewing. So my idea here is to make this 'small' pr where I set the foundation of the base package, some simple usages, and then do this kind of 'strategy' pattern migration step by step.

What you mention was already done for some functions (e.g linkedListChanges) as an example.

packages/base/jest.config.js
hbandura: If it has to be fixed in both packages, I'd rather do that in a different PR

packages/utils/src/address.ts
hbandura: This is exactly what I didn't want, but we can discuss if it makes sense or not. I explicitly exported just what it existed before to ensure backwards compatibility, but the idea is for people to use base, not utils, when possible. Therefore, after a while we could even delete this exports here (after carefully deprecating and searching for their usages). If a new function is added in base, I didn't want people to point to utils to use it. Does that make sense?

packages/contractkit/src/wrappers/LockedGold.ts
hbandura: done

packages/utils/src/collections.ts
hbandura: done

hbandura:
> Check the comment of the use of libraries as parameters.
> I'm also curious about how many utils function are still in contractKit that we could make this sort of, "just add the lib and consume from base", that will only pick what is necessary

Probably a lot! I only moved a few as an example. In a different PR I will continue to move references

packages/utils/src/address.ts
gastonponti: makes sense

gastonponti (approved):
Keep on destroooooying the utiiiilssss, ahah

PR #4654, opened 2020-08-10
Add decrypt to RPCSigner

nambrot:
### Description

Sister PR to https://github.com/celo-org/ganache-cli/pull/25 to support decryption in ContractKit's wallet interface


gastonponti (approved):
LGTM
Just remember, that this will allow the contractKit to consume an endpoint, that we are not 100% that it will be in the node (due to blockchain releases).

Don't know which service actually consume this, but if is required for something that already was in use, you should assume a possible error or add a feature flag for this. We faced the same problem with the `gasPrice` with different currency, that was used by the `params-normalizer` which it was used in every Tx and basically made us remove a package version because it was unusable.

Again, don't know this case, but it's just a warning

nambrot:
Thanks @gastonponti for the warning. Since there is no actual usage of this, I would consider this OK for now.

PR #4665, opened 2020-08-11
Signing fixes

codyborn:
### Description

This PR fixes some issues with the AWS signer:
1. It was failing to init the wallet since it was erring on non-ECC keys (there’s one in our KMS).
2. When calling `recoverKeyIndex`, it failed because `ecdsaRecover` expects the first byte of the public key to be `0x4` to indicate the key is not compressed.
3. Mock signer wasn't signing correctly. I ripped out the `crypto` library approach and used `ethUtil` instead. To support this, I added a new method `asn1FromPublicKey` for getting the key into the same format as KMS.
4. Mock object failure handling needed updating (ex. invalid key check),

### Other changes

Updated Azure Key Vault mock response to match the updated API.

### Tested

Tested AWS and Azure HSM wallets. For both I tested using the mock objects as well as actual deployed instances.

### Related issues

- Fixes #4275 

### Backwards compatibility

yes


packages/contractkit/src/utils/signing-utils.ts
```diff
@@ -243,3 +247,15 @@ export function recoverKeyIndex(
   }
   throw new Error('Unable to generate recovery key from signature.')
 }
+
+/**
+ * Maps the publicKey to its address
+ */
+export function getAddressFromPublicKey(publicKey: BigNumber): Address {
```
tkporter, line 256: There's already function (that takes a string instead of a bn) that does this in `packages/utils/src/address.ts`, not sure if it helps

packages/contractkit/src/wallets/aws-hsm-wallet.ts
```diff
@@ -29,20 +36,23 @@ export default class AwsHsmWallet extends RemoteWallet implements Wallet {
...
-          publicKeyToAddress(bigNumberToBuffer(publicKey, 64).toString('hex')),
-          new AwsHsmSigner(this.kms, KeyId!, publicKey)
-        )
+        const publicKey = await this.getPublicKeyFromKeyId(KeyId)
+        const address = getAddressFromPublicKey(publicKey)
+        addressToSigner.set(address, new AwsHsmSigner(this.kms, KeyId, publicKey))
       } catch (e) {
-        // todo: what does the error look like here
-        throw e
+        // Safely ignore non-secp256k1 keys
```
tkporter, line 55: should we log or something if this if condition isn't true?

What's the benefit of not throwing sometimes?

packages/contractkit/src/wallets/aws-hsm-wallet.ts
```diff
@@ -57,11 +67,20 @@ export default class AwsHsmWallet extends RemoteWallet implements Wallet {
       throw new Error('AwsHsmWallet needs to be initialised first')
     }
     const { PublicKey } = await this.kms.getPublicKey({ KeyId: keyId }).promise()
-    return publicKeyFromAsn1(Buffer.from(PublicKey as Uint8Array))
+    const pubKey = publicKeyFromAsn1(Buffer.from(PublicKey as Uint8Array))
+    const pkbuff = bigNumberToBuffer(pubKey, 64)
```
tkporter: 64 is hardcoded in a couple places (in and outside of this PR), is something that could be put into a constant?

tkporter (approved):

packages/contractkit/src/wallets/aws-hsm-wallet.ts
codyborn: Yes, good idea on logging. We don't want to throw the exception if it fails because of non-secp256k1 keys. If you try to call `getPublicKey` on a key that doesn't support it, then it will throw the `UnsupportedOperationException` exception. We can ignore this type of exception since it won't happen with secp256k1 keys. If it's another type of exception, it's unexpected and we should throw.

packages/contractkit/src/utils/signing-utils.ts
codyborn: Thanks for pointing this out.. we can and should definitely re-use this. Let me check this into Alex's branch and then I'll address this in his PR.

PR #4676, opened 2020-08-11
Fix undefined serialization in URIs

yorhodes:
### Description

Fixes in-app QR code generation for user sharing such that undefined URI data values (like currency code) are not included in the generated URI. 

### Other changes

### Tested

Added unit tests.

### Related issues

- Fixes scanning issues seen in onboarding

### Backwards compatibility

URIs are now always URL encoded for interoperability.


packages/mobile/src/qrcode/schema.ts
```diff
@@ -29,13 +29,15 @@ export const uriDataFromJson = (obj: object): UriData => {
...
 
 enum UriMethod {
   pay = 'pay',
 }
 
 export const urlFromUriData = (data: Partial<UriData>, method: UriMethod = UriMethod.pay) => {
-  const params = new URLSearchParams(Object(data))
-  return `celo://wallet/${method.toString()}?${params.toString()}`
+  const params = new URLSearchParams(
+    Object(Object.entries(data).filter(([k, v], _) => v !== undefined))
```
annakaz: Nit: Can you add a comment I find this difficult to read

annakaz (approved):
Approving to get this in the latest build / a fix out as soon as we can!

Have you tested on device? If not, I can test locally using two devices

annakaz:
Tested QR send on two devices 💯

packages/mobile/src/qrcode/schema.ts
jmrossy: +1, or consider breaking this apart.

PR #4678, opened 2020-08-11
Prevent nonce assignment collisions when sending multiple transactions

### LINKED ISSUE #1111: Error: Transaction was not mined within 50 blocks, please make sure your transaction was properly sent. ...
Sentry Issue: [CELO-MOBILE-1YF](https://sentry.io/organizations/celo/issues/1788077634/?referrer=github_integration)

```
Error: Transaction was not mined within 50 blocks, please make sure your transaction was properly sent. Be aware that it might still be mined!
  at callImmediates ([native code])
  at value ([native code])
  at value ([native code])
...
(8 additional frame(s) were not displayed)
```

nategraf:
### Description

When sending multiple transactions at once, if signing a transaction and adding it to the transaction pool are not done atomically, then multiple of the transactions may be assigned the same nonce. Currently we see this issue when trying to complete the verification flow, where 3 transactions are submitted in parallel, one for each attestation to be completed. https://github.com/celo-org/celo-blockchain/issues/1111

This PR adds a simple implementation of a mutex lock, and uses it to ensure that once a nonce is assigned to a transaction, no other transaction can have a nonce assigned until that transaction is in the node transaction pool, or has been rejected by the node.

### Other changes

* Changed initialization of Geth to allow for fetching static nodes file and genesis file in parallel.

### Tested

Added a unit test for the lock, and used a manual test recreating the error observed during verification completion to ensure that error was fixed.

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain/issues/1111

### Backwards compatibility

100%


i1skn (approved):
Great work on identifying and then fixing it!

packages/mobile/src/transactions/contract-utils.ts
```diff
@@ -172,7 +172,7 @@ export async function sendTransactionAsync<T>(
       feeCurrency: feeCurrencyAddress,
       // Hack to prevent web3 from adding the suggested gold gas price, allowing geth to add
       // the suggested price in the selected feeCurrency.
-      gasPrice: gasPrice ? gasPrice : '0',
+      gasPrice: gasPrice ?? '0',
```
asaj, line 175: You learn something new every day!

gastonponti (approved):
Nice!
I agree that this is waaaay better than what we have already. So, nice adding!
But the nonce only depends on the `from`, right? (maybe I'm wrong)

Couldn't we only acquire the lock of the specific `from`, having a map in the `lock`. Something like `acquire(from)`.
Otherwise this will lock every send, and could be from different addresses, with different `nonces`

As I said, either way, this is better than what we have, so it's your call.

packages/mobile/src/geth/geth.ts
```diff
@@ -123,12 +123,19 @@ export async function initGeth(sync: boolean = true): Promise<typeof gethInstanc
...
 
   try {
-    if (!(await ensureGenesisBlockWritten())) {
-      throw FailedToFetchGenesisBlockError
-    }
-    if (!(await ensureStaticNodesInitialized(sync))) {
-      throw FailedToFetchStaticNodesError
-    }
+    // Write static node and genesis block, canceling if either fail.
+    await Promise.all(
```
aslawson: nice :)

packages/contractkit/src/providers/celo-provider.ts
```diff
@@ -1,3 +1,4 @@
+import { Lock } from '@celo/utils/lib/lock'
```
aslawson, line 1: think you need to bump utils version and publish to npm

packages/utils/src/lock.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+
+      // Wait for an event emitted when releasing the lock.
+      const callback = () => {
+        if (this.tryAcquire()) {
+          this.emitter.removeListener(LockEvent.Unlock, callback)
+          resolve()
+        }
+      }
+      this.emitter.on(LockEvent.Unlock, callback)
```
aslawson, line 43: curious -- what happens if you have multiple requests calling `emitter.on`?  

if an unlock happens and all their callbacks are triggered, only one will successfully acquire, and it will remove the listener (hopefully not for all pending requests...).

packages/contractkit/src/providers/celo-provider.ts
nategraf: Do you know the most up to date process for this? When should this be published?

packages/utils/src/lock.ts
nategraf: Because the callback is created in the function call, the function (and it's closure) are identity unique, so only the callback for the one that will acquire the lock will be removed. (If this were not the case, the race in the unit test would deadlock.)

nategraf:
@gastonponti I started to work on adding a lock per address, and it can be done but there are some weird edge cases involved, such as if an account is accessed by address, and by index in parallel. Generally, I think we assume `contractkit` is being used by one or a small handful of accounts at any given time, so I would favor simplicity over this optimization.

packages/utils/src/lock.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  // Attempt to aquire the lock without blocking.
+  //
+  // @returns {boolean} True if the lock was aquired.
+  tryAcquire(): boolean {
+    if (!this.locked) {
+      this.locked = true
+      return true
+    }
+    return false
```
timmoreton, line 24: did you think about reentrancy behavior? right now it looks like this is non-reentrant but could be made reentrant here.

packages/utils/src/lock.ts
```diff
@@ -0,0 +1,54 @@
...
+      // Attempt to grab the lock without waiting.
+      if (this.tryAcquire()) {
+        resolve()
+        return
+      }
+
+      // Wait for an event emitted when releasing the lock.
+      const callback = () => {
+        if (this.tryAcquire()) {
+          this.emitter.removeListener(LockEvent.Unlock, callback)
```
timmoreton, line 39: could this ever fail? do you need `try`..`catch`?

packages/utils/src/lock.ts
```diff
@@ -0,0 +1,54 @@
+import { EventEmitter } from 'events'
+
+enum LockEvent {
+  Unlock = 'unlock',
+}
+
+// Lock which can be used to ensure mutual exclusion in concurrent code.
+export class Lock {
```
timmoreton, line 8: surprised there isn't an obvious library we could use for this?

timmoreton (approved):
LGTM

packages/utils/src/lock.ts
nategraf: Reentrancy would require something analogous to thread-local storage, which based on a quick search is implemented by `async_hooks` which is an experimental feature since Node 8. Given that it's not very simple to implement with the tools available, and there isn't a need for it at the moment, I would opt not to implement it right now.
https://nodejs.org/api/async_hooks.html

I did add a comment about it being non-reentrant though.

packages/contractkit/src/providers/celo-provider.ts
aslawson: [How to publish a new npm package
](https://github.com/celo-org/celo-monorepo/blob/f97e19fc1311dad65c1039ec97051f93fd7e937e/README-dev.md
)
IIUC, I think it's simpler since the references here are all suffixed with `-dev.` so just bumping them all in this rb and then publishing in npm should be fine.  there's only an issue if there's a mobile release and they remove the `dev` suffix before the npm gets published (which would result in a build failure for them).

packages/utils/src/lock.ts
nategraf: We could pull in a library, but I opted not to when I realized the [most popular implementation](https://www.npmjs.com/package/async-lock) has a big disclaimer about how the maintainer is not active 😕 . I figure it's not worth the incremental attack surface increase and we are better off having a simple implementation locally.

aslawson: I thought the same and was trying to google search for one 🤔

packages/utils/src/lock.ts
aslawson: >  and there isn't a need for it at the moment

why? are we assuming single threading...?

packages/utils/src/lock.ts
aslawson: i see -- so it just removed that callback rather than full listener.  makes sense 👍

aslawson: i see -- so it just removed that callback rather than full listener. makes sense 👍

packages/utils/src/lock.ts
nategraf: It's not needed because we can ensure when we use the lock that `lock.acquire` is never called from the "thread" that holds the lock. (Although it's true that we only have one true thread in Node, so what the "thread" I am referring to here is an `async`/`await` chain of execution. I am not sure what the real name is for it.) (As a reference, reentrancy is not a required feature. Golang, for instance, [does not implement it](https://stackoverflow.com/questions/14670979/recursive-locking-in-go))

PR #4680, opened 2020-08-12
Mobile discovery

tkporter:
### Description

Should not be merged until https://github.com/celo-org/react-native-geth/pull/28 is merged (which itself relies upon https://github.com/celo-org/celo-blockchain/pull/1132), and then the react-native-geth version in this PR will be updated.

This uses the enodes that are used for static nodes as bootnodes for discovery. The current bootnodes that are deployed do not support the type of discovery required by mobile clients (discovery v5), but the static nodes do. Having more nodes as bootnodes will also help with discovery speed & diversity of the nodes discovered.

When testing on baklava with both static nodes and discovery enabled, peering was nearly immediate. When only testing with discovery enabled and no static nodes, peering was very quick (< a few seconds). A few times I found when only relying upon discovery that no peers would be found, so I recommend keeping static nodes & discovery for the time being.

### Other changes

I added some instructions on how to attach to the geth instance on the app.

### Tested

Ran a lot on android & iOS devices

### Backwards compatibility

Requires updates from react-native-geth (https://github.com/celo-org/react-native-geth/pull/28) and celo-blockchain (https://github.com/celo-org/celo-blockchain/pull/1132) to work as intended


nategraf (approved):

packages/mobile/src/geth/geth.ts
```diff
@@ -274,6 +308,7 @@ function getStaticNodesFile(nodeDir: string) {
 async function writeStaticNodes(nodeDir: string, enodes: string) {
   console.info(`writeStaticNodes enodes are "${enodes}"`)
   const staticNodesFile = getStaticNodesFile(nodeDir)
+  console.info(`writeStaticNodes static nodes file is ${staticNodesFile}`)
```
annakaz: Nit: Logger instead of console, though I realize the old line above sets the wrong example
```suggestion
  Logger.info('Geth@ writeStaticNodes',`static nodes file is ${staticNodesFile}`)
```

packages/mobile/src/geth/geth.ts
```diff
@@ -78,24 +78,44 @@ function getFolder(filePath: string) {
...
     ipcPath: IPC_PATH,
+    noDiscovery: !useDiscovery,
+  }
+
+  if (useDiscovery) {
+    Logger.debug('Geth@newGeth', 'Using discovery, bootnodes = ' + bootnodeEnodes)
+    gethOptions.bootnodeEnodes = bootnodeEnodes
+  }
+
+  if (GETH_START_HTTP_RPC_SERVER) {
```
annakaz: Can include `if (__DEV__)` check to ensure this is only used in dev mode

packages/mobile/src/config.ts
```diff
@@ -79,3 +79,8 @@ export const MOONPAY_PUBLIC_KEY = keyOrUndefined(
   'MOONPAY_PUBLIC_KEY'
 )
 export const MOONPAY_RATE_API = `https://api.moonpay.io/v3/currencies/celo/price?apiKey=${MOONPAY_PUBLIC_KEY}`
+
+export const GETH_USE_FULL_NODE_DISCOVERY = true
```
annakaz: For env specific variables (for example, if this would not work on `alfajores` but would on `mainnet`), we use the mobile .env , .env.mainnet, .env.alfajores files for this configuration: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/.env

For more feature flag type variables that are constant across networks, we use: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/flags.ts

Would these flags fit in one of those files instead? I think .env is most natural as that's where other geth-related flags like `DEFAULT_SYNC_MODE` and `FORNO_ENABLED_INITIALLY` live

packages/mobile/src/geth/geth.ts
```diff
@@ -182,25 +213,28 @@ export function isProviderConnectionError(error: any) {
...
-    return true
+    throw FailedToFetchStaticNodesError
   }
-  return false
+}
+
+// Writes static nodes to the correct location
+async function initializeStaticNodesFile(staticNodes: string[]): Promise<void> {
+  const { nodeDir } = networkConfig
+  Logger.debug('Geth@ensureStaticNodesInitialized', 'initializing static nodes')
```
annakaz: Nit: tag should be updated
```suggestion
  Logger.debug('Geth@initializeStaticNodesFile', 'initializing static nodes')
```

packages/mobile/src/geth/geth.ts
```diff
@@ -274,6 +308,7 @@ function getStaticNodesFile(nodeDir: string) {
 async function writeStaticNodes(nodeDir: string, enodes: string) {
   console.info(`writeStaticNodes enodes are "${enodes}"`)
```
annakaz:
```suggestion
  Logger.info('Geth@writeStaticNodes',`enodes are "${enodes}"`)
```

annakaz (approved):
Excited to have this enabled in Valora!

Left some nits, but LGTM. Let me know if there's anything I can do to be helpful in getting this merged in time for the 1.1.0 release cut

annakaz (approved):
✨

PR #4686, opened 2020-08-12
PGPNP to ODIS rename Part 3 [ContractKit and part of Mobile]

jmrossy:
### Description

Rename PGPNP references in ContractKit to ODIS
This required corresponding updates in Mobile as well, but mobile still has some 'salt' references' that we should update too at some point.

### Test

Did full onboarding and send flow in wallet

### Related issues

- #4609 

### Backwards compatibility

Yes


packages/mobile/src/identity/privateHashing.ts
```diff
@@ -79,7 +80,7 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
     account,
     selfPhoneHash
   )
-  yield put(updateE164PhoneNumberSalts({ [e164Number]: details.salt }))
+  yield put(updateE164PhoneNumberSalts({ [e164Number]: details.pepper }))
```
codyborn, line 83: Is it possible to rename this? I assume we'll need to reference both for backward compatibility until everyone has upgraded to the latest version

codyborn (approved):
I'm sure this wasn't easy.. thanks!

packages/mobile/src/identity/privateHashing.ts
jmrossy: @codyborn Definitely, i plan handle the remaining 'salt' references in mobile next. There's quite a few, would have bloated this PR too much

PR #4688, opened 2020-08-12
added cUSD documentation

henry-zhang-bohan:
cUSD documentation: https://docs.google.com/document/d/1YuLM1tyKwX7k7-TmHG_qIMZ97jfuHr1F8bxSKT2mpBA/edit?usp=sharing


packages/docs/celo-holder-guide/cusd.md
```diff
@@ -0,0 +1,50 @@
...
+```
+celocli exchange:celo --value <VALUE-TO-EXCHANGE> --from $CELO_ACCOUNT_ADDRESS
+```
+
+## Transfer cUSD
+
+When you have sufficient balance, you can send cUSD to other accounts. Note that the unit of value is cUSD Wei (1 cUSD = 10^18 cUSD Wei).
+
+```
+celocli transfer:dollars --from $CELO_ACCOUNT_ADDRESS --to <RECIPIENT-ADDRESS> --value <VALUE-TO-TRANSFER>
```
yorhodes, line 48: I just added `--usdGas` so you can pay for fees in cUSD but it doesn't work with ledger yet

henry-zhang-bohan: @yorhodes would most users need to add this option?

yorhodes: if they don't have any CELO they would need it

YazzyYaz (approved):

PR #4689, opened 2020-08-12
adds timestamp and retries to ODIS salt request in contractkit and wallet

alecps:
### Description

Adds timestamp to odis pepper query. This timestamp is used by the signer as part of the replay attack prevention scheme. 

### Other changes

adds retries to odis pepper query

### Tested

e2e using emulator

### Related issues

- https://github.com/celo-org/celo-monorepo/issues/4465

### Backwards compatibility

Yes


packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
```diff
@@ -97,27 +98,39 @@ export async function postToPhoneNumPrivacyService<ResponseType>(
...
+    if (res.ok) {
+      debug('Response ok. Parsing.')
+      const response = await res.json()
+      return response as ResponseType
+    }
+
     debug(`Response not okay. Status ${res.status}`)
+
     switch (res.status) {
       case 403:
```
alecps: Would appreciate feedback on this part. Are there any other error codes we want to immediately throw after (and not retry) ?

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
```diff
@@ -97,27 +98,39 @@ export async function postToPhoneNumPrivacyService<ResponseType>(
...
-      Authorization: authHeader,
-    },
-    body: bodyString,
-  })
-
-  if (!res.ok) {
+
+  const tries = 3
+  let lastError
+  for (let i = 1; i <= tries; i++) {
```
jmrossy: Why do we need retry logic? are we seeing intermittent failures?

packages/mobile/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celo/mobile",
-  "version": "1.8.2",
+  "version": "1.8.3-dev",
```
jmrossy: Don't change the mobile package version

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
```diff
@@ -34,6 +34,7 @@ export interface PhoneNumberPrivacyRequest {
 
 export interface SignMessageRequest extends PhoneNumberPrivacyRequest {
   blindedQueryPhoneNumber: string
+  timestamp: number
```
jmrossy: If this isn't made optional, i think this will break the mobile app which currently doesn't include it. You'd have to update that too or make it optional for now.

codyborn: Good call out. I think we should make it optional to make it backward compatible with the existing mobile apps in the wild.

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
codyborn: This is more preemptive since network calls commonly fail. We couldn't add it before without potentially consuming the user's quota. Now with Alec's change, we can safely retry with the same timestamp without worrying about quota consumption.

codyborn: There's a retry method in our utils library that we should use:
https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/async.ts#L35
More code re-use and it has a configurable back-off. This prevents our server from getting pounded with traffic whenever there's a drop in availability (which can sometimes make the problem worse).

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
alecps: If the mobile app is using these changes won't it also include the change above in `phone-number-identifier.ts` that sets` timestamp: Date.now() ? `

alecps: I agree its probably better to have this optional just trying to understand everything fully

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
codyborn: Today, we also return `400` (user input error), `401` (unauthenticated), and `500` (unknown server error). I think `400` and `401` we should not retry, since they're guaranteed to fail.

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
jmrossy: I'm not sure polluting the code with retry logic is necessary here, we would usually reserve it for cases where the system is known to be flaky which I think isn't the case for ODIS. As far as general network flakiness goes, TCP should handle things mostly for us anyway. 
My concern is partly just to keep the code simple as possible and partly to avoid noise in partner's signers until they get the update.

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
jmrossy: @alecps ah I missed that bit in `phone-number-identifier.ts`. In that case we should be fine so long as the combiner/signers keep it optional.

packages/contractkit/src/utils/phone-number-lookup/phone-number-lookup.ts
alecps: Right I looked at that. That method won't allow us to automatically throw on certain error codes. Is that okay or should I implement a new helper?

packages/utils/src/async.test.ts
```diff
@@ -23,6 +23,23 @@ describe('retryAsync()', () => {
   })
 })
 
+describe('selectiveRetryAsyncWithBackOff()', () => {
```
jmrossy, line 26: No need to test this here and in base

jmrossy (approved):

PR #4691, opened 2020-08-12
Add usdGas flag to base command

### LINKED ISSUE #4544: [CLI] Add command to change default fee token to cUSD
### Expected Behavior

Users SBAT send cUSD from CLI without needing CELO to pay for tx fees. This can be done by setting the default fee token to cUSD (or to be more extensible, any token)

### Current Behavior

Users need to have some CELO in the same account to be able to send txns even if they only wish to send cUSD.

yorhodes:
### Description

Should work seamlessly with `tx-params-normalizer` for locally signed transactions by populating `tx.feeCurrency`

### Other changes

### Tested

### Related issues

- Fixes #4544

### Backwards compatibility

### DIFF
--- a/packages/contractkit/src/kit.ts
+++ b/packages/contractkit/src/kit.ts
@@ -75,6 +75,7 @@ export interface NetworkConfig {
 export interface KitOptions {
   gasInflationFactor: number
   gasPrice: string
+  gasPriceSuggestionMultiplier: number
   feeCurrency?: Address
   from?: Address
 }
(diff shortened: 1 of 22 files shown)


packages/contractkit/src/kit.ts
```diff
@@ -75,6 +75,7 @@ export interface NetworkConfig {
 export interface KitOptions {
   gasInflationFactor: number
   gasPrice: string
+  gasPriceSuggestionMultiplier: number
```
gastonponti, line 79: I would add a TODO here too, for removing it when all the nodes support the gasPrice endpoint

packages/contractkit/src/kit.ts
```diff
@@ -335,10 +344,6 @@ export class ContractKit {
       gasPrice: this.config.gasPrice,
     }
 
-    if (this.config.feeCurrency) {
```
gastonponti, line 338: I think that this is helpful. Not everyone will set it as a default, but don't see why we should remove it.

gastonponti (approved):

PR #4695, opened 2020-08-13
adds warnings to ODIS signer response

alecps:
### Description

When a request is found to already exist in the signer's db or when the signer fails to increment the user's query count a successful (200) response is returned, but the error field is populated with a warning.

### Other changes

None

### Tested

Unit tests added. e2e tests passing

### Related issues

None

### Backwards compatibility

Yes


packages/phone-number-privacy/common/src/interfaces/error-utils.ts
```diff
@@ -20,4 +20,7 @@ export enum WarningMessage {
   EXCEEDED_QUOTA = 'CELO_PNP_WARN_03 QUOTA Requester exceeded service query quota',
   UNVERIFIED_USER_ATTEMPT_TO_MATCHMAKE = 'CELO_PNP_WARN_04 QUOTA Unverified user attempting to matchmake',
   DUPLICATE_REQUEST_TO_MATCHMAKE = 'CELO_PNP_WARN_05 QUOTA Attempt to request >1 matchmaking',
+  DUPLICATE_REQUEST_TO_GET_PARTIAL_SIG = 'CELO_PNP_WARN_06 BAD_INPUT Attempt to replay partial sig request',
+  FAILURE_TO_STORE_REQUEST = 'CELO_PNP_WARN_07 QUOTA Failed to store partial sig request',
+  FAILURE_TO_INCREMENT_QUERY_COUNT = 'CELO_PNP_WARN_08 QUOTA Failed to increment user query count',
```
jmrossy: I'd recommend moving these to Err. Though the don't fail requests, they are issues that signer operators should alert on immediately (as opposed to semi-expected issues like these other warnings).

packages/phone-number-privacy/signer/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "dependencies": {
     "@celo/contractkit": "0.4.5",
-    "@celo/phone-number-privacy-common": "1.0.6",
+    "@celo/phone-number-privacy-common": "1.0.7-dev",
```
jmrossy: Have you confirmed that the docker image still builds correctly when you use a dev version here?

packages/phone-number-privacy/signer/src/signing/get-partial-signature.ts
```diff
@@ -106,7 +125,8 @@ export async function handleGetBlindedMessagePartialSig(
     } else {
       signMessageResponse = signMessageResponseSuccess
     }
-    response.json(signMessageResponse)
+    logger.debug('Signature retrieval success')
+    response.status(200).json(signMessageResponse)
```
jmrossy: wasn't this the default without the `status(200)`?

packages/phone-number-privacy/signer/src/signing/get-partial-signature.ts
```diff
@@ -79,16 +79,35 @@ export async function handleGetBlindedMessagePartialSig(
...
+      didIncrementQueryCount = !!(await incrementQueryCount(account))
+      if (!didIncrementQueryCount) {
+        errorMsg = WarningMessage.FAILURE_TO_INCREMENT_QUERY_COUNT
+      }
+    }
+    if (!didStoreRequest) {
+      logger.debug('Did not store request.')
+    }
+    if (!didIncrementQueryCount) {
+      logger.debug('Did not increment query count.')
```
jmrossy: Can these be DRY'ed up with the two if statements above which check the same thing?

packages/phone-number-privacy/signer/package.json
alecps: Actually I know that it will _not_ build successfully haha. I'll change these versions once I've published the common package

packages/phone-number-privacy/signer/src/signing/get-partial-signature.ts
alecps: yes, was just trying to be explicit since we are populating the error field but am happy to switch it back :)

packages/phone-number-privacy/signer/src/signing/get-partial-signature.ts
alecps: I wrote it this way so that the logging will happen regardless of why the update didn't happen. That is, if the db already contained the request vs. the db insert just failed.

jmrossy: My intuition is that there may be a more concise way to structure this but it's really nbd :)

jmrossy (approved):

PR #4698, opened 2020-08-13
[Wallet] SetClock bugfix

gnardini:
### Description

When the clock is not synced with the real time, we redirect users to the `SetClock` screen. In that screen, we have a button to go to the date settings and show the user's timezone. 
- When fetching the device's timezone, we were accessing `i18n.language` which might not be initialized yet, added a null check just in case.
- Tried to make iOS go to the settings page as well, but it risks getting the app rejected by the App Store, so I left it commented out.

### Related issues

- Fixes Sentry crash https://sentry.io/organizations/celo/issues/1816186985


annakaz (approved):
Looks great thanks for fixing this!

Agree with the call to leave the iOS redirect commented out for now.

PR #4699, opened 2020-08-13
[Wallet] Fixed performance tracking bug for Android

tarikbellamine:
### Description
Fixed the bug that was preventing us from logging react/app load times on Android.

### Other changes
Will look into enabling iOS performance logging in separate PR.

### Tested
Tested on Android device and emulator.

### Related issues
Build's on @etuleu previous PR #4636 that worked only on emulators.

### Backwards compatibility
Yes


packages/mobile/src/app/App.tsx
```diff
@@ -52,32 +57,42 @@ BigNumber.config({
 // }
 
 export class App extends React.Component {
+  state: State = {
+    reactInitTime: undefined,
+    reactLoadTime: undefined,
+  }
+
+  appStartListener: EmitterSubscription = new NativeEventEmitter().addListener(
```
annakaz: Does this work on both Android and iOS? If only on Android can you leave a comment

packages/mobile/src/app/App.tsx
```diff
@@ -52,32 +57,42 @@ BigNumber.config({
...
+    let reactLoadDuration
+    let appLoadDuration
+
+    if (this.state.reactInitTime && this.state.reactLoadTime) {
+      const appLoadedTime = Date.now()
+      reactLoadDuration = (this.state.reactLoadTime - this.state.reactInitTime) / 1000
+      appLoadDuration = (appLoadedTime - this.state.reactInitTime) / 1000
     }
 
+    ValoraAnalytics.startSession(AppEvents.app_launched, {
```
annakaz, line 103: Thoughts on a separate app_loaded event with the loadDuration payload, to make it easier to query where those values are non-null? This would also allow for const assignment of reactLoadDuration and appLoadDuration

annakaz (approved):
LGTM, left a few suggestions

packages/mobile/src/app/App.tsx
tarikbellamine: This should work for both Android and iOS but we are only emitting an event from Android native code at the moment. The hope is that we can use this same listener to capture iOS events once we figure out how to emit an event in iOS native code.

jmrossy: Is there a chance that by the time React has created App and this listener is hooked on, we've missed the event? Should this go in index.js instead?

tarikbellamine: Good question - the event is emitted from `onReactContextInitialized` in MainActivity so I think that means the order is ok. We tested using a physical device and emulator without issue.

Is there an edge case you have in mind for why it might work on some devices but not others?

jmrossy: No particular edge case in mind, just thinking out loud.
If it works it works :)

PR #4705, opened 2020-08-14
[Wallet] Fix Language screen transitions

gnardini:
### Description

The `Language` screen is used in three places: 
- It's the initial one when one opens the app for the first time if the language cannot be inferred.
- Shows up when one presses the "English" button on the top-rigth of the first onboarding screen.
- From the settings menu once logged in.

In the first and third cases, it should use default animations. In the case of the screen that pops up when pressing the language button, it should use a bottom-up animation on iOS. Before this PR this animation was being used everywhere.

### Tested

On the simulator, since there was a bug recently regarding the initial Language screen, I made sure that the screen shows up correctly on a fresh install when there is no language by removing the language inference. The bottom-up animation only shows up when the language button is pressed on the onboarding flow.


annakaz (approved):
Looks great thanks for fixing this!

jmrossy (approved):

PR #4706, opened 2020-08-14
Script to deploy smart contract changes to a network via governance

asaj:
### Description

This PR introduces the `make-release` script, which reads a backwards compatibility report against the latest release generated by `check-version`, deploys any changed contracts to the specified network, and outputs a JSON of the proposal that should be submitted to governance in order to promote the release.

This script is part of a series of tools that are intended to make smart contract deployments as easy and safe as possible. These include:

1. *check-deployed*: Takes smart-contract source code and confirms that the specified network is running that code. Used to verify that the latest tagged release is in fact what has been released to the network.
2. *check-versions*: Takes two branches and checks that, given the smart contract versions and source code of branch A, branch B has correctly set the smart contract versions given the source code changes. Outputs a summary of all smart contract changes between branch A and B. Used to sanity check that the developers' understanding of all smart contract changes from A to B are fully understood, and to specify what contracts need to be upgraded in order to release the contracts in branch B onto a network that is currently running the contracts in branch A.
3. *make-release*: Takes the compatibility report output by *check-versions*, deploys the specified contracts, and outputs a governance proposal. Used to propose a release of the contracts in branch B onto a network that is currently running the contracts in branch A.
4. *verify-release*: Takes two branches and a proposal ID, creates a compatibility report with *check-versions*, and confirms that the on-chain governance proposal is the result of  *make-release* on branches A and B. Used by CELO holders to verify contract upgrade proposals before voting.

## Other changes

- Adds proxies for all linked libraries so that they can be upgraded in the future without having to upgrade contracts that link them. There is some special casing to redeploy all linked libraries with proxies in release 1, which should be removed after the release is made.
- Adds versions to linked libraries
- Flag rename in `check-versions`
- Use 'Test' suffix to denote test contracts everywhere
- Don't ignore linked list libraries in contract version check
  
### Tested
First, spin up a ganache network with the contracts/0 release:
```
git checkout release/contracts/0
yarn devchain generate-tar .tmp/devchain.tar.gz && yarn devchain run-tar .tmp/devchain.tar.gz
```

Then, populate the initialization data for DowntimeSlasher:
```
echo '{"DowntimeSlasher": ["0x000000000000000000000000000000000000ce10", 10, 5, 20]}' > initialize_data.json
```

Finally, run the script:
```
yarn make-release -a release/contracts/0 -b asaj/make-release -n development -i initialize_data.json -p proposal.json -r report.json
```

### Related issues

-Part of broader work towards #4426

### Backwards compatibility

That's the goal!


packages/contractkit/src/base.ts
```diff
@@ -26,29 +26,7 @@ export enum CeloContract {
...
-  'GovernanceApproverMultiSigProxy',
-  'GovernanceProxy',
-  'LockedGoldProxy',
-  'ReserveProxy',
-  'ReserveSpenderMultiSigProxy',
-  'StableTokenProxy',
-  'SortedOraclesProxy',
-  'RegistryProxy',
-]
+export const ProxyContracts = Object.keys(CeloContract).map((c) => `${c}Proxy`)
```
yorhodes, line 29: 💯

packages/protocol/contracts/common/proxies/AddressLinkedListProxy.sol
```diff
@@ -0,0 +1,6 @@
+pragma solidity ^0.5.3;
+
+import "../Proxy.sol";
+
+/* solhint-disable no-empty-blocks */
+contract AddressLinkedListProxy is Proxy {}
```
yorhodes, line 6: any reason why these need to be explicitly separate instead of just reusing `Proxy`?

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,219 @@
...
+  public get = (contract: string): string[] => {
+    return this.dependencies.has(contract) ? this.dependencies.get(contract) : []
+  }
+}
+
+class ContractAddresses {
+  constructor(public addresses: Map<string, Address>) {}
+  static async create(contracts: string[], registry: any) {
+    const addresses = new Map()
+    contracts.forEach(async (contract: string) => {
```
nambrot: This does not wait for the function in `forEach` which I don't think is what you want

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,219 @@
...
+        if (linksLibrariesWithoutChanges[i]) {
+          console.log(
+            `${c} links ${dependencies.get(c)} and needs to be upgraded to link proxied libraries.`
+          )
+        }
+      })
+      throw new Error('All contracts linking libraries should be upgraded in release 1')
+    }
+
+    const release = async (contractName: string) => {
```
nambrot, line 213: This is a pretty massive function, I would recommend splitting it up for readability purposes

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,219 @@
...
+      throw new Error(`Unable to find address for ${contract}`)
+    }
+  }
+
+  public set = (contract: string, address: Address) => {
+    this.addresses.set(contract, address)
+  }
+}
+
+module.exports = async (callback: (error?: any) => number) => {
```
nambrot, line 189: Might be out of scope, but it would be cool to have a `dryRun` option that outputs the transactions this script would do when not dry running

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,219 @@
...
+        const Contract = await artifacts.require(contractName)
+        await Promise.all(contractDependencies.map((d) => Contract.link(d, addresses.get(d))))
+
+        // This is a hack that will re-deploy all libraries with proxies, whether or not they have
+        // changes to them.
+        // TODO(asa): Remove `isLibrary` for future releases.
+        const isLibrary = Object.keys(linkedLibraries).includes(contractName)
+        const deployImplementation = Object.keys(report.contracts).includes(contractName)
+        const deployProxy =
+          deployImplementation && report.contracts[contractName].changes.storage.length
```
nambrot: Do you truly want to just be checking for the existence of the `length` property or assert a `length > 0`?

nambrot (commented):
Catching up right now, so please forgive any inaccurate statements.

If I understand correctly, the dilemma we are facing is that in `master`, changes to linked libraries imply that all core contracts that rely on the changes would need to be explicitly upgraded as well?

The solution that you have outlined here is to make all linked libraries proxied, so that changes to libraries can be done via repointing the library proxy instead of having to upgrade the core contracts itself.

I couldn't tell from the PR how exactly you changed the core contracts to access the linked library via proxies vs. directly?

Additionally, would you mind elaborating again why you felt that upgrading all linked libraries was favorable in your opinion? Vs. Having core contracts being able to explicitly rely on possibly older versions of libraries. I assume it is because the invariant of having the source code at a commit represents the contract bytecode in totality on a network?

asaj:
> If I understand correctly, the dilemma we are facing is that in `master`, changes to linked libraries imply that all core contracts that rely on the changes would need to be explicitly upgraded as well?

Correct, that is the implication. However, because changes to a library do not cause bytecode changes to any contracts that link that library, our current compatibility tool will not recognize this implication.  

> The solution that you have outlined here is to make all linked libraries proxied, so that changes to libraries can be done via repointing the library proxy instead of having to upgrade the core contracts itself.

Exactly
 
> I couldn't tell from the PR how exactly you changed the core contracts to access the linked library via proxies vs. directly?

There are two pieces to this.

The *first* is to ensure changes to all contracts that link libraries, so that the compatibility tool flags them as needing to be upgraded. There is a sanity check in `make-release.ts` that confirms that all contracts that link libraries are included in the compatibility report.

The *second* is to deploy the proxied libraries *before* any of the contracts that link them, and to explicitly link the contracts to those proxied libraries before deploying the contracts. 

> Additionally, would you mind elaborating again why you felt that upgrading all linked libraries was favorable in your opinion? Vs. Having core contracts being able to explicitly rely on possibly older versions of libraries. I assume it is because the invariant of having the source code at a commit represents the contract bytecode in totality on a network?

Correct, having core contracts rely on older versions of libraries breaks the invariant that the latest release branch represents exactly the contract code that is running on public networks.

packages/protocol/scripts/truffle/make-release.ts
asaj: Yeah I thought about that. It's a bit tricky since we depend on the addresses of the contracts we deploy. Let me see what I can do.

packages/protocol/scripts/truffle/make-release.ts
asaj: The latter, is `0` not falsey? Changing to > 0

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -107,6 +112,77 @@ module.exports = async (callback: (error?: any) => number) => {
       throw new Error('All contracts linking libraries should be upgraded in release 1')
     }
 
+    const deployImplementation = async (contractName: string, Contract: any) => {
```
nambrot: If possible I would suggest for these helper methods to move outside of the exported function, otherwise it makes following this function quite challenging

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,244 @@
...
+    const addresses = await ContractAddresses.create(contracts, registry)
+    const released: Set<string> = new Set([])
+    const proposal = []
+    // Release 1 will deploy all libraries with proxies so that they're more easily
+    // upgradable. All contracts that link libraries should be upgraded to instead link the proxied
+    // library.
+    // To ensure this actually happens, we check that all contracts that link libraries are marked
+    // as needing to be redeployed.
+    // TODO(asa): Remove this check after release 1.
+    const linksLibrariesWithoutChanges = contracts.map(
```
nambrot: I originally interpreted this as `linkedLibrariesWithoutChanges` but it sounds like its more like `contractLinksLibrariesWithoutChanges`

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,244 @@
...
+          await release(dependency)
+        }
+        // 2. Link dependencies.
+        const Contract = await artifacts.require(contractName)
+        await Promise.all(contractDependencies.map((d) => Contract.link(d, addresses.get(d))))
+
+        // This is a hack that will re-deploy all libraries with proxies, whether or not they have
+        // changes to them.
+        // TODO(asa): Remove `isLibrary` for future releases.
+        const isLibrary = Object.keys(linkedLibraries).includes(contractName)
```
nambrot: Can you not just directly check against the key here with `linkedLibraries[contractName]`?

packages/protocol/scripts/truffle/make-release.ts
nambrot: What do you think about also extracting this into its own function?

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,244 @@
...
+    // To ensure this actually happens, we check that all contracts that link libraries are marked
+    // as needing to be redeployed.
+    // TODO(asa): Remove this check after release 1.
+    const linksLibrariesWithoutChanges = contracts.map(
+      (c) =>
+        dependencies.get(c).length &&
+        !Object.keys(report.contracts).includes(c) &&
+        !c.endsWith('Test')
+    )
+    if (linksLibrariesWithoutChanges.some((b) => b)) {
```
nambrot: Optional Nit: I personally had trouble following the code, what do you think about this one

```typescript

function ensureAllContractsWithLibrariesAreUpgraded() {
  let anyContractViolates = false
  contracts.map((contract) => {
    const hasDependency = dependencies.get(contract).length > 0
    const specifiedToUpgrade = report.contracts[contract]
    const isTest = contract.endsWith("Test")

    if (hasDependency && !specifiedToUpgrade && !isTest) {
      console.log(
        `${contract} links ${dependencies.get(contract)} and needs to be upgraded to link proxied libraries.`
      )
      anyContractViolates = true
    }
  });

  if (anyContractViolates) {
    throw new Error('All contracts linking libraries should be upgraded in release 1')
  }
}


```

nambrot (commented):
Just minor nits

packages/protocol/scripts/truffle/make-release.ts
asaj: You tell me!

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+ *   network: The network for which artifacts should be
+ *
+ * Run using truffle exec, e.g.:
+ * truffle exec scripts/truffle/make-release \
+ *   --network alfajores --build_directory build/alfajores/ --report report.json \
+ *   --initialize_data initialize_data.json --proposal proposal.json
+ *
+ */
+
+class ContractDependencies {
```
yorhodes, line 25: I think there are some libraries for this like https://github.com/ConsenSys/surya
not suggesting we use it at this stage but good to be aware of in the future
or similarly publishing these scripts as our own npm package for truffle projects

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+    }
+  } else {
+    if (!dryRun) {
+      await proxy._setImplementation(contract.address)
+    }
+  }
+  // This makes essentially every contract dependent on Governance.
+  console.log(`Transferring ownership of ${contractName}Proxy to Governance`)
+  if (!dryRun) {
+    await proxy._transferOwnership(addresses.get('Governance'))
```
yorhodes, line 166: nit but we could pass in only the governance address to this function and fetch it once outside

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+    const report: ASTDetailedVersionedReport = readJsonSync(argv.report).report
+    const initializationData = readJsonSync(argv.initialize_data)
+    const dependencies = new ContractDependencies(linkedLibraries)
+    const contracts = readdirSync(join(argv.build_directory, 'contracts')).map((x) =>
+      basename(x, '.json')
+    )
+    const registry = await artifacts.require('Registry').at(REGISTRY_ADDRESS)
+    const addresses = await ContractAddresses.create(contracts, registry)
+    const released: Set<string> = new Set([])
+    const proposal = []
```
yorhodes: can we strongly type this?

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+        return
+      } else {
+        // 1. Release all dependencies.
+        const contractDependencies = dependencies.get(contractName)
+        for (const dependency of contractDependencies) {
+          await release(dependency)
+        }
+        // 2. Link dependencies.
+        const Contract = await artifacts.require(contractName)
+        await Promise.all(contractDependencies.map((d) => Contract.link(d, addresses.get(d))))
```
yorhodes, line 224: does this mutate the artifacts?
do we know if we store the artifacts linked or unlinked?

yorhodes (approved):
forgot to click submit review!

packages/protocol/contracts/common/proxies/AddressLinkedListProxy.sol
asaj: I think we wanted separate build artifacts for each. Happy to revisit this question outside the scope of this PR

packages/protocol/scripts/truffle/make-release.ts
asaj: Fortunately our migrations will fail if our hardcoded list of dependencies don't have complete coverage, so it's not as fragile as these sorts of things normally are.

packages/protocol/scripts/truffle/make-release.ts
asaj: Yes I believe it does. They start off unlinked, and then they get linked here

packages/protocol/contracts/common/Signatures.sol
```diff
@@ -3,6 +3,14 @@ pragma solidity ^0.5.3;
 import "openzeppelin-solidity/contracts/cryptography/ECDSA.sol";
 
 library Signatures {
+  /**
+   * @notice Returns the storage, major, minor, and patch version of the contract.
+   * @return The storage, major, minor, and patch version of the contract.
+   */
+  function getVersionNumber() external pure returns (uint256, uint256, uint256, uint256) {
```
m-chrzan, line 10: Are a deployed library's `external` functions normally callable? Now that I think about it, I guess they should be since that's how linking works, but just flagging this for a double check since I personally haven't thought about doing that to a library before.

packages/protocol/contracts/common/linkedlists/AddressLinkedList.sol
```diff
@@ -12,6 +12,14 @@ library AddressLinkedList {
   using LinkedList for LinkedList.List;
```
m-chrzan, line 12: Should this PR also be the one that makes nested libraries inlined rather than linked? Otherwise, we'll be wasting a lot of gas on Proxy + linked library indirection.

packages/protocol/contracts/common/proxies/AddressLinkedListProxy.sol
m-chrzan: Yeah, this definitely made things easier when relying on Truffle for contract deployments. I guess going forward we'll be doing upgrade deployments somewhat differently (using this PR's script), but keeping with this convention might still be good to keep things organized, keep track of which contracts are proxied, + we still use Truffle for any dev deployments/new testnets.

packages/protocol/contracts/identity/test/AttestationsTest.sol
```diff
@@ -7,7 +7,7 @@ import "../Attestations.sol";
  * but mocks the implementations of the validator set getters. Otherwise we
  * couldn't test `request` with the current ganache local testnet.
  */
-contract TestAttestations is Attestations {
+contract AttestationsTest is Attestations {
```
m-chrzan, line 10: Hmm, definitely a major nit, but to me `XTest` sounds like "file containing tests for X", `TestX` like "a version of X used for testing". Most of these files I think are the latter, so just wondering why the rename?

packages/protocol/scripts/bash/check-versions.sh
```diff
@@ -5,22 +5,28 @@ set -euo pipefail
...
+REPORT=""
 
-while getopts 'o:n:' flag; do
+while getopts 'a:b:n:r:' flag; do
   case "${flag}" in
-    o) BRANCH_1="${OPTARG}" ;;
-    n) BRANCH_2="${OPTARG}" ;;
+    a) BRANCH_1="${OPTARG}" ;;
+    b) BRANCH_2="${OPTARG}" ;;
+    n) NETWORK="$OPTARG" ;;
```
m-chrzan: Nit: `${OPTARG}` for consistency?

packages/protocol/scripts/bash/make-release.sh
```diff
@@ -0,0 +1,41 @@
...
+NETWORK=""
+REPORT=""
+
+while getopts 'a:b:n:p:i:r:' flag; do
+  case "${flag}" in
+    a) BRANCH_1="${OPTARG}" ;;
+    b) BRANCH_2="${OPTARG}" ;;
+    n) NETWORK="$OPTARG" ;;
+    p) PROPOSAL="$OPTARG" ;;
+    i) INITIALIZE_DATA="$OPTARG" ;;
```
m-chrzan: Ditto.

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * Expects the following flags:
+ *   report: The filepath of the backwards compatibility report
+ *   network: The network for which artifacts should be
+ *
+ * Run using truffle exec, e.g.:
+ * truffle exec scripts/truffle/make-release \
+ *   --network alfajores --build_directory build/alfajores/ --report report.json \
+ *   --initialize_data initialize_data.json --proposal proposal.json
+ *
```
m-chrzan: Extraneous newline

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+ */
+
+class ContractDependencies {
+  dependencies: Map<string, string[]>
+  constructor(libraries: any) {
```
m-chrzan: Can this be typed? Looks like it's a `{ [library: string]: string[] }` interface that could be defined somewhere.

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+    })
+  }
+
+  public get = (contract: string): string[] => {
+    return this.dependencies.has(contract) ? this.dependencies.get(contract) : []
+  }
+}
+
+class ContractAddresses {
+  static async create(contracts: string[], registry: any) {
```
m-chrzan: Can `registry` be typed? If using truffle, import `RegistryInstance` from `"types"`.

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+      anyContractViolates = true
+    }
+  })
+
+  if (anyContractViolates) {
+    throw new Error('All contracts linking libraries should be upgraded in release 1')
+  }
+}
+
+const deployImplementation = async (contractName: string, Contract: any, dryRun: boolean) => {
```
m-chrzan: Can `Contract` by typed with something from Truffle?

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+  // before or after other contracts in this script?), and that ordering is not being
+  // checked, fail if there are storage incompatible changes to Governance.
+  if (contractName === 'Governance') {
+    throw new Error(`Storage incompatible changes to Governance are not yet supported`)
+  }
+  console.log(`Deploying ${contractName}Proxy`)
+  const Proxy = await artifacts.require(`${contractName}Proxy`)
+  // Hack to trick truffle, which checks that the provided address has code
+  const proxy = await (dryRun ? Proxy.at(REGISTRY_ADDRESS) : Proxy.new())
+  const initializeAbi = (contract as any).abi.find(
```
m-chrzan, line 141: Isn't `contract` already `any`? Also, wondering if it could be typed with something from `truffle-typings`.

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+      await proxy._setImplementation(contract.address)
+    }
+  }
+  // This makes essentially every contract dependent on Governance.
+  console.log(`Transferring ownership of ${contractName}Proxy to Governance`)
+  if (!dryRun) {
+    await proxy._transferOwnership(addresses.get('Governance'))
+  }
+  const proxiedContract = await artifacts.require(contractName).at(proxy.address)
+  const transferOwnershipAbi = (contract as any).abi.find(
```
m-chrzan, line 169: Ditto.

packages/protocol/scripts/truffle/make-release.ts
m-chrzan: On a related note, though not related to this PR directly, have we decided how we're going to be managing artifacts within the release process?

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -0,0 +1,265 @@
...
+              ],
+              value: '0',
+              description: `Registry: ${contractName} -> ${proxy.address}`,
+            })
+          } else {
+            proposal.push({
+              contract: `${contractName}Proxy`,
+              function: '_setImplementation',
+              args: [contract.address],
+              value: '0',
```
m-chrzan, line 261: Do we also want a human readable description like above?

m-chrzan (approved):
Looks awesome! Left some comments, mostly nits.

packages/protocol/contracts/common/Signatures.sol
asaj: Believe so, yes

packages/protocol/contracts/common/linkedlists/AddressLinkedList.sol
asaj: Yeah good call, done

packages/protocol/contracts/identity/test/AttestationsTest.sol
asaj: Just to be consistent, happy to rename to TestX rather than XTest but would prefer not to block this PR on that

packages/protocol/scripts/truffle/make-release.ts
asaj: I think it's sufficiently readable, the problem with the registry repointing is that the registry ID is not human readable

packages/protocol/scripts/truffle/make-release.ts
asaj: Did it this way to cover the future case where the governance address might change due to a new governance proxy being deployed

PR #4711, opened 2020-08-14
Ensure every transaction has a unique ID for logging

nategraf:
### Description

Currently in the mobile application transactions are IDs when created, but in many cases these IDs are not unique or descriptive enough to provide context, or track an individual transaction.
In trying to track the performance of transaction, and diagnose errors from logging and analytics, this is a challenge.

This PR adds a new `TransactionContext` interface and ensures that every transaction has a unique ID and short description.

### Tested

Updated unit tests and manual testing of the application on my device.

### Backwards compatibility

Transaction ID format changed from a SHA256 hash to a UUID.


packages/mobile/package.json
```diff
@@ -72,13 +72,13 @@
...
     "dot-prop-immutable": "^1.4.0",
     "ethereumjs-util": "^6.2.0",
+    "fp-ts": "2.1.1",
     "fuzzysort": "^1.1.4",
     "google-libphonenumber": "^3.2.4",
     "graphql": "^14.1.1",
     "hermes-engine": "0.4.2-rc1",
     "i18next": "^19.4.3",
     "io-ts": "2.0.1",
-    "fp-ts": "2.1.1",
```
annakaz, line 81: 🙏 👨‍🎨  it's the little things, thank you for alphabetizing

packages/mobile/src/transactions/types.ts
```diff
@@ -26,6 +28,27 @@ export interface TransferStandby {
 
 export type StandbyTransaction = ExchangeStandby | TransferStandby
 
+// Context used for logging the transaction execution flow.
+export interface TransactionContext {
+  // Unique identifier used for tracking a transaction within logging.
+  // Note that this is not the transaction hash, which is unknown when creating a new transaction.
+  id: string
```
annakaz, line 35: Longer term could consider precomputing the transaction hash? If all fields are specified before we start tracking

annakaz (approved):
LGTM

packages/mobile/src/transactions/types.ts
nategraf: In cases where we really do know all the transaction fields on the outset, this would definitely be a nice way to handle tracking. It's likely that would be a rare case though because fields like `nonce` and `gas` require a chain state (or at least a model of chain state) to set correctly, so we currently rely on the node to fill them. It would also break our current abstraction layers to calculate the hash before signing, as that's delegated to the wallet interface. It's important that we can link the transaction hash and this local identifier, and the current method is that in the logs, and in the analytics events, the hash is included, alongside the ID, when it is received after signing as part of the `transaction_hash_received` event.

packages/mobile/package.json
nategraf: Don't thank me. Thank automation 🤖

Issue #4658, opened 2020-08-10
[Wallet] No error shown in CELO exchange flow if balance is not yet loaded 


annakaz:
### Expected Behavior

If balance is not yet loaded, show error so users understand why they cannot proceed with their CELO purchase

### Current Behavior

From App Store testing: 
"We discovered one or more bugs in your app when reviewed on iPad running iOS 13.6 on Wi-Fi.

Specifically, when we attempted to purchase “Celo” via the “Cell” tab, we could not proceed after adding an amount."
![attachment Screenshot-0810-122850](https://user-images.githubusercontent.com/16907027/89816346-45263600-dafb-11ea-9734-2e5badbc8c45.png)

annakaz:
I believe the user had 0 balance, so the "Buy" and "Sell" buttons in CELO home should have both been hidden. The issue may be that "Buy" is not being hidden when there is a 0 cUSD balance

tarikbellamine:
Addressed by #4714

PR #4714, opened 2020-08-14
[Wallet] Hide CELO buy button when there's no dollar balance

### LINKED ISSUE #4708: [Wallet] Users with 0 cUSD balance SNBAT see CELO buy button
### Expected Behavior

Buy button is hidden for users with 0 cUSD balance

### Current Behavior

Buy button is shown for users with 0 cUSD balance, but when they click buy they cannot proceed and encounter this issue https://github.com/celo-org/celo-monorepo/issues/4658 which led to the iOS build being rejected from the App Store 
release review

Screenshot_20200814-092706

gnardini:
### Description

When the user has 0 cUSD, the `buy` button on the CELO screen is now hidden. It's also hidden until the exchange rate is fetched, since user can't confirm the transaction until that happens.

### Other changes

Moved the buy and sell buttons to a new component.

### Tested

On a simulator by using a fresh account with 0 funds and then giving the account funds.

### Related issues

- Fixes #4708 & #4658


packages/mobile/src/stableToken/selectors.ts
```diff
@@ -0,0 +1,5 @@
+import { RootState } from 'src/redux/reducers'
```
annakaz: We have an existing a dollar balance selector here: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/stableToken/reducer.ts#L34
Can you use this function or deprecate it?

packages/mobile/src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -191,24 +167,7 @@ function ExchangeHomeScreen({ navigation }: Props) {
...
-              <Button
-                size={BtnSizes.FULL}
-                text={t('sell')}
-                onPress={goToBuyDollars}
-                style={styles.button}
-                type={BtnTypes.TERTIARY}
-              />
-            )}
-          </View>
+          <CeloExchangeButtons navigation={navigation} />
```
annakaz, line 170: 🧑‍🎨  thanks for cleaning this up!

packages/mobile/src/exchange/CeloExchangeButtons.tsx
```diff
@@ -0,0 +1,95 @@
...
+  navigation: StackNavigationProp<StackParamList, Screens.ExchangeHomeScreen>
+}
+
+export default function CeloExchangeButtons({ navigation }: Props) {
+  const { t } = useTranslation(Namespaces.exchangeFlow9)
+
+  const dollarBalance = useSelector(getDollarBalance)
+  const goldBalance = useSelector(getCeloBalance)
+
+  const hasDollars = new BigNumber(dollarBalance || 0).isGreaterThan(0)
```
annakaz: Could add a TODO here to replace with the minimum display or exchange transaction amount (which we don't have hard coded anywhere), for example if a user sees 0.00 but has 0.001 available it could be a confusing experience

packages/mobile/src/goldToken/selectors.ts
```diff
@@ -0,0 +1,5 @@
+import { RootState } from 'src/redux/reducers'
+
+export function getCeloBalance(state: RootState) {
```
annakaz: Nit: matter of style preference, but most of our selectors follow the () => {} syntax

annakaz (approved):
Minor comments but LGTM!

packages/mobile/src/exchange/CeloExchangeButtons.tsx
gnardini: Used `DOLLAR_TRANSACTION_MIN_AMOUNT` and `GOLD_TRANSACTION_MIN_AMOUNT`

PR #4723, opened 2020-08-17
Switch to use 64 bit emulator

cmcewen:
### Description

Running on an x86_64 emulator would crash due to some android syscall permissions. This fork fixes those issues (https://github.com/celo-org/sys) and enables the emulator to run much more quickly.


packages/mobile/scripts/sync_branding.sh
```diff
@@ -21,7 +21,7 @@ echo $mobile_root
 cd "$mobile_root"
 
 # Please update the sha when valora branding updates are needed
-valora_branding_sha=d58569f
+valora_branding_sha=2d60f89
```
jeanregisser, line 24: Why changing this in this PR?

jeanregisser (approved):
Awesome! 🚀

PR #4729, opened 2020-08-17
[mobile] add balance check and error message for requesting ODIS pepper

aslawson:
### Description

* Added balance check before making pepper request to ODIS since ODIS requires minimum balance of $0.10
* Right now in the verification flow, zero balance requesters for their own pepper results in a `saltQuotaExceededError` which can be misleading since their quota is calculated as 0.  The real issue is insufficient balance.  The `saltQuotaExceeded` state now encourages the user to buy more quota, but this actually wouldn't help here.  What they need is an increased balance.  So we should put them in `insufficientBalance` state.
* Wanted to differentiate InsufficentBalance errors to make it clear if the balance was insufficient for ODIS or the needed attestation requests (which can vary based on # of attestations needed). Plus, not all requesters actually need additional attestations 

* EG. A zero balance user that is already verified on-chain should be able to proceed through the flow (restore use case)
* EG. A user that has $0.07 and only needs one more attestation should be able to proceed even if that is less the ODIS requirement 

Current flow:
1. Get own pepper
2. Use pepper to see if already verified on chain
3. If verified — good, update app to show it.
4. If not verified — check for sufficient balance to make needed attestations.

### Other changes

NA

### Tested

No -- I'm not sure how

### Related issues

Issue described above.

### Backwards compatibility

Yes


packages/mobile/src/identity/privateHashing.ts
```diff
@@ -70,6 +72,9 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
   }
 
   Logger.debug(`${TAG}@fetchPrivatePhoneHash`, 'Salt was not cached, fetching')
+  if (!balanceSufficientForQuotaRetrieval()) {
```
annakaz: To wait for this function to resolve, should use `yield call(balanceSufficientForQuotaRetrieval)`

annakaz (commented):
Thank you for adding this! 

To run the test suite you can run `yarn run test identity`. It looks like there's a similar test here: https://github.com/celo-org/celo-monorepo/blob/amyslawson/errorCaseBalanceForQuota/packages/mobile/src/identity/verification.test.ts#L290

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -84,6 +89,11 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
   return details
 }
 
+export function* balanceSufficientForQuotaRetrieval() {
+  const userBalance = yield select(stableTokenBalanceSelector)
+  return new BigNumber(userBalance).isGreaterThanOrEqualTo(ODIS_MINIMUM_DOLLAR_BALANCE)
```
jmrossy, line 101: Please leave a TODO to check for CELO balance too. Once the signer update goes live, users with just CELO can query.

packages/mobile/src/config.ts
```diff
@@ -52,6 +52,8 @@ export const ESCROW_PAYMENT_EXPIRY_SECONDS = 86400 // 1 days
 export const DEFAULT_TESTNET = Config.DEFAULT_TESTNET || 'integration'
 export const DAILY_PAYMENT_LIMIT_CUSD = 500
 export const SMS_RETRIEVER_APP_SIGNATURE = Config.SMS_RETRIEVER_APP_SIGNATURE
+// ODIS minimum dollar balance for pepper quota retrieval
+export const ODIS_MINIMUM_DOLLAR_BALANCE = 0.1
```
jmrossy, line 57: Please leave a TODO to change this to 0.01 and add the CELO min balance one signer update goes out

packages/mobile/src/identity/privateHashing.ts
jmrossy: It would be better for this to live in the Contract kit ODIS utils but if we're in a hurry to get this merged feel free to just leave a TODO

packages/mobile/src/app/ErrorMessages.ts
```diff
@@ -27,6 +27,7 @@ export enum ErrorMessages {
   ODIS_QUOTA_ERROR = 'odisQuotaError',
   SALT_FETCH_FAILURE = 'saltFetchFailure',
   SALT_QUOTA_EXCEEDED = 'saltQuotaExceededError',
+  ODIS_INSUFFICIENT_BALANCE = 'odisInsufficientBalance',
```
jmrossy, line 30: These should correspond to strings in the localized JSON files otherwise if the error message is intercepted and changed (like is done for verification but not other places) the error would just show 'odisInsufficientBalance'

packages/mobile/src/identity/privateHashing.ts
jmrossy, line 101: Can you confirm the user balance is stored in dollars and not wei? It may be, I just can't remember.

aslawson: all other stableTokenBalanceSelector calls treat is as if it is stored in dollars

marekolszewski (commented):
lgtm

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -71,6 +73,10 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
   }
 
   Logger.debug(`${TAG}@fetchPrivatePhoneHash`, 'Salt was not cached, fetching')
+  const isBalanceSufficientForQuota = yield call(balanceSufficientForQuotaRetrieval)
+  if (!isBalanceSufficientForQuota) {
+    throw new Error(ErrorMessages.ODIS_INSUFFICIENT_BALANCE)
```
tarikbellamine, line 81: I don't think this error message is going to be propagated all the way to the verification flow because of the catch block of `fetchPhoneHashPrivate`. The way the catch block there handles errors from `doFetchPhoneHashPrivate` will only ever result in throwing with `SALT_QUOTA_EXCEEDED` or `SALT_FETCH_ERROR`. Probably easiest to add an `else if` clause to the catch block of `fetchPhoneHashPrivate` to handle then propagate `ODIS_INSUFFICIENT_BALANCE`.

packages/mobile/src/config.ts
nityas: Is it possible to have this not require a re-release once the change is made within ODIS?

tarikbellamine (dismissed):

packages/mobile/locales/en-US/global.json
```diff
@@ -122,6 +122,7 @@
   "addressLookupFailure": "Failed to look up phone number address",
   "saltFetchFailure": "Failed to retrieve phone number secret",
   "saltQuotaExceededError": "Your phone number lookup quota is used up",
+  "odisInsufficientBalance": "Your balance is too low to make this transaction",
```
nityas: nit: "Your balance is too low to complete this process" would be a bit more user friendly

jeanregisser (approved):
Awesome! 💯 
Did the following successful tests:
- ✅ popup is displayed when trying verification with a brand new account with balance < 0.1
- ✅ popup is displayed when trying verification with a restored account which was already verified, with balance < 0.1
- ✅ popup is NOT displayed when trying verification with a restored account which was already verified, with balance >= 0.1, it just checks the account is verified and quickly proceeds to the home screen
- ✅ normal verification works (balance needs to be > 0.15)

jmrossy (approved):

PR #4730, opened 2020-08-17
[Phone Number Privacy] Include celo utils as explicit dependency

annakaz:
### Description

Adding @celo/utils as explicit dependency in `phone-number-privacy/common` as it is used in input-validation.ts

Was seeing "src/utils/input-validation.ts(1,47): error TS2307: Cannot find module '@celo/utils/lib/address'." when running `yarn build`

### Tested

Ran yarn build

### Backwards compatibility

Yes


codyborn (approved):
Thanks!

Issue #4731, opened 2020-08-18
" QR code is invalid" validation message is not shown when user scan invalid QR code while withdrawing CELO


Lss-Ankit:
**Frequency:** 100%
**Repro App version:**  Valora Mainnet build v1.0.0 (11)
**Repro on device:** Google Pixel XL(8.1) , Vivo 1606 (6.0.1) 
**Test Account:** (680) 333-1978/ [Backup key. ](https://docs.google.com/spreadsheets/d/1Iy41GGA6e2pES2XHjBn7nb8X3n8zZsu7EW1jvwD2e70/edit#gid=0)
**Repro Steps:**
1. Launch the Application and sign in with test account
2. Tap on the Hamburger menu. 
3. Tap on CELO option
4. Tap on Withdraw CELO 
5. Tap on QR code icon. 
6. Scan invalid QR code. 
7. Observe. 

**Current Behavior:** User is redirected to the Withdraw CELO screen after scanning invalid QR code. 
**Expected Behavior:** QR code is Invalid message should be shown 
**Investigation:** Issue does not repro when user Scan invalid QR code while sending or requesting payment using QR code. 
**Impact:** User will not get idea why they back to the Withdraw CELO screen 
**Attachment:** [Android_QRCode_Validation_CELO.mp4](https://drive.google.com/file/d/1keV4jEE7eX13EgRh6rCWvsOkLdNoXIIg/view?usp=sharing) (Bug observed at 05 sec)

Lss-Ankit:
@annakaz , @gnardini verified issue with latest Test Flight build Valora app build  v1.0.0 (15) , Android Valora Mainnet build v1.0.0 (16) and found that issue is fixed as " QR code read failed: Reason: wallet address invalid" message is shown when user scan invalid QR code. 
Verified devices: iPhone 7+ (13.3.1), Google Pixel Xl (8.1)

PR #4740, opened 2020-08-18
[Wallet] Witdraw CELO QR scan fixes

### LINKED ISSUE #4732: Review button is not getting enabled when a user scans the QR code of another user while withdrawing CELO.
**Frequency:** 100%
**Repro App version:**  Valora Mainnet build v1.0.0 (11)
**Repro on device:** Google Pixel XL(8.1) , Vivo 1606 (6.0.1)  
**Test Account:** (680) 333-1978/ [Backup key.](https://docs.google.com/spreadsheets/d/1Iy41GGA6e2pES2XHjBn7nb8X3n8zZsu7EW1jvwD2e70/edit#gid=0) 
**Repro Steps:**
1. Launch the Application and sign in with test account
2. Tap on the Hamburger menu. 
3. Tap on CELO option
4. Tap on Withdraw CELO 
5. Tap on the QR code icon. 
6. Scan the valid QR code. 
7. Make sure other user’s Account no is shown in Account No field. 
8. Enter an amount in the Amount (CELO) field. 
9. Observe Review button. 

**Current Behavior:** Review button is not getting enabled
**Expected Behavior:** Review button should enable 
**Investigation:** Issue does not repro when the user sends CELO by pasting Account no. of other users. 
**Impact:** Users are blocked to proceed for sending CELO using QR code. 
**Attachment:** [Android_ReviewButton_CELO.mp4](https://drive.google.com/file/d/10lriPZNWf3hzLvLgHVTl5pYU5tC4ur43/view?usp=sharing) (Bug observe at 21 sec)

gnardini:
### Description

The withdraw CELO PR (https://github.com/celo-org/celo-monorepo/pull/4646) used the old QR format that was modified by another PR (https://github.com/celo-org/celo-monorepo/pull/4536/files). With the new format, scanning a QR is not returning the correct address. This PR fixes that and adds an error message if an invalid QR is scanned

### Tested

On an Android physical phone.

### Related issues

- Fixes #4732 and #4731

### DIFF
--- a/packages/mobile/src/exchange/WithdrawCeloQrScannerScreen.tsx
+++ b/packages/mobile/src/exchange/WithdrawCeloQrScannerScreen.tsx
@@ -6,13 +6,18 @@ import { componentStyles } from '@celo/react-components/styles/styles'
 import { StackScreenProps } from '@react-navigation/stack'
 import { memoize } from 'lodash'
 import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { useDispatch } from 'react-redux'
+import { showMessage } from 'src/alert/actions'
+import { ErrorMessages } from 'src/app/ErrorMessages'
 import BackButton from 'src/components/BackButton.v2'
-import i18n from 'src/i18n'
+import i18n, { Namespaces } from 'src/i18n'
 import { nuxNavigationOptions } from 'src/navigator/Headers.v2'
 import { navigateBack } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { StackParamList } from 'src/navigator/types'
 import QRScanner from 'src/qrcode/QRScanner'
+import { uriDataFromUrl } from 'src/qrcode/schema'
 import { QrCode } from 'src/send/actions'
 import Logger from 'src/utils/Logger'
 
@@ -22,7 +27,7 @@ type Props = StackScreenProps<StackParamList, Screens.WithdrawCeloQrScannerScree
 
 function fetchAddressFromQrData(data: string): string | null {
   try {
-    return JSON.parse(data).address
+    return uriDataFromUrl(data).address
   } catch (error) {
     return null
   }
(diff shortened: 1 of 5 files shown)


packages/mobile/src/exchange/WithdrawCeloQrScannerScreen.tsx
```diff
@@ -22,7 +27,7 @@ type Props = StackScreenProps<StackParamList, Screens.WithdrawCeloQrScannerScree
 
 function fetchAddressFromQrData(data: string): string | null {
   try {
-    return JSON.parse(data).address
+    return uriDataFromUrl(data).address
```
annakaz, line 30: 👍

packages/mobile/src/exchange/WithdrawCeloQrScannerScreen.tsx
```diff
@@ -31,13 +36,20 @@ function fetchAddressFromQrData(data: string): string | null {
...
       const qrData = qrCode.data
       Logger.debug(TAG, 'Bar code detected: ' + qrData)
       const address = fetchAddressFromQrData(qrData) || qrData
-      onAddressScanned(address)
-      navigateBack()
+      if (address.startsWith('0x') && address.length === 42) {
+        onAddressScanned(address)
+        navigateBack()
+      } else {
+        dispatch(showMessage(t(ErrorMessages.QR_FAILED_INVALID_ADDRESS)))
```
annakaz: Any reason to use showMessage over showError, so the red banner appears instead of blue?

packages/mobile/src/qrcode/utils.ts
```diff
@@ -103,7 +104,7 @@ export function* handleBarcode(
   try {
     qrData = uriDataFromUrl(barcode.data)
   } catch (e) {
-    yield put(showMessage(ErrorMessages.QR_FAILED_INVALID_ADDRESS))
+    yield put(showMessage(i18n.t(`global:${ErrorMessages.QR_FAILED_INVALID_ADDRESS}`)))
```
annakaz: Not a new change but could use showError instead of showMessage?

packages/mobile/src/exchange/WithdrawCeloQrScannerScreen.tsx
```diff
@@ -31,13 +36,20 @@ function fetchAddressFromQrData(data: string): string | null {
...
+  const dispatch = useDispatch()
+
   const onBarCodeDetected = memoize(
     (qrCode: QrCode) => {
       const qrData = qrCode.data
       Logger.debug(TAG, 'Bar code detected: ' + qrData)
       const address = fetchAddressFromQrData(qrData) || qrData
-      onAddressScanned(address)
-      navigateBack()
+      if (address.startsWith('0x') && address.length === 42) {
```
annakaz: Nit: defining 42 as address length constant to make it clear where it comes from

annakaz (approved):
Thanks for fixing this. Minor comment on using showError

I can cherry pick this commit to be included in the new build

Unrelated to changes in this PR but thinking out loud- right now CELO exchange providers Bittrex only allows for the address to be pasted (no QR), will want to make sure they use the same encoding scheme when they add QR support. cc @yorhodes

packages/mobile/src/exchange/WithdrawCeloQrScannerScreen.tsx
gnardini: No reason, I just copied it from the saga, I'll change it in both :)

PR #4742, opened 2020-08-18
Web Touchups

aaronmgdr:
### Description

note that these are "Early Backers"

update to the new cLabs Logo

### Tested

update snapshot where needed.


tarikbellamine (approved):
Send it!

PR #4747, opened 2020-08-19
CIP-21: Governable LookbackWindow Smart Contract changes

andres-dg:
### Description

* Implementation of [CIP-21](https://github.com/celo-org/celo-proposals/pull/145) smart contract changes.
* celo-blockchain related PR: https://github.com/celo-org/celo-blockchain/pull/1136

Adds the `uptimeLookbackWindow` to the `BlockchainParameters` contract. This is only ever used on the `celo-blockchain` client to determine the size of the uptime lookback window which is used to compute the validator [uptimeScore](https://docs.celo.org/celo-codebase/protocol/proof-of-stake/epoch-rewards/validator-rewards#calculating-uptime-score)

An important invariant, is that `UptimeLookbackWindow` must remain constant during an epoch. Thus, changes to the parameter only take effect on the next epoch. There's an exception to this rule during migration/deployment of this change; there `uptimeLookbackWindow` should be initialized to chain's current value for the parameter which is part of the `ChainConfig` on the genesis block. 

To initialize the value:
  * On a new testnet: As part of the migrations, `BlockchainParamets.initialize()` will do the initialization
  * On an existing testnet: After upgrading the contract, first call to `setLookbackWindow()` will the value for the current epoch.

Since `BlockchainParameters` won't be reinitialized when this change is deployed, a patch was made to the `getUptimteLookbackWindow()` contract to return `12` (mainnet, baklava & alfajores value for lookbackWindow) in the case of the value not being initalized. This patch can be removed in the future, since it's only necessary until `setLoobackWindow()` is called and first epoch since deployment has passed

### Tested

* Smart Contract unit tests


.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ e2e-defaults: &e2e-defaults
   docker:
     - image: celohq/circleci:geth1.9-2
   environment:
-    CELO_BLOCKCHAIN_BRANCH_TO_TEST: master
```
nambrot, line 34: This probably should be back to master right?

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -75,6 +88,15 @@ contract BlockchainParameters is Ownable, Initializable {
     emit IntrinsicGasForAlternativeFeeCurrencySet(gas);
   }
 
+  /**
+   * @notice Sets the uptime lookback window.
+   * @param window New window.
+   */
+  function setUptimeLookbackWindow(uint256 window) public onlyOwner {
```
nambrot, line 108: You should check that the input is within reasonable bounds

packages/protocol/test/governance/network/blockchainparams.ts
```diff
@@ -114,30 +116,58 @@ contract('BlockchainParameters', (accounts: string[]) => {
...
+    })
+    it('only owner should be able to set', async () => {
+      await assertRevert(
+        blockchainParameters.setUptimeLookbackWindow(lookbackWindow, {
+          from: accounts[1],
+        })
+      )
+    })
+  })
+
```
nambrot, line 186: Add a check for malformed input

packages/protocol/contracts/governance/BlockchainParameters.sol
prestwich: what would reasonable bounds be for this parameter?

andres-dg: The proposed limits, according to the [CIP](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0021.md), are [12 (1 minute) - 720 (1 hour)], I'm setting those for now.

Edit: This is assuming the block period is 5 seconds.

nambrot: I assume a window of 0 or 1 day should be impossible?

mcortesi: Added check as specified on https://github.com/celo-org/celo-proposals/pull/145/files

.circleci/config.yml
mcortesi: Not for the moment, since it's not merged there yet

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -75,6 +98,51 @@ contract BlockchainParameters is Ownable, Initializable {
...
+    require(window >= 12 && window <= 720, "UptimeLookbackWindow must be within safe range");
+    require(
+      window <= getEpochSize() - 2,
```
prestwich: edge case: if `epochSize` is 1 or 0 this will underflow

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -75,6 +98,51 @@ contract BlockchainParameters is Ownable, Initializable {
...
+      // we need to hardcode the default value here
+      // can be removed after the deploy
+      lookbackWindow = 12;
+    } else if (getEpochNumber() >= uptimeLookbackWindow.nextValueActivationEpoch) {
+      lookbackWindow = uptimeLookbackWindow.nextValue;
+    } else {
+      lookbackWindow = uptimeLookbackWindow.oldValue;
+    }
+
+    return lookbackWindow;
```
prestwich: nit: unnecessary return statement

packages/protocol/contracts/governance/BlockchainParameters.sol
nambrot: +1, use SafeMath

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -75,6 +98,51 @@ contract BlockchainParameters is Ownable, Initializable {
...
+    uptimeLookbackWindow.nextValue = window;
+
+    // epoch 0 is the genesis block, thus implies it has not been initialized
+    if (uptimeLookbackWindow.nextValueActivationEpoch == 0) {
+      // on initialize we set the value for the current epoch which configurer
+      // must make it so it's SAME as geth's chainConfig value
+      uptimeLookbackWindow.nextValueActivationEpoch = getEpochNumber();
+    } else {
+      // changes only take place on the next epoch
+      uptimeLookbackWindow.nextValueActivationEpoch = getEpochNumber() + 1;
```
nambrot: SafeMath

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -75,6 +98,51 @@ contract BlockchainParameters is Ownable, Initializable {
...
+    emit UptimeLookbackWindowSet(window, uptimeLookbackWindow.nextValueActivationEpoch);
+  }
+
+  /**
+   * @notice Gets the uptime lookback window.
+   */
+  function getUptimeLookbackWindow() public view returns (uint256 lookbackWindow) {
+    // epoch 0 is the genesis block, thus implies it has not been initialized
+    if (uptimeLookbackWindow.nextValueActivationEpoch == 0) {
+      // since on mainet,baklava,alfajores we don't reinitialize the contract
```
nambrot:
```suggestion
      // since on mainet,baklava,alfajores we haven't initialized the contract
```

packages/protocol/test/governance/network/blockchainparams.ts
```diff
@@ -114,30 +125,144 @@ contract('BlockchainParameters', (accounts: string[]) => {
...
+      // wait an epoch to find new value
+      await mineBlocks(EPOCH, web3)
+      assert.equal((await blockchainParameters.getUptimeLookbackWindow()).toNumber(), 40)
+    })
+
+    it('if unset should set the variable for the current epoch', async () => {
+      await blockchainParameters.setUptimeLookbackWindow(24)
+      assert.equal((await blockchainParameters.getUptimeLookbackWindow()).toNumber(), 24)
+    })
```
nambrot: new line in between tests

packages/protocol/test/governance/network/blockchainparams.ts
```diff
@@ -114,30 +125,144 @@ contract('BlockchainParameters', (accounts: string[]) => {
...
+
+      // still old value
+      assert.equal((await blockchainParameters.getUptimeLookbackWindow()).toNumber(), 24)
+
+      // wait an epoch to find new value
+      await mineBlocks(EPOCH, web3)
+      assert.equal((await blockchainParameters.getUptimeLookbackWindow()).toNumber(), 40)
+    })
+
+    it('if unset should set the variable for the current epoch', async () => {
```
nambrot: I think this would benefit in readability if these pre-conditions are set with sub-describe's and beforeEach's

packages/protocol/test/governance/network/blockchainparams.ts
```diff
@@ -114,30 +125,144 @@ contract('BlockchainParameters', (accounts: string[]) => {
...
+      await blockchainParameters.setUptimeLookbackWindow(24)
+      assert.equal((await blockchainParameters.getUptimeLookbackWindow()).toNumber(), 24)
+    })
+    it('if unset, second set should set the variable for the next epoch', async () => {
```
nambrot: Is this not the same test as on line 146?

packages/protocol/test/governance/network/blockchainparams.ts
```diff
@@ -114,30 +125,144 @@ contract('BlockchainParameters', (accounts: string[]) => {
...
+        blockchainParameters.setUptimeLookbackWindow(lookbackWindow, {
+          from: accounts[1],
+        })
+      )
+    })
+
+    it('should fail when using value lower than safe minimum', async () => {
+      await assertRevert(
+        blockchainParameters.setUptimeLookbackWindow(11, {
+          from: accounts[1],
```
nambrot: This will revert just because it's not the owner

packages/protocol/test/governance/network/blockchainparams.ts
```diff
@@ -114,30 +125,144 @@ contract('BlockchainParameters', (accounts: string[]) => {
...
+        })
+      )
+    })
+
+    it('should fail when using value greater than safe maximum', async () => {
+      await assertRevert(
+        blockchainParameters.setUptimeLookbackWindow(721, {
+          from: accounts[1],
```
nambrot: ditto

packages/protocol/test/governance/network/blockchainparams.ts
```diff
@@ -114,30 +125,144 @@ contract('BlockchainParameters', (accounts: string[]) => {
...
+        })
+      )
+    })
+
+    it('should fail when using value greater than epochSize-2', async () => {
```
nambrot:
```suggestion
    it('should fail when using value greater than epochSize - 2', async () => {
```

packages/protocol/test/governance/network/blockchainparams.ts
mcortesi: dumb :roll_eyes: 
Thanks

packages/protocol/test/governance/network/blockchainparams.ts
mcortesi: No, on this case we do 2 sets within the same epoch, and one is the initial one

packages/protocol/test/governance/network/blockchainparams.ts
```diff
@@ -141,94 +141,95 @@ contract('BlockchainParameters', (accounts: string[]) => {
...
+        assert.equal((await blockchainParameters.getUptimeLookbackWindow()).toNumber(), 50)
+      })
+    })
 
-      // current epoch still initial value
-      assert.equal((await blockchainParameters.getUptimeLookbackWindow()).toNumber(), 24)
+    describe('when not initialized', () => {
+      it('should set the variable for the current epoch', async () => {
+        await blockchainParameters.setUptimeLookbackWindow(initialValue)
+        assert.equal(
```
nambrot: You'd probably technically want to make this assertion before setting in case initialValue === 12

.circleci/config.yml
nambrot: I guess for this to merge into master, this needs to be master no?

nambrot (approved):

packages/protocol/test/governance/network/blockchainparams.ts
mcortesi: There's already a tests for that:
```ts
    it('should return 12 if not initalized', async () => {
      assert.equal((await blockchainParameters.getUptimeLookbackWindow()).toNumber(), 12)
    })
```

prestwich (approved):

packages/protocol/test/governance/network/blockchainparams.ts
nambrot: What I'm saying is that if `initialValue` is 12, then a faulty setUptimeLookbackWindow (i.e. set in the next epoch) would still make this test pass,right?

oneeman: Do you mean before setting to have the same assert using `.notEqual` to ensure the before and after values can't coincidentally be equal?

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -75,6 +101,49 @@ contract BlockchainParameters is Ownable, Initializable {
...
+    );
+
+    uptimeLookbackWindow.oldValue = getUptimeLookbackWindow();
+    uptimeLookbackWindow.nextValue = window;
+
+    // epoch 0 is the genesis block, thus implies it has not been initialized
+    if (uptimeLookbackWindow.nextValueActivationEpoch == 0) {
+      // on initialize we set the value for the current epoch which configurer
+      // must make it so it's SAME as geth's chainConfig value
+      uptimeLookbackWindow.nextValueActivationEpoch = getEpochNumber();
```
oneeman: Why is this necessary?  Suppose we always used `getEpochNumber().add(1)`, would that be problematic?

mcortesi: For the case where you deploy the contract mid-epoch. 

The change we just did on celo-blockchain helps with that... but only if the contract is already there when donut is activated.

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -109,35 +109,31 @@ contract BlockchainParameters is Ownable, Initializable, UsingPrecompiles {
...
     require(
       window <= getEpochSize().sub(2),
-      "UptimeLookbackWindow must be smaller or equal to epochSize-2"
+      "UptimeLookbackWindow must be smaller or equal to epochSize - 2"
     );
 
-    uptimeLookbackWindow.oldValue = getUptimeLookbackWindow();
+    uptimeLookbackWindow.oldValue = _getUptimeLookbackWindow();
+
+    // changes only take place on the next epoch
```
oneeman, line 117: This should also be made clear in the function's @notice above.

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -109,35 +109,31 @@ contract BlockchainParameters is Ownable, Initializable, UsingPrecompiles {
...
   function getUptimeLookbackWindow() public view returns (uint256 lookbackWindow) {
-    // epoch 0 is the genesis block, thus implies it has not been initialized
-    if (uptimeLookbackWindow.nextValueActivationEpoch == 0) {
-      // since on mainet, baklava, alfajores we haven't initialized the contract with lookbackWindow
-      // we need to hardcode the default value here
-      // can be removed after the deploy
-      return 12;
-    } else if (getEpochNumber() >= uptimeLookbackWindow.nextValueActivationEpoch) {
+    lookbackWindow = _getUptimeLookbackWindow();
+    require(lookbackwindow != 0, "UptimeLookbackWindow is not initialized");
```
oneeman: I'd avoid the word 'initialized' here since it can be initialized but not yet in effect.  If we want to be super-helpful we could give a different error in the case where it hasn't been initialized and the case where it has been initialized but we haven't yet passed to the next epoch.  Otherwise, I would phrase the error as "UptimeLookbackWindow not set for current epoch" or something like that.

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -109,35 +109,31 @@ contract BlockchainParameters is Ownable, Initializable, UsingPrecompiles {
...
-      uptimeLookbackWindow.nextValueActivationEpoch = getEpochNumber();
-    } else {
-      // changes only take place on the next epoch
-      uptimeLookbackWindow.nextValueActivationEpoch = getEpochNumber().add(1);
-    }
     emit UptimeLookbackWindowSet(window, uptimeLookbackWindow.nextValueActivationEpoch);
   }
 
   /**
    * @notice Gets the uptime lookback window.
```
oneeman, line 125: Call out that this reverts if there is no lookback window in effect for the current epoch (I don't know if there is a special way to do that or if it's just mentioning that in the @notice)

packages/protocol/contracts/governance/BlockchainParameters.sol
```diff
@@ -109,35 +109,31 @@ contract BlockchainParameters is Ownable, Initializable, UsingPrecompiles {
...
-      // we need to hardcode the default value here
-      // can be removed after the deploy
-      return 12;
-    } else if (getEpochNumber() >= uptimeLookbackWindow.nextValueActivationEpoch) {
+    lookbackWindow = _getUptimeLookbackWindow();
+    require(lookbackwindow != 0, "UptimeLookbackWindow is not initialized");
+  }
+
+  /**
+   * @notice Gets the uptime lookback window.
```
oneeman, line 133: Call out that this returns 0 if there is no lookback window value in effect for the current epoch

PR #4756, opened 2020-08-21
Salt was renamed to pepper in the PhoneNumberHashDetails

bowd:
### Description

Looks like there's a leftover `salt` -> `pepper` here that wasn't caught because our generators aren't strictly typed. This feels dangerous and should be treated more broadly, but I know generators are notoriously hard to type because of the yield typing semantics.

For now I introduced a typed const to ensure correct typing -- feels like this should be a pattern instead of directly returning objects from generators as that will never complain about structure.

### Tested

Curious why tests didn't flag this - we don't have cached pepper related tests?

### DIFF
--- a/packages/mobile/src/identity/privateHashing.ts
+++ b/packages/mobile/src/identity/privateHashing.ts
@@ -66,7 +66,8 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
   if (cachedSalt) {
     Logger.debug(`${TAG}@fetchPrivatePhoneHash`, 'Salt was cached')
     const phoneHash = getPhoneHash(e164Number, cachedSalt)
-    return { e164Number, phoneHash, salt: cachedSalt }
+    const cachedDetails: PhoneNumberHashDetails = { e164Number, phoneHash, pepper: cachedSalt }
+    return cachedDetails
   }
 
   Logger.debug(`${TAG}@fetchPrivatePhoneHash`, 'Salt was not cached, fetching')


packages/mobile/src/identity/privateHashing.ts
```diff
@@ -66,7 +66,8 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
   if (cachedSalt) {
     Logger.debug(`${TAG}@fetchPrivatePhoneHash`, 'Salt was cached')
     const phoneHash = getPhoneHash(e164Number, cachedSalt)
-    return { e164Number, phoneHash, salt: cachedSalt }
+    const cachedDetails: PhoneNumberHashDetails = { e164Number, phoneHash, pepper: cachedSalt }
```
jeanregisser, line 69: Is `cachedSalt` still the appropriate name?
I don't know the specifics of this code and when I look at it I understand `salt` becomes `pepper` which might be right but it's hard to tell without more context on this.

jeanregisser (commented):
Thank you @bowd!! Great catch!! 🥇 

This is a good reminder that we don't yet have e2e tests for verification (planned for upcoming milestone 6).

Also one more reason to agree on a solution for `redux-saga` typing issue.
@jmrossy spent some time on this recently and suggested using [typed-redux-saga](https://celo-org.slack.com/archives/CL7BVQPHB/p1596664825144900?thread_ts=1596664825.144900).

I found out about this lib after bumping in 3 issues caused by this and was initially against it because the syntax is slightly different (we'd have to use `yield*` instead of `yield`) but I'm now thinking it's probably the best way to fix this for now. 
I'm wondering if it would enforce correct usage of `yield*` though, if it doesn't, then it might still hide real issues.

jeanregisser (approved):

packages/mobile/src/identity/privateHashing.ts
bowd: It's not the right name, no, but at the same time that's a rabbit hole, 'cose currently nothing around the _salt_ cache was renamed to pepper. I felt like doing this as part of this bug-fix wasn't warranted.

jeanregisser: Yes of course I didn't mean for you to do the salt/pepper change rename in this PR!
Thanks for the details.

jeanregisser: I've opened #4758 to track this.

PR #4760, opened 2020-08-21
[Wallet] Add back trigger for enabling dev mode in setting menu

jmrossy:
### Description

Add back trigger for enabling dev mode in setting menu. Clicks show no user feedback.

### Tested

In iOS simulator

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/account/reducer.ts
+++ b/packages/mobile/src/account/reducer.ts
@@ -3,6 +3,7 @@ import { Actions, ActionTypes } from 'src/account/actions'
 import { DEV_SETTINGS_ACTIVE_INITIALLY } from 'src/config'
 import { features } from 'src/flags'
 import { getRehydratePayload, REHYDRATE, RehydrateAction } from 'src/redux/persist-helper'
+import Logger from 'src/utils/Logger'
 import { getRemoteTime } from 'src/utils/time'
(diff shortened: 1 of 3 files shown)


jmrossy:
TODO: add a toast message

_Edit_: Done.

packages/mobile/src/account/reducer.ts
```diff
@@ -111,10 +112,14 @@ export const reducer = (
       }
     case Actions.DEV_MODE_TRIGGER_CLICKED:
       const newClickCount = (state.devModeClickCount + 1) % 6
+      const devModeActive = newClickCount >= 3
+      if (devModeActive) {
+        Logger.showMessage('Debug Mode Activated')
+      }
```
jeanregisser, line 118: nit: though this is harmless here, reducers shouldn't have any side effect.
This should be at the component level or saga.

jeanregisser (commented):
Thanks! Just one minor thing.

i1skn (approved):
LGTM after @jeanregisser feedback is addressed :)

packages/mobile/src/account/reducer.ts
jmrossy: I considered that but that would require us to create a watcher for this event which duplicates the logic in this reducer. Just not worth it for 'purity' in this case I think

PR #4765, opened 2020-08-21
Enable usdGas + useLedger and refactor usdGas to gasCurrency

yorhodes:
### Description

Add base `gasCurrency` flag and config for use with `cUSD` and `CELO` fee currencies
Also supports an `auto` mode which defaults to whatever is available on the sender's balance
Config `gasCurrency` defaults to `auto` before being set by the user

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/packages/cli/src/commands/config/set.ts
+++ b/packages/cli/src/commands/config/set.ts
@@ -19,6 +19,7 @@ export default class Set extends LocalCommand {
     'set  --node ws://localhost:2500',
     'set  --node <geth-location>/geth.ipc',
     'set --gasCurrency cUSD',
+    'set --gasCurrency CELO',
   ]
 
   async run() {
--- a/packages/cli/src/commands/config/set.ts
+++ b/packages/cli/src/commands/config/set.ts
@@ -10,14 +10,16 @@ export default class Set extends LocalCommand {
     // Overrides base command node flag.
     node: flags.string({
       char: 'n',
-      required: false,
       description: 'URL of the node to run commands against',
-      default: 'http://localhost:8545',
     }),
     gasCurrency: BaseCommand.flags.gasCurrency,
   }
 
-  static examples = ['set  --node ws://localhost:2500', 'set  --node <geth-location>/geth.ipc']
+  static examples = [
+    'set  --node ws://localhost:2500',
+    'set  --node <geth-location>/geth.ipc',
+    'set --gasCurrency cUSD',
+  ]
 
   async run() {
     const res = this.parse(Set)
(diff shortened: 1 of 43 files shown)


packages/cli/src/commands/config/set.ts
```diff
@@ -10,19 +10,25 @@ export default class Set extends LocalCommand {
...
   static examples = ['set  --node ws://localhost:2500', 'set  --node <geth-location>/geth.ipc']
 
   async run() {
     const res = this.parse(Set)
-    const config: CeloConfig = {
-      nodeUrl: res.flags.node,
-    }
-    await writeConfig(this.config.configDir, config)
+    const curr = readConfig(this.config.configDir)
+    const node = res.flags.node ?? curr.node
```
gastonponti, line 28: Double check this.
Node has a `default: 'http://localhost:8545'` so that value will always show the default.
I can't really remember but I think that something like this should help:
```
        const hasNode = res.raw.some(
          (value) => (value as any).flag === 'node'
        )
        const node = hasNode ? res.flags.node : curr.node
```

packages/cli/src/commands/transfer/dollars.ts
```diff
@@ -30,17 +31,28 @@ export default class TransferDollars extends BaseCommand {
     this.kit.defaultAccount = from
```
gastonponti: You can remove this one too, right?

gastonponti (commented):
Could you check the node default flag thing?

packages/cli/src/commands/config/set.ts
yorhodes: I'm checking if an override is provided but defaulting to the config (localhost)
I have tested this with setting and getting alfajores forno node url

packages/cli/src/commands/config/set.ts
```diff
@@ -10,19 +10,28 @@ export default class Set extends LocalCommand {
     // Overrides base command node flag.
     node: flags.string({
       char: 'n',
-      required: true,
       description: 'URL of the node to run commands against',
-      default: 'http://localhost:8545',
```
gastonponti, line 15: Mmmm, if you decide to remove the defaults from the commands (for node and gasCurrency) but you are adding those from the config utils, you should at least add in the descriptions that those are the defaults.

It's kind of weird having the default here and in the configFile and it seems wrong, I know, but I think that this default helps a lot for the documentation.

gastonponti (approved):

PR #4766, opened 2020-08-22
Lower min balance && support CELO min balance

codyborn:
### Description

- Lowering min balance to reduce out of quota errors (min balance that's greater than tx fees is not a real cost)
- Support min balance of CELO

### Tested

New unit tests

### Backwards compatibility

Yes

### DIFF
--- a/packages/phone-number-privacy/signer/src/signing/query-quota.ts
+++ b/packages/phone-number-privacy/signer/src/signing/query-quota.ts
@@ -90,8 +90,21 @@ async function getQueryQuota(account: string, hashedPhoneNumber?: string) {
     )
   }
 
-  const cUSDAccountBalance = await getDollarBalance(account)
-  const celoAccountBalance = await getCeloBalance(account)
+  let cUSDAccountBalance = new BigNumber(0)
+  let celoAccountBalance = new BigNumber(0)
+
+  await Promise.all([
+    new Promise((resolve) => {
+      resolve(getDollarBalance(account))
+    }),
+    new Promise((resolve) => {
+      resolve(getCeloBalance(account))
+    }),
+  ]).then((values) => {
+    cUSDAccountBalance = values[0] as BigNumber
+    celoAccountBalance = values[1] as BigNumber
+  })
+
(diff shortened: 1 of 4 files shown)


aslawson (approved):
lgtm, thanks!

packages/phone-number-privacy/signer/src/signing/query-quota.ts
```diff
@@ -90,8 +90,13 @@ async function getQueryQuota(account: string, hashedPhoneNumber?: string) {
...
 
-  const accountBalance = await getDollarBalance(account)
-  if (accountBalance.isGreaterThanOrEqualTo(config.quota.minDollarBalance)) {
+  const cUSDAccountBalance = await getDollarBalance(account)
+  const celoAccountBalance = await getCeloBalance(account)
+  // Min balance can be in either cUSD or CELO
+  if (
+    cUSDAccountBalance.isGreaterThanOrEqualTo(config.quota.minDollarBalance) ||
+    celoAccountBalance.isGreaterThanOrEqualTo(config.quota.minCeloBalance)
+  ) {
```
jmrossy: Nit: i would be more efficient to check them individually, that way we short circuit the 2nd state read in the case of wallets with min cUSD balance.

jmrossy (approved):
One optional comment

packages/phone-number-privacy/signer/src/signing/query-quota.ts
codyborn: Good point. What do you think about running both checks in parallel? This would optimize for the 0 cUSD balance case as well.

jmrossy: Yeah `Promise.all` would work similarly well but we still double the forno load per request. I'm sure it's fine either way :)

alecps (approved):
Looks good to me!

packages/phone-number-privacy/signer/src/signing/query-quota.ts
codyborn: Thanks! Went with the promise.all approach.

PR #4777, opened 2020-08-24
Add ODIS db migrations to signer init

alecps:
### Description

Add ODIS db migrations to signer init

### Other changes

Disable ODIS retries on 400s in contractkit 

### Tested

On Alfajores

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes


packages/base/src/async.test.ts
```diff
@@ -27,7 +27,7 @@ describe('selectiveRetryAsyncWithBackOff()', () => {
   test('tries only once if error is in dontRetry array', async () => {
     const mockFunction = jest.fn()
     mockFunction.mockImplementation(() => {
-      throw new Error('test')
+      throw new Error('test ')
```
jmrossy: intentional?

packages/contractkit/src/identity/odis/query.ts
```diff
@@ -106,6 +106,7 @@ export async function queryOdis<ResponseType>(
     ErrorMessages.ODIS_QUOTA_ERROR,
     ErrorMessages.ODIS_AUTH_ERROR,
     ErrorMessages.ODIS_INPUT_ERROR,
+    'Unknown Client Error',
```
jmrossy: Nit: consider moving this to a const var at the top.

packages/contractkit/src/identity/odis/query.ts
```diff
@@ -136,6 +137,10 @@ export async function queryOdis<ResponseType>(
         case 401:
           throw new Error(ErrorMessages.ODIS_AUTH_ERROR)
         default:
+          if (Math.floor(res.status / 400) === 1) {
```
jmrossy: This would be true for anything > 400 right? Including 500s

codyborn: A common approach that I've seen is 
`if (res.stats >= 400 && res.status < 500)`

packages/base/src/async.test.ts
alecps: Yea.. I suppose I could be more explicit here

packages/contractkit/src/identity/odis/query.ts
alecps: 😂 Yep! I'll fix it

codyborn (approved):

packages/contractkit/src/identity/odis/query.ts
```diff
@@ -136,6 +138,10 @@ export async function queryOdis<ResponseType>(
         case 401:
           throw new Error(ErrorMessages.ODIS_AUTH_ERROR)
         default:
+          if (Math.floor(res.status / 100) === 4) {
```
jmrossy: Any reason we can't do res.status >= 400 && <= 500 like Cody suggested?

packages/contractkit/src/identity/odis/query.ts
```diff
@@ -136,6 +138,10 @@ export async function queryOdis<ResponseType>(
...
+            // Don't retry error codes in 400s
+            throw new Error(ErrorMessages.ODIS_CLIENT_ERROR + ' ' + res.status)
```
jmrossy:
```suggestion
            throw new Error(`${ErrorMessages.ODIS_CLIENT_ERROR} ${res.status}`)
```

packages/phone-number-privacy/signer/migrations/20200330212224_create-accounts-table.ts
```diff
@@ -1,4 +1,4 @@
-import * as Knex from 'knex'
+import Knex from 'knex'
```
jmrossy: why change these? In theory they're equivalent but just a bit wary to change since knex itself auto-gened most of the content in these files

alecps: Yea I'm actually facing an issue with this rn. I'll ping you to discuss

packages/contractkit/src/identity/odis/query.ts
alecps: Oh I didn't see that suggestion. That works for me

packages/phone-number-privacy/signer/src/migrations/20200330212224_create-accounts-table.ts
```diff
@@ -0,0 +1,21 @@
...
+    return knex.schema.createTable(ACCOUNTS_TABLE, (t) => {
+      t.string(ACCOUNTS_COLUMNS.address)
+        .notNullable()
+        .primary()
+      t.dateTime(ACCOUNTS_COLUMNS.createdAt).notNullable()
+      t.integer(ACCOUNTS_COLUMNS.numLookups).unsigned()
+      t.dateTime(ACCOUNTS_COLUMNS.didMatchmaking)
+    })
+  }
+  return new Promise((resolve) => resolve(null))
```
jmrossy: Was this Promise wrapper necessary? Al async functions return promises by default, wondering if `return null` is equivalent here.

jmrossy (approved):
LGTM! Excited to have db migrations rolled into initialization, this was def a pain point when we first onboarded our partner signers. Thanks for sticking with it on this.

packages/phone-number-privacy/signer/src/migrations/20200330212224_create-accounts-table.ts
alecps: Good call! Looks like it's not necessary

codyborn (approved):
Looks great! Thanks for the deep investigation on this

PR #4778, opened 2020-08-25
[WIP] Verify source code matches deployed bytecode

### LINKED ISSUE #4496: Script for checking a contracts directory corresponds to deployed bytecode
### Expected Behavior

It should be possible to verify that a given set of Solidity sources corresponds to the currently deployed bytecodes on a Celo network. This is necessary for automated backwards compatibility checking to make sense. 

### Current Behavior

No such script.

m-chrzan:
Note to reviewers: putting [WIP] on this again while I work on also verifying a governance proposal - that work will live on top of these changes and I'll end up rewriting/refactoring some of the code.

### Description

This PR adds a script that verifies that a given set of Solidity sources matches bytecodes deployed to a Celo network.

This check will rely on:
* Each (non-library) core contract being registered in the Registry
* Each linked library being proxied

The deployed Celo contracts do not currently satisfy these assumptions, so there is some temporary code included to handle the current case. The checks are also not 100% complete for the current RC1 deployment (e.g. the Registry itself is missed since it's not registered).

The main check happens in the `verifyBytecodesDfs` function, which takes a list of core contract names and a build directory as input. The DFS works as follows:

1. If we're on a core contract, look up it's address in the Registry. We keep track of library addresses as we encounter them.
2. Find all instances of `__LibraryName__...` in the compiled bytecode.
3. Ensure that each `__LibraryName__...` link corresponds to the same address in the onchain bytecode.
4. Link the compiled bytecode and verify it matches the onchain bytecode.
5. Add any freshly encountered libraries to the DFS stack.

The biggest remaining piece of work on this front is also verifying *proposed* changes, that is, given a Celo network and a Governance proposal, verifying that the given sources correspond to what the Celo system would look like after applying the proposal.

### Tested

Unit tests for the library doing the heavy lifting.

For a manual test against baklava,

    git checkout baklava
    yarn build
    mv build/contracts build/baklava-contracts
    git checkout m-chrzan/deployed-bytecode-check
    # point the script to build/baklava-contracts
    yarn build:ts && yarn truffle exec --network baklava scripts/truffle/verify-bytecode.js

### Other changes

Moved some duplicated code out into a new `lib/bytecode` library.

### Related issues

- Fixes #4496 

### Backwards compatibility

Very much so.


m-chrzan:
Closing this as #4996 is a superset of this diff.

Issue #4655, opened 2020-08-10
Continuous loading indicator is shown when user paste received invite code on “Invite Code” screen


Lss-Ankit:
**Frequency:** 100%
**Repro App version:**  Celo Alfajores Wallet (Internal early Access) v1.9.4 (105094704) 
Does not repro on: Play store build v1.8.2.
**Repro on device:** Google Pixel XL (8.1) 
**Pre-Condition:** User should receive Invite code from friends 

**Repro Steps:**
1. Launch the Application and complete the walkthrough. 
2. Enter Full and select country and enter phone number. 
3. Tap on Next button. 
4. Create a Security PIN
5. Paste Received Invite Code. 
6. Observe 

**Current Behavior:** Continuous loading indicator is shown 
**Expected Behavior:** User should be able to sign up with Invite code. 
**Impact:** User is blocked to sign up with invite code. 
**Attachment:** [Android_InviteCodeissue.MP4 ](https://drive.google.com/file/d/1lp2Ly0hRyg05balUWjbaRlLdl7KzIEEz/view?usp=sharing)(Bug observed at 30 sec)

Lss-Ankit:
I have verified this issue on Valora app build v1.0.0 (10) and in iOS test flight build v1.0.0 (11) and found issue is fixed

annakaz:
@Lss-Ankit can you confirm whether you're seeing this in the latest build?

annakaz:
Seeing invite codes failing on the most recent build as well. See https://docs.google.com/document/d/1NmEdo4lyLox_2t0efuzuiXSBhVeLLINJ-d5MnKe7ej4/edit?disco=AAAAG1oO_TI for details

dominiquejane:
Additional details from Aug 19 onboarding session:
Users were sent an invite code, however, neither copy/pasting the entire link nor using just the passcode in the link worked. Roughly 70% of the users were unable to use the invite code.

I saw some messages on slack about the operation timing out after 2 min, which may have been the case for many of the users.

Lss-Ankit:
Hi @annakaz I have verified Invite flow using Android Valora app build v1.0.0 (12) & IOS test flight build v1.0.0 (11) and found the following: 
- "Invite code has no funds" error message is shown when user proceed for sign up with using received invitation code. 
Note: I have sent invitation from user A who has balance and after sent invitation $0.25 also cut for User A

[...]

PR #4780, opened 2020-08-25
[Wallet] Reduce invite redemption time

tarikbellamine:
### Description
Successfully redeeming an invite code currently takes ~44 seconds on average, which results in a large amount of users dropping off. This PR does four things to try to speed the process up:
- Runs invite redemption generators in parallel to the extent possible
- Hard codes the gas estimate for moving funds to $0.005, an order of magnitude higher than the $0.0003 it costs when recorded
- Hard codes the gas estimate for the send transaction to 200K, which is rounded up from the actual cost of 191775
- Moves the DEK registration to after the invite redemption is successful. In practice DEK registration will now run simultaneously to verification

### Other changes
Added in more event logging so we can better diagnose bottlenecks. Removed a migration hack from the DEK registration process that we no longer need.

Fixes an unrelated [Sentry error](https://sentry.io/organizations/celo/issues/1860869526/?environment=co.clabs.valora&project=1250733)

### Tested
Tested on device

### Related issues
- Partially addresses #4655 

### Backwards compatibility
Yes

IMG_0137

IMG_0138


packages/mobile/src/invite/saga.ts
```diff
@@ -284,10 +284,13 @@ export function* redeemInviteSaga({ inviteCode }: RedeemInviteAction) {
...
   })
 
-  if (result === true) {
+  if (result.success === true) {
     Logger.debug(TAG, 'Redeem Invite completed successfully')
     yield put(redeemInviteSuccess())
     navigate(Screens.VerificationEducationScreen)
+    // Note: We are ok with this succeeding or failing silently in the background,
+    // user will have another chance to register DEK when sending their first tx
+    yield call(registerAccountDek, result.newAccount)
```
i1skn: Will user be able to verify later if this fails?

packages/mobile/src/invite/saga.ts
```diff
@@ -284,10 +284,13 @@ export function* redeemInviteSaga({ inviteCode }: RedeemInviteAction) {
     timeout: delay(REDEEM_INVITE_TIMEOUT),
   })
 
-  if (result === true) {
+  if (result.success === true) {
```
i1skn, line 298: I'm wondering about this. As far as I got `doRedeemInvite` returns `false | { success: boolean; newAccount: string; }`. So in case it's `false` we should see an error, no?

packages/mobile/src/invite/saga.ts
jmrossy: Consider `yield spawn` here so this saga can close out.

packages/mobile/src/invite/saga.ts
```diff
@@ -397,13 +402,17 @@ export function* moveAllFundsFromAccount(
...
-    recipientAddress: toAccount,
-    amount: tempAccountBalance,
-    comment,
-  })
-  // Inflate fee by 10% to harden against minor gas changes
-  const sendTokenFee = divideByWei(sendTokenFeeInWei).times(1.1)
+
+  // Temporarily hardcoding fee estimate to be 1/2 a cent to save time on estimation. Actual
+  // fees are currently an order of magnitude smaller (0.0003)
+  const sendTokenFee = new BigNumber(0.005)
```
jmrossy: It's a bit ironic cause this is what we used to do. Then @jeanregisser and I overhauled fee estimation across the whole app, replacing the hard-coding with a properly calculated fee. And now where we are again :)

jmrossy: Suggest hoisting this contact to the top of the file for clarity

packages/mobile/src/invite/saga.ts
tarikbellamine: That's a good call. I should do `result && result.success`. I'll type check the `yield race` results to avoid this mistake

packages/mobile/src/invite/saga.ts
tarikbellamine: Yeah if they arent registered here, then they register their DEK either during verification or with their first transfer transaction

packages/mobile/src/invite/saga.ts
```diff
@@ -275,19 +277,29 @@ function* sendInviteSaga(action: SendInviteAction) {
   }
 }
 
+function* registerAccountDek(newAccount: string) {
```
jmrossy: Why create this wrapper?

nategraf:
Reading this I am now realizing that we actually estimate gas twice for every transaction, which is less than ideal. Instead of hardcoding the gas _fee_ here, we need to be hard-coding the gas _amount_ and using that to derive the fee if we hope to avoid the cost of both gas estimations. I can provide more context if needed. As this PR is written it will eliminate gas estimation for the fee that is shown the the user, but will still estimate before sending the transaction unless a gas value is added to the transaction before sending.

packages/mobile/src/invite/saga.ts
tarikbellamine: My bad, made an incorrect assumption around the input type of `yield spawn`. Will revert this change and just call `yield spawn` directly with the unwrapped `registerAccountDek`

packages/mobile/src/invite/saga.ts
```diff
@@ -327,10 +339,11 @@ export function* doRedeemInvite(inviteCode: string) {
       CURRENCY_ENUM.DOLLAR,
       SENTINEL_INVITE_COMMENT
     )
-    yield call(registerAccountDek, newAccount)
+    ValoraAnalytics.track(OnboardingEvents.invite_redeem_move_funds_complete)
+
     yield put(fetchDollarBalance())
     ValoraAnalytics.track(OnboardingEvents.invite_redeem_complete)
-    return true
+    return { success: true, newAccount }
```
jeanregisser, line 349: nit: What do you think of unifying the return type? for instance `{ success: false } | { success: true, newAccount: string }` or simply `string | null`?

packages/mobile/src/invite/EnterInviteCode.tsx
```diff
@@ -67,6 +67,7 @@ export class EnterInviteCode extends React.Component<Props, State> {
         // Note: redux state reset is handled by UseBackToWelcomeScreen
         // tslint:disable-next-line: jsx-no-lambda
         onPress={() => navigate(Screens.Welcome)}
+        eventName={OnboardingEvents.invite_redeem_cancel}
```
jeanregisser: You can also cancel by hitting the hardware back button on Android or doing a swipe back on iOS.

This event should be moved to `<UseBackToWelcomeScreen ... />` which handles all of these cases, though when you've already started the redeem process we should send both `OnboardingEvents.invite_redeem_cancel` and `OnboardingEvents.create_account_cancel`. 
`UseBackToWelcomeScreen` can currently only send one event so it would need to be modified to be able to send more than one.

jeanregisser (commented):
Awesome! 👍  Looking almost good!!

packages/mobile/src/invite/EnterInviteCode.tsx
tarikbellamine: Didn't realize we had that. Thanks, will use it instead!

packages/mobile/src/utils/time.ts
```diff
@@ -319,6 +319,6 @@ function quickFormat(timestamp: number, i18next: i18nType, formatRule: string) {
     Math.abs(differenceInYears(timestamp, Date.now())) > 40 ? timestamp * 1000 : timestamp
 
   return format(millisecondsSinceEpoch, formatRule, {
-    locale: i18next.language.includes('es') ? es : enUS,
+    locale: i18next?.language.includes('es') ? es : enUS,
```
tarikbellamine, line 322: Small fix for unrelated Sentry error I was seeing

packages/mobile/locales/en-US/onboarding.json
```diff
@@ -38,7 +38,9 @@
     "label": "Invite Code",
     "codePlaceholder": "Enter invite code",
     "noCode": "Don’t have an Invite Code",
-    "nodeCodeInviteLink": "Don’t have an invite code? Request one at <0>celo.org/developers/wallet</0> or <1>continue without</1>"
+    "nodeCodeInviteLink": "Don’t have an invite code? Request one at <0>celo.org/developers/wallet</0> or <1>continue without</1>",
```
aaronmgdr, line 41: are we still wanting to have this in the public build? its a dev only feature that doesnt even work right now

jeanregisser: Thanks for checking.
This is controlled by the `SHOW_GET_INVITE_LINK` config, which is appropriately set to `false` for public builds.

jeanregisser (approved):
🚀

Lss-Ankit:
@tarikbellamine @jeanregisser  i have verified task using TestFlight 1.1.0 build (21) and Play Store internal (1004294309) and observed that invite code redemption time is ~30-35 sec for me. 
Note: I have check 2 -2 times on Android and IOS platform.

Issue #4450, opened 2020-07-15
Contracts SDK module as wrapper for typechain/contract abis/tx data builder


yorhodes:
### Expected Behavior

Please describe the behavior you are expecting

### Current Behavior

What is the current behavior?

aslawson:
@gastonponti was this fixed in  [Sdk Modularization #4790](https://github.com/celo-org/celo-monorepo/pull/4790)?
The PR says it fixes this.

aslawson:
@aslawson to follow up with @yorhodes

PR #4790, opened 2020-08-25
Sdk Modularization

### LINKED ISSUE #3993: Types in ContractKit

gastonponti:
### Description

Divide the ContractKit in multiple packages, to reduce the bundle size required as default.

Package descriptions (and probably the way to face this PR)

- `Connect` (Connection)
This package basically handles how we communicate to the our chain nodes. It wraps the `web3` library and has its own `rpcCaller` class, to make custom calls to the node. It's the layer in charge of knowing how and which parameters are added by Celo, connect to the node, build the message, sent it and handle those responses.
One big change that this package/layer brings, is that we are not relying anymore on the typing extension of web3 (which generated a ton of typing problems for developers trying to use our library). This layer is the one in charge of exporting the "celo types" used to make transactions/receipts, translates those types to use the web3 lib, and then translate the response from web3 to re-use our types again, which will maintain all our types all along the implementations.
Also, it won't have the web3 library as a dependency (just as a dev dependency for typing) which will reduce the size, but will allow us not only to update the library easier, but also to generate a light web3 library or even will make a future migration of the core library easier.
Edit: We decide to add web3 as a dependency in the `ContractKit`, because we indirectly ask the user to add it, so it will be easier for migrations and for the first steps of a dev coding for Celo. In `Connect` will remain as a `peerDependecy` 

- `ContractKit`
This contractKit package, is a reduce sub set of the old contractKit. This is the layer in charge of loading and using our core contracts. Internally uses the `connect` package.
It has our contracts generated from the abis, its wrappers and also the logic to make claims (used from the attestation wrapper for example)

- Complementary packages
Basically are the packages that consume our core contracts, but won't be part of the common use case of the `contractKit` library.
The new packages are:
  - `explorer` (Block and Log explorers)
  - `governance` (proposal generator)
  - `identity` (odis, phone-number-identifier)
  - `network-utils` (utils for genesis-block, static-nodes, google-storage)
  - `transactions-uri` (tx-uri generator)

- A lot of Wallets
One of the biggest problem of our bundle size was that every wallet that we use, requires heavy libraries and we had them all in the same package. With this division, every user will just require the wallet that is going to use, and will avoid all the other "wallet libraries".
This is the list of wallets that the user could use:
  - `wallet-hsm-azure`
  - `wallet-hsm-aws`
  - `wallet-ledger`
  - `wallet-local`
  - `wallet-rpc`
Then we have a three more wallet packages, that are mostly abstract clases and utils that the previous wallets use. This libraries are: `wallet-base`, `wallet-hsm` and `wallet-remote`

### Other changes

Fix Types
All the tools were moved to a `sdk` folder
Increase web3 version (to 1.3.0)
Prune some unnecessary dependencies

### Notes ###

Non of the docs autogenerated files were commited, just to avoid having the double of files changed. This should and be added after the PR have been approved (this will show the `contractKit` tests failing, because of the docs check, but is something working locally)

### Related issues

- Fixes #3993, #4450

### Backwards compatibility

No (we can add a few things to make it backwards compatible, but the question is if we want to modify that, knowing that will make the user to get used to a non properly way to use the sdk in general).
Edit: We added those changes to make it easier for the user to use the new sdk, but as big part of the imports changed, and as the CK is not creating a web3 object anymore (just receive it), there are some changes that won't be totally backward compatible.


dockerfiles/attestation-service/Dockerfile
```diff
@@ -10,7 +10,9 @@ COPY patches/ patches/
 COPY packages/base/package.json packages/base/
 COPY packages/utils/package.json packages/utils/
 COPY packages/typescript/package.json packages/typescript/
+COPY packages/typechain-target-web3-v1-celo/package.json packages/typechain-target-web-v1-celo/
```
jmrossy: Does this need typechain? @nambrot trimmed this down a while back but then reverted I believe. Maybe he can shed some light.

dockerfiles/leaderboard/Dockerfile
```diff
@@ -10,7 +10,9 @@ COPY patches/ patches/
 COPY packages/typescript/package.json packages/typescript/
 COPY packages/base/package.json packages/base/
 COPY packages/utils/package.json packages/utils/
+COPY packages/typechain-target-web3-v1-celo/package.json packages/typechain-target-web-v1-celo/
```
jmrossy: ditto for many of these

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -192,7 +192,7 @@ class AttestationRequestHandler {
       salt
     )
 
-    return kit.web3.eth.sign(message, getAttestationSignerAddress())
+    return kit.communication.web3.eth.sign(message, getAttestationSignerAddress())
```
jmrossy: I wonder if a `getWeb3()` method on kit would be more elegant since in the short-to-medium term this pattern will be common

packages/base/src/collections.ts
```diff
@@ -20,6 +20,10 @@ export function zip3<A, B, C>(as: A[], bs: B[], cs: C[]) {
   return res
 }
 
+export function zeroRange(to: number): number[] {
```
jmrossy, line 23: Nit, just call this `range`?

packages/blockchain-api/package.json
```diff
@@ -17,8 +17,8 @@
     "deploy": "./deploy.sh"
   },
   "dependencies": {
-    "@celo/contractkit": "0.3.6",
-    "@celo/utils": "0.1.9",
+    "@celo/contractkit": "0.4.14-dev",
```
jmrossy: Blockchain api gets deployed to app engine which does npm installs and builds in the cloud, so it can only use public packages, not the -dev version.

Same constraint for notification-service, the ODIS combiner, and the faucet

jmrossy: It would still be good to upgrade this, though maybe not in this already big PR :)

packages/celotool/src/cmds/account/faucet.ts
```diff
@@ -69,10 +70,11 @@ export const handler = async (argv: FaucetArgv) => {
   const addresses = argv.account
 
   const cb = async () => {
-    const kit = newKit('http://localhost:8545')
-    const account = (await kit.web3.eth.getAccounts())[0]
+    const web3: Web3 = new Web3('http://localhost:8545')
```
jmrossy: why the explicit types here and next line?

packages/celotool/src/lib/geth.ts
```diff
@@ -189,7 +186,7 @@ export const checkGethStarted = (dataDir: string) => {
 }
 
 export const getWeb3AndTokensContracts = async () => {
-  const kit = newKit('http://localhost:8545')
+  const kit = newKitFromWeb3(new Web3('http://localhost:8545'))
```
jmrossy, line 189: I wonder if the newKit interface should be:
`const kit = new ContractKit(Web3, 'myUrl')`
This feels more natural and now would be a good time to revisit this.

packages/cli/src/base.ts
```diff
@@ -124,22 +121,15 @@ export abstract class BaseCommand extends LocalCommand {
...
-    return nodeUrl && nodeUrl.endsWith('.ipc')
-      ? new Web3(new Web3.providers.IpcProvider(nodeUrl, net))
-      : new Web3(nodeUrl)
-  }
-
   get kit() {
     if (!this._kit) {
-      this._kit = newKitFromWeb3(this.web3, this._wallet)
+      this._kit = newKitFromWeb3(this.web3)
+      this._kit.communication.wallet = this._wallet
```
jmrossy: This feel a bit unnatural to me. How does this differ from / overlap with `addAccount`?

packages/communication/.gitignore
```diff
@@ -0,0 +1,4 @@
+lib/
```
jmrossy: Could this go under `packages/kits/*` or maybe `packages/sdk/*` so we can start structuring that subfolder as a workspace? Similar to what was done for phone-privacy

jmrossy (commented):
I've only reviewed the first 1/4 or so but some early thoughts:

1. I'd recommend we rethink the addition of ContractKit-Extenders. It's unclear what they are, it feels like a bag of random stuff. Just renaming it wouldn't solve the issue I think. We need to break them apart and find the right homes for them. Some should be their own package. Others should be moved to @celo/utils. Happy to chat about them case-by-case.

2. Instead of `communication`, how about `connection` or just `link`?

3. This may be a good time to revisit the `newKit` interface. Something like `new ContractKit()` feels more intuitive.

dockerfiles/attestation-service/Dockerfile
gastonponti: Actually, we have a fork that I created a time ago (https://github.com/celo-org/typechain-target-web3-v1-celo) because we had typings problems with the the web3 types extensions. And as we have it as an actual package, docker was just installing it, that's why it was not needed to copying it anymore. 
Now that typechain uses our communication library for the typing, I basically added to the repo for two things:
a) it was easier to generate changes and test the flows without re publishing our typechain a lot of times
b) as it is a custom made for us that is only used by the protocol package, I thought that having it in the monorepo could be easier to add changes.

But, as I made all the class generation part is stable, I could add the changes in our fork, publish it again, and use it as a dependency
(I'll change that)

packages/communication/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@celo/communication",
+  "version": "0.0.1-dev",
+  "description": "Celo's wrapper to interact with Celo network",
```
jmrossy:
```suggestion
  "description": "Toolkit for communicating with the Celo network",
```

packages/communication/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@celo/communication",
```
jmrossy: I said this above but how about `@celo/connect`?

packages/communication/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "main": "./lib/index.js",
+  "types": "./lib/index.d.ts",
+  "author": "Celo",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/celo-org/celo-monorepo/tree/master/packages/communication",
+  "repository": "https://github.com/celo-org/celo-monorepo/tree/master/packages/communication",
+  "keywords": [
+    "celo",
+    "blockchain",
+    "defi"
```
jmrossy: maybe misleading

packages/communication/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "dependencies": {
+    "@celo/utils": "0.1.20-dev",
+    "debug": "^4.1.1"
+  },
+  "devDependencies": {
+    "web3": "1.2.11",
+    "web3-core": "1.2.11",
+    "web3-eth": "1.2.11",
+    "web3-eth-contract": "1.2.11",
+    "web3-utils": "1.2.11",
```
jmrossy: Why all these web3 items when `web3` includes all of them?

packages/communication/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@celo/flake-tracker": "0.0.1-dev",
+    "@types/debug": "^4.1.5",
+    "@types/ethereumjs-util": "^5.2.0"
```
jmrossy: intentional?

packages/communication/src/abi.ts
```diff
@@ -0,0 +1,35 @@
+import { AbiInput, AbiItem } from 'web3-utils'
+import { EventLog } from './commons'
```
jmrossy: I'm going to assume that there aren't significant code changes and just focus on structure. Please lmk if there's any files you want me to check for code.

packages/communication/src/commons.ts
```diff
@@ -0,0 +1,70 @@
+import { PromiEvent, Transaction, TransactionConfig, TransactionReceipt } from 'web3-core'
```
jmrossy: Looks like these are all types. Can we name this `types.ts`?

packages/communication/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+export * from './utils/celo-transaction-object'
+export * from './utils/rpc-caller'
+export * from './utils/tx-result'
+export * from './wallet'
+
+/**
+ * Creates a new instance of `CommunicationWrapper` with a web3 instance
+ * @param web3 Web3 instance
+ */
+export function newCommunicationWrapperFromWeb3(web3: Web3, wallet?: ReadOnlyWallet) {
```
jmrossy: yikes! Similar to what I wrote above, instead of this factory pattern how about we let users just call `new` and we do that currentProvider check in the constructor?

If we really want the factories still, maybe from `fromWeb3` would suffice.

packages/communication/src/node-communication-wrapper.ts
```diff
@@ -0,0 +1,379 @@
...
+  Provider,
+} from './commons'
+import { decodeStringParameter } from './utils/abi-utils'
+import { hasProperty } from './utils/provider-utils'
+import { DefaultRpcCaller, RpcCaller } from './utils/rpc-caller'
+import { TxParamsNormalizer } from './utils/tx-params-normalizer'
+import { toTxResult, TransactionResult } from './utils/tx-result'
+import { ReadOnlyWallet } from './wallet'
+
+const debugGasEstimation = debugFactory('communication:gas-estimation')
```
jmrossy: Is this right?

packages/communication/src/node-communication-wrapper.ts
```diff
@@ -0,0 +1,379 @@
...
+
+export interface CommunicationOptions {
+  gasInflationFactor: number
+  gasPrice: string
+  feeCurrency?: Address
+  from?: Address
+}
+
+export class NodeCommunicationWrapper {
```
jmrossy: Suggest `Connection` or `ConnectionManager`

packages/communication/src/utils/tx-params-normalizer.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  private async getChainId(): Promise<number> {
+    if (this.chainId === null) {
+      this.chainId = await this.communication.chainId()
+    }
+    return this.chainId
+  }
+
+  // @ts-ignore - see comment above
```
jmrossy: what comment is this referring to?

packages/communication/src/wallet.ts
```diff
@@ -0,0 +1,26 @@
+import { EIP712TypedData } from '@celo/utils/lib/sign-typed-data-utils'
+import { Address, CeloTx, EncodedTransaction, RLPEncodedTx } from './commons'
+
+export interface ReadOnlyWallet {
+  getAccounts: () => Address[]
```
jmrossy: Maybe it's too late or out of scope but should a Wallet be 1:1 with an account like Ethers.Js does:
https://docs.ethers.io/v5/api/signer/#Wallet

It would simplify this interface, we wouldn't need callers to pass in their own address every time. But it may create complexity in other places, just thinking out loud.

packages/communication/tslint.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": ["@celo/typescript/tslint.json"],
+  "rules": {
+    "no-global-arrow-functions": false,
+    "no-console": false,
```
jmrossy: intentional?

packages/celotool/src/lib/geth.ts
gastonponti: The problem that I see with adding the web3 to the constructor, is that we are trying to abstract that with the communication/connect layer.
The natural way, using constructors for this will be:
```
const web3 = new Web3(url)
const connect = new Connect(web3)
const kit = new ContractKit(connect)
```
that why the factory way of reducing that

packages/communication/src/commons.ts
gastonponti: 👍

packages/communication/src/utils/tx-params-normalizer.ts
gastonponti: Hahaha, nice catch. It was a comment removed from master 5 months ago:
```
    Right now, Forno does not expose a node's coinbase so we can't
    set the gatewayFeeRecipient. Once that is fixed, we can reenable
    this.
```
and its ignored because it's not used in the class
I will add again the comment, but this was basically a copy/paste, haha

packages/communication/package.json
gastonponti: Removed. Copy/paste from the kit

packages/communication/package.json
gastonponti: My only concern with this, is that if I remove those versions, and for example, we are using another version of the submodules in one of our packages (and it will happen when I fix again the versions for the notification-service, the ODIS combiner and the faucet), yarn could use those libs instead of the ones of the Web3 1.2.11.
It happened to me before and I tried to avoid that.

Worst case scenario, it's redundant, because it will download only the web3 package, but this will surely avoid us cross-packages versions with already dowloaded libs

dockerfiles/attestation-service/Dockerfile
gastonponti: Removed

packages/communication/.gitignore
gastonponti: Moved to `sdk`
and all the wallets to `sdk/wallets`

packages/communication/package.json
gastonponti: changed to `@celo/connect` :)

aaronmgdr:
2 questions. 

What is "the explorer"? given that we sometimes refer to blockscout as the explorer I could see that name getting confusing. 

Also not directly related to this but when are we going to have a 1.0 version?

packages/communication/package.json
gastonponti: nice catch. Removed

packages/communication/src/node-communication-wrapper.ts
gastonponti: `Connection`

packages/communication/tslint.json
gastonponti: removed

gastonponti:
> 2 questions.
> 
> What is "the explorer"? given that we sometimes refer to blockscout as the explorer I could see that name getting confusing.
> 
> Also not directly related to this but when are we going to have a 1.0 version?

Yes, I've just moved the explorer folder, didn't change the name. Basically, it's a block and logs consumer with the tools. Don't really know if blockscout is using that part of the old contractKit to show things (I have to check it). But basically could allow any user to consume that info in a simpler way, without the need of consume blockscout. More like the lib part of those kind of apps.

Good question, the 1.0 version decision is something more of the devX circle. Technically speaking, these changes doesn't mean that our lib is in a "production" state. Maybe we already think that we are on that state... I was not part of those discussions, but probably we could have a list of "must things" that we feel could change that version (adding @barbaraliau that has more context than me)

aaronmgdr:
Thanks for the context

barbaraliau:
I would imagine that it makes sense to move to 1.0 now that we are on mainnet and should be fairly stable. We haven’t discussed this In detail within devX but I’ll bring it up at the next meeting. It might make sense to have it coincide with the hard fork if possible

aaronmgdr:
@barbaraliau  yes, i've been wondering about when we were going to start releasing 1.0 and higher versions

dockerfiles/leaderboard/Dockerfile
gastonponti: Removed to use a released version of that package

packages/attestation-service/src/requestHandlers/attestation.ts
gastonponti: Add a `get web3` to also cover backwards compatibility

.circleci/config.yml
```diff
@@ -403,7 +405,32 @@ jobs:
...
+              --ignore @celo/env-tests \
+              --ignore @celo/web \
+              --ignore @celo/identity \
+              --ignore @celo/transactions-uri \
+              run test
+
+      - store_test_results:
+          path: test-results
+
+  web-test:
```
jmrossy, line 424: @aaronmgdr can you have a look at this? I vaguely remember web having some weirdness around it's CI testing a while ago. Does this look right to you?

packages/faucet/package.json
```diff
@@ -18,19 +18,19 @@
...
     "firebase": "^7.8.0",
     "firebase-admin": "^8.10.0",
     "firebase-functions": "^3.6.0",
     "twilio": "^3.39.3",
-    "web3": "1.2.4",
-    "web3-core-helpers": "1.2.4"
+    "web3": "1.2.11"
   },
   "devDependencies": {
+    "@celo/connect": "0.0.2-dev",
```
jmrossy: Why is this a dev dependency? If it's because we only use the types, I'd still recommend making it a runtime dep. It's easy for someone to not realize it's just a dev dep and then use actual functionality in the future, then not understand why it's broken.

packages/mobile/metro.config.js
```diff
@@ -8,7 +8,7 @@ const cwd = path.resolve(__dirname)
 const root = path.resolve(cwd, '../..')
 const escapedRoot = escapeStringRegexp(root)
 const celoRegex = new RegExp(
-  `${escapedRoot}\/packages\/(?!mobile|utils|base|contractkit|komencikit|react-components).*`
+  `${escapedRoot}\/packages\/(?!mobile|sdk\/utils|sdk\/base|sdk\/contractkit|komencikit|react-components|sdk\/connect|sdk\/identity|sdk\/network-utils|sdk\/wallets\/wallet-rpc|sdk\/wallets\/wallet-remote|sdk\/wallets\/wallet-base|sdk\/wallets\/wallet-local).*`
```
jmrossy: Nit: reorder to keep all the sdk ones together. The order doesn't matter here.

packages/mobile/package.json
```diff
@@ -149,13 +152,14 @@
...
     "vm-browserify": "^1.1.2",
-    "web3": "1.2.4"
+    "web3": "1.2.11"
   },
   "devDependencies": {
     "@apollo/react-testing": "^3.1.3",
     "@babel/plugin-proposal-decorators":"^7.12.1",
     "@celo/flake-tracker": "0.0.1-dev",
     "@celo/typescript": "0.0.1",
+    "@celo/connect": "0.0.2-dev",
```
jmrossy: Ditto feedback as for faucet here.

packages/mobile/src/icons/Logo.v2.tsx
```diff
@@ -0,0 +1,45 @@
+import colors from '@celo/react-components/styles/colors'
```
jmrossy: This should not be added. It's likely here because you had it hanging around from before the logo assets were moved out. Please remove.

packages/notification-service/src/blockscout/decode.ts
```diff
@@ -1,16 +1,16 @@
 import { notEmpty } from '@celo/utils/lib/collections'
 import * as utf8 from 'utf8'
-import Web3 from 'web3'
+import web3Abi, { AbiCoder } from 'web3-eth-abi'
 import { hexToUtf8 } from 'web3-utils'
 import { Log, Transfer } from './blockscout'
 
+const abiCoder = (web3Abi as unknown) as AbiCoder
```
jmrossy, line 7: Why is this double cast necessary?

packages/notification-service/src/blockscout/decode.ts
```diff
@@ -56,19 +56,9 @@ function decodeTransferLog(log: Log): Transfer | null {
...
-  /**
-   * Decode using the parameter signature for an ERC20 Transfer event
-   * For unknown reasons, blockscout includes an extra unknown param in the log's topics list
-   * Including this unknown param in the input list or decoding won't work
-   */
   try {
-    const decodedLog = web3.eth.abi.decodeLog(
+    const decodedLog = abiCoder.decodeLog(
       [
-        {
```
jmrossy, line 67: Is this change because blockscout no longer includes that extra unknown param in the topic list or because the updated abiDecoder handles it for us?

packages/faucet/package.json
```diff
@@ -18,19 +18,19 @@
     "build:rules": "firebase-bolt database-rules.bolt"
   },
   "dependencies": {
-    "@celo/contractkit": "0.3.6",
-    "@celo/utils": "0.1.9",
+    "@celo/contractkit": "0.4.19-dev",
+    "@celo/utils": "0.1.23-dev",
```
jmrossy: This package can't use `-dev` versions of the kits because those aren't published. This would break the deployment of the firebase function. We need to either revert these for now or get signoff from the package owners to have deployment broken until the new kit changes are all published.

packages/notification-service/package.json
```diff
@@ -16,8 +16,8 @@
     "deploy": "./deploy.sh"
   },
   "dependencies": {
-    "@celo/contractkit": "0.3.6",
-    "@celo/utils": "0.1.9",
+    "@celo/contractkit": "0.4.19-dev",
+    "@celo/utils": "0.1.23-dev",
```
jmrossy: Same as for the faucet package, this is an app engine service so it can only use published packages as deps.

packages/phone-number-privacy/combiner/package.json
```diff
@@ -24,17 +24,18 @@
     "db:migrate": "NODE_ENV=dev ts-node scripts/run-migrations.ts"
   },
   "dependencies": {
-    "@celo/contractkit": "0.4.17",
-    "@celo/phone-number-privacy-common": "1.0.14",
-    "@celo/utils": "0.1.21",
+    "@celo/contractkit": "0.4.19-dev",
+    "@celo/phone-number-privacy-common": "1.0.15-dev",
+    "@celo/utils": "0.1.23-dev",
```
jmrossy: Same as for the faucet package, this is a firebase function so it can only use published packages as deps.

packages/sdk/base/package.json
```diff
@@ -23,6 +23,7 @@
     "@celo/flake-tracker": "0.0.1-dev",
     "@celo/typescript": "0.0.1",
     "bignumber.js": "^9.0.0",
-    "elliptic": "^6.5.3"
+    "elliptic": "^6.5.3",
+    "web3-utils": "1.2.11"
```
jmrossy: just checking, was this intentional?

packages/sdk/connect/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "dependencies": {
+    "@celo/utils": "0.1.23-dev",
+    "debug": "^4.1.1"
+  },
+  "devDependencies": {
+    "web3": "1.2.11",
+    "web3-core": "1.2.11",
+    "web3-eth": "1.2.11",
+    "web3-eth-contract": "1.2.11",
+    "web3-utils": "1.2.11",
```
jmrossy: Should we make web3 a `peerDependency`?

packages/sdk/dappkit/package.json
```diff
@@ -9,11 +9,12 @@
     "@celo/utils": "0.1.23-dev",
     "expo": "^36.0.2",
     "expo-contacts": "8.0.0",
-    "libphonenumber-js": "^1.7.22",
-    "web3-eth-contract": "1.2.4"
+    "libphonenumber-js": "^1.7.22"
   },
   "devDependencies": {
+    "@celo/connect": "0.0.2-dev"
```
jmrossy: ditto as above here RE this being a dev dep

packages/sdk/identity/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@celo/identity",
+  "version": "0.0.1-dev",
```
jmrossy: We talked at one point about having the versions of all the packages match. Should this be higher? No strong preference, just raising the question.

packages/sdk/identity/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "description": "Celo's identity protocol",
```
jmrossy:
```suggestion
  "description": "Utilities for interacting with Celo's identity protocol",
```

packages/sdk/network-utils/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@celo/network-utils",
+  "version": "0.0.1-dev",
+  "description": "Celo's network utils",
```
jmrossy:
```suggestion
  "description": "Utilities for fetching static information about the Celo network",
```

jmrossy (commented):
In addition to my two previous review rounds, I've looked over the Circle CI, blockchain-api, faucet, mobile, notification-service, phone-number-privacy, dappkit, sdk/identity, sdk/network-utils, utils.

We still need reviews from the owners of: attestation-service, celotool, cli, dev-utils, docs, env-tests, komencikit, metadata-crawler, protocol, sdk/tx-uri, sdk/wallets.

And we should get addition reviews on mobile from the wallet team and the SDK changes itself from more dev-ex members.

packages/notification-service/src/blockscout/decode.ts
gastonponti: Our AbiCoder, that is the one that we have in our actual CK, basically changes function parameter (I think the AbiItem) with something that we add. So it make it incompatible (at least for types) when we try to cast it to our AbiCoder

packages/notification-service/src/blockscout/decode.ts
gastonponti: Well this is the change that I've told you about. It's not that blockscout add it an unknown one, it's that apparently, if the event has an owner, the first parameter is the topic of an address, but is not expected to be add that topic to the decoder.
That's why, I'm removing it:
```
// Docs: An array with the index parameter topics of the log, without the topic[0] if its a non-anonymous event.`
log.topics.slice(1).filter(notEmpty)
```
You can check it here: https://web3js.readthedocs.io/en/v1.2.11/web3-eth-abi.html#decodelog

What I think that was happening, is that we added that unknown for the input, because we were adding the first topic, without knowing this owner thing. This way, this first topic, basically was adding garbage to that `unknown` field, and as we didn't use it, it was discarded.

In that version of web3 they were probably not checking if it was valid, but in this case yes. That's why I require to use it as it should, and it's returning the same exactly data as before

packages/sdk/contractkit/package.json
```diff
@@ -11,68 +11,45 @@
   "keywords": [
```
medhakothari, line 11: Didn't we want to bump this (and all other packages) to 1.0.0?

packages/sdk/explorer/scripts/driver.ts
```diff
@@ -59,7 +60,8 @@ async function main() {
   // )
 
   // console.log(JSON.stringify(genEvents, null, 2))
-  ;(kit.web3.currentProvider as any).disconnect()
+  // ;(kit.connection.web3.currentProvider as any).disconnect()
```
medhakothari: stray comment?

packages/sdk/contractkit/src/wrappers/DowntimeSlasher.ts
```diff
@@ -210,6 +210,7 @@ export class DowntimeSlasherWrapper extends BaseSlasher<DowntimeSlasher> {
         ? startBlock + length - 1
         : (await this.kit.web3.eth.getBlockNumber()) - 2 // latest grandparent
```
medhakothari: should this be `this.kit.connection.getBlockNumber()`?

packages/celotool/src/cmds/account/invite.ts
```diff
@@ -48,7 +50,7 @@ export const handler = async (argv: InviteArgv) => {
     // TODO: this default gas price might not be accurate
     const gasPrice = 100000000000
 
-    const temporaryWalletAccount = await kit.web3.eth.accounts.create()
+    const temporaryWalletAccount = web3.eth.accounts.create()
```
medhakothari, line 53: Just for my understanding, is there a reason we didn't add this function as a part of `connection.ts?`

packages/celotool/src/e2e-tests/governance_tests.ts
```diff
@@ -1041,13 +1039,13 @@ describe('governance tests', () => {
...
-      await kit.web3.eth.personal.unlockAccount(validator, '', 1000000)
+      const validator = (await kit.connection.getAccounts())[0]
+      await kit.connection.web3.eth.personal.unlockAccount(validator, '', 1000000)
       const freezer = await kit._web3Contracts.getFreezer()
       await freezer.methods.freeze(epochRewards.options.address).send({ from: validator })
       blockFrozen = await web3.eth.getBlockNumber()
       epoch = new BigNumber(await validators.methods.getEpochSize().call()).toNumber()
-      await waitForBlock(kit.web3, blockFrozen + epoch * 2)
+      await waitForBlock(kit.connection.web3, blockFrozen + epoch * 2)
       latestBlock = await web3.eth.getBlockNumber()
```
medhakothari: Should `blockFrozen` and `latestBlock ` use `kit.connection.getBlockNumber()`?

packages/cli/src/commands/transfer/celo.ts
```diff
@@ -29,19 +28,19 @@ export default class TransferCelo extends BaseCommand {
     const value = new BigNumber(res.flags.value)
 
     this.kit.defaultAccount = from
-    const celoToken = await this.kit.contracts.getGoldToken()
+    const usdToken = await this.kit.contracts.getStableToken()
```
medhakothari: I'm a bit confused here, isn't this the command to transfer CELO? Why is it being changed to the stable token?

medhakothari (commented):
Looked through celotool, explorer, cli, docs, env-tests, dev-utils, wallet  and some other stray files. Everything pretty much looks good to go! 🎉  Left a couple small comments. Would be good to have relevant codeowners look over the attestation-service, komencikit, metadata-crawler, protocol, and wallets some more. 

Also some small notes (not important):
* Wanted to note that I see we used `web3.eth.getTransaction` and `web3.eth.getTransactionReceipt` in two files in celotool (describe.ts and list.ts). Wondering if for consistency we should use `kit.connection.getTransaction()` and `kit.connection.getTransactionReceipt()` instead?
* Noticed this with other functions as well, like `web3.eth.getAccounts` and `web3.eth.personal.unlockAccount` in celotool’s geth.ts file
* Similarly, cli's network/info.ts uses `kit.web3.eth.getBlockNumber()`, should we change it to `kit.connection.getBlockNumber()`?

packages/sdk/governance/src/proposals.ts
```diff
@@ -1,31 +1,45 @@
...
+  hotfixToParams,
+  Proposal,
+  ProposalTransaction,
+} from '@celo/contractkit/lib/wrappers/Governance'
+import { BlockExplorer, obtainKitContractDetails } from '@celo/explorer'
+import { isValidAddress } from '@celo/utils/lib/address'
+import { BigNumber } from 'bignumber.js'
+import debugFactory from 'debug'
+import { keccak256 } from 'ethereumjs-util'
+import * as inquirer from 'inquirer'
```
yorhodes, line 32: this isn't in the `package.json`

packages/sdk/governance/src/proposals.ts
```diff
@@ -144,18 +158,36 @@ export class ProposalBuilder {
...
     to: params.to,
     input: tx.encodeABI(),
   })
 
+  /**
+   * Adds a transaction to set the implementation on a proxy to the given address.
+   * @param contract Celo contract name of the proxy which should have its implementation set.
+   * @param newImplementationAddress Address of the new contract implementation.
+   */
+  addProxyRepointingTx = (contract: CeloContract, newImplementationAddress: string) => {
```
yorhodes, line 171: this is a relic, can be removed

packages/sdk/governance/src/proposals.ts
```diff
@@ -1,31 +1,45 @@
-import { Address, isHexString, trimLeading0x } from '@celo/utils/lib/address'
```
yorhodes, line 1: is there a `governance/proxy-utils` missing here?

yorhodes (commented):
`governance` module review

packages/sdk/wallets/wallet-base/src/signing-utils.ts
```diff
@@ -250,36 +245,3 @@ export function decodeSig(sig: any) {
     s: ethUtil.toBuffer(s) as Buffer,
   }
 }
-
-/**
- * Attempts each recovery key to find a match
- */
-export function recoverKeyIndex(
```
yorhodes, line 257: I assume this was moved to `wallet-hsm`

packages/sdk/wallets/wallet-ledger/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "lint": "tslint -c tslint.json --project ."
+  },
+  "dependencies": {
+    "@celo/utils": "0.1.23-dev",
+    "@celo/wallet-base": "0.0.1-dev",
+    "@celo/wallet-remote": "0.0.1-dev",
+    "eth-lib": "^0.2.8",
+    "ethereumjs-util": "^5.2.0",
+    "debug": "^4.1.1",
+    "@ledgerhq/hw-app-eth": "^5.11.0",
```
yorhodes, line 28: 🚀

patches/web3-core-helpers+1.2.4.patch
```diff
@@ -1,13 +0,0 @@
-patch-package
```
yorhodes, line 1: 🙏

yorhodes (commented):
`wallet` modules review

packages/sdk/transactions-uri/src/tx-uri.test.ts
```diff
@@ -1,25 +1,25 @@
-import { Tx } from 'web3-core'
-import { CeloContract } from '../base'
-import { newKit } from '../kit'
+import { CeloTx } from '@celo/connect'
+import { CeloContract, newKitFromWeb3 } from '@celo/contractkit'
+import { testWithGanache } from '@celo/dev-utils/lib/ganache-test'
 import { buildUri, parseUri } from './tx-uri'
 
-describe('URI utils', () => {
+testWithGanache('URI utils', (web3) => {
```
yorhodes, line 6: 🚀

yorhodes (commented):
`transactions-uri` review
I wonder if this should just be included in the contractkit itself but defer for now :)

packages/sdk/contractkit/package.json
```diff
@@ -11,68 +11,45 @@
...
-    "secp256k1": "^4.0.0",
-    "web3": "1.2.4",
-    "web3-core": "1.2.4",
-    "web3-core-helpers": "1.2.4",
-    "web3-eth": "1.2.4",
-    "web3-eth-abi": "1.2.4",
-    "web3-eth-contract": "1.2.4",
-    "web3-utils": "1.2.11",
-    "inquirer": "^7.0.5"
+    "inquirer": "^7.0.5",
```
yorhodes: I think this should be moved to `governance` module

packages/sdk/contractkit/package.json
```diff
@@ -11,68 +11,45 @@
...
+    "io-ts": "2.0.1"
```
yorhodes: I think this should be moved to `identity` module

packages/sdk/contractkit/package.json
```diff
@@ -11,68 +11,45 @@
...
+    "@celo/connect": "0.0.2-dev",
     "@celo/utils": "0.1.23-dev",
-    "@ledgerhq/hw-app-eth": "^5.11.0",
-    "@ledgerhq/hw-transport": "^5.11.0",
-    "@types/asn1js": "^0.0.2",
-    "asn1js": "^2.0.26",
-    "aws-sdk": "^2.705.0",
+    "@celo/wallet-local": "0.0.1-dev",
     "bignumber.js": "^9.0.0",
-    "blind-threshold-bls": "https://github.com/celo-org/blind-threshold-bls-wasm#e1e2f8a",
```
yorhodes, line 41: love to see it

packages/sdk/contractkit/package.json
```diff
@@ -11,68 +11,45 @@
...
   },
   "devDependencies": {
     "@celo/dev-utils": "0.0.1-dev",
     "@celo/flake-tracker": "0.0.1-dev",
     "@celo/ganache-cli": "git+https://github.com/celo-org/ganache-cli.git#21652da",
     "@celo/protocol": "1.0.0",
     "@types/debug": "^4.1.5",
-    "@types/ethereumjs-util": "^5.2.0",
-    "@types/qrcode": "^1.3.4",
+    "@types/inquirer": "^6.5.0",
```
yorhodes: ditto

packages/sdk/contractkit/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export { Address, NULL_ADDRESS } from '@celo/base/lib/address'
+export { AllContracts, CeloContract, CeloToken, RegisteredContracts } from './base'
+export { IdentityMetadataWrapper } from './identity'
```
yorhodes, line 3: this is a weird inclusion

packages/sdk/contractkit/src/proxy.ts
```diff
@@ -129,3 +129,8 @@ export const getInitializeAbiOfImplementation = (
   }
   return initializeAbi
 }
+
+export const setImplementationOnProxy = (address: string, web3: Web3) => {
```
yorhodes, line 133: I think this should be moved to the `governance` module

packages/sdk/contractkit/src/kit.ts
```diff
@@ -252,242 +208,135 @@ export class ContractKit {
...
+  }
+
+  // *** NOTICE ***
+  // Next functions exists for backwards compatibility
+  // These should be consumed via connection to avoid future deprecation issues
+
   addAccount(privateKey: string) {
-    assertIsCeloProvider(this.web3.currentProvider)
-    this.web3.currentProvider.addAccount(privateKey)
+    this.connection.addAccount(privateKey)
```
yorhodes, line 283: are these exposed on `contractkit` for backwards compatibility?
any reason why `kit.connection` isn't sufficient

packages/sdk/contractkit/src/wrappers/Accounts.ts
```diff
@@ -203,18 +200,22 @@ export class AccountsWrapper extends BaseWrapper<Accounts> {
...
+    const account =
+      this.kit.connection.defaultAccount || (await this.kit.connection.getAccounts())[0]
     if (await validators.isValidator(account)) {
-      const message = this.kit.web3.utils.soliditySha3({ type: 'address', value: account })
-      const prefixedMsg = hashMessageWithPrefix(message)
+      const message = this.kit.connection.web3.utils.soliditySha3({
+        type: 'address',
+        value: account,
+      })
+      const prefixedMsg = hashMessageWithPrefix(message!)
```
yorhodes: why did these become optionals?

yorhodes (commented):
`contractkit` module review

codyborn (approved):
Signing off for identity/ODIS changes.

packages/faucet/package.json
gastonponti: Well, I just double checked why is it needed and basically is for a debugging wrapper in the provider, that we added some time ago in the actual CeloProvider for the contractKit.

So it seems that it's outdated. I can remove that unnecessary wrapper and remove the dep 😄

packages/faucet/package.json
gastonponti: (this was discussed outside github but adding the response here to keep track in the PR)

Before merging (after checking that everything works as it should), we are going to release those dependencies packages with a `-beta` release, merge it, release the non-beta packages for everything, create a new PR pointing those packages to the fixed versions of the `non-beta packages`, and merge that.
After that all those packages versions will be fixed again. But it's easy to have all the migration changes in this PR, that revert those here, and add them in another PR

packages/sdk/base/package.json
gastonponti: yes, the `phoneNumbers.test.ts` uses `we3-utils` (don't know how was working, haha)

packages/notification-service/package.json
gastonponti: Same: https://github.com/celo-org/celo-monorepo/pull/4790#discussion_r527606408

packages/phone-number-privacy/combiner/package.json
gastonponti: same: https://github.com/celo-org/celo-monorepo/pull/4790#discussion_r527606408

packages/celotool/src/cmds/account/invite.ts
gastonponti: This is an account creation using web3. We don't have any of ours wallet that make that. So in a way, if the user set a wallet (the azure-hsm, for example), I don't know if it will be clear for the user, that we are just calling a pub/priv key creation, instead of the wallet set.  
I agree that is something that we could eventually add, but as it was never part of our kit, I prefer to leave it to a new feature/PR

packages/sdk/contractkit/src/wrappers/DowntimeSlasher.ts
gastonponti: Yes!
This was probably a merge after I added the `.web3` to the kit for backwards compatibility. That's why it does not fail, but as we probably we could deprecate that in the future, is something that we better migrate.
Nice catch! 😄

packages/sdk/explorer/scripts/driver.ts
gastonponti: Oh. The next line:
```
kit.connection.stop()
```
is the one that will make the disconnection. I'd should erased that line, not comment it, haha

packages/cli/src/commands/transfer/celo.ts
gastonponti: Wooow. Nice catch!
Was a change added in a merge, without any conflict.... (1e649a88b3b9c5c07df6bf2b9190c58bcb1c1477)

Ok, this does not reassure me one bit.......

I'm changing this as it is in master. And I will review all the differences again just in case something like this happened with other file.

Again, amazing catch

gastonponti: Ok, double checked. It was my mistake and it was in this commit (cf27409a1350a3efb52a491bf4ee5c15a2485d51)

So, my bad. I don't know what I was thinking...
At least my faith in git merging was restored, haha

.circleci/config.yml
aaronmgdr: as far as i can tell looks right, althoguh im not sure what attach_workspace does

aaronmgdr: should the name be web tests / more descriptive?

aaronmgdr: instead of jest test

packages/sdk/connect/src/wallet.ts
```diff
@@ -1,5 +1,16 @@
 import { EIP712TypedData } from '@celo/utils/lib/sign-typed-data-utils'
-import { RLPEncodedTx } from '../../utils/signing-utils'
+import { Address, CeloTx, EncodedTransaction, RLPEncodedTx } from './types'
+
+export interface ReadOnlyWallet {
```
mcortesi, line 4: doesn't look "readonly" to me.

it has remove account, it can sign... what's the read ony about ?

packages/sdk/connect/src/connection.ts
```diff
@@ -0,0 +1,403 @@
...
+    // Required otherwise is not backward compatible
+    const parameter = feeCurrency ? [feeCurrency] : []
+
+    // Reference: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gasprice
+    const response = await this.rpcCaller.call('eth_gasPrice', parameter)
+    const gasPriceInHex = response.result.toString()
+    return gasPriceInHex
+  }
+
+  async getBlockNumber(): Promise<number> {
```
mcortesi: i like how we replaced `web3` with `rpcCaller` for other methods. I would like this and the methosds below to also be replaced. Basically you need an rpcCaller (or something on top) that is able to parse inputs and parse outputs to proper types; probably more outputs that inputs). 

but it's a good idea to decouple from web3

packages/sdk/connect/src/celo-provider.ts
```diff
@@ -1,16 +1,15 @@
...
-
-const debug = debugFactory('kit:provider:connection')
-const debugPayload = debugFactory('kit:provider:payload')
-const debugResponse = debugFactory('kit:provider:response')
+import { Connection } from './connection'
+import { Callback, EncodedTransaction, JsonRpcPayload, JsonRpcResponse, Provider } from './types'
+import { hasProperty, stopProvider } from './utils/provider-utils'
+import { rpcCallHandler } from './utils/rpc-caller'
+
+const debug = debugFactory('provider:connection')
```
mcortesi, line 8: nit: we could move to a `debug.payload()` etc scheme :P

packages/sdk/contractkit/src/kit.ts
mcortesi: maybe we need a depreacted decorator, that logs a warning first time we use this method

mcortesi (commented):
reviewed contractkit, connection, and minor look at the rest of sdk pkgs

packages/notification-service/src/blockscout/decode.ts
gastonponti: I've checked another line, sorry.

This basically is web3 being web3. For some reason the `web3-eth-abi` lib exposes a `coder` as default, that is not fully compatible with the type they are defining. In these particular cases won't bring any problem, but if we wanted to use it with types we had to do this doble cast

packages/mobile/src/icons/Logo.v2.tsx
gastonponti: Remove all the file, or the import?
The `colors` lib is used in that file

gastonponti: Ok, I've checked with master and the whole file was removed.
Removing it

packages/sdk/governance/src/proposals.ts
gastonponti: I've added a Proxy file in the CK, because the proxies are things used in a few parts
Maybe you are referring to that

packages/sdk/governance/src/proposals.ts
gastonponti: we mention it as an example in celotool/src/cmds/deploy/upgrade/hotfix.ts
Maybe we can remove it in another PR. Don't want to keep adding things to this.

packages/sdk/wallets/wallet-base/src/signing-utils.ts
gastonponti: Yes!

packages/sdk/contractkit/package.json
gastonponti: As the claims and a few other things are used by the attestation wrapper, the identity package got divided in a weird way (I'm not 100% convinced about it). That's why we require that lib in both packages.

packages/sdk/contractkit/src/wrappers/Accounts.ts
gastonponti: Web3Utils can return a null, and we use it in hashMessageWithPrefix.
I've just changed it to force the utils to return a string

jmrossy (approved):
Approving for blockchain-api, faucet, mobile, notification-service, phone-number-privacy, dappkit, sdk/identity, sdk/network-utils, and utils.
There are still a few kinks to work out but I think this has reached a level of quality that we should move forward with it and sort out any remaining loose ends in separate PRs. Great job here @gastonponti !

barbaraliau (approved):
LGTM. Couple small things that can be fixed in another PR.

packages/sdk/connect/src/types.ts
```diff
@@ -6,7 +6,7 @@ export type Address = string
 export interface CeloParams {
   feeCurrency: string
   gatewayFeeRecipient: string
-  gatewayFee: string
+  gatewayFee: number | string
```
mcortesi: not sure if you want to make this a number, since numbers in js don't have enough precision to handle eth numbers. So it's a better practice to use string or BigNumbers

mcortesi (approved):
connection module.

PR #4793, opened 2020-08-25
Dappkit Redesign

### LINKED ISSUE #4644: [Wallet] DappKit should use new design system
### Expected Behavior

Uses styles similar to the rest of Valora

Design: [Figma](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet-Refresh?node-id=7424%3A468)

### Current Behavior

Uses old design which looks out of place

medhakothari:
### Description

Uses styles similar to the rest of Valora, follows the designs [here](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet-Refresh?node-id=7424%3A468)

### Other changes

n/a

### Related issues

- Fixes #4644 

### Backwards compatibility

none


packages/mobile/src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -119,42 +114,40 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     justifyContent: 'center',
     marginHorizontal: '15%',
+    marginVertical: '50%',
```
jeanregisser: I was surprised by this at first and did some more checks to be sure it works in all situations, 
turns out it starts clipping the content on smaller screens, like iPhone SE:



Here's the full styles that make it work on all screen sizes, no matter what the content is, the trick is with `flexGrow: 1` and fixing the button taking the full height:

```jsx
  container: {
    flex: 1,
  },
  scrollContainer: {
    flexGrow: 1,
    alignItems: 'center',
    justifyContent: 'center',
    marginHorizontal: '15%',
  },
  header: {
    ...fontStyles.h1,
    alignItems: 'center',
    paddingBottom: 16,
  },
  share: {
    ...fontStyles.subSmall,
    color: colors.gray4,
    fontSize: 16,
    textAlign: 'center',
  },
  sectionDivider: {
    alignItems: 'center',
    width: 200,
  },
  sectionHeaderText: {
    ...fontStyles.sectionLabelNew,
    fontSize: 13,
    marginTop: 16,
  },
  bodyText: {
    ...fontStyles.regular,
    fontSize: 16,
    color: colors.gray4,
    textAlign: 'center',
  },
  button: {
    marginTop: 24,
  },
})
```

Result on iPhone SE and iPhone 11:

packages/mobile/src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -1,4 +1,4 @@
-import Button, { BtnTypes } from '@celo/react-components/components/Button'
+import Button, { BtnSizes, BtnTypes } from '@celo/react-components/components/Button.v2'
 import colors from '@celo/react-components/styles/colors'
 import fontStyles from '@celo/react-components/styles/fonts'
```
jeanregisser: Redesigned screens should use `fonts.v2`.

packages/mobile/src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -119,42 +114,40 @@ const styles = StyleSheet.create({
...
-    marginBottom: 20,
+    paddingBottom: 16,
   },
   share: {
-    ...fontStyles.bodySecondary,
-    fontSize: 13,
-    alignSelf: 'center',
+    ...fontStyles.subSmall,
+    color: colors.gray4,
+    fontSize: 16,
```
jeanregisser: Once this is switched to `fonts.v2`, we should avoid redefining `fontSize`, unless absolutely unique and necessary.

packages/mobile/src/dappkit/DappKitSignTxScreen.tsx
```diff
@@ -1,4 +1,4 @@
-import Button, { BtnTypes } from '@celo/react-components/components/Button'
+import Button, { BtnSizes, BtnTypes } from '@celo/react-components/components/Button.v2'
 import colors from '@celo/react-components/styles/colors'
 import fontStyles from '@celo/react-components/styles/fonts'
```
jeanregisser: ditto about using `font.v2`

packages/mobile/src/dappkit/DappKitSignTxScreen.tsx
```diff
@@ -121,24 +117,18 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     justifyContent: 'center',
     marginHorizontal: '15%',
+    marginVertical: '50%',
```
jeanregisser: ditto about using `flexGrow: 1` to prevent content clipping on smaller screen sizes.

packages/mobile/src/dappkit/DappKitSignTxScreen.tsx
```diff
@@ -147,22 +137,25 @@ const styles = StyleSheet.create({
...
+    color: colors.gray4,
     textAlign: 'center',
   },
   underLine: {
     textDecorationLine: 'underline',
   },
+  button: {
+    flex: 1,
+    flexDirection: 'column',
+    alignItems: 'center',
```
jeanregisser: These can be removed when using `flexGrow: 1` in the `contentContainerStyle`.

jeanregisser (changes requested):
Thanks!! 👍 
Found an issue on smaller screen sizes and added recommendations to simplify font styles following the redesign.

medhakothari:
Thanks for the helpful comments @jeanregisser! Didn't know fonts v2 existed, switched to that now! Also changed the button styling and tested on smaller screens, works now :)

Screen Shot 2020-08-28 at 6 44 26 PM

packages/mobile/src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -76,10 +76,13 @@ class DappKitAccountAuthScreen extends React.Component<Props> {
     const { dappName } = route.params.dappKitRequest
     return (
       <SafeAreaView style={styles.container}>
+        <TopBarTextButton
+          title={t('cancel')}
+          onPress={this.cancel}
+          titleStyle={{ color: colors.dark }}
```
jeanregisser: nit: We try to avoid declaring styles inline with the component, unless necessary (animation or dynamic style).

> - By moving styles away from the render function, you're making the code easier to understand.
> - Naming the styles is a good way to add meaning to the low level components in the render function.

Source: https://reactnative.dev/docs/stylesheet.html

Another reason is about performance (though not the case here).
If the underlying component uses `PureComponent / React.memo`, passing a new style like this will defeat the render optimization when the parent component re-render.

packages/mobile/src/dappkit/DappKitSignTxScreen.tsx
```diff
@@ -74,10 +74,13 @@ class DappKitSignTxScreen extends React.Component<Props> {
 
     return (
       <SafeAreaView style={styles.container}>
+        <TopBarTextButton
+          title={t('cancel')}
+          onPress={this.cancel}
+          titleStyle={{ color: colors.dark }}
```
jeanregisser: ditto about the inline style

jeanregisser (approved):
Looks great! 👍 
Left one optional comment.

annakaz (approved):
🚀

LGTM overall! +1 to Jean's comment on avoiding inline styles

Lss-Ankit:
@jeanregisser  Can you please provide a build details to verify task. thanks!

PR #4805, opened 2020-08-25
Prevent accidental rotation of ECDSA validator key without BLS key

nategraf:
### Description

One of the most [serious downtime events](https://bisontrails.co/celo-incident-report-jul-09-2020/) on Celo Mainnet occurred because the validator accidentally rotated their ECDSA key without rotating their BLS key with it. Although this is a technically correct operation, there is no known user of this flow, and it is a sharp edge that does not need to exist.

This PR adds checks to the `celocli` commands which authorize new validator keys to require that the BLS key is always rotated with the ECDSA, unless the hidden `--force` flag is provided.

### Other changes

* Change `hasEnoughGold` to `hasEnoughCelo`

### Backwards compatibility

Does not break any known users workflow as the recommendation is to always provide the BLS key when rotating the ECDSA key.


packages/cli/src/commands/validator/update-bls-public-key.ts
```diff
@@ -4,8 +4,11 @@ import { displaySendTx } from '../../utils/cli'
 import { Flags } from '../../utils/command'
 
 export default class ValidatorUpdateBlsPublicKey extends BaseCommand {
-  static description =
-    'Update the BLS public key for a Validator to be used in consensus. Regular (ECDSA and BLS) key rotation is recommended for Validator operational security.'
+  static description = `Update the BLS public key for a Validator to be used in consensus.
+
+Regular (ECDSA and BLS) key rotation is recommended for Validator operational security.
+
+WARNING: By default, the BLS key used by the validator node is derived from the ECDSA private key. As a result, rotating the BLS key without rotating the ECDSA key will result in validator downtime without special configuration. use this method only if you know what you are doing.`
```
gastonponti:
```suggestion
WARNING: By default, the BLS key used by the validator node is derived from the ECDSA private key. As a result, rotating the BLS key without rotating the ECDSA key will result in validator downtime without special configuration. Use this method only if you know what you are doing.`
```

gastonponti (approved):
Minor suggestion.
LGTM

PR #4821, opened 2020-08-26
AWS full nodes

tkporter:
### Description

AWS full nodes! Huge shoutout to @jasonrudianto1 for starting this and making a ton of progress before his internship ended. This adds support for running the `celotooljs deploy (initial|upgrade|destroy) oracle-fullnode` command on AWS!

### Code structure

Now that we need to support both Azure & AWS for oracle full nodes, I created a new way of switching to clusters. There's a new directory, `packages/celotool/lib/k8s-cluster` that includes:
* `BaseClusterManager`: an abstract class that has common code for switching to a cluster that has a context kubectl is already aware of. Has a common interface for switching to a cluster where subclasses can have more specific code for switching to the subscription or getting the proper kubectl credentials
* `AKSClusterManager`: subclass of `BaseClusterManager` for Azure/AKS
* `AWSClusterManager`: subclass of `BaseClusterManager` for AWS/EKS
* a util function for getting the correct cluster manager given a cloud platform

I also created a similar approach for deploying full nodes. There's a new directory, `packages/celotool/lib/k8s-fullnode` that includes:
* `BaseFullNodeDeployer`: an abstract class that has common code for deploying/upgrading/destroying the full node helm chart, helm parameters that are common to any cloud provider, and an interface to allow subclasses to have more specific helm parameters or ways of managing IP addresses
* `AKSFullNodeDeployer`: a subclass of `BaseFullNodeDeployer` for Azure/AKS
* `AWSFullNodeDeployer`: a subclass of `BaseFullNodeDeployer` for AWS/EKS
* a util function for getting the correct full node deployer given a cloud platform

My general philosophy was to do my best in keeping cluster managers and full node deployers agnostic to an oracle context and related environment variables. Instead, they just require some config interfaces that can be generated in any way. Most of the oracle context related stuff (like reading oracle-specific env variables) happens in `oracle.ts`. The cloud provider (AWS or Azure) is taken from the prefix of the oracle context name. This means that an oracle context must start with either `AWS` or `AZURE`

#### NodePort

A previous iteration of this PR used `LoadBalancer` services, which (with a non-trivial amount of effort) worked well with TCP. However, there is no way for EKS network load balancers to have TCP & UDP ingress traffic for the same port. There is a restriction on the geth client where you can only configure a single port that is used for both TCP & UDP (in other words, you can't have 1 port be TCP and another be UDP).

Instead, I went with a NodePort service, which involves a single port that each VM in the cluster uses and routes traffic to the service. A node port is picked for each full node, and the geth client itself also uses that port. This is so the geth client properly advertises the correct `<ip>:<node-port>` combo for discovery.

One thing I'm looking for feedback on is that if there are 3 full nodes with node ports 30000, 30001, and 30002 respectively, because of the limitations of statefulsets each pod will expose all 3 of those ports, even though only 1 is actually used, so in theory another pod on the cluster could send traffic to the pod on a port that isn't actually listening for traffic. This was the easiest approach, but another approach is to have the geth client listen on the node port (for example, 30000), and then have another process on the port redirect all traffic to port 30303 to 30000. That way, only 30303 has to be exposed on the pod. This felt a little fragile though.

### Other changes

I discovered a problem where if a node (like a full node, or tx node, or whatever) was failing its readiness check (let's say the pod is not synced for some reason), then any services that point to it will not route traffic to the pod. This is very much intended for certain services, like the ClusterIP RPC service the oracles use, or forno, etc. However, this was also affecting the `LoadBalancer`/`NodePort` services that would point to each individual node to allow ingress TCP/UDP protocol traffic to the node. If a node isn't syncing well we definitely want the load balancer to still work so that other nodes can help the node sync! The fix to this is just to add `publishNotReadyAddresses: true` to the spec of the service. 

### Tested

Lots of AWS deployments. Need to test Azure deployments in more detail (semi-blocked because of the pod security policy enforcement on the test cluster that hasn't been merged to master yet), but no glaring issues.

### Backwards compatibility

Yes :)


packages/celotool/src/lib/helm_deploy.ts
```diff
@@ -175,13 +175,26 @@ export async function redeployTiller() {
...
   }
 }
 
+async function initHelm(serviceAccountName: string) {
+  // If the server version is >= 1.16, we need to modify the helm init command.
+  // See https://github.com/helm/helm/issues/6374#issuecomment-533427268
+  let cmd = `helm init --service-account=${serviceAccountName}`
+  const kubeServerVersion = await getServerVersion()
+  if (kubeServerVersion.major >= 1 && kubeServerVersion.minor >= 16) {
+    cmd = `${cmd} --override spec.selector.matchLabels.'name'='tiller',spec.selector.matchLabels.'app'='helm' --output yaml | sed 's@apiVersion: extensions/v1beta1@apiVersion: apps/v1@' | kubectl apply -f -`
```
jcortejoso, line 193: Nice to have a oneliner for this.

packages/helm-charts/common/templates/_helpers.tpl
```diff
@@ -143,6 +143,8 @@ release: {{ .Release.Name }}
     ADDITIONAL_FLAGS="${ADDITIONAL_FLAGS} --pprof --pprofport {{ .pprof_port }} --pprofaddr 0.0.0.0"
     {{- end }}
 
+{{ .extra_setup }}
```
jcortejoso, line 152: Nice using this + template for the value

packages/celotool/src/lib/k8s-fullnode/aws.ts
```diff
@@ -0,0 +1,77 @@
...
+      ...(await super.additionalHelmParameters()),
+      `--set aws=true`,
+      `--set storage.storageClass=gp2`,
+    ]
+  }
+
+  /**
+   * Prints action required to remove the node ports, and removes the chart
+   */
+  async removeChart() {
```
jcortejoso, line 35: Nice

packages/celotool/src/lib/k8s-fullnode/aws.ts
```diff
@@ -0,0 +1,77 @@
...
+    await super.removeChart()
+  }
+
+  /**
+   * In order for NodePort to work on an EKS cluster, you need to manually
+   * allow ingress traffic to the specific ports for each VM in the cluster.
+   * This prints a message showing which ports need to be added and how to modify
+   * the security group that is common to all VMs in a cluster.
+   */
+  printNodePortsActionRequired(nodePorts: number[], destroy: boolean = false) {
```
jcortejoso: I am OK with not automating this and taking as a manual action, but I am curious if you think this could be automated to authorizing here.

jcortejoso (approved):
Wow! Awesome work :clap:

packages/celotool/src/lib/k8s-fullnode/aws.ts
tkporter: Tbh didn't really think of it-- it feels very automatable (if that's a word?) so I'll look into it

PR #4850, opened 2020-08-26
[Wallet] Hide CTAs with verification when insufficient balance

### LINKED ISSUE #4748: [Wallet] Users SNBAT see "Confirm number" call to action if they have 0 balance
### Expected Behavior

"Confirm number" call to action is hidden if a user has 0 balance

### Current Behavior

"Confirm number" call to action is shown on home page regardless of user's balance. If they have 0 balance, they will not be able to verify after clicking this call to action

### LINKED ISSUE #4797: Verification Prompts should not show if a user has zero balance
### Expected Behavior

A user should not see prompts to verify their number if they have a zero balance

### Current Behavior

- Notification shown in the home screen
- prompt to confirm number shown in send and request flows

i1skn:
### Description

This hides CTAs with verification when balance is insufficient to retrieve a pepper if it's not cached.

### Tested

iOS simulator

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/4748
- Fixes https://github.com/celo-org/celo-monorepo/issues/4797

### Backwards compatibility

Yes


packages/contractkit/src/identity/odis/phone-number-identifier.test.ts
```diff
@@ -0,0 +1,7 @@
+import { isSufficientBalanceForQuotaRetrieval } from './phone-number-identifier'
+describe(isSufficientBalanceForQuotaRetrieval, () => {
+  it('identifies sufficient balance correctly', () => {
+    expect(isSufficientBalanceForQuotaRetrieval(0.09)).toBe(false)
```
aslawson: nice :) thanks for adding!

packages/mobile/src/identity/verification.ts
```diff
@@ -64,6 +68,13 @@ export interface AttestationCode {
   issuer: string
 }
 
+export function* isSufficientBalance() {
+  const e164Number: string = yield select(e164NumberSelector)
+  const cachedSalt = yield call(getSaltCache, e164Number)
+  const isSufficientForQuota = cachedSalt ? true : yield call(balanceSufficientForQuotaRetrieval)
+  yield put(setVerificationPossible(isSufficientForQuota))
+}
```
jeanregisser: I feel we shouldn't introduce an action and state for this.
If it can be derived from the existing state, we should derive it directly and use a selector.

jeanregisser (changes requested):
Thanks! See my comment.

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -60,8 +63,8 @@ export function* fetchPhoneHashPrivate(e164Number: string) {
 function* doFetchPhoneHashPrivate(e164Number: string) {
   const account: string = yield call(getConnectedAccount)
   Logger.debug(`${TAG}@fetchPrivatePhoneHash`, 'Fetching phone hash details')
-  const saltCache: E164NumberToSaltType = yield select(e164NumberToSaltSelector)
-  const cachedSalt = saltCache[e164Number]
+
+  const cachedSalt = yield getSaltCache(e164Number)
```
jmrossy: should be yield call

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -85,6 +92,17 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
   return details
 }
 
+export function* getSaltCache(e164Number: string) {
+  const saltCache: E164NumberToSaltType = yield select(e164NumberToSaltSelector)
+  return saltCache[e164Number]
+}
+
+export function* balanceSufficientForQuotaRetrieval() {
```
jmrossy: Duped with the one in CK now right?

packages/mobile/src/identity/verification.ts
```diff
@@ -64,6 +68,13 @@ export interface AttestationCode {
   issuer: string
 }
 
+export function* isSufficientBalance() {
```
jmrossy: Nit: the `is` prefix is misleading for a saga that will update state.

packages/mobile/src/identity/verification.ts
```diff
@@ -64,6 +68,13 @@ export interface AttestationCode {
...
+  const e164Number: string = yield select(e164NumberSelector)
+  const cachedSalt = yield call(getSaltCache, e164Number)
```
jmrossy: Use the `getUserSelfPhoneHashDetails` function in privateHashing so we can keep all interaction with these salt related states exlcusively in the privateHashing.ts file

packages/mobile/src/identity/verification.ts
```diff
@@ -64,6 +68,13 @@ export interface AttestationCode {
...
+  const isSufficientForQuota = cachedSalt ? true : yield call(balanceSufficientForQuotaRetrieval)
+  yield put(setVerificationPossible(isSufficientForQuota))
```
jmrossy: It seems like we didn't need new state and a new watcher to compute this. We could have used a reselect selector to compute it since it's purely derived from other state.

packages/mobile/src/identity/saga.ts
```diff
@@ -108,6 +109,12 @@ function* watchFetchDataEncryptionKey() {
   yield takeLeading(Actions.FETCH_DATA_ENCRYPTION_KEY, fetchDataEncryptionKeyWrapper)
 }
 
+function* watchBalanceChange() {
+  // TODO: add check for CELO balance as well
+  yield isSufficientBalance()
```
jmrossy: yield call

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -85,6 +92,17 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
   return details
 }
 
+export function* getSaltCache(e164Number: string) {
```
jmrossy: Nit: getCachedSalt is more accurate

packages/contractkit/src/identity/odis/phone-number-identifier.test.ts
```diff
@@ -0,0 +1,7 @@
+import { isSufficientBalanceForQuotaRetrieval } from './phone-number-identifier'
```
jmrossy: Thanks for the test!

packages/mobile/src/identity/verification.ts
i1skn: I've tried to use `verificationStatus` for this, but then in the situation when we need to change it by the reasons unrelated to quota - it got mixed up, so decided to have it as a separate flag. It felt for me not as a status, but rather a precondition. Do you have a better idea on how we can derive it?

jeanregisser: From what I see, this logic could be done in a selector instead.
Happy to have a deeper look at this together (slack or call).

packages/mobile/src/identity/privateHashing.ts
i1skn: This one is a saga and the one in CK is a pure function. `balanceSufficientForQuotaRetrieval` is a client of `isSufficientBalanceForQuotaRetrieval`

i1skn:
@jeanregisser @jmrossy thanks for the great feedback! I've addressed it - PTAL.

packages/mobile/src/app/selectors.ts
```diff
@@ -19,3 +23,13 @@ export const getLastTimeBackgrounded = (state: RootState) => {
...
+
+export const verificationPossibleSelector = (state: RootState) => {
+  const e164Number = e164NumberSelector(state)
+  const dollarBalance = stableTokenBalanceSelector(state)
+  const saltCache = e164NumberToSaltSelector(state)
+  return !!(
+    (e164Number && saltCache[e164Number]) ||
+    (dollarBalance && isSufficientBalanceForQuotaRetrieval(dollarBalance))
+  )
+}
```
jeanregisser: great!
Wondering, should we also include the check for `!state.app.numberVerified` in there?

packages/mobile/src/app/selectors.test.ts
```diff
@@ -0,0 +1,41 @@
+import { verificationPossibleSelector } from 'src/app/selectors'
+import { getMockStoreData } from 'test/utils'
+import { mockE164Number, mockE164NumberPepper } from 'test/values'
+
+describe(verificationPossibleSelector, () => {
+  it('returns true', () => {
```
jeanregisser: nit: it's more helpful when the description gives more details about the expectations. 
for instance `it returns true when the number pepper is already cached or when the balance is sufficient to fetch it`.

jeanregisser (approved):
Awesome! 💯  Left two optional comments.

packages/mobile/src/app/selectors.ts
i1skn: I think we should keep this separate, cause they mean different things - "is it possible?" and "was it done already?". Although I'm not sure if it was already done - is it still possible 🤔 ?

packages/mobile/src/app/selectors.ts
```diff
@@ -19,3 +23,13 @@ export const getLastTimeBackgrounded = (state: RootState) => {
 export const sessionIdSelector = (state: RootState) => {
   return state.app.sessionId
 }
+
+export const verificationPossibleSelector = (state: RootState) => {
```
jmrossy, line 28: Should this take into account the $ required for verification itself? The ODIS min balance just dropped to 0.01 so if a user has 3 cents this will still return true though verification isn't actually possible.

packages/contractkit/src/identity/odis/phone-number-identifier.ts
```diff
@@ -11,6 +12,10 @@ import {
   SignMessageResponse,
 } from './query'
 
+// ODIS minimum dollar balance for pepper quota retrieval
+// TODO change this to new ODIS minimum dollar balance once deployed
+export const ODIS_MINIMUM_DOLLAR_BALANCE = 0.1
```
jmrossy: As of this morning, this is now 0.01 cUSD or 0.005 CELO. Can you adjust this and add gold?

packages/contractkit/src/identity/odis/phone-number-identifier.ts
```diff
@@ -85,3 +90,10 @@ export function getPepperFromThresholdSignature(sigBuf: Buffer) {
     .digest('base64')
     .slice(0, PEPPER_CHAR_LENGTH)
 }
+
+/**
+ * Check if balance is sufficient for quota retrieval
+ */
+export function isSufficientBalanceForQuotaRetrieval(balance: BigNumber.Value) {
```
jmrossy: I should have pointed this out in Amy's PR but the name here is misleading. Recommend `isBalanceSufficientForSigRetrieval`

packages/mobile/src/identity/privateHashing.ts
```diff
@@ -94,11 +92,10 @@ function* doFetchPhoneHashPrivate(e164Number: string) {
   return details
 }
 
-// TODO move to Contract kit ODIS utils
 export function* balanceSufficientForQuotaRetrieval() {
```
jmrossy: ditto about name

jmrossy (commented):
Just a few more tweaks needed I think!

jmrossy (approved):

Lss-Ankit:
Hi @jeanregisser  i have verified task using Test flight build v1.1.0 (20) and Android play store build v1.1.0 (1004294308) and found below new issue: 

**Test scenarios:**
- Verify user not be able to see "Phone number verification CTA" on home screen, "Start Verification" option from Send & receive screen , "Confirm Phone Number" option from Setting when user sign up without invite code with 0 balance. 
- Verify user is able to see "Phone number verification CTA" on home screen, "Start Verification" option from Send & receive screen , "Confirm Phone Number" option from Setting when user received $ 0.01 cUSD from any other user. 
- Verify Confirm phone number CTA is removed when user spent all balance and now balance issue 0 now (Issue #5018) 
- Verify "Confirm Phone Number" screen is open with "Your current Balance is not high enough..." description when user try to confirm phone number with low balance. 
- Verify Confirm phone number CTA is not shown when user sign in with Account has 0 balance.
- 

**New Issue:**
- https://github.com/celo-org/celo-monorepo/issues/5018

**Need Clarification:**
- Confirm Phone number CTA is shown even though user has only $ 0.01 cUSD, can you please confirm is it expected?  

Thanks !

PR #4856, opened 2020-08-26
[cli] Remove abstract release-gold command

gastonponti:
### Description

The releasegold:release-gold is not a valid command (just a base one). Remove it from the cli


yorhodes (approved):
LGTM

PR #4858, opened 2020-08-26
[Wallet] Prevent getting stuck on PIN screen

### LINKED ISSUE #4486: User Stuck in Enter PIN page while trying to set up Account Key for second time  by tapping Account Key option from Menu
**Frequency:** 100%
**Repro App version:**  IOS Test flight Celo integration build v1.9.1 (41)
**Repro on device:** iphone 7(13.5.1), iphone 11(14.0)
**Repro Steps:**

1. Launch the Application and signup with test account
2. Tap on Set Up Now button of Your Account Key In-app notification
3. Make sure user redirect to Your Account key page
4. Tap on Hamburger Menu and tap on Account Key option
5. Enter the correct PIN 
6. Observe 

**Current Behavior:** User is getting stuck in Enter PIN page while trying to set up account key for second time  
**Impact:** User is not able to set up account key
**Expected Behavior:** After entering PIN user should redirect to Your Account Key page
**Attachment:** [StuckinEnterPINscreenfromNotificationIssue.mp4](https://drive.google.com/file/d/1aYrgv7S9YBOY5jpFwJ0R8CCMtMPEdP6v/view?usp=sharing)

gnardini:
### Description

The Account Key drawer item is 'protected' which means that when a user presses it the PIN enter screen shows up. The problem was that if you are already on that screen and you select the Account Key item again the pin is requested but when you finish and navigate to the Account Key screen nothing happens because you are already there.

### Tested

On simulator

### Related issues

- Fixes #4486


jmrossy (approved):

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -78,6 +80,7 @@ function CustomDrawerItemList({
   ...passThroughProps
 }: CustomDrawerItemListProps) {
   const buildLink = useLinkBuilder()
+  const navigationState = useNavigationState((navState) => navState)
```
jeanregisser: This will cause this component to re-render on every navigation state change, which might be ok but should better be avoided.

Since this is used only in `onPress`, what do you think of using `navigation.dangerouslyGetState()` there instead?
(though maybe this won't work quite right because of nesting, and we might need to get the parent navigation).

jeanregisser (commented):
Nice!

packages/mobile/src/navigator/DrawerNavigator.tsx
gnardini: Thanks for catching this! It did actually work since we were already passing the parent navigation :)

jeanregisser (approved):

Lss-Ankit:

Test scenarios:

1. enter PIN is not asked now when user taps on Account Key option from hamburger menu if user is already on Account Key screen from CTA
2. Enter PIN is asked when user taps on Account option from Hamburger menu if user is not in Account key screen

Devices: iPhone 6+ (12.4.5),

Fixed issues:

#4486 

Need Clarification: 
- Enter PIN is not asked when user taps on "Set up Now" button from Your account key.." CTA 

**can you please confirm is it expected?** 

Thanks !

PR #4861, opened 2020-08-26
adds tests for missing timestamp, fix log error

alecps:
### Description

Adds test to confirm backwards compatibility. Fixes minor reporting error: Signer logs were saying that query count was not incremented when in fact it was.

### Other changes

None

### Tested

With e2e tests

### Related issues

NA

### Backwards compatibility

Yes


packages/phone-number-privacy/signer/src/database/wrappers/account.ts
```diff
@@ -41,6 +41,7 @@ export async function incrementQueryCount(account: string) {
       await accounts()
         .where(ACCOUNTS_COLUMNS.address, account)
         .increment(ACCOUNTS_COLUMNS.numLookups, 1)
+      return true
```
alecps, line 44: This fixes the super small logging error. It wasn't going to break anything, but might have led to confusion while debugging.

codyborn (approved):

jmrossy (approved):

PR #4872, opened 2020-08-27
[Wallet] Reorganize and add e2e tests

gnardini:
### Description

All e2e were in a single file called `Transfer.spec.js`. This was fine when we had few tests, but as we added more having everything in a single file was starting to get messy and it was going to get worse.

I tried different ways to separate tests into different files, but all tests depend on the onboarding flow, so we couldn't have truly standalone test files unless we wanted to go through the onboarding flow on each one which would make running the tests take significantly longer.

In the end, I added one main file `TestRunner.spec.js` which simply calls the tests in all other files in the order that we want. This way, the app is not restarted between tests so they run quicker and it's easy to comment out or skip some if we want to run a single test file.

### Other changes

- Added a test for the 'Reset Valora' flow.
- Added a `-d` option to the `run_e2e.sh` script to simplify development.
- Updated README.


packages/mobile/src/account/saga.ts
```diff
@@ -109,9 +109,15 @@ function* clearStoredAccountSaga({ account }: ClearStoredAccountAction) {
     yield call(removeAccountLocally, account)
     yield call(clearStoredMnemonic)
     yield call(ValoraAnalytics.reset)
-    yield call(firebaseSignOut, firebase.app())
     yield call(deleteNodeData)
 
+    // Ignore error if it was caused by Firebase.
+    try {
+      yield call(firebaseSignOut, firebase.app())
```
jeanregisser, line 117: Why would this fail?

jeanregisser (approved):
w00t! Great changes! 🚀

packages/mobile/src/account/saga.ts
gnardini: If Firebase isn't initialized it fails. It failed for me on the test build, since I guess it's not enabled there, and just in case something happens in the production build I think it's better to have this check because if this fails the account is actually deleted but the user is stuck on the settings page so they need to kill the app and open it again to have it work.

jeanregisser: Oh yes, firebase init is under a config.
This looks good then, however we should only log the error `if (FIREBASE_ENABLED)` to prevent confusions as to what error is expected.

PR #4898, opened 2020-08-30
Allow overriding the hostAddress and hostPort for truffle configs

prasincs:
### Description

This PR keeps the behavior for running truffle commands the same for initializing the test network. We run multiple instances of celo for testing or similar networks and have to override the ports for testing, this allows us to be able to point to a different host/port combination for this, while keeping the default as is.

### Other changes

None, but if any other tool uses NODE_ADDRESS or NODE_PORT for similar function, it might conflict with this changes, I'm open to making a CELO_ prefix in addition 

### Tested

tested by running yarn init

### Related issues

None

### Backwards compatibility

Should be backwards compatible for cases other than I mentioned above


packages/protocol/truffle-config.js
```diff
@@ -38,10 +38,12 @@ const BAKLAVASTAGING_FROM = '0x4588ABb84e1BBEFc2BcF4b2296F785fB7AD9F285'
 
 // Gas limit is doubled for initial contract deployment.
 const gasLimit = argv.reset ? 20000000 : 12500000
+const hostAddress = process.env.NODE_ADDRESS || "127.0.0.1"; 
+const hostPort = parseInt(process.env.NODE_PORT || "8545");
```
nategraf:
```suggestion
const hostAddress = process.env.CELO_NODE_ADDRESS || "127.0.0.1"; 
const hostPort = parseInt(process.env.CELO_NODE_PORT || "8545");
```

nategraf (approved):
Good tweak! Thanks for submitting a PR! I would actually suggest using the `CELO_` prefix. I've added a comment with the suggestion so you can just hit commit suggestion. 🙏

barbaraliau:
@prasincs is this PR still relevant? If so, can you resolve the conflicts?

prasincs:
@barbaraliau it is indeed, sorry for the delay I have resolved the conflict

nambrot (approved):

PR #4903, opened 2020-08-31
[Wallet] Add daily limit to CELO withdrawals

### LINKED ISSUE #4895: Apply 500 cUSD daily limit to CELO Withdrawals & cUSD Requests
### Expected Behavior

Apply the daily limit of 500 cUSD to CELO Withdrawals & cUSD Requests in mobile app

### Current Behavior

500 cUSD limit applies to cash in & cUSD send, but not cUSD request + CELO Withdrawal

NOTE: Blocker for v1.1, not v1.0.x

gnardini:
### Description

Added a limit to CELO withdrawals. They contribute to and use the same limit that cUSD transfers do. 

### Other changes

- Limited the limit error message to 2 decimal places since it was very long otherwise because of exchange rates.
- Added a $500 limit to requests as well, since they would never be able to be completed otherwise.

### Tested

In the simulator sending/withdrawing different amount of funds.

### Related issues

- Fixes #4895


packages/mobile/src/tokens/saga.ts
```diff
@@ -24,24 +24,24 @@ const contractWeiPerUnit: { [key in CURRENCY_ENUM]: BigNumber | null } = {
   [CURRENCY_ENUM.DOLLAR]: null,
 }
 
-function* getWeiPerUnit(token: CURRENCY_ENUM) {
+async function getWeiPerUnit(token: CURRENCY_ENUM) {
```
gnardini: I had to change these generator functions to async ones because I needed to call them from a React component, if there's a different or better way to do this that we've been using and I missed please let me know :)

packages/mobile/locales/en-US/global.json
```diff
@@ -125,6 +125,7 @@
   "odisInsufficientBalance": "Your balance is too low to complete this process",
   "matchmakingQuotaExceededError": "Matchmaking can be done only once",
   "paymentLimitReached": "The daily limit is {{dailyLimitcUSD}} cUSD ({{currencySymbol}}{{dailyLimit}})\n{{dailyRemainingcUSD}} cUSD ({{currencySymbol}}{{dailyRemaining}}) remaining",
+  "requestLimitError": "Can't request more than the {{limit}} cUSD daily limit.",
```
tarikbellamine, line 128: Can we repurpose `paymentLimitReached` message to fit this?

gnardini: Do you mean the specific wording? `paymentLimitReached` ends with `{X} cUSD remaining` but in the case of requesting the remaining part doesn't make much sense I think, right?

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
```diff
@@ -43,11 +55,25 @@ function WithdrawCeloScreen({ navigation }: Props) {
...
+    const now = Date.now()
+    const isLimitReached = isPaymentLimitReached(now, recentPayments, cUsdAmount.toNumber())
+    if (isLimitReached) {
+      dispatch(
+        showLimitReachedError(now, recentPayments, localCurrencyExchangeRate, localCurrencySymbol)
+      )
+      setLoading(false)
+      return
+    }
+
```
tarikbellamine: Can we organize this into a separate function?

gnardini: Yup, probably a good idea to reuse this logic between this file and the one in the Send flow too.

packages/mobile/src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -79,10 +78,7 @@ export function ExchangeTradeScreen({ navigation, route }: Props) {
   }
 
   function goNext() {
-    const now = Date.now()
-    const isLimitReached = isPaymentLimitReached(now, recentPayments, dollarAmount.toNumber())
-    if (isLimitReached) {
-      dispatch(showLimitReachedError(now, recentPayments, localExchangeRate, localCurrencySymbol))
+    if (isPaymentLimitReached(dollarAmount)) {
```
tarikbellamine: Nit: I think this is much cleaner overall. Only thing is it's not intuitive to me that the error is being shown within `isPaymentLimitReached`. I'd suggest either changing the function name to try to capture that or breaking out the `showError` to be within the if block (i.e., `isPaymentLimitReached` would only return a boolean but not display an error either way)

packages/mobile/src/send/utils.ts
```diff
@@ -93,13 +96,29 @@ function dailySpent(now: number, recentPayments: PaymentInfo[]) {
   return amount
 }
 
-export function isPaymentLimitReached(
-  now: number,
-  recentPayments: PaymentInfo[],
-  initial: number
+export function isPaymentLimitReached(amountToTransfer: BigNumber): boolean {
+  return _isPaymentLimitReachedWithStore(store.getState(), amountToTransfer)
```
tarikbellamine: Nit: I think it's cleaner to have `store.getState` called within the inner function rather than here as a function argument. The function name `_isPaymentLimitReachedWithStore` implies that we are checking the store and there realistically isn't ever another value we'd want to use instead as an argument

packages/mobile/src/send/utils.ts
```diff
@@ -108,12 +127,15 @@ export function showLimitReachedError(
   localCurrencyExchangeRate: string | null | undefined,
   localCurrencySymbol: LocalCurrencySymbol | null
 ) {
-  const dailyRemainingcUSD = dailyAmountRemaining(now, recentPayments)
-  const dailyRemaining = convertDollarsToLocalAmount(dailyRemainingcUSD, localCurrencyExchangeRate)
+  const dailyRemainingcUSD = dailyAmountRemaining(now, recentPayments).toFixed(2)
```
tarikbellamine, line 153: Nice catch here with the decimals!

packages/mobile/src/fiatExchanges/FiatExchangeAmount.tsx
gnardini: I agree it's not intuitive, the reason I wanted to leave it here is that the `showLimitReachedError` has a lot of dependencies that have to be imported in every place it's used just to show the error message and I though the usage would be much simpler if we could 'hide' those dependencies within this method. I'll try to think of a rename to the function to make clear what's happening.

packages/mobile/src/send/utils.ts
gnardini: It's true it may lead to confusion, I use it from the test and it's easier to call it with a mock state, so I just renamed the function to say `state` instead of `store`, hope that solves the concern.

gnardini: Writing here for other reviewers, I ended up changing this code to be a custom hook instead `useDailyTransferLimitValidator`, I think it's cleaner this way :)

packages/mobile/src/send/utils.ts
```diff
@@ -93,13 +99,41 @@ function dailySpent(now: number, recentPayments: PaymentInfo[]) {
...
+      return amount
+    } else {
+      const exchangeRate = getRateForMakerToken(
+        exchangeRatePair,
+        CURRENCY_ENUM.DOLLAR,
+        CURRENCY_ENUM.GOLD
+      )
+      return goldToDollarAmount(amount, exchangeRate) || new BigNumber(0)
+    }
+  }, [amount])
```
tarikbellamine: Think you want currency here as well for the edge case that the currency changes but the amount doesn't

packages/mobile/src/send/utils.ts
```diff
@@ -93,13 +99,41 @@ function dailySpent(now: number, recentPayments: PaymentInfo[]) {
...
+
+  const recentPayments = useSelector(getRecentPayments)
+  const localCurrencyExchangeRate = useSelector(getLocalCurrencyExchangeRate)
+  const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
+
+  const now = Date.now()
```
tarikbellamine, line 127: Nit nit: if you aren't reusing `now`, you can call `Date.now` in the function argument to avoid this extra variable/line

tarikbellamine (approved):
Good stuff! Left some minor comments

packages/mobile/src/send/utils.ts
gnardini: It's actually being used twice (lines 129 & 133) :)

jeanregisser (approved):
Great! 👍

Lss-Ankit:
Need fund to verify this task

PR #4912, opened 2020-08-31
Attestations Service: Support Nexmo applications

timmoreton:
### Description

Enable support for passing `applicationId` in to the Nexmo provider.

### Tested

Created a Nexmo application with a status URL and cleared the default callback URL. Tested on a local instance. Got delivery receipts.

### Backwards compatibility

Yes. Parameter defaults to empty so that configs need not be changed.


packages/docs/operations-manual/attestation-service.md
```diff
@@ -227,7 +231,9 @@ celocli account:get-metadata $CELO_VALIDATOR_RG_ADDRESS
...
 
-Nexmo requires manual configuration to enable this. Go to [Settings](https://dashboard.nexmo.com/settings), and under `Delivery Receipts`, enter the external URL of your Attestation Service appended by `/delivery_status_nexmo` -- for example `http://1.2.3.4:80/delivery_status_nexmo`. This should correspond to the URL printed when Attestation Service is started.
+Nexmo requires manual configuration to enable this. If you have not configured a [Nexmo application](https://dashboard.nexmo.com/applications/), go to [Settings](https://dashboard.nexmo.com/settings), and under `Delivery Receipts`, enter the external URL of your Attestation Service appended by `/delivery_status_nexmo` -- for example `http://1.2.3.4:80/delivery_status_nexmo`. This should correspond to the URL printed when Attestation Service is started.
+
+If you have configured [Nexmo applications](https://dashboard.nexmo.com/applications/), open the matching application, click `Edit`, then enter this value as the `Status URL` (you may also be required to enter an `Inbound URL`, though it will be unused).
```
aslawson, line 236: The messages toggle says "Communicate with WhatsApp, Facebook Messenger, MMS and Viber."  The URLs entered here still work for for SMS delivery?

aslawson (approved):

packages/docs/operations-manual/attestation-service.md
timmoreton: yes, it's a bad label name, but appears to be actually the same api.

nambrot:
Not knowing anything about Nexmo applications, can one not use nexmo for multiple instances without nexmo applications? What's the error?

timmoreton:
> Not knowing anything about Nexmo applications, can one not use nexmo for multiple instances without nexmo applications? What's the error?

@nambrot without applications there is only one URL you can enter at which to receive delivery callbacks. Applications allows you to share one account with multiple Applications and use one Application for each Attestation Service front door. (e.g if you have two validators with separate Attestation Service setups)

PR #4922, opened 2020-09-01
Fix group attestation scores

### LINKED ISSUE #4905: [Web] Accurately calculate attestation score
### Expected Behavior

The validator explorer accurately calculates the attestation score of a validator group.

### Current Behavior

The validator explorer logic to calculate the attestation score has a bug -- if the group contains a validator with 0/0 it will average a 0% in the average and bring the group score way down.  This happens frequently as groups will over register validators that are never elected (in preparation for if they get enough votes to elect another validator).  The result is a discouraging score for many of our validators.
I'd think expected behavior would be: sum(all validator successful attestations)/sum(all validator attestation requested)
https://github.com/celo-org/celo-monorepo/blob/master/packages/web/src/dev/ValidatorsList.tsx#L247

ErinHales:
### Description

More accurately calculates the attestation scores of validator groups.

### Tested

Manually tested.

### Related issues

- Fixes #4905


timmoreton (dismissed):
Validators should be excluded if they are not elected. Ones with score zero could need including if they are elected (consider an elected validator that it totally misconfigured and hasn't done any successfully)

ErinHales:
> Validators should be excluded if they are not elected. Ones with score zero could need including if they are elected (consider an elected validator that it totally misconfigured and hasn't done any successfully)

I did as you suggested and excluded all that are not elected.  But I think that logic misrepresents groups like this (showing 0.0% for the group since none were elected, when the score for their validator is actually quite high: 83.3%). Suggestions?
Screen Shot 2020-09-01 at 11 34 03 AM

ErinHales:
Hm, looking at it, I can tell if a validator has never been elected (if the `lastElected` field is null).  That seems to be the best determinant for whether a validator should be excluded from the group score.

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -242,6 +242,7 @@ class ValidatorsList extends React.PureComponent<Props, State> {
                 usd: weiToDecimal(+usd),
                 gold: weiToDecimal(+lockedGold),
                 elected: lastElected >= latestBlock,
+                neverElected: !lastElected,
                 online: lastOnline >= latestBlock,
                 uptime: (+score * 100) / 10 ** 24,
                 attestation:
```
aslawson: While its still unlikely that an elected validator would have 0/0 attestations, I think the best logic here would be to have the individual validator maintain `attestationsFulfilled` and `attestationsRequested` fields and then the validator group would do
SUM_ALL(attestationsFulfilled) / SUM_ALL(attestationsRequested)

This would solve for all cases, and I think it most appropriately calculates what that number represents.

If that is too difficult given the current implementation structure, what you have here should work most of the time -- it just might cause a groups score to drop dramatically when a new validator gets elected until it receives its first attestation request (once we have higher traffic this shouldn't be very long).

WDYT?

ErinHales: Because of the way this data is setup, it would be a pain to calculate it that way.  But if that's the most accurate way to represent it, then I will make it work :)

aslawson (commented):
this looks good to me!  I'll let @aaronmgdr ship as the expert on all web stuff

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -207,7 +207,14 @@ class ValidatorsList extends React.PureComponent<Props, State> {
             .multipliedBy(100)
           const votesPer = new BigNumber(votes).dividedBy(receivableVotes).multipliedBy(100)
           const votesAbsolutePer = receivableVotesPer.multipliedBy(votesPer).dividedBy(100)
+          const totalFulfilled = affiliates.edges.reduce((acc, obj) => {
+            return acc + obj.node.attestationsFulfilled
```
aaronmgdr: is there any chance attestationsFulfilled will be undefined? if so are we good with `NaN`

ErinHales: Good point here.  There are definitely several instances where attestationsFulfilled is `null`, which this handles just fine. I don't think it is ever `undefined`, but returning `NaN` would be an issue.  I'll adjust this to be extra safe.

aaronmgdr (approved):
LGTM

PR #4924, opened 2020-09-01
[Wallet] Add button spinner to Send Amount screen when loading

### LINKED ISSUE #4852: Do not block enabling Review CTA in send/request flow on secure send lookup
### Expected Behavior

The "review" button should be enabled once a valid payment amount is input, independent of secure send completion
once cta is pressed if more time is needed to complete the secure send check, the spinner state on the button should be displayed until moving to the next screen of the send/request flow.

### Current Behavior

review button on a payment is sometimes grayed out for up to ~5s, making it unclear to the user whether changes to the amount input need to be made or the root cause of "review" not being clickable

tarikbellamine:
### Description
Two checks used to happen in order to enable the "Review" button on the Send Amount screen:
1. The amount the user has entered is valid
2. The "secure send check" for the intended recipient is completed

This created a confusing user experience where sometimes 2 would take a while and the user would be uncertain as to why the button was still disabled despite entering in a valid amount.

The PR splits out the two checks:
1 still applies to the "Review" button but if 2 is not done by the time the user hits the button, then the button will have a loading spinner in it until it's done.

### Other changes
No

### Tested
Yes

### Related issues
Closes #4852 

### Backwards compatibility
Yes


packages/mobile/src/send/SendAmount.tsx
```diff
@@ -105,7 +106,9 @@ function SendAmount(props: Props) {
 
   const { t } = useTranslation(Namespaces.sendFlow7)
 
-  const [amount, setAmount] = React.useState('')
+  const [amount, setAmount] = useState('')
+  const [reviewButtonWasPressed, setReviewButtonToPressed] = useState(false)
```
annakaz: nit: would find `reviewButtonPressed` and `setReviewButtonPressed` clearer

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -105,7 +106,9 @@ function SendAmount(props: Props) {
...
+  const [isFetchingAddressMapping, setIsFetchingAddressMapping] = useState(true)
```
annakaz: nit: would find name `fetchingAddressMapping` without `is` clearer

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -202,9 +209,26 @@ function SendAmount(props: Props) {
     }
   }, [props.route, localCurrencyCode, localCurrencyExchangeRate, dollarAmount])
 
+  const onReviewButtonPressed = async () => {
+    setReviewButtonToPressed(true)
+    // Wait until verification status is known to proceed
+    while (recipientVerificationStatus === RecipientVerificationStatus.UNKNOWN) {
```
annakaz, line 223: Are there any cases where the RecipientVerificationStatus stays UNKOWN? If so could add a timeout?

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -202,9 +209,26 @@ function SendAmount(props: Props) {
...
+      await wait(250)
+    }
+
+    setIsFetchingAddressMapping(false)
```
annakaz: Without the context of the PR, I don't know if I would have realized that fetchingAddressMapping is equivalent to getting the recipientVerificationStatus

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -202,9 +209,26 @@ function SendAmount(props: Props) {
     }
   }, [props.route, localCurrencyCode, localCurrencyExchangeRate, dollarAmount])
 
+  const onReviewButtonPressed = async () => {
+    setReviewButtonToPressed(true)
+    // Wait until verification status is known to proceed
+    while (recipientVerificationStatus === RecipientVerificationStatus.UNKNOWN) {
+      await wait(250)
```
annakaz: We use `sleep` from `sleep-promise` elsewhere - any particular reason to use wait here instead?

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -168,8 +171,12 @@ function SendAmount(props: Props) {
   const isAmountValid = parsedLocalAmount.isGreaterThanOrEqualTo(DOLLAR_TRANSACTION_MIN_AMOUNT)
   const isDollarBalanceSufficient = isAmountValid && newAccountBalance.isGreaterThanOrEqualTo(0)
 
-  const reviewBtnDisabled =
-    !isAmountValid || recipientVerificationStatus === RecipientVerificationStatus.UNKNOWN
+  const reviewButtonText =
```
annakaz: nit: was confused by the loading indicator being in a variable called text

packages/mobile/src/send/SendAmount.tsx
tarikbellamine: Gotcha - will use that instead

packages/mobile/src/send/SendAmount.tsx
tarikbellamine: Renamed to `reviewButtonInnerElement`

packages/mobile/src/send/SendAmount.tsx
tarikbellamine: Makes sense. Renamed to `fetchingValidationStatus` to make that more clear

packages/mobile/src/send/SendAmount.tsx
tarikbellamine: I don't think that should ever happen. It is only `UNKNOWN` when the `fetchAddressesAndValidateSaga` is running. When that saga completes or fails, then the value will resolve to either `VERIFIED` or `UNVERIFIED`.

annakaz (approved):
🚀

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -210,14 +217,32 @@ function SendAmount(props: Props) {
     CURRENCY_ENUM.DOLLAR
   )
 
+  const onReviewButtonPressed = async () => {
+    setReviewButtonPressed(true)
+    // Wait until verification status is known to proceed
+    while (recipientVerificationStatus === RecipientVerificationStatus.UNKNOWN) {
+      await sleep(250)
+    }
```
jeanregisser, line 225: I don't understand how this can work.
`recipientVerificationStatus` is a new var declared and set on each render, once you press the button, it's value will never change.
What am I missing?

Lss-Ankit:
Hi @jeanregisser i have verified task using Test flight build v1.1.0 (20) and Android play store build v1.1.0 (1004294308) and found below new issue:

**Test scenarios:**

1. Verify user is able to see the send button on home page 
2. Verify user redirect to Send contacts page after tapping on send button
3. Verify user should able to search any contact 
4. Verify user is able to select any searched contact
5. Verify user should redirect to send page after selecting any contact
6. Verify user is able to see the review button on Send page 
7. Verify Review button should show Disable when user enter invalid amount 
8. verify Review button should show enable when user enter valid amount 
9. Verify user is able to see the spinner loading indicator when tapping on Review button (Issue: [#5019](https://github.com/celo-org/celo-monorepo/issues/5019) )
10. Verify user is already selected then loading should not shown and review button enable immediately after entering amount
11. Verify user should redirect to sending page after tapping on review button 

**New issues:**

-  [#5019](https://github.com/celo-org/celo-monorepo/issues/5019)

**Note:** we are blocked to perform this task on android because app get crash when user tap on Review button 

Thanks !

PR #4925, opened 2020-09-01
Display articles in all user languages on Press page

### LINKED ISSUE #4791: Display secondary Lang Press Articles
on celo.org/press 

display articles if they are in the language of the website OR a language the user has as acceptable in their browser. 

currently only do the first.

isabellewei:
### Description

Currently the Press page only displays articles in the primary language that is set, now it should display articles in all languages that the user uses.

### Other changes

changed the enum in `languages.ts`, but I checked and it isn't being used anywhere else so should be fine.

### Tested

I set my Chrome settings to also include Chinese, Chinese articles appeared on the press page

### Related issues

- Fixes #4791


packages/web/src/press/PressPage.tsx
```diff
@@ -20,29 +20,40 @@ export interface PressArticleFields {
 
 interface Props {
   press: PressArticleFields[]
+  languages: string
 }
 
 class PressPage extends React.PureComponent<I18nProps & Props> {
   static async getInitialProps(context) {
-    let press = []
```
aaronmgdr, line 27: can we leave this as defaulting to an array? if something messes up at least then we get an empty array to call array methods on and not `undefined doesnt have function` errors

aaronmgdr:
good work. just a few refinements

packages/web/src/press/PressPage.tsx
```diff
@@ -20,29 +20,40 @@ export interface PressArticleFields {
...
+    const langList = new Set(
+      languages
+        .toLowerCase()
+        .split(',')
+        .map((s) => s.substr(0, 2))
+    )
 
     const formated = press
-      .filter((article) => article.language === Languages[i18n.language])
+      .filter((article) => langList.has(Languages[article.language]))
```
aaronmgdr: Id like to do an `||` here with the old way of using lang from i18n just in case. as they will be browsing in that language anyway.

packages/web/src/press/PressPage.tsx
```diff
@@ -20,29 +20,43 @@ export interface PressArticleFields {
...
     } catch {
-      return { press }
+      return { press, languages }
     }
   }
   render() {
-    const { t, press, i18n } = this.props
+    const { t, press, languages, i18n } = this.props
+    const langList = new Set(
+      languages
```
aaronmgdr, line 49: nit: consider moving the string parsing into the getInitialProps so it only needs to run once rather than on each render. 

keep the creation of the Set here though.

aaronmgdr (approved):

PR #4929, opened 2020-09-01
[L09] Misleading comments and typographical errors

yorhodes:
### Description

### Other changes

### Tested

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/605

### Backwards compatibility

### DIFF
--- a/packages/protocol/contracts/common/TransferWhitelist.sol
+++ b/packages/protocol/contracts/common/TransferWhitelist.sol
@@ -34,7 +34,7 @@ contract TransferWhitelist is Ownable, UsingRegistry {
 
   /**
    * @notice Remove an address from the whitelist.
-   * @param removedAddress The address to add.
+   * @param removedAddress The address to remove.
    * @param index Index of address in the whitelist.
    */
   function removeAddress(address removedAddress, uint256 index) external onlyOwner {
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/governance/Validators.sol
```diff
@@ -918,7 +918,7 @@ contract Validators is
     uint256 balance = getLockedGold().getAccountTotalLockedGold(account);
     // Add a bit of "wiggle room" to accommodate the fact that vote activation can result in a 1
     // wei rounding error.
-    return balance.add(10) >= getAccountLockedGoldRequirement(account);
+    return balance.add(1) >= getAccountLockedGoldRequirement(account);
```
yorhodes: @asaj any rationale for 10 as opposed to 1?
does it have to do with multiple vote activations?
this feels like an edge case that we don't account for elsewhere

packages/protocol/contracts/common/TransferWhitelist.sol
```diff
@@ -34,7 +34,7 @@ contract TransferWhitelist is Ownable, UsingRegistry {
 
   /**
    * @notice Remove an address from the whitelist.
-   * @param removedAddress The address to add.
```
aaronmgdr, line 37: ‼️

packages/protocol/contracts/governance/Validators.sol
m-chrzan: My guess would be to for sure overshoot whatever the rounding error can be? But yeah, would love @asaj's input before this gets merged.

asaj: Yeah it was just for an additional margin of safety. I would argue for keeping it given I don't see much of a benefit of reducing it to 1, but I don't feel terribly strongly

PR #4932, opened 2020-09-01
[Wallet] Build number changes from 1.0.1 release

annakaz:
### Description

Bump build number to 1.0.1

### Other changes

Update fastlane script to skip uploading assets, as we want to keep what was added manually by Nitya to the App and Play Stores


gnardini (approved):

packages/mobile/android/gradle.properties
```diff
@@ -22,8 +22,8 @@ CELO_RELEASE_STORE_FILE=celo-release-key.keystore
 CELO_RELEASE_KEY_ALIAS=celo-key-alias
 
 # Setting this manually based on version number until we have this deploying via Cloud Build
-# Example: v1.5.1 deployment number 1 = 100500101 (1 005 001 01)
-VERSION_CODE=100800202
+# Example: v1.5.1 deployment number 1 = 1005001001 (1 005 001 001)
+VERSION_CODE=1000100001
```
jmrossy, line 26: Should this be 1000001001 According to the new scheme here?

packages/mobile/ios/celo.xcodeproj/project.pbxproj
```diff
@@ -847,7 +847,7 @@
 				CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = celo/celo.entitlements;
-				CURRENT_PROJECT_VERSION = 34;
+				CURRENT_PROJECT_VERSION = 17;
```
jmrossy, line 850: Was cutting this in half intentional?

packages/mobile/ios/celo/Info.plist
```diff
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.8.2</string>
+	<string>$(MARKETING_VERSION)</string>
```
jmrossy, line 20: Just throwing it out there. It's not too late to bump everything up to 2.0.0 and save us the complexity of a separate marketing version!

jmrossy (dismissed):
See comments above

packages/mobile/ios/celo.xcodeproj/project.pbxproj
annakaz: No haha this went from 1 -> 2 -> ... 17, but none of those build number changes from me or @cmcewen got merged into master

packages/mobile/android/gradle.properties
annakaz: This seems like the right number of zeroes to me: 1 000 100 001 - is this different from how you interpreted the scheme above?

packages/mobile/ios/celo/Info.plist
annakaz: We love our 1.0.0 ... if we went with 2.0.0 then it would be too late to decide we wanted to go back to 1.x.x though 😄

annakaz:
Lmk if the comments above make sense @jmrossy !

packages/mobile/android/gradle.properties
annakaz: Actually looking at this again - yes it should be but since the build has been deployed we can't bump it down. Fortunately the next build will be `1 100 000 001` and this will be resolved

packages/mobile/ios/celo/Info.plist
jeanregisser: wondering is this kept when we use the script to bump the version (`pre-deploy.sh`)?

jeanregisser (approved):
LGTM 👍

PR #4951, opened 2020-09-02
[Wallet] Ability to send events from native ios to react

etuleu:
### Description

Similar to what we did on Android we want to send an event as soon as the app started. On iOS it is slightly trickier. Still doing a bit more testing and trying to figure out the best place to fire the event from.

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


packages/mobile/src/app/App.tsx
```diff
@@ -45,71 +46,30 @@ BigNumber.config({
...
-  eventName: string,
-  callback: (reactInitTime: string) => void
-) => EmitterSubscription | null = (
-  eventName: string,
-  callback: (reactInitTime: string) => void
-): EmitterSubscription | null => {
+function getEventEmitter() {
   if (Platform.OS === 'android') {
-    return new NativeEventEmitter().addListener(eventName, callback)
+    return new NativeEventEmitter()
```
tarikbellamine: nit: More of a stylistic thing but may make sense to replace the `if...else` with a ternary to keep it clean

packages/mobile/src/app/App.tsx
```diff
@@ -45,71 +46,30 @@ BigNumber.config({
...
-      deviceHeight: height,
-      deviceWidth: width,
-      reactLoadDuration,
-      appLoadDuration,
-    })
 
-    if (this.appStartListener) {
-      this.appStartListener.remove()
-    }
+    const appLoadedTime = Date.now()
```
tarikbellamine: Instead of having this as an input for `getAppStartTimeFromNative`, can we instead define it inside the function? I think it would be helpful to contain as much of the variables and logic as possible within that function

packages/mobile/src/app/App.tsx
```diff
@@ -118,6 +78,29 @@ export class App extends React.Component {
     Linking.removeEventListener('url', this.handleOpenURL)
   }
 
+  getAppStartTimeFromNative(appLoadedTime: number) {
```
tarikbellamine: nit: Can we rename to `logAppLoadDuration`? I think `get` implies that we expect the function to return something whereas `log` makes it clearer that we are only logging the results

tarikbellamine (approved):
Looks good to me! Left some smaller stylistic suggestions and comments.

Please double check that this works as expected on both Android and iOS devices. Given we don't have many native code experts to vet this, the best thing we can do is thoroughly test before pushing this out into production.

packages/mobile/src/app/App.tsx
etuleu: I found a Platform.select function that seems to be pretty useful and makes it look cleaner.

etuleu:
> Looks good to me! Left some smaller stylistic suggestions and comments.
> 
> Please double check that this works as expected on both Android and iOS devices. Given we don't have many native code experts to vet this, the best thing we can do is thoroughly test before pushing this out into production.

Thanks for the review! I made the suggested changes and am doing a bunch more testing on emulator, simulator and real devices. I am actually holding on pushing this because in parallel I am also implementing the log sending through the bridge init and then we can compare and decide which one we like more.

etuleu:
> > Looks good to me! Left some smaller stylistic suggestions and comments.
> > Please double check that this works as expected on both Android and iOS devices. Given we don't have many native code experts to vet this, the best thing we can do is thoroughly test before pushing this out into production.
> 
> Thanks for the review! I made the suggested changes and am doing a bunch more testing on emulator, simulator and real devices. I am actually holding on pushing this because in parallel I am also implementing the log sending through the bridge init and then we can compare and decide which one we like more.

Actually it looks like there is a bug with the initial props passing strategy: https://github.com/facebook/react-native/issues/20115

So probably best to stick with this working code for now. We can monitor the react-native issue and see if it gets fixed in the future.

etuleu:
> > > Looks good to me! Left some smaller stylistic suggestions and comments.
> > > Please double check that this works as expected on both Android and iOS devices. Given we don't have many native code experts to vet this, the best thing we can do is thoroughly test before pushing this out into production.
> > 
> > 
> > Thanks for the review! I made the suggested changes and am doing a bunch more testing on emulator, simulator and real devices. I am actually holding on pushing this because in parallel I am also implementing the log sending through the bridge init and then we can compare and decide which one we like more.
> 
> Actually it looks like there is a bug with the initial props passing strategy: [facebook/react-native#20115](https://github.com/facebook/react-native/issues/20115)
> 
> So probably best to stick with this working code for now. We can monitor the react-native issue and see if it gets fixed in the future.

Did this get auto merged? I don't think I pushed the merge button. I need to do some more testing on it. Not sure if I should revert it now or leave it and test further. What do you think? @jeanregisser @tarikbellamine

PR #4952, opened 2020-09-02
Support for running testnets with multiple versions of geth

### LINKED ISSUE #1148: Nightly Build Test - Mixed Version Test
Create an **automated** test that runs a testnet with multiple version of celo-blockchain 

- [ ]  Test Runs every night using last commit on master
- [ ]  Be able to start test manually

This is needed prior a release to know if a new release is compatible with mainnet versions. Also to know if current master is compatible

### Implementation Ideas

We need to modify `testnet` helm chart so that we can use different versions. 

* Use `RID` as a selector of which version to run
* Two ideas for choosing a contiainer:
    * Have 2 or more validator stateful sets (seems hard & complex)
    * Build a docker container that contains geth binaries for more than one version, and then choose the version as part of the docker run command

### Assertions

* Testnet doesn't stall
* Avg Block Time is 5s (or whatever we configure)
* All validators are mining
* All validator have 100% uptime

mrsmkl:
### Description

I'll probably add the actual script to run everything to the celo-blockchain repo.

### Related issues

- Fixes https://github.com/celo-org/celo-blockchain/issues/1148

### DIFF
--- a/.env.testold
+++ b/.env.testold
@@ -1,0 +1,120 @@
+# These are the default values that should apply to most environments, especially dev ones
+ENV_TYPE="development"
+#GETH_DEBUG=true
+GETH_VERBOSITY=2
+VM_BASED=false
+KUBERNETES_CLUSTER_NAME="celo-networks-dev"
+KUBERNETES_CLUSTER_ZONE="us-west1-a"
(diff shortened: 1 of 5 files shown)


jcortejoso:
Closing in favour of changes from https://github.com/celo-org/infrastructure/

PR #4955, opened 2020-09-02
Changed to google analytics

### LINKED ISSUE #4950: Web analytics update
### Expected Behavior

We send directly to GA as opposed to using Segment to save quota

### Current Behavior

Send via segment

ErinHales:
### Description

Change Segment to Google Analytics

### Tested

Manually tested and saw events come through

### Related issues

- Fixes #4950


packages/web/src/analytics/analytics.ts
```diff
@@ -2,41 +2,30 @@ import Cookies from 'js-cookie'
...
 }
 
 export async function initializeAnalytics() {
-  if (!isBrowser() || !(await canTrack())) {
-    return {
-      track: noTrack,
-      page: noTrack,
-    }
-  }
+  ReactGA = await import('react-ga').then((mod) => mod.default)
```
aaronmgdr: can you add canTrack call here and only load ReactGA if it is true?

packages/web/src/analytics/analytics.ts
```diff
@@ -2,41 +2,30 @@ import Cookies from 'js-cookie'
...
-    segmentPromise = import('load-segment').then((mod) => mod.default)
-  }
-  if (!analytics) {
-    const Segment = await segmentPromise
-    const { publicRuntimeConfig } = getConfig()
-    analytics = Segment({ key: publicRuntimeConfig.__SEGMENT_KEY__ })
-  }
-  return analytics
+  ReactGA.initialize(publicRuntimeConfig.GA_KEY)
+  ReactGA.pageview('initial pageview')
```
aaronmgdr: I think we can remove the initial page view as it was just for testing right?

packages/web/src/analytics/analytics.ts
aaronmgdr: or make sure where this is called there is a canTrack check first?

packages/web/src/analytics/analytics.ts
ErinHales: I was originally going to remove it, but then I realized we only track page views on route change.  Thought it would be useful to have analytics for the initial visit.
```
    this.props.router.events.on('routeChangeComplete', async () => {
      await analytics.page()
    })
```

ErinHales: Although, I should add a `canTrack` check to that as well...

aaronmgdr: Oh all evidence ive seen shows that page loads are being tracked even on initial load. 

and if that is not the case id like like to not have the call to track happen inside of the initialize function as it should be idempotent so that we can call it in track function to ensure analytics are available.

ErinHales: I could be wrong, but I'm not seeing any analytics pop up on page load other than this "initial pageview".  Where would you like to have the call happen then?

aaronmgdr: ok if we need to add, I think the best spot would be in side of `_app.tsx` in the componentDidMount function. right after it calls initialize.

packages/web/src/analytics/analytics.ts
```diff
@@ -2,41 +2,31 @@ import Cookies from 'js-cookie'
 import getConfig from 'next/config'
 import { isBrowser } from 'src/utils/utils'
 
-let analytics: {
-  track: (key: string, properties?: object, options?: object) => void
-  page: () => void
-}
+const { publicRuntimeConfig } = getConfig()
```
aaronmgdr, line 5: hmm whenever ive tried to call this top level (not in a function) its ended up crashing. 

Maybe the recent upgrade fixed that. 🤔

ErinHales: Hm, I haven't had any issues with it... Is it crashing currently? Or it just has in the past?

aaronmgdr: just run it for myself seem to be working.  know this now works is a great find.

aaronmgdr (approved):

PR #4961, opened 2020-09-03
Add script to fund e2e tests account when balance is low

gnardini:
### Description

Yesterday the e2e tests started failing because the e2e test account run out of funds.

Wrote a quick script that checks its balance and adds to it if it's running low.

For now, it's not running automatically anywhere but my computer. In the future, we could add this as a CircleCI cron job.


etuleu (approved):
looks good to me

I don't think my approval is enough though so maybe you can ping one of the people listed as well?

jeanregisser (approved):
Thanks!

PR #4973, opened 2020-09-03
[Wallet] Payment requests must use lowercase addresses

### LINKED ISSUE #4886: Payment requests made via address are not received
### Expected Behavior
Payment requests work.

### Current Behavior
When making a payment request with `address` as the identifier (as opposed to phone number), requests don't appear to be received by the recipient. Appears to only happen when the requester is not verified (though unsure of this). Possibly related to #1533. 

Double checked that the address included in the request matches the recipients address.

gnardini:
### Description

When receiving a payment request, we look on Firebase for requests made to our address using lowercase letters. If the request was made using uppercase letters, we ignore that request. With this change, we are making sure requests are always made using lowercase addresses.

### Tested

Sent a payment request without this change, it didn't arrive. Made the change, made a new request, it arrived. 
Also did the same thing but manually changing from upper to lowercase on Firebase and the requests started showing up after doing so.

### Related issues

- Fixes #4886

### Backwards compatibility

Requests that failed in the past won't start working, new payment requests will need to be made. I think that's for the better though.

### DIFF
--- a/packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
+++ b/packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
@@ -117,7 +117,7 @@ class PaymentRequestConfirmation extends React.Component<Props> {
       timestamp: new Date(),
       requesterAddress: address,
       requesterE164Number: this.props.e164PhoneNumber ?? undefined,
-      requesteeAddress,
+      requesteeAddress: requesteeAddress.toLowerCase(),
       currency: currencyToShortMap[CURRENCY_ENUM.DOLLAR],
       status: PaymentRequestStatus.REQUESTED,
       notified: false,
--- a/packages/mobile/src/recipients/RecipientItem.tsx
+++ b/packages/mobile/src/recipients/RecipientItem.tsx
@@ -5,7 +5,7 @@ import fontStyles from '@celo/react-components/styles/fonts.v2'
 import variables from '@celo/react-components/styles/variables'
 import * as React from 'react'
 import { StyleSheet, Text, View } from 'react-native'
-import { getRecipientThumbnail, Recipient, RecipientKind } from 'src/recipients/recipient'
+import { getRecipientThumbnail, Recipient } from 'src/recipients/recipient'
 
 interface Props {
   recipient: Recipient
@@ -15,9 +15,7 @@ interface Props {
 class RecipientItem extends React.PureComponent<Props> {
   onPress = () => {
     const recipient = this.props.recipient
-    if (recipient.kind === RecipientKind.Address) {
-      recipient.address = recipient.address.toLowerCase()
-    }
+    recipient.address = recipient.address?.toLowerCase()
     this.props.onSelectRecipient(recipient)
   }
 
--- a/packages/mobile/src/recipients/RecipientItem.tsx
+++ b/packages/mobile/src/recipients/RecipientItem.tsx
@@ -5,7 +5,7 @@ import fontStyles from '@celo/react-components/styles/fonts.v2'
 import variables from '@celo/react-components/styles/variables'
 import * as React from 'react'
 import { StyleSheet, Text, View } from 'react-native'
-import { getRecipientThumbnail, Recipient } from 'src/recipients/recipient'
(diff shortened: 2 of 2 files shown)


packages/mobile/src/recipients/RecipientItem.tsx
```diff
@@ -14,7 +14,11 @@ interface Props {
 
 class RecipientItem extends React.PureComponent<Props> {
   onPress = () => {
-    this.props.onSelectRecipient(this.props.recipient)
+    const recipient = this.props.recipient
```
gnardini, line 17: With the change on the other file it suffices for the purpose of this bug, but I thought having this here might avoid similar problems in the future.

tarikbellamine: This would be completely redundant with the other `.toLowercase` operation right?

Instead of having two checks like this, can we instead have one happen on the import/filtering of payment requests? I think that gives us the added benefit of helping users "recover" old payment requests that were getting incorrectly filtered out before

gnardini: For the purpose of payment requests it's redundant, yes, but this is used for sending as well and might be used for other things, I just thought that making sure that addresses are lowercase is not a bad thing and could possibly avoid issues in the future.

I thought of recovering old payment requests, but if they didn't show up before I think it might be a bit weird for users for requests to suddenly pop up out of nowhere when it didn't before, especially if they sent several ones when it didn't work. The number of users is quite low for now anyways, so I guess it doesn't matter much either way.

At the same time, I'm not sure how possible it is to do it on the other side. Right now it's doing a simple `.equalTo(address)`. Firebase doesn't have a lot of flexibility in the operations it allows and this would require for them to accept a case insensitive equals, which doesn't seem supported from a quick look.

tarikbellamine: Fair points! In that case, should we lowercase all addresses instead of only when `recipient.kind === RecipientKind.Address`?

gnardini: Yup, for some reason thought that the address field is only populated when the kind is address, but better to map it always :)

tarikbellamine (approved):
Nice work!

packages/mobile/src/recipients/RecipientItem.tsx
```diff
@@ -14,7 +14,9 @@ interface Props {
...
+    recipient.address = recipient.address?.toLowerCase()
+    this.props.onSelectRecipient(recipient)
```
jeanregisser, line 19: This is dangerous! We're mutating an existing object. It's very likely this comes from the redux state and we're directly mutating it, bypassing the immutability assumptions of the redux state.

gnardini: Addressed on https://github.com/celo-org/celo-monorepo/pull/4979

packages/mobile/src/paymentRequest/PaymentRequestConfirmation.tsx
```diff
@@ -117,7 +117,7 @@ class PaymentRequestConfirmation extends React.Component<Props> {
       timestamp: new Date(),
       requesterAddress: address,
```
jmrossy, line 118: Should this address be normalized too?

gnardini: It's already doing that on the selector:
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/web3/selectors.ts#L3
```
export const currentAccountSelector = (state: RootState) =>
  (state.web3.account && state.web3.account.toLowerCase()) || null
```

Lss-Ankit:
@gnardini @jeanregisser Verified issue using Android internal build v1.1.0 (1004294309)& iOS test flight build v1.1.0(21) and found that user B is able to get payment request when user A sent the payment request. 

- Verified user B is getting payment request when user user A sent payment request using user B's phone number. 
- Verified user B is getting payment request when user A sent payment request using user B's address / account No. 
- Verified user B is getting payment request (Verified User B) when non verified user A sent payment request using user B's address 
- Verified User B is getting payment request (Non verified user B) when verified user A sent payment request using user B's address and phone number.

PR #4976, opened 2020-09-04
Adds ODIS identifier query to celocli

alecps:
### Description

Adds support for querying a phone number identifier and pepper from ODIS via celocli. 

### Other changes

Fixes random flakey test that just needed a longer timeout https://github.com/celo-org/celo-monorepo/issues/4988

### Tested

Locally on alfajores. Ensured that queries with an unregistered/empty account or with an invalid phone number output appropriate error messages. 

### Related issues

NA

### Backwards compatibility

Yes


packages/cli/src/commands/identity/identifier.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    cli.action.start('Querying ODIS for identifier')
+
+    const authSigner: AuthSigner = {
+      authenticationMethod: OdisUtils.Query.AuthenticationMethod.WALLET_KEY,
+      contractKit: this.kit,
+    }
+
+    let serviceContext: ServiceContext
+    switch (env) {
```
alecps: I considered adding these as constants in OdisUtils. Thoughts?

codyborn: Yeah, I like the idea! You can either define it in this class or in OdisUtils.

packages/cli/src/commands/identity/identifier.ts
```diff
@@ -0,0 +1,84 @@
...
+import { AuthSigner, ServiceContext } from '@celo/contractkit/lib/identity/odis/query'
+import { flags as oFlags } from '@oclif/command'
+import { cli } from 'cli-ux'
+import { BaseCommand } from '../../base'
+import { newCheckBuilder } from '../../utils/checks'
+import { printValueMap } from '../../utils/cli'
+import { Flags } from '../../utils/command'
+export default class IdentifierQuery extends BaseCommand {
+  static description =
+    'Queries ODIS for the on-chain identifier and pepper corresponding to a given phone number.'
```
codyborn, line 12: We should also add a description for the identity commands for when someone just runs `celocli`. Looks like it was missing:
https://github.com/celo-org/celo-monorepo/blob/alecps/odisCli/packages/cli/package.json#L106

codyborn (approved):
Tried it out, works like a charm! 👏

packages/cli/src/commands/identity/identifier.ts
jmrossy: +1, it would prevent the duplication in mobile/src/network-config.ts too

packages/cli/src/commands/identity/identifier.ts
```diff
@@ -0,0 +1,84 @@
...
+      required: true,
+      description: 'The address from which to perform the query',
+    }),
+    phoneNumber: Flags.phoneNumber({
+      required: true,
+      description: 'The phone number for which to query the identifier (w/ country code)',
+    }),
+    env: oFlags.string({
+      required: false,
+      description: 'mainnet (default), alfajores, or alfajores_staging',
```
jmrossy: should be alfajoresstaging one word

packages/cli/src/commands/identity/identifier.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  static flags = {
+    ...BaseCommand.flags,
+    from: Flags.address({
+      required: true,
+      description: 'The address from which to perform the query',
+    }),
+    phoneNumber: Flags.phoneNumber({
+      required: true,
+      description: 'The phone number for which to query the identifier (w/ country code)',
```
jmrossy: You should mention the number must be in e164 format

packages/cli/src/commands/identity/identifier.ts
```diff
@@ -0,0 +1,84 @@
+import { OdisUtils } from '@celo/contractkit'
+import { AuthSigner, ServiceContext } from '@celo/contractkit/lib/identity/odis/query'
+import { flags as oFlags } from '@oclif/command'
+import { cli } from 'cli-ux'
+import { BaseCommand } from '../../base'
+import { newCheckBuilder } from '../../utils/checks'
+import { printValueMap } from '../../utils/cli'
+import { Flags } from '../../utils/command'
+export default class IdentifierQuery extends BaseCommand {
```
jmrossy, line 10: nit spacing above

jmrossy (approved):
Looks good just minor things!

packages/cli/src/commands/identity/identifier.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    let serviceContext: ServiceContext
+    switch (env) {
+      case 'alfajores':
+        serviceContext = {
+          odisUrl: 'https://us-central1-celo-phone-number-privacy.cloudfunctions.net',
+          odisPubKey:
+            'kPoRxWdEdZ/Nd3uQnp3FJFs54zuiS+ksqvOm9x8vY6KHPG8jrfqysvIRU0wtqYsBKA7SoAsICMBv8C/Fb2ZpDOqhSqvr/sZbZoHmQfvbqrzbtDIPvUIrHgRS0ydJCMsA',
+        }
+      case 'staging':
```
jmrossy: should be alfajoresstaging

PR #4977, opened 2020-09-04
[Wallet] Fix bug to allow users to skip invite redemption after canceling

tarikbellamine:
### Description
If a user presses "Cancel" while an invite redemption is in progress, it should reset the invite redemption state variables. Currently, `isRedemeeingInvite` is never flipped back to `false`. Additionally, when a invite redemption fails it often does so after an account has been created but before the funds have been transferred from the temp invite account to the user's account. The combination of `account` and `isRedeemingInvite` being truthy results in the user being stuck on the `EnterInviteCode` screen indefinitely.

### Other changes
- When a user copy/pastes the full invite text message, the input form used to get populated with the private key that was extracted from the parsed invite code. This was confusing users so instead we are now populating the input form with the parsed invite code instead of the private key
- Added a cancel action, which gets triggered when the user presses "cancel" on the `EnterInviteCode` screen,  to the send transaction within `moveAllFundsFromAccount` 
- Changed the condition by which `addTempAccountToWallet` detects whether an account has already been created. I think this may have gotten overlooked during the recent geth changes

### Tested
Added tests and ran through various conditions on a physical device.

### Related issues
NA

### Backwards compatibility
Yes


packages/mobile/src/invite/actions.ts
```diff
@@ -11,6 +11,7 @@ export enum Actions {
   REDEEM_INVITE = 'INVITE/REDEEM_INVITE',
   REDEEM_INVITE_SUCCESS = 'INVITE/REDEEM_INVITE_SUCCESS',
   REDEEM_INVITE_FAILURE = 'INVITE/REDEEM_INVITE_FAILURE',
+  CANCEL_REDEEM_INVITE = 'CANCEL_REDEEM_INVITE',
```
i1skn: Shall we use "INVITE/CANCEL_REDEEM_INVITE"?

packages/mobile/src/invite/EnterInviteCode.tsx
```diff
@@ -192,7 +186,7 @@ export class EnterInviteCode extends React.Component<Props, State> {
                       <TextButton
                         style={styles.bottomButton}
                         onPress={this.onPressSkip}
-                        disabled={isRedeemingInvite || !!account}
+                        disabled={isRedeemingInvite}
```
tarikbellamine: @jeanregisser is there a reason why were we disabling on `account` truthy here? Looks to me that the `skipInvite` saga can handle the case of when an account has already been created.

packages/mobile/src/onboarding/UseBackToWelcomeScreen.tsx
```diff
@@ -59,6 +60,7 @@ export function useBackToWelcomeScreen({ backAnalyticsEvents }: Props) {
           ValoraAnalytics.track(analyticsEvent)
         })
         dispatch(cancelCreateOrRestoreAccount())
+        dispatch(cancelRedeemInvite())
```
i1skn: What is the difference between this event and the once above it?

tarikbellamine: The new action switches `isRedeemingInvite` to `false`. When a user was hitting cancel, this could be remaining as `true`, which would prevent the user from being able to skip to the next screen. These actions route to separate reducers so didn't want to mix them.

packages/mobile/src/invite/EnterInviteCode.tsx
jeanregisser: I think the `!!account` check was there to prevent incorrect state to proceed.
If for some reason you end up on this screen (though it looks impossible today) and you already have an account, you shouldn't be able to skip to the next screen. As it might cause other unwanted state changes.

But you're right, `skipInvite` is able to handle that case and uses the existing account.
So your change looks fine to me.

@jmrossy what's your take on this? Just want to be sure we don't miss something here.

packages/mobile/src/onboarding/UseBackToWelcomeScreen.tsx
jeanregisser: In general, using consecutive `dispatch` should be avoided for 2 reasons:
- I think it's important to have meaningful actions and it's good to have such actions cause different parts of the state to change and cause multiple side effects.
- We don't want incorrect state to be possible. Dispatching 2 actions consecutively makes it possible to end up with incorrect state at least temporarily. And definitely if there's a crash/app being killed between the two, though quite unlikely of course.

Conceptually we're still creating the account here. And it looks normal to have the `cancelCreateOrRestoreAccount` cancel/reset any state linked to it.

I propose we remove the newly introduced `cancelRedeemInvite` action and react to `cancelCreateOrRestoreAccount` in the redeem invite saga.

packages/mobile/src/invite/saga.ts
```diff
@@ -286,25 +296,32 @@ export function* redeemInviteSaga({ inviteCode }: RedeemInviteAction) {
...
     // Note: We are ok with this succeeding or failing silently in the background,
     // user will have another chance to register DEK when sending their first tx
     yield spawn(registerAccountDek, result.newAccount)
-  } else if (result.success === false) {
+  } else if (result?.success === false) {
     Logger.debug(TAG, 'Redeem Invite failed')
     yield put(redeemInviteFailure())
+  } else if (cancel) {
+    ValoraAnalytics.track(OnboardingEvents.invite_redeem_cancel)
+    yield put(redeemInviteFailure())
```
jeanregisser: Following what I said above, I think we shouldn't dispatch a failure here, it's not really a failure, it's just the user who decided to cancel. The reducer should update the state for the `CANCEL_CREATE_OR_RESTORE_ACCOUNT` action.

jeanregisser (changes requested):
Thanks for looking into this! 👍 
I propose some small changes.

packages/mobile/src/onboarding/UseBackToWelcomeScreen.tsx
tarikbellamine: Good suggestion. Will do.

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -504,7 +503,7 @@ export function MainStackScreen() {
     } else if (!e164Number || !acceptedTerms || pincodeType === PincodeType.Unset) {
       // User didn't go far enough in onboarding, start again from education
       initialRoute = Screens.OnboardingEducationScreen
-    } else if (!redeemComplete && !account) {
+    } else if (!redeemComplete) {
```
tarikbellamine, line 506: @jeanregisser similar to what we discussed earlier, I don't think this `account` condition is necessary. I am removing it because if there is an invite code redemption failure, it often fails after an account has been created but before the invite funds have been transferred. So if a user closes the app after an invite failure, they are skipped ahead to verification instead of staying on the screen and having the opportunity to try again to redeem

gnardini (approved):
I'm not very familiar with the invite flow and what you guys discussed this morning, but the code itself LGTM :)

packages/mobile/src/invite/saga.ts
```diff
@@ -377,17 +394,17 @@ export function* skipInvite() {
...
     const wallet: UnlockableWallet = yield call(getWallet)
-    const account = privateKeyToAddress(inviteCode)
+    const account = privateKeyToAddress(tempAccountPrivateKey)
     const password: string = yield call(getPasswordSaga, account, false, true)
-    const tempAccount = yield call([wallet, wallet.addAccount], inviteCode, password)
+    const tempAccount = yield call([wallet, wallet.addAccount], tempAccountPrivateKey, password)
     Logger.debug(TAG + '@addTempAccountToWallet', 'Account added', tempAccount)
   } catch (e) {
-    if (e.toString().includes('account already exists')) {
+    if (e.message === ErrorMessages.GETH_ACCOUNT_ALREADY_EXISTS) {
```
tarikbellamine, line 407: I think the error message may have changed with recent changes to geth. What I've done seems to produce the desired outcome

packages/mobile/src/invite/saga.ts
```diff
@@ -442,7 +459,14 @@ export function* moveAllFundsFromAccount(
...
   const context = newTransactionContext(TAG, 'Transfer from temp wallet')
   // Temporarily hardcoding gas estimate to save time on estimation
-  yield call(sendTransaction, tx.txo, account, context, SEND_TOKEN_GAS_ESTIMATE)
+  yield call(
+    sendTransaction,
+    tx.txo,
+    account,
+    context,
+    SEND_TOKEN_GAS_ESTIMATE,
+    AccountActions.CANCEL_CREATE_OR_RESTORE_ACCOUNT
```
tarikbellamine, line 468: Adding a "cancel" action for the transfer transaction that gets triggered when the user pressed the top left "Cancel" button during invite redemption

jeanregisser (approved):

Lss-Ankit:
HI @tarikbellamine and @jeanregisser i have verified task using Android internal build v1.1.0  (1004294309) and Test flight build v1.1.0 (21). 
Device used: Google Pixel XL (8.1) , iPhone 7+ (13.3.1) 

Test scenarios:
- User is able to redeem code again if user tap on cancel button before receiving funds in app notification. 
- Redeem invite code fails when user cancel after receiving funds in app notification. 

**Please let me know if i need to verify any thing else.**
 
**Need Clarification:**
- User is able to see alpha numeric number after pasting invite code. 
Attachment: 
Invite_Code

PR #4979, opened 2020-09-04
[Wallet] Fix payment request comments decryption

### LINKED ISSUE #4851: “Comment Unavailable” text is shown on sent request section when user send request with adding description
**Frequency:** 100%
**Repro App version:**  Valora Mainnet build v1.0.0 (14), Test flight build v1.0.0 (15) 
**Does not Repro on  App version:** Test flight Valora v1.0.0(6) 
**Repro on device:** iPhone 6(12.4.7), Samsung Galaxy A5(7.0)

Pre-conditions: Make sure user has verified phone number.

**Repro Steps:**
1. Launch the Android app and sign in with a test account. 
2. Launch the iOS app and sign up. 
3. Tap on Request button
4. Select any contact
5. Enter valid amount
6. Tap on Review button
7. Tap on send button 
8. Go to home page & observe on sent notification

**Current Behavior:** Comment unavailable text is shown when user sent request adding with description 
**Expected Behavior:**  Entered description should show on sent request section
**Investigation:** Issue does not repro if user has not verified number. 
**Impact:** User not able to see entered description 
**Attachment:** [CommentUnavailable.png](https://drive.google.com/file/d/18YnB1olK5Tq4gF6itvyLcS-o59IYvAQi/view?usp=sharing)

gnardini:
### Description

Outgoing payment requests had their comments showing as "Comment Unavailable". Added a flag so that outgoing requests can be decrypted correctly.

### Other changes

Addressed a comment that @jeanregisser made post-merge here: https://github.com/celo-org/celo-monorepo/pull/4973 regarding state immutability. Since they are both related to payment requests I included it here, I can separate it into a different PR if needed.

### Tested

On two simulators, one verified and one not, sending requests both ways and seeing that the comments are now visible.

### Related issues

- Fixes #4851

### Backwards compatibility

Yes


packages/mobile/src/recipients/RecipientItem.tsx
```diff
@@ -14,8 +14,13 @@ interface Props {
 
 class RecipientItem extends React.PureComponent<Props> {
   onPress = () => {
-    const recipient = this.props.recipient
-    recipient.address = recipient.address?.toLowerCase()
+    let recipient = this.props.recipient
+    if (recipient.address) {
```
jmrossy: Instead of modifying the the address casing here in the view layer (which shouldn't concern itself with data details liek this), how about we move this normalization into the sagas?

jmrossy (commented):
Looks correct to me, nice catch!

packages/mobile/src/recipients/RecipientItem.tsx
gnardini: That's a good point

gnardini: I moved this to call `toLowerCase` on the address search string, which is the only one that's using an address directly here.

jeanregisser (approved):
Awesome!! 🎖️ 👍

jmrossy (approved):

Lss-Ankit:
@gnardini @jeanregisser Verified with latest Android internal build v1.1.0 (1004294309)& iOS test flight build v1.1.0(21) and found that issue is fixed as user is able to see comment on payment request section. 
- Also i have verified with one verified user and one unverified user and both are able to see comment on payment request.

PR #4980, opened 2020-09-04
[M02] Data lengths alteration could lead to malicious transaction exe…

ewilz:
…cutions

### Description

oz audit fix:
MetaTransactionWallet.executeTransactions will include data length prepended to actual data in `data` parameter.

### Other changes

N/A

### Tested

Changed parameters in tests to account for data parameter update. And tested revert condition for update.

### Related issues

https://github.com/celo-org/celo-labs/issues/593

### Backwards compatibility

are not because of updated param in MetaTransactionWallet

### DIFF
--- a/packages/protocol/contracts/common/MetaTransactionWallet.sol
+++ b/packages/protocol/contracts/common/MetaTransactionWallet.sol
@@ -198,7 +198,7 @@ contract MetaTransactionWallet is
    * @notice Executes multiple transactions on behalf of the signer.`
    * @param destinations The address to which each transaction is to be sent.
    * @param values The CELO value to be sent with each transaction.
-   * @param data The concatenated data to be sent in each transaction.
+   * @param data The concatenated transaction data with their respective lengths <length+data>.
    * @param dataLengths The length of each transaction's data.
    */
   function executeTransactions(
@@ -213,8 +213,16 @@ contract MetaTransactionWallet is
     );
     uint256 dataPosition = 0;
     for (uint256 i = 0; i < destinations.length; i++) {
-      executeTransaction(destinations[i], values[i], sliceData(data, dataPosition, dataLengths[i]));
-      dataPosition = dataPosition.add(dataLengths[i]);
+      require(
+        uint8(data[dataPosition]) == dataLengths[i],
+        "Data length must match specified length"
+      );
+      executeTransaction(
+        destinations[i],
+        values[i],
+        sliceData(data, dataPosition.add(1), dataLengths[i])
+      );
+      dataPosition = dataPosition.add(dataLengths[i].add(1));
     }
   }
 
@@ -227,6 +235,7 @@ contract MetaTransactionWallet is
    */
   function sliceData(bytes memory data, uint256 start, uint256 length)
     internal
+    pure
     returns (bytes memory)
   {
     // When length == 0 bytes.slice does not seem to always return an empty byte array.
--- a/packages/protocol/contracts/common/MetaTransactionWallet.sol
+++ b/packages/protocol/contracts/common/MetaTransactionWallet.sol
@@ -217,6 +217,8 @@ contract MetaTransactionWallet is
         uint8(data[dataPosition]) == dataLengths[i],
         "Data length must match specified length"
       );
+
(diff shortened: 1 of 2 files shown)


packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -304,7 +304,15 @@ contract('MetaTransactionWallet', (accounts: string[]) => {
             await wallet.executeTransactions(
               transactions.map((t) => t.destination),
```
ewilz, line 305: attempted to DRY up this parameter generation, but I clearly need to get up to speed on Typescript...

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -213,8 +213,16 @@ contract MetaTransactionWallet is
...
-      executeTransaction(destinations[i], values[i], sliceData(data, dataPosition, dataLengths[i]));
-      dataPosition = dataPosition.add(dataLengths[i]);
+      require(
+        uint8(data[dataPosition]) == dataLengths[i],
+        "Data length must match specified length"
+      );
+      executeTransaction(
+        destinations[i],
+        values[i],
+        sliceData(data, dataPosition.add(1), dataLengths[i])
```
yorhodes, line 225: what is this `dataPosition + 1`?

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -213,8 +213,16 @@ contract MetaTransactionWallet is
...
+      );
+      dataPosition = dataPosition.add(dataLengths[i].add(1));
```
yorhodes, line 227: ditto `dataLengths + 1`?

packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -328,7 +336,15 @@ contract('MetaTransactionWallet', (accounts: string[]) => {
               .executeTransactions(
                 transactions.map((t) => t.destination),
                 transactions.map((t) => t.value),
-                ensureLeading0x(transactions.map((t) => trimLeading0x(t.data)).join('')),
+                ensureLeading0x(
+                  transactions
+                    .map((t) => {
+                      const data = trimLeading0x(t.data) // @ts-ignore
```
yorhodes: why do we need the @ts-ignore?

packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -354,6 +370,29 @@ contract('MetaTransactionWallet', (accounts: string[]) => {
             assertEqualBN(await wallet.nonce(), 1)
           })
         })
+
+        describe('when lengths in length array do not match up with lengths in data', async () => {
+          it('reverts with correct message', async () => {
```
yorhodes: where are we checking the revert message?

ewilz: I was planning on it, but noticed we don't check messages, then forgot to change my description back. Perhaps it's something we SHOULD do though..

packages/protocol/test/common/metatransactionwallet.ts
ewilz: for the annoying reason of a shadow variable, which IIIIII think it perfectly fine(!!!) but it failed the build, so I will have to change this anyway.

packages/protocol/contracts/common/MetaTransactionWallet.sol
ewilz: there's a byte that now accounts for the length of the following byte segment. So when we move the cursor down, we now have to account for this extra one byte "metadata" telling us the length.

ewilz: NOT backwards compatible though. Does that cause any problems?

packages/protocol/contracts/common/MetaTransactionWallet.sol
ewilz: added a code comment

ewilz:
We will not move foward with this fix as we've come to the conclusion that the extra lengths + check is redundant.

PR #4981, opened 2020-09-04
[L16] Transfer of CELO may unexpectedly fail

ewilz:
### Description
Change from using address.transfer() to address.sendValue() from openzeppelin library

### Tested

everything should still work as expected. No test changes.
### Related issues

https://github.com/celo-org/celo-labs/issues/595

### Backwards compatibility

everything should continue working as expected

### DIFF
--- a/packages/protocol/contracts/governance/Governance.sol
+++ b/packages/protocol/contracts/governance/Governance.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.5.3;
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import "openzeppelin-solidity/contracts/math/Math.sol";
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "openzeppelin-solidity/contracts/utils/Address.sol";
 
 import "./interfaces/IGovernance.sol";
 import "./Proposals.sol";
@@ -34,6 +35,7 @@ contract Governance is
   using SafeMath for uint256;
   using IntegerSortedLinkedList for SortedLinkedList.List;
   using BytesLib for bytes;
+  using Address for address;
 
   uint256 private constant FIXED_HALF = 500000000000000000000000;
 
@@ -754,7 +756,7 @@ contract Governance is
     require(value > 0, "Nothing to withdraw");
     require(value <= address(this).balance, "Inconsistent balance");
     refundedDeposits[msg.sender] = 0;
-    msg.sender.transfer(value);
+    msg.sender.sendValue(value);
     return true;
   }
 
--- a/packages/protocol/contracts/governance/Governance.sol
+++ b/packages/protocol/contracts/governance/Governance.sol
@@ -35,7 +35,7 @@ contract Governance is
   using SafeMath for uint256;
   using IntegerSortedLinkedList for SortedLinkedList.List;
   using BytesLib for bytes;
-  using Address for address;
+  using Address for addressssss;
 
   uint256 private constant FIXED_HALF = 500000000000000000000000;
 
--- a/packages/protocol/contracts/governance/Governance.sol
+++ b/packages/protocol/contracts/governance/Governance.sol
@@ -35,7 +35,7 @@ contract Governance is
   using SafeMath for uint256;
   using IntegerSortedLinkedList for SortedLinkedList.List;
   using BytesLib for bytes;
-  using Address for addressssss;
+  using Address for address payable; // prettier-ignore
 
   uint256 private constant FIXED_HALF = 500000000000000000000000;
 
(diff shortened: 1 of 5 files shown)


packages/protocol/contracts/governance/Governance.sol
```diff
@@ -754,7 +756,7 @@ contract Governance is
     require(value > 0, "Nothing to withdraw");
     require(value <= address(this).balance, "Inconsistent balance");
     refundedDeposits[msg.sender] = 0;
-    msg.sender.transfer(value);
+    msg.sender.sendValue(value);
```
yorhodes, line 759: do we need to use the `toPayable` in collaboration with this?
https://docs.openzeppelin.com/contracts/2.x/api/utils#Address-toPayable-address-

yorhodes: I wonder if this is related to the prettier issue

packages/protocol/specs/harnesses/GoldTokenHarness.sol
```diff
@@ -11,7 +13,7 @@ contract GoldTokenHarness is GoldToken {
     require(to != address(this), "GoldToken cannot transfer to itself");
     require(value <= balanceOf(msg.sender), "Transfer value exceeded balance of sender");
 
-    address(uint160(to)).transfer(value); // TRANSFER.call.value(0).gas(gasleft())(abi.encode(msg.sender, to, value));
+    address(uint160(to)).sendValue(value); // TRANSFER.call.value(0).gas(gasleft())(abi.encode(msg.sender, to, value));
```
yorhodes, line 16: ditto on `toPayable`
cc @asaj re implications of changing this

yorhodes (commented):
some questions about the zeppelin lib

packages/protocol/contracts/governance/Governance.sol
yorhodes: the docs also flag taking care not to introduce reentrancy issues

ewilz: * `msg.sender` is inherently `address payable` so I don't think so.....? I think it would have failed to compile
* I was thinking the prettier issue may have had to do with the solidity linter linting an earlier version of solidity before there was a separate `address payable` type.... (though that still doesn't make complete sense given it's used elsewhere)
* will take a second look at the reentrancy stuff

packages/protocol/specs/harnesses/GoldTokenHarness.sol
ewilz: https://solidity.readthedocs.io/en/v0.5.3/types.html#address


>Implicit conversions from address payable to address are allowed, whereas conversions from address to address payable are not possible (the only way to perform such a conversion is by using an intermediate conversion to uint160).

^--- so intuitive ^_^ (sarcasm) I'm guessing this must be casting successfully?

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -378,7 +380,7 @@ contract LockedGold is
     address communityFund = registry.getAddressForOrDie(GOVERNANCE_REGISTRY_ID);
     address payable communityFundPayable = address(uint160(communityFund));
     require(maxSlash.sub(reward) <= address(this).balance, "Inconsistent balance");
-    communityFundPayable.transfer(maxSlash.sub(reward));
+    communityFundPayable.sendValue(maxSlash.sub(reward));
```
ewilz, line 383: this looks like reentrancy is possible. We could add the modifier? I'm unfamiliar with why it may have been omitted here though... @yorhodes

ewilz: wondering if reentrancy in general brings enough benefits to the table to makeup for all the pain it causes ^_^

packages/protocol/contracts/stability/Reserve.sol
```diff
@@ -332,7 +334,7 @@ contract Reserve is
    */
   function _transferGold(address payable to, uint256 value) internal returns (bool) {
     require(value <= getUnfrozenBalance(), "Exceeding unfrozen reserves");
-    to.transfer(value);
+    to.sendValue(value);
```
emaG3m, line 337: this isn't protected against reentrancy, but it seems to be handled in the logic in the external function, and the other function can only be called by one of our registered functions..

packages/protocol/specs/harnesses/GoldTokenHarness.sol
ewilz: we could use zeppelin's function if it makes this line more explicit in it's intentions, no strong opinions here

packages/protocol/contracts/governance/LockedGold.sol
ewilz: ohhh actually, we're sending to `GOVERNANCE_REGISTRY_ID` here which is our own contract. we could add nonReentrant for peace of mind though?

m-chrzan: There's no storage changes after the transfer happens. Where are you seeing the possibility of reentrancy?

ewilz: you're right!

ewilz:
I think this one should be good to go, unless there's still any reentrancy concerns, which I think everything's been accounted for nudge: @yorhodes @m-chrzan

yorhodes (approved):
🚀

PR #4982, opened 2020-09-04
[Wallet] Fix incorrect verification screen buttons in some cases

jeanregisser:
### Description

If the balance was updated after the verification state was fetched, the user would see a `skip` button even if the balance was enough to verify.

This was reproducible more easily on the Android emulator in the e2e test where it restores an account from a seed phrase.

I removed `isBalanceSufficient` from the verification state in redux and used a selector instead which derives the value when the balance changes.

### Other changes

None

### Tested

Updated tests. E2e test doesn't get stuck on the verification screen randomly.

### Backwards compatibility

Yes.


packages/mobile/src/identity/reducer.ts
```diff
@@ -349,7 +362,42 @@ export const secureSendPhoneNumberMappingSelector = (state: RootState) =>
...
+  const celoBalance = celoTokenBalanceSelector(state) || 0
+  return isBalanceSufficientForSigRetrieval(dollarBalance, celoBalance)
+}
+
+function isBalanceSufficientForAttestations(state: RootState, attestationsRemaining: number) {
+  const userBalance = stableTokenBalanceSelector(state) || 0
+  return new BigNumber(userBalance).isGreaterThan(
+    attestationsRemaining * ESTIMATED_COST_PER_ATTESTATION
+  )
+}
```
jeanregisser, line 377: This was copied from the existing code, but it looks like this is gonna fail for users with only CELO balances, right?

i1skn: Yeap, afaik we require cUSD to pay for attestations

packages/mobile/src/identity/reducer.ts
```diff
@@ -349,7 +362,42 @@ export const secureSendPhoneNumberMappingSelector = (state: RootState) =>
...
+
+const identityVerificationStateSelector = (state: RootState) => state.identity.verificationState
+
+const isBalanceSufficientSelector = (state: RootState) => {
+  const verificationState = state.identity.verificationState
+  const { phoneHashDetails, status, actionableAttestations } = verificationState
+  const attestationsRemaining = status.numAttestationsRemaining - actionableAttestations.length
+  const isBalanceSufficient = !phoneHashDetails.phoneHash
```
i1skn, line 385: If:
* we do not have `phoneHash`
* we do have enough money to retrieve it
* we do not have enough money for attestations

this would return `TRUE`, right?

i1skn (commented):
Great work on this!

packages/mobile/src/identity/reducer.ts
jeanregisser: Yes, it doesn't seem right.

Hum problem is we should sum the required amounts which is not directly possible with the current helpers.

jeanregisser: hum even summing amounts wouldn't be right
cause you can have just enough to fetch the pepper
but if you're already verified you don't need more for attestations
it's really a 2 steps process

do you see a different way to solve this?

i1skn: Actually, the only case when the issue, I’ve pointed, would occur is when fetching pepper has returned an error. User then would be able to start the process even if it was not enough money for attestations. But then if we fetch the pepper eventually - we would show them that balance is insufficient for attestations as a popup anyway. I thinks this is ok for the edge case, when we can not estimate amount needed for attestations.

i1skn (approved):
LGTM

Lss-Ankit:
@jeanregisser  Verified task on latest Android internal build v1.1.0  (1004294309) and Test flight build v1.1.0 (21) and found the following: 
Devices: Google Pixle Xl  (8.1) , iPhone 7+ (13.3.1). 

Tested Scenarios: 
- Skip button is shown on Confirm screen while verifying phone number on on-boarding flow  
- Skip button on Confirm screen is not shown when user taps on Confirm Phone number option from setting page. 
- "Skip for now"  button is shown with "Your current balance is not enough..." description when user dont have enough balance. 

**Please let me know if i need to verify anything else.** 
Thanks!

PR #4985, opened 2020-09-04
Removes css processors we dont need

aaronmgdr:
### Description

removes two CSS processors which were around from the old old days. 

### Tested

ran the tests and manually checked out a good sample of pages. 

### Related issues
fixes https://github.com/celo-org/celo-monorepo/pull/4975

### Backwards compatibility

😇


jmrossy (approved):
Nice 👍

PR #4989, opened 2020-09-04
[Wallet] Add e2e tests and improve flakiness

gnardini:
### Description

Added e2e tests for buying and selling CELO. Since they happen in the same screen as withdrawing, I united the three under a single test file so that the commons stuff (opening the CELO screen and going through onboarding if necessary) is run only once and tests don't take longer than necessary.

Added a secondary test file that runs test logging in as a new account. (ie. creating an account instead of importing it). This test goes through onboarding, sets up the account key and checks that the buy CELO button is not visible because of the lack of funds.

### Other Changes

Fixed some sources of flakiness:
- There were some `await` missing.
- Having tests too separated made them more error-prone and take longer. Most flows run on a single `it` now (only onboarding remains split in several ones).
- If tests got delayed and took too long, the PIN was sometimes required again. We now check in all places where the PIN could be required if the PIN screen popped up and input the PIN if so.

Also:
- Updated Detox version
- Added a `-t` option to the `run_e2e` script to run a single test file.
- Updated the README

Fixes flakiness due to `Switch` taps not registering some %  of the time:
https://github.com/celo-org/celo-monorepo/issues/5023 & https://github.com/celo-org/celo-monorepo/issues/5022


packages/mobile/e2e/src/FundedAccount.spec.js
```diff
@@ -0,0 +1,14 @@
+import dismissBanners from './utils/banners'
+import Onboarding from './usecases/Onboarding'
```
jeanregisser: nit: what do you think of naming it something like `RestoreAccountOnboarding`?

packages/mobile/e2e/src/usecases/ExchangeCelo.js
```diff
@@ -0,0 +1,71 @@
+import { enterPinUiIfNecessary } from '../utils/utils'
+import { DEFAULT_RECIPIENT_ADDRESS } from '../utils/consts'
+import { celoOnboarding } from '../utils/celoOnboarding'
```
jeanregisser: nit: what do you think of naming it `celoEducation`? this is the naming we've been using in the app / design so far.

packages/mobile/e2e/src/usecases/Onboarding.js
```diff
@@ -24,15 +24,11 @@ export default Onboarding = () => {
 
   it('Terms', async () => {
     await element(by.id('scrollView')).scrollTo('bottom')
-    expect(element(by.id('AcceptTermsButton'))).toBeVisible()
+    await expect(element(by.id('AcceptTermsButton'))).toBeVisible()
```
jeanregisser: 👍

packages/mobile/src/backup/BackupPhraseContainer.tsx
```diff
@@ -91,7 +91,11 @@ export class BackupPhraseContainer extends React.Component<Props> {
         </View>
         {mode === BackupPhraseContainerMode.READONLY && (
           <View style={styles.phraseContainer}>
-            {!!words && <Text style={styles.phraseText}>{words}</Text>}
+            {!!words && (
+              <Text style={styles.phraseText} testID="accountKeyWords">
```
jeanregisser: nit: I see we usually use `Capitalized` test IDs, though we've been inconsistent in other places.

packages/mobile/src/components/DevSkipButton.tsx
```diff
@@ -18,7 +18,11 @@ interface DispatchProps {
   devModeTriggerClicked: typeof devModeTriggerClicked
 }
 
-type Props = { nextScreen: keyof StackParamList; onSkip?: () => void } & StateProps & DispatchProps
+type Props = {
+  nextScreen?: keyof StackParamList
+  onSkip?: () => void
```
jeanregisser: nit: would be nice to separate the 2 use cases and enforce them with TypeScript.
if `nextScreen` is not specified, `onSkip` should be set.

```tsx
type Props = {
  nextScreen: keyof StackParamList
  onSkip?: () => void
} | {  onSkip: () => void }
```

jeanregisser (approved):
Awesome!! 💯 
Left some small nits.

packages/mobile/e2e/src/usecases/ResetAccount.js
```diff
@@ -0,0 +1,44 @@
+import { enterPinUiIfNecessary, waitForElementId, sleep } from '../utils/utils'
+import { SAMPLE_BACKUP_KEY } from '../utils/consts'
+
+export default ResetAccount = () => {
+  it('Reset Account by doing the Account Key quiz', async () => {
+    // This test has very high flakiness on Android. I did my best to fix it, but
+    // locally it works every time but on CI it fails 50%+ of the time.
+    // TODO: Keep investigating the source of flakiness of this test on Android.
+    if (device.getPlatform() === 'android') {
+      return
```
etuleu, line 10: I am hoping that upgrading our Circle instance will improve on this. I believe @annakaz is looking into that.

packages/mobile/scripts/run_e2e.sh
```diff
@@ -80,16 +83,18 @@ preloadBundle() {
 runTest() {
   reuse_param=""
   if [[ $DEV_MODE == true ]]; then
+    export NUM_RETRIES=0
```
etuleu, line 86: leftover?

etuleu (approved):
Looks good to me! nice work!

packages/mobile/src/backup/BackupPhraseContainer.tsx
gnardini: Good to know :)

packages/mobile/scripts/run_e2e.sh
gnardini: This is behind the `DEV_MODE` flag so that the flakey test runner doesn't run when running tests locally. The problem with it is that when tests are retried, the retry often fails because the state in the device is different than expected, so the error shown is usually not the correct one.

For example, let's say we have screens A and B, and a button in A.
Tests starts, presses button in A and fails while doing some check in B. Now the simulator is in B screen.
When running the retry, the error we see is "button not found in A" because the simulator is in B, but the real error was in the B screen.

Having the flakey test checker disabled in dev mode allows us to get the real error and this is the way I found to do that :)

PR #4990, opened 2020-09-04
[Wallet] Send attestation messages sequentially on Android

i1skn:
### Description

As a short-term solution for https://github.com/celo-org/celo-labs/issues/578 we will send attestation messages sequentially with a 5sec delay, so auto-import would not run into race condition. Real fix require some work with a native code for Android and will be done a bit later.

### Other changes

Changing `SMS_RETRIEVER_APP_SIGNATURE`, so the app does parse attestation messages on the dev environment

### Tested

Android and iOS device

### Related issues

- Only a short term solution to https://github.com/celo-org/celo-labs/issues/578

### Backwards compatibility

Yes

### Video

movie-name


packages/mobile/src/identity/verification.ts
```diff
@@ -6,7 +6,7 @@ import {
   AttestationsWrapper,
   UnselectedRequest,
 } from '@celo/contractkit/lib/wrappers/Attestations'
-import { retryAsync } from '@celo/utils/src/async'
+import { retryAsync, sleep } from '@celo/utils/src/async'
```
jeanregisser: nit: `redux-saga` provides a `delay` helper.
Usage: `yield delay(delayInMs)`

jeanregisser (approved):
LGTM 👍

Lss-Ankit:
Hi @i1skn @jeanregisser I have verified task using Android internal build v1.1.0 (1004294309) and found the following : 

- Attestation messages sequentially received with a ~5 sec delay and it is auto-import one by one. 

Note: I have verified with Indian number as Auto import is not working for me in US number because for US number i am  using third party app.

PR #4995, opened 2020-09-06
[L13] Not using the SafeMath library

ewilz:
### Description

Add safemath where possible

### Other changes

Added `ICeloVersionedContract` interface to `UsingPrecompiles`

### Tested

N/A should remain working as expected

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/603
### Backwards compatibility

N/A should remain working as expected

### DIFF
--- a/packages/protocol/specs/harnesses/LockedGoldHarness.sol
+++ b/packages/protocol/specs/harnesses/LockedGoldHarness.sol
@@ -1,10 +1,13 @@
 pragma solidity ^0.5.8;
 
 import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "contracts/governance/LockedGold.sol";
 import "./GoldTokenHarness.sol";
 
 contract LockedGoldHarness is LockedGold {
+  using SafeMath for uint256;
+
   GoldTokenHarness goldToken;
 
   /* solhint-disable-next-line no-empty-blocks */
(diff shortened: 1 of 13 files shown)


packages/protocol/specs/harnesses/LockedGoldHarness.sol
```diff
@@ -1,10 +1,13 @@
 pragma solidity ^0.5.8;
 
 import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
```
yorhodes, line 4: these changes are probably harmless but to clarify, these specs are only used in the `certora-test` formal verification and the contracts are never deployed on network afaik

yorhodes (approved):
LGTM

packages/protocol/specs/harnesses/LockedGoldHarness.sol
ewilz: thanks, useful information

packages/protocol/specs/harnesses/RegistryHarness.sol
```diff
@@ -123,23 +126,23 @@ contract RegistryHarness is IRegistryExtended {
...
-    return randomBoolMap[randomIndex];
+    randomIndex.add(1);
+    return randomBoolMap[randomIndex.add(1)
   }
 
   function getRandomUInt256() public returns (uint256) {
-    randomIndex++;
-    return randomUInt256Map[randomIndex];
+    randomIndex.add(1);
+    return randomUInt256Map[randomIndex.add(1)
```
ewilz: oops

PR #4996, opened 2020-09-06
Verify source code matches deployed or proposed bytecode

### LINKED ISSUE #4496: Script for checking a contracts directory corresponds to deployed bytecode

### LINKED ISSUE #4817: Upgrade proposal verification from contract source code commits
### Expected Behavior

### Current Behavior

What is the current behavior?

m-chrzan:
### Description

This PR adds a script that verifies that a given set of Solidity sources matches bytecodes deployed to a Celo network. An optional `proposal.json` can also be passed in, representing proposed changes that are not yet reflected by onchain Registry/proxies' state.

This check will rely on:
* Each (non-library) core contract being registered in the Registry
* Each linked library being proxied

The deployed Celo contracts do not currently satisfy these assumptions, so there is some temporary code included to handle the current case. The checks are also not 100% complete for the current RC1 deployment (e.g. the Registry itself is missed since it's not registered).

The main check happens in the `verifyBytecodes` function, which takes a list of core contract names and a build directory as input. The DFS works as follows:

1. If we're on a core contract, look up its address in the Registry. We keep track of library addresses as we encounter them.
2. Find all instances of `__LibraryName__...` in the compiled bytecode.
3. Ensure that each `__LibraryName__...` link corresponds to the same address in the onchain bytecode.
4. Link the compiled bytecode and verify it matches the onchain bytecode.
5. Add any freshly encountered libraries to the DFS stack.

The biggest remaining piece of work on this front is also verifying *proposed* changes, that is, given a Celo network and a Governance proposal, verifying that the given sources correspond to what the Celo system would look like after applying the proposal.

### Tested

Unit tests for the library doing the heavy lifting.

For a manual test against baklava,

    git checkout baklava
    yarn build
    mv build/contracts build/baklava-contracts
    git checkout m-chrzan/deployed-bytecode-check
    # point the script to build/baklava-contracts
    yarn build:ts && yarn truffle exec --network baklava scripts/truffle/verify-bytecode.js

### Other changes

Moved some duplicated code out into a new `lib/bytecode` library.

### Related issues

- Fixes #4496 
- Fixes #4817

### Backwards compatibility

Very much so.


packages/protocol/initialize-data.json
```diff
@@ -0,0 +1,3 @@
+{
```
nambrot: Would it make sense to version this to the release, i.e. this being `initiailze-data-release-1.json`?

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+
+const verifyLibraryPrefix = (bytecode: string, address: string) => {
+  if (bytecode.slice(2, 4) !== '73') {
+    throw new Error(`Library bytecode doesn't start with address load`)
+  } else if (bytecode.slice(4, 44) !== address) {
+    throw new Error(`Library bytecode loads unexpected address at start`)
+  }
+}
+
+interface VerificationContext {
```
nambrot, line 22: Move interface definition up

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  Object.keys(sourceLibraryPositions).forEach(library => {
+    if (!visited.has(library)) {
+      queue.push(library)
+      visited.add(library)
+    }
+  })
+}
+
+export const verifyBytecodesDfs = async (contracts: string[], artifacts: BuildArtifacts, registry: RegistryInstance, proposal: ProposalTx[], Proxy: Truffle.Contract<ProxyInstance>, web3: Web3, isBeforeRelease1: boolean = false) => {
```
nambrot: Optional Nit: What do you think about omitting the dfs suffix here as consumers shouldn't really care about the implementation detail?

packages/protocol/lib/bytecode.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  const match = bytecode.match(GENERAL_METADATA_REGEXP)
+  if (match === null) {
+    throw new Error('Only support stripping metadata from bytecodes generated by solc up to v0.5.13 with no experimental features.')
+  }
+  return match[1]
+}
+
+export interface LibraryLinks {
```
nambrot, line 37: An explanation of the value would be nice here

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+  // These contracts are not in the Registry (before release 1)
+  'ReserveSpenderMultiSig',
+  'GovernanceApproverMultiSig'
+]
+
+export interface LibraryPositions {
+  [library: string]: number[]
+}
+
+const addPosition = (positions: LibraryPositions, library: string, position: number) => {
```
nambrot: There is a fair number of global constants/functions in this file. I wonder if this file could benefit from a bit of a rearrangement/grouping. For example,  `addPosition` is a helper method for `collectLibraryPositions`, but in between are constants irrelevant to that tie

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+  while (match != null) {
+    // The firt capture group is the library's name
+    addPosition(libraryPositions, match[1], match.index)
+    match = libraryLinkRegExp.exec(bytecode)
+  }
+
+  return libraryPositions
+}
+
+export const collectLibraryAddresses = (bytecode: string, libraryPositions: LibraryPositions, libraryAddresses: LibraryAddresses = {}): LibraryAddresses => {
```
nambrot: Nit: consider adding a comment here, it's not immediately clear what this method does

nambrot: Just my 2 cents, but it was not clear from reading this method that it would mutate `libraryAddresses`

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+  return context.registry.getAddressForString(contract)
+}
+
+const getProxiedAddress = async (address: string, context: VerificationContext): Promise<string> => {
+  const proxy: ProxyInstance = await context.Proxy.at(context.web3.utils.toChecksumAddress(address))
+  return proxy._getImplementation()
+}
+
+const verifyLibraryPrefix = (bytecode: string, address: string) => {
+  if (bytecode.slice(2, 4) !== '73') {
```
nambrot: Please add explanations for these magical constants and consider naming the constants

packages/protocol/lib/compatibility/verify-bytecode.ts
nambrot: Or conversely, this is arguably the meat of the functionality, it would be nice to have a high-level description of the approach of this verification as a comment to this function

packages/protocol/initialize-data.json
ewilz: also, if we keep a history of our initialize data: `initialize-data-release-{i}.json` it might make sense to give it a directory?

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
+import { linkLibraries, stripMetadata } from '@celo/protocol/lib/bytecode'
+import { ProposalTx } from '@celo/protocol/scripts/truffle/make-release'
+import { ProxyInstance, RegistryInstance } from 'types'
+
+import { BuildArtifacts, } from '@openzeppelin/upgrades'
+
+import Web3 from 'web3'
+
+const ignoredContracts = [
```
ewilz, line 13: does `MetaTransactionWallet` need to be added here, or is that different?

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+ * Returns a LibraryPositions object, which, for each yet to be linked library,
+ * contains the bytecode offsets of where the library address should be
+ * inserted.
+ */
+export const collectLibraryPositions = (bytecode: string): LibraryPositions => {
+  const libraryLinkRegExp = new RegExp(libraryLinkRegExpString, 'g')
+  const libraryPositions: LibraryPositions = {}
+  let match = libraryLinkRegExp.exec(bytecode)
+  while (match != null) {
+    // The firt capture group is the library's name
```
ewilz: first* ?

packages/protocol/lib/bytecode.ts
```diff
@@ -0,0 +1,47 @@
...
+ * The Solidity compiler appends a Swarm Hash of compilation metadata to the end
+ * of bytecode. We find this hash based on the specification here:
+ * https://solidity.readthedocs.io/en/develop/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode
+ */
+
+const CONTRACT_METADATA_REGEXPS = [
+  // 0.5.8
+  'a165627a7a72305820.*0029',
+  // 0.5.12
+  'a265627a7a72315820.*64736f6c6343.*0032'
```
yorhodes: we are pinning to `0.5.13` in the audit response
can we add that here?

packages/protocol/lib/compatibility/verify-bytecode.ts
yorhodes: +1

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+const isImplementationChanged = (contract: string, context: VerificationContext): boolean => {
+  return context.proposal.some((tx: ProposalTx) => {
+    return tx.contract === `${contract}Proxy` && tx.function === '_setImplementation'
+  })
+}
+
+const isProxyChanged = (contract: string, context: VerificationContext): boolean => {
+  const registryId = context.web3.utils.soliditySha3({ type: 'string', value: contract })
+  return context.proposal.some((tx: ProposalTx) => {
+    return tx.contract === `Registry` && tx.function === 'setAddressFor' && tx.args[0] === registryId
```
yorhodes: perhaps these could be abstracted as `isRepointTx`

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+  })
+}
+
+const getProposedProxyAddress = (contract: string, context: VerificationContext): string => {
+  const registryId = context.web3.utils.soliditySha3({ type: 'string', value: contract })
+  const relevantTx = context.proposal.find((tx: ProposalTx) => {
+    return tx.contract === `Registry` && tx.function === 'setAddressFor' && tx.args[0] === registryId
```
yorhodes: ditto

packages/protocol/lib/registry-utils.ts
```diff
@@ -1,3 +1,5 @@
+export const celoRegistryAddress = '0x000000000000000000000000000000000000ce10'
```
yorhodes, line 1: +1

packages/protocol/lib/bytecode.ts
```diff
@@ -0,0 +1,47 @@
...
+]
+const GENERAL_METADATA_REGEXP = new RegExp(`^(.*)(${CONTRACT_METADATA_REGEXPS.map(r => '(' + r + ')').join('|')})$`, 'i')
```
yorhodes: I think you don't need the `^(.*)` and it will match starting from the end using `$` already

packages/protocol/lib/bytecode.ts
yorhodes: any reason this is an interface?

packages/protocol/lib/bytecode.ts
yorhodes: comment the regexp flags usage

packages/protocol/lib/bytecode.ts
```diff
@@ -0,0 +1,47 @@
...
+ * bytes of the address that should be substituted in).
+ */
+const padForLink = (name: string): string => {
+  return `__${name}`.padEnd(40, '_')
+}
+
+export const linkLibraries = (bytecode: string, libraryLinks: LibraryLinks): string => {
+  Object.keys(libraryLinks).forEach(libraryName => {
+    const linkString = padForLink(libraryName)
+    bytecode = bytecode.replace(RegExp(linkString, 'g'), libraryLinks[libraryName])
```
yorhodes, line 54: ditto

packages/protocol/lib/bytecode.ts
```diff
@@ -0,0 +1,47 @@
+/*
```
ewilz, line 2: biggest nit ever, I'm almost embarrassed, but all the other files in this directory end in `-utils` wondering if `bytecode-utils` will make it more obvious this is a file of helper functions.

packages/protocol/lib/compatibility/verify-bytecode.ts
yorhodes: +1

packages/protocol/lib/compatibility/verify-bytecode.ts
yorhodes: extract this to `isRegistryRepointTx`?

packages/protocol/lib/compatibility/verify-bytecode.ts
yorhodes: ditto

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+
+const getProposedImplementationAddress = (contract: string, context: VerificationContext): string => {
+  const relevantTx = context.proposal.find((tx: ProposalTx) => {
+    return tx.contract === `${contract}Proxy`
+  })
+
+  return relevantTx.args[0]
+}
+
+const getSoureBytecode = (contract: string, context: VerificationContext): string => {
```
yorhodes: `getSourceBytecode`?

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+const getProposedProxyAddress = (contract: string, context: VerificationContext): string => {
+  const registryId = context.web3.utils.soliditySha3({ type: 'string', value: contract })
+  const relevantTx = context.proposal.find((tx: ProposalTx) => {
+    return tx.contract === `Registry` && tx.function === 'setAddressFor' && tx.args[0] === registryId
+  })
+
+  return relevantTx.args[1]
+}
+
+const getProposedImplementationAddress = (contract: string, context: VerificationContext): string => {
```
yorhodes: should these be methods of a `VerificationContext` class to improve readability?

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+  Object.keys(sourceLibraryPositions).forEach(library => {
+    if (!visited.has(library)) {
+      queue.push(library)
+      visited.add(library)
+    }
+  })
+}
+
+export const verifyBytecodesDfs = async (contracts: string[], artifacts: BuildArtifacts, registry: RegistryInstance, proposal: ProposalTx[], Proxy: Truffle.Contract<ProxyInstance>, web3: Web3, isBeforeRelease1: boolean = false) => {
+  const queue = [...contracts]
```
yorhodes: could we filter the `ignoredContracts` out here?

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+
+const getOnchainBytecode = async (contract: string, isLibrary: boolean, context: VerificationContext): Promise<string> => {
+  const implementationAddress = await getImplementationAddress(contract, isLibrary, context)
+  const onchainBytecodeWithMetadata = await context.web3.eth.getCode(implementationAddress)
+  return stripMetadata(onchainBytecodeWithMetadata)
+}
+
+const dfsStep = async (queue: string[], visited: Set<string>, context: VerificationContext) => {
+  const contract = queue.pop()
+  if (ignoredContracts.includes(contract)) {
```
yorhodes: see other note about filtering

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+    return
+  }
+
+  const isLibrary = !context.contracts.includes(contract)
+
+  const sourceBytecode = getSoureBytecode(contract, context)
```
yorhodes: typo

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,277 @@
...
+  const onchainBytecode = await getOnchainBytecode(contract, isLibrary, context)
+
+  const sourceLibraryPositions = collectLibraryPositions(sourceBytecode)
+  collectLibraryAddresses(onchainBytecode, sourceLibraryPositions, context.libraryAddresses)
```
yorhodes: not clear these mutate
can we do any of these as a map across the contracts outside the DFS step, and only do the actual linking within the DFS?

packages/protocol/lib/compatibility/verify-bytecode.ts
yorhodes: +1

packages/protocol/initialize-data.json
m-chrzan: Do we want to keep historical initialize-data? I was thinking about adding an invariant that with this particular file, you can deploy whatever contracts are currently checked in.

packages/protocol/lib/compatibility/verify-bytecode.ts
ewilz: agreed. (even b4 I saw your comments). I'm having trouble understanding the purpose of this function

packages/protocol/test/compatibility/verify-bytecode.ts
```diff
@@ -0,0 +1,427 @@
...
+}
+
+const deployProxiedContract = async (Contract: any, from: string) => {
+  const proxy = await Proxy.new()
+  const contract = await Contract.new({ from })
+  await proxy._setImplementation(contract.address)
+  return Contract.at(proxy.address)
+}
+
+contract('', (accounts) => {
```
yorhodes, line 37: thanks for making these comprehensive! really increases my confidence

yorhodes (commented):
great work!

packages/protocol/lib/bytecode.ts
m-chrzan: > any reason this is an interface?

I guess just for clarity when writing out types for functions that consume this? What alternative would you suggest?

packages/protocol/lib/compatibility/verify-bytecode.ts
m-chrzan: Refactoring LibraryPositions and LibraryAddresses into classes.

packages/protocol/initialize-data.json
m-chrzan: For now renamed to `example-initialize-data.json` to make it clear this is not the official data used for deployment, just to make testing possible.

packages/protocol/lib/compatibility/verify-bytecode.ts
m-chrzan: This is only to filter out contracts that appear in the `CeloContractName` enum in `lib/registry-utils`. MetaTransactionWallet is a contract that would be deployed on a per-user basis - it's not a part of the core Celo system.

packages/protocol/lib/compatibility/verify-bytecode.ts
yorhodes: this is still unresolved

packages/protocol/lib/compatibility/verify-bytecode.ts
m-chrzan: Would you say that making `LibraryAddresses` a class, and `collect` a method on it, makes it clear enough that this could mutate?

I think in future releases we could modify this to not even be a DFS. With libraries being proxied, looks like Asa is having even them be entered in the Registry. At that point we can just feed this script a list of all core contracts and libraries, and like you're suggesting, gather addresses in a preprocessing step, using the Registry as a source of truth for that.

I suggest we leave this as a future improvement.

packages/protocol/lib/compatibility/verify-bytecode.ts
m-chrzan: I think should be fine on most recent diff? If you click the link of this comment I think it shows you the original commit.

packages/protocol/lib/bytecode.ts
m-chrzan: The `^(.*)` capture group is necessary to extract the non-metadata part of bytecode after a match.

packages/protocol/lib/bytecode.ts
m-chrzan: Always appreciate a good nit! I did do this deliberately because I've never been a huge fan of the `-utils` suffix going on everything in `lib/`, but could change it for consistency.

yorhodes (approved):
LGTM :)

nambrot:
You'll want https://github.com/celo-org/celo-monorepo/pull/5235 once its merged

PR #5000, opened 2020-09-07
[Wallet] Fix user seeing the enter invite code screen after upgrading to 1.1.0

### LINKED ISSUE #4997: [Wallet] After upgrading to 1.1.0 app displays the enter invite code screen
### Expected Behavior

Upgrading to 1.1.0 with an already configured account, shouldn't prompt the user to enter an invite code.

### Current Behavior

Issue was first reported by @marekolszewski in https://celo-org.slack.com/archives/CL7BVQPHB/p1599428329407100

Here are the steps to reproduce:
1. install v1.0.0
1. create new account, skip invite code
1. install v1.1.0 update
1. launch app
1. observe that the screen that opens is prompting to enter an invite code, instead of the wallet home feed with the transactions

This issue can be reproduced on iOS and Android.
It also happens when upgrading from v1.0.1 to v1.1.0

Issue does NOT reproduce when first installing v1.0.0 or v1.0.1 and restoring an existing account, and then upgrading to v1.1.0.

jeanregisser:
### Description

There was a regression introduced by the change in https://github.com/celo-org/celo-monorepo/pull/4977/files#diff-55bba9805f8b51290d1be787ca2cbca6L507-R506

See details in #4997

Though it might looks confusing, `redeemComplete` needs to be set to true when skipping the invite code to avoid this problem.
This is similar to what is done when restoring an account from a seed phrase in https://github.com/celo-org/celo-monorepo/blob/910ab0a5218c05f712422b774dcff1aedf495289/packages/mobile/src/import/saga.ts#L79

Additional note: In the long term I think we would benefit from using more meaningful actions which update multiple parts of the state tree in an atomic manner.
Currently there are parts of the code which dispatch multiple actions sequentially, i.e. using actions as "glorified setters"
See description of the problem here: https://twitter.com/dan_abramov/status/800310164792414208?s=20

### Other changes

None.

### Tested

Added new test.

### Related issues

- Fixes #4997

### Backwards compatibility

Yes.

### DIFF
--- a/packages/mobile/src/invite/saga.ts
+++ b/packages/mobile/src/invite/saga.ts
@@ -45,6 +45,8 @@ import {
   sendInviteFailure,
   sendInviteSuccess,
   SENTINEL_INVITE_COMMENT,
+  skipInviteFailure,
+  skipInviteSuccess,
   storeInviteeData,
 } from 'src/invite/actions'
 import { createInviteCode } from 'src/invite/utils'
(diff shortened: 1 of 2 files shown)


packages/mobile/src/invite/saga.ts
```diff
@@ -382,6 +382,11 @@ export function* skipInvite() {
   try {
     ValoraAnalytics.track(OnboardingEvents.invite_redeem_skip_start)
     yield call(getOrCreateAccount)
+    // TODO: refactor this, the multiple dispatch calls are somewhat confusing
+    // (`redeemInviteSuccess` though we're actually skipping and `setHasSeenVerificationNux` though the user hasn't seen it),
+    // we should prefer a more atomic approach with a meaningful action type
+    // Set redeem invite complete so user isn't brought back into nux flow
+    yield put(redeemInviteSuccess())
```
gnardini: There's a `skipInviteSuccess` action that is not called anywhere and sets `redeemComplete` to `true` as well, would it be better to use that one?

gnardini (approved):
Great job on the quick fix, agree on the refactor TODO :)

packages/mobile/src/invite/saga.ts
jmrossy: +1 for `skipInviteSuccess` thought they seem equivalent

jmrossy (approved):

packages/mobile/src/invite/saga.ts
jeanregisser: Yes thanks `skipInviteSuccess` is better 👍 
I've also added back the use of `skipInviteFailure` so it resets the state too in case of errors.

Lss-Ankit:

Test scenarios:

1. Verify user should able to install valora V1.0.1 build from play store
2. Verify user is able to upgrade valore v1.0.1 to internal build v1.1.0
3. Verify Invite code page should not show when user upgrade

Devices: iPhone 6+ (12.4.5), Google pixel Xl 8.1

Fixed issues:

#4997

Thanks !

PR #5002, opened 2020-09-07
[Wallet] Add migration to set redeemComplete to true if account exists

### LINKED ISSUE #4997: [Wallet] After upgrading to 1.1.0 app displays the enter invite code screen

gnardini:
### Description

Added a migration for the Redux store that sets `redeemComplete` to `true` if there is an account saved on the store.

### Tested

1- Manually set `redeemComplete` to false in the store and opened the app. 
2- I saw the invite screen even though I had an account. 
3- Add the migrations in the PR and opened the app again. 
4- Saw on the logs that migrations run and I appeared on the Home screen.

### Related issues

- Fixes #4997

### Backwards compatibility

That's the point of the PR :D

### DIFF
--- a/packages/mobile/src/redux/migrations.test.ts
+++ b/packages/mobile/src/redux/migrations.test.ts
@@ -90,4 +90,17 @@ describe('Redux persist migrations', () => {
     expect(migratedSchema.web3.dataEncryptionKey).toBe('key')
     expect(migratedSchema.web3.commentKey).toBe(undefined)
   })
+
+  it('works for v5 to v6', () => {
+    const v5Stub = {
+      invite: {
+        redeemComplete: false,
+      },
+      web3: {
+        account: 'some_account',
+      },
+    }
+    const migratedSchema = migrations[6](v5Stub)
+    expect(migratedSchema.invite.redeemComplete).toBe(true)
+  })
 })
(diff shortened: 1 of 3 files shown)


jeanregisser (approved):
Perfect! Thanks for this!

Lss-Ankit:

**Test scenarios:**

1. Verify user should able to install valora V1.0.1 build from play store 
2. Verify user is able to upgrade valore v1.0.1 to internal build v1.1.0
3. Verify Invite code page should not show when user upgrade 


**Devices:** iPhone 6+ (12.4.5), Google pixel Xl 8.1

Fixed issues:

#4997 

Thanks !

PR #5011, opened 2020-09-08
[Attestation Service] get_attestations endpoint, HA support for delivery receipts

timmoreton:
### Description

All delivery state is now persisted in the database. Supports HA setups with multiple Attestation Service instances.

Deliveries can be re-requested, regardless of their delivery status. It kicks off the next attempt, assuming retries are available, with the next provider.

Test attestation endpoint now accepts `provider` parameter to force provider to test:

```
tim@Tims-MacBook-Pro attestation-service % curl -H "Content-Type: application/json" -d '{ "phoneNumber":"+1415xxxxxxx","message":"hello","signature":"", "provider":"nexmo"}' localhost:3000/test_attestations

{"success":true,"identifier":"0xfc992dad242cca0dd8607aad3cab6ff4e2f9e47b0cd5c34df1ca19d7d641a373","account":"0xxxx","issuer":"0xxxx","attempt":0,"countryCode":"US","status":"Sent","salt":"9f68a7c35c9d2b04cbb39a714096d66c927132edb72f4be93dd92b919c8db47a","provider":"nexmo"}
```

It also randomly generates a "salt" so that tests can be stored alongside regular attestations in the database, so they can share more code and be accessible via `get_attestations`. 

New endpoint `/get_attestations` to retrieve delivery info:

```
tim@Tims-MacBook-Pro attestation-service % curl -i -H "Content-Type: application/json"  'localhost:3000/get_attestations?account=0xxxx&issuer=0xxxx&phoneNumber=%2B1415xxxxxxx&salt=9f68a7c35c9d2b04cbb39a714096d66c927132edb72f4be93dd92b919c8db47a'
HTTP/1.1 200 OK
X-Powered-By: Express
X-Request-Id: ba5f9a0c-dc47-40dd-a6b9-df8c2b31caeb
X-RateLimit-Limit: 50
X-RateLimit-Remaining: 48
Date: Tue, 08 Sep 2020 23:01:13 GMT
X-RateLimit-Reset: 1599606076
Content-Type: application/json; charset=utf-8
Content-Length: 278
ETag: W/"116-lQIpekV90rYPlJEkNfRYo95+NcA"
Connection: keep-alive

{"success":true,"identifier":"0xfc992dad242cca0dd8607aad3cab6ff4e2f9e47b0cd5c34df1ca19d7d641a373","account":"0xxxx","issuer":"0xxxx","attempt":0,"countryCode":"US","status":"Delivered","provider":"nexmo","duration":2274}
 ```

Delivery info includes time taken to final delivery success or failure, in ms.

The `celocli identity:test-attestation-service` command now also uses the `/get_attestations` endpoint to await a delivery receipt, so that validators have another way of testing that receipts are configured correctly.

### Other changes

* new metric `attestation_service_healthy` for overall service health
* If an attestation is not found on-chain, the request retries up to 10 secs to avoid the case where the full node is running behind a block or two (not if the attestation is found on-chain but is complete)
* Parameter `DB_RECORD_EXPIRY_MINS` to set db record expiry, and now done on a timer, not when attestations happen
* Fix default config's setting of `MAX_DELIVERY_ATTEMPTS`.  Delivery attempts now striped across providers, i.e try other provider next before retry with same provider 
* Fix broken metrics `attestation_requests_believed_delivered_sms` and `attestation_requests_failed_to_deliver_sms`
* Setting `NODE_ENV=dev` allows signature and other checks to be skipped (do not run in production)
* mention TheCelo in docs

### Tested

Baklava validator.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### CODE REVIEW

File: packages/attestation-service/src/db.ts [hunk 1]
```diff
@@ -1,24 +1,42 @@
...
       logging: (msg: string, sequelizeLogArgs: any) =>
         rootLogger.debug({ sequelizeLogArgs, component: 'sequelize' }, msg),
     })
     rootLogger.info('Initializing Database')
-    return sequelize.authenticate() as Promise<void>
+    await sequelize.authenticate()
+
+    // Check four times every `dbRecordExpiryMins` period to delete records older than dbRecordExpiryMins
+    dbRecordExpiryMins = parseInt(fetchEnvOrDefault('DB_RECORD_EXPIRY_MINS', '60'), 10)
+    const dbRecordExpiryCheckEveryMs = dbRecordExpiryMins * 15 * 1000
```
aslawson, line 38: why times 15?  shouldn't it be `dbRecordExpiryMins * 60 (sec/min) * 1000 (ms/sec)` ?

packages/attestation-service/src/index.ts
```diff
@@ -50,6 +59,11 @@ async function init() {
     res.send('Ready').status(200)
   })
   app.get('/healthz', rateLimiter, asyncHandler(handleLivenessRequest))
+  app.get(
+    '/get_attestations',
+    rateLimiter,
```
aslawson: if mobile client will use this after ever request, do we want it ratelimit with the same ratelimiter as the healthz/status/ready endpoint?  

we don't currently ratelimit the `/attestations`, but these should be limited similarly

packages/attestation-service/migrations/20200903211921-attestation-v3.js
```diff
@@ -0,0 +1,90 @@
...
+        allowNull: true,
+      })
+
+      await queryInterface.addColumn('Attestations', 'ongoingDeliveryId', {
+        type: Sequelize.STRING,
+        allowNull: true,
+      })
+
+      await queryInterface.addColumn('Attestations', 'providers', {
+        type: Sequelize.STRING,
```
aslawson, line 30: why not `Sequelize.ARRAY(Sequelize. STRING)`?  avoid having to manually parse with a delimiter

packages/attestation-service/src/models/attestation.ts
```diff
@@ -28,16 +46,46 @@ export default (sequelize: Sequelize) => {
...
+  model.prototype.key = function(): AttestationKey {
+    return { account: this.account, identifier: this.identifier, issuer: this.issuer }
+  }
+
+  model.prototype.provider = function(): string | null {
+    const pl = this.providers.split(',')
+    return this.providers ? pl[this.attempt % pl.length] : null
+  }
+
+  model.prototype.recordError = function(error: string) {
```
aslawson, line 72: where does this get used? i don't see any references to it

packages/attestation-service/migrations/20200903211921-attestation-v3.js
```diff
@@ -0,0 +1,90 @@
...
+
+    try {
+      await queryInterface.removeColumn('Attestations', 'smsProvider')
+
+      await queryInterface.addColumn('Attestations', 'countryCode', {
+        type: Sequelize.STRING,
+        allowNull: true,
+      })
+
+      await queryInterface.addColumn('Attestations', 'phoneNumber', {
```
nambrot, line 14: So we at some point removed phone numbers from the database to enhance privacy. Not saying we have to maintain that, but we should probably think twice whether we want to reintroduce or not

packages/attestation-service/migrations/20200903211921-attestation-v3.js
```diff
@@ -0,0 +1,90 @@
...
+        type: Sequelize.STRING,
+        allowNull: true,
+      })
+
+      await queryInterface.addColumn('Attestations', 'message', {
```
nambrot, line 19: Why do we persist the message? It seems risky as we are adding more surface areas for attestation messages to be leaked?

packages/attestation-service/migrations/20200903211921-attestation-v3.js
```diff
@@ -0,0 +1,90 @@
...
+
+      await queryInterface.addColumn('Attestations', 'errors', {
+        type: Sequelize.STRING,
+        allowNull: true,
+      })
+
+      await queryInterface.addIndex(
+        'Attestations',
+        ['ongoingDeliveryId'],
+        { fields: ['ongoingDeliveryId'], unique: false },
```
nambrot: the ID is not unique?

packages/attestation-service/src/request.ts
```diff
@@ -14,6 +16,43 @@ export enum ErrorMessages {
   NODE_IS_STUCK = 'Full node is not up to date',
 }
 
+export const AttestationRequestType = t.type({
```
nambrot: I originally put it in a common package since requestors should have access to this type as well. By separating it out, we lose the benefit of keeping the type in sync?

packages/attestation-service/src/request.ts
```diff
@@ -14,6 +16,43 @@ export enum ErrorMessages {
...
+
+  // Returned for errors in 1.0.x
+  error: t.union([t.undefined, t.string]),
+  errors: t.union([t.undefined, t.string]),
+
+  // Returned for successful send in 1.0.x
+  provider: t.union([t.undefined, t.string]),
+
+  // New fields
+  identifier: t.union([t.undefined, t.string]),
```
nambrot: Why return these fields, wouldn't the caller know these?

packages/attestation-service/src/requestHandlers/get_attestation.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  async getAttestationRecord() {
+    return findAttestationByKey(this.key, {
+      logging: this.sequelizeLogger,
+    })
+  }
+}
+
+export async function handleGetAttestationRequest(
```
nambrot, line 46: There seems to be no authentication in this endpoint, is that intentional?

packages/attestation-service/src/db.ts
timmoreton: the comment tried to explain it...

packages/attestation-service/src/index.ts
timmoreton: I think we do want to rate limit it.  i'll add rate limiting to `/attestations`

packages/attestation-service/migrations/20200903211921-attestation-v3.js
timmoreton: Only available in Postgres -would reduce range of databases we could use.

packages/attestation-service/src/models/attestation.ts
timmoreton: `src/sms/index.ts` a couple of places

packages/attestation-service/src/db.ts
timmoreton: maybe I should have done `* 60 / 4` and let the compiler optimize it

packages/attestation-service/migrations/20200903211921-attestation-v3.js
timmoreton: the message needs persisting because a separate instance may receive the callback saying that delivery failed, and then need to resend the next SMS to the next provider

timmoreton: I did consider removing the contents of the message on eventual delivery success or failure (running out of retries).  The number and message are never returned to clients.

packages/attestation-service/migrations/20200903211921-attestation-v3.js
timmoreton: indeed, see comment below.

packages/attestation-service/migrations/20200903211921-attestation-v3.js
timmoreton: I'll make it unique

packages/attestation-service/src/request.ts
timmoreton: yeah but would be nice if metadata crawler could also call the get endpoint, get these fields and use them to cross reference against client side data.

packages/attestation-service/src/requestHandlers/get_attestation.ts
timmoreton: Yes, I felt that the initial supplied fields were sufficient. There's nothing returned that is sensitive.  Do you feel differently?  (Also means the metadata crawler could use this endpoint)

packages/attestation-service/src/db.ts
aslawson: oh i see, so that we check within 15 seconds of its expiry.

packages/attestation-service/src/models/attestation.ts
aslawson: oh gotcha -- github had defaulted the toggle of that file to closed 🙄

packages/attestation-service/migrations/20200903211921-attestation-v3.js
aslawson: to clarify -- the separate instances will be expected to share the DB then?

timmoreton: yes, exactly

packages/attestation-service/migrations/20200903211921-attestation-v3.js
timmoreton: actually, this will be null for many records. So it can't be unique AFAIU

packages/attestation-service/migrations/20200903211921-attestation-v3.js
timmoreton: I will add a section to the docs explaining this architecture

packages/attestation-service/src/index.ts
timmoreton: now done.

packages/attestation-service/migrations/20200903211921-attestation-v3.js
nambrot: I believe you should be able to specify a unique index while allowing NULL values?

packages/attestation-service/src/request.ts
nambrot: How would the crawler get this response without requesting it, for which it  needs to know the identifier?

packages/attestation-service/src/requestHandlers/get_attestation.ts
nambrot: Not directly, just wanted to make sure it's explicit. We are "leaking" additional information in terms of deliverability status  and country code

timmoreton: Actually, you require the salt, so only the initial requester has that.

packages/attestation-service/src/request.ts
timmoreton: In fact, you're right, we can't use it for metadata crawler since it requires the `salt`.

packages/attestation-service/migrations/20200903211921-attestation-v3.js
timmoreton: done!

packages/attestation-service/src/models/attestation.ts
```diff
@@ -28,16 +46,51 @@ export default (sequelize: Sequelize) => {
...
-      AttestationStatus.FAILED,
-      AttestationStatus.UNABLE_TO_SERVE,
-    ].includes(this.status)
+  model.prototype.key = function(): AttestationKey {
+    return { account: this.account, identifier: this.identifier, issuer: this.issuer }
+  }
+
+  model.prototype.provider = function(): string | null {
+    const pl = this.providers.split(',')
+    return this.providers ? pl[this.attempt % pl.length] : null
```
aslawson, line 69: what is this.attempt initialized as?  wondering if
1. this will ever be `null` or 
2. if on first attempt, this will actually set this to `1` in which case you wouldn't start with the 0 index provider

packages/attestation-service/src/index.ts
```diff
@@ -33,23 +42,27 @@ async function init() {
...
-  app.use([requestIdMiddleware(), loggerMiddleware])
-  const port = process.env.PORT || 3000
-  app.listen(port, () => rootLogger.info({ port }, 'Attestation Service started'))
-
   const rateLimiter = rateLimit({
     windowMs: 5 * 60 * 100, // 5 minutes
     max: 50,
   })
+  const app = express()
+  app.use([requestIdMiddleware(), loggerMiddleware, rateLimiter])
```
aslawson, line 53: so this now limits all requests with the same ratelimiter?

packages/attestation-service/src/metrics.ts
```diff
@@ -61,6 +61,11 @@ export const Counters = {
     labelNames: ['provider', 'country', 'code'],
     help: 'Counter for error code of each failed delivery attempt by provider and country',
   }),
+  attestationRequestsByNumberType: new Counter({
+    name: 'attestation_requests_number_type',
+    labelNames: ['country', 'type'],
+    help: 'Counter for requests by country and type of phone number',
```
aslawson, line 71: what do you mean type of phone number? is one of these the prefix and the other the country code itself?

packages/attestation-service/src/models/attestation.ts
```diff
@@ -28,16 +46,51 @@ export default (sequelize: Sequelize) => {
...
+      this.status == AttestationStatus.NotSent.valueOf() ||
+      // tslint:disable-next-line: triple-equals
+      this.status == AttestationStatus.Failed.valueOf()
+    )
+  }
+
+  model.prototype.currentError = function() {
+    if (this.failure()) {
+      const errors = this.errors ? JSON.parse(this.errors) : {}
+      return errors[this.attempt]?.error ?? errors[this.attempt - 1]?.error ?? undefined
```
aslawson, line 94: why does this look at the past two errors?

packages/attestation-service/src/sms/index.ts
```diff
@@ -100,197 +120,300 @@ export function unsupportedRegionCodes() {
...
+  const transaction = await sequelize!.transaction({ logging: sequelizeLogger })
+
+  try {
+    const parsedNumber = phoneUtil.parse(phoneNumber)
+    const numberType = parsedNumber ? phoneUtil.getNumberType(parsedNumber) : null
+    const countryCode = parsedNumber ? phoneUtil.getRegionCodeForNumber(parsedNumber) : null
+
+    let providers: SmsProvider[] = countryCode
+      ? smsProvidersFor(countryCode, phoneNumber, logger)
+      : []
```
aslawson, line 203: any reason not to just have the default provider list here? or are we certain that providers will fail to send if phoneUtil can not get the country code?

aslawson (approved):

packages/attestation-service/src/index.ts
timmoreton: yep. I will raise this number however, as a load balancer might well ping every 10secs by itself

packages/attestation-service/src/sms/index.ts
timmoreton: in fact this is a failsafe. Parsing the request itself fails if the phone number is not E164 format.

packages/attestation-service/src/metrics.ts
timmoreton: ok I've now added string literals so you get ```attestation_requests_number_type{country="US",type="fixed_Line_or_mobile"} 1```

packages/attestation-service/src/models/attestation.ts
timmoreton: because we need to differentiate between waiting for the final attempt's delivery failure, and receiving that and not attempting to send any more. If all retries are exhausted, then the final error will be recorded on the preceding attempt.

packages/attestation-service/src/models/attestation.ts
timmoreton: Always zero: `sms/index.ts:183`

packages/attestation-service/src/db.ts
```diff
@@ -219,3 +225,49 @@ async function purgeExpiredRecords() {
...
+    if (failureReason) {
+      rootLogger.warn(`Health check failed: ${failureReason}`)
+    }
+  } catch {
+    rootLogger.warn(`Health check failed`)
+  }
+}
+
+// Check health and return failure reason, null on success.
+export async function doHealthCheck(): Promise<string | null> {
```
nambrot, line 242: Should this live in this `db` file?

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -68,18 +70,28 @@ class AttestationRequestHandler {
...
-      const state = await attestations.getAttestationState(this.key.identifier, account, issuer)
+    if (isDevMode()) {
+      return
+    }
 
-      if (state.attestationState !== AttestationState.Incomplete) {
-        Counters.attestationRequestsWOIncompleteAttestation.inc()
-        throw new Error(NO_INCOMPLETE_ATTESTATION_FOUND_ERROR)
+    // Check the on-chain status of the attestation. If it's marked Complete, don't do it.
+    // If it's missing, the full node could be behind by a block or two. Try a few times before erroring.
```
nambrot, line 78: Would it be better to check the age of the latest block here?

packages/attestation-service/src/db.ts
timmoreton: I considered moving it, but `db` has always had everything to do with ck and the full node in it too. will consider splitting in future

packages/attestation-service/migrations/20200924232226-attestation-v4.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+module.exports = {
```
aslawson, line 2: do we need another migration in the same release?  why not just add this in the previous migration?

packages/attestation-service/src/requestHandlers/attestation.ts
aslawson: Have we been seeing issues for validators that the attestation is missing because the full node is behind?  We already wait 5 blocks worth between when the client requests the issuers and it selects them. Is waiting 2 more blocks worth likely to help those cases or are they typically very behind?

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -50,36 +51,48 @@ class AttestationRequestHandler {
...
+    const attestations = await kit.contracts.getAttestations()
+    for (let i = 0; i < 4; i++) {
+      const state = await attestations.getAttestationState(this.key.identifier, account, issuer)
+      if (state.attestationState === AttestationState.Incomplete) {
```
aslawson: what is the return value of `state` if the attestation doesn't exist yet? 

i feel like you'd get an error...in which case you need a try/catch to all for retry.

or if it returns null, this `null.attestationState` would cause an exception as well.

packages/docs/operations-manual/attestation-service.md
```diff
@@ -34,7 +34,11 @@ Attestation Service exposes a HTTP endpoint, but it is strongly recommended that
...
 
-Attestation Service currently does not support setups where multiple instances reside behind a load balancer. This will be addressed in a future release.
+Attestation Service is a stateless service that uses a database to persist status of current and recently completed SMS delivery attempts. The most straightforward deployment architecture is to have a single machine or VM running three containers: one the attestation service, a Celo Blockchain node, and a single database instance.
+
+For a high availability setup, multiple instances can be deployed behind a load balancer configured with a round robin routing policy using the instances' `/healthz` endpoint as a healthcheck. Deploying a high availability database setup is beyond the scope of these instructions, but is straightforward with most cloud providers.  In this setup, if a delivery report for an SMS issued by one instance is received by another instance, that instance can act on it and resend if necessary.
```
aslawson: Maybe its inferred, but it could be beneficial to explicitly say that all attestation service instances should share a single database instance/setup.

packages/docs/operations-manual/attestation-service.md
```diff
@@ -325,4 +342,6 @@ Administrative metrics:
 
 ### Blockchain
 
-The number of requested and entirely completed attestations is in effect recorded on the blockchain. The values can be seen by entering the Validator's address on the [Celo Explorer](https://explorer.celo.org) and clicking the 'Celo Info' tab.  
+The number of requested and entirely completed attestations is in effect recorded on the blockchain. The values can be seen at [Celo Explorer](https://explorer.celo.org): enter the Validator's address and click the 'Celo Info' tab.  
+
+[TheCelo](https://thecelo.com/?tab=attestations) tracks global attestation counts and success rates, and shows detailed information about recent attestation attempts.
```
aslawson, line 347: nice plug for thecelo! :)

packages/attestation-service/src/db.ts
```diff
@@ -131,6 +132,11 @@ export async function initializeKit() {
   }
 }
 
+export async function startPeriodicHealthCheck() {
+  await tryHealthCheck()
+  setInterval(tryHealthCheck, 60 * 1000)
```
aslawson, line 137: should the rate be configurable?

aslawson (approved):
Looks mostly good.  Some thoughts above, but nothing urgent.

packages/attestation-service/src/requestHandlers/attestation.ts
timmoreton: We have seen it last month. After calling `selectIssuers` we don't wait, and it's the result of that that needs to be visible to the Attestation Service

packages/attestation-service/migrations/20200924232226-attestation-v4.js
timmoreton: because I had it deployed on baklava already

packages/attestation-service/src/db.ts
timmoreton: I felt 1min was ok.

packages/docs/operations-manual/attestation-service.md
timmoreton: edited to clarify this

packages/attestation-service/src/requestHandlers/attestation.ts
timmoreton: hmm possibly! Edited to catch this.

packages/attestation-service/src/requestHandlers/attestation.ts
aslawson: sounds like we need to dig into why the client is not waiting

PR #5029, opened 2020-09-09
Adds replica e2e tests and docs

trianglesphere:
### Description

Adds a new e2e test for hot swapping replica and adding docs for that.

### Other changes

Some changes to the e2e geth runner to more fully support proxies and also support secondaries/replicas.
Also add the ability to deploy testnets w/ replicas/secondaries.
 - Add the line `SECONDARIES=N` to the env file. There will be secondaries for validator `0-N` and they will connect to proxies as needed.

### Tested

replica test runs and passes

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


mcortesi (approved):
web3-istanbul.ts file is empty, can be deleted

PR #5031, opened 2020-09-09
OZ Audit Phase 4: M02 + M03

ewilz:
### Description

M02: Instead of having extra `dataLength` parameter in `MetaTransactionWallet.executeTransactions()`, `data` contains lengths internally to guarantee matchup.
M03: function reverts if data has extra unused bytes appended at the end

### Other changes

N/A

### Tested

Test added for revert, otherwise everything should perform as expected.
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/597 
https://github.com/celo-org/celo-labs/issues/593

### Backwards compatibility

Parameters changed on MetaTransactionWallet.executeTransactions() but we haven't officially published an initial version, so I don't think this matters

### DIFF
--- a/packages/protocol/contracts/common/MetaTransactionWallet.sol
+++ b/packages/protocol/contracts/common/MetaTransactionWallet.sol
@@ -198,7 +198,7 @@ contract MetaTransactionWallet is
    * @notice Executes multiple transactions on behalf of the signer.`
    * @param destinations The address to which each transaction is to be sent.
    * @param values The CELO value to be sent with each transaction.
-   * @param data The concatenated data to be sent in each transaction.
+   * @param data The concatenated transaction data with their respective lengths <length+data>.
    * @param dataLengths The length of each transaction's data.
    */
   function executeTransactions(
@@ -213,8 +213,16 @@ contract MetaTransactionWallet is
     );
     uint256 dataPosition = 0;
     for (uint256 i = 0; i < destinations.length; i++) {
-      executeTransaction(destinations[i], values[i], sliceData(data, dataPosition, dataLengths[i]));
-      dataPosition = dataPosition.add(dataLengths[i]);
+      require(
+        uint8(data[dataPosition]) == dataLengths[i],
+        "Data length must match specified length"
+      );
+      executeTransaction(
+        destinations[i],
+        values[i],
+        sliceData(data, dataPosition.add(1), dataLengths[i])
+      );
+      dataPosition = dataPosition.add(dataLengths[i].add(1));
     }
   }
 
@@ -227,6 +235,7 @@ contract MetaTransactionWallet is
    */
   function sliceData(bytes memory data, uint256 start, uint256 length)
     internal
+    pure
     returns (bytes memory)
   {
     // When length == 0 bytes.slice does not seem to always return an empty byte array.
--- a/packages/protocol/contracts/common/MetaTransactionWallet.sol
+++ b/packages/protocol/contracts/common/MetaTransactionWallet.sol
@@ -217,6 +217,8 @@ contract MetaTransactionWallet is
         uint8(data[dataPosition]) == dataLengths[i],
         "Data length must match specified length"
       );
+
+      // dataPositions.add(1) accounts for preceding 1 byte signifying length of following data
       executeTransaction(
         destinations[i],
         values[i],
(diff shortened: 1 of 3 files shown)


packages/protocol/contracts/common/interfaces/IMetaTransactionWallet.sol
```diff
@@ -4,12 +4,7 @@ interface IMetaTransactionWallet {
   function setSigner(address) external;
   function setEip712DomainSeparator() external;
   function executeTransaction(address, uint256, bytes calldata) external returns (bytes memory);
-  function executeTransactions(
-    address[] calldata,
-    uint256[] calldata,
-    bytes calldata,
-    uint256[] calldata
-  ) external;
+  function executeTransactions(address[] calldata, uint256[] calldata, bytes calldata) external;
```
ewilz, line 7: oh wow,  I think the linter one-lined this. (I took a param out)

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -198,24 +198,27 @@ contract MetaTransactionWallet is
...
-      "Input arrays must be same length"
-    );
+    require(destinations.length == values.length, "Input arrays must be same length");
+
     uint256 dataPosition = 0;
     for (uint256 i = 0; i < destinations.length; i++) {
-      executeTransaction(destinations[i], values[i], sliceData(data, dataPosition, dataLengths[i]));
-      dataPosition = dataPosition.add(dataLengths[i]);
+      uint256 dataLength = uint256(uint8(data[dataPosition]));
+      // dataPositions.add(1) accounts for preceding 1 byte signifying length of following data
```
m-chrzan, line 213: Typo, `dataPosition`?

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -198,24 +198,27 @@ contract MetaTransactionWallet is
...
-      destinations.length == values.length && values.length == dataLengths.length,
-      "Input arrays must be same length"
-    );
+    require(destinations.length == values.length, "Input arrays must be same length");
+
     uint256 dataPosition = 0;
     for (uint256 i = 0; i < destinations.length; i++) {
-      executeTransaction(destinations[i], values[i], sliceData(data, dataPosition, dataLengths[i]));
-      dataPosition = dataPosition.add(dataLengths[i]);
+      uint256 dataLength = uint256(uint8(data[dataPosition]));
```
m-chrzan, line 212: Is one byte for the length enough?

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -198,24 +198,27 @@ contract MetaTransactionWallet is
    * @notice Executes multiple transactions on behalf of the signer.`
    * @param destinations The address to which each transaction is to be sent.
    * @param values The CELO value to be sent with each transaction.
-   * @param data The concatenated data to be sent in each transaction.
-   * @param dataLengths The length of each transaction's data.
+   * @param data The concatenated transaction data with their respective lengths <length+data>.
```
m-chrzan, line 201: Include exact specification of how this data is laid out (in particular, how many bytes the lengths take up).

m-chrzan (commented):
Have some questions about how this should be implemented.

In general, also wondering how necessary this change is, would love to hear others' thoughts. `executeTransactions` can only be called when all inputs have been signed by the wallet's owner. Sure, a faulty client could generate parameters incorrectly, but that's true of any function. Do these extra checks need to happen onchain?

It would be a critical problem if a malicious submitter could submit only a portion of a signed `executeTransactions` call, but I don't think that was possible in our original scheme, right?

emaG3m:
> a

Yorke and I had the these same thoughts. As I was coding, I put the lengths in both the dataLengths array AND the data as suggested. But this seemed very redundant, so I just moved them to the data param. But if the client was malicious, it could just mess up the lengths in both. Basically we've come to the same conclusion as you, we think the code is fine as is, however, I could check to make sure there's no extra data at the end. Either way, that data would just get thrown out and not be executed.

ewilz:
> Have some questions about how this should be implemented.
> 
> In general, also wondering how necessary this change is, would love to hear others' thoughts. `executeTransactions` can only be called when all inputs have been signed by the wallet's owner. Sure, a faulty client could generate parameters incorrectly, but that's true of any function. Do these extra checks need to happen onchain?
> 
> It would be a critical problem if a malicious submitter could submit only a portion of a signed `executeTransactions` call, but I don't think that was possible in our original scheme, right?

@m-chrzan Yorke and I had the these same thoughts. As I was coding, I put the lengths in both the dataLengths array AND the data as suggested. But this seemed very redundant, so I just moved them to the data param. But if the client was malicious, it could just mess up the lengths in both. Basically we've come to the same conclusion as you, we think the code is fine as is, however, I could check to make sure there's no extra data at the end. Either way, that data would just get thrown out and not be executed.

ewilz:
I will issue a new PR for M03. We will not proceed with the M02 fix, because we have found the check to be redundant.

PR #5032, opened 2020-09-09
[L14] Solidity compiler version is not pinned and it is not the same for all contracts

yorhodes:
### Description

### Other changes

### Tested

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/594

### Backwards compatibility

### DIFF
--- a/packages/protocol/contracts/stability/SortedOracles.sol
+++ b/packages/protocol/contracts/stability/SortedOracles.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.3;
+pragma solidity ^0.5.13;
 
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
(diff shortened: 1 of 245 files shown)


m-chrzan:
Note, this appears to break Certora tests. Can they not handle 0.5.13, or is it just that they expected a different version and need to be explicitly told to expect the new one?

packages/protocol/contracts/stability/SortedOracles.sol
```diff
@@ -1,11 +1,11 @@
-pragma solidity ^0.5.3;
+pragma solidity ^0.5.13;
 
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import "./interfaces/ISortedOracles.sol";
 import "../common/FixidityLib.sol";
 import "../common/Initializable.sol";
-pragma solidity ^0.5.3;
+pragma solidity ^0.5.13;
```
m-chrzan: Why is there a (second) `pragma` statement in the middle of the file here? Lol.

m-chrzan (approved):
What a fascinating diff to look through. Thanks! Please take a look at the Certora issue before merging.

PR #5036, opened 2020-09-10
[N05] Lack of outcome validation 

yorhodes:
### Description

### Other changes

### Tested

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/612

### Backwards compatibility

### DIFF
--- a/packages/protocol/contracts/governance/Validators.sol
+++ b/packages/protocol/contracts/governance/Validators.sol
@@ -508,9 +508,7 @@ contract Validators is
         .multiply(groups[group].slashInfo.multiplier);
       uint256 groupPayment = totalPayment.multiply(groups[group].commission).fromFixed();
       uint256 validatorPayment = totalPayment.fromFixed().sub(groupPayment);
-      IStableToken stableToken = IStableToken(
-        registry.getAddressForOrDie(STABLE_TOKEN_REGISTRY_ID)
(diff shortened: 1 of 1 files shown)


aaronmgdr:
I notice that there have been lots of PRs that haven't filled out the template. Whats the reason for that? Especially as this is an Open source project it seems important that each PR explains what it is for. even if it only fills out a part of the template and removes the rest. 

Also is master the right base for this PR?

yorhodes:
> I notice that there have been lots of PRs that haven't filled out the template. Whats the reason for that? Especially as this is an Open source project it seems important that each PR explains what it is for. even if it only fills out a part of the template and removes the rest.
> 
> Also is master the right base for this PR?

master is the wrong base, thanks
I will make an effort to fill out the templates in the future, though some of these changes contain potentially sensitive information from the smart contract audit
hence the link to the clabs repo issues

m-chrzan (approved):
I think it's fine to include this just for completeness, although it's worth noting that these functions only ever return `true` if they return anything - they revert on a `require` otherwise.

yorhodes:
> I think it's fine to include this just for completeness, although it's worth noting that these functions only ever return `true` if they return anything - they revert on a `require` otherwise.

The specific revert messages are nice I think

PR #5041, opened 2020-09-10
L10: Missing Docstrings

ewilz:
### Description

Add missing docstrings to contracts

### Other changes

N/A

### Tested

not related to tests
### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/601

### Backwards compatibility

yup

### DIFF
--- a/packages/protocol/contracts/governance/LockedGold.sol
+++ b/packages/protocol/contracts/governance/LockedGold.sol
@@ -278,6 +278,11 @@ contract LockedGold is
     return (values, timestamps);
   }
 
+  /**
+   * @notice Returns the number of pending withdrawals from unlocked gold for an account.
+   * @param account The address of the account.
+   * @return total number of pending withdrawals
+   */
   function getTotalPendingWithdrawals(address account) external view returns (uint256) {
     uint256 pendingWithdrawalSum = 0;
     PendingWithdrawal[] memory withdrawals = balances[account].pendingWithdrawals;
--- a/packages/protocol/contracts/governance/LockedGold.sol
+++ b/packages/protocol/contracts/governance/LockedGold.sol
@@ -279,9 +279,9 @@ contract LockedGold is
   }
 
   /**
-   * @notice Returns the number of pending withdrawals from unlocked gold for an account.
+   * @notice Returns the total amount to withdraw from unlocked gold for an account.
    * @param account The address of the account.
-   * @return total number of pending withdrawals
+   * @return Total amount to withdraw.
    */
   function getTotalPendingWithdrawals(address account) external view returns (uint256) {
     uint256 pendingWithdrawalSum = 0;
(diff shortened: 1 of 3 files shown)


ewilz:
ready for merge if it looks good! I fixed those mentioned in the audit and found some extras. We should give a reminder to devs to add any missing docstrings they stumble upon while developing though

packages/protocol/contracts/governance/LockedGold.sol
```diff
@@ -277,6 +277,11 @@ contract LockedGold is
     return (values, timestamps);
   }
 
+  /**
+   * @notice Returns the number of pending withdrawals from unlocked gold for an account.
+   * @param account The address of the account.
+   * @return total number of pending withdrawals
```
m-chrzan: Wording: I think this should be worded in terms of "amount to withdraw" rather than "number of withdrawals" - this is summing an amount of CELO I believe.

packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -90,6 +90,8 @@ contract StableToken is
    * @param registryAddress Address of the Registry contract.
    * @param inflationRate weekly inflation rate.
    * @param inflationFactorUpdatePeriod how often the inflation factor is updated.
+   * @param initialBalanceAddresses array of addresses with an initial balance.
+   * @param initialBalanceValues array of balance values corresponding to initialBalanceAddresses.
```
m-chrzan: Nit: looks like there's some leftover inconsistency (in this natspec and potentially throughout the file) - have param descriptions start with capital letter, end with period.

packages/protocol/contracts/stability/StableToken.sol
```diff
@@ -392,6 +394,7 @@ contract StableToken is
 
   /**
    * @notice Returns the units for a given value given the current inflation factor.
+   * @param inflationFactor the current inflation factor.
```
m-chrzan: Ditto.

packages/protocol/contracts/governance/LockedGold.sol
ewilz: right. I misunderstood the function

m-chrzan (approved):

yorhodes (approved):

ewilz:
It seems on all my PR's I get the same beforeEach hook failing but for a different test:

```
 1) Contract: Attestations
       "before each" hook for "should revert with a non-requested issuer":
     Error: Invalid JSON RPC response: ""
```

```
1) Contract: Attestations
       "before each" hook for "should delete the unselected request":
     Error: Invalid JSON RPC response: ""
```

```
  1) Contract: Attestations
       "before each" hook for "should revert an already completed request":
     Error: Invalid JSON RPC response: ""
```

```
  1) Contract: Attestations
       "before each" hook for "should emit the AttestationExpiryBlocksSet event":
     Error: Invalid JSON RPC response: ""
```
@yorhodes @m-chrzan  what do you think of merging PR's when this is the only failure (or are they run out of order and the first one fails)? Have we changed anything on our branch to affect this `beforeEach` or is this common?

```
        beforeEach(async () => {
          const requestBlockNumber = await web3.eth.getBlockNumber()
          await random.addTestRandomness(requestBlockNumber + selectIssuersWaitBlocks, '0x1')
        })
```

PR #5042, opened 2020-09-10
[Wallet] Remove fee calculation for CELO withdrawals

gnardini:
### Description

The fee estimation was being done using cUSD so users were seeing a "Insufficient balance" error if they didn't have a cUSD balance even if the withdrawal was possible anyways.

In the exchange flow the fee is hard coded as a 0, we don't have fee estimation calculation for CELO yet, so I switched the calculation for a hard coded 0 here as well for now.

### Tested

In the simulator

I'll add e2e tests soon for an account without cUSD balance that will cover this case. Depends on this PR: https://github.com/celo-org/celo-monorepo/pull/4989


packages/mobile/src/exchange/WithdrawCeloSummary.tsx
```diff
@@ -29,7 +24,8 @@ export default function WithdrawCeloSummary({
   recipientAddress,
 }: WithdrawCeloProps) {
   const { t } = useTranslation(Namespaces.exchangeFlow9)
-  const currentAccount = useSelector(currentAccountSelector)
+  // TODO: Estimate real fee.
+  const fee = new BigNumber(0)
```
tarikbellamine, line 28: Could we instead have this be a small non-zero number? My thinking is it's better to overestimate rather than underestimate when showing the user costs. I think this would also be good because it will ensure the user has enough CELO to pay for gas.

tarikbellamine (approved):
LGTM - left a small suggestion but could see it going either way.

packages/mobile/src/exchange/WithdrawCeloSummary.tsx
gnardini: I'll preface this by saying that I don't fully understand how the fees are calculated, but:

- This number is for showing on the UI only, not the real one. I tried using the `max` number and there are no fee issues. Not sure why. Withdrawing without cUSD balance also works.
- If we put something non-zero here, it would still be rounded to 0 since we only show three decimals, and whatever fee is charged is lower than that. So leaving it as 0 seems fine to me. 
- Also, the buy/sell CELO flow has a 0 too and I don't think we've received any complaints or issues from that.

tarikbellamine: I get what you mean - the number is only an estimate for display and doesn't impact the actual transaction at all. If the buy/sell CELO flow is 0 as well, I think leaving it at 0 here makes sense. We can keep it consistent for now and change all of them if/when we decide to.

Lss-Ankit:
@tarikbellamine I have verified this issue on latest test flight build (25) & Android internal build Android internal build 1.2.0(1004294313) found that user is able to withdraw CELO even though user 0 cUSD. 
Also, observed that esitmation fee is remain 0 while withdrawing CELO 
Device: iPhone 11 (14iOS), Samsung Galaxy A70(9.0)

Please let me know if you want to test any thing else.

PR #5043, opened 2020-09-10
upgrade React-native-web to 13

### LINKED ISSUE #3005: upgrade RNW to 0.13.x
Upgrade react-native-web to latest version 0.12.2 

https://www.npmjs.com/package/react-native-web/v/0.12.2

aaronmgdr:
### Description

upgrades react-native-web from 0.11 to 0.13. 

### Other changes

only made changes nessesary to get the tests and functionality to work 

* remove the custom `onPress` testing fireEvent as `fireEvent.click` now works 

* Changed what Form onInput expects from a native event to a custom`{name, value}` as `nativeEvent` no longer included the input `name` nor was it to be found  elsewhere on the event.

 

### Tested

ran the tests, fixed the test

went to every page and interacted, filled out forms. etc. 

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/3005

### Backwards compatibility
should be


packages/web/src/about/VideoCover.tsx
```diff
@@ -1,5 +1,11 @@
 import * as React from 'react'
-import { createElement, Image, StyleSheet, View, ViewStyle } from 'react-native'
+import {
+  Image,
+  StyleSheet,
+  unstable_createElement as createElement,
```
ErinHales, line 5: Should we be using `createElement` anymore if it is now titled `unstable_createElement`?  Is there a new alternative?

ErinHales:
Also, noticed `Pin` is now black instead of white in `/validators/explore`.  Probably missed some styles when you were updating there :)
Screen Shot 2020-09-15 at 5 55 30 PM

packages/web/src/about/VideoCover.tsx
aaronmgdr: there isnt. and funnily enough they still use it in the docs for things like this.

ErinHales: Weird, ok then 😂

ErinHales:
Error when starting up the site
Screen Shot 2020-09-17 at 3 04 28 PM

ErinHales (approved):
My bad! Didn't run yarn, looks good!

PR #5057, opened 2020-09-10
[Wallet] Add e2e tests for account with no balance

gnardini:
### Description

Added a secondary test file that runs test logging in as a new account. (ie. creating an account instead of importing it). This test goes through onboarding, sets up the account key and checks that the buy button is not visible because of the lack of funds.

### Other changes

- Added a `-t` option to the `run_e2e` script to run a single test file.
- Updated the README


gnardini:
Merged this to https://github.com/celo-org/celo-monorepo/pull/4989

PR #5062, opened 2020-09-11
[Wallet] Fix unable to send transactions after upgrading to 1.1.0 on Android (internal)

### LINKED ISSUE #5061: Unable to send transactions after upgrading to 1.1.0 on Android (internal)
### Expected Behavior

Users should be able to send transactions successfully after upgrading from 1.0.0 or 1.0.1 to 1.1.0 on Android and iOS.

### Current Behavior

Issue was reported by @codyborn on Android

> I'm seeing consistent errors when sending funds in cUSD and CELO even for accounts that I've sent to previously. What info can I provide to help debug?
> 
> 

Logs contained the following:
```
stableToken/saga :: Error transfering token :: Could not find address 0x405... in Error: Could not find address 0x4052...
    at anonymous (address a
```

### Steps to reproduce

On Android:
1. start fresh, no valora installed
1. install 1.0.1 (version code 1004294305)
1. restore account I have
1. send tx to self
1. observe the transaction succeeds
1. install 1.1.0 (version code 1004294309)
1. send tx to self
1. observe the error banner saying the transaction failed

Similar steps on iOS don't reproduce the problem so far.

jeanregisser:
### Description

See details in https://github.com/celo-org/celo-monorepo/issues/5061

It depends on https://github.com/celo-org/react-native-geth/pull/33

This is an alternative to the fix in https://github.com/celo-org/celo-monorepo/pull/5060

### Other changes

None.

### Tested

Manually upgrading from 1.0.1 to a build containing this change fixes the original issue.

### Related issues

- Fixes #5061

### Backwards compatibility

Yes for users on 1.0.x
But not for Android users who created a new account on 1.1.0 which was only published internally. They will have to clear they app data and restore their account.
We choose this because we only expect a few internal users to be affected.


gnardini (approved):

packages/mobile/package.json
```diff
@@ -105,7 +105,7 @@
     "react-native-flag-secure-android": "git://github.com/kristiansorens/react-native-flag-secure-android#e234251",
     "react-native-fs": "^2.16.6",
     "react-native-gesture-handler": "^1.6.1",
-    "react-native-geth": "https://github.com/celo-org/react-native-geth#1ba8817",
+    "react-native-geth": "https://github.com/celo-org/react-native-geth#0fcf62f",
```
jmrossy: If you're using this commit be sure to merge your RN-geth PR without squash so it still refers to master. We've made this mistake in the past :)

jmrossy (approved):

jmrossy:
Need to update the yarn lock ;)

packages/mobile/package.json
jeanregisser: Yep! I've merged the PR and updated the sha.

PR #5063, opened 2020-09-11
Fix getting languages on Press page for client side

### LINKED ISSUE #5015: [web] TypeError: Cannot read property 'toLowerCase' of null
It looks like this is related to latest changes on press page. based on when this started showing up. 

Sentry Issue: [CELO-WEB-S9](https://sentry.io/organizations/celo/issues/1873789876/?referrer=github_integration)

```
TypeError: Cannot read property 'toLowerCase' of null
  at a.value (/_next/static/chunks/pages/press-629e27ca90b35e464e29.js:1:1346)
  at Ji (/_next/static/chunks/framework.c7f9f565e2e122ea20f6.js:1:94421)
  at Gi (/_next/static/chunks/framework.c7f9f565e2e122ea20f6.js:1:94214)
  at Ro (/_next/static/chunks/framework.c7f9f565e2e122ea20f6.js:1:129660)
  at zu (/_next/static/chunks/framework.c7f9f565e2e122ea20f6.js:1:113942)
...
(40 additional frame(s) were not displayed)
```

isabellewei:
### Description

When on client side, `languages` was a null object, it now fetches the correct data.

### Tested

navigated to press page from other page

### Related issues

- Fixes #5015 

### Backwards compatibility

### DIFF
--- a/packages/web/src/press/PressPage.tsx
+++ b/packages/web/src/press/PressPage.tsx
@@ -39,7 +39,7 @@ class PressPage extends React.PureComponent<I18nProps & Props> {
       } else {
         const res = await fetch(`/api/press`)
         press = await res.json()
-        languages = navigator.languages.map((s) => s.substr(0, 2))
+        languages = navigator.languages ? navigator.languages.map((s) => s.substr(0, 2)) : []
       }
       return { press, languages }
     } catch {
--- a/packages/web/src/press/PressPage.tsx
+++ b/packages/web/src/press/PressPage.tsx
@@ -26,17 +26,20 @@ interface Props {
 class PressPage extends React.PureComponent<I18nProps & Props> {
   static async getInitialProps(context) {
     let press = []
-    let languages
+    let languages = []
     try {
       // when run on server import fetching code and run. on client send req to api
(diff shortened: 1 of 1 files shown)


packages/web/src/press/PressPage.tsx
```diff
@@ -26,17 +26,20 @@ interface Props {
...
         press = await getpress.default()
         languages = context.req.headers['accept-language']
+          .toLowerCase()
+          .split(',')
+          .map((s) => s.substr(0, 2))
       } else {
         const res = await fetch(`/api/press`)
         press = await res.json()
-        languages = res.headers.get('accept-language')
+        languages = navigator.languages.map((s) => s.substr(0, 2))
```
aaronmgdr: because `navigator.languages` doesnt have 100% support can you make so even if its undefined this wont break? in that case our languages array would just be empty.

isabellewei: ok! I set it to an empty list if it's undefined, and then on line 56, it'll just use the i18n language instead

aaronmgdr (approved):
🧩

PR #5071, opened 2020-09-12
Migrate authentication logic and export test methods

codyborn:
### Description

- Move common auth logic to common package. Added `authenticationMethod` property check which may be set by the client.
- Export test methods for use elsewhere.

### Tested

Will test when these methods are consumed by the signer/combiner. The logic is mostly the same as before with some minor tweaks.

### Backwards compatibility

Yes


packages/phone-number-privacy/common/src/utils/authentication.ts
```diff
@@ -0,0 +1,73 @@
...
+  const messageSignature = request.get('Authorization')
+  const message = JSON.stringify(request.body)
+  const signer = request.body.account
+  const authMethod = request.body.authenticationMethod
+
+  if (!messageSignature || !signer) {
+    return false
+  }
+
+  if (authMethod && authMethod === AuthenticationMethod.ENCRYPTION_KEY) {
```
codyborn, line 34: This line is new (as well as line 28), everything else is the same as what is in identity.ts

alecps (approved):
Looks good to me!

packages/phone-number-privacy/common/package.json
```diff
@@ -16,10 +16,17 @@
...
+    "@celo/contractkit": "0.4.14-dev",
+    "@celo/utils": "0.1.20-dev",
+    "bignumber.js": "^9.0.0",
+    "elliptic": "6.5.3",
+    "@types/elliptic": "^6.4.12",
+    "firebase-functions": "^3.6.0"
+  },
+  "devDependencies": {
+    "ts-node": "^8.3.0",
+    "typescript": "^3.8.3"
```
jmrossy: We generally avoid putting these here and instead rely on the ones at the monorepo root package.json to ensure we're all using the same version

packages/phone-number-privacy/common/package.json
```diff
@@ -16,10 +16,17 @@
...
   ],
   "dependencies": {
-    "@celo/utils": "^0.1.20-dev",
-    "bignumber.js": "^9.0.0"
+    "@celo/contractkit": "0.4.14-dev",
+    "@celo/utils": "0.1.20-dev",
+    "bignumber.js": "^9.0.0",
+    "elliptic": "6.5.3",
+    "@types/elliptic": "^6.4.12",
+    "firebase-functions": "^3.6.0"
```
jmrossy: Can we avoid a dep on firebase-functions? This will add this dep to the signer which doesn't need it.

packages/phone-number-privacy/common/package.json
```diff
@@ -16,10 +16,17 @@
     "lib/**/*"
   ],
   "dependencies": {
-    "@celo/utils": "^0.1.20-dev",
-    "bignumber.js": "^9.0.0"
+    "@celo/contractkit": "0.4.14-dev",
+    "@celo/utils": "0.1.20-dev",
+    "bignumber.js": "^9.0.0",
+    "elliptic": "6.5.3",
+    "@types/elliptic": "^6.4.12",
```
jmrossy: types should be in dev deps

packages/phone-number-privacy/common/src/index.ts
```diff
@@ -7,3 +7,5 @@ export {
 export { toBool, toNum } from './utils/config-utils'
 export * from './utils/constants'
 export * from './utils/input-validation'
+export * from './test/utils'
+export * from './test/values'
```
jmrossy: This isn't a big deal but it's polluting the 'source' namespace with 'test' stuff. Consider wrapping these under a Test object or something like that.

packages/phone-number-privacy/common/src/utils/authentication.ts
```diff
@@ -0,0 +1,73 @@
+import { ContractKit } from '@celo/contractkit'
+import { AuthenticationMethod } from '@celo/contractkit/lib/identity/odis/query'
+import { AccountsWrapper } from '@celo/contractkit/lib/wrappers/Accounts'
+import { trimLeading0x } from '@celo/utils/lib/address'
+import { retryAsyncWithBackOff } from '@celo/utils/lib/async'
+import { verifySignature } from '@celo/utils/lib/signatureUtils'
+import { ec as EC } from 'elliptic'
+import { Request } from 'firebase-functions'
```
jmrossy: If we're only using this for the type, consider hard-coding in a type def and removing this dependency

jmrossy: Which would also help because the Signer's request type may be slightly different (uses Express's request instead of firebase-function's, which are almost identical though)

packages/phone-number-privacy/common/src/utils/authentication.ts
```diff
@@ -0,0 +1,73 @@
...
+  const authMethod = request.body.authenticationMethod
+
+  if (!messageSignature || !signer) {
+    return false
+  }
+
+  if (authMethod && authMethod === AuthenticationMethod.ENCRYPTION_KEY) {
+    const registeredEncryptionKey = await getDataEncryptionKey(signer, contractKit)
+    if (!registeredEncryptionKey) {
+      logger.info(`Account ${signer} does not have registered encryption key`)
```
jmrossy: logger.warn here ?

jmrossy: Should we return false in this case? Even if they accidentally signed it with the wallet key but declared auth type as DEK, seems like we should reject that request

packages/phone-number-privacy/common/src/utils/authentication.ts
```diff
@@ -0,0 +1,73 @@
...
+    } else {
+      logger.info(`Found DEK ${registeredEncryptionKey} for ${signer}`)
+      try {
+        const key = ec.keyFromPublic(trimLeading0x(registeredEncryptionKey), 'hex')
+        const parsedSig = JSON.parse(messageSignature)
```
jmrossy, line 43: Why do we JSON parse here?

jmrossy (commented):
Mostly looks great! Some minor comments

packages/phone-number-privacy/common/package.json
codyborn: good point.. we're just using `request` which is from express, so we can just use express as a dependency.

packages/phone-number-privacy/common/src/utils/authentication.ts
codyborn: Yeah, that's a good point. Better to fail fast and let the dev know.

packages/phone-number-privacy/common/src/utils/authentication.ts
codyborn: Chatted offline. It looks like the sender is serializing a DER object but it's not clear whether it's a string or not. We'll leave as-is for now since we'll be migrating this at some point to EIP712 standard.

packages/phone-number-privacy/common/src/utils/authentication.ts
codyborn: I changed it to use express instead so it's more compatible with both the combiner and signer, thanks!

jmrossy (approved):
Ship it!

PR #5075, opened 2020-09-14
Revert "[Wallet] Ability to send events from native ios to react"

etuleu:
Reverts celo-org/celo-monorepo#4951 because we are testing a different implementation of the same thing which might be simpler.

Reverting for now until I do more testing.


jeanregisser (approved):
Be sure to indicate why you want this reverted.

PR #5076, opened 2020-09-14
[Wallet] Log app load time using props instead of events

etuleu:
### Description

As per Jean's suggestion this is a different way, maybe simpler, of getting the app load time log.

### Other changes

Remove the event listener since it's unused with this change.

### Tested

Launched app on iOS simulator and android emulator.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


packages/mobile/android/app/src/main/java/org/celo/mobile/MainActivity.java
```diff
@@ -62,9 +62,7 @@ public void onPause() {
 
   @Override
   public void onReactContextInitialized(ReactContext context) {
-    context
-      .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
-      .emit("AppStartedLoading", "" + appStartedMillis);
+    // pass
```
tarikbellamine: Nit: I think it could be helpful to write a little more in the comment to describe the historical context + why it's empty

packages/mobile/android/app/src/main/java/org/celo/mobile/MainActivity.java
```diff
@@ -75,6 +73,14 @@ protected ReactActivityDelegate createReactActivityDelegate() {
       protected ReactRootView createRootView() {
         return new RNGestureHandlerEnabledRootView(MainActivity.this);
       }
+
+      @Override
+      protected Bundle getLaunchOptions() {
+        // this is used to pass props (in this case app start time) to react
```
tarikbellamine: Nit: Capitalize start of comment and "React"

tarikbellamine (approved):
Works on both my iOS and Android devices. Left some very minor comments otherwise this looks good to me!

packages/mobile/android/app/src/main/java/org/celo/mobile/MainActivity.java
etuleu: Actually just realized that we can remove the listener for this since it's unused with this change. Looking at it I also realized that we subscribed and unsubscribed in onPause / onResume instead of onCreate / onDestroy which means we would probably miss some events in the rare case of screen going dark. I don't think that was a huge amount but seems better not having to rely on that.

PR #5081, opened 2020-09-14
N04: Gas optimization possible in ReentrancyGuard.sol

ewilz:
### Description

Use Zeppelin's ReentrancyGuard contract to get better gas optimization. Copy and pasted due to version incompatibility

### Other changes

n/a

### Tested

everything should continue working as expected

- Fixes https://github.com/celo-org/celo-labs/issues/616

### Backwards compatibility

yes


ewilz:
This makes a huge storage change so I must close this. RIP

PR #5091, opened 2020-09-15
CIP8 encryption

nambrot:
### Description

This PR introduces encryption of offchain metadata via CIP8. For a better overview please follow the [design document](https://docs.google.com/document/d/1TPI4hy_TCenvL5EvLMWwEU-WbdfABX5tS5yOuDlPHcc/edit#heading=h.tusf13fjf4du).

As part of the implementation we've included two sample schemas:

- `NameSchema` for encrypting a simple name string to others
- `ProfilePictureSchema` as an example of encrypting binary data.

### Other changes

New methods on our `Wallet` interface:
- `computeSharedSecret: (address: Address, publicKey: string) => Promise<Buffer>` (for now only implemented for the `LocalWallet`)
- `removeAccount: (address: Address) => void`
### Tested

Tests found in the file `packages/contractkit/src/identity/offchain-data-wrapper.test.ts`

### Related issues

N/A

### Backwards compatibility

N/A


packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +26,228 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+  async read(account: string): Promise<Result<DataType, SchemaErrors>> {
+    const rawData = await this.wrapper.readDataFromAsResult(account, this.dataPath)
+    if (!rawData.ok) {
+      return Err(new OffchainError(rawData.error))
+    }
+
+    return this.deserialize(rawData.result)
+  }
+
+  async readEncrypted(
```
nambrot: A consumer should not need to specify encrypted or not, right?

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
```diff
@@ -145,10 +163,102 @@ testWithGanache('Offchain Data', (web3) => {
...
+          throw new Error('Should not get here')
+        } catch (e) {
+          expect(e.message).toEqual(`Could not find address ${readerAddress.toLowerCase()}`)
+        }
+
+        kit.removeAccount(writerAddress)
+      })
+    })
+
+    describe('using keys in the wallet', () => {
```
nambrot: It seems like there should be two separate cases here?
1. The writer has the key accessible, but the reader doesn't
2. Both have their key accessible?

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
```diff
@@ -145,10 +163,102 @@ testWithGanache('Offchain Data', (web3) => {
...
+        kit.addAccount(readerPrivate)
+        kit.addAccount(reader2Private)
+
+        const testname = 'test'
+        const payload = { name: testname }
+
+        const nameAccessor = new NameAccessor(wrapper)
+
+        await nameAccessor.writeEncrypted(payload, writerAddress, readerAddress)
+        const receivedName = await nameAccessor.readEncrypted(writerAddress, readerAddress)
```
nambrot: Like mentioned below, the consumer of the API here shouldn't need to specify whether the read is encrypted or not, right?

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +26,228 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+}
+
+export const writeEncrypted = async (
+  wrapper: OffchainDataWrapper,
+  dataPath: string,
+  data: Buffer,
+  fromAddress: string,
+  toAddress: string
+) => {
+  const [fromPubKey, toPubKey] = await Promise.all([
```
nambrot: Unless I'm mistaken this is pretty critical:

I.e. we should not be using the account addresses to compute the corresponding public keys and thus shared secrets. The reason being that any of the accounts might represent smart contracts and thus are not able to use the corresponding private key. Instead we should always be retrieving an accounts `dataEncryptionKey` with a future improvement being that such key can be registered off-chain

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -55,13 +65,18 @@ export class SimpleSchema<DataType> {
   deserialize(buf: Buffer): Result<DataType, SchemaErrors> {
     return deserialize(this.type, buf)
   }
+  async sign(data: DataType) {
```
nambrot: Does this need to be a public function?

nambrot: In general `SimpleSchema` seems to expose quite a fair number of methods. As the schema is the main interface for developers, it might make sense to slim the interface down to what they would actually use? I.e. just read + write?

packages/cli/src/commands/account/register-data-encryption-key.ts
```diff
@@ -0,0 +1,49 @@
...
+    publicKey: flags.string({
+      required: true,
+      description: 'The public key you want to register',
+    }),
+    force: flags.boolean({ description: 'Ignore metadata validity checks' }),
+  }
+
+  static examples = [
+    'register-metadata --url https://www.mywebsite.com/celo-metadata --from 0x47e172F6CfB6c7D01C1574fa3E2Be7CC73269D95',
+  ]
```
nategraf: Looks like these descriptions need to be updated.

packages/cli/src/commands/account/register-data-encryption-key.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  async run() {
+    // const priv1 = '3fd72bd680c5a2c697bea4035dd39f375b58ddfc6b9e460a3274fb27740259fe'
+    // const pub1 = '0x049ba928b8ee77c08294c6483e2054b9acbddd1aae05d4462c226a1ca6eef15bea0f45af0edd046a007a7ea0037a24fc66e8eee2fe5717ce01d78e137e31329cf2'
+
+    // const priv2 = '92c88a27773d43bf7d3dd2852ca60800248559f4140e33bebf1079e471f630f5'
+    // const pub2 = '0x0411b60fc071f0b038ab4d8388de6788ede47ed8bf9399c425d398606105d1e3391fc2616ac899642f009e6a6de25f3ee1b56b18ff846535498f6f88f40df63010'
```
nategraf: Should these be left here, or are they just leftover from development?

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -61,9 +62,9 @@ export default class OffchainDataWrapper {
       return Err(new NoStorageRootProvidedData())
     }
 
-    const item = (await Promise.all(storageRoots.map((_) => _.readAsResult(dataPath)))).find(
-      (_) => _.ok
-    )
+    const item = (
+      await Promise.all(storageRoots.map((_) => _.readAndVerifySignature(dataPath, buildTypedData)))
+    ).find((_) => _.ok)
```
nategraf: `_` usually means an unused variable, or the `lodash` library. I suggest using a single-letter variable here.
```suggestion
      await Promise.all(storageRoots.map((s) => s.readAndVerifySignature(dataPath, buildTypedData)))
    ).find((s) => s.ok)
```

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +75,60 @@ export default class OffchainDataWrapper {
 
   readDataFrom = makeAsyncThrowable(this.readDataFromAsResult.bind(this))
 
-  async writeDataTo(data: string, dataPath: string) {
+  async writeDataTo(data: Buffer, signature: string, dataPath: string) {
```
nategraf: `OffchainDataWrapper` has access to the wallet, so we might simplify this function a bit by taking the signing address and producing the signature internal to this function. It would be a bit closer to a generic read/write API.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +75,60 @@ export default class OffchainDataWrapper {
...
+    const guessedSigner = guessEIP712TypedDataSigner(typedData, signature)
+    if (eqAddress(guessedSigner, this.account)) {
       return Ok(body)
     }
 
     // The signer might be authorized off-chain
     // TODO: Only if the signer is authorized with an on-chain key
-    return this.readAsResult(`/account/authorizedSigners/${toChecksumAddress(signer)}`)
+    return this.readAndVerifySignature(
+      `/account/authorizedSigners/${toChecksumAddress(guessedSigner)}`,
```
nategraf: Should this be extracted as a schema? Similar to profile names and pictures.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +75,60 @@ export default class OffchainDataWrapper {
...
+    const typedData = await buildTypedData(body)
+    const guessedSigner = guessEIP712TypedDataSigner(typedData, signature)
+    if (eqAddress(guessedSigner, this.account)) {
       return Ok(body)
     }
 
     // The signer might be authorized off-chain
     // TODO: Only if the signer is authorized with an on-chain key
-    return this.readAsResult(`/account/authorizedSigners/${toChecksumAddress(signer)}`)
+    return this.readAndVerifySignature(
```
nategraf: I don't understand this piece of the function. If the signature does not verify against the the primary signer, it returns the contents of the authorized signer file?

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+}
+export type EIP712Schema = Array<{ name: string; type: string }>
+const binaryEIP712Schema: EIP712Schema = [
+  { name: 'path', type: 'string' },
+  { name: 'hash', type: 'string' },
+]
+
+export class SimpleSchema<DataType> {
   constructor(
     readonly wrapper: OffchainDataWrapper,
```
nategraf: Including the `OffchainDataWrapper` in this breaks the idea of a schema a bit. If this is not included, schema objects can be included as library components for my application, but with this as a member variable, the schema can only be constructed at runtime, and counter to what one might assume from the name, it needs access to the wallet and other runtime data.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+
+  private async sign(data: DataType) {
+    const typedData = await buildEIP712TypedData(this.wrapper, this.type, this.dataPath, data)
+    const wallet = this.wrapper.kit.getWallet()
+    return wallet.signTypedData(this.wrapper.self, typedData)
+  }
+
+  async write(data: DataType) {
+    if (!this.type.is(data)) {
+      return
```
nategraf: This should return some indication of an error.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+  [t.number._tag]: 'uint256',
+  [t.boolean._tag]: 'bool',
+}
+export type EIP712Schema = Array<{ name: string; type: string }>
+const binaryEIP712Schema: EIP712Schema = [
+  { name: 'path', type: 'string' },
+  { name: 'hash', type: 'string' },
+]
+
+export class SimpleSchema<DataType> {
```
nategraf: It would be very beneficial to include `tsdoc` comments on this class and methods. It's not exactly a simply object, we we might expect external developers to use it as a basis for their own off-chain storage needs.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+
+  async writeEncrypted(data: DataType, toAddress: string) {
+    if (!this.type.is(data)) {
+      return
+    }
+
+    return writeEncrypted(this.wrapper, this.dataPath, this.serialize(data), toAddress)
+  }
+
+  async writeWithSymmetric(data: DataType, toAddresses: string[], symmetricKey?: Buffer) {
```
nategraf: I'd recommend that any given schema either be a direct asymmetric encryption scheme, or one that uses the message key indirection implemented here. This would simplify the interface of a given scheme, such that there is only one encrypt option, and it would prevent edge cases such as someone created a direct and indirect encryption for the same recipient, which could lead to undefined behavior.

Further, I'd recommend that the schema also encode whether this is an encrypted, or unencrypted message type. It could be problematic if a storage owner write both encrypted and unencrypted payloads for the same data. It's also a bit of a footgun in the case that a developer is using a schema someone else wrote, and doesn't realize the message _should_ be encrypted.

packages/contractkit/src/identity/offchain/schema-utils.ts
nategraf: In the case of data that is stored at a number of paths, such as transaction comments, the fact that the path in included in the schema object may be a bit confusing because it would be constructing a new schema object for every comment your want to read even though all the comments are encoded according to the same "schema". Maybe only I find this to be an odd framing though.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+    return writeEncryptedWithSymmetric(
+      this.wrapper,
+      this.dataPath,
+      this.serialize(data),
+      toAddresses,
+      symmetricKey
+    )
+  }
+
+  async readAsResult(account: string): Promise<Result<DataType, SchemaErrors>> {
```
nategraf: Related to my comment above, this method could be simplified if the schema encode whether or not this message is encrypted.

packages/contractkit/src/identity/offchain/schema-utils.ts
nategraf: Could this be made a bit more concise by making `SimpleSchema` build on `BinarySchema`?

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+export function getDataPath(
+  path: string,
+  sharedSecret: Buffer,
+  senderPublicKey: string,
+  receiverPublicKey: string
+) {
+  const senderPublicKeyBuffer = Buffer.from(trimLeading0x(senderPublicKey), 'hex')
+  const receiverPublicKeyBuffer = Buffer.from(trimLeading0x(receiverPublicKey), 'hex')
+
+  return createHmac('sha3-256', sharedSecret)
```
nategraf: After some additional feedback from Psi we decided to switch to BLAKE2s
```suggestion
  return createHmac('blake2s256', sharedSecret)
```

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
 
-export const readWithSchema = makeAsyncThrowable(readWithSchemaAsResult)
+// label = PRF(ECDH(A, B), A || B || data path)
+// ciphertext path = "/cosmetic path/" || base64(label)
+export function getDataPath(
+  path: string,
+  sharedSecret: Buffer,
+  senderPublicKey: string,
+  receiverPublicKey: string
+) {
```
nategraf: This signature contains a lot of implementation detail. It may be better to avoid exporting it if we can provide a higher-level abstraction.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+  const senderPublicKeyBuffer = Buffer.from(trimLeading0x(senderPublicKey), 'hex')
+  const receiverPublicKeyBuffer = Buffer.from(trimLeading0x(receiverPublicKey), 'hex')
```
nategraf: Although the compressed encoding is the most common, an equivalent key can be represented in uncompressed form. We should either handle that case, or make a note (and possibly an assert check) that the given key is only passed in compressed form.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+
+  private async readEncrypted(account: string) {
+    return readEncrypted(this.wrapper, this.dataPath, account)
   }
 }
 
-export const readWithSchema = makeAsyncThrowable(readWithSchemaAsResult)
+// label = PRF(ECDH(A, B), A || B || data path)
+// ciphertext path = "/cosmetic path/" || base64(label)
+export function getDataPath(
```
nategraf: Following the spec, this function seems to compute the ciphertext label (with the data path as an input).
```suggestion
export function getCiphertextLabel(
```

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+  const [fromPubKey, toPubKey] = await Promise.all([
+    accounts.getDataEncryptionKey(wrapper.self),
+    accounts.getDataEncryptionKey(toAddress),
+  ])
+  const wallet = wrapper.kit.getWallet()
+  const sharedSecret = await wallet.computeSharedSecret(publicKeyToAddress(fromPubKey), toPubKey)
+
+  const computedDataPath = getDataPath(dataPath, sharedSecret, fromPubKey, toPubKey)
+  const encryptedData = Encrypt(
+    Buffer.from(trimPublicKeyPrefix(toPubKey), 'hex'),
```
nategraf: It's assumed here that the public key is in uncompressed form. We should verify that this is correct. `Accounts.sol` mentions that the compressed form is "preferred" but it is not required https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/Accounts.sol#L91

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+    Buffer.from(data)
+  )
+
+  const prefixedDataPath = '/ciphertexts/' + computedDataPath
+  const signature = await signBuffer(wrapper, prefixedDataPath, encryptedData)
+  await wrapper.writeDataTo(encryptedData, signature, prefixedDataPath)
+}
+
+//  file_ciphertext = IV || E(message key, IV, data)
+export const writeEncryptedWithSymmetric = async (
```
nategraf: We probably not export this method given it's interface complexity.
```suggestion
const writeEncryptedWithSymmetric = async (
```

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+) => {
+  // if explicitly passing in a symmetric key, use that.
+  // else check for existing key
+  // otherwise generate new one
+  let key: Buffer
+  if (symmetricKey) {
+    key = symmetricKey
+  } else {
+    const keyResult = await readEncrypted(wrapper, `${dataPath}.enc`, wrapper.self)
+    key = keyResult.ok ? keyResult.result : randomBytes(16)
```
nategraf: It seems like the type of error should matter here. If the result was a `404` error, we should generate a new key, if the result was `500` error, we should abort.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+  }
+
+  const iv = randomBytes(16)
+  const cipher = createCipheriv('aes-128-ctr', key, iv)
+  const payload = Buffer.concat([iv, cipher.update(data), cipher.final()])
+
+  const signature = await signBuffer(wrapper, `${dataPath}.enc`, payload)
+  await wrapper.writeDataTo(payload, signature, `${dataPath}.enc`)
```
nategraf: If we are reusing the existing key and the encrypted message exists (i.e. adding new recipients to an existing message) we do not need to rewrite/re-encrypt the data payload.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+
+  await Promise.all(
+    toAddresses.map(async (toAddress) => writeEncrypted(wrapper, dataPath, key, toAddress))
+  )
+}
+
+const readEncrypted = async (
+  wrapper: OffchainDataWrapper,
+  dataPath: string,
+  senderAddress: string
```
nategraf: One way to simplify this API a bit would be to encode the sender address in the object that encodes the storage tree (e.g. `OffchainDataWrapper`) so it doesn't have to be passed in separately here and the caller has a simpler interface to work with. (I am assuming with this that every storage tree has a single address that writes data to it, so if that is not true, then this would not be valid)

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+    accounts.getDataEncryptionKey(senderAddress),
+  ])
+
+  if (readerPubKey === null) {
+    return Err(new UnavailableKey())
+  }
+  const readerPublicKeyAddress = publicKeyToAddress(readerPubKey)
+
+  if (!wallet.hasAccount(readerPublicKeyAddress)) {
+    return Err(new UnavailableKey())
```
nategraf: It would be very useful to report the address for the key we couldn't access here.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+
+  const iv = randomBytes(16)
+  const cipher = createCipheriv('aes-128-ctr', key, iv)
+  const payload = Buffer.concat([iv, cipher.update(data), cipher.final()])
+
+  const signature = await signBuffer(wrapper, `${dataPath}.enc`, payload)
+  await wrapper.writeDataTo(payload, signature, `${dataPath}.enc`)
+
+  await Promise.all(
+    toAddresses.map(async (toAddress) => writeEncrypted(wrapper, dataPath, key, toAddress))
```
nategraf: We should append `dataPath` with the `.key` suffix to indicate that it is a key, and not the data itself, here. (This is what we have in the spec)

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+  if (!wallet.hasAccount(readerPublicKeyAddress)) {
+    return Err(new UnavailableKey())
+  }
+
+  const sharedSecret = await wallet.computeSharedSecret(readerPublicKeyAddress, senderPubKey)
+
+  const computedDataPath = getDataPath(dataPath, sharedSecret, senderPubKey, readerPubKey)
+
+  const encryptedPayloadPath = `${dataPath}.enc`
+  const encryptedPayloadOrKeyPath = '/ciphertexts/' + computedDataPath
```
nategraf: Instead of prepending the `ciphertexts` prefix each time the path is needed, it may be better to include this in a `getCiphertextPath` function, which take the same args are `getDataPath` does now.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+  private async readEncrypted(account: string) {
+    return readEncrypted(this.wrapper, this.dataPath, account)
   }
 }
 
-export const readWithSchema = makeAsyncThrowable(readWithSchemaAsResult)
+// label = PRF(ECDH(A, B), A || B || data path)
+// ciphertext path = "/cosmetic path/" || base64(label)
+export function getDataPath(
+  path: string,
```
nategraf: We should probably normalize the path given here. If we are assuming URI path syntax, then `foo/bar`, `/foo/bar`, and `/foo//baz/../bar` should all be normalized to the same.

I previously had the following note in the spec, which would be one way of specifying this normalization:
> The path should be normalized to the UTF8 encoding of the absolute URL path component, as specified in RFC 3986 § 3.3. Note that path does not include the storage root, query parameters, or fragment.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+      senderAddress,
+      (buf) => buildBinaryEIP712TypedData(wrapper, encryptedPayloadPath, buf),
+      encryptedPayloadPath
+    ),
+    wrapper.readDataFromAsResult(
+      senderAddress,
+      (buf) => buildBinaryEIP712TypedData(wrapper, encryptedPayloadOrKeyPath, buf),
+      encryptedPayloadOrKeyPath
+    ),
+  ])
```
nategraf: I think we should avoid making these two network calls, and have whether of not the message is directly or indirectly asymmetric encrypted. This could be done either be adding a parameter to this function or by making the direct and indirect cases use different functions.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
   }
-  const serializedData = JSON.stringify(data)
-  await wrapper.writeDataTo(serializedData, dataPath)
-  return
+
+  const payloadOrKey = await wallet.decrypt(readerPublicKeyAddress, encryptedPayloadOrKey.result)
+
+  // encrypted with symmetric key
+  if (encryptedPayload.ok) {
+    const key = payloadOrKey
```
nategraf: Might be useful to add a length check here.

packages/contractkit/src/identity/offchain/schema-utils.ts
```diff
@@ -21,62 +29,399 @@ export class OffchainError extends RootError<SchemaErrorTypes.OffchainError> {
...
+    },
+    primaryType: 'Claim',
+    message: {
+      path,
+      payload: (data as unknown) as EIP712Object,
+    },
+  }
+}
+
+export const buildBinaryEIP712TypedData = async (
```
nategraf: It seems like `buildEIP712TypedData` and `buildBinaryEIP712TypedData` could be combined such that one is a special case of the other. It help reduce the amount of code and simplify the mental model of this module.

nategraf (commented):
Overall I found a couple of places where the behavior may be undefined, and noted those. Additionally I think the abstractions could use some clarification, but the vast majority of the functionality is on-point.

Some general notes:
* Although the data is encoded as `Buffer`, it seems that signatures are encoded as hexadecimal strings throughout this PR. Why is that?
* None of the `write` methods have return values. Because they could fail at a number of steps, especially the network calls, we should return an optional error value.
* View calls are actually quite expensive, including 10 or more network round trips for a simple call. We should prefetch and/or cache the data encryption keys where possible. Additionally, it would be good to allow applications to interact with CIP8 storage without active usage of a blockchain node. This would allow for creating more automation and tooling around CIP8 storage without every tool needing to make requests to a blockchain node, if they can access keys in another way.
* Distinguishing between the two encryption modes as "symmetric" and not symmetric is a bit misleading because the "symmetric" case is still a form of asymmetric encryption. (In fact it's very similar to the internal workings of ECIES) I don't exactly have a better solution, but in this review I refer to them as "direct" and "indirect" encrpption.
* I imagine that a number of developers, many in cLabs but hopefully externally as well, will want to use these APIs. We should add good `TSDoc` annotations that can be included in our reference docs.

packages/contractkit/src/identity/offchain-data-wrapper.ts
AlexBHarley: Unfortunately to construct the signature it needs to know about the data shape. I'd rather do that outside the "storage layer". In `OffchainDataWrapper.readDataFromAsResult` we pass in a function to construct the signature which is of course the other way to do it.

packages/contractkit/src/identity/offchain/schema-utils.ts
AlexBHarley: What if we're updating the picture? Do we want to have the compare logic of two objects done here? I'd prefer (for V1 at least) that we just overwrite every time.

packages/contractkit/src/identity/offchain/schema-utils.ts
AlexBHarley: Yeah, we do specifically want signatures to include the path though so that they can't be shuffled around. I personally am not worried about "constructing a new schema".

packages/utils/src/signatureUtils.ts
```diff
@@ -129,6 +130,20 @@ export function parseSignatureWithoutPrefix(
   throw new Error(`Unable to parse signature (expected signer ${signer})`)
 }
 
+export function guessEIP712TypedDataSigner(typedData: EIP712TypedData, signature: string): string {
+  const dataBuff = generateTypedDataHash(typedData)
```
bowd, line 137: So yeah this looks like we're recovering the /wrong/ way and I'm guessing we ended up with this implementation because that's how the WalletBase was signing. 

We need to make a decision on this topic and I would err towards doing it the regular way. Which means that we would need to either fix this after merging #5311 or wait for that to merged first :-/

Thoughts @nambrot, @AlexBHarley ?

AlexBHarley: The issue is that then, everything done between now and when we fix it is incorrect... 

How much stuff would merging your PR break?

bowd: That's the question -- I wasn't able to track down any usages of the `signTypedData` so I'm not sure what will break. That was an open question I had as well in the PR. 

At the moment I updated it, fixed the tests and also figured out the Ledger has a specialised method for EIP712 and used that in the signer as well. So I think it's in a good spot, but because of the nature of the beast the tests are a bit brittle / irrelevant because basically the tests for the Azure HSM/AWS HSM/Ledger signers mock the actual signing logic so it's basically just testing that the wrappers behave correctly. I guess we would still need to manually verify that signing happens correctly with the respective key managers - but, at least in the HSM case I think it should be ok because we're constructing the message hash in javascript-land, which leaves the Ledger where we're trusting their internal library's implementation.

Anyway coming back to the actual question I'm not sure what would break now by merging that, it looks like, at least in `celo-monorepo` nothing is actually consuming the `signTypedData` logic - except for the MetaTransactionWallet which expects the correct version, but I'm unsure if there are other consumers of ContractKit that will be affected.

One option maybe worth considering would be to expose both signing schemes for the sake of backwards compatibility but I'm worried that it would compound the confusion.

AlexBHarley: +1 on compounding confusion. I'm also under the suspicion no one is using the current implementation. So let's merge.

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
```diff
@@ -223,10 +224,10 @@ testWithGanache('Offchain Data', (web3) => {
       })
 
       it("the writer can encrypt data directly to the reader's dataEncryptionKey", async () => {
-        const nameAccessor = new NameAccessor(wrapper)
-        await nameAccessor.writeEncrypted(testPayload, readerAddress)
+        const nameAccessor = new EncryptedNameAccessor(wrapper)
```
nambrot: This no longer matches the test description right?

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -157,40 +205,45 @@ export const readEncrypted = async (
...
-    const encryptedData = payload.slice(IV_LENGTH)
-    const decipher = createDecipheriv('aes-128-ctr', key, iv)
-
-    return Ok(Buffer.concat([decipher.update(encryptedData), decipher.final()]))
+  if (key.result.length !== KEY_LENGTH) {
+    return Err(new InvalidKey())
   }
 
-  return keyOrPayload
+  const iv = payload.result.slice(0, IV_LENGTH)
```
nambrot: Maybe separate the "cryptography" into a separate function, it seems like different layers of concerns here

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +89,81 @@ export default class OffchainDataWrapper {
...
+    const typedData = await buildTypedData(body)
+    const guessedSigner = guessEIP712TypedDataSigner(typedData, signature)
 
-    if (!signature.ok) {
-      return Err(new InvalidSignature())
+    if (eqAddress(guessedSigner, this.account)) {
+      return Ok(body)
     }
 
-    // TODO: Compare against registered on-chain signers or off-chain signers
```
nambrot, line 106: Remembering this TODO, I believe we will still need this if the account is a smart contract (since it cannot sign itself, not create an `AuthorizedSigner` payload

nambrot (commented):
Just a few small comments

packages/contractkit/src/identity/offchain-data-wrapper.ts
AlexBHarley: I'm not totally sure what to do. How do we check if it's a smart contract?

nategraf: Calling the `getCode` RPC is the simplest method. https://ethereum.stackexchange.com/a/28527

nambrot: I wouldn't check if its a contract or not, I would just check if the account is registered on `Accounts.sol` and if so, check against the authorized signers, similarly to how we do it in metadata: https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/identity/metadata.ts#L56

packages/cli/src/commands/account/register-data-encryption-key.ts
```diff
@@ -0,0 +1,43 @@
+import { flags } from '@oclif/command'
+import { BaseCommand } from '../../base'
+import { newCheckBuilder } from '../../utils/checks'
+import { displaySendTx } from '../../utils/cli'
+import { Flags } from '../../utils/command'
+
+export default class RegisterDataEncryptionKey extends BaseCommand {
+  static description =
+    'Register a data encryption key for an account where users will be able to retieve the metadata file and verify your claims'
```
nategraf: Command description looks like it needs to be updated. Same goes for the `from` flag. We may want to make sure we mention the fact that the data encryption key will be stored on-chain.

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
```diff
@@ -1,63 +1,102 @@
...
+    )
+
+    readerWrapper = new OffchainDataWrapper(readerAddress, kit)
+    reader2Wrapper = new OffchainDataWrapper(reader2Address, kit)
   })
 
-  afterEach(async () => {
+  afterEach(() => {
     fetchMock.reset()
+    kit.removeAccount(writerAddress)
```
nategraf: Related to my comment on `kit.ts`, instead of using `removeAccount` here, it make be best to simply make a new `kit` in the `beforeEach`.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -36,34 +37,48 @@ class NoStorageRootProvidedData extends RootError<OffchainErrorTypes.NoStorageRo
...
+  signer: string
 
-  constructor(readonly self: string, readonly kit: ContractKit) {}
+  constructor(readonly self: string, readonly kit: ContractKit, signer?: string) {
+    this.signer = signer || self
+  }
 
   async readDataFromAsResult(
     account: string,
+    buildTypedData: (x: Buffer) => Promise<EIP712TypedData>,
```
nategraf: Every call site I've seen seems to use the `buildEIP712TypedData` function. Unless there is a compelling reason this needs to be a callback, I would suggest replacing this arg with the data type (i.e. the own that would be passed to `buildEIP712TypedData`) and calling `buildEIP712TypedData` internally.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +89,93 @@ export default class OffchainDataWrapper {
 
   readDataFrom = makeAsyncThrowable(this.readDataFromAsResult.bind(this))
 
-  async writeDataTo(data: string, dataPath: string) {
+  async writeDataTo(
+    data: Buffer,
+    signature: string,
```
nategraf: It may be better to define the signature as a Buffer here instead of a hex string. If we want to do this, it will be much easier to do now than later.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +89,93 @@ export default class OffchainDataWrapper {
...
 class StorageRoot {
-  constructor(readonly account: string, readonly address: string) {}
-
-  // TODO: Add decryption metadata (i.e. indicates ciphertext to be decrypted/which key to use)
-  async readAsResult(dataPath: string): Promise<Result<string, OffchainErrors>> {
-    const data = await fetch(this.address + dataPath)
-    if (!data.ok) {
+  constructor(
+    readonly wrapper: OffchainDataWrapper,
+    readonly account: string,
```
nategraf: Using the Address type here would be good for type safety and clarity of the signature. I was initial unsure if account was a Celo address, or something else.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +89,93 @@ export default class OffchainDataWrapper {
...
+    readonly address: string
```
nategraf: IIUC from the code below, this is a URL. If so, could we change the name to `url` or maybe `root`. From this signature I assumed this was a Celo address, and was confused.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +89,93 @@ export default class OffchainDataWrapper {
...
+  async readAndVerifySignature(
+    dataPath: string,
+    buildTypedData: (x: Buffer) => Promise<EIP712TypedData>
+  ): Promise<Result<Buffer, OffchainErrors>> {
+    let dataResponse, signatureResponse
+
+    try {
+      ;[dataResponse, signatureResponse] = await Promise.all([
+        fetch(this.address + dataPath),
+        fetch(this.address + dataPath + '.signature'),
```
nategraf: We should use `url.resolve` to concatenate the data path with the URL root instead of simple concatenation. https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_url_resolve_from_to

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +89,93 @@ export default class OffchainDataWrapper {
...
+      ])
+    } catch (error) {
       return Err(new FetchError())
```
nategraf: Can we embed the caught error into the `FetchError` object? It would make debugging easier when this happens.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +89,93 @@ export default class OffchainDataWrapper {
...
-    const body = await data.text()
-
-    const signature = await fetch(this.address + dataPath + '.signature')
-
-    if (!signature.ok) {
+    if (!dataResponse.ok) {
+      return Err(new FetchError())
+    }
+    if (!signatureResponse.ok) {
       return Err(new InvalidSignature())
```
nategraf: It seems like this should also a `FetchError` as we never actually validated the signature.

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -74,46 +89,93 @@ export default class OffchainDataWrapper {
...
+        accounts.getVoteSigner(this.account),
+        accounts.getValidatorSigner(this.account),
+        accounts.getAttestationSigner(this.account),
+      ])
+      if (signers.some((signer) => signer === guessedSigner)) {
+        return Ok(body)
+      }
+
+      const authorizedSignerAccessor = new AuthorizedSignerAccessor(this.wrapper)
+      const authorizedSigner = await authorizedSignerAccessor.readAsResult(
```
nategraf: It seems like this may result in an infinite loop because `readAndVerifySignature` is indirectly called by the authorized signature accessor, which will return to here and repeat.

packages/contractkit/src/identity/offchain/accessors/authorized-signer.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+    const dataPath = this.basePath + '/' + toChecksumAddress(signer)
+    const typedData = await buildEIP712TypedData(
+      this.wrapper,
+      dataPath,
+      payload,
+      AuthorizedSignerSchema
+    )
+    const signature = await this.wrapper.kit.getWallet().signTypedData(this.wrapper.self, typedData)
+    await this.wrapper.writeDataTo(Buffer.from(JSON.stringify(payload)), signature, dataPath)
```
nategraf: Errors should be handled here.

packages/contractkit/src/identity/offchain/accessors/name.ts
```diff
@@ -0,0 +1,21 @@
...
+import OffchainDataWrapper from '../../offchain-data-wrapper'
+import { EncryptedSimpleSchema, SimpleSchema } from '../schemas'
+
+const NameSchema = t.type({
+  name: t.string,
+})
+
+export type NameType = t.TypeOf<typeof NameSchema>
+
+export class NameAccessor extends SimpleSchema<NameType> {
```
nategraf: It's unclear to me what "Schema" and "Accessor" and defined to mean in this code. Here is seems that an "Accessor" is implements a "Schema", but there is also a "Schema" right above this that describes a data type and does not satisfy the "Schema" interface defined in `utils.ts`.

packages/contractkit/src/identity/offchain/accessors/name.ts
```diff
@@ -0,0 +1,21 @@
...
+  constructor(readonly wrapper: OffchainDataWrapper) {
+    super(wrapper, NameSchema, '/account/name')
+  }
+}
+
+export class EncryptedNameAccessor extends EncryptedSimpleSchema<NameType> {
```
nategraf: Instead of `EncryptedNameSchema` and `NameSchema`, I would suggest `PrivateNameSchema` and `PublicNameSchema` or something along those lines. In particular, calling out that the unencrypted version is **public**. We want to make sure folks don't accidentally publicly attach their name to their account if they do not mean to. This is also from the fact that, at least in the applications we are creating immediately, encryption is the default rather than being unusual. Same comment applies to the profile picture schema.

packages/contractkit/src/identity/offchain/schemas/binary.ts
```diff
@@ -0,0 +1,58 @@
...
+  readEncrypted,
+  Schema,
+  signBuffer,
+  writeEncrypted,
+} from './utils'
+
+/**
+ * Schema for writing any generic binary data
+ */
+export class BinarySchema implements Schema<Buffer> {
```
nategraf: Related to my other comment, we might want to call this `UnencryptedBinarySchema`.

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+  path: string,
+  sharedSecret: Buffer,
+  senderPublicKey: string,
+  receiverPublicKey: string
+) {
+  const senderPublicKeyBuffer = Buffer.from(trimLeading0x(senderPublicKey), 'hex')
+  const receiverPublicKeyBuffer = Buffer.from(trimLeading0x(receiverPublicKey), 'hex')
+
+  const label = createHmac('blake2s256', sharedSecret)
+    .update(Buffer.concat([senderPublicKeyBuffer, receiverPublicKeyBuffer, Buffer.from(path)]))
```
nategraf: IIRC I made a comment before about normalizing the path. I see the call to `path.normalize` below, but it's possible there was some confusing. I was discussing using it on the `path` input to this function, which is used to form the label here.

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+    .digest('base64')
+  return normalize(join('ciphertexts', label))
```
nategraf: So this may be problematic because `/` is a valid base 64 character. For instance, if the label begins with `/`, or has a repeated `/` in the middle, normalize with break the path by changing `//` to `/`. It also reveals we may already have an issue, with or without normalization, in that Google Storage may interpret repeated `/` incorrectly as it needs to be treated as different than a single `/`. We may need to use and require use of RFC 4648 Section 5 alphabet https://tools.ietf.org/html/rfc4648#section-5, or think of another way to avoid this issue.

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+
+// Assumes that the wallet has the dataEncryptionKey of wrapper.self available
+// TODO: Should check and throw a more meaningful error if not
+
+/**
+ * Encrypts the symmetric key `key` to `toAddress`'s data encryption key and uploads it
+ * to the computed storage path.
+ *
+ * @param wrapper the offchain data wrapper
+ * @param dataPath path to where the encrypted data is stored. Used to derive the key location
```
nategraf: Rather than being the path where the encrypted data is stored, its more of the logical path for the data. (e.g. The encrypted profile picture may be stored at `/profile/picture.jpg.enc`, but the data path is `/profile/picture.jpg`)

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+  const accounts = await wrapper.kit.contracts.getAccounts()
+  const [fromPubKey, toPubKey] = await Promise.all([
+    accounts.getDataEncryptionKey(wrapper.self),
+    accounts.getDataEncryptionKey(toAddress),
+  ])
+  const wallet = wrapper.kit.getWallet()
+  const sharedSecret = await wallet.computeSharedSecret(publicKeyToAddress(fromPubKey), toPubKey)
+
+  const computedDataPath = getCiphertextLabel(`${dataPath}.key`, sharedSecret, fromPubKey, toPubKey)
+  const encryptedData = Encrypt(Buffer.from(trimPublicKeyPrefix(toPubKey), 'hex'), Buffer.from(key))
```
nategraf: It seems like `Buffer.from(key)` may be a typo because `key` should already a `Buffer`.

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+  senderAddress: string
+): Promise<Result<Buffer, SchemaErrors>> => {
+  const accounts = await wrapper.kit.contracts.getAccounts()
+  const wallet = wrapper.kit.getWallet()
+  const [readerPubKey, senderPubKey] = await Promise.all([
+    accounts.getDataEncryptionKey(wrapper.self),
+    accounts.getDataEncryptionKey(senderAddress),
+  ])
+
+  if (readerPubKey === null) {
```
nategraf: Should this be `senderPubKey`? If so, the a check may need to be added for `readerPubKey` before sending it to `publicKeyToAddress`

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+    return Err(new OffchainError(encryptedPayload.error))
+  }
+
+  const payload = await wallet.decrypt(readerPublicKeyAddress, encryptedPayload.result)
+  return Ok(payload)
+}
+
+/**
+ * Reads and decrypts a payload that has been encrypted to your data encryption key. Will
+ * resolving the symmetric key used to encrypt the payload.
```
nategraf: s/resolving/resolve
```suggestion
 * resolve the symmetric key used to encrypt the payload.
```

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+  }
+  if (!key.ok) {
+    return Err(key.error)
+  }
+
+  if (key.result.length !== KEY_LENGTH) {
+    return Err(new InvalidKey())
+  }
+
+  return Ok(
```
nategraf: We should add a minimum length check before this point to make sure that the payload is at least `IV_LENGTH` bytes.

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+    AES128Decrypt(key.result, payload.result.slice(0, IV_LENGTH), payload.result.slice(IV_LENGTH))
```
nategraf: This is inconsistent with what we do for encryption above, which is written using only the Node `crypto` APIs. It does seems to be the same, but is a bit confusing. Id recommend either splitting the `AES128EncryptAndHMAC` function in `ecies.ts`to create an `AES128Encypt` function, or implement this section using the Node API (possibly just copy-pasting the code from `ecies.ts`) to make this match.

packages/contractkit/src/identity/offchain/schemas/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+    }
+  }
+
+  return {
+    types: {
+      EIP712Domain,
+      ...types,
+    },
+    domain: {
+      name: 'CIP8 Claim',
```
nategraf: Is the term "claim" used here as a carry-over from the metadata system? If so, we may want to change it to "message" or "published data"

packages/contractkit/src/identity/offchain/storage-writers.ts
```diff
@@ -1,20 +1,20 @@
...
-  async write(data: string, dataPath: string): Promise<void> {
+  async write(data: Buffer, dataPath: string): Promise<void> {
     return this.writeToFs(data, dataPath)
   }
 
-  protected async writeToFs(data: string, dataPath: string): Promise<void> {
+  protected async writeToFs(data: string | Buffer, dataPath: string): Promise<void> {
     await new Promise((resolve, reject) => {
       const directory = parse(dataPath).dir
       mkdirSync(this.root + directory, { recursive: true })
```
nategraf: If I understand JS Promises, we should use the callback version here to avoid blocking. Additionally, I believe the sync versions here may throw, which is nasty because `reject` will never get called.

packages/contractkit/src/identity/offchain/storage-writers.ts
```diff
@@ -30,9 +30,9 @@ export class LocalStorageWriter extends StorageWriter {
 }
 
 export class GitStorageWriter extends LocalStorageWriter {
-  async write(data: string, dataPath: string): Promise<void> {
+  async write(data: Buffer, dataPath: string): Promise<void> {
     await this.writeToFs(data, dataPath)
-    execSync(`git add ${dataPath.substr(1)}`, {
+    execSync(`git add ${dataPath}`, {
```
nategraf: Using `exec` here and below is a massive footgun, potentially leading to remote execution, in case we ever path untrusted input to these functions. We should use `spawn` instead (without `shell` mode) https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options

packages/contractkit/src/identity/offchain/storage-writers.ts
```diff
@@ -41,12 +41,32 @@ export class GitStorageWriter extends LocalStorageWriter {
...
+    super(local)
+    this.bucket = bucket
+  }
+
+  async write(data: Buffer, dataPath: string): Promise<void> {
+    await this.writeToFs(data, dataPath)
+    execSync(`gsutil cp ${normalize(this.root + '/' + dataPath)} gs://${this.bucket}${dataPath}`, {
+      cwd: this.root,
+    })
+    await new Promise((resolve) => setTimeout(resolve, 2000))
```
nategraf: Why this this here?

packages/contractkit/src/identity/offchain/storage-writers.ts
```diff
@@ -41,12 +41,32 @@ export class GitStorageWriter extends LocalStorageWriter {
...
+
 export class MockStorageWriter extends LocalStorageWriter {
   constructor(readonly root: string, readonly mockedStorageRoot: string, readonly fetchMock: any) {
     super(root)
   }
-  async write(data: string, dataPath: string): Promise<void> {
+  async write(data: Buffer, dataPath: string): Promise<void> {
     await this.writeToFs(data, dataPath)
-    this.fetchMock.mock(this.mockedStorageRoot + dataPath, data)
+    this.fetchMock.mock(this.mockedStorageRoot + dataPath, data, {
```
nategraf: You might want to use `url.resolve` here instead of simple concatenation. Obviously it may not matter as much because its in a test, but still removes a sharp edge.

packages/contractkit/src/kit.ts
```diff
@@ -204,6 +209,11 @@ export class ContractKit {
     this.web3.currentProvider.addAccount(privateKey)
   }
 
+  removeAccount(address: Address) {
```
nategraf: Based on my searching, this is only used for testing. If that is true I don't think we should add it to all of our interfaces.

packages/utils/src/ecdh.ts
```diff
@@ -0,0 +1,18 @@
+import { trimLeading0x } from '@celo/base/lib/address'
+import { createECDH } from 'crypto'
+
+export function computeSharedSecret(privateKey: string, publicKey: string): Buffer {
+  const ecdh = createECDH('secp256k1')
+  ecdh.setPrivateKey(Buffer.from(trimLeading0x(privateKey), 'hex'))
+  return ecdh.computeSecret(Buffer.from(ensurePublicKeyPrefix(publicKey), 'hex'))
+}
+
+function ensurePublicKeyPrefix(publicKey: string): string {
```
nategraf: All of this assumes the key is already in compressed form. We either need to make sure all callers of this function pass in a compressed key, or we compress it ourselves before adding the `04` prefix. It's unclear to me from the `crypto` module docs what encodings the `computeSecret` function will accept, but that may be worth verifying. (Additional context: The `Accounts` contracts does not specify whether users should use compressed or uncompressed forms for their key)

packages/utils/src/signatureUtils.ts
```diff
@@ -129,6 +130,18 @@ export function parseSignatureWithoutPrefix(
   throw new Error(`Unable to parse signature (expected signer ${signer})`)
 }
 
+export function guessEIP712TypedDataSigner(typedData: EIP712TypedData, signature: string): string {
```
nategraf: Its not really a "guess", so to align with Ethereum terminology I would recommend calling this `recoverEIP712TypedDataSigner`

nategraf (commented):
Gave it another pass and I have a number of new comments.

General comments:
* I noticed `packages/contractkit/src/identity/offchain/schemas/custom.ts` is empty. Is that intentional?

packages/contractkit/src/identity/offchain/schemas/utils.ts
AlexBHarley: I think making an `AES128Encrypt` function makes sense, one question, should the new `AES128Encrypt` and `AES128Decrypt` functions take the `iv` parameter? 

The other option is that `AES128Encrypt` always generates a new one, `AES128Decrypt` always looks for the `iv` as the first 16 bytes.

AlexBHarley: I think that makes sense, if not I will revert

packages/contractkit/src/identity/offchain-data-wrapper.test.ts
nambrot: IIRC, the problem is that the account persists on a new kit instance since its on `web3`?

packages/utils/src/ecdh.ts
AlexBHarley: To be clear, the 04 prefix is for uncompressed keys. What we'll do is have an `ensureCompressed` function so whatever is coming back from the blockchain / elsewhere ends us how we need it.

AlexBHarley: Our ECIES `Encrypt` function takes non prefixed uncompressed public keys, and appends the prefix itself, I don't think we want to change that in this PR. So we now have

`ensureCompressed`
`ensureUncompressed`
`trimUncompressedPrefix`
`ensureUncompressedPrefix`

packages/contractkit/src/identity/offchain/schema-utils.ts
AlexBHarley: The route I've gone for now is handling both. We could always require compressed though also.

packages/contractkit/src/identity/offchain/schemas/utils.ts
AlexBHarley: I'm torn, I agree it doesn't make so much sense anymore all the time. But we _could_ be publicly/privately claiming a name/picture, in this case it still makes sense. Generally though (like publishing some private data associated with a transaction) it doesn't make sense

packages/contractkit/src/identity/offchain/schemas/utils.ts
AlexBHarley: I don't know if this concerns us.

How the google storage interface displays "folders" is irrelevant for us. The file `gs://hello/test//test` is just as readable and writeable as `gs://hello/test/test`

packages/contractkit/src/identity/offchain-data-wrapper.ts
AlexBHarley: Hmm so I added another param, `checkOffchainSignatures: boolean`, but I dunno if that's the best way to do this...

nategraf: Another option, to avoid adding another parameter to one of the main methods schema writers may call, to add another function named something like `readUnverifiedDataFromAsResult`. I also don't see a good way to do this without a more major refactor, so either way I think this is fine for version 1.

packages/contractkit/src/identity/offchain/schemas/utils.ts
nategraf: Option 2 is the way to go. the IV is generally considered part of the ciphertext in most crypto abstractions, and it's a detail that doesn't need to be controlled by the user of a high-level library.

packages/contractkit/src/identity/offchain/schemas/utils.ts
nategraf: I would consider a "claim" to be a conceptual subset of "published data" (since a claim must be published, but published data need not be a claim). So I would advocate for calling this something more generic to highlight that we intend for this to be a general purpose storage framework for dApps.

packages/contractkit/src/identity/offchain/storage-writers.ts
```diff
@@ -15,28 +18,20 @@ export class LocalStorageWriter extends StorageWriter {
...
+    await mkdir(join(this.root, directory), { recursive: true })
+    await writeFile(join(this.root, dataPath), data)
   }
 }
 
 export class GitStorageWriter extends LocalStorageWriter {
   async write(data: Buffer, dataPath: string): Promise<void> {
     await this.writeToFs(data, dataPath)
-    execSync(`git add ${dataPath}`, {
+    spawnSync('git', ['add', dataPath], {
```
nategraf, line 30: Thanks for addressing these! 🙏

packages/contractkit/src/identity/offchain/storage-writers.ts
```diff
@@ -1,6 +1,9 @@
-import { execSync } from 'child_process'
-import { mkdirSync, writeFile } from 'fs'
-import { normalize, parse } from 'path'
+import { spawnSync } from 'child_process'
+import { promises } from 'fs'
+import { join, parse } from 'path'
+import { resolve } from 'url'
+
+const { writeFile, mkdir } = promises
```
nategraf, line 6: Did not know this existed! I'll have to use these for now on 🙌

packages/contractkit/src/identity/offchain-data-wrapper.ts
```diff
@@ -57,10 +60,11 @@ export default class OffchainDataWrapper {
...
 
-  async readDataFromAsResult(
-    account: string,
-    buildTypedData: (x: Buffer) => Promise<EIP712TypedData>,
-    dataPath: string
+  async readDataFromAsResult<DataType>(
+    account: Address,
+    dataPath: string,
+    checkOffchainSigners: boolean,
+    type?: t.Type<DataType>
```
nategraf, line 66: Minor suggestion: I imagine the `checkOffchainSigners` will almost always be `true`, so moving it to the last position and providing a default will make caller code slightly more concise.
```suggestion
    type?: t.Type<DataType>,
    checkOffchainSigners: boolean = true,
```

packages/contractkit/src/identity/offchain/schemas/utils.ts
nategraf: Sorry I didn't make this point very clear. I think we have a potential correctness issue in the edge case that the `base64` ciphertext label either starts with a slash, or contains a repeated slash. How this potential issue manifests is largely dependent on the specific handling by the storage provider, so for now just GCS, but in the future we want to support arbitrary implementations.

I realize now that using `base64` is likely a mistake because of this sharp-edge. Instead I think we should change  line 49 here to use `hex`, and back propagate the change to the spec. (IIUC, this should be a one-line change here now, but would be a must bigger issue if we ever wanted to migrate in the future) I think the only downside to using `hex` is that the ciphertext label will be 64 characters instead of 42, but that seems trivial.

nategraf (approved):
Looks good to me! I had one significant comment, about changing the encoding for ciphertext labels to `hex`.

packages/contractkit/src/identity/offchain/utils.ts
```diff
@@ -0,0 +1,350 @@
...
+// label = PRF(ECDH(A, B), A || B || data path)
+// ciphertext path = "/cosmetic path/" || base64(label)
+function getCiphertextLabel(
+  path: string,
+  sharedSecret: Buffer,
+  senderPublicKey: string,
+  receiverPublicKey: string
+) {
+  const senderPublicKeyBuffer = Buffer.from(trimLeading0x(senderPublicKey), 'hex')
+  const receiverPublicKeyBuffer = Buffer.from(trimLeading0x(receiverPublicKey), 'hex')
```
nategraf: Actually one more thing. We need to ensure these are in a consistent encoding, either compressed or uncompressed (If uncompressed we can update the spec) because the on-chain Account does not specify.

packages/contractkit/src/identity/offchain-data-wrapper.ts
AlexBHarley: That doesn't work because you then need to pass a type, this API isn't great I'll give you that. The absence of a type means it's binary data and therefore io-ts cannot help.

With your proposed changes, to pass `checkOffchainSigners` you'd need to pass a type

packages/contractkit/src/identity/offchain/utils.ts
AlexBHarley: ICYMI https://github.com/celo-org/celo-monorepo/pull/5091#discussion_r509317862. I've made it so we support either as per `Accounts.sol`. Do we want to force one or the other?

packages/utils/src/ecdh.ts
```diff
@@ -0,0 +1,48 @@
...
+    'secp256k1',
+    'hex',
+    'hex',
+    'uncompressed'
+  ) as string
+}
+
+export function trimUncompressedPrefix(publicKey: string) {
+  const noLeading0x = trimLeading0x(publicKey)
+  if (noLeading0x.length === 130) {
```
kobigurk: Maybe this heuristic can be transformed into a check that the prefix is `04` and fail otherwise?

packages/utils/src/signatureUtils.ts
```diff
@@ -129,6 +130,21 @@ export function parseSignatureWithoutPrefix(
   throw new Error(`Unable to parse signature (expected signer ${signer})`)
 }
 
+export function recoverEIP712TypedDataSigner(
+  typedData: EIP712TypedData,
+  signature: string
+): string {
+  const dataBuff = generateTypedDataHash(typedData)
+  const { r, s, v } = parseSignatureAsRsv(signature.slice(2))
```
kobigurk, line 138: Keep in mind that signature utils generates vrs signatures: https://github.com/celo-org/celo-monorepo/blob/e22f1c02f1d60bae7f12bee0abe3e81e12d244df/packages/base/src/signatureUtils.ts#L25

packages/contractkit/src/identity/offchain/utils.ts
```diff
@@ -0,0 +1,350 @@
...
+
+  const label = createHmac('blake2s256', sharedSecret)
```
kobigurk, line 34: Nit: we're using sha256 hmac in other places. Any specific reason for the difference?

packages/contractkit/src/identity/offchain/utils.ts
```diff
@@ -0,0 +1,350 @@
...
+ */
+const distributeSymmetricKey = async (
+  wrapper: OffchainDataWrapper,
+  dataPath: string,
+  key: Buffer,
+  toAddress: Address
+): Promise<void | OffchainErrors> => {
+  const accounts = await wrapper.kit.contracts.getAccounts()
+  const [fromPubKey, toPubKey] = await Promise.all([
+    accounts.getDataEncryptionKey(wrapper.self),
```
kobigurk, line 60: What happens if this fails? Do we throw an exception, get undefined and fail later or we get some default value? Or something else? I'm concerned about `toPubKey`

packages/contractkit/src/identity/offchain/utils.ts
```diff
@@ -0,0 +1,350 @@
...
+  symmetricKey?: Buffer
+): Promise<SchemaErrors | void> => {
+  const fetchKey = await fetchOrGenerateKey(wrapper, dataPath, symmetricKey)
+  if (!fetchKey.ok) {
+    return fetchKey.error
+  }
+
+  const iv = randomBytes(16)
+  const payload = AES128Encrypt(fetchKey.result, iv, data)
+  const signature = await signBuffer(wrapper, `${dataPath}.enc`, payload)
```
kobigurk, line 148: I bet we already handle it, though not obvious to me directly - do we care about people "stealing" data? This ciphertext is signed but an attacker can take the ciphertext and replace the signature with their own. 

Specifically I can also set my data encryption key to be w/e I want, if I'm reading correctly: https://github.com/celo-org/celo-monorepo/blob/54197498c3890674f7068aca3b8b39d5d709a3f4/packages/protocol/contracts/common/Accounts.sol#L165

I didn't dive deeper into it since it's probably already thought about by someone, so I'd be happy to understand where it fails.

kobigurk (commented):
The crypto parts LGTM in general, assuming that we trust the crypto package in nodejs.
A few comments and a question that I'd be happy to discuss.

packages/contractkit/src/identity/offchain/utils.ts
AlexBHarley: Advised by @psivesely, I can't personally give any reason why though.

packages/contractkit/src/identity/offchain/utils.ts
AlexBHarley: Fixed. Added a similar check to when we read.

packages/utils/src/signatureUtils.ts
AlexBHarley: Here I wanted to keep it consistent with our `guessSigner` method below. 

Also the ethereumjs-util lib we use for the in memory signer returns rsv signatures.

packages/utils/src/ecdh.ts
AlexBHarley: I wanted this to act like the `trimLeading0x` and `ensureLeading0x` functions. ie, do nothing if it's already correct.

packages/contractkit/src/identity/offchain/utils.ts
AlexBHarley: Yes the storage provider could be malicious and replace data, that is true of anywhere your data is stored that you don't own. In my mind this is alleviated by the fact that when you're requesting Bob's name, you always verify the signature is from his account's public key or an authorized signer, let me know if I'm missing something.

Yes that's also true your DEK can be any buffer >= 33 bytes. The doc says `secp256k1 public key for data encryption. Preferably compressed` which @nategraf has mentioned before. It's true that we could add additional checks to this function but that's out of scope for this PR.

packages/contractkit/src/identity/offchain/utils.ts
AlexBHarley: @nategraf maybe you'd be able to provide some context

packages/utils/src/ecdh.ts
kobigurk: Got it, got the intention. `trimLeading0x` checks for leading `0x` though. Maybe instead we can check for leading `04`?

packages/contractkit/src/identity/offchain/utils.ts
kobigurk: I'm imagining the following:
* I'm another user who wants to "steal" data and present it as my own.
* I take the encrypted blob and replace the signature with my account key.
* I set my data encryption key to be the person I stole it from.

packages/utils/src/ecdh.ts
AlexBHarley: Could a compressed key have a leading `04`? I think they're always `02` or `03` right? I will add an additional check for `04`.

packages/contractkit/src/identity/offchain/utils.ts
nategraf: Yes. My point here is that IIUC, [the caller](https://github.com/celo-org/celo-monorepo/pull/5091/files/3144e537d9279ea17ce4e73daadd7fe90fa1a7c1#diff-e7d55cae43eddd87cf3d520332a28a94c3259940b5d1f8e77355eb1a73d8ab1eR66) of this function is passing in the direct output of the view call to retrieve the DEK. There needs to be a normalization step at some point between retrieving it from the chain and [concatenating it with the shared secret](https://github.com/celo-org/celo-monorepo/pull/5091/files/3144e537d9279ea17ce4e73daadd7fe90fa1a7c1#diff-e7d55cae43eddd87cf3d520332a28a94c3259940b5d1f8e77355eb1a73d8ab1eR35) as input to the MAC.

packages/contractkit/src/identity/offchain/utils.ts
nategraf: Here is the context on the conversation between Psi and I. https://docs.google.com/document/d/1TPI4hy_TCenvL5EvLMWwEU-WbdfABX5tS5yOuDlPHcc/edit?disco=AAAAG7ce8WA

My impression in that it doesn't really matter too much either way. BLAKE2s is more performant, especially on lower-end devices. SHA256 or SHA3 or more standardized, but we already use BLAKE2s anyway in BLS signatures, so there was less of a concern about it not being available to the device.

packages/contractkit/src/identity/offchain/utils.ts
AlexBHarley: @kobigurk are you talking about the situation where a malicious user has write access to storage? I've spent a decent amount of time thinking about this and I think we're still OK. The DEK is not used for signing, the attacker cannot simply replace the signature with theirs. It needs to be done with an authorized signer or account key.

Otherwise there's nothing stopping people from claiming things that aren't true. I believe the case you're thinking of is covered by verifying on-chain.

##### Name claim
- User A claims their name
- X "steals" their name
- B tries to read A's name and see's A
- when they read X's name they also see A. Not the end of the world

##### TX comment
 - User A sends a TX to B with a private off-chain comment 
- X "steals" their TX comment
- B receives a transaction, reads the comment from (`<User A storage root>/<txHash>/comment`) and sees the correct TX comment
- if A decided to read the TX comment from X's storage root they also see the comment data

Do those handle what you're thinking of?

packages/contractkit/src/identity/offchain/utils.ts
AlexBHarley: Ah I see what you're talking about now. Good point. I've used `ensureCompressed` as input to the MAC

packages/utils/src/ecdh.ts
kobigurk: Nope! See section 2.2 in https://tools.ietf.org/html/rfc5480. Uncompressed keys have 04 as a prefix.

packages/contractkit/src/identity/offchain/utils.ts
kobigurk: Yep, I'm less worried about security, more about inconsistency and maintainability, but not in a big way.

packages/contractkit/src/identity/offchain/utils.ts
kobigurk: I think they do! Thanks for walking me through it.

packages/utils/src/ecdh.ts
AlexBHarley: Perfect, I don't think _also_ checking for length 130 is invalid though right?

Issue #5086, opened 2020-09-15
Wallet displays redeemed escrow payments to user


tarikbellamine:
### Expected Behavior
User will not see notifications about redeemable escrow payments if they do not have any.

### Current Behavior
User will see a notification about redeemable escrow payments when in fact they do not have any. This is occurring because `doFetchSentPayments` is not being called for some reason.

PR #5092, opened 2020-09-15
[Wallet] Fixing fatal error when escrow gas estimation fails

tarikbellamine:
### Description
Gas estimation failure throws with a string rather than an error object, which was causing the app the to crash.

This PR just addresses the fatal error. The root cause is that that the `doFetchSentPayments` saga is sometimes getting disrupted by geth connectivity issues. When geth connection is reestablished, the saga is not continued. This results in redeemed escrow payments erroneously showing up as reclaimable for the user, which in turn was causing the fatal error.

It appears to only be happening with `doFetchSentPayments` rather than `tokenFetchFactory`, which implies the issue may be with `getContractKit`.

### Other changes
N/A

### Tested
Yes

### Related issues
- Sentry: https://sentry.io/organizations/celo/issues/1890055944/?environment=co.clabs.valora&project=1250733&query=c.error.message.include

- Partially addresses #5086

### Backwards compatibility
Yes


annakaz (approved):

Lss-Ankit:
@tarikbellamine I have verified this task on latest Android internal build v1.1.0(1004294311) & iOS test flight build v1.1.0(23) and observed the following
- Escrow payment received notification is not shown after pasting invite code. 
- Escrow payment Received notification is shown after verifying phone number. 
Please let me know if i need to verify any thing else? 
Thanks!

tarikbellamine:
@Lss-Ankit I think that is good for now!

PR #5094, opened 2020-09-16
[Wallet] Adding Geth connectivity checks as part of getContractKit

### LINKED ISSUE #5086: Wallet displays redeemed escrow payments to user

### LINKED ISSUE #5262: Ensure Geth is synced before initiating a transaction
### Expected Behavior
We should check that Geth is synced before initiating a transaction.

### Current Behavior
We are not checking if Geth is synced before initiating a transaction. In most instances, we are not even checking if Geth is connected.

This is one cause of the "Transaction was not mined within 50 blocks" error that we repeatedly see on Sentry.

https://sentry.io/organizations/celo/issues/1928730892/?environment=co.clabs.valora&project=1250733&query=is%3Aunresolved+not+mined&statsPeriod=14d

tarikbellamine:
### Description
Follow up PR to #5092

It appears that sometimes Contract Kit methods are getting called when Geth is not connected. This negative impact is often mitigated by our use of `getConnectedAccount` at the start of sagas but we don't do that consistently (see `doFetchSentPayments` as an example).

Rather than calling `getConnectedAccount` at the start of the saga, we should instead make sure that Geth is connected (and wait if it's not) when retrieving, and subsequently calling methods on, Contract Kit.

### Other changes
N/A

### Tested
On devices

### Related issues
- Closes #5086
- Closes #5262
- https://sentry.io/organizations/celo/issues/1970861133/?environment=co.clabs.valora&project=1250733&query=Transaction+was+not+mined+within+50+blocks&statsPeriod=14d

### Backwards compatibility
Yes


packages/mobile/src/geth/saga.ts
```diff
@@ -98,6 +100,21 @@ export function* waitForGethConnectivity() {
   }
 }
 
+export async function waitForGethConnectivityAsync() {
+  let connected = isGethConnectedSelector(store.getState())
+  if (connected) {
+    return
+  }
+
+  while (true) {
```
annakaz, line 113: Can you add some number of retries similar to the waitForContractKit function to prevent getting stuck in this loop? https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/web3/contracts.ts#L126

annakaz (commented):
I like the move towards ensuring contractKit's dependencies (geth) are ready before returning it rather than assuming that it will be only called when it is safe to call it. 

Can we replace getConnectedAccount with getAccount?

In forno mode there's no need to wait for geth connectivity. Would it make sense to only wait for geth connectivity if not in forno?

jmrossy:
This does nicely solve a few problems but I think it may introduce a few new ones.
The contract kit is currently more than for just using contracts. Think Odis utils and other things. Blocking access to the whole kit until Geth is ready may be undesirable in certain cases. I think this problem becomes easier once @gastonponti's kit refactoring PR lands, which will start to break up the kit.

packages/mobile/src/geth/saga.ts
tarikbellamine: I considered doing that but opted to mirror the design after the generator function I was copying here: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/geth/saga.ts#L80

I think it's probably a good idea to have a max number of retries but I wasn't sure if I was missing context for why we had the `while (true)` condition.

tarikbellamine:
@annakaz I think we can probably move from `getConnectedAccount` to `getAccount` if we implement this change. Good point about forno mode...in the meantime we can at least put a conditional clause in `getConnectedAccount` to avoid the check while in forno mode. `getConnectedAccount` is used so widely we should be careful when getting rid of it.

@jmrossy that's a good point and was my biggest concern. `getContractKit` is called in so many places that it was hard to understand all the different use cases. The error I was hoping to solve happens with enough consistency that a few users have reported it but not so many that it needs an immediate resolution imo. Do you think it makes sense to hold on this until @gastonponti's work is done? If so, do you mind linking his PR so I can keep tabs on it?

jmrossy:
@tarikbellamine We may be able to mitigate the concern by having separate `getContractKit` and `getConnectedContractKit` methods. Or, if you prefer, we could pass in a param that defaults to true for if the getter should wait. I think we still need to inspect every place this is called to be sure it's the right thing, not sure of a way around that. 

And here's the PR I'm referring to: #4790

jeanregisser (commented):
@jmrossy made a good point. This is a critical code path and I'd recommend we do what he suggested too.

tarikbellamine:
@jmrossy @jeanregisser I've looked more into it and can't find many instances in the mobile directory where we call `getContractKit` but do not require it to be connected. Can you please point an example or two out to me?

The only place I found is in the [sendInvite generator](https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/invite/saga.ts#L173) where we are using the contractKit's web3 method to create a temporary account address.

packages/mobile/src/geth/saga.ts
```diff
@@ -98,6 +100,21 @@ export function* waitForGethConnectivity() {
   }
 }
 
+export async function waitForGethConnectivityAsync() {
+  let connected = isGethConnectedSelector(store.getState())
```
nategraf: We may want to wait for the node to be synced rather than simply connected.

An example of when this distinction is relevant is if the user has the app in the background for 5+ minutes, then opens the app again and immediately makes a transaction. In this case, the node will become connected again quickly, but it will be at least 60 blocks behind the head. If the transaction is sent immediately, then sync occurs, an error will be reported that the transaction "was not mined within 50 blocks" because the chain head just "jumped" 60 blocks.

There isn't a ready-made redux state entry for this, but the [chain head selector](https://github.com/celo-org/celo-monorepo/blob/94f47c0fa17308273ce30df86964cc88a0f409bb/packages/mobile/src/geth/selectors.ts#L9) provides the data needed, and the [sync status saga](https://github.com/celo-org/celo-monorepo/blob/94f47c0fa17308273ce30df86964cc88a0f409bb/packages/mobile/src/geth/saga.ts#L302) provides an example of how it can be used.

tarikbellamine: What's a realistic upper bound for sync time? I agree with your point and have seen the `not mined within 50 blocks` relatively often in Sentry. Just wondering what the UX implications are for adding more wait time without any visual cues to the user

packages/mobile/src/geth/saga.ts
```diff
@@ -40,6 +42,7 @@ const TAG = 'geth/saga'
 const INIT_GETH_TIMEOUT = 15000 // ms
 const NEW_BLOCK_TIMEOUT = 30000 // ms
 const GETH_MONITOR_DELAY = 5000 // ms
+const GETH_SYNC_TIMEOUT = 30000 // ms
```
tarikbellamine, line 45: @nategraf is this a reasonable timeout value?

packages/mobile/src/geth/saga.ts
```diff
@@ -98,6 +101,76 @@ export function* waitForGethConnectivity() {
...
+    if (blockIsFresh(head)) {
+      return true
+    }
+
+    head = (yield take(Actions.SET_CHAIN_HEAD)).head
+  }
+}
+
+export function* waitForGethSync() {
+  const fornoEnabled = yield select(fornoSelector)
```
bowd, line 123: I think you could remove some of the duplication here and only implement the `async` one and then call into it from the generator. Albeit, the use of race is nice and I'd be sad to see it go :D

bowd (commented):
I get what's happening but somehow I feel like a new tentacle sprouted on my pet octopus. But don't get this wrong, I feel the same about my last PR that dealt with geth init. We need to bring some TLC to this part of the codebase.

Now specifically this seams like a pretty innocuous change, I would imagine only a small % runs will go down this code path. I agree that we might be introducing a delay in operations that don't require a connected contractKit but at the same time I feel like having a way to get a potentially disconnected contractKit instance is a 🦶 🔫 scenario as well.

Therefore my instinct would be to go with the current implementation and carefully evaluate the impact through analytics.

packages/mobile/src/geth/saga.ts
tarikbellamine: That's a good suggestion. After thinking a little more about it, I think I will leave it as is because:
1. The generator's method is more efficient than the async polling method. We can simply wait for the relevant action to dispatch instead of repeatedly polling. Maybe it works similarly under the hood but I think it's safe to assume it's done more efficiently than my simple polling function
2. I also like the race method better ;)

packages/mobile/src/geth/saga.ts
```diff
@@ -98,6 +101,76 @@ export function* waitForGethConnectivity() {
...
+  if (!head) {
+    head = (yield take(Actions.SET_CHAIN_HEAD)).head
+  }
+
+  while (true) {
+    if (blockIsFresh(head)) {
+      return true
+    }
+
+    head = (yield take(Actions.SET_CHAIN_HEAD)).head
```
nategraf, line 118: IIRC the chain head is set to null if Forno is toggled on. We should account for what might happen here if we are in this loop.

packages/mobile/src/geth/saga.ts
```diff
@@ -98,6 +101,76 @@ export function* waitForGethConnectivity() {
...
+  }
+}
+
+export function* waitForGethSync() {
+  const fornoEnabled = yield select(fornoSelector)
+  if (fornoEnabled) {
+    return
+  }
+
+  const { result } = yield race({
```
nategraf, line 128: Why implement a timeout here instead of at the call site? I think there are tradeoffs to both, but I think that having a constant timeout here may lead to issues for cases where the caller is willing to wait for longer periods of time. Maybe a good compromise is to make the timeout configurable.

packages/mobile/src/geth/saga.ts
```diff
@@ -98,6 +101,76 @@ export function* waitForGethConnectivity() {
...
+    result: call(pollGethStatusUntilSynced),
+    timeout: delay(GETH_SYNC_TIMEOUT),
+  })
+
+  if (!result) {
+    ValoraAnalytics.track(GethEvents.geth_sync_timeout)
```
nategraf: Why track this event independently? We should get an error event from whichever flow fails due to this timeout, and we can get sync timing information from the `network_sync_*` events, so it's unclear to me what additional information this event provides.

packages/mobile/src/web3/contracts.ts
```diff
@@ -143,6 +143,7 @@ export function* getContractKit() {
       yield call(initContractKit)
     }
   }
+  yield call(waitForGethSync)
```
nategraf: I don't think is nessasrly the place to block. Some functions of `contractkit`, such as unlocking the wallet, do not require sync and it's useful to be able to do them offline. This is something that gave me trouble when implementing offline transactions PoC for the hackathon, and it's also relevant in cases when we loose temporarily during a transaction send.

packages/mobile/src/geth/saga.ts
tarikbellamine: Yup, I have a check for forno mode upstream in `waitForGethSync` that kicks you out of the flow. I could see it going either place. Do you think it would make more sense in the helper function?

packages/mobile/src/geth/saga.ts
tarikbellamine: That's an interesting idea that I didn't consider. I was thinking about it in a binary way: either you need Contract Kit to be synced or you don't. Which use cases do you think would benefit from configurable timeout durations?

packages/mobile/src/geth/saga.ts
tarikbellamine: I agree...will remove this

packages/mobile/src/web3/contracts.ts
tarikbellamine: What do you think about passing a `waitForSync` parameter to `getContractKit` that defaults to `true` because we do want to wait for sync in the majority of cases?

Open to other suggestions for where this check could go. To the extent possible, I'm trying to avoid changes that will inflate the file count of the diff.

packages/mobile/src/geth/saga.ts
```diff
@@ -98,6 +101,74 @@ export function* waitForGethConnectivity() {
...
+}
+
+function pollGethSyncStatusAsync() {
+  return new Promise(async (resolve, reject) => {
+    const dateLimit = Date.now() + GETH_SYNC_TIMEOUT
+
+    while (Date.now() <= dateLimit) {
+      const head = chainHeadSelector(store.getState())
+      if (!head) {
+        await sleep(500)
```
codyborn: We can make this a constant as well

packages/mobile/src/geth/saga.ts
```diff
@@ -98,6 +101,74 @@ export function* waitForGethConnectivity() {
...
+  if (!result) {
+    throw Error(`Geth failed to sync within ${GETH_SYNC_TIMEOUT / 1000} seconds`)
+  }
+}
+
+function pollGethSyncStatusAsync() {
+  return new Promise(async (resolve, reject) => {
+    const dateLimit = Date.now() + GETH_SYNC_TIMEOUT
+
+    while (Date.now() <= dateLimit) {
```
codyborn, line 142: This polling pattern should be generic and created as a util function if not already

codyborn (approved):
LGTM, added some small requests.

Lss-Ankit:
Hey @gnardini  Can you please provide build details along with testing information to verify this task. Thanks!

gnardini:
Hi @Lss-Ankit this is a behind the scenes thing, so it will be tested by just testing everything else, there's nothing explicit that needs to be done to check this.

Lss-Ankit:
Okay thanks @gnardini for the update.

PR #5100, opened 2020-09-16
[Web] Fix minor bugs on celo.org

### LINKED ISSUE #5093: [Web] Fix minor errors on celo.org
### Expected Behavior

There shouldn't be any errors in the console when launching celo.org

### Current Behavior

There are currently 2 minor errors:
 - [ ] Missing end `<div>` tag
 - [x] `<a>` cannot appear as a descendant of `<a>`
 - [ ] You have not declared a namespacesRequired array on your page-level component: withI18nextTranslation(Home). This will cause all namespaces to be sent down to the client, possibly negatively impacting the performance of your app. For more info, see: https://github.com/isaachinman/next-i18next#4-declaring-namespace-dependencies
Warning: viewport meta tags should not be used in _document.js's <Head>. https://err.sh/next.js/no-document-viewport-meta

The second one should be resolved by changing `MediumLogo` to just the logo instead of a link.  Double check that this doesn't break any other links, but I don't think it should.

ErinHales:
### Description

Solves a few minor bugs and warnings

### Related issues

- Fixes #5093

### Backwards compatibility

Yes


packages/web/src/shared/Progress.tsx
```diff
@@ -98,8 +98,8 @@ const styles = StyleSheet.create({
     backgroundColor: colors.red,
   },
   bar: {
-    transformProperty: 'background-color',
-    transformDuration: '1s',
+    transitionProperty: 'backgroundColor',
```
aaronmgdr: my experience is that these values need to be hyphenated to work. because its passed as is to the actual dom and then references the dom style which has hyphens

aaronmgdr: mind you i guess it was broken before any way

aaronmgdr (approved):
Thank you for fixing these

PR #5101, opened 2020-09-16
adds nexmo applications to attestations-service helm chart

alecps:
### Description

Updates the attestation-service helm chart to support setting a different NEXMO_APPLICATION value for each instance

### Other changes

None

### Tested

Deployed to alfajores without error

### Related issues

NA

### Backwards compatibility

Yes


timmoreton (approved):
LGTM!

PR #5106, opened 2020-09-17
[Wallet] Group activity feed items by date

### LINKED ISSUE #5105: Group items on the Activity Feed by date
### Expected Behavior

Items on the activity feed are grouped by date followign the [Figma](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet-Refresh?node-id=2609%3A4659) design

### Current Behavior

All items show without grouping

gnardini:
### Description

Group items in the activity feed by date.

[Figma link](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet-Refresh?node-id=2609%3A4659)

### Tested

On the simulator

Screen Shot 2020-09-17 at 13 24 03

### Related issues

- Fixes #5105


packages/mobile/src/transactions/TransactionFeed.tsx
```diff
@@ -116,7 +122,23 @@ export class TransactionFeed extends React.PureComponent<Props> {
     }
 
     if (data && data.length > 0) {
-      return <FlatList data={data} keyExtractor={this.keyExtractor} renderItem={this.renderItem} />
+      if (kind === FeedType.HOME) {
+        const sections = groupFeedItemsInSections(data)
```
jeanregisser: I feel we should memoize this instead of recomputing it on each render.
What do you think?

packages/mobile/src/transactions/TransactionFeed.tsx
```diff
@@ -116,7 +122,23 @@ export class TransactionFeed extends React.PureComponent<Props> {
...
+        return (
+          <SectionList
+            renderItem={this.renderItem}
+            renderSectionHeader={this.renderSectionHeader}
+            sections={sections}
+            keyExtractor={this.keyExtractor}
+            initialNumToRender={30}
```
jeanregisser: The default wasn't good enough?

packages/mobile/src/utils/time.ts
```diff
@@ -258,6 +258,12 @@ export const formatFeedDate = (timestamp: number, i18next: i18nType) => {
   return quickFormat(timestamp, i18next, 'MMM d')
 }
 
+export const formatFeedSectionTitle = (timestamp: number, i18next: i18nType) => {
+  const title = quickFormat(timestamp, i18next, 'MMMM yyyy')
+  const currentYear = new Date().getFullYear().toString()
+  return title.endsWith(currentYear) ? title.substring(0, title.length - 5) : title
```
jeanregisser: This is ok for now, but is problematic if we localize this format. For some locales the year isn't at the end.

```
> new Date().toLocaleDateString('fr-FR', {
  year: 'numeric',
  month: 'long',
})
< "septembre 2020"
> new Date().toLocaleDateString('ru-RU', {
  year: 'numeric',
  month: 'long',
})
< "сентябрь 2020 г."
```

Could you add a small note?

jeanregisser (approved):
Awesome! 👍

packages/mobile/src/transactions/TransactionFeed.tsx
gnardini: I copied this from `RecipientPicker` and for some reason decided to leave it in, but I don't see any reason to do so, removed :)

packages/mobile/src/utils/time.ts
gnardini: I thought this wouldn't happen because we're explicitly passing `MMMM yyyy` as a format. Updated the code to use a different format if the years match to avoid problems in the future :)

packages/mobile/src/transactions/TransactionFeed.tsx
gnardini: I made `TransactionFeed` a functional component and memoized this call but I think there are still many extra renders.
 
The problem comes from the `data` prop which is re-created each time in `TransactionsList`.
I think this is good enough for now but at some point we should measure the performance of this screen and decide if we need to optimize the number of renders here.

packages/mobile/src/utils/time.ts
```diff
@@ -300,25 +309,35 @@ export const getLocalTimezone = () => {
   return timezoneGuess.format('zz')
 }
 
+const oneSecondInMillis = 1000
+const oneMinuteInMillis = 60 * oneSecondInMillis
+const oneHourInMillis = 60 * oneMinuteInMillis
+const oneDayInMillis = 24 * oneHourInMillis
```
jeanregisser: nit: should be `ONE_SECOND_IN_...` though again we've been inconsistent in the codebase 😅

packages/mobile/src/transactions/TransactionFeed.tsx
```diff
@@ -103,24 +71,58 @@ export class TransactionFeed extends React.PureComponent<Props> {
...
+    }
+
+    ${ExchangeFeedItem.fragments.exchange}
+    ${TransferFeedItem.fragments.transfer}
+  `,
+}
+
+// TODO: We should do a better comparison of the |data| prop to decide when to re-render.
+// Right now |data| always changes (returns a different ref) which causes many extra renders.
+export default memo(TransactionFeed)
```
jeanregisser: Maybe we shouldn't use `memo` for now then?

jeanregisser (approved):
All good! 👍 
Some (really) optional comments.

packages/mobile/src/transactions/TransactionFeed.tsx
gnardini: I guess that makes sense, I used it because we were using `PureComponent` and wanted to maintain an equivalence to that, but I'll remove it :)

Lss-Ankit:
Hi @tarikbellamine i have verified task using Test flight build v1.2.0 (24) and found the following:

Test scenarios:
- Verify user is able to see the Activity feed in home page 
- Verify user is able to see the activity feed items by date 

Thanks !

Lss-Ankit:
Hi @tarikbellamine i have verified task using Android internal build 1.2.0(1004294313) and found the following

**Device:** Samsung Galaxy A70(9.0)

**Test scenarios:**

- Verify user is able to see the Activity feed in home page
- Verify user is able to see the activity feed items by date

Thanks !

PR #5107, opened 2020-09-17
MetaTransactionWallet upgrades

ewilz:
### Description

M03: Check for unused data in executeTransactions data param
add indexing in events
emit event when receiving ether

### Other changes

added extra tests to multisig

### Tested

new tests for event emissions and any new cases
### Related issues

- Fixes 
https://github.com/celo-org/celo-labs/issues/597 
https://github.com/celo-org/celo-labs/issues/611 
https://github.com/celo-org/celo-labs/issues/600
https://github.com/celo-org/celo-labs/issues/597

### Backwards compatibility

yes.

### DIFF
--- a/packages/protocol/contracts/common/MetaTransactionWallet.sol
+++ b/packages/protocol/contracts/common/MetaTransactionWallet.sol
@@ -27,11 +27,17 @@ contract MetaTransactionWallet is
   uint256 public nonce;
   address public signer;
 
-  event SignerSet(address signer);
+  event SignerSet(address indexed signer);
   event EIP712DomainSeparatorSet(bytes32 eip712DomainSeparator);
-  event TransactionExecution(address destination, uint256 value, bytes data, bytes returnData);
+  event Deposit(address indexed sender, uint256 value);
+  event TransactionExecution(
+    address indexed destination,
+    uint256 value,
+    bytes data,
+    bytes returnData
+  );
   event MetaTransactionExecution(
-    address destination,
+    address indexed destination,
     uint256 value,
     bytes data,
     uint256 nonce,
@@ -41,7 +47,11 @@ contract MetaTransactionWallet is
   /**
    * @dev Fallback function allows to deposit ether.
    */
-  function() external payable {}
+  function() external payable {
+    if (msg.value > 0) {
+      emit Deposit(msg.sender, msg.value);
+    }
+  }
 
   /**
    * @notice Returns the storage, major, minor, and patch version of the contract.
@@ -217,6 +227,7 @@ contract MetaTransactionWallet is
       executeTransaction(destinations[i], values[i], sliceData(data, dataPosition, dataLengths[i]));
       dataPosition = dataPosition.add(dataLengths[i]);
     }
+    require(dataPosition == data.length, "data cannot have extra bytes appended");
   }
 
   /**
--- a/packages/protocol/contracts/common/MetaTransactionWallet.sol
+++ b/packages/protocol/contracts/common/MetaTransactionWallet.sol
@@ -209,23 +209,36 @@ contract MetaTransactionWallet is
    * @param values The CELO value to be sent with each transaction.
    * @param data The concatenated data to be sent in each transaction.
    * @param dataLengths The length of each transaction's data.
+   * @return The return values of all transactions appended as bytes and an array of the length
+   *         of each transaction output which will be 0 if a transaction had no output 
    */
   function executeTransactions(
     address[] calldata destinations,
     uint256[] calldata values,
     bytes calldata data,
     uint256[] calldata dataLengths
-  ) external {
+  ) external returns (bytes memory, uint256[] memory) {
     require(
       destinations.length == values.length && values.length == dataLengths.length,
       "Input arrays must be same length"
     );
+
+    bytes memory returnValues;
+    uint256[] memory returnLengths = new uint256[](destinations.length);
     uint256 dataPosition = 0;
     for (uint256 i = 0; i < destinations.length; i = i.add(1)) {
-      executeTransaction(destinations[i], values[i], sliceData(data, dataPosition, dataLengths[i]));
+      bytes memory returnVal = executeTransaction(
+        destinations[i],
+        values[i],
+        sliceData(data, dataPosition, dataLengths[i])
+      );
+      returnValues = abi.encodePacked(returnValues, returnVal);
+      returnLengths[i] = returnVal.length;
       dataPosition = dataPosition.add(dataLengths[i]);
     }
+
     require(dataPosition == data.length, "data cannot have extra bytes appended");
+    return (returnValues, returnLengths);
   }
 
   /**
--- a/packages/protocol/contracts/common/MetaTransactionWallet.sol
+++ b/packages/protocol/contracts/common/MetaTransactionWallet.sol
@@ -40,7 +40,7 @@ contract MetaTransactionWallet is
     address indexed destination,
     uint256 value,
     bytes data,
-    uint256 nonce,
+    uint256 indexed nonce,
     bytes returnData
   );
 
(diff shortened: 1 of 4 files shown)


packages/protocol/test/common/multisig.ts
```diff
@@ -40,6 +40,32 @@ contract('MultiSig', (accounts: any) => {
     })
   })
 
+  describe('fallback function', () => {
```
ewilz, line 48: zeppelin suggested emulating multisig event emission for consistency, added tests to the new case and went in and added these same tests for existing multisig functionality

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -27,11 +27,17 @@ contract MetaTransactionWallet is
...
+    uint256 value,
+    bytes data,
+    bytes returnData
+  );
   event MetaTransactionExecution(
-    address destination,
+    address indexed destination,
     uint256 value,
     bytes data,
     uint256 nonce,
```
yorhodes: do we think nonce would be useful as an indexed parameter?

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -201,22 +209,36 @@ contract MetaTransactionWallet is
    * @param values The CELO value to be sent with each transaction.
    * @param data The concatenated data to be sent in each transaction.
    * @param dataLengths The length of each transaction's data.
+   * @return The return values of all transactions appended as bytes and an array of the length
```
yorhodes, line 212: amazing! contract composability ftw

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -201,22 +209,36 @@ contract MetaTransactionWallet is
...
+    uint256[] memory returnLengths = new uint256[](destinations.length);
     uint256 dataPosition = 0;
     for (uint256 i = 0; i < destinations.length; i = i.add(1)) {
-      executeTransaction(destinations[i], values[i], sliceData(data, dataPosition, dataLengths[i]));
+      bytes memory returnVal = executeTransaction(
+        destinations[i],
+        values[i],
+        sliceData(data, dataPosition, dataLengths[i])
+      );
+      returnValues = abi.encodePacked(returnValues, returnVal);
```
yorhodes, line 235: this is basically an append right?
would it be useful to ensure that `returnValues.length before == returnValues.length after - returnVal.length` as we expect?

packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -300,24 +339,80 @@ contract('MetaTransactionWallet', (accounts: string[]) => {
...
+            const returnValues = await wallet.executeTransactions.call(
               transactions.map((t) => t.destination),
               transactions.map((t) => t.value),
               ensureLeading0x(transactions.map((t) => trimLeading0x(t.data)).join('')),
               transactions.map((t) => trimLeading0x(t.data).length / 2),
               { from: signer }
             )
+
+            assert.equal(
+              returnValues[0],
```
yorhodes, line 386: can we add some comments about indices 0 and 1 corresponding to the order of `transactions.push`
or alternatively using a normal array init instead of `push`

packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -300,24 +339,80 @@ contract('MetaTransactionWallet', (accounts: string[]) => {
...
+              // return values are padded to equal 32 bytes
+              `0x${signerAddrReturned.padStart(64, '0')}${boolTrue.padStart(64, '0')}`
+            )
+            for (let i in returnValues[1]) {
+              // tx's with index 3 and 4 have return values
+              if (parseInt(i) == 3 || parseInt(i) == 4) {
```
yorhodes: can we initialize these constants where we initialize the transactions

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -41,7 +47,9 @@ contract MetaTransactionWallet is
   /**
    * @dev Fallback function allows to deposit ether.
    */
-  function() external payable {}
+  function() external payable {
+    if (msg.value > 0) emit Deposit(msg.sender, msg.value);
```
yorhodes, line 51: 🤑

yorhodes (approved):
a few comments but otherwise LGTM!

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -201,22 +209,36 @@ contract MetaTransactionWallet is
    * @param values The CELO value to be sent with each transaction.
    * @param data The concatenated data to be sent in each transaction.
    * @param dataLengths The length of each transaction's data.
+   * @return The return values of all transactions appended as bytes and an array of the length
+   *         of each transaction output which will be 0 if a transaction had no output 
    */
   function executeTransactions(
```
yorhodes, line 215: just had a random thought
we could have 2 versions of this function or a wrapper function which saves on gas by not building the return values array
lmk if you think that would actually be useful or if the gas discrepancy is insignificant

packages/protocol/contracts/common/MetaTransactionWallet.sol
ewilz: can you remind me the role of a nonce here?

packages/protocol/contracts/common/MetaTransactionWallet.sol
ewilz: Yes it appends the second argument onto the first.  the length of `returnValues` is actually pretty unpredictable since some transactions do not return anything

packages/protocol/test/common/metatransactionwallet.ts
ewilz: I initialized the array without the push (which seems to make more sense anyway). Not sure if this addresses this comment?

packages/protocol/test/common/metatransactionwallet.ts
ewilz: I assigned these numbers to constants within the test which refer to what these numbers actually mean. This adds to the readability. Let me know if this addresses your concerns

packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -300,24 +346,81 @@ contract('MetaTransactionWallet', (accounts: string[]) => {
 
       describe('when the caller is the signer', () => {
         describe('when the transactions are executed directly', () => {
-          beforeEach(async () => {
-            await wallet.executeTransactions(
+          describe('', () => {
```
ewilz, line 349: assigning the `beforeEach` to only a subset of tests here

packages/protocol/contracts/common/MetaTransactionWallet.sol
ewilz: with today's ethereum prices, it would add 10 cents onto the $3.80 transaction in our test. So obv .....half the gas price 5 cents.  Lol where can I find the celo gas prices at the moment?? I think it's fiine,  but maybe a feature we can add if it's needed?

packages/protocol/contracts/common/MetaTransactionWallet.sol
yorhodes: differentiate sequence of transactions signed

packages/protocol/contracts/common/MetaTransactionWallet.sol
yorhodes: sgtm

packages/protocol/test/common/metatransactionwallet.ts
yorhodes: yeah that helps
I was also referring to the `returnValues[0]`, `returnValues[1]` "magic indices"

ewilz: ohhhh. I think it's just how web3 handles multiple arguments being returned, since this function returns 2 args. `[0]` is the actual data, and `[1]` is the lengths array

PR #5108, opened 2020-09-17
Bring FAQ to Contentful 

### LINKED ISSUE #4803: Contentful Part 2
Make at least one celo.org page that isn't a kit run off Contentful.

aaronmgdr:
### description
replaces FAQ where content was kept with code with a version where content is fetched from contentful.com. This will allow it to be kept up to date more easily. 

### Other changes
Due to option chaining syntax `?.` causing jest to fail I ended up changing the target for the tscofig.jest.ts to be es2016

### Tested

updated the test

### Related issues

- Fixes #4803 

### Backwards compatibility

yes


dominiquejane (approved):
Looks good!

PR #5112, opened 2020-09-17
Display currency of payment request without changing local currency settings

### LINKED ISSUE #4869: Currency Field in QR Code should not permanently change the local currency setting
### Expected Behavior

localCurrency field should set the transaction amount to the local currency indicated, but outside of this transaction the user's pre-set local currency should not change

### Current Behavior

scanning a QR code in with localCurrency=PHP changes app local currency setting to PHP going forward

Impact: currently this QR/Deeplink field is only used in an isolated case, this is much more serious once the localCurrency parameter is more widely adopted

isabellewei:
### Description

The local currency setting was being changed when a QR code/deeplink was used. Refactored components to use a `currencyInfo` parameter if passed instead of defaulting to using the local currency from state.

### Other changes

None

### Tested

Used a deeplink for a transaction in a different currency from my app currency.

### Related issues

- Fixes #4869 

### Backwards compatibility


packages/mobile/src/send/utils.ts
```diff
@@ -188,34 +184,32 @@ export function* handleSendPaymentData(
...
-    }
-  }
-
-  if (data.amount) {
+  if (data.amount && data.currencyCode) {
     // TODO: integrate with SendConfirmation component
-    const exchangeRate = yield select(getLocalCurrencyExchangeRate)
-    const amount = convertLocalAmountToDollars(data.amount, exchangeRate)
-    if (!amount) {
+    // const exchangeRate = yield select(getLocalCurrencyExchangeRate)
```
tarikbellamine: Please delete if no longer needed

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -55,6 +55,11 @@ import { fetchDollarBalance } from 'src/stableToken/actions'
 import Logger from 'src/utils/Logger'
 import { currentAccountSelector, isDekRegisteredSelector } from 'src/web3/selectors'
 
+export interface CurrencyInfo {
+  currencyCode: LocalCurrencyCode
+  exchangeRate?: string | null
```
tarikbellamine: When is exchange rate `null` or `undefined`? It looks like the `fetchExchangeRate` async function always resolves to a `string`?

packages/mobile/src/navigator/types.tsx
```diff
@@ -181,6 +182,7 @@ export type StackParamList = {
     transactionData: TransactionDataInput
     addressJustValidated?: boolean
     isFromScan?: boolean
+    currencyInfo?: CurrencyInfo
```
tarikbellamine, line 185: Why is this optional? Won't it always be defined?

packages/mobile/src/components/TotalLineItem.v2.tsx
```diff
@@ -11,16 +11,23 @@ import { CURRENCIES, CURRENCY_ENUM } from 'src/geth/consts'
 import { Namespaces } from 'src/i18n'
 import { LocalCurrencyCode } from 'src/localCurrency/consts'
 import { useExchangeRate, useLocalCurrencyCode } from 'src/localCurrency/hooks'
+import { CurrencyInfo } from 'src/send/SendConfirmation'
 
 interface Props {
   title?: string
   amount: MoneyAmount
   hideSign?: boolean
+  currencyInfo?: CurrencyInfo
```
tarikbellamine, line 20: Why is this optional?

packages/mobile/src/components/TotalLineItem.v2.tsx
```diff
@@ -11,16 +11,23 @@ import { CURRENCIES, CURRENCY_ENUM } from 'src/geth/consts'
...
-  const localCurrencyCode = useLocalCurrencyCode()
-  const localCurrencyExchangeRate = useExchangeRate()
+export default function TotalLineItem({ title, amount, hideSign, currencyInfo }: Props) {
+  let localCurrencyCode = useLocalCurrencyCode()
+  let localCurrencyExchangeRate = useExchangeRate()
+  if (currencyInfo) {
+    localCurrencyCode = currencyInfo.currencyCode
+    localCurrencyExchangeRate = currencyInfo.exchangeRate
+  }
+
```
tarikbellamine: What do you think about simplifying the code a bit and defining these variables by using a ternary like you did elsewhere?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -357,8 +364,18 @@ export class SendConfirmation extends React.Component<Props, State> {
             feeLoading={asyncFee.loading}
             feeHasError={!!asyncFee.error}
             totalFee={fee}
+            currencyInfo={{
```
tarikbellamine: Consider defining `currencyInfo` once up top then referencing that variable throughout these prop assignments

packages/mobile/src/send/utils.ts
```diff
@@ -188,34 +184,32 @@ export function* handleSendPaymentData(
   }
   yield put(storeLatestInRecents(recipient))
 
-  if (data.currencyCode) {
```
tarikbellamine, line 192: So this logic within the conditional was the root of the problem, right?

packages/mobile/src/send/utils.ts
```diff
@@ -188,34 +184,32 @@ export function* handleSendPaymentData(
...
-
-  if (data.amount) {
+  if (data.amount && data.currencyCode) {
     // TODO: integrate with SendConfirmation component
-    const exchangeRate = yield select(getLocalCurrencyExchangeRate)
-    const amount = convertLocalAmountToDollars(data.amount, exchangeRate)
-    if (!amount) {
+    // const exchangeRate = yield select(getLocalCurrencyExchangeRate)
+    const currency = data.currencyCode as LocalCurrencyCode
+    const exchangeRate = yield call(fetchExchangeRate, currency)
```
tarikbellamine: Can you please define the type of `exchangeRate`? I believe it's a string but `convertLocalAmountToDollars` expects a BigNumber

packages/mobile/src/components/FeeDrawer.tsx
```diff
@@ -26,6 +27,7 @@ interface Props {
   feeHasError?: boolean
   totalFee?: BigNumber
   testID?: string
+  currencyInfo?: CurrencyInfo
```
tarikbellamine, line 30: Why is this optional?

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -131,9 +133,14 @@ export default function CurrencyDisplay({
...
+  currencyInfo,
 }: Props) {
-  const localCurrencyCode = useLocalCurrencyCode()
-  const dollarToLocalRate = useDollarToLocalRate()
+  let localCurrencyCode = useLocalCurrencyCode()
+  let dollarToLocalRate = useDollarToLocalRate()
+  if (currencyInfo) {
+    localCurrencyCode = currencyInfo.currencyCode
+    dollarToLocalRate = currencyInfo.exchangeRate
+  }
```
tarikbellamine: What do you think about simplifying the code a bit and defining these variables by using a ternary like you did elsewhere?

tarikbellamine (commented):
Nice work! Left a couple of comments that are mostly stylistic but the logic looks good to me

packages/mobile/src/send/SendConfirmation.tsx
isabellewei: in the case that we're not navigating to `SendConfirmation` from `handleSendPaymentData` (ie. from `SendAmount` maybe), the `CurrencyInfo` variable is undefined, and the `localCurrencyExchangeRate` is set by `getLocalCurrencyExchangeRate(state)` instead, which could be null. So Typescript throws an error if I don't include null here

packages/mobile/src/components/TotalLineItem.v2.tsx
isabellewei: I tried doing that but it won't let me put React hooks in conditional operators :( I also hate this section of code and am totally open to other workarounds

packages/mobile/src/send/utils.ts
isabellewei: Mhm, I think `selectPreferredCurrency` was being called just as a temp workaround for the grameen pilot

packages/mobile/src/send/SendConfirmation.tsx
annakaz: If exchangeRate is only passed sometimes, could it make sense to be passing currencyCode and exchangeRate directly rather than as a CurrencyInfo object to make it clear when exchangeRate is included or not?

packages/mobile/src/send/utils.ts
annakaz: +1, note that we don't have automatic types for sagas so being explicit with `const exchangeRate: string = yield call(fetchExchangeRate, currency)` is needed to type check

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -131,9 +133,14 @@ export default function CurrencyDisplay({
...
   style,
+  currencyInfo,
 }: Props) {
-  const localCurrencyCode = useLocalCurrencyCode()
-  const dollarToLocalRate = useDollarToLocalRate()
+  let localCurrencyCode = useLocalCurrencyCode()
+  let dollarToLocalRate = useDollarToLocalRate()
+  if (currencyInfo) {
+    localCurrencyCode = currencyInfo.currencyCode
+    dollarToLocalRate = currencyInfo.exchangeRate
```
annakaz: Is there ever a case when currencyInfo is defined but `currencyInfo.exchangeRate` is null?

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -131,9 +133,14 @@ export default function CurrencyDisplay({
...
   hideFullCurrencyName,
   style,
+  currencyInfo,
 }: Props) {
-  const localCurrencyCode = useLocalCurrencyCode()
-  const dollarToLocalRate = useDollarToLocalRate()
+  let localCurrencyCode = useLocalCurrencyCode()
+  let dollarToLocalRate = useDollarToLocalRate()
+  if (currencyInfo) {
+    localCurrencyCode = currencyInfo.currencyCode
```
annakaz: Could you name the currencyInfo something that makes it clear that it's transaction-specific? Maybe transactionCurrencyCode?

packages/mobile/src/send/utils.ts
```diff
@@ -188,34 +184,30 @@ export function* handleSendPaymentData(
...
-    navigate(Screens.SendConfirmation, { transactionData, isFromScan: true })
+    const currencyInfo: CurrencyInfo = {
+      currencyCode: currency,
+      exchangeRate,
+    }
+
+    navigate(Screens.SendConfirmation, {
+      transactionData,
+      isFromScan: true,
+      currencyInfo,
```
annakaz: I would find this more readable if currencyCode and exchangeRate were explicitly passed

annakaz (commented):
Looks good overall!

packages/mobile/src/send/SendConfirmation.tsx
annakaz: exchangeRate is always passed, resolving

packages/mobile/src/icons/Logo.v2.tsx
```diff
@@ -0,0 +1,65 @@
+import colors from '@celo/react-components/styles/colors'
```
annakaz: Was this file accidentally included? I think it was deprecated

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -45,251 +48,180 @@ import { navigate } from 'src/navigator/NavigationService'
...
+  const validatedRecipientAddress = getSecureSendAddress(
+    transactionData.recipient,
+    secureSendPhoneNumberMapping
+  )
+  const account = useSelector(currentAccountSelector)
+  const isSending = useSelector(isSendingSelector)
+  // tslint:disable-next-line: react-hooks-nesting
+  const dollarBalance = useSelector(stableTokenBalanceSelector) || '0'
+  const appConnected = useSelector(isAppConnected)
+  const isDekRegistered = useSelector(isDekRegisteredSelector) ?? false
```
annakaz, line 113: Is the tslint disable not needed here?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -45,251 +48,180 @@ import { navigate } from 'src/navigator/NavigationService'
...
+type OwnProps = StackScreenProps<StackParamList, Screens.SendConfirmation>
+type Props = OwnProps
 
-interface State {
-  modalVisible: boolean
-  encryptionDialogVisible: boolean
-  buttonReset: boolean
-  comment: string
-}
+export const sendConfirmationScreenNavOptions = () => ({
```
annakaz, line 69: Thanks for refactoring this into a functional component!

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -45,251 +48,180 @@ import { navigate } from 'src/navigator/NavigationService'
...
-  }
-
-  cancelModal = () => {
-    this.setState({ modalVisible: false, buttonReset: true }, () => {
-      this.setState({ buttonReset: false })
-    })
+  const cancelModal = () => {
+    setButtonReset(true)
+    setModalVisible(false)
+    // this.setState({ modalVisible: false, buttonReset: true }, () => {
```
annakaz: Can this be removed if unused?

annakaz (commented):
Looks good overall! Left some minor comments

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -144,7 +144,7 @@ function FiatExchangeOptions({ route, navigation }: Props) {
   }
 
   const { isAddFunds } = route.params
-  const { t } = useTranslation('fiatExchangeFlow')
+  const { t } = useTranslation(Namespaces.fiatExchangeFlow)
```
jeanregisser, line 147: 👍

packages/mobile/src/identity/reducer.ts
```diff
@@ -383,6 +383,8 @@ export const acceptedAttestationCodesSelector = (state: RootState) =>
   state.identity.acceptedAttestationCodes
 export const e164NumberToAddressSelector = (state: RootState) => state.identity.e164NumberToAddress
 export const addressToE164NumberSelector = (state: RootState) => state.identity.addressToE164Number
+export const addressToDataEncryptionKeySelector = (state: RootState) =>
+  state.identity.addressToDataEncryptionKey
```
jeanregisser, line 387: This looks unrelated to this PR, maybe a merge result of a bad merge conflict?

packages/mobile/src/send/utils.ts
```diff
@@ -189,34 +184,26 @@ export function* handleSendPaymentData(
...
+      amount: dollarAmount,
       reason: data.comment,
       type: TokenTransactionType.PayPrefill,
     }
-    navigate(Screens.SendConfirmation, { transactionData, isFromScan: true })
+
+    navigate(Screens.SendConfirmation, {
+      transactionData,
+      isFromScan: true,
+      currencyInfo: { localCurrencyCode: currency, localExchangeRate: exchangeRate },
```
jeanregisser, line 205: From the conversation we had with @corey and @nityas, my understanding was that we would go with the solution where "Everyone sees things in their own currency—Jean requests  €100, Nitya sees a request for $110. No changes to drilldown."
In that case the `currencyInfo` override here isn't needed, right?

jeanregisser (commented):
Thanks for fixing this, see my comments below.

packages/mobile/src/identity/reducer.ts
isabellewei: I had to define this selector to use in `SendConfirmation`!

packages/mobile/src/send/SendConfirmation.tsx
isabellewei: I guess it's not considered a conditional operator?

packages/mobile/src/send/utils.ts
isabellewei: oh hm, from what I understood, that's usually the case if the transaction just specifies an amount in cusd, but if the amount is given in a specific currency (ex. when scanning a QR code for a specific transaction), then we default to that currency.

packages/mobile/src/identity/reducer.ts
jeanregisser: Oh right, missed that because the GitHub UI collapsed that file 👀

packages/mobile/src/send/utils.ts
jeanregisser: Can you check with @nityas / @coreycelo to confirm what we want?

isabellewei: Seems to be what we want :) 

Screen Shot 2020-10-21 at 1 18 00 PM

jeanregisser: Sounds good! let's merge this!

jeanregisser (approved):
Great! :shipit: 👍

Lss-Ankit:
@jeanregisser  to verify this task, i need to sync with @evan for OTP to login into BeanAndGo web app. 
will sync it and update the status here

Lss-Ankit:
@jeanregisser @isabellewei verified task with Test flight build v1.5.1 (30) and Play store internal build 1.5.1 (1004294319) and found the following. 
- PHP currency is shown on Sending screen when user taps on OPEN VALORA APP button from B&G site. 
- Verified USD/by default selected in app currency is shown after navigate back from the sending screen 
- PHP currency is shown on Sending screen when user scan the QR code from B&G site. 
- Verified USD/by default selected in app currency is shown after navigate back from the sending screen 

**Need Clarification:**
- PHP amount is shown on sending screen when user taps on OPEN VALORA APP or Scan QR code from B&G site even though user has selected other currency from app setting. 
can you please let us now is it expected or not? 

Also, we have not purchased any items after tapping on OPEN VALORA APP or scanning QR code from B&G site as we have not much balance. 
Please let us know if we need to proceed further to verify this task. 
Thanks!!

jeanregisser:
Yes that's the expected behavior, thanks @Lss-Ankit

Lss-Ankit:
Okay thanks @jeanregisser for the update.

PR #5118, opened 2020-09-21
Add MetaTransactionWalletDeployer

bowd:
### Description

This is a factory contract that is responsible for deploying and initialising MetaTransactionWalletProxy contracts and setting up the implementation.
It also keeps track of deployed instances in a mapping so that each external account will have at most one wallet deployed.

### Other changes

I've added this line ...
```solidity
```
... to the MetaTransactionWallet itself at the beginning of the `initialize` function because when the contract is set as an implementation to a proxy the `_owner` is `0x0` initially as it's not set in the constructor and thus the `initialize` call always fails because `setSigner` is `onlyOwner`. 
This way when we're calling `initialize` on the proxy, the owner is temporarily set to the `msg.sender` in order to be able to run the rest of the setup and then ownership gets transferred to the contract itself. 

### Tested

Added tests for all flows.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

New functionality (and a fix that's backwards compatible)


asaj (commented):
What is the advantage of a factory contract over just deploying things "manually"?

bowd:
> What is the advantage of a factory contract over just deploying things "manually"?

It's cheaper and it also makes more sense as a place to keep the mapping between external accounts and wallets which makes it easily recoverable for Valora as well.

@diminator - I know you brought this up initially, am I missing some other advantages?

diminator:
Not much besides. Was mostly a matter of having efficient gas in feeless onboarding.

bowd:
@asaj There's one more reason I totally spaced on yesterday (tired):
We wanted the API to behave in a predictable manner for clients in the sense that all API calls that result in chain interaction will return a TX hash back to the client within a reasonably small amount of time and then the client will wait for that TX to complete at which point it can move to the next step in the flow.
If we wanted to do things manually it means that the call to deploy would need to make two transactions:
1. Deploy the proxy
2. Set implementation and initialise with the signer

So we had three options here: 
1. Wait for the contract to be deployed and return to the client with the TX hash of the 2nd operation - not really good because the API call could stay open at the worst 5s
2. Split the deploy action into two API calls (deploy and initialise) which are both called by the client
3. Use the factory contract which would result in the deploy operation being a single transaction and would maintain the invariant that any API request that touches the chain results in a single TX.

Going with (3) seamed appropriate because it also gave us the efficiency win and provide a good source of truth for the EOA-to-Wallet mapping.

packages/protocol/contracts/common/MetaTransactionWalletDeployer.sol
```diff
@@ -0,0 +1,98 @@
...
+
+  event WalletDeployed(address indexed owner, address indexed wallet, address implementation);
+  event DeployerStatusGranted(address indexed addr);
+  event DeployerStatusRevoked(address indexed addr);
+
+  /**
+     * @dev Verifies that the sender is allowed to deploy a wallet
+     */
+  modifier onlyCanDeploy() {
+    require(msg.sender == owner() || canDeploy[msg.sender] == true, "not-allowed");
```
ewilz: I'm unfamiliar with any conventions clabs has on revert messages, but if I were debugging I might appreciate something more specific like: "sender not authorized to deploy wallet". "not-allowed", if used elsewhere too, could make debugging difficult?

packages/protocol/contracts/common/MetaTransactionWalletDeployer.sol
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  /**
+     * @notice Used to deploy a MetaTransactionWalletProxy, set the implementation,
+     * initialize, transfer ownership and emit an event.
+     * @param owner The external account which will act as signer and owner of the proxy
+     * @param implementation The address of the implementation which the proxy will point to
+     */
+  function deploy(address owner, address implementation, bytes calldata initCallData)
```
ewilz, line 86: we should probably add `initCallData` to the docstrings above!

packages/protocol/contracts/common/MetaTransactionWalletDeployer.sol
```diff
@@ -0,0 +1,98 @@
...
+      _changeDeployerAllowance(initialDeployers[i], true);
+    }
+  }
+
+  /**
+     * @notice Change the deployer status of an address
+     * @param target The address to be allowed as a deployer
+     * @param allowedToDeploy toggle whether the address is allowed or not
+     */
+  function changeDeployerAllowance(address target, bool allowedToDeploy) external onlyOwner {
```
ewilz: `allowance` sounds like an allotted sum of money, especially given this is a monetary app. I feel I never hear the word outside that context. `Permissions` sounds clearer to me. But this is a nitpick and not a blocker if other people are OK with `allowance`

bowd:
@emaG3m Thanks for the feedback! You're right that naming was a bit weird. I also handled the rest of the comments.

packages/protocol/contracts/common/MetaTransactionWalletDeployer.sol
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /**
+     * @notice Implementation of permission change
+     * @param target The address to be allowed as a deployer
+     * @param allowedToDeploy toggle whether the address is allowed or not
+     */
+  function _changeDeployerPermission(address target, bool allowedToDeploy) internal {
+    canDeploy[target] = allowedToDeploy;
+    if (allowedToDeploy == true) {
```
codyborn:
```suggestion
    if (allowedToDeploy) {
```

packages/protocol/contracts/common/MetaTransactionWalletDeployer.sol
```diff
@@ -0,0 +1,102 @@
...
+  event WalletDeployed(address indexed owner, address indexed wallet, address implementation);
+  event DeployerStatusGranted(address indexed addr);
+  event DeployerStatusRevoked(address indexed addr);
+
+  /**
+     * @dev Verifies that the sender is allowed to deploy a wallet
+     */
+  modifier onlyCanDeploy() {
+    require(
+      msg.sender == owner() || canDeploy[msg.sender] == true,
```
codyborn:
```suggestion
      msg.sender == owner() || canDeploy[msg.sender],
```

packages/protocol/test/common/metatransactionwalletdeployer.ts
```diff
@@ -0,0 +1,186 @@
...
+        })
+      })
+    })
+
+    describe('when permission is granted', async () => {
+      beforeEach(async () => {
+        changePermissionRes = await deployer.changeDeployerPermission(otherAllowedDeployer, true)
+      })
+
+      it('should have added the allownace and emit an event', async () => {
```
codyborn:
```suggestion
      it('should have added the allowance and emit an event', async () => {
```

codyborn (approved):
LGTM

Issue #4835, opened 2020-08-26
Quality / bug fixing


jeanregisser:
- [ ] Continue addressing existing bugs and other minor improvements, see https://github.com/orgs/celo-org/projects/150
- [ ] Transaction feeds fixes (unknown, qr scan, etc) 
- [ ] Test iOS 14 beta (new API for clipboard monitoring

gnardini:
Closing since it's an epic from a previous milestone

PR #5121, opened 2020-09-21
[Wallet] Update Clipboard for iOS 14

### LINKED ISSUE #5120: Update Clipboard for iOS 14
iOS 14 brought a new API for the clipboard which we need to implement

### Expected Behavior

A notification shouldn't appear every second on iOS 14+ when on a screen with an input where we can paste.

### Current Behavior

A notification appears every second.

gnardini:
### Description

On iOS 14 when we read the clipboard the user gets a notification. Since we're reading it every second to detect changes, it's quite annoying. This PR's goal is to have everything work the way it used to except on iOS devices running version 14+. In those devices, the paste button/icons are always visible if the clipboard is not empty and is hidden only if the clipboard is empty. In short, we lose the "smart" paste icons on iOS 14+ that showed up only if the format of the content of the clipboard matches the expected format.

### Other changes

Updated the clipboard library.

### Tested

On the simulator using different iOS versions.

### Related issues

- Fixes #5120

Part of this epic: https://github.com/celo-org/celo-monorepo/issues/4835


packages/mobile/package.json
```diff
@@ -45,6 +45,7 @@
     "@celo/react-native-sms-retriever": "git+https://github.com/celo-org/react-native-sms-retriever#b88e502",
     "@celo/utils": "0.1.20-dev",
     "@react-native-community/async-storage": "^1.9.0",
+    "@react-native-community/clipboard": "git+https://github.com/gnardini/clipboard#01d1c13",
```
gnardini: I opened a PR on their repo: https://github.com/react-native-community/clipboard/pull/73 but I'm not sure they'll accept it since I didn't implement the whole new API, just the method we needed. Should I move the fork to celo-org?

packages/mobile/src/utils/useClipboard.ts
```diff
@@ -1,17 +1,25 @@
...
+export function useClipboard(): [boolean, string] {
+  const [forceShowingPasteIcon, setForceShowingPasteIcon] = useState(false)
   const [clipboardContent, setClipboardContent] = useState('')
 
   useEffect(() => {
     let isMounted = true
 
     async function checkClipboardContent() {
       try {
+        const majorVersionIOS = parseInt(Platform.Version.toString(), 10)
```
aaronmgdr: noticing that this line and the following  lines are duplicated in withPasteAware.

packages/mobile/package.json
jeanregisser: Nice! Would be good to move to celo-org indeed 👍

packages/mobile/src/invite/EnterInviteCode.test.tsx
```diff
@@ -37,9 +38,18 @@ const INVALID_REFERRER_INVITE_URL = 'http://example.com?invite-code=abc'
 SendIntentAndroid.openSMSApp = jest.fn()
 
 const clipboardGetStringMock = (Clipboard.getString = jest.fn())
+const clipboardHasStringMock = (Clipboard.hasString = jest.fn())
+Platform.OS = 'ios'
```
jeanregisser: It would be better to only change this in a specific `describe` section for testing this.
Otherwise it's gonna be applied to all tests here. Which currently is `android` by default.

packages/mobile/src/utils/useClipboard.ts
```diff
@@ -1,17 +1,25 @@
+import Clipboard from '@react-native-community/clipboard'
 import { useEffect, useState } from 'react'
-import { AppState, Clipboard } from 'react-native'
+import { AppState, Platform } from 'react-native'
 import Logger from 'src/utils/Logger'
 
 const CLIPBOARD_CHECK_INTERVAL = 1000 // 1sec
 
-export function useClipboard(): string {
+export function useClipboard(): [boolean, string] {
```
jeanregisser: The new API for this helper hook is getting confusing with this new use case.
If the returned boolean is `true`, you have to use `await Clipboard.getString()` outside of this and not use the 2nd value here.
 
I'm trying to see if there's a way to make it more obvious / less error prone.
What do you think?

jeanregisser (commented):
Thanks for updating this!

I'm only concerned with the new API not mapping very nicely with our existing usage.
I don't yet have a proposition for it though :)

packages/mobile/src/utils/useClipboard.ts
gnardini: Agreed, how about instead of a string we return a function that returns the contents of the clipboard? The API would still look weird (why a function instead of a string directly?) but it would reduce the error-proneness imo.

gnardini: I took another pass as this, but I'm not convinced by the result. It still looks weird to me. Maybe different naming can help? I added a comment to try to clarify a bit..

packages/mobile/src/utils/useClipboard.ts
gnardini: Added a helper method to reuse logic :)

packages/mobile/src/utils/useClipboard.ts
jeanregisser: Ok I've thought about something. I'll detail it tomorrow. 
In the mean time, I think it's fine to still merge this as is.

packages/mobile/src/utils/useClipboard.ts
```diff
@@ -1,17 +1,31 @@
+import { deviceIsIos14OrNewer } from '@celo/react-components/components/WithPasteAware'
```
jeanregisser: nit: would be nicer to have this util in another file.

packages/react-components/components/WithPasteAware.tsx
```diff
@@ -20,6 +21,11 @@ interface PasteAwareState {
   clipboardContent: string | null
 }
 
+export function deviceIsIos14OrNewer() {
+  const majorVersionIOS = parseInt(Platform.Version.toString(), 10)
```
cmcewen: it it safer to use parseFloat here for iOS 14.1, etc? i guess it does seem like parseInt rounds down (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) but just may be a little confusing

jeanregisser (approved):
Looking good! 👍

cmcewen (approved):

packages/react-components/components/WithPasteAware.tsx
gnardini: My understanding of `parseInt` is that it just dismisses the decimal part, like a cast to int does in other languages, but since it can cause confusion I jut changed it to `parseFloat` :)

PR #5122, opened 2020-09-21
Fix the way we're signing EIP712 typed data

### LINKED ISSUE #5097: MTW Signature doesn't follow EIP712 correctly
### Overview

Extracted from Slack:

Reading the [spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#specification), it sounds like EIP712 signatures should be signature of the message:

```
encode(domainSeparator : 𝔹²⁵⁶, message : 𝕊) = "\x19\x01" ‖ domainSeparator ‖ hashStruct(message)
``` 
However, under the RPC spec, they specify...

>  The sign method calculates an Ethereum specific signature with: sign(keccak256("\x19Ethereum Signed Message:\n" + len(message) + message))).

...which seems to contradict the message spec, and possible is just copied from the actual personal_sign method. The [test cases](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#test-cases) agree, that the Ethereum Signed Message prefix should not be used, instead it should be `0x1901`.

However, it looks like in MetaTransactionWallet.sol, we are using a mixture of both, since we are first creating the [“right message”](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/MetaTransactionWallet.sol#L125), but then also prepend `Ethereum Signed Message` [here.](https://github.com/celo-org/celo-monorepo/blob/master/packages/protocol/contracts/common/MetaTransactionWallet.sol#L149)

One last data point, `celo-blockchain` seems to have the implementation “per spec” as well.

Actually one more thing, it looks like there isn’t any consensus on EIP712 in the Ethereum ecosystem? Geth hasn’t added it to their RPC API yet. web3.js and ethers.js haven’t added it. The only wallet to support EIP712 [seems to be metamask](https://medium.com/metamask/eip712-is-coming-what-to-expect-and-how-to-use-it-bb92fd1a7a26), it is unclear to me that EIP712 [is actually a standard](https://www.reddit.com/r/ethereum/comments/absnnx/eip712_and_state_channels/ed2s113/).

Metamask’s own documentation page [highlights the discrepancy](https://docs.metamask.io/guide/signing-data.html#signing-data-with-metamask) (hardware wallets can’t sign the spec and metamask itself has 3 active versions of signTypedData)

bowd:
### Description

See Issue #5097 

I opted for moving all the generic logic to `Sigantures.sol` and thus rely on that library to construct the typed data hash with the correct prefix and etc. Which required some small changes on the MetaTransactionWallet side. I think this provides a bit of a more generic structure for typed data usage in the future.

### Tested

Specs updated to reflect change in signature

### Related issues

- Fixes #5097 

### Backwards compatibility

No concerns here


ewilz:
Hi, I'm Emily, new to clabs/contracts ^_^  I have some questions to help me better understand EIP-712 and where we stand with it in MTW after this PR. Nice refactor. Looking at this it seems like the only change in functionality is `getSignerOfTypedDataHash`/ `toEthSignedTypedDataHash` to match the 712 prefix, as well as exposing `getMetaTransactionStructHash` logic. Is this the fix that @asaj was referring to, or do we still have more on getting this fully EIP712 compliant?

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -99,6 +99,32 @@ contract MetaTransactionWallet is
...
 
+  /**
+   * @notice Returns the struct hash of the MetaTransaction
+   * @param destination The address to which the meta-transaction is to be sent.
+   * @param value The CELO value to be sent with the meta-transaction.
+   * @param data The data to be sent with the meta-transaction.
+   * @param _nonce The nonce for this meta-transaction local to this wallet.
+   * @return The digest of the provided meta-transaction.
+   */
+  function getMetaTransactionStructHash(
```
ewilz: if this is public, per recent zeppelin audit, it should probably be listed in the `IMetaTransactionWallet` interface as well :)

packages/protocol/contracts/common/Signatures.sol
```diff
@@ -49,4 +49,44 @@ library Signatures {
     bytes32 prefixedHash = ECDSA.toEthSignedMessageHash(messageHash);
     return ECDSA.recover(prefixedHash, signature);
   }
+
+  /**
+  * @notice Given a domain separator and a structHash, construct the typed data hash
+  * @param eip712DomainSeparator Context specific domain separator
+  * @param structHash hash of the typed data struct
+  * @return The EIP712 singed typed data hash
```
ewilz: signed**

packages/protocol/contracts/common/MetaTransactionWallet.sol
bowd: Yeah I'm not actually sure if it makes that much sense to be public. I think `getMetaTransactionDigest` is enough so we can mark this as internal :-/

bowd:
> Hi, I'm Emily, new to clabs/contracts ^_^ I have some questions to help me better understand EIP-712 and where we stand with it in MTW after this PR. Nice refactor. Looking at this it seems like the only change in functionality is `getSignerOfTypedDataHash`/ `toEthSignedTypedDataHash` to match the 712 prefix, as well as exposing `getMetaTransactionStructHash` logic. Is this the fix that @asaj was referring to, or do we still have more on getting this fully EIP712 compliant?

Hey Emily and welcome to clabs! I'm also fairly new, started a few months ago :)
So EIP712 is also fairly new to me, but here's my understanding:

Before the fix we were constructing this hash...
```solidity
keccak256(abi.encodePacked("\x19\x01", eip712DomainSeparator, structHash))
```
... as part of the `MetaTransactionWallet` which is the correct EIP712 hash, but when recovering we were using `Signatures.getSignerOfMessageHash` which in turn used `ECDSA.toEthSignedMessageHash` passing in the above hash and thus we were recovering the signature from:
```solidity
keccak256(abi.encodePacked(
  "\x19Ethereum Signed Message:\n32", 
  keccak256(abi.encodePacked("\x19\x01", eip712DomainSeparator, structHash)
)));
```

This is also apparent in the specs (pre-fix) where we were relying on the `eth_sign` RPC call and passing it the hash of the typedData and that RPC call prefixes with `Ethereum Signed Message` and signs.

What I did to mitigate this was to extend `Signatures.sol` with functions that are specific for EIP712 operations:
-  `toEthSignedTypedDataHash(bytes32 eip712DomainSeparator, bytes32 structHash)`
Which is the TypedData counterpart for the `ECDSA.toEthSignedMessageHash` and is responsible for constructing the correct digest with the EIP712 prefix
- `getSignerOfTypedDataHash(bytes32 eip712DomainSeparator, bytes32 structHash, uint8 v, bytes32 r, bytes32 s)`
Which is the TypedData counterpart for the `Signatures.getSignerOfMessageHash` and is responsible for recovering the signer of a typed data digest.

My mental modal was to construct functions for these two cases:
- `Ethereum Signed Message` where the input is the `messageHash`
- `EIP712 Signed Typed Data` where the inputs are: `eip712DomainSeparator` and `structHash`

Also post-fix I've changed the tests to rely `eth_signTypedData` for the signature. This RPC call's implementation is at parity with what's happening in Signatures.sol.

Let me know if that makes sense - like I said I'm also not super in the know here, so I'm grateful if you point out any cracks in the logic.

ewilz:
got it! do we have any tests that would fail if we were to switch it back to the way it was before?? Ideally signing one way, and recovering another way would've broke a test....

bowd:
> got it! do we have any tests that would fail if we were to switch it back to the way it was before?? Ideally signing one way, and recovering another way would've broke a test....

Yup after changing the smart contract logic all the tests that were signing meta transactions failed because they were signing using the `eth_sign` RPC. The PR also changes that logic to start singing using the `eth_signTypedData` RPC call which hashes things in the same way as the improved contract logic.

packages/protocol/contracts/common/Signatures.sol
```diff
@@ -49,4 +49,44 @@ library Signatures {
...
     return ECDSA.recover(prefixedHash, signature);
   }
+
+  /**
+  * @notice Given a domain separator and a structHash, construct the typed data hash
+  * @param eip712DomainSeparator Context specific domain separator
+  * @param structHash hash of the typed data struct
+  * @return The EIP712 typed data hash
+  */
+  function toEthSignedTypedDataHash(bytes32 eip712DomainSeparator, bytes32 structHash)
```
nambrot, line 59: I don't know off the top of my head, but would it be possible for calllers to pass something closer to the struct itself instead of the hash? Ideally we'd mirror the sign interface? The encoding + hashing is ideally something a caller shouldn't have to think about?

bowd: Yeah I tried going in that direction that's why I setup up this function to move more of that hash construction away from the consumer of `Signatrues.sol` but I don't see a way to move away unless we create another level of abstraction for each type of signed data that we want, and introduce a new library contract that's responsible for constructing these hashes, which would mean splitting that functionality away from the `MetaTransactionWallet` contract into somewhere else. That way other contracts can rely on that library to construct the hashes and so on without consuming the full `MetaTransactionWallet` but I don't see a really clear benefit to that.

nambrot: Gotcha, what about just passing the bytes of the `abi.encode`? Or would that not really "hide" the implementation details enough?

bowd: Doesn't feel like a huge win but would just maybe add some bloat to the traces (not 100% sure how library calls are encoded). 

But I could be convinced. Any other folk want to chime in: @asaj / @m-chrzan ?

nambrot: Yeah, I don't think its a huge difference, and i think its fine to keep this way

emaG3m: to me it seems appropriate that the `caller`/`client` would encode and hash the input  off-chain. it's cheaper, more efficient, and a good client would hopefully make it intuitive for the non-technical end-user.

ewilz: merp forgot about email notifications...but deleted my last comment bc it didn't really apply

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -99,6 +99,32 @@ contract MetaTransactionWallet is
...
+    bytes memory data,
+    uint256 _nonce
+  ) internal view returns (bytes32) {
+    return
+      keccak256(
+        abi.encode(
+          EIP712_EXECUTE_META_TRANSACTION_TYPEHASH,
+          destination,
+          value,
+          keccak256(data),
```
ewilz, line 122: everything looks good to me. There was a side conversation (@nambrot was there) in slack that mentioned that this keccak256 is extra and useless. Getting rid of it might complicate things on your end, so I don't think it's a big deal, but if it's an easy delete, we don't need this.

EDIT: THIS IS WRONG. totally different contract with nested keccak.

ewilz (approved):

PR #5124, opened 2020-09-21
Fixes backward compatibility issues in cli

medhakothari:
### Description

Fixes the bug in the CLI package mentioned [here](https://discord.com/channels/600834479145353243/600840423958904842/751439853756874863). 

Basically in the PR linked below, the `nodeUrl` key was renamed to `node`  and a new key, `gasCurrency`, was introduced. In `cli/utils/config.ts`, the `readConfig` function would assume that if the config json exists, then we have everything. But that's not true for anyone using cli versions <= 0.0.53, because they would only have one config key — `nodeUrl`/`node`. Their config wouldn't have `gasCurrency` set. This is what causes the "null" in the error message. 

So, this PR mitigates two issues:
- the rename of the `nodeUrl` key
- allows backward compatibility for any new keys added to the config

### Tested

- Set the config using CLI Version 0.0.53 —> Used the CLI using 0.0.57-dev (no error) vs 0.0.56 (error)
- Set the config using Version 0.0.56 —> Used the CLI using 0.0.57-dev (no error) vs 0.0.53 (error)

### Related issues

- #4765 

### Backwards compatibility

- adds in backward compatibility for any new keys added to cli's config


packages/cli/src/utils/config.ts
```diff
@@ -22,7 +22,13 @@ export function configPath(configDir: string) {
 
 export function readConfig(configDir: string): CeloConfig {
   if (fs.pathExistsSync(configPath(configDir))) {
-    return fs.readJSONSync(configPath(configDir))
+    const existingConfig = fs.readJSONSync(configPath(configDir))
+    const combinedConfig = { ...existingConfig, ...defaultConfig }
+    if (combinedConfig.hasOwnProperty('nodeUrl')) {
+      /* tslint:disable:no-string-literal */
+      combinedConfig['node'] = combinedConfig['nodeUrl']
```
aaronmgdr: would this work? `combinedConfig.node = combinedConfig.nodeUrl`

I think then no tslint disabled is needed.

packages/cli/src/utils/config.ts
```diff
@@ -22,7 +22,13 @@ export function configPath(configDir: string) {
 
 export function readConfig(configDir: string): CeloConfig {
   if (fs.pathExistsSync(configPath(configDir))) {
-    return fs.readJSONSync(configPath(configDir))
+    const existingConfig = fs.readJSONSync(configPath(configDir))
+    const combinedConfig = { ...existingConfig, ...defaultConfig }
```
gastonponti: Watch out, this will override the existing with the default.
```
j1= { a: 'j1', b: null }
j2= {a: 'j2', b: 'j2' }
{...j1, ...j2} => {a: 'j2', b: 'j2' }
{...j2, ...j1} => { a: 'j1', b: null }
```
So, basically overrides the keys that has in the object, it doesn't matter if it's null/undefined, if the key exists, it will step on it. And it doesn't take in count if the key already had a valid value
for `{...j1, ...j2}` instead of doing
`jr['a'] = j1['a'] ?? j2['a']`
it will make
```
jr['a'] = j1['a']
jr['a'] = j2.hasOwnProperty('a')? j2['a'] : jr['a']
```

In this case, I think that you want to use the default as the base configuration, and override it, with what you have in the file
`{ ...defaultConfig, ...existingConfig }`

packages/cli/src/utils/config.ts
gastonponti: Adding a little to what Aaron said,
Typescript when you've checked that an object already has a property, in that block of the if, it will assume that the property is in the object. So, it will let you do `combinedConfig.nodeUrl` without any warning.

So yes, in this case, you could use it like this: `combinedConfig.node = combinedConfig.nodeUrl`

gastonponti (changes requested):

packages/cli/src/utils/config.ts
medhakothari: great catch, fixed!

packages/cli/src/utils/config.ts
medhakothari: good catch!

gastonponti (approved):
LGTM 👍 🚀

Issue #4563, opened 2020-07-29
Verification UI Experience Improvements


nityas:
### Expected Behavior

- [x] Users should be able to enter in received text messages - restarting verification just requests more
- [x] Onboarding should check if I’m already verified before prompting. I have an account that I restored that is already verified to the number I entered in
- [x] Possible manual re-request attestation CTA (eg. button next to each import message box that requests from new issuer)
~~- [ ] change mobile's expiry definition for an attestation request to 5min and re-request new attestation on resume~~

Addressed more fully by an ongoing Verification Enhancements Feature: [figma](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet--Refresh?node-id=8047%3A238)

### Current Behavior

What is the current behavior?

i1skn:
@nityas do you think we can remove "change mobile's expiry definition for an attestation request to 5min and re-request new attestation on resume" in light of the recent discussions?

i1skn:
@nityas is there things left to do in this one?

i1skn:
Closing this one as it seems to be completed

PR #5134, opened 2020-09-22
[Wallet] Add resend button to VerificationInputScreen

### LINKED ISSUE #5216: Add "resend messages" option to the Verification flow
### Expected Behavior
Users SBAT request more attestations if they did not receive all 3 text messages as intended.

### Current Behavior
Users are unable to explicitly request more text messages.

i1skn:
### Description

* Remove "I already received three codes" and merge its logic into "Resume" button on the VerificationEducation screen.
* Add "Resend N messages" on the VerificationInput screen. Button is only available after a minute since the last reveal. When pressed we will try to reveal already existing attestations first. If HTTP call to attestation-service returns anything, but `OK` we request another issuer and try to reveal the number to it. Those we increase number of actionable attestations till the limit 5, after which user need to wait until some attestations expire to request more.
* Updated text for "Do I need to confirm" dialog in English.

See [figma](https://www.figma.com/file/zt7aTQ5wuXycIwxq5oAmF9/Wallet-Refresh?node-id=7420%3A2384) for more details.

### Tested

iOS & Android with actual requesting of SMSs

### Related issues

- Part of https://github.com/celo-org/celo-monorepo/issues/4563
- Closes #5216

### Backwards compatibility

Yes

### Screenshots

Simulator Screen Shot - iPhone 11 Pro - 2020-09-22 at 16 57 36
Simulator Screen Shot - iPhone 11 Pro - 2020-09-22 at 16 25 06
Simulator Screen Shot - iPhone 11 Pro - 2020-09-22 at 16 22 08


packages/mobile/src/identity/verification.ts
```diff
@@ -534,52 +530,100 @@ function* isCodeAlreadyAccepted(code: string) {
...
+
+  const e164Number: string = yield select(e164NumberSelector)
+
+  const { phoneHashDetails }: VerificationState = yield select(verificationStateSelector)
+  const phoneHash = features.USE_PHONE_NUMBER_PRIVACY
+    ? phoneHashDetails.phoneHash
+    : getPhoneHash(e164Number)
+  const account: string = yield call(getConnectedUnlockedAccount)
+
+  const actionableAttestations: ActionableAttestation[] = yield call(
```
aslawson: So this appears to only try to re-reveal the already selected issuers (unless I'm missing something in the PR).

We also wanted to request an additional issuer for the case that the SMS could not be received. I think without a "successful response cache" or Tim's new `/get_attestations` endpoint yet, we could accomplish this by re-revealing to the original issuer, and if it returns an error then request a new issuer. WDYT?

i1skn: @aslawson  Yeap, you got it right! 

In this case it's pretty hard to predict if user has sufficient balance in advance. So, after user presses this button and we try to request another issuer with insufficient balance - we would show an error (or should we silently handle this and do nothing if balance is insufficient?). If we show an error, how can we explain to user why suddenly they need more funds?

i1skn:
There is a vulnerability which does not have a fix yet https://www.npmjs.com/advisories/1560/versions and it blocks the build

packages/mobile/src/identity/types.ts
```diff
@@ -12,8 +12,7 @@ export enum VerificationStatus {
   Prepping = 1,
   GettingStatus = 2,
   RequestingAttestations = 3,
-  RevealingNumber = 4,
-  RevealAttemptFailed = 5,
+  CompletingAttestations = 4,
   Done = 6,
```
gnardini, line 16: Maybe this is on purpose and this is just a nit but there is a jump between 4 and 6.

packages/mobile/locales/en-US/onboarding.json
```diff
@@ -51,18 +51,18 @@
...
     "title": "Are you sure?",
     "body": "Confirming allows you to send and receive funds easily to your phone number.\n\nUnconfirmed accounts can only send payments using Celo addresses or QR codes.",
     "cancel": "Go Back",
     "confirm": "Skip for now"
   },
+  "verificationPrematureRevealMessage": "Please wait one minute before resending",
   "verificationLearnMoreDialog": {
-    "title": "Phone Number Confirmation",
-    "body": "Phone number confirmation works by associating your account with your phone number.\n\n<0>Do I need to complete this?</0>\n\nConfirmation is not required, however, unconfirmed accounts can only send payments using Celo addresses or QR codes.\n\n<1>Security and Privacy</1>\n\nTo protect your privacy, only an obfuscated version of your phone number is stored on the Celo blockchain.",
+    "title": "Phone Numbers and Valora",
```
jeanregisser, line 64: Should be `{{appName}}` instead of `Valora`

packages/mobile/locales/es-419/onboarding.json
```diff
@@ -51,15 +51,15 @@
...
   "verificationSkipDialog": {
     "title": "Estas seguro",
     "body": "Confirmar te permite enviar y recibir fondos fácilmente a tu número de teléfono.\n\nLas cuentas no confirmadas solo pueden enviar pagos utilizando direcciones de Celo o códigos QR.",
     "cancel": "Regresa",
     "confirm": "Saltar por ahora"
   },
+  "verificationPrematureRevealMessage": "Por favor espera 1 minuto antes de volver a enviar",
   "verificationLearnMoreDialog": {
     "title": "Confirmación de número de teléfono",
     "body": "La confirmación del número de teléfono funciona asociando tu cuenta con tu número de teléfono.\n\n<0>¿Necesito completar esto?</0>\n\nNo se requiere confirmación, sin embargo, las cuentas no confirmadas solo pueden enviar pagos utilizando direcciones Celo o códigos QR.\n\n<1>Seguridad y privacidad</1>\n\nPara proteger tu privacidad, solo una versión ofuscada de tu número de teléfono se almacena en la cadena de bloques de Celo.",
```
jeanregisser, line 65: Looks like these need to be updated too.

packages/mobile/src/identity/verification.ts
```diff
@@ -193,7 +191,35 @@ export function* startVerification({ withoutRevealing }: StartVerificationAction
...
+    if (restart) {
+      isRestarted = true
+    } else {
+      return verification
+    }
+    Logger.debug(TAG, 'Verification has been restarted')
+  }
+}
+
 export function* doVerificationFlow(withoutRevealing: boolean = false) {
```
gnardini, line 225: This method looks really important. I didn't have previous knowledge of it so I appreciate the code comments, but it's still 150 lines long. I know this was already like this before but I wonder if some of the logic could be extracted to other methods to make this a bit easier to read. No need to change anything now, just thinking out loud.

packages/mobile/src/verify/VerificationEducationScreen.test.tsx
```diff
@@ -45,39 +45,6 @@ describe('VerificationEducationScreen', () => {
     expect(queryByTestId('VerificationEducationAlready')).toBeFalsy()
   })
 
-  it('shows the `continue` and `already received` buttons when there are actionable attestations', () => {
```
jeanregisser, line 48: Shouldn't we update the test instead of deleting it, for the state when there are actionable attestations?

gnardini (approved):
I get a bit lost on the details of `packages/mobile/src/identity/verification.ts` but it looks good to me.

packages/mobile/src/identity/verification.ts
i1skn: I agree, the whole saga related logic needs a refactoring 👍

packages/mobile/src/identity/types.ts
i1skn: I did it mainly for back compatibility reasons, so I would not need right now to check do we persist this status in redux-storage and potentially creating migrations.

packages/mobile/src/verify/VerificationInputScreen.tsx
```diff
@@ -206,6 +219,18 @@ class VerificationInputScreen extends React.Component<Props, State> {
     navigateHome()
   }
 
+  onPressResend = () => {
```
tarikbellamine, line 222: Why are we showing an error message when `!isRevealAllowed` instead of hiding the button?

packages/mobile/src/identity/verification.ts
```diff
@@ -193,7 +191,35 @@ export function* startVerification({ withoutRevealing }: StartVerificationAction
...
+    })
+    yield put(resetVerification())
+    yield call(getConnectedAccount)
+    if (isRestarted || (yield select(isVerificationStateExpiredSelector))) {
+      yield call(fetchVerificationState)
+    }
+
+    const { verification, restart } = yield race({
+      verification: call(doVerificationFlow, withoutRevealing),
+      restart: take(Actions.RESEND_ATTESTATIONS),
```
tarikbellamine, line 210: What analytics events should we add to best capture data on "restarts"? I think we'll definitely want to be able to track the efficacy of this feature moving forward

packages/mobile/src/identity/verification.ts
```diff
@@ -193,7 +191,35 @@ export function* startVerification({ withoutRevealing }: StartVerificationAction
   yield put(refreshAllBalances())
 }
 
+export function* restartableVerification(initialWithoutRevealing: boolean) {
+  let isRestarted = false
+  while (true) {
+    const withoutRevealing = !isRestarted && initialWithoutRevealing
+    yield call(navigate, Screens.VerificationLoadingScreen, {
```
tarikbellamine, line 199: Are there any negative repercussions to calling `navigate` in a loop like this? What happens to the screen stack if `VerificationLoadingScreen` is already at the top when it's called? I'm not sure but maybe `replace` is more appropriate in that instance?

packages/mobile/src/verify/VerificationInputScreen.tsx
i1skn: Well, I guess Corey or Nitya can answer this better, but I think the idea is similar to why we do not disable send button, when there is not enough funds: user knows that such thing exists and also knows its limitations.

packages/mobile/src/identity/verification.ts
```diff
@@ -222,61 +248,106 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
...
-      yield put(setVerificationStatus(VerificationStatus.RevealingNumber))
-      ValoraAnalytics.track(VerificationEvents.verification_reveal_all_attestations_start)
-      yield all([
-        // Request codes for the attestations needed
-        call(
-          revealNeededAttestations,
+      if (!withoutRevealing) {
+        ValoraAnalytics.track(VerificationEvents.verification_reveal_all_attestations_start)
+        // Request codes for the already existing attestations if any.
+        // We check after which ones were successfull
```
tarikbellamine: Nit: spelling of `successful`

packages/mobile/src/identity/verification.ts
```diff
@@ -222,61 +248,106 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
...
+            issuers,
+          })
+
+          // start listening for the new list of attestations
+          receiveMessageTask?.cancel()
+          issuers = attestations.map((a) => a.issuer)
+          receiveMessageTask = yield takeEvery(
+            Actions.RECEIVE_ATTESTATION_MESSAGE,
+            attestationCodeReceiver(attestationsWrapper, phoneHash, account, issuers)
+          )
```
tarikbellamine, line 315: I notice we are listening for attestations only after all of them have been requested. Is there a risk that we might miss the first or second attestation with this approach?

What's the harm in starting to listen before we start requesting attestations?

packages/mobile/src/identity/verification.ts
i1skn: As far as I understand, if you call navigate with the current screen it would not result into any modification to the stack. If this is not true, we should re-think our approach here https://github.com/celo-org/celo-monorepo/blob/22cd8449fe457101badcde844c292b0763ee4c8e/packages/mobile/src/import/saga.ts#L64 as well, where we navigate to the same screen just with different params.

packages/mobile/src/identity/verification.ts
i1skn: There was no explicit requirement, so I will create one with the name which make sense :)

packages/mobile/src/identity/verification.ts
```diff
@@ -222,61 +248,106 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
...
+      yield put(setVerificationStatus(VerificationStatus.CompletingAttestations))
+      yield race({
+        actionableAttestationCompleted: all([
+          call(completeAttestations, attestationsWrapper, account, phoneHashDetails, attestations),
+          // Set acccount and data encryption key in Accounts contract
+          // This is done in other places too, intentionally keeping it for more coverage
+          call(registerAccountDek, account),
+        ]),
+        // This is needed, because we can have more actionableAttestations than NUM_ATTESTATIONS_REQUIRED
+        requiredAttestationsCompleted: call(function*() {
```
tarikbellamine: Nit: I think ideally this generator would be split out into a helper rather than defined inline

packages/mobile/src/identity/verification.ts
```diff
@@ -549,52 +626,83 @@ function* isCodeAlreadyAccepted(code: string) {
...
   phoneHashDetails: PhoneNumberHashDetails,
-  attestations: ActionableAttestation[],
-  withoutRevealing: boolean = false
+  attestations: ActionableAttestation[]
 ) {
-  Logger.debug(TAG + '@revealNeededAttestations', `Revealing ${attestations.length} attestations`)
-  const delayPeriod = 5000
+  Logger.debug(TAG + '@revealAttestations', `Revealing ${attestations.length} attestations`)
   let i = 0
+  const delayPeriod = 5000
```
tarikbellamine: Nit: should we define this at the top of the file with the same pattern as the other constants we use?

packages/mobile/src/identity/verification.ts
i1skn: We do start listen to messages before, but in case the actionable attestations list is changes we resubscribe :) 

See slightly above: https://github.com/celo-org/celo-monorepo/blob/22cd8449fe457101badcde844c292b0763ee4c8e/packages/mobile/src/identity/verification.ts#L259-L262

packages/mobile/src/identity/verification.ts
```diff
@@ -549,52 +626,83 @@ function* isCodeAlreadyAccepted(code: string) {
...
+    const result = yield call(
+      revealAttestation,
+      attestationsWrapper,
+      account,
+      phoneHashDetails,
+      attestation,
+      // TODO (i1skn): remove this method and uncomment revealNeededAttestations above
+      // when https://github.com/celo-org/celo-labs/issues/578 is resolved
+      // send messages with 5000ms delay on Android
+      Platform.OS === 'android' ? delayPeriod * i : 0
```
tarikbellamine: Isn't it more simple to add a `delay` or `sleep` call in this for loop instead of passing an extra `delayInMs` parameter? That would simplify the `revealAttestation` generator, eliminate the need for `i`, and make it really easy to eliminate the delay when we no longer need it

packages/mobile/src/identity/verification.ts
```diff
@@ -549,52 +626,83 @@ function* isCodeAlreadyAccepted(code: string) {
...
+  attestations: ActionableAttestation[]
 ) {
-  Logger.debug(TAG + '@revealNeededAttestations', `Revealing ${attestations.length} attestations`)
-  const delayPeriod = 5000
+  Logger.debug(TAG + '@revealAttestations', `Revealing ${attestations.length} attestations`)
   let i = 0
+  const delayPeriod = 5000
+  const reveals = []
+  for (const attestation of attestations) {
+    const result = yield call(
```
tarikbellamine: Nit: if I understand correctly, `result` is always a boolean. If so, consider renaming to `success` because `result` implies an outcome and makes me think it's either some data or `undefined`

packages/mobile/src/identity/verification.ts
tarikbellamine: Awesome, thanks! Will be great to have data on this

packages/mobile/src/identity/verification.ts
tarikbellamine: If I am understanding [this thread](https://github.com/react-navigation/react-navigation/issues/2882) correctly, it looks like this method will continue to push the same screen onto the top of the stack with new params.

If so, this probably has negative implications on performance as well as creating some unwanted consequences of breaking the "back" carat. Maybe you can use the `setParams` method instead?

packages/mobile/src/verify/VerificationInputScreen.tsx
tarikbellamine: Ok cool, just wanted to make sure this was discussed! No objection from me if this was the product/design decision.

packages/mobile/src/identity/verification.ts
i1skn: From here https://reactnavigation.org/docs/navigating/#navigate-to-a-route-multiple-times

> ... The navigate function roughly means "go to this screen", and if you are already on that screen then it makes sense that it would do nothing.

Also, in our case we will navigate to the same screen only if we are stuck on `VerificationLoadingScreen`.

From the same page:

> Each time you call push we add a new route to the navigation stack. When you call navigate it first tries to find an existing route with that name, and only pushes a new route if there isn't yet one on the stack.

So, I assume it does not add to the stack already existing routes when you navigate to them.

Lmk if that makes sense for you?

tarikbellamine: My understanding based on [this comment in the thread](https://github.com/react-navigation/react-navigation/issues/2882#issuecomment-656466559) was that if you're passing new params, then it would create a new screen.

To be honest, I'm not too sure about this either way. If you didn't notice any adverse effects when you were testing, I am ok with merging as is given we do this in other parts of the code base.

i1skn: So, I've logged `route.key` when navigating to the route with `{withoutRevealing: true}` and then again with `{withoutRevealing: false}` and key stays the same. I assume it does not create a new screen, but I might be wrong...

tarikbellamine (approved):
Awesome work!

packages/mobile/locales/en-US/onboarding.json
tarikbellamine: Just out of curiosity, why is that necessary?

packages/mobile/src/verify/VerificationEducationScreen.test.tsx
i1skn: Visually there is no difference in rendering, so I've decided to drop this test. The only difference that button click dispatch a slightly different event. Do you think it makes sense to test it here?

Lss-Ankit:
Hi @tarikbellamine Can you please give access to below id for figma screen (https://www.figma.com/) to verify this task
Account: Celoqa@lodestoneco.com

Lss-Ankit:
Hi @tarikbellamine I have verified this task test flight build (25) & Android internal build Android internal build 1.2.0(1004294313) and observe the following 
**Device:** iPhone 11 (14iOS), Samsung Galaxy A70(9.0)

**Test plan link:** https://docs.google.com/spreadsheets/d/1phHo7MUMQNLjuJOic1TlZwoHM2T0iecT07VLoKtiqMg/edit#gid=1211519342

**Need Clarification:**
- Blank fields is shown on First & second Confirm code fields when user go back and open confirm page 
- "Verification Error " message is shown when user paste third code from Resend messages
can you please confirm that  i need file  issues for the same

PR #5136, opened 2020-09-22
[Wallet] Change Podfile source

gnardini:
### Description

There's an issue in the CocoaPods CDN: https://github.com/CocoaPods/CocoaPods/issues/10078

which causes `pod install` to intermittently fail. The linked issue suggests reverting this change in a few days when everything stabilizes.


gnardini:
Some check didn't like the fix and it looks like it was fixed by CocoaPods: https://github.com/CocoaPods/CocoaPods/issues/10078 so closing this :)

Issue #4084, opened 2020-06-12
User avatars often show up as blank circles


Lss-Ankit:
**Frequency:** 100%
** Repro App version:** Celo integration build v1.8.2 (36)
**Does not repro:** app store build v1.8.0
**Repro on device:**iPhone XS max (13.1), iphone 6+ (12.4.5)
**Backup Key** 503 447 5958/ [Backup key](https://docs.google.com/spreadsheets/d/1Iy41GGA6e2pES2XHjBn7nb8X3n8zZsu7EW1jvwD2e70/edit#gid=0)

**Repro Steps:** 
1) Launch app and login with tested account. 
2) Tap on Hamburger menu. 
3.) Observe.
 
**Current Behavior:**  Blank circle is shown instead first letter of user name.
**Expected Behavior:** First letter should be shown. 
**Attachment:**[ ProfilePicIssue .png ](https://drive.google.com/file/d/1sIZa9Gg_nHfnkK8VRMiaBgvId_6hJR07/view?usp=sharing)

Lss-Ankit:
Verified with latest IOS test flight build v1.0.0(4) and observed that issue is still Repro
- Blank circle is shown instead first letter of user name on Profile pic from left panel menu.
- **Device Used:**  iPhone 6(12.4.7)

Lss-Ankit:
Verified with latest IOS test flight build v1.0.0 (15) and found that issue is still reproduced. 
- Blank circle is shown instead first letter of user name on Profile pic from left panel menu.
Note: issue does not repro on Android valora app build v1.0.0 (14) 
Verified devices: iPhone 7+ (13.3.1)

Lss-Ankit:
Verified with latest IOS test flight build v1.3.0 (27) and found that issue is still reproduced.

- Blank circle is shown instead first letter of user name on Profile pic from left panel menu.

Verified devices on: iPhone SE 2 (13.5), 

Does not repro on Android  build v1.3.0(1004294316) Google pixel Xl (8.1).

ole-nov:
will be fixed with CIP8 profile photo storage

Lss-Ankit:
Hi @aleksey-nov @nityas I hve Verified this issue with latest Android Internal Build V1.9.1(1004294332) & Test Flight Build V1.9.1(42) ) and found that issue is still reproduced.

- Random profile picture is shown instead of first letter of user name on Profile pic when user clicks on hamburger menu

[...]

PR #5154, opened 2020-09-23
[Wallet] Add display names of scanned QRs to cache

### LINKED ISSUE #1005: Users SBAT see the name of a person in a QR code transaction in the feed
### Expected Behavior

If you've scanned someone's QR code which has their name in it, you should be able to see their name labeled in feed txs

### Current Behavior

Just see sent or received

gnardini:
### Description

When scanning a QR, the link embedded on the QR has a display name. This PR stores that info on a cache to use it on the activity feed to show the name of the person we scanned the QR from.

### Tested

Read a QR code to make a transfer and saw the display name of the scanned QR on the transactions feed.

### Related issues

- Fixes #1005
Part of #4835


packages/mobile/src/send/utils.ts
```diff
@@ -171,21 +180,44 @@ export function showLimitReachedError(
   return showError(ErrorMessages.PAYMENT_LIMIT_REACHED, ALERT_BANNER_DURATION, translationParams)
 }
 
+// exported for testing
+export function* _updateRecipientCache(recipient: RecipientWithQrCode) {
```
tarikbellamine: Please correct me if I am wrong but I don't think we use the underscore convention as part of the wallet code base. I understand you want to show that this is a private helper function but I don't want it to be confusing because we aren't doing it anywhere else

packages/mobile/src/send/utils.ts
```diff
@@ -171,21 +180,44 @@ export function showLimitReachedError(
...
+  }
+  const addressToE164Number: AddressToE164NumberType = yield select(addressToE164NumberSelector)
+  const recipientCache: NumberToRecipient = yield select(recipientCacheSelector)
+
+  recipientCache[recipient.e164PhoneNumber] = {
+    contactId: '',
+    ...recipient,
+    kind: RecipientKind.Contact,
+  }
+  addressToE164Number[recipient.address] = recipient.e164PhoneNumber
```
tarikbellamine: This makes me nervous because it overwrites any existing mapping without having any type of check for security or correctness. I don't know the best way to mitigate this concern atm.

I am imagining a scenario where someone creates an account with a popular but unverified phone number. Anyone that scan's that QR, will have that phone number overwritten with the QR owner's address even though the phone number is not verified

tarikbellamine (commented):
Left a couple of comments. 

How would you feel about holding off on this PR until the next release? Personally, I don't feel like we've worked out the UX considerations enough to make this a "must have" for this release.

packages/mobile/src/send/utils.ts
gnardini: Yeah, I agree, I was a bit lazy tbh, I found this code in some old PR and took it from there. I don't think we should be using phone numbers from this without checking if they're verified, how about this:

Instead of reusing these caches that use phone numbers, create a new one that maps `address -> display name` that's used as fallback if none of the others have info about the transfer. This way the display name in the QR would only be used if the alternative is to show 'Unknown' in the feed.

gnardini:
> How would you feel about holding off on this PR until the next release?

No strong opinion, sounds good to me :)

packages/mobile/src/send/utils.ts
gnardini: Did the changes I suggested above, let me know what you think :)

i1skn (approved):
Great work! LGTM

packages/mobile/src/send/utils.ts
jeanregisser: It's not used a lot but we mention this in our [TypeScript style guide](https://github.com/celo-org/celo-monorepo/blob/master/packages/docs/celo-codebase/style-guide/typescript.md#exporting-variables-only-for-testing)

packages/mobile/src/identity/reducer.ts
```diff
@@ -42,6 +42,10 @@ export interface AddressToDataEncryptionKeyType {
   [address: string]: string | null // null means no DEK registered
 }
 
+export interface AddressToDisplayNameType {
+  [address: string]: string
```
jeanregisser:
```suggestion
  [address: string]: string | undefined
```

packages/mobile/src/send/utils.ts
```diff
@@ -178,14 +179,16 @@ export function* handleSendPaymentData(
...
     displayId: data.e164PhoneNumber,
     displayName: data.displayName || cachedRecipient?.displayName || 'anonymous',
     e164PhoneNumber: data.e164PhoneNumber,
     phoneNumberLabel: cachedRecipient?.phoneNumberLabel,
     thumbnailPath: cachedRecipient?.thumbnailPath,
     contactId: cachedRecipient?.contactId,
   }
+
+  yield put(addAddressToDisplayName(recipient.address, recipient.displayName))
   yield put(storeLatestInRecents(recipient))
```
jeanregisser: nit: I think it's an anti-pattern to dispatch multiple actions consecutively, but we're guilty of this in other places 😢 
This is because we have a bunch of actions that are just setters.

See https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters

packages/mobile/src/identity/reducer.ts
```diff
@@ -89,6 +93,8 @@ export interface State {
   e164NumberToAddress: E164NumberToAddressType
   e164NumberToSalt: E164NumberToSaltType
   addressToDataEncryptionKey: AddressToDataEncryptionKeyType
+  // Doesn't contain all known addresses, use only as a fallback.
+  addressToDisplayName: AddressToDisplayNameType
```
jeanregisser: Could you add a `TODO` note about removing this once we have CIP8 working in the app?

packages/mobile/__mocks__/react-i18next.js
```diff
@@ -34,17 +34,28 @@ const renderNodes = (reactNodes) => {
...
+const printParamInsteadOfKey = {
+  feedItemSentTitle: 'nameOrNumber',
+}
+const translationFunction = (key, params) => {
+  const paramToPrint = printParamInsteadOfKey[key]
+  return paramToPrint ? params[paramToPrint] || key : key
+}
+
+const useMock = [translationFunction, {}]
+useMock.t = translationFunction
```
jeanregisser, line 49: Nothing to do now, but it looks like we should revisit the decision of printing only the keys during unit test.
And print the actual EN strings.
What do you think?

jeanregisser (approved):
LGTM :shipit:  👍

packages/mobile/__mocks__/react-i18next.js
gnardini: That does sound like it would be better, it's one of the things I considered but I think there were some issues around values that may change (dates maybe? I think there was something else). They can be definitely normalised though and it'd be nice to use real strings :)

packages/mobile/src/send/utils.ts
gnardini: We can reuse the `storeLatestInRecents` action to avoid sending two with the same parameter :)

packages/mobile/src/send/utils.ts
tarikbellamine: So should we try to actively enforce this moving forward? There are plenty of places where we are exporting for testing only but I've never seen the underscore pattern used.

I'm happy either way but think we should strive for consistency.

packages/mobile/src/send/utils.ts
tarikbellamine: I like it! I think this is a good short-term solution until we get CIP8 integrated. Please remind whoever implements the CIP8 work to remove this cache.

Lss-Ankit:
Hi @gnardini We have verified task on latest Android play store internal build v1.3.0(1004294316) & Test Flight build v1.3.0(27) and found the following. 

- Scanned user name is shown on transaction feed when user sending an amount by scanning QR code. 
- Also, verify that User name is shown on transaction feed for receiver as well. 

However, we have updated below issues related to name as Unknown text is shown on transaction feed and profile pic as blank profile pic is shown: 
- https://github.com/celo-org/celo-monorepo/issues/4086
- https://github.com/celo-org/celo-monorepo/issues/4084

Please let us know if we need to verify any thing else. 
Thanks!

PR #5156, opened 2020-09-23
MetaTransactionWallet ContractKit wrapper

### LINKED ISSUE #22: ContractKit supports creating/submitting MetaTxs
Dev SBAT create both inner tx, meta-tx, and submit them to network using ContractKit

bowd:
### Description

Create a contract wrapper for the MetaTransactionWallet.
It also implements a couple of convenience methods that allow for MetaTransaction building, included as a call to it's own `executeTransactions` method in order to create a batch of operations.
These signed meta transactions can then be submitted by anybody.

#### Nonce

One consideration here is whether we want the wrapper to do any fancy logic to keep track of the on-chain nonce of MetaTransactions. I briefly considered it but I think it would just introduce more source of confusion. I think we should thus leave the nonce setting/management at the discretion of the consumer of the wrapper. 

### Tested

Specs included covering a full range of usage.

### Related issues

- Fixes https://github.com/celo-org/komenci/issues/22

### Backwards compatibility

New addition no bw comp concerns.


packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,279 @@
...
+  CeloTransactionObject,
+  numericToHex,
+  proxyCall,
+  proxySend,
+  stringIdentity,
+  toTransactionObject,
+  valueToInt,
+} from './BaseWrapper'
+
+export interface MTWTransaction {
```
asaj: Rather than define a new type, why not follow the approach that the `MultiSig` wrapper took, which was to take `TransactionObject`s?

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,279 @@
...
+/**
+ * Class that wraps the MetaTransactionWallet
+ */
+export class MetaTransactionWalletWrapper extends BaseWrapper<MetaTransactionWallet> {
+  /**
+   * Execute a transaction originating from the MTW
+   * Reverts if the caller is not a signer
+   * @param tx a MTWTransaction
+   */
+  executeTransaction(tx: MTWTransaction) {
```
asaj: nit: Explicitly define return types here and below

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,279 @@
...
+    )
+  }
+
+  /**
+   * Signs a meta transaction as EIP712 typed data
+   * @param mtx a MTWTransaction
+   * @returns smtx a MTWSignedMetaTransaction
+   */
+  async signMetaTransaction(
+    signer: Address,
```
asaj: Should the signer be given in the constructor? Or be part of the MTW state somehow?

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,279 @@
...
+      )
+    )
+  }
+
+  /**
+   * Signs a meta transaction as EIP712 typed data
+   * @param mtx a MTWTransaction
+   * @returns smtx a MTWSignedMetaTransaction
+   */
+  async signMetaTransaction(
```
asaj: Consider adding a `signMetaTransactions` function, that batches the transactions into a single `executeTransactions` call and then signs that as a meta transaction.

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,279 @@
...
+  executeTransactions(txs: MTWTransaction[]) {
+    return toTransactionObject(this.kit, this.buildExecuteTransactionsTx(txs))
+  }
+
+  /**
+   * Builds the TransactionObject for executeTransactions
+   * Reverts if the caller is not a signer
+   * @param txs An array of MTWTransactions
+   */
+  buildExecuteTransactionsTx(txs: MTWTransaction[]) {
```
asaj: Can we mark internal functions as private or protected? It will help clarify the wrapper API, of which >95% of the usage will be:
```
executeTransaction
executeTransactions
executeMetaTransaction
signMetaTransaction
signMetaTransactions
```

asaj: Or maybe the question should be, why make this function, as well as those below, public?
```
buildExecuteTransactionsWrapperTx
getMetaTransactionTypedData
buildMetaTxTypedData
```

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
asaj: Does the `TransactionObject` have a value and destination address? In `MultiSig` we just use it to encode the data but maybe we can do better.

bowd: Yeah I was thinking of a way to improve this. I guess I also wanted to support base network transfers in the contract API but I guess we can wrap those as contract calls to the Gold contract and not need to support that. 
It's true we can also extract the destination from the TransactionObject, it's there but it's not public in the typescript interface `tx._parent._address`, so I'm wondering if there are cases where that isn't there, but I think it should always be.

I'd still like a wrapper type but I'll refactor to something a bit cleaner and see how that looks like.

bowd:
@asaj Changed the interface to use TransctionObject like you suggested, it's cleaner, but I still kept a wrapper object because I don't think there's any way to encode a value transfer inside of the TransactionObject. And I like having a wrapper interface because I can have an Array of those for the functions that execute batches.

bowd:
One other approach would be to have something like:
```typescript
interface TransactionObjectWithValue<T> {
   txo: TransactionObject<T>
   value: BigNumber.Value
}

type TransactionInput = TransactionObjectWithValue<any> | TransactionObject<any>
```

And then have the functions accept `TransactionInput` and `Array<TransactionInput>` respectively, and they can then discriminate on the `value` attribute essentially. Which would make the API a bit cleaner instead of always passing in the wrapper type :-/

bowd:
Gave that idea a spin and I like it I think it has the best ergonomics of what I tried so far with the only downside of some input massaging here and there.

bowd:
Made another change to the wrapper API because I realised we need an easy way to execute transactions that were signed somewhere and then passed over the wire to another entity which is doing the submitting and ended up with three types of way to pass in txs.

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,289 @@
...
+  _signer?: Address
+  public async signer() {
+    if (this._signer === undefined) {
+      this._signer = await this.getSigner()
+    }
+    return this._signer
+  }
+
+  /**
+   * Turns any possible way to pass in a tranasction into the raw values
```
codyborn:
```suggestion
   * Turns any possible way to pass in a transaction into the raw values
```

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,289 @@
...
+   * that are actually required. This is used both internally to normalize
+   * ways in which transactions are passed in but also public in order
+   * for one instance of ContractKit to serialize a meta transaction to
+   * send over the wire and be consumed somewhere else.
+   * @param tx TransactionInput<any> union of all the ways we expect transactions
+   * @returns a RawTransactions that's serializable
+   */
+  public toRawTransaction(tx: TransactionInput<any>): RawTransaction {
+    if ('destination' in tx) {
```
codyborn: The string based check seems error prone to me. Can we do something like this?
```suggestion
    if ((tx as RawTransaction).destination) {
```

bowd: Could you elaborate on why the string check feels error prone?

Theoretically it wouldn't make sense to do a cast at that point because we're relying on the compiler's understanding of the object structure. So `tx` starts as being TransactionInput<any> which is a combination of 3 possible object structures. By checking whether props exist on the object we are discriminating between the types.

So the compile would complain if none of the branches had a `destination` prop and on the flip side it actually uses that if condition to discriminate, so inside of the first if statement `tx` becomes a `RawTransaction` automatically. 

So the point is we're letting the compiler tell us if our assertion that `destination` being a field of the input object is enough for it to determine that tx must be a `RawTransaction`. If one of the other branches were to add a `destination` field the compiler would start complaining but if we cast to `RawTransaction` we won't get that benefit and we'd be depriving the compiler of the ability to discriminate between fields.

---

Checking for the presence of a key feels ok to me as long as we're ensuring other data boundaries correctly. Aka we are not passing in manually cast data of uncertain structure - but that's true of any situation in which we're expecting a certain object structure and we might be shooting ourselves in the foot.

bowd:
@codyborn @asaj / anybody can I get a :shipit: on this 😺

packages/contractkit/src/wrappers/MetaTransactionWallet.test.ts
```diff
@@ -0,0 +1,262 @@
...
+testWithGanache('MetaTransactionWallet Wrapper', (web3) => {
+  MetaTransactionWallet.setProvider(web3.currentProvider)
+
+  const deployWallet = async (deployer: Address, signer: Address): Promise<Address> => {
+    const instance = await MetaTransactionWallet.new({ from: deployer })
+    await instance.initialize(signer, { from: deployer })
+    return instance.address
+  }
+
+  // Ganache returns 1 in chainId assembly code
```
m-chrzan, line 27: But the "official" network ID is set to something different, right? Definitely nothing to worry about in this PR, but wondering if this should be fixed in ganache eventually?

packages/contractkit/src/wrappers/MetaTransactionWallet.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  beforeAll(async () => {
+    accounts = await web3.eth.getAccounts()
+    walletSigner = accounts[0]
+    kit.defaultAccount = walletSigner
+    rando = accounts[1]
+    gold = await kit.contracts.getGoldToken()
+  })
+
+  beforeEach(async () => {
+    const walletAddress = await deployWallet(walletSigner, walletSigner)
```
m-chrzan: Probably not super necessary, but would making deployer and signer different addresses make for some extra interesting test cases?

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,292 @@
...
+      toRawTransaction(tx),
+      nonce,
+      await this.chainId()
+    )
+    // TODO: This should be cached by the CeloProvider and executed through the wallets.
+    //       But I think the GethNativeBridgeWallet currently doesn't support this, it's
+    //       in the works.
+    const signer = await this.signer()
+    const signature = await new Promise<string>((resolve, reject) => {
+      ;(this.kit.web3.currentProvider as Pick<HttpProvider, 'send'>).send(
```
m-chrzan: Do we do signing anywhere else in contractkit? This feels like something that could be abstracted out of this function, at least for code cleanup.

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,292 @@
...
+    this.contract.methods.setSigner
+  )
+
+  setEip712DomainSeparator: () => CeloTransactionObject<void> = proxySend(
+    this.kit,
+    this.contract.methods.setEip712DomainSeparator
+  )
+
+  /**
+   * Get an cache the chain ID -- assume it's static for a kit instance
```
m-chrzan: Typo: *and*

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,292 @@
...
+    )
+  }
+
+  /**
+   * Signs a meta transaction as EIP712 typed data
+   * @param tx a TransactionWrapper
+   * @param nonce Optional -- will query contract state if not passed
+   * @returns signature a Signature
+   */
+  async signMetaTransaction(tx: TransactionInput<any>, nonce?: number): Promise<Signature> {
```
tarikbellamine: Is there purposefully not a `public` or `private` distinction for this?

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,292 @@
...
+/**
+ * Turns any possible way to pass in a transaction into the raw values
+ * that are actually required. This is used both internally to normalize
+ * ways in which transactions are passed in but also public in order
+ * for one instance of ContractKit to serialize a meta transaction to
+ * send over the wire and be consumed somewhere else.
+ * @param tx TransactionInput<any> union of all the ways we expect transactions
+ * @returns a RawTransactions that's serializable
+ */
+export const toRawTransaction = (tx: TransactionInput<any>): RawTransaction => {
```
tarikbellamine, line 215: The way you have the conditional setup will work but I prefer to use `instanceof` when possible so it's explicit that we are routing based on type.

More info on `instanceof` if you need it: https://www.ibrahima-ndaw.com/blog/advanced-typescript-cheat-sheet/#type-guards

tarikbellamine (commented):
LGTM but I probably don't have enough context to give final approval

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
bowd: I agree in principle but in this situation all of the types in the union are interfaces, not classes so `instanceof` can't be used here. The link you shared also has a section on discriminating between interfaces which uses the `in` operator.

m-chrzan (approved):
Looks good, just a few small comments.

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
codyborn: Sorry for the late reply. I was worried that if `destination` string had a typo then the compiler wouldn't catch it as it would with the casting approach. However, based on what you're saying here, it sounds like that's not the case:
> compile would complain if none of the branches had a destination prop

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
```diff
@@ -0,0 +1,292 @@
...
+   * @param txs An array of TransactionInput
+   */
+  public executeTransactions(
+    txs: Array<TransactionInput<any>>
+  ): CeloTransactionObject<{ 0: string; 1: string[] }> {
+    const rawTxs: RawTransaction[] = txs.map(toRawTransaction)
+    const destinations = rawTxs.map((rtx) => rtx.destination)
+    const values = rawTxs.map((rtx) => rtx.value)
+    const callData = ensureLeading0x(rawTxs.map((rtx) => trimLeading0x(rtx.data)).join(''))
+    const callDataLengths = rawTxs.map((rtx) => trimLeading0x(rtx.data).length / 2)
```
codyborn: Optional: can be made into a util method

codyborn (approved):
LG thanks!

packages/contractkit/src/wrappers/MetaTransactionWallet.test.ts
bowd: Yeah that's my understanding as well, but I didn't dive deep into this. @asaj first bumped into this when working on the MetaTransactionWallet.sol implementation itself and used this static chainId in the specs there. I just took his word for it and did the same thing in the wrapper specs 😄

packages/contractkit/src/wrappers/MetaTransactionWallet.test.ts
bowd: Yeah that's true makes sense. I think I meant to do that from the beginning (the reason behind having separate arguments). But not sure why I ended up using the same one. Will add a different deployer so the specs cover more ground.

packages/contractkit/src/wrappers/MetaTransactionWallet.ts
bowd: Yeah agree I'll take another look to clean this up a bit.

bowd: I'm not sure where the best place to move it would be, tho. I still haven't gotten a good sense of how utils are distributed in contractkit. I could see this being exposed as a ContractKit method so we could do:
```
kit.signTypedData(signer, typedData)
```

But I'm not sure if we're ok with adding that to ContractKit as a toplevel method.

bowd: I made the change like this. I like the ergonomics on having that method exposed on `ContractKit` but I'll wait before merging this to get feedback on the choice.

PR #5177, opened 2020-09-23
[Wallet] Go back to VerificationEducation screen without updating the state

### LINKED ISSUE #5166: “Skip for now” button is shown instead on “Resume” & “I already received three codes” buttons on Confirm page when user tap on < button from Confirm Code pasting page
**Frequency:** 100%
**Repro App version:**  Android Valora Mainnet build v1.1.0 (1004294311), Test Flight build Valora app build  v1.1.0 (223) 
**Repro on device:** iPhone 11 (14 iOS), Pixel 3a (11)
**Repro Steps:**

1. Launch the Application and sign in with a test account 
2. Make sure user should able to see the “Receive payments directly to your phone number” notification in home page 
3. Tap on confirm Now button 
4. Tap on start button 
5. Make sure user should receive 3 code
6. Tap on < button on confirm pasting code page 
7. Observe skip for now button is shown 

**Investigation:** Resume & i already received three codes buttons is shown when user go back and and again tap on confirm button 

**Current Behavior:** “Skip for now” button is shown when user tap on < button from Confirm code roasting page 
**Expected Behavior:** Resume & i already received three codes buttons should show when user tap on < button from confirm pasting code page 

**Attachment:** [ResumeButtons.mp4](https://drive.google.com/file/d/1exYXcJcR2V0zdPdrEilBUUImtLBveCAY/view?usp=sharing) (Bug observed at 1:48 )

i1skn:
### Description

Add fetching verification state also on going back to VerificationEducation
screen, not only at the first time screen loads.

### Other changes

* Disable "Back" button on Verification screen during onboarding. If user landed there it means account is already created (phrase imported or invite redeemed).

### Tested

iOS simulator

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/5166

### Backwards compatibility

Yes


cmcewen (approved):
This looks good to me. Wondering if we can add some sort of test coverage for all the edge cases you've been working through - I know @gnardini has been working on e2es for verification, but just would be nice to make sure these fixes stay fixed 😄

gnardini (approved):
LGTM :)

For the edge cases, I think we could add a unit test for this rather than an e2e one, I see there are some for this screen but none with the `hideOnboardingStep` flag as true.

e2e tests for the verification flow are going to be very tough to get right and will probably be extremely flaky because of the SMS verification

Issue #4841, opened 2020-08-26
Address tech debt


jeanregisser:
- [ ] Consider using typed-redux-saga to fix recurrent issues with this
- [x] Continue removing obsolete code following the redesign
- [x] Simplify switch between envs (dev/prod)

PR #5191, opened 2020-09-24
[Wallet] Make it easier to build and run the different envs

### LINKED ISSUE #5294: Developers should be able to easily build and run the app for the different supported envs
### Expected Behavior

Developers should be able to easily build and run the app for the different supported envs.

### Current Behavior

Right now the command for building and running the app, `yarn dev:ios` or `yarn dev:android`, allows specifying a different network with the `-n $NETWORK` flag but it's still using `packages/mobile/.env` and updates the corresponding google services plist/json to set the realtime db url to use.
This still uses the deprecated `celo-org-mobile` project, so some functionalities like push notifications and payment requests won't work as expected.
`celo-mobile-mainnet` and `celo-mobile-alfajores` projects should be used instead depending on what network the app is running.

Also, it's not easy to run the other envs we have, for instance, the prod config from `packages/mobile/.env.mainnet`, the developer has to manually update things for this to work.

Similarly, it's not easy to run the app in release mode where the JS bundle and assets are directly included in the app.

jeanregisser:
### Description

This PR makes it easier to build the app in the different envs we have.
- We now have the following env files, which are mapped to Xcode schemes/Android flavors:
  - `.env.alfajores`
  - `.env.alfajoresdev`
  - `.env.mainnet`
  - `.env.mainnetdev`
  - `.env.test`
- Supports both `debug`/`release` for all envs (for release builds, the js bundle and assets are included in the app package)
- Supports "dev" variants for day-to-day use by developers, can be installed alongside the store apps
- Usage: `yarn dev:ios -e %envName%` or `yarn dev:android -e %envName%`, (`%envName%` can be `alfajores`, `alfajoresdev`, `mainnet`, `mainnetdev` or `test`), add the `-r` flag to produce a release build which doesn't need the packager running
- Other `run_app.sh` script improvements:
  - can build iOS too from the command line (no need to have Xcode running 🎉) and now launches the packager in a different terminal window respecting the user choice of terminal
  - doesn't need to dynamically update the google services plist anymore. This should prevent using the wrong firebase db
  - App can still be launched through Xcode: Screenshot 2020-09-25 at 10 57 28
  - App can also be run from Android Studio if needed: Screenshot 2020-09-25 at 15 22 02
- Removed unused `pilot` and `integration` networks
- Removed google service configs referencing the deprecated `celo-org-mobile` Firebase project.

Finally it is relatively simple to add new envs if needed.

### Other changes

- Upgraded `react-native-config`

### Tested

Successfully ran:
- `yarn dev:ios`
- `yarn dev:ios -e aflajoresdev`
- `yarn dev:ios -e aflajores`
- `yarn dev:ios -e mainnetdev`
- `yarn dev:ios -e mainnet`
- `yarn dev:ios -r`
- `yarn dev:android`
- `yarn dev:android -e aflajoresdev`
- `yarn dev:android -e aflajores`
- `yarn dev:android -e mainnetdev`
- `yarn dev:android -e mainnet`
- `yarn dev:android -r`
- `yarn test:e2e:ios`
- `yarn test:e2e:ios -r`
- `yarn test:e2e:android` (release mode (`-r`) option doesn't work yet as it fails at runtime due to proguard not keeping some Kotlin classes used by detox).
- `bundle exec fastlane ios mainnet` (though manually disabling upload)
- `bundle exec fastlane android mainnet` (though manually disabling upload)

### Related issues

- Part of #4841
- Fixes #5294

### Backwards compatibility

Yes


packages/mobile/.env.mainnetdev
```diff
@@ -1,18 +1,21 @@
-ENVIRONMENT=integration
-DEFAULT_TESTNET=alfajores
-SMS_RETRIEVER_APP_SIGNATURE=l5k6LvdPDXS
+DEFAULT_TESTNET=mainnet
+SMS_RETRIEVER_APP_SIGNATURE=bU9E4ctGtIW
```
jeanregisser: @i1skn Do we need a different signature for this one?

gnardini (approved):
This looks amazing!! Do you think we'll be able to use release builds for e2e tests now? That would reduce some of the flakiness that is caused by intermittent warnings that cover the bottom of the screen.

tarikbellamine (commented):
This is awesome! Please open an github issue for this to close.

jeanregisser:
> This looks amazing!! Do you think we'll be able to use release builds for e2e tests now? That would reduce some of the flakiness that is caused by intermittent warnings that cover the bottom of the screen.

Yes, iOS e2e tests in release mode already works!
Android is still problematic but very close. 
I manually made the e2e test release build work by enabling multidex, but then it failed at runtime due to proguard not keeping some Kotlin classes used by detox. Which could be easily solved by disabling proguard in that configuration.
Though there might be other issues 👀

PR #5195, opened 2020-09-24
[Wallet] Add exchanges to Cash In / Out options

### LINKED ISSUE #4837: Cash in / out changes
- [ ] ~Wyre / Indacoin integration~ (on hold)
- [ ] Surfacing exchanges in the feed

annakaz:
### Description

Add list of exchanges to Cash In / Out options. These links bring the user to the exchange's website with celo or cUSD if possible, otherwise just the exchange website. 

Longer term we may want to be able to update the list of exchanges dynamically, for now they are hardcoded in the config file

The link-arrow icon is not considered protected so I didn't encrypt it in the valora assets repository

### Other changes

Updated format of account key to have second line centered (see screenshots below) based on conversation with @nityas and @coreycelo 

### Tested

On emulator (see screenshots below) and updated snapshot tests

Did not test purchasing on each exchange, as that is outside of the app

### Related issues

- Fixes #4837 

### Backwards compatibility

Yes

Screen Shot 2020-09-24 at 3 00 20 PM
Screen Shot 2020-09-24 at 3 00 02 PM
Screen Shot 2020-09-24 at 2 59 46 PM


packages/mobile/src/app/saga.ts
```diff
@@ -41,8 +41,8 @@ export function* appInit() {
 
   if (isDeprecated) {
     Logger.warn(TAG, 'App version is deprecated')
-    navigate(Screens.UpgradeScreen)
-    return
+    // navigate(Screens.UpgradeScreen)
```
gnardini: We probably don't want this commented out :)

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -24,6 +24,10 @@ function FiatExchange() {
     navigation.navigate(Screens.FiatExchangeAmount, { isAddFunds: true })
   }
 
+  function goToExternalExchanges() {
+    navigation.navigate(Screens.ExternalExchanges, { isAddFunds: false })
```
gnardini: I think the `isAddFunds` doesn't belong here

packages/mobile/src/fiatExchanges/ExternalExchanges.tsx
```diff
@@ -0,0 +1,111 @@
...
+  return (
+    <ScrollView style={styles.container}>
+      <SafeAreaView>
+        <Text style={styles.pleaseSelectProvider}>{t('youCanTransfer')}</Text>
+        <View style={styles.accountNumberContainer}>
+          <View style={styles.accountNoTextContainer}>
+            <Text style={styles.accountNoText}>Account</Text>
+            <Text style={styles.accountNoText}>No.</Text>
+          </View>
+          <AccountNumber address={account || ''} />
```
gnardini, line 59: I wonder how many people realize that they can copy the address by tapping on it and wether adding something related to that to the `youCanTransfer` changes the number of taps on the address.

gnardini (approved):
Looks great! 🚀

packages/mobile/src/account/FiatExchange.tsx
annakaz: Thanks for catching this!

packages/mobile/src/fiatExchanges/ExternalExchanges.tsx
annakaz: Good call - will update the copy after a check with Nitya and Corey

packages/mobile/src/app/saga.ts
annakaz: 😅 fixed

packages/mobile/src/fiatExchanges/ExternalExchanges.tsx
annakaz: Updated!

packages/mobile/src/account/FiatExchange.tsx
```diff
@@ -24,6 +24,10 @@ function FiatExchange() {
     navigation.navigate(Screens.FiatExchangeAmount, { isAddFunds: true })
   }
 
+  function goToExternalExchanges() {
+    navigation.navigate(Screens.ExternalExchanges)
```
tarikbellamine, line 28: Should we use our custom `navigate` function from NavigationService instead?

packages/mobile/src/analytics/Properties.tsx
```diff
@@ -577,6 +578,12 @@ interface CeloExchangeEventsProperties {
   }
 }
 
+interface FiatExchangeEventsProperties {
+  [FiatExchangeEvents.external_exchange_link]: {
+    link: string
```
tarikbellamine, line 584: Can we pass the exchange name instead of the link? Think that would be easier for us to do analysis on

packages/mobile/src/analytics/Events.tsx
```diff
@@ -288,6 +288,10 @@ export enum CeloExchangeEvents {
   celo_withdraw_error = 'celo_withdraw_error', // when there's an error on the withdrawal transaction
 }
 
+export enum FiatExchangeEvents {
```
tarikbellamine, line 291: I think this should also be added to `AnalyticsEventType` at the bottom of this file

packages/mobile/src/fiatExchanges/ExternalExchanges.tsx
```diff
@@ -0,0 +1,111 @@
...
+import AccountNumber from 'src/components/AccountNumber'
+import BackButton from 'src/components/BackButton.v2'
+import { EXCHANGE_PROVIDER_LINKS } from 'src/config'
+import i18n from 'src/i18n'
+import LinkArrow from 'src/icons/LinkArrow'
+import { emptyHeader } from 'src/navigator/Headers.v2'
+import { navigateToURI } from 'src/utils/linking'
+import { currentAccountSelector } from 'src/web3/selectors'
+
+export const externalExchangesScreenOptions = () => {
```
tarikbellamine, line 20: I believe the pattern we are trying to move towards is including the navigation options as a property of the component. See the Welcome screen pattern as an example

packages/mobile/src/fiatExchanges/ExternalExchanges.tsx
```diff
@@ -0,0 +1,111 @@
...
+          <View style={styles.accountNoTextContainer}>
+            <Text style={styles.accountNoText}>Account</Text>
+            <Text style={styles.accountNoText}>No.</Text>
+          </View>
+          <AccountNumber address={account || ''} />
+        </View>
+        <View style={styles.providersContainer}>
+          {providers.map((provider, idx) => {
+            return (
+              <ListItem key={idx} onPress={goToProvider(provider.link)}>
```
tarikbellamine: I think using index as a key is generally not advised: https://reactjs.org/docs/lists-and-keys.html

How about we use the provider link or name instead?

tarikbellamine (changes requested):
Looks mostly good. Had a couple of suggestions that I'd love to get addressed.

Pulled the branch and tested on iOS and Android devices and it looked/worked as expected!

packages/mobile/src/fiatExchanges/ExternalExchanges.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const providers = EXCHANGE_PROVIDER_LINKS // TODO Dynamically fetch exchange provider links so they can be updated between releases
+
+  return (
+    <ScrollView style={styles.container}>
+      <SafeAreaView>
+        <Text style={styles.pleaseSelectProvider}>{t('youCanTransfer')}</Text>
+        <View style={styles.accountNumberContainer}>
+          <View style={styles.accountNoTextContainer}>
+            <Text style={styles.accountNoText}>Account</Text>
+            <Text style={styles.accountNoText}>No.</Text>
```
tarikbellamine, line 57: I think we need to run this text through the t function to get them properly translated. I believe we also have an `AccountNumber` component purpose built for this.

annakaz: This matches what we're using in the drawer navigator, as we've chosen not to translate it: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/navigator/DrawerNavigator.tsx#L166
The AccountNumber component doesn't contain the "Account No." text as the formatting we'd like relative to the number changes

packages/mobile/src/account/FiatExchange.tsx
annakaz: We don't consistently use either throughout the codebase, but I agree that it would make sense to settle on one throughout

packages/mobile/src/analytics/Properties.tsx
annakaz: We were going back and forth on what link to include (whether to link directly to the exchange site, or to CELO or cUSD within the exchange site), so I think this detail would be helpful as we might change the link for a given exchange. Can include the exchange too

packages/mobile/src/analytics/Events.tsx
annakaz: Good catch, updated

packages/mobile/src/fiatExchanges/ExternalExchanges.tsx
annakaz: Interesting, we use this in the FiatExchange screen as well where there's not as clear an index, but happy to update here with the provider name

packages/mobile/src/fiatExchanges/ExternalExchanges.tsx
annakaz: For now will keep this consistent with the rest of the fiat exchange-related flows but agree that we should have something consistent across the codebase

tarikbellamine (approved):

Lss-Ankit:
@tarikbellamine I have verified this task using android internal build 1.2.0(1004294313) & test flight build (25) and found the following 
Devices: iPhone 11 (14iOS), Samsung Galaxy A70(9.0)

Test plan Link: https://docs.google.com/spreadsheets/d/1phHo7MUMQNLjuJOic1TlZwoHM2T0iecT07VLoKtiqMg/edit#gid=756061759 

Please let me know if  i need to verify anything else for this task.
Thanks !

PR #5198, opened 2020-09-25
[Wallet] Validate mnemonic in the app's language

### LINKED ISSUE #5192: Mnemonic invalid error for non-English mnemonics
### Expected Behavior
Users SBAT to validate their mnemonic regardless of langauge.

### Current Behavior
If a user has a Spanish mnemonic, they run into the "invalid mnemonic" error trying to import their account.

gnardini:
### Description

Validating the mnemonic was defaulting to english, so it was not possible to restore accounts created in Spanish or other languages.

It is possible now to restore in Spanish, but only if the app's language is set to Spanish. I wonder if we could do something smarter and figure out the language of the mnemonic based on the first word/s so we don't depend on the language chosen by the user.

### Tested

On the simulator

### Related issues

- Fixes #5192

### Backwards compatibility

Yes


jeanregisser (approved):
💯 Thanks for fixing this critical issue!

tarikbellamine (changes requested):
Great catch! Would love to get this in 1.2.0 if the issue was actually that any non-English mnemonics were resulting in errors.

Can we please add some copy to explain to the user that their app language must match their mnemonic language? That should help mitigate that confusion for now.

tarikbellamine (approved):
Tested on both iOS and Android just in case and everything looks good. Great work :)

Lss-Ankit:
@tarikbellamine I have verified this issue on latest test flight build (25) & Android internal build Android internal build 1.2.0(1004294313) and observe issue is still repro 
Device: iPhone 11 (14iOS), Samsung Galaxy A70(9.0)
**Attachment:** [Validatemnemonic.mp](https://drive.google.com/file/d/1wQtjCOuRG5kCwbxKR2POkorcsyvqmmYS/view?usp=sharing)

gnardini:
@Lss-Ankit Thanks for testing this! What you see is expected, this change makes it so you can only use an account key of the same language you chose for the app. So, if the app is in Spanish you need a spanish Account Key. To create one, you need to create an account with the language as spanish and the created account key will be in spanish.

Lss-Ankit:
Hi @gnardini thank you so much for giving more information about this task 
- I have again verified this task using latest test flight build (25) & Android internal build Android internal build 1.2.0(1004294313) by following scenarios 
- User is able to signin using English account key when user has set up account in English language after sign up.
- User is able to signin with Spanish account key when user set up account in Spanish language after sign up
- User not able to signin with English Account key when user has set up account in Spanish language after sign up. 
(User friendly message is shown :Clave de Cuenta invalida  ) 
- User not able to signin with Spanish Account key when user has set up account in English language after sign up. 
(User friendly message is shown :Invalid account key ) 
- User is able to sign up without invite code when user set app language to Spanish 
- User is able to sign up without invite code when user set app language to English

Can you please let me know if i need to test anything else Thanks!

tarikbellamine:
Sounds like you've covered everything @Lss-Ankit. I think there is no further action needed on this PR :)

Lss-Ankit:
Thanks @tarikbellamine for the confirmation.

PR #5201, opened 2020-09-25
Add check to CI deps audit for deps changed

yorhodes:
### Description

- Moves `yarn audit` vulnerabilities check from blocking CI to cronjob
- Makes `yarn audit` run in CI when dependencies have changed

### Other changes

- outputs yarn audit results in human readable format in CI

### Tested

Screen Shot 2020-10-08 at 4 19 20 PM

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,28 +93,7 @@ commands:
       - run:
           name: Fail if generated dependency graph doesn't match committed
           command: ./scripts/ci_check_dependency_graph_changed.sh
-      - run:
-          name: Fail if there are vulnerabilities in packages
-          command: |
-            generate_post_data()
-            {
-              cat <<EOF
-            {
-              "summary": "Vulnerability found in CI during yarn audit",
-              "details": "$REPORT",
-              "targets": [{
-                "type": "EscalationPolicy",
-                "slug": "pol-bu0pGzGe3mMQ1W8D",
-              }]
-            }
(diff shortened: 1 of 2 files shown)


aaronmgdr (approved):

.circleci/config.yml
```diff
@@ -96,8 +96,10 @@ commands:
       - run:
           name: Fail if there are vulnerabilities in packages
           command: |
-            /bin/bash -c 'yarn audit --groups dependencies --level high; [[ $? -ge 8 ]] && exit 1 || exit 0'
-
+            if [ -e ~/.tmp/yarn_deps_have_changed ]; then
```
gnardini: This looks good to me, but it will fail if any dependencies change, right? So if dependency `A` starts failing it will not be possible to add or modify any others until `A` is fixed?

It's still an improvement even if it works that way :)

gnardini (approved):
Thanks for the quick fix! 👏

.circleci/config.yml
yorhodes: yes, very true

cmcewen (changes requested):
i would not feel comfortable merging this unless we have another plan to ensure vulnerabilities get addressed immediately

aaronmgdr:
i believe we decided not to do this so im gonna hit close

PR #5208, opened 2020-09-26
[Wallet] Check for vulnerable webview version

etuleu:
### Description

Addressing vulnerability https://github.com/react-native-webview/react-native-webview/security/advisories/GHSA-36j3-xxf7-4pqg.
We use WebView for the MoonPay screen. The current workaround is to launch the url outside of the application if we are running on a vulnerable version (less than 83.0.4103.106).

### Other changes

None

### Tested

On the emulator manually adjusting the version string.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


packages/mobile/src/fiatExchanges/MoonPay.tsx
```diff
@@ -59,6 +62,19 @@ function FiatExchangeWeb({ route }: Props) {
...
 
+  // XSS vulnerability in Android WebView
+  // https://github.com/react-native-webview/react-native-webview/security/advisories/GHSA-36j3-xxf7-4pqg
+  if (
+    uri &&
+    Platform.OS === 'android' &&
+    version &&
+    compareVersions.compare(version, '83.0.4103.106', '<')
+  ) {
+    navigateToURI(uri)
```
jeanregisser, line 73: This side effect can't happen in render and should be wrapped in a `useEffect` instead.
I think this logic should be moved to where we navigate to this screen instead.

jeanregisser (changes requested):
Thanks for looking into this!

jeanregisser:
From what I understand from the report in https://alesandroortiz.com/articles/uxss-android-webview-cve-2020-6506/
We're not vulnerable to this given we don't have any WebView which allows loading arbitrary URLs.

Given this, I wouldn't change the code we have. Except for bypassing the CI check for this specific vulnerability.

etuleu:
Based on the call today we decided to fork the react-native-webview since the vulnerability only affects apps that allow navigating to any URLs, some of which could be compromised. @gnardini is going to look into the fork. Closing this PR.

PR #5210, opened 2020-09-28
ODIS monitoring improvements + Komenci support

alecps:
### Description

Komenci Changes: Update the process by which ODIS signers calculate quota to include balances and txCounts of walletAddresses. Before quota is calculated the celoBalance, cusdBalance and txCount inputs are summed across the accountAddress and walletAddress. This change is necessary so that users who are onboarded with komenci can query odis even though their balance is held in a walletAddress. 

Monitoring Changes: Adds a monitoring bot that regularly queries the ODIS combiner to confirm that the system is functioning properly. Adds structured logging across the entire phone-number-privacy package. Updates Azure templates to use a user defined identity to fix issue where key retrieval was failing on restart. 

### Other changes

Refactors various logic into the common package 

### Tested

On alfajoresstaging

### Related issues

NA

### Backwards compatibility

Yes


packages/phone-number-privacy/combiner/src/signing/get-threshold-signature.ts
```diff
@@ -225,7 +231,7 @@ function requestSignature(
 
 function getMajorityErrorCode(errorCodes: Map<number, number>) {
   if (errorCodes.size > 1) {
-    logger.error(ErrorMessage.INCONSISTENT_SIGNER_RESPONSES)
+    logger.error({ errorCodes }, ErrorMessage.INCONSISTENT_SIGNER_RESPONSES)
```
codyborn, line 234: Based on [Bunyan docs](https://github.com/trentm/node-bunyan#log-method-api), it sounds like this should be a named parameter.
```suggestion
    logger.error({ errorCodes: errorCodes }, ErrorMessage.INCONSISTENT_SIGNER_RESPONSES)
```

packages/phone-number-privacy/combiner/src/signing/get-threshold-signature.ts
```diff
@@ -180,7 +185,7 @@ function logResponseDiscrepancies(responses: SignMsgRespWithStatus[]) {
           totalQuota: response.signMessageResponse.totalQuota,
         }
       })
-      logger.error(`Discrepancy found in signers' measured quota values ${values}`)
+      logger.error({ values }, `Discrepancy found in signers' measured quota values`)
```
codyborn, line 188: I think this should be named as well.

packages/phone-number-privacy/common/src/utils/authentication.ts
```diff
@@ -24,7 +24,7 @@ export async function authenticateUser(
   // https://tools.ietf.org/html/rfc7235#section-4.2
   const messageSignature = request.get('Authorization')
   const message = JSON.stringify(request.body)
-  const signer = request.body.account
+  const signer = JSON.stringify(request.body.account)
```
codyborn: Why was this necessary? Just wondering because in the case that this wasn't working before, we shouldn't take a dependency on this published version of pgpnp-common for this method.

packages/phone-number-privacy/common/src/utils/config-utils.ts
```diff
@@ -1,5 +1,21 @@
...
+
+export function fetchEnv(name: string): string {
+  if (process.env[name] === undefined || process.env[name] === '') {
+    throw new Error(`ENV var '${name}' was not defined`)
+  }
+  return process.env[name] as string
+}
+
+export function fetchEnvOrDefault(name: string, defaultValue: string): string {
+  return process.env[name] === undefined || process.env[name] === ''
```
codyborn, line 18:
```suggestion
  return process.env[name] || process.env[name] === ''
```

packages/attestation-service/src/metrics.ts
```diff
@@ -1,72 +0,0 @@
-import { Counter, Gauge } from 'prom-client'
```
codyborn: Was this file intentionally removed?

packages/phone-number-privacy/monitor/src/query.ts
```diff
@@ -0,0 +1,27 @@
...
+contractKit.addAccount(privateKey)
+contractKit.defaultAccount = accountAddress
+
+export const queryOdisForSalt = () => {
+  const authSigner: AuthSigner = {
+    authenticationMethod: OdisUtils.Query.AuthenticationMethod.WALLET_KEY,
+    contractKit,
+  }
+
+  return OdisUtils.PhoneNumberIdentifier.getPhoneNumberIdentifier(
```
codyborn, line 21: Can you pass in a client version string with something that indicates this is coming from a canary test and not the mobile client (ex. "canary:1.0.0")? This way we can tell from the Combiner logs that the out-of-quotas are not coming from real users.

packages/phone-number-privacy/monitor/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@celo/phone-number-privacy-monitor",
+  "version": "1.22.5",
```
codyborn: Since this is new, we can start it on 1.0.0.

packages/phone-number-privacy/signer/src/database/wrappers/account.ts
```diff
@@ -76,7 +75,7 @@ export async function getDidMatchmaking(account: string): Promise<boolean> {
  * Set did matchmaking to true in database.  If record doesn't exist, create one.
  */
 export async function setDidMatchmaking(account: string) {
-  logger.debug('Setting did matchmaking')
+  logger.debug({ account }, 'Setting did matchmaking')
```
codyborn, line 81: Should be a named param?

packages/phone-number-privacy/signer/src/database/wrappers/request.ts
```diff
@@ -12,9 +11,10 @@ export async function getRequestExists(
   request: GetBlindedMessagePartialSigRequest
 ): Promise<boolean> {
   if (!request.timestamp) {
+    logger.debug('request does not have timestamp')
     return false // TODO(Alec) make timestamps required
   }
-  logger.debug('Checking if request exists')
+  logger.debug({ request }, 'Checking if request exists')
```
codyborn, line 17: Same here

packages/phone-number-privacy/common/src/utils/logger.ts
```diff
@@ -1,12 +1,27 @@
-// tslint:disable: no-console
+import Logger, { createLogger, levelFromName, LogLevelString, stdSerializers } from 'bunyan'
+import bunyanDebugStream from 'bunyan-debug-stream'
+import { createStream } from 'bunyan-gke-stackdriver'
```
codyborn, line 3: Did you see who created this package?
https://www.npmjs.com/package/bunyan-gke-stackdriver
Thanks @mcortesi 😄

packages/phone-number-privacy/signer/.env
```diff
@@ -17,4 +17,13 @@ DB_USE_SSL=true
 KEYSTORE_TYPE=MockSecretManager
 KEYSTORE_GOOGLE_PROJECT_ID=mockProjectId
 KEYSTORE_GOOGLE_SECRET_NAME=mockSecretName
-KEYSTORE_GOOGLE_SECRET_VERSION=latest
\ No newline at end of file
+KEYSTORE_GOOGLE_SECRET_VERSION=latest
+# Options: json, human (default), stackdriver
+LOG_FORMAT=stackdriver
```
codyborn, line 22: We run our signers in Azure Container Instance and pipe the logs to [Azure Log Analytics](https://portal.azure.com#@7cb7628a-e37c-4afb-8332-2029e418980e/blade/Microsoft_Azure_Monitoring_Logs/LogsBlade/resourceId/%2Fsubscriptions%2F97e2b592-255b-4f92-bce0-127257163c36%2Fresourcegroups%2Fpgpnp-alfajores%2Fproviders%2Fmicrosoft.operationalinsights%2Fworkspaces%2Fpgpnp-alfajores/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA3POzytJzMxLLfLMKy5JzEtO9clPj3f2AQCR58T%252FFwAAAA%253D%253D/timespan/P1D). We should double check that this formatting is compatible with [our alerting](https://portal.azure.com/#@mcelo.onmicrosoft.com/resource/subscriptions/97e2b592-255b-4f92-bce0-127257163c36/resourceGroups/pgpnp-alfajores/providers/Microsoft.OperationalInsights/workspaces/pgpnp-alfajores/alerts).

packages/phone-number-privacy/signer/src/index.ts
```diff
@@ -18,6 +18,6 @@ async function start() {
 }
 
 start().catch((e) => {
-  logger.error('Fatal error occured. Exiting', e)
+  logger.error({ e }, 'Fatal error occured. Exiting')
```
codyborn:
```suggestion
  logger.error({ err: e }, 'Fatal error occured. Exiting')
```

packages/phone-number-privacy/combiner/src/index.ts
```diff
@@ -1,31 +1,36 @@
...
 
+require('dotenv').config()
+
 // DEPRECATED, TODO: Remove once clients are all on contract kit version 4.11
 export const getDistributedBlindedSalt = functions
   .region('us-central1', 'europe-west3')
-  .https.onRequest(async (request, response) => {
-    logger.info('Begin getDistributedBlindedSalt request')
-    return handleGetBlindedMessageSig(request, response)
+  .https.onRequest(async (req, res) => {
```
codyborn, line 12: In general, it might be useful to include a sessionId/correlationId in each logging event so that we can connect the events together after the fact. We don't have to do it in this PR but just something to think about.

codyborn (changes requested):
Nice work! Added some small feedback.

packages/phone-number-privacy/combiner/src/signing/get-threshold-signature.ts
alecps: I *think* that the field will assume the name of the variable based on how this is done in the attestation-service https://github.com/celo-org/celo-monorepo/blob/e1a6a102dd1654a6774d47198d29e8ccb2360887/packages/attestation-service/src/db.ts#L105 I'll make sure to check.

packages/attestation-service/src/metrics.ts
alecps: Nope! Good catch 😅

packages/phone-number-privacy/common/src/utils/logger.ts
alecps: 😂 🙏 a true hero

packages/phone-number-privacy/combiner/src/index.ts
alecps: Good idea. Are you thinking something like each request has an id so that we can link all the logs originating from that request together?

packages/phone-number-privacy/common/src/utils/config-utils.ts
alecps: Did you mean to put '!' in front of process.env? I think we might want to be careful in case of env variables with the value 'false' or '0'. I *think* js always treats these as strings so probably fine.

packages/phone-number-privacy/common/src/utils/authentication.ts
alecps: It probably wasn't necessary come to think of it. I was just trying to make sure that we didn't try and give bunyan something other than a string there but I don't think it matters

packages/phone-number-privacy/signer/.env
alecps: Good idea. I'll test on staging and use the 'json' format if not

packages/phone-number-privacy/monitor/src/query.ts
alecps: Good idea

.env
```diff
@@ -175,8 +175,3 @@ VOTING_BOT_SCORE_SENSITIVITY=5
 # is not easily represented.
 # This schedule is "every 15 minutes", which is close enough. For some epochs, it will run twice
 VOTING_BOT_CRON_SCHEDULE="*/15 * * * *"
-
-# ODIS Service Tests
-ODIS_SIGNER_SERVICE_URL=https://staging-pgpnp-signer0.azurefd.net
```
codyborn, line 180: We should keep these here so that Vscode debugging works. We can have a copy in both locations if needed.

packages/phone-number-privacy/combiner/src/signing/get-threshold-signature.ts
alecps: I've verified that the variable name becomes the name of the parameter

.env
alecps: 👍

packages/phone-number-privacy/signer/azure-templates/container-parameters.json
```diff
@@ -90,6 +98,9 @@
         },
         "logAnalyticsWorkspaceKey": {
             "value": "TODO"
-        }
+        },
+        "userAssignedIdentity": {
```
codyborn, line 102: Thanks!! Can you also update the `azure-templates/README.md` with the prerequisite step to deploy the UserAssignedIdentity?

alecps: done!

packages/phone-number-privacy/signer/src/signing/query-quota.ts
```diff
@@ -80,25 +90,41 @@ export async function getRemainingQueryCount(
  * If the caller is not verified, they must have a minimum balance to get the unverifiedQueryMax.
  */
 async function getQueryQuota(account: string, hashedPhoneNumber?: string) {
-  if (hashedPhoneNumber && (await isVerified(account, hashedPhoneNumber))) {
-    logger.debug('Account is verified')
-    const transactionCount = await getTransactionCountFromAccount(account)
-    return (
+  let walletAddress = await getWalletAddress(account)
+  logger.debug({ account, walletAddress }, 'begin getQueryQuota')
+  if (account === walletAddress) {
```
codyborn: It's probably safest to do a normalized comparison
```suggestion
  if (account.toLowerCase() === walletAddress.toLowerCase()) {
```

codyborn (approved):
Looks great!

PR #5213, opened 2020-09-28
Allow ignoring known vulnerabilities

### LINKED ISSUE #5220: Allow ignoring known vulnerabilities
### Expected Behavior

We have a CI step that checks for vulnerabilities so we are aware of them. We should be able to ignore one if we already know about it and are working on it or have already mitigated it's effects.

### Current Behavior

There's no way to mark a vulnerability as "known" so that CI isn't blocked.

gnardini:
### Description

There is currently a vulnerability on `react-native-webview` that is blocking merging.
While we work on analysing the issue, we want to unblock CI. This PR adds a mechanism to add vulnerabilities as "known" so that CI doesn't fail.

Fixes #5220


tarikbellamine (approved):
LGTM! Thanks for taking this on!

PR #5222, opened 2020-09-28
Add redeem invite code and verification e2e tests

### LINKED ISSUE #741: Add e2e tests for verifying a phone number
### Expected Behavior

There are e2e tests that cover verifying a phone number.

### Current Behavior

There was an attempt at adding e2e tests for the flow here: https://github.com/celo-org/celo-monorepo/pull/5222

However, that branch hasn't worked since the Komenci changes. Mainly because of the recaptcha, we can't solve the recaptcha in e2e tests, there was an idea at some point of whitelisting a key on alfajores and using that one on e2e tests, but that hasn't been implemented yet.

We can reuse part of that code when the above or some other alternative is live.

### LINKED ISSUE #5223: Add e2e tests for redeeming invite code and verifying a phone number
### Expected Behavior

There are e2e tests that cover redeeming an invite code and verifying a phone number.

### Current Behavior

There are no such tests.

Part of #4836

gnardini:
### Description

Adds a test that generates an invite code using celotool, redeems it to get funds and verifies the phone number.

### Related issues

- Fixes #5223


packages/mobile/e2e/src/RedeemInviteAndVerify.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    if (!RUN_VERIFICATION_TEST) {
+      return
+    }
+    for (let i = 0; i < 3; i++) {
+      await element(by.id('Education/progressButton')).tap()
+    }
+
+    await element(by.id('CreateAccountButton')).tap()
+
+    // Accept Terms
```
i1skn, line 24: I wonder if we can re-use https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/e2e/src/usecases/NewAccountOnboarding.js here?

gnardini:
This branch hasn't worked since the Komenci changes :( Mainly because of the recaptcha, we can't solve the recaptcha in e2e tests, there was an idea at some point of whitelisting a key on alfajores and using that one on e2e tests, but that hasn't been implemented yet.

We can reuse part of this code when the above or some other alternative is live.

PR #5225, opened 2020-09-28
Celo Art Project for Kuneco

aaronmgdr:
### Description

Art inspired by Celo mission and for the Celo community. to be shown at Kuneco

staging.celo.org/flowers

### Tested

Screen Shot 2020-09-28 at 3 55 11 PM
Screen Shot 2020-09-28 at 3 55 19 PM
Screen Shot 2020-09-28 at 3 55 31 PM
Screen Shot 2020-09-28 at 3 55 49 PM

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/187


ErinHales:
This is beautiful, I love that this exists 💐🌼🌸 Also, love the animation, but wondering if we can shrink the initial size of the flower to about 3/4 to 1/2 the size it is currently? Also, would it be possible to center the flower until the animation was complete?  We could possibly have the animation occur on page load rather than during scroll.  Seems like the scroll leads to a jerky animation.

aaronmgdr:
The size is by design so i don't think we will change it. 

I thought i had gotten rid of the jerk however i see loading it ff now that it is still with us.

aaronmgdr:
closing in favor of https://github.com/celo-org/celo-monorepo/pull/5244

Issue #4812, opened 2020-08-26
Conduct contract release 1


yorhodes:
This issue is tracking Celo's first smart contract upgrade release. While executing upon this release according to the process laid out in https://docs.celo.org/community/release-process/smart-contracts, we are simultaneously going to improve it as we go and reference those changes here as well

Process Checklist:

- [x] [Merge all changes that should go into release 1](https://github.com/celo-org/celo-monorepo/issues/4812#issuecomment-691184631)
- [x] [Cut release branch and submit branch to auditors](https://github.com/celo-org/celo-monorepo/issues/4812#issuecomment-691186072)
- [x] Receive Audit report
- [x] Address issues found in audit
- [x] Submit audit fixes to auditors @emaG3m @yorhodes 
- [x] [Receive final audit report](https://github.com/celo-org/celo-monorepo/issues/4812#issuecomment-706883667)
- [x] Write up release/governance proposal notes @alecps 
- [x] [Make release on Github and add @alecps's release notes draft](https://github.com/celo-org/celo-monorepo/issues/4812#issuecomment-699127362) @nambrot 
- [x] [(Sep 29) Propose GasLimit increase on Baklava](https://github.com/celo-org/celo-monorepo/issues/4812#issuecomment-700888483) @alecps 
- [x] [(Oct 6) Propose GasLimit increase on Alfajores](https://github.com/celo-org/celo-monorepo/issues/4812#issuecomment-704501034) @aslawson @emaG3m 
- [x] [(Oct 13) Propose GasLimit increase on Mainnet](https://github.com/celo-org/celo-monorepo/issues/4812#issuecomment-708021473) @aslawson 
- [x] Deploy contracts on Baklava @yorhodes 
- [x] Verify Deploy on Baklava (i.e. test downtime slasher) @yorhodes 
- [x] Propose contract upgrade on Baklava 
- [x] Deploy contracts on Alfajores 
- [x] Propose contract upgrade on Alfajores
- [x] Verify Deploy on Alfajores (i.e. test downtime slasher)
- [x] Deploy contracts on Mainnet
- [x] Propose contract upgrade on Mainnet

Blockers:
- [x] [Testing that raising the block gas limit is safe](https://github.com/celo-org/celo-monorepo/issues/4812#issuecomment-696272192)

To-Dos:
- [x] Review [@m-chrzan's verify-bytecode](https://github.com/celo-org/celo-monorepo/pull/4996/files) PR @emaG3m @yorhodes 
[...]

PR #5235, opened 2020-09-29
Update CI with hardcoded release tag

ewilz:
### Description

As releases are happening, master will check against the latest release
candidate (pre audit-fixes), while changes to the latest release candidate (such changes are typically audit fixes) will have to still track changes against the currently deployed, in-use release.

### Tested

Make build pass with correct versions

### Related issues

- Checklist item here: https://github.com/celo-org/celo-monorepo/issues/4812

### Backwards compatibility

yah


packages/protocol/contracts/governance/SlasherUtil.sol
```diff
@@ -21,6 +21,14 @@ contract SlasherUtil is Ownable, Initializable, UsingRegistry, UsingPrecompiles
 
   event SlashingIncentivesSet(uint256 penalty, uint256 reward);
 
+  /**
```
yorhodes: I don't think we want this SlasherUtil to be versioned as multiple versioned contracts inherit

packages/protocol/contracts/governance/EpochRewards.sol
```diff
@@ -82,6 +82,14 @@ contract EpochRewards is
 
   event TargetVotingYieldUpdated(uint256 fraction);
 
+  /**
+  * @notice Returns the storage, major, minor, and patch version of the contract.
+  * @return The storage, major, minor, and patch version of the contract.
+  */
+  function getVersionNumber() external pure returns (uint256, uint256, uint256, uint256) {
```
yorhodes: should be implementing `ICeloVersionedContract`

packages/protocol/contracts/governance/DoubleSigningSlasher.sol
```diff
@@ -13,6 +13,14 @@ contract DoubleSigningSlasher is SlasherUtil {
   event SlashingIncentivesSet(uint256 penalty, uint256 reward);
   event DoubleSigningSlashPerformed(address indexed validator, uint256 indexed blockNumber);
 
+  /**
+  * @notice Returns the storage, major, minor, and patch version of the contract.
+  * @return The storage, major, minor, and patch version of the contract.
+  */
+  function getVersionNumber() external pure returns (uint256, uint256, uint256, uint256) {
```
yorhodes: ditto
should be `ICeloVersionedContract`

packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -16,6 +16,14 @@ contract UsingPrecompiles {
   address constant GET_PARENT_SEAL_BITMAP = address(0xff - 10);
   address constant GET_VERIFIED_SEAL_BITMAP = address(0xff - 11);
 
+  /**
+  * @notice Returns the storage, major, minor, and patch version of the contract.
+  * @return The storage, major, minor, and patch version of the contract.
+  */
+  function getVersionNumber() external pure returns (uint256, uint256, uint256, uint256) {
```
yorhodes: ditto
I don't think we want this to be versioned as multiple versioned contracts inherit

packages/protocol/contracts/common/GoldToken.sol
```diff
@@ -28,6 +28,14 @@ contract GoldToken is Initializable, CalledByVm, Freezable, IERC20, ICeloToken {
 
   event Approval(address indexed owner, address indexed spender, uint256 value);
 
+  /**
+   * @notice Returns the storage, major, minor, and patch version of the contract.
+   * @return The storage, major, minor, and patch version of the contract.
+   */
+  function getVersionNumber() external pure returns (uint256, uint256, uint256, uint256) {
```
yorhodes: should be `ICeloVersionedContract`

.circleci/config.yml
```diff
@@ -474,7 +474,7 @@ jobs:
             mkdir -p ~/.ssh
             ssh-keyscan github.com >> ~/.ssh/known_hosts
             git fetch --all
-            RELEASE_BRANCH=$(git branch -a --list *release/contracts/* | tail -1 | sed -e 's/remotes\/origin\///g')
+            RELEASE_BRANCH=release/contracts/1
```
yorhodes: won't this only pass after we merge the pre-release branch? or did we already?

yorhodes (changes requested):
this diff is understandably confusing
why do we need to make so many contract changes?

.circleci/config.yml
ewilz: thanks for the review. My PR was not up to date because I keep forgetting to push to my fork (instead of this repo), from which I'm now issuing PR's, womp. It's up to date now. It needs to point to the base release, from before any audit fixes. I'm just waiting to see how this affects the build. THen will need feedback on the best way to reference this base commit (I have it as a branch for now)

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -48,7 +48,7 @@ contract MetaTransactionWallet is
    * @return The storage, major, minor, and patch version of the contract.
    */
   function getVersionNumber() public pure returns (uint256, uint256, uint256, uint256) {
-    return (1, 1, 0, 0);
+    return (1, 1, 0, 1);
```
ewilz: Not sure why this is still not up to date. It would seem to me this latest commit should've required a updated patch version, but it passed the build without it: https://github.com/celo-org/celo-monorepo/commit/009292124bf941d73f85fea2a354394a322874e9

nambrot: So when that build ran, I did not yet have cut the branch which would have caused it to compare against the release branch

.circleci/config.yml
```diff
@@ -474,7 +474,7 @@ jobs:
             mkdir -p ~/.ssh
             ssh-keyscan github.com >> ~/.ssh/known_hosts
             git fetch --all
-            RELEASE_BRANCH=$(git branch -a --list *release/contracts/* | tail -1 | sed -e 's/remotes\/origin\///g')
+            RELEASE_BRANCH=release/contracts/base_release
```
nambrot: If we use `base_release` in this case, how would we handle it when cutting the next release branch? Wouldn't need we two different `base_releases`?

packages/protocol/contracts/common/MetaTransactionWallet.sol
emaG3m: either way, it clearly changed the bytecode, so I'd think it would want to be bumped??

nambrot: Yes bumped make sense!

.circleci/config.yml
ewilz: changing this to a tag.

nambrot (approved):

yorhodes (approved):
lgtm

nambrot:
Can be closed as https://github.com/celo-org/celo-monorepo/pull/5254 went in

PR #5238, opened 2020-09-29
Validator Explorer Improvements

### LINKED ISSUE #5102: Replace "Gold" with "Celo" on the validator explorer
On the validator UI (validator.celo.org) there are some places where it still says Gold when it should not. 

Also check all the tooltips

replace with the corret language 
Screen Shot 2020-09-16 at 4 04 26 PM

ErinHales:
### Description

 - "Celo Gold" to "CELO"
 - Refactors and cleans up validator code, adds `utils/validators`
 - Fixes sort functionality on validator table
 - Validators that have never been elected show an attestation score of 'n/a' instead of 0
 - Adds transitions to arrows in table

### Tested

Tests will be included in separate PR

### Related issues

- Fixes #5102 

### Backwards compatibility

Yes


packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -293,26 +148,14 @@ class ValidatorsList extends React.PureComponent<Props, State> {
...
-
-  isPinned({ address }: any) {
-    const list = (localStorage.getItem(localStoragePinnedKey) || '').split(',') || []
-    return +list.includes(address)
-  }
+    this.cachedCleanData = (data || []).sort((a, b) => dir * compare(accessor(a), accessor(b)))
 
-  onPinned() {
-    this.setState({ update: Math.random() } as any)
+    return (data || []).sort((a, b) => dir * compare(accessor(a), accessor(b)))
```
ErinHales: I really really wanted to simplify these two lines, but no matter what I try, it causes a weird bug to occur.  Really wanted to either return data or assign to `this.cachesCleanData` (instead of doing both) or do something like
```
const sorted = (data || []).sort((a, b) => dir * compare(accessor(a), accessor(b)))
this.cachesCleanData = sorted
return sorted
```
but all of those cause a bug with the sort 😒

packages/web/src/utils/validators.ts
```diff
@@ -0,0 +1,184 @@
...
+    elected: boolean
+    neverElected: boolean
+    online: boolean
+    uptime: number
+    attestation: number
+    claims: string[]
+  }>
+}
+
+export function isPinned({ address }: any) {
```
aaronmgdr: any chance this can be typed more explicitly?

packages/web/src/dev/ValidatorsListRow.tsx
```diff
@@ -85,61 +48,33 @@ class ValidatorsListRow extends React.PureComponent<Props & I18nProps, State> {
...
-            <View style={[styles.pin, isPinned ? styles.pinned : {}]} />
-          </View>
           <View style={[styles.tableCell, styles.tableCellTitle]}>
-            <Text style={[styles.defaultText, styles.tableCell, styles.tableCellTitleArrow]}>
+            <Text
+              style={[
+                styles.defaultText,
+                styles.tableCell,
+                styles.tableCellTitleArrow,
+                ...(expanded ? [styles.tableCellTitleArrowExpanded] : []),
```
aaronmgdr: if im reading this right i think `expanded && styles.tableCellTitleArrowExpanded` is all you need here. Saves on making arrays

packages/web/src/dev/ValidatorsList.tsx
aaronmgdr: yeah using a cache like this is bound to be buggy with react. 

my first thought is that we should not be overwriting the cachedData. it should be a constant. 

my second thought is that all this the cleanData function should be called on the data one layer up like right when its obtained rather than in this component. and that the concept of cached data shouldnt be needed in the sense of a giant cache like this.

aaronmgdr (commented):
thanks for taking this on.

packages/web/src/dev/ValidatorsListRow.tsx
ErinHales: Oh yes, much better!

packages/web/src/utils/validators.ts
ErinHales: Yes!  Added address type, let me know if that looks sufficient

packages/web/src/dev/ValidatorsListApp.tsx
```diff
@@ -172,7 +173,7 @@ class ValidatorsListApp extends React.PureComponent<Props> {
                     </View>
                   )
                 }
-                return <ValidatorsList data={data} isLoading={loading} />
+                return <ValidatorsList data={cleanData(data)} isLoading={loading} />
```
aaronmgdr, line 176: yes!

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -27,24 +26,21 @@ class HeaderCell extends React.PureComponent<HeaderCellProps, { hover: boolean }
...
 
   render() {
-    const { style, name, order, onClick, tooltip } = this.props
+    const { asc, style, name, ordered, orderFn, tooltip } = this.props
     const { hover } = this.state
     return (
-      <Hoverable onHoverIn={this.onHoverIn} onHoverOut={this.onHoverOut}>
-        <View onClick={onClick} style={[styles.tableHeaderCell, ...((style || []) as any)]}>
+      <Hoverable onHoverIn={this.onHoverIn} onHoverOut={this.onHoverOut} onPress={orderFn}>
+        <View style={[styles.tableHeaderCell, ...((style || []) as any)]}>
```
aaronmgdr:
```suggestion
<View style={[styles.tableHeaderCell, style]}>
```

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -321,77 +134,87 @@ class ValidatorsList extends React.PureComponent<Props, State> {
...
               name="Attestation"
-              order={orderBy === 'attestation' ? orderAsc : null}
+              ordered={orderKey === 'attestation'}
+              asc={orderAsc}
             />
           </View>
           {validatorGroups.map((group, i) => (
-            <div key={group.id} onClick={this.expand.bind(this, i)}>
-              <ValidatorsListRow onPinned={onPinned} group={group} expanded={expanded === i} />
+            <div key={i} onClick={this.expand.bind(this, i)}>
```
aaronmgdr: its generally better to use a key that will not change if things are reordered

packages/web/src/dev/ValidatorsListStyles.tsx
```diff
@@ -113,6 +113,11 @@ export const styles = StyleSheet.create({
   },
   tableHeaderCellArrowVisible: {
     opacity: 0.6,
+    transitionProperty: 'transform',
+    transitionDuration: '0.4s',
+  },
+  tableHeaderCellArrowDesc: {
+    transform: [{ rotateZ: '180deg' }, { translateX: -5 }, { translateY: -2 }],
```
aaronmgdr, line 120: nice!

packages/web/src/utils/validators.ts
```diff
@@ -0,0 +1,256 @@
...
+  const dir = asc ? 1 : -1
+
+  const compare = (a, b): number => {
+    if (a === null) {
+      return 1
+    }
+    if (b === null) {
+      return -1
+    }
+    return a > b ? 1 : -1
```
aaronmgdr, line 253: thoughts on returning 0 for cases where they are equal?

packages/web/src/utils/validators.ts
```diff
@@ -0,0 +1,256 @@
...
+  rewards: (_) => _.rewards || 0,
+  uptime: (_) => _.uptime || 0,
+  attestation: (_) => _.attestation || 0,
+}
+
+/**
+ * Determines whether a validator is pinned
+ * Pinned validators are set to the top of the table, regardless of how the table is sorted
+ * @param  {Address}    address The address of a given Validator
+ * @return {boolean}    Whether the given validator is pinned
```
aaronmgdr: I believe that this actualy returns a number

packages/web/src/utils/validators.ts
```diff
@@ -0,0 +1,256 @@
...
+ */
+export function isPinned(address: Address) {
+  const list = (localStorage.getItem(localStoragePinnedKey) || '').split(',') || []
+  return +list.includes(address)
```
aaronmgdr, line 115: `+` coercese to a number so return will be 1 or 0? i think (not boolean like return doc says

ErinHales: Yes, good catch

packages/web/src/dev/ValidatorsList.tsx
```diff
@@ -58,261 +54,78 @@ class HeaderCell extends React.PureComponent<HeaderCellProps, { hover: boolean }
...
-      if (a === null) {
-        return 1
-      }
-      if (b === null) {
-        return -1
-      }
-      return a > b ? 1 : -1
+  expand(expanded: number) {
+    if (this.state.expanded === expanded) {
+      return this.setState({ expanded: undefined })
```
aaronmgdr, line 114: oh! was about to write that it looked like both setStates were being called since i didnt see this return statement .

packages/web/src/utils/validators.ts
ErinHales: Hm, I don't think it's necessary.  Thought maybe just returning `a - b` would be a good way to simplify, but that doesn't work for some reason

aaronmgdr: for `a - b` to work that operation would need to return a number but `a` and `b` could be strings. (mostly likely are strings) 

also does sort work with any number returned? i thought it needed to be 1 or -1 or 0 so even if they are numbers im not sure it would work.

aaronmgdr (approved):

packages/web/src/utils/validators.ts
ErinHales: That would work for numbers.  It sorts it based on if it's negative, positive or 0.  But you're right, it doesn't work for strings.  I think this is fine as is.

packages/web/src/_page-tests/validators/explore.test.tsx
```diff
@@ -161,7 +162,7 @@ describe('ValidatorsList', () => {
     const tree = renderer
       .create(
         <TestProvider>
-          <ValidatorsList data={mock.data} />
+          <ValidatorsList data={cleanData(mock.data as ValidatorsListData)} />
```
aaronmgdr: It would be safer to set data to its correct type when it is defined.

packages/web/src/_page-tests/validators/explore.test.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react'
 import * as renderer from 'react-test-renderer'
 import { TestProvider } from 'src/_page-tests/test-utils'
 import ValidatorsList from 'src/dev/ValidatorsList'
+import { cleanData, ValidatorsListData } from 'src/utils/validators'
 
 const mock = {
```
aaronmgdr: for instnace `const data: ValidatorsListData = ...`

ErinHales: Fair enough, that was kind of lazy 😅

PR #5240, opened 2020-09-29
Added fees to the activity feed

### LINKED ISSUE #4843: Blockchain API changes
- [ ] Display correct fees in transactions drilldowns from the feed
- [ ] Query escrow contract for comments of escrowed transactions

carterqw2:
### Description

Adds fees to the the feed response.

### Other changes

Removes unused verification rewards functionality.

### Tested

Updated snapshots and added more transactions to test aggregation of fees.

### Related issues
Closes #4843.

### Backwards compatibility

Backwards compatible as the fees is an optional field.


packages/blockchain-api/src/helpers/EventBuilder.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  static formatFees(fees: Fee[], timestamp: number): FormattedFee[] {
+    return fees.map((fee) => ({
+      type: fee.type,
+      amount: {
+        currencyCode: fee.currencyCode,
+        timestamp,
+        value: fee.value
+          .multipliedBy(-1)
```
jeanregisser: I think these should be positive.

packages/blockchain-api/test/blockscout.test.ts
```diff
@@ -244,6 +370,47 @@ describe('Blockscout', () => {
           },
           "block": "117451",
           "comment": "",
+          "fees": Array [
+            Object {
+              "amount": Object {
+                "currencyCode": "cGLD",
+                "timestamp": 1566481000000,
+                "value": "-1e-8",
```
jeanregisser: Wondering if we should disable the exponential notation. What do you think?

jeanregisser (commented):
Really nice work sofar! 👍

packages/blockchain-api/test/blockscout.test.ts
carterqw2: Wasn't sure what is the expected format on the app side. Do I need to round them up to the second decimal?

packages/blockchain-api/package.json
```diff
@@ -7,6 +7,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "test": "jest --ci --silent --coverage --runInBand",
+    "test:fast": "jest --ci  --runInBand",
```
jeanregisser: nit: it's even faster without `--runInBad`

packages/blockchain-api/test/InputDecoder.test.ts
```diff
@@ -23,25 +24,40 @@ const savedInput =
...
   describe('format-comment-string', () => {
     it('should decode comment correctly', () => {
-      const decoded = formatCommentString(input)
+      const decoder = new InputDecoder(contractAddresses, Input.fromString(input))
+      const decoded = decoder.getTransactionComment()
       expect(decoded).toEqual(comment)
     })
-    it('should return empty on too short input', () => {
-      const decoded = formatCommentString('0x10')
+    xit('should return empty on too short input', () => {
```
jeanregisser: Why skipping?

packages/blockchain-api/test/InputDecoder.test.ts
```diff
@@ -23,25 +24,40 @@ const savedInput =
...
+      const decoder = new InputDecoder(contractAddresses, Input.fromString('0x10'))
+      const decoded = decoder.getTransactionComment()
       expect(decoded).toEqual('')
     })
-    it('should return empty on invalid function selector', () => {
-      const decoded = formatCommentString(
-        '0x095ea7b30000000000000000000000000000000000000000000000000000000000000' +
-          'abe0000000000000000000000000000000000000000000000000214e8348c4f0000'
+    xit('should return empty on invalid function selector', () => {
```
jeanregisser: Why skipping?

packages/blockchain-api/src/transaction/Transaction.ts
```diff
@@ -0,0 +1,138 @@
...
+      // If this is raised, something is wrong with our assertion
+      throw new Error(`Fee transfers don't add up for tx ${this.blockscoutTx.transactionHash}`)
+    }
+  }
+
+  private calculateSecurityFee(): void {
+    if (!this.isCeloTransaction()) {
+      // transfer to validator should be the last one
+      this.transactionNavigator.popLastTransfer()
+      this.transactionNavigator.popTransferTo(Contracts.Governance)
```
jeanregisser: Why not popping the last 2 transfers?
I'm afraid this could be wrong in the unlikely case the tx has a real transfer to the governance contract.

Also it would be nice to leave a note for "future us" about why we do this special case for non Celo transactions.

packages/blockchain-api/src/transaction/Transaction.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (existingFeeTypeIndex !== -1) {
+      const existingFee = this.fees[existingFeeTypeIndex]
+      existingFee.value = existingFee.value.plus(newFee.value)
+    } else {
+      this.fees.push(newFee)
+    }
+  }
+
+  isCeloTransaction(): boolean {
+    return this.blockscoutTx.feeToken !== undefined && this.blockscoutTx.feeToken === CGLD
```
jeanregisser: Not sure how blockscout treats this, but from the TX side, an empty feeToken means CELO (CGLD) is used to pay for fees, so would suggest to change this to the following to be on the safe side.
```suggestion
    return !this.blockscoutTx.feeToken || this.blockscoutTx.feeToken === CGLD
```

packages/blockchain-api/src/transaction/Transaction.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    this.transactionFees.push({
+      type: FeeType.SECURITY_FEE,
+      value: new BigNumber(this.blockscoutTx.gasPrice).multipliedBy(this.blockscoutTx.gasUsed),
+      currencyCode: this.blockscoutTx.feeToken,
+    })
+  }
+
+  private calculateGatewayFee(): void {
+    if (!this.isCeloTransaction()) {
```
jeanregisser, line 134: Can we add a note about why we do this special treatment?

packages/blockchain-api/src/transaction/Transaction.ts
```diff
@@ -0,0 +1,138 @@
...
+      this.transactionNavigator.popTransferTo(this.blockscoutTx.gatewayFeeRecipient)
```
jeanregisser: Unlikely edge case but what if the main TX is actually to this particular gatewayFeeRecipient?

Previous implementation used the fact that the fees were always last in the reported transfers.

packages/blockchain-api/src/transaction/TransactionNavigator.ts
```diff
@@ -0,0 +1,101 @@
+import { BlockscoutCeloTransfer } from '../blockscout'
+import { ContractAddresses, Contracts } from '../utils'
+import { TransferCollection } from './TransferCollection'
+
+const MINTED_TOKEN_ADDRESS = '0x0000000000000000000000000000000000000000'
+
+export class TransactionNavigator {
```
jeanregisser: Thinking this could be renamed, maybe `TransfersNavigator`?

jeanregisser (commented):
Awesome! See comments below.

packages/blockchain-api/src/transaction/Transaction.ts
carterqw2: I just wanted to be precise about what I'm removing from the list of transfers. If there was a reliable way to detect that the last transaction is a transfer to a validator, I would do that too.

packages/blockchain-api/package.json
carterqw2: Nice, thanks!

packages/blockchain-api/test/InputDecoder.test.ts
carterqw2: Leftover, good catch, thank you!

packages/blockchain-api/test/InputDecoder.test.ts
carterqw2: Reverted.

packages/blockchain-api/src/transaction/Transaction.ts
carterqw2: I changed the implementation of `popTransferTo` to search from the end and added a test.

packages/blockchain-api/src/transaction/Transaction.ts
carterqw2: Thanks, I accepted the changed and added some tests.

packages/blockchain-api/src/transaction/Transaction.ts
carterqw2: Done.

packages/blockchain-api/src/transaction/Transaction.ts
carterqw2: I added a test for this edge case, thanks for the idea!

packages/blockchain-api/src/transaction/TransactionNavigator.ts
carterqw2: Makes sense to me, done.

jeanregisser (approved):
💯 🚢 :shipit:

PR #5241, opened 2020-09-30
ODIS Common: Add missing dependencies

codyborn:
### Description

Some dependencies were missed when migrating the test common files under src.


alecps (approved):
Thanks!

PR #5254, opened 2020-10-01
Fix Contract versioning CI checks on master

nambrot:
### Description

- Creating the `release/contracts/1` branch wrecked havoc as it changed the reference of all CI checks. It did bring up the fact that on current master, the reference for checks should not be mainnet anymore with release 1 under way
- As per our release process, @emaG3m and I figured that the right move should be `celo-core-contracts-v1.pre-audit` as that should serve as the right reference for changes on master
- However, that proved to be problematic too, because master only later got the version changes that we then cherry-picked onto `oz-prerelease`.
- Thus the correct reference becomes 8a8b82252e2da9a4afc743b1e27260d521729f74 which is the commit at which the version changes were introduced on master.

### Other changes

- `Signatures.sol` version was bumped in minor, since we did not check that in in #5122


packages/protocol/package.json
```diff
@@ -100,6 +100,7 @@
     "nodemailer": "^4.6.8",
     "targz": "^1.0.1",
     "tmp": "^0.1.0",
+    "ts-node": "8.3.0",
```
ewilz, line 103: ¿por qué?

.circleci/config.yml
```diff
@@ -478,7 +478,7 @@ jobs:
             mkdir -p ~/.ssh
             ssh-keyscan github.com >> ~/.ssh/known_hosts
             git fetch --all
-            RELEASE_BRANCH=$(git branch -a --list *release/contracts/* | tail -1 | sed -e 's/remotes\/origin\///g')
+            RELEASE_BRANCH=8a8b82252e2da9a4afc743b1e27260d521729f74
```
ewilz, line 482: to me this seems a bit cryptic. With the deleted code you at least kind of get the gist of:  `*release/contracts/*`. (and this possibly isn't really a release anymore?). Might merit a code comment?

ewilz:
Is the significance of this newly referenced commit, the fact that `getVersionNumber()` was added to all the contracts that didn't implement it yet?

As far as documenting this 0_o...in the future will we still be using the pre-audit release cut? Or will that still present problems?  (I think maybe I'm understanding that the versions numbers could fork from the audit fix process?....seems edgy though)

packages/protocol/package.json
nambrot: Scripts in the protocol package require `ts-node`, but it seems like they have been working so far as most of us + CI has it installed globally? I figured I'd add it

.circleci/config.yml
nambrot: Yeah, I added an echo to hopefully highlight this special case

nambrot:
Exactly that. In the future, the pre-audit release cut should work. We were just in this weird limbo state since when we cut the release, we didn't yet have everything in place for the version checks

ewilz (approved):

PR #5255, opened 2020-10-01
Create cli command for claiming github username

isabellewei:
### Description
 
`account:claim-github` requires from the user a `metadata.json` file, their account address, and their github username, and generates a json with a signature that is to be uploaded onto the user's github.

### Tested

not tested, needs variable names from `github.ts` in contractKit

### Related issues

- Fixes #[issue number here]


isabellewei:
I meant to make this PR in a fork of `celo-monorepo`, not here

PR #5263, opened 2020-10-02
[Wallet] Include phone number in DappKit auth 

### LINKED ISSUE #5264: [Wallet] DappKit is not explicit about requesting phone number permission
### Expected Behavior

DappKit explicitly states that it needs phone number permission during auth deeplink

### Current Behavior

DappKit allows users to share their Valora addresses and phone numbers, but only tells users their addresses are being shared

annakaz:
### Description

DappKit account auth was returning a Valora user's phone number and address to the requesting Dapp, but only stated that the address would be shared. This adds in phone number so that users are aware that it is shared
Screen Shot 2020-10-02 at 1 31 47 PM

Before
Screen Shot 2020-10-02 at 1 09 09 PM

### Other changes

Reformat address to match address display throughout app

## Issues
- Fixes https://github.com/celo-org/celo-monorepo/issues/5264

### Tested

By looking at the screen on a Pixel 2 

### Backwards compatibility

Yes


medhakothari (approved):
Looks good to me! thanks for adding this 🎉

Lss-Ankit:
Hey @gnardini  can you please provide build details to verify this task. 
Thanks!

gnardini:
Hi @Lss-Ankit to open the screen on the screenshot you need to click on this link on a phone with Valora installed: celo://wallet/dappkit?type=account_address&dappName=toca&callback=toca://route&requestId=hello and the opened screen should have the phone number of the Valora account.

Lss-Ankit:
Hi @gnardini  We have opened given link on IOS and Android devices and found the following: 
Builds: Test flight build v1.3.0(27) and Android play store internal build v1.3.0(1004294316).
Devices: Google Pixel XL (8.1), iPhone SE 2 (14.1) 

**For Android:**
- User is not redirect to Valora app after pasting given link into chrome browser. 
[Attachment: ](https://drive.google.com/file/d/1D18Ek2GlJXuPKuN2VU5vYqp47v-wXLEs/view?usp=sharing)

**For IOS:**
- Same link also not worked in IOS chrome browser as well. 
- Redirected to Valora app and able to see below screen after pasting given link into safari Mweb browser. 
- Also,  verified that Phone number and Account No. (Address) are shown correct on opened screen. 

Attachment: 
IOS_safari

gnardini:
Hi @Lss-Ankit 

It looks like the issue is in chrome's browser which is not redirecting, this is something that is not used directly from the browser, so it's not unexpected.

For context, this is used from other apps that want to integrate with Valora.

Since this is only a UI change and you could verify it for iOS, I think it can be considered approved.

Lss-Ankit:
Okay thanks @gnardini for the information. 
As issue is in chrome browser, do you want to raise a ticket for the same for other better alternative?

PR #5113, opened 2020-09-17
Multi-cluster k8s forno

tkporter:
### Description

This PR builds upon #4821, that's part of the reason why the diff is so huge!

#### Overview

This PR allows devs to use celotool to easily deploy full nodes in different clusters for the same celo env by allowing each celo env to have multiple contexts. For example, `rc1` could be the celo env and it could have contexts `gcp-us-west` and `gcp-asia-northeast`. It also introduces a new Terraform template for deploying a forno load balancer that routes traffic to full nodes in specified contexts.

#### Contexts

Contexts names are declared in the `.env` file along with their relevant configurations.
I think examples are the best way of showing what is going on:
```
# names of all supported contexts. 2 GCP context, one azure context, and 1 aws context
CONTEXTS=gcp-us-west,gcp-asia-northeast,azure-eastus,aws-us-west

# forno for this env will point to both GCP contexts
FORNO_CONTEXT=gcp-us-west1-a,gcp-asia-northeast2-a
FORNO_DOMAINS=this-is-a-test-forno.celo-networks-dev.org.
# all clusters must be in this VPC
FORNO_VPC_NETWORK_NAME=default

# definition of gcp-us-west ------------
# name of the GCP project
GCP_US_WEST_GCP_PROJECT_NAME=celo-testnet
# name of the GCP zone for the context
GCP_US_WEST_GCP_ZONE=us-west4-a
# name of the k8s cluster for the context
GCP_US_WEST_KUBERNETES_CLUSTER_NAME=federated-dev-us-west4-a
# how many full nodes in this context
GCP_US_WEST_FULL_NODES_COUNT=1
# the disk size of the full nodes in this context
GCP_US_WEST_FULL_NODES_DISK_SIZE=10

# definition of gcp-asia-northeast ------------
# name of the GCP project
GCP_ASIA_NORTHEAST_GCP_PROJECT_NAME=celo-testnet
# name of the GCP zone for the context
GCP_ASIA_NORTHEAST_GCP_ZONE=asia-northeast2-a
# name of the k8s cluster for the context
GCP_ASIA_NORTHEAST_KUBERNETES_CLUSTER_NAME=federated-dev-asia-northeast2-a
# how many full nodes in this context
GCP_ASIA_NORTHEAST_FULL_NODES_COUNT=1
# the disk size of the full nodes in this context
GCP_ASIA_NORTHEAST_FULL_NODES_DISK_SIZE=10

# definition of azure-eastus ----------------
# Azure subscription ID
[...]

PR #5266, opened 2020-10-03
Select static node based on region resolved from timezone

nategraf:
### Description

Building on the recent work to deploy Forno full nodes to multiple regions (e.g. https://github.com/celo-org/celo-monorepo/pull/5113), this PR adds a support in contractkit for simple resolution of the users timezone to inform what region they are in. (Note that this method is by no means perfect, but it should work in most cases) On mainnet, and other networks if more regions are deployed in the future, a region-specific `static-node.json` file is fetched from GCP to have the app establish connection to relatively nearby set of peers.

### Other changes

* Add the latest block number to the `network_sync_waiting` Segment event

### Tested

Added a number of new unit tests. Manually ran the new code on a development build and confirmed it connected to the right nodes for my region. Also changed by timezone and observed that it connected to the correct nodes for the region of the new timezone.


jeanregisser:
Cool!
Maybe naive question, but couldn't we use anycast IPs for this? or use domains and geo-location based load balancing?
This way we would manage a single static-nodes.json like today.

nategraf:
@jeanregisser We should definitely use one of these solutions, but it's a bit tricky to make this work with the p2p protocol and we will need more time to get this all worked out. I'm proposing this as a temporary solution until then.

packages/contractkit/scripts/fetch_timezones.ts
```diff
@@ -0,0 +1,97 @@
+import 'isomorphic-fetch'
+import fs from 'fs'
+import path from 'path'
+import { Timezone } from '../src/utils/timezone'
+
+const TZ_URL = 'https://data.iana.org/time-zones/tzdb/zone.tab'
+const OUTPUT_PATH = path.normalize(path.join(__dirname, '..', 'src', 'utils', 'timezones.json'))
+
+// Parse a lattitude coordinate as encoded in the zone.tab file.
+function parseLattitude(lat: string): number {
```
tkporter: nit: `latitude` here and throughout

packages/contractkit/src/network-utils/static-node-utils.ts
```diff
@@ -1,22 +1,155 @@
...
+      lattitude: { min: 0, max: 90 },
+      longitude: { min: -105, max: -25 },
+    },
+    {
+      name: 'gcp-us-west1',
+      lattitude: { min: 0, max: 90 },
+      longitude: { min: -170, max: -105 },
+    },
+  ],
+  // Alias for mainnet
```
tkporter, line 69: Wdyt of putting the array into a variable and using it as the value for `mainnet` and `rc1`? Just thinking to minimize the chance of things breaking later

Also I'm curious what the process is like for determining these values. What should we do if we remove/add a cluster? Maybe something to add to the runbook

packages/contractkit/src/network-utils/static-node-utils.ts
```diff
@@ -1,22 +1,155 @@
...
+   * may be replaced in the future.
+   */
+  static getStaticNodeRegion(networkName: string): string {
+    // Get the lattitude and longitude of the timezone locations.
+    // Note: This is the location of the city that the user has the timzone set to.
+    const tz = timezone()
+    const coords = tz.coordinates
+    if (coords === undefined) {
+      debug('Could not resolve region from timezone %s', tz.name)
+      return '' // Use the default region of static nodes
```
tkporter, line 95: this is a good idea-- i just realized that rn the file just called `rc1` (or `mainnet`) contains the enodes for the VM-based nodes. I assume we'd like to no longer have to maintain those a few months out from now (some time to account for old valora versions). Any take on which enodes we should put in the default one? Could be a basket of enodes (this would be much more annoying from an implementation perspective though) or simply from a specific context/region

tkporter (approved):
Looks great! Also hope you don't mind but I added the 1 liner to use `mainnet` instead of `rc1` for static node uploading into this branch

packages/contractkit/scripts/fetch_timezones.ts
nategraf: Ya, spelling is definitely not my strong suit.

packages/contractkit/src/network-utils/static-node-utils.ts
nategraf: Consolidated this into a variable, and yes, we will need to adjust this if we add or remove regions. Can you point me to the run book where I can add this?

Ad far as how these values were determined. It's a bit rough, but it's designed to follow the results I shared in this Slack message: 
Screen Shot 2020-10-14 at 2 55 07 PM

packages/contractkit/src/network-utils/static-node-utils.ts
nategraf: Because we have them deployed, using a basket of nodes makes the most sense to me. My understanding of LES as it stands is that it does not prioritize sending requests to nodes by latency, so we may want to exclude `southamerica-east1` from that basket, because is not as well connected to the rest of the world and users in Africa or the Phillipines would experience poor performance if their client routes requests there. If we want to use regions which will be workable for all users, `us-west1` and `us-east1` are the best connected globally. (On Google's network, all roads lead to Atlanta, and more recently, to Iowa and Oregon. There is also only a low bandwidth link between India and the Middle East, so traffic from travelling from Asia to Europe is often forced to go through America) https://cloud.google.com/about/locations

packages/contractkit/src/network-utils/static-node-utils.ts
tkporter: https://docs.google.com/document/d/1ZYYAlWPgabwltjfYTV4a_KuK6dqIO2qP3HN9g-gk1Wk/edit#heading=h.4unz11mx0ftl
I've been meaning to port this to notion, but feel free to just add it in here

packages/contractkit/src/network-utils/static-node-utils.ts
tkporter: Hmm okay, I will keep this in mind. Maybe there can be an env variable that lists the contexts that the default static nodes will consist of, like
```
FORNO_STATIC_NODE_DEFAULT_CONTEXTS=gcp-us-west1,gcp-us-east1
```
and whenever there's an update to one of those contexts, the static node file is reuploaded

packages/protocol/truffle-config.js
```diff
@@ -169,6 +169,8 @@ const networks = {
   },
   baklava: {
     ...defaultConfig,
+    provider: ipcProvider('/Users/victor/nodes/baklava/geth.ipc'),
```
tkporter: Unintentional?

tkporter (approved):
Lgtm!

packages/contractkit/src/network-utils/static-node-utils.ts
nategraf: I added a note to the runbook 👍 . Let me know what you think.

PR #5270, opened 2020-10-06
Attestation Service 1.1.0: Verification security codes

### LINKED ISSUE #5356: Attestation 8-digit Code Support
The attestation service sends an 8 digit code similar to a centralized offering and have the app send that back to the attestation service over HTTPS. If the correct code is sent back, the server can respond with the signed message.

timmoreton:
### Description

Adds security code functionality to attestation service.  Flow would be as follows:

- client reveal phone number to the attestation service with the `/attestations` endpoint with a `securityCodePrefix` (prefix includes a single digit number for the order of attestations revealed to make it easier for users to discern different codes) and an optional `language` parameter.
- attestations service sends the message with the security code prefix and the random security code via SMS
- user inputs the code and the client peels off the prefix, then includes the code with a subsequent request to the attestation service at the endpoint `/get_attestations`
- attestation service response will include the delivery status and signature
- the client sends a on-chain transaction to complete the attestation with the signature

### Other changes

* Bumps version to 1.1.0
* Minor docs changes

### Tested
**/attestations request**
`curl -H "Content-Type: application/json" -d '{ "phoneNumber":"+1XXXXXX","securityCodePrefix":"1","signature":"","account":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","issuer":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","salt":"b"}' localhost:3000/attestations`

`{"success":true,"identifier":"0x406a7dd34fd358953a6f423d0d09f6a634c7b683f0e85ce742dedb71aa210703","account":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","issuer":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","attempt":0,"countryCode":"US","status":"Sent","provider":"twilio"}`

**/get_attestations request**
`curl -H "Content-Type: application/json" 'localhost:3000/get_attestations?account=0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd&issuer=0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd&phoneNumber=%2B1XXXXXX&salt=b&securityCode=7295427'
`

`{"success":true,"identifier":"0x406a7dd34fd358953a6f423d0d09f6a634c7b683f0e85ce742dedb71aa210703","account":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","issuer":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","attempt":0,"countryCode":"US","status":"Delivered","provider":"twilio","duration":1527,"attestationCode":"celo://wallet/v/IzwU6oxtIFB5IlYTRcluz1rp9ynloC5w1ypuUp34EMbim`

Confirmed that without `securityCodePrefix`, the original message is sent as an SMS. ✅

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/5356

### Backwards compatibility

Still supports returning the signed message if the request to `/attestations` does not include a security code prefix


jmrossy:
@timmoreton Very cool! This is something I've been hoping to see for a while cause it can potentially enable auto-reading of messages on iOS (as you probably know)! How does it work roughly? Is the code a password to some encrypted value in a centralized storage location?

aslawson:
> @timmoreton Very cool! This is something I've been hoping to see for a while cause it can potentially enable auto-reading of messages on iOS (as you probably know)! How does it work roughly? Is the code a password to some encrypted value in a centralized storage location?

@jmrossy I added some detail to the PR description.  The security code is currently just stored in plaintext in the db.  The subsequent request from the client checks that the code the client sends is correct, and then returns the signature.

For extra security, we plan to enforce TSL and remove the firebase function so that the code is not exposed in transit. The same checks apply in that signed message includes the account address which must correspond to msg.sender for complete transaction, so not much you could do if intercepting the code. But would love your thoughts if you perceive further vulnerabilities here. We could ensure the attestation service only accepts requests from the requestor by including an auth header/code from the requesting account (similar to ODIS).

~The one component that IS missing from this PR is a limit on the attempts to try codes (since the entropy of 7 digits is not impossibly high to guess the code).~  <-- EDITED nvm looks like Tim already added this

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -141,22 +150,51 @@ class AttestationRequestHandler {
...
-        this.attestationRequest.smsRetrieverAppSig
-      )
+      // Determine if we're sending a security code, or the full deep link.
+      let messageBase, securityCode
+      if (this.attestationRequest.securityCodePrefix) {
+        if (this.attestationRequest.securityCodePrefix.length !== 1) {
+          throw new ErrorWithResponse('Invalid securityCodePrefix', 422)
+        }
+
+        // Client is requesting a security code SMS. Generate a challenge and just store the deeplink.
```
aslawson, line 173: hmm any reason we can't just sign the attestation when they are retrieving it during the `/get_attestation` request?  looks like all it needs is phoneNumber, salt, and account which we will have in that request as well.

just think we should persist as few codes as possible.

packages/attestation-service/src/requestHandlers/get_attestation.ts
```diff
@@ -36,10 +36,47 @@ class GetAttestationRequestHandler {
     this.key = getAttestationKey(getRequest)
   }
 
-  async getAttestationRecord() {
-    return findAttestationByKey(this.key, {
+  async withAttestationAndSecurityCodeChecked(f: (attestation: AttestationModel) => any) {
+    const transaction = await sequelize!.transaction({
       logging: this.sequelizeLogger,
+      type: Transaction.TYPES.IMMEDIATE,
```
aslawson, line 47: is IMMEDIATE not the default for db transaction options?  what does having it here ensure?

packages/attestation-service/migrations/20201006231054-attestation-v6.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
```
aslawson, line 1: can consolidate these migration files

packages/attestation-service/src/requestHandlers/attestation.ts
aslawson: nvm -- I guess we already persist the message today which is the signature. perhaps its not worth the added complexity for the `/get_attestation` method

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -37,6 +32,20 @@ function getAttestationKey(attestationRequest: AttestationRequest): AttestationK
   }
 }
 
+// TODO proper localization
+function getSecurityCodeText(language: string | undefined) {
+  switch (language) {
+    case 'es':
+    case 'es-419':
+    case 'es-LA': {
+      return 'Código de seguridad de Celo'
```
aslawson: i'd recommend aligning with the strings for mobile i18n
https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/i18n.ts#L86

so `es-US` and `es-LA`

packages/attestation-service/migrations/20201006231054-attestation-v6.js
aslawson: i'm guessing you may have already deployed this to a network, so I won't change this

packages/attestation-service/src/requestHandlers/get_attestation.ts
timmoreton: It fixes issues with sqlite described here: https://activesphere.com/blog/2018/12/24/understanding-sqlite-busy

aslawson: oh interesting, "IMMEDIATE behaviour uses a different kind of lock which allows concurrent readers, but blocks other concurrent writers". you'd think that would be accomplished just by obtaining the lock..

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -41,6 +41,10 @@ function getSecurityCodeText(language: string | undefined) {
     case 'es-LA': {
       return 'Código de seguridad de Celo'
     }
+    case 'pt':
```
aslawson, line 44: I chatted with the mobile circle and there isn't really a standardization for how local is represented across various services.  The language for the phone’s OS is in the format of 'es-US' 'es-LA' .  But for Valora's language settings, it’s actually es-419.  The Valora app could probably pass either to this endpoint, and I'd think that the Valora language settings makes the most sense.

I don't see anything wrong with supporting all cases for now and cleaning up later (in case we find one format to be preferable).

Additionally, we could do something like:

```
delimiterIndex = language.indexOf("-")
language.substring(0, delimiter >=0 ? delimiter : language.length - 1 )
```

so we reduce our cases to just the language 'es', 'pt', ect.

aslawson (approved):
LGTM

packages/attestation-service/src/requestHandlers/attestation.ts
timmoreton: I was thinking we may end up with different messages for e.g. LatAm vs Castilian spanish so we might prefer to just keep the flexible structure right now and make sure that whatever Valora sends matches one of these values.

aslawson: kk seems fine to keep as it for now 👍

packages/attestation-service/src/db.ts
```diff
@@ -48,7 +48,17 @@ export function isDBOnline() {
   }
 }
 
-export let kit: ContractKit
+let kit: ContractKit
+
+// Wrapper that on error tries once to reinitialize connection to node.
+export async function useKit<T>(f: (kit: ContractKit) => T): Promise<T> {
+  try {
+    return f(kit)
```
tkporter: I think if you don't await here, the rejected promise from `f` won't be caught. At least not if the rejection is from an asynchronous call within `f`

packages/attestation-service/src/db.ts
```diff
@@ -48,15 +48,32 @@ export function isDBOnline() {
   }
 }
 
-export let kit: ContractKit
+let kit: ContractKit | undefined
+
+// Wrapper that on error tries once to reinitialize connection to node.
+export async function useKit<T>(f: (kit: ContractKit) => T): Promise<T> {
```
nambrot, line 54: This is probably more meta, but this feels like something that should be addressed at the ContractKit level? I.e. a config on kit itself? cc @celo-org/devx

packages/attestation-service/src/request.ts
```diff
@@ -82,7 +82,8 @@ export function respondWithAttestation(
   res: express.Response,
   attestation: AttestationModel,
   alwaysSuccess?: boolean | undefined,
-  salt?: string | undefined
+  salt?: string | undefined,
+  showAttestationCode?: boolean | undefined
```
nambrot: This feels a bit dangerous for me to include the attestation code with a simple boolean variable. I 'm worried that someone down the line will accidentally enable it in the public endpoint. It might be nice to merge the payload by passing it in explicitly?

packages/attestation-service/src/requestHandlers/attestation.ts
```diff
@@ -141,22 +155,51 @@ class AttestationRequestHandler {
...
-      )
+      // Determine if we're sending a security code, or the full deep link.
+      let messageBase, securityCode
+      if (this.attestationRequest.securityCodePrefix) {
+        if (this.attestationRequest.securityCodePrefix.length !== 1) {
+          throw new ErrorWithResponse('Invalid securityCodePrefix', 422)
+        }
+
+        // Client is requesting a security code SMS. Generate a challenge and just store the deeplink.
+        securityCode = randomBytes(7)
```
nambrot, line 174: Optional Nit: There is a decent amount of string/code processing happening that I wonder would be beneficial to happen in helper methods?

packages/attestation-service/src/requestHandlers/get_attestation.ts
```diff
@@ -36,10 +36,47 @@ class GetAttestationRequestHandler {
     this.key = getAttestationKey(getRequest)
   }
 
-  async getAttestationRecord() {
-    return findAttestationByKey(this.key, {
+  async withAttestationAndSecurityCodeChecked(f: (attestation: AttestationModel) => any) {
```
nambrot: Nit: I personally like shorter variable names, but I believe the js convention is to call these `callback`, especially in more critical functions like this

packages/attestation-service/src/requestHandlers/get_attestation.ts
```diff
@@ -2,13 +2,13 @@ import { PhoneNumberUtils } from '@celo/utils'
...
-import { respondWithAttestation, respondWithError } from '../request'
+import { Transaction } from 'sequelize'
+import { findAttestationByKey, makeSequelizeLogger, sequelize, SequelizeLogger } from '../db'
+import { AttestationKey, AttestationModel } from '../models/attestation'
+import { ErrorWithResponse, respondWithAttestation, respondWithError } from '../request'
 import { obfuscateNumber } from '../sms/base'
 
-export const VERSION = process.env.npm_package_version as string
-export const SIGNATURE_PREFIX = 'attestation-service-status-signature:'
+const maxSecurityCodeAttempts = 5
```
nambrot: Nit: Convention calls for the constant names to be capitalized. Also there seems to be enough constants now that it would maybe make sense to have a centralized `constants.ts` file?

packages/attestation-service/src/requestHandlers/get_attestation.ts
```diff
@@ -36,10 +36,47 @@ class GetAttestationRequestHandler {
...
+        transaction,
+        lock: Transaction.LOCK.UPDATE,
+      })
+
+      if (!attestation) {
+        throw new ErrorWithResponse('Attestation not found', 404)
+      }
+
+      // Check security code is supplied and attempts allow
+      if (!this.getRequest.securityCode) {
```
nambrot, line 64: I'm probably missing something, but does this mean that if the get request does not contain a security code, then the getRequest is approved? Doesn't this prove problematic?

packages/attestation-service/src/requestHandlers/get_attestation.ts
nambrot: Optional Nit: This seems like a confusing interface where success means a void return, but an error throws with an error payload response?

packages/attestation-service/src/requestHandlers/get_attestation.ts
timmoreton: "Doesn't this prove problematic" is an understatement!  pls check the updated logic.

packages/attestation-service/src/db.ts
timmoreton: I agree, would prefer It on the CK level, but I feel out of scope rn

packages/attestation-service/src/requestHandlers/get_attestation.ts
aslawson: yikes good catch!

aslawson:
tested locally and it does not return the attestationCode if the security code is not provided

```
curl -H "Content-Type: application/json" -d '{ "phoneNumber":"+1XXX","securityCodePrefix":"1","signature":"","account":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","issuer":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","salt":"b"}' localhost:3000/attestations
```

```
{"success":true,"identifier":"0x406a7dd34fd358953a6f423d0d09f6a634c7b683f0e85ce742dedb71aa210703","account":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","issuer":"0x8e41FA2c0117eDCB977364E92cF38Dcc37074dFd","attempt":0,"countryCode":"US","status":"Sent","provider":"twilio"}
```

PR #5282, opened 2020-10-07
cli->0.0.59-dev & Utils->0.1.22-dev & ContractKit->0.4.16-dev

medhakothari:
### Description

cli->0.0.59-dev
Utils->0.1.22-dev
ContractKit->0.4.16-dev

### DIFF
--- a/packages/phone-number-privacy/combiner/package.json
+++ b/packages/phone-number-privacy/combiner/package.json
@@ -24,7 +24,7 @@
     "db:migrate": "NODE_ENV=dev ts-node scripts/run-migrations.ts"
   },
   "dependencies": {
-    "@celo/contractkit": "0.4.14",
+    "@celo/contractkit": "0.4.12",
     "@celo/phone-number-privacy-common": "1.0.7",
     "@celo/utils": "0.1.20",
     "blind-threshold-bls": "https://github.com/celo-org/blind-threshold-bls-wasm#e1e2f8a",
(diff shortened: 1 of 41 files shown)


packages/phone-number-privacy/combiner/package.json
```diff
@@ -24,9 +24,9 @@
     "db:migrate": "NODE_ENV=dev ts-node scripts/run-migrations.ts"
   },
   "dependencies": {
-    "@celo/contractkit": "0.4.12",
```
gastonponti: The cases where you have fixed versions, you should probably leave them as they are. Remember that we are actually not managing a "real" semantic versioning.
A version change could require some changes/fixes, until we "normalize" the semantic versioning again.
In the case of the `-dev`, as every package is pointing to master, changing the version is sort of an alias change, but all the compatibilities changes should be addressed before. But this is not the case with fixed versions.

So, I recommend to get back to the old versions. And in the future, if those packages owners want to use newer versions, will probably be a card in the milestone

packages/phone-number-privacy/common/package.json
```diff
@@ -16,7 +16,7 @@
     "lib/**/*"
   ],
   "dependencies": {
-    "@celo/contractkit": "0.4.15-dev",
+    "@celo/contractkit": "0.4.16-dev",
     "@celo/utils": "0.1.21-dev",
```
gastonponti: I think that this is the one making the `dependencies install` job fail

aslawson: looks like it also needs to be bump to `0.1.22-dev`

aslawson (commented):
+1 to @gastonponti mention about fixed versions

then i'll approve, thanks for updating!

packages/phone-number-privacy/common/package.json
medhakothari: oops my bad!

aslawson (approved):

PR #5296, opened 2020-10-09
Celo.org Cleanup

### LINKED ISSUE #5026: Remove cross-fetch
next.js provides its own cross functional fetch polyfill 

https://nextjs.org/blog/next-9-4#improved-built-in-fetch-support

### LINKED ISSUE #5297: replace react-reveal with awesome-react-reveal
react-reveal is not maintained

aaronmgdr:
### Description

Upgrades packages and cleans up ~most~ *all* error/warnings when running `yarn test`

#### package changes
- remove `cross-fetch` as next.js now provides a polyfill by default
- upgrade `fecha` (data library) to 4
- upgrades `yiq` (color library) to 4
- upgrade `react-autosuggest` to 10
- replace `react-reveal` with `react-awesome-reveal` as the former is not maintained and used deprecated react lifecycle methods. while there apis are not one to one, i was able to create custom animations that kept visual behavior consistent. 

#### test cleanup 
- wraps all renderings of components in tests with `TestProvider` which provides i18n context. this removes warning in tests and also means that  rather than tests rendering i18n keys we get the values of translation strings.
- use the react `htmlFor` instead of `for` 
- use `waitFor` instead of other deprecated `wait` variants from testing-library

### Other changes

fixes some places where `key` prop was not unique

Screen Shot 2020-10-14 at 2 45 13 PM

### Tested
yarn --audit showed no changes from master
as well as existing test manual testing shows correct behavior
### Related issues

- Fixes #5026 
- fixes #5297


packages/web/src/_page-tests/experience/__snapshots__/color.test.tsx.snap
```diff
@@ -6030,6 +6038,7 @@ exports[`Experience/Color renders 1`] = `
               }
             }
           >
+            Read more in Celo
```
ErinHales, line 6041: Should be `Read more in Celo's Terms of Service` or just `Read more in Terms of Service`?

packages/web/src/_page-tests/experience/__snapshots__/icons.test.tsx.snap
```diff
@@ -1401,7 +1401,9 @@ exports[`Experience/Icons renders 1`] = `
                   }
                 }
               >
-                icons.headline
+                Used to simplify the complexities of Celo technology, icons are useful visual tools that can educate, clarify, and instruct in ways simple text can not.
```
ErinHales: One word "cannot"?

packages/web/src/_page-tests/experience/__snapshots__/typography.test.tsx.snap
```diff
@@ -3093,7 +3102,7 @@ exports[`Experience/Typography renders 1`] = `
                     }
                   }
                 >
-                  typography.h4Example
+                  Subtitle / larger text that is one size larger than the body, lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor.
```
ErinHales: Should there be a break between the text and the lorem ipsum text?

packages/web/src/_page-tests/experience/__snapshots__/typography.test.tsx.snap
```diff
@@ -3809,7 +3819,7 @@ exports[`Experience/Typography renders 1`] = `
                 }
               }
             >
-              footer.technology
+              Technology
```
ErinHales, line 3823: How come we switched from using variables to using plain text?  Seems more likely we would have a typo somewhere or it would lead to more work if we decided to update one of these links

packages/web/jestSetup.js
```diff
@@ -17,3 +6,23 @@ Math.random = function random() {
...
+  }
+
+  takeRecords() {
+    return null
+  }
+
+  unobserve() {
+    return null
+  }
+}
```
ErinHales, line 28: What is the purpose of this code?

packages/web/src/_page-tests/experience/__snapshots__/typography.test.tsx.snap
aaronmgdr: I wrapped the tests in TestProdiver to get rid of a warning about there not being an i18n context available when tests were run. 

Most test already had this these ones just were created without it.

packages/web/jestSetup.js
aaronmgdr: It's a mock since interaction observer is not available in test env

packages/web/src/_page-tests/experience/__snapshots__/typography.test.tsx.snap
aaronmgdr: Beyond that benefit of using the real strings is it's closer to user behavior. Also if we had a typo in the key it will be more obvious since it will show up in the snapshots.

packages/web/src/dev/LeaderBoard.test.tsx
```diff
@@ -1,19 +1,36 @@
 import { fireEvent, render } from '@testing-library/react'
 import * as React from 'react'
+import { TestProvider } from 'src/_page-tests/test-utils'
 import LeaderBoard from 'src/dev/LeaderBoard'
 
-const leaders = new Array(12).fill({ points: 10, identity: 'annie', address: '0x1f3f' })
+const leaders = new Array(12)
+  .fill({ points: 10, identity: 'annie', address: '0x1f3f' })
```
aaronmgdr, line 7: Added the index so that keys would be unique

packages/web/src/shared/AwesomeFade.tsx
```diff
@@ -0,0 +1,72 @@
+import { keyframes } from '@emotion/core'
+import * as React from 'react'
+import Reveal from 'react-awesome-reveal'
```
aaronmgdr, line 3: This file is worth looking over

aaronmgdr: @ErinHales

packages/web/src/_page-tests/experience/__snapshots__/typography.test.tsx.snap
aaronmgdr: sounds like a good improvement

packages/web/src/shared/AwesomeFade.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+type Direction = 'up' | 'right'
```
ErinHales: 'X' or 'Y' might make more sense here since we're using translations

packages/web/src/shared/AwesomeFade.tsx
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export default React.memo(function AwesomeFade({
+  children,
+  reverse,
+  delay,
+  distance,
+  direction,
+  duration,
+  fraction,
```
ErinHales, line 27: What does "Fraction" do?

packages/web/src/shared/AwesomeFade.tsx
```diff
@@ -0,0 +1,72 @@
...
+    </Reveal>
+  )
+})
+
+function getKeyFrames(distance: number | string, direction: Direction, reverse: boolean) {
+  let from = `translate3d(0, ${distance}, 0)`
+
+  if (direction === 'right') {
+    from = `translate3d(${distance}, 0, 0)`
+  }
```
ErinHales, line 50: Then we could do: 
```
const from = `translate${direction}(${distance})`
```

packages/web/src/shared/AwesomeFade.tsx
```diff
@@ -0,0 +1,72 @@
...
+    from {
+      opacity: 0;
+      transform: ${from};
+    }
+    to {
+      opacity: 1;
+      transform: translate3d(0, 0, 0);
+    }
+  `
+}
```
ErinHales, line 62: Would like to shorten this to a single return statement using variables, but I don't have any good ideas...
``` 
if (reverse) {
    return keyframes`
    from {
      opacity: 1;
      transform: translate3d(0, 0, 0);
    }
    to {
      opacity: 0;
      transform: ${from}; 
    }
  `
  }

  return keyframes`
    from {
      opacity: 0;
      transform: ${from};
    }
    to {
      opacity: 1;
      transform: translate3d(0, 0, 0);
    }
  `
```

ErinHales:
```
const to = 'translate3d(0, 0, 0)`

return keyframes`
  from {
    opacity: ${+reverse};
    transform: ${reverse ? to : from};
  } to {
    opacity: ${+(!reverse)};
    transform: ${reverse ? from : to};
  }
`
```
Eh, that's kind of what I was thinking, but makes it harder to read.  It's probably fine as is :)

packages/web/src/shared/AwesomeFade.tsx
aaronmgdr: it's passed to the intersection observer. Its the fraction of an element that needs to be visible before it is considered onscreen

packages/web/src/shared/AwesomeFade.tsx
aaronmgdr: oh interesting using the coersion power of `+` to turn boolean in to 1 or 0. clever

packages/web/src/_page-tests/experience/__snapshots__/color.test.tsx.snap
aaronmgdr: the "terms of service" is linked so its in the node below

packages/web/src/shared/AwesomeFade.tsx
aaronmgdr: i like it

packages/web/src/shared/AwesomeFade.tsx
aaronmgdr: yeah originally i was match the API of the react-awesome-reveal but this makes sense

isabellewei (approved):
LGTM! There are indeed no warnings when running `yarn test` :)

PR #5301, opened 2020-10-09
Fix network:contracts CLI command

jmrossy:
### Description

Fix error when showing contract addresses and remove hard-coded list of contracts.

### Other changes

Reorder CLI commands in package.json alphabetically

### Tested

Locally on alfajores and mainnet

### Backwards compatibility

Yes


yorhodes (approved):
LGTM, thanks for the fix!

PR #5304, opened 2020-10-10
Build wallet command that builds with relevant dependencies only

### LINKED ISSUE #5303: Add wallet-specific build command to celo-monorepo
### Expected Behavior
Running the `build:wallet` command will allow wallet devs to easily build with the least amount of wait time.

### Current Behavior
The `build` command takes too long to run. However, running the command while only including filtered dependencies is faster.

tarikbellamine:
### Description
Adding a `build:wallet` command to the root repo so wallet devs can lower wait time by building only what they need.

### Other changes
N/A

### Tested
Yes

### Related issues
- Fixes #5303

### Backwards compatibility
Yes


jeanregisser (approved):
🎉

PR #5307, opened 2020-10-12
Support Smart Contract Array Arguments

daoauth:
Now ```transfer(address,uint256)``` is no problem. But ```transfers(address[2],uint256[2])``` case is not supported.
And  RegExt has similar problem. ```ABI_TYPE_REGEX``` supports only dynamic array.

So encoding and decoding functionArgs with ```JSON``` and ```BASE64```, and edit regEx. It can be supported multi dimension array eg. ```uint256[2][2]```)

And if this module isn't used a lot yet, we might consider giving up backwards compatibility.

https://github.com/celo-org/celo-monorepo/issues/5308


medhakothari:
Closing this PR because it's stale -- please feel free to update to current code and reopen the PR!

PR #5315, opened 2020-10-13
[Wallet] Report to analytics reveal statuses from validators

i1skn:
### Description

Implementing a second part of https://github.com/celo-org/celo-labs/issues/574. The logic is the following:
- we need to fetch reveal status for each attestation. Reasons fo initiate fetch any(code entered,  go away from the enter codes screen)
- call once again get-attestations for the issuer if not 200 returned on reveal attempt immediately.
- add firebase proxy function `proxyRevealStatus`.

### TODO
- [x] When merged, `yarn deploy` needs to be run in `packages/attestation-proxy` to deploy firebase function.

### Other changes

* Small refactoring
* Remove `features. USE_PHONE_NUMBER_PRIVACY`

### Tested

- [x] Complete verification flow (all use-cases for triggering the report worked out well)
- [x] Deployed `proxyRevealStatus` and `proxyReveal` to `celo-mobile-alfajores` and tested forcing using proxy in the codebase on alfajores.

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/574

### Backwards compatibility

Yes


timmoreton (approved):
Nice! I would say we want `status`, `attempt`, `provider`, `duration` (if present), `errors` (if present).

packages/mobile/ios/Podfile.lock
```diff
@@ -849,4 +849,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 470c646f25b89e063843bdec1cd889a649b38781
 
-COCOAPODS: 1.9.1
+COCOAPODS: 1.9.3
```
aslawson: what's this?

packages/mobile/src/analytics/ValoraAnalytics.ts
```diff
@@ -26,17 +26,17 @@ async function getDeviceInfo() {
     LastUpdateTime: await DeviceInfo.getLastUpdateTime(),
     Manufacturer: await DeviceInfo.getManufacturer(),
     MaxMemory: await DeviceInfo.getMaxMemory(),
-    Model: await DeviceInfo.getModel(),
-    ReadableVersion: await DeviceInfo.getReadableVersion(),
-    SystemName: await DeviceInfo.getSystemName(),
-    SystemVersion: await DeviceInfo.getSystemVersion(),
+    Model: DeviceInfo.getModel(),
```
aslawson, line 29: why do some still need the `await` while these do not?

packages/mobile/src/identity/verification.ts
```diff
@@ -881,3 +966,25 @@ function* waitForAttestationCode(issuer: string) {
     }
   }
 }
+
+function* getPhoneHashData(e164Number: string) {
```
aslawson: nice refactor.

out of scope, but I wonder do we still even need the USE_PHONE_NUMBER_PRIVACY flag?

aslawson (commented):
looks good.  i would say we'd want all the data you specified in the PR overview.

i'm going to sync with tim to understand the schema for multiple attempts and whether it makes sense to parse at the level of throwing the event here or to keep in the "status" field.

i'll get back to you on it!

packages/mobile/src/analytics/Properties.tsx
```diff
@@ -273,6 +273,10 @@ interface VerificationEventsProperties {
     issuer: any
     error: string
   }
+  [VerificationEvents.verification_reveal_attestation_status]: {
+    issuer: any
+    status: any
```
aslawson: ok, chatted with Tim.  Can we make this event reflect the fields that the attestation service returns?  iow, just parse the first level of the JSON response field-for-field, and then the parsing/transformation of the individual errors should be doable in our metrics pipeline.

```
  issuer: any,
  identifier: any,
  success: boolean,
  account:  any,
  attempt: number,
  countryCode: string,
  status: string,
  provider: string,
  duration: number,
  errors: any
```

^ though, I think some of these already are automatically added as generic Valora events like 'identifier' and 'account' but not sure where that happens

packages/mobile/src/analytics/ValoraAnalytics.ts
i1skn: The ones, which return a `Promise` need `await` - others do not.

packages/mobile/ios/Podfile.lock
i1skn: Unnecessary update of CocoaPods, will remove it.

packages/mobile/src/analytics/Properties.tsx
aslawson: I see you just pass forward the entire json response from the attestation service to this event.  Does the event thrown include all fields even though only these two are defined in the Properties.tsx?

I'm not super familiar with this part of the metrics pipeline and how segment may use this file, but I wonder if we need to explicitly include each property here for segment to include the additional fields... I would think @annakaz @tarikbellamine know

packages/mobile/src/identity/verification.ts
```diff
@@ -866,6 +873,79 @@ async function postToAttestationService(
...
+    )
+    const body = yield call(response.json.bind(response))
+    if (response.ok) {
+      Logger.debug(
+        `${TAG}@reportAttestationRevealStatus`,
+        `Successful for service url ${attestationServiceUrl}`
+      )
+      ValoraAnalytics.track(VerificationEvents.verification_reveal_attestation_status, body)
+    } else {
+      Logger.debug(
```
aslawson: any way to parse the response to ignore if the `get_attestation` endpoint doesn't exist for the service issuer -- to make this backwards compatible for older versions of attestation services that don't have this endpoint yet rather than logging it as a failure?

packages/attestation-proxy/src/index.ts
```diff
@@ -33,3 +31,39 @@ export const proxyReveal = functions.https.onCall(async (data, context) => {
...
+
+export const proxyRevealStatus = functions.https.onCall(async (data, context) => {
+  // refuse requests that do not have basic auth
+  checkAuth(context)
+
+  const { attestationServiceUrl, phoneNumber, account, issuer, pepper } = data
+  try {
+    const urlParams = new URLSearchParams({
+      phoneNumber,
+      salt: pepper ?? '',
```
jmrossy, line 43: Out of scope here but just a nit, would be nice to have salt renamed in attestation service too: https://github.com/celo-org/celo-monorepo/blob/master/packages/utils/src/io.ts#L109

Since there's been changes there anyway seems like there's an opportunity to do that (in a backwards compat way) before the next release. cc @aslawson @timmoreton

packages/mobile/src/identity/actions.ts
```diff
@@ -226,6 +236,7 @@ export type ActionTypes =
   | UpdateVerificationState
   | ResendAttestations
   | SetLastRevealAttempt
+  | ReportRevealStatus
```
jmrossy: Nit these interfaces typically end in `Action`, though it seems a few others in this file already break that convention

packages/mobile/src/identity/verification.ts
```diff
@@ -881,3 +994,15 @@ function* waitForAttestationCode(issuer: string) {
     }
   }
 }
+
+function* getPhoneHashData(e164Number: string) {
+  let phoneHash: string
+  let phoneHashDetails: PhoneNumberHashDetails
+  phoneHashDetails = yield call(fetchPhoneHashPrivate, e164Number)
+  phoneHash = phoneHashDetails.phoneHash
+  return { phoneHash, phoneHashDetails }
```
jmrossy: Nit, why not just return phoneHashDetails which has the hash?

jmrossy (commented):
Looks good to me! Just curious, are there plans to deprecate the proxy entirely? It was only meant to be used while validators got their attestation services set up properly with TLS.

aslawson:
> Looks good to me! Just curious, are there plans to deprecate the proxy entirely? It was only meant to be used while validators got their attestation services set up properly with TLS.

@jmrossy yup, we are currently in the process of urging validators to enable TLS now

packages/mobile/src/identity/verification.ts
```diff
@@ -881,8 +881,32 @@ export function* reportRevealStatusSaga({
...
+      const proxyReveal = functions().httpsCallable('proxyRevealStatus')
+      const response = yield call(proxyReveal, body)
+      const { status, data } = response.data
+      const ok = status >= 200 && status < 300
+      aggregatedResponse = { ok, status, body: JSON.parse(data) }
+    } catch (error) {
+      Logger.error(`${TAG}@reportAttestationRevealStatus`, 'Error calling proxyRevealStatus', error)
+      // The httpsCallable throws on any HTTP error code instead of
+      // setting response.ok like fetch does, so catching errors here
+      aggregatedResponse = { ok: false, status: 500, body: error }
```
aslawson, line 913: interesting -- does the error code relate to the appropriate status code?  rather than hardcoding it to a 500 status response here?

aslawson (approved):
lgtm!

packages/attestation-proxy/src/index.ts
aslawson: we've already published the attestation service, so unfortunately can't do it this iteration :/

i1skn:
@jmrossy I've addressed the feedback, PTAL.

Lss-Ankit:
@jeanregisser  can you please provide more details if we need to verify this task. 
Thanks!!

jeanregisser:
No need to verify @Lss-Ankit  thanks!

Lss-Ankit:
Thanks @jeanregisser  for the update.

PR #5316, opened 2020-10-14
[celotool] Add celostats config for multiproxy

gastonponti:
### Description

Change templates in celotool, to make the first proxy of a multiproxy configuration to be the one that send metrics to celostats

### Tested

Manually

In the custom environment set (for example)
```
GETH_NODE_DOCKER_IMAGE_TAG=<multiproxy commit>
VALIDATORS="4"
VALIDATOR_PROXY_COUNTS=1:3,1:2,1:1,1:0
```
then deploy the `testnet` and the `celostats`

### Backwards compatibility

Yes


jcortejoso (approved):
Lgtm

PR #5326, opened 2020-10-15
Use parsed deposit for value

nambrot:
### Description

This will support the use of scientific notation for the deposit of a governance proposal


yorhodes (approved):
LGTM

PR #5336, opened 2020-10-15
Fix celocli's attestation service test result output

kphed:
### Description

Successfully testing the attestation service (v1.0.5) with the celocli (v0.0.58) currently results in a `Request was not successful` output. The false-negative seems to be caused by the test response-checking condition (see change).

### Tested

Tested using the following command: `celocli identity:test-attestation-service`.

### Backwards compatibility

Change does not affect anything other than the celocli's identity test output.


timmoreton (approved):
LGTM! thanks for submitting

kphed:
> LGTM! thanks for submitting

Thanks Tim. I rebased the latest master but it looks like the CI's vuln checks keep failing. Any suggestions?

PR #5337, opened 2020-10-15
Allow user to use AWS HSM Wallet without having DescribeKey permission for all keys

tkporter:
### Description

The AWS HSM wallet will list all keys and then attempt to describe all the keys. If a user doesn't have `DescribeKey` permission for all keys, initializing the wallet will fail. This change will skip keys that the user does not have permissions for.

Also, to keep with the pattern in contractkit of not exporting classes as the default module, `AwsHsmSigner` and `AwsHsmWallet` are now not default exports. This is a breaking change if anyone is currently relying on the aws wallet (not that I know of?)

### Other changes

n/a

### Tested

Ran on oracle

### Related issues

n/a

### Backwards compatibility

`AwsHsmSigner` and `AwsHsmWallet` are now not default exports


AlexBHarley (approved):
Makes sense!

PR #5340, opened 2020-10-15
Remove version numbers from wrong contracts/libraries

nambrot:
### Description

There are a few libraries/contracts which should not have `getVersionNumbers` as they are not deployed/linked.

### DIFF
--- a/packages/protocol/contracts/common/UsingPrecompiles.sol
+++ b/packages/protocol/contracts/common/UsingPrecompiles.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.5.13;
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "../common/interfaces/ICeloVersionedContract.sol";
 
-contract UsingPrecompiles is ICeloVersionedContract {
+contract UsingPrecompiles {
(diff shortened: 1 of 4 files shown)


packages/protocol/contracts/common/UsingPrecompiles.sol
```diff
@@ -17,14 +17,6 @@ contract UsingPrecompiles is ICeloVersionedContract {
   address constant GET_PARENT_SEAL_BITMAP = address(0xff - 10);
   address constant GET_VERIFIED_SEAL_BITMAP = address(0xff - 11);
 
-  /**
```
yorhodes, line 20: also remove `ICeloVersionedContract`

packages/protocol/contracts/governance/SlasherUtil.sol
```diff
@@ -28,14 +28,6 @@ contract SlasherUtil is
 
   event SlashingIncentivesSet(uint256 penalty, uint256 reward);
 
-  /**
```
yorhodes, line 31: ditto

yorhodes (approved):
2 issues

PR #5345, opened 2020-10-15
Add fee estimation for Withdrawing CELO and pay for Withdrawing and Selling CELO fees in CELO

### LINKED ISSUE #5258: Exchange and Withdraw CELO transcations fail when moving full balance
### Expected Behavior
Users SNBAT withdraw their full balance because they won't have any left to pay gas with

### Current Behavior
When attempting to exchange CELO for cUSD (or withdraw CELO), our balance check has hard coded 0 as the fee estimate. This results in the transaction being reverted if the user attempts to exchange or withdraw their full balance.

https://celo-org.slack.com/archives/C018VLA3YAK/p1601666037020700

gnardini:
### Description

Add a fee calculation for withdrawing CELO. This will be removed soonish when we do offline fee calculations (it's an epic in the upcoming milestone). This is meant as a quick fix in the meantime to not have a 0 hard-coded.
When withdrawing CELO or selling CELO. The fee is now paid for with CELO instead of cUSD.

I *didn't* add fee estimations for exchanging because it uses more than one transaction and made the calculation somewhat more complex, it will probably be improved doing offline estimations.

### Tested

Tested it in an emulator, just looking at the estimate and paying attention to the exact CELO balance in the logs when doing different operations to see that the balance changes more than expected (meaning that gas was paid for in CELO) only when doing withdrawals and selling CELO.

### Related issues

- Fixes #5258


packages/mobile/src/exchange/WithdrawCeloScreen.tsx
```diff
@@ -25,14 +25,20 @@ import { HeaderTitleWithBalance, headerWithBackButton } from 'src/navigator/Head
...
 import useSelector from 'src/redux/useSelector'
+import { getSendFee } from 'src/send/saga'
 import { useDailyTransferLimitValidator } from 'src/send/utils'
 import DisconnectBanner from 'src/shared/DisconnectBanner'
+import { divideByWei } from 'src/utils/formatting'
+import Logger from 'src/utils/Logger'
 
 type Props = StackScreenProps<StackParamList, Screens.WithdrawCeloScreen>
 
+const RANDOM_ADDRESS = '0xDCE9762d6C1fe89FF4f3857832131Ca18eE15C66'
```
gnardini, line 36: We don't have an address here yet to do the estimation, and estimating requires an address, so I used a random one. Could that cause any issues?

packages/mobile/src/transactions/send.ts
```diff
@@ -114,6 +115,15 @@ export function* sendTransactionPromises(
   const fornoMode: boolean = yield select(fornoSelector)
   let gasPrice: BigNumber | undefined
 
+  // If stableToken is prefered to pay fee, use it unless its balance is Zero,
+  // in that case use CELO to pay fee.
+  // TODO: Make it transparent for the user.
+  // TODO: Check for balance should be more than fee instead of zero.
+  const feeCurrency =
+    preferredFeeCurrency === CURRENCY_ENUM.DOLLAR && stableTokenBalance.isGreaterThan(0)
```
gnardini, line 123: This was working like this before and I was worried about changing it in case someone only has CELO and needs to pay for gas for something, but I would understand not wanting to have it anymore.

packages/mobile/src/transactions/send.ts
```diff
@@ -126,19 +136,18 @@ export function* sendTransactionPromises(
...
     }
 
-    gasPrice = yield getGasPrice(CURRENCY_ENUM.DOLLAR)
+    gasPrice = yield getGasPrice(feeCurrency)
   }
+
+  const feeCurrencyAddress =
+    feeCurrency === CURRENCY_ENUM.DOLLAR
+      ? yield call(getCurrencyAddress, CURRENCY_ENUM.DOLLAR)
+      : undefined
```
jeanregisser: nit: Would be nice to add a small comment explaining it's normal that we pass `undefined` when the `feeCurrency` is `CELO`.

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
```diff
@@ -51,6 +57,25 @@ function WithdrawCeloScreen({ navigation }: Props) {
...
+      false
+    )
+      .then((estimate: BigNumber) => {
+        setFeeEstimate(divideByWei(estimate))
+      })
+      .catch((error) => {
+        Logger.warn('WithdrawCeloScreen', `Error found while estimating gas fee: ${error}`)
+      })
+  }, [])
+
```
jeanregisser: Async functions should be avoided in general in `useEffect` because they can trigger actions after the component has been unmounted.

`useAsync` from [react-async-hook](https://github.com/slorber/react-async-hook) solves this.

Now we have [CalculateFee](https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/fees/CalculateFee.tsx) which uses `useAsync` under the hood and mostly does what you want here.

I originally created it as a component because most components we had at the time, were class components.
But we could export a `useSendFee` hook [from it now](https://github.com/celo-org/celo-monorepo/blob/e67eba1cc9a54cdc6ed38425eea98a35e743aa6f/packages/mobile/src/fees/CalculateFee.tsx#L92-L111).

Added benefit is it provides loading/error states so we can display a spinner while it's calculating and it displays an error if it fails.

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
jeanregisser: This is fine, we already do this [here as well](https://github.com/celo-org/celo-monorepo/blob/2588f4de9691b2944e8f18069d3cf44a77bb3ec1/packages/mobile/src/fees/saga.ts#L24).

jeanregisser (commented):
Great work! 👍

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
gnardini: Thanks for this! I had run into this many times and always solved it in this inelegant way, the `useAsync` hook looks great.

packages/mobile/src/exchange/WithdrawCeloReviewScreen.test.tsx
```diff
@@ -14,6 +14,7 @@ const SAMPLE_AMOUNT = new BigNumber(5.001)
 const mockScreenProps = getMockStackScreenProps(Screens.WithdrawCeloReviewScreen, {
   recipientAddress: SAMPLE_ADDRESS,
   amount: SAMPLE_AMOUNT,
+  feeEstimate: new BigNumber(1),
```
nategraf, line 17: Is this 1 wei, and if so, is that displayed to the user at all? If the fee was actually 1 wei, it would likely not be significant enough to warrant mentioning (unless they are trying to withdraw 100% of their balance, in which case we might help craft a request that won't leave any dust behind.)

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
```diff
@@ -27,15 +29,18 @@ import { StackParamList } from 'src/navigator/types'
...
 
+const RANDOM_ADDRESS = '0xDCE9762d6C1fe89FF4f3857832131Ca18eE15C66'
+
 function WithdrawCeloScreen({ navigation }: Props) {
   const [accountAddress, setAccountAddress] = useState('')
   const [celoInput, setCeloToTransfer] = useState('')
 
   const goldBalance = useSelector((state) => state.goldToken.balance)
-  const goldBalanceNumber = new BigNumber(goldBalance || 0)
+  const goldBalanceNumber = useMemo(() => new BigNumber(goldBalance || 0), [goldBalance])
```
tarikbellamine: Why use `useMemo` here?

My understanding for the best use cases of this is for when there are expensive calculations involved, otherwise the storage costs are not worth it.

nategraf (commented):
All the parts that I am familiar with look good. I am less familiar with the react native components, so I am going to defer to @jeanregisser for final approval.

packages/mobile/src/exchange/WithdrawCeloReviewScreen.test.tsx
gnardini: Thanks for reviewing! The wei to CELO mapping happens in `WithdrawCeloScreen`, [line 67](https://github.com/celo-org/celo-monorepo/pull/5345/files#diff-976dad371e7a9454192fe31ac3ed6c2a45ada63cee14937bdcfc48f6162832deR67). The converted fee is then passed to this review screen and this is a test of it where I put a != 0 number just to have something there.

packages/mobile/src/transactions/send.ts
tarikbellamine: Doesn't this create a potential discrepancy between what we show users and what we actually do? For example, if someone has both cUSD and CELO, we will always show them the fee in CELO but then will always deduct their fee from their cUSD balance.

If my understanding is correct, I think we need to get rid of this. It looks like you've set the default currency = cUSD for the param in the relevant function calls but I'd definitely test it out to be safe.

tarikbellamine (commented):
Nice work! One comment I think needs to be addressed

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
gnardini: Yeah, great catch, I'd like not to have it either, the problem is that we're creating a new `BigNumber` instance on each render and we pass this to the `useSendFee` hook, which uses it to see whether dependencies changed or not. If we pass different instances it assumes it changed and re-calculates the fee, which is undesired. The `useMemo` caches the instance of `BigNumber` so we use always the same unless the total balance changes for some reason.

tarikbellamine: Oh, that's an interesting problem and a creative solution. 

I don't feel strongly about doing this now but an alternative is to pass an `int` value instead and convert to `BigNumber` within the `useSendFee` function. If this would require a bunch of other changes, I'd understand if you'd rather not do it. But if that's the case, please leave a "to-do" so we fix this later.

As a side note, do you think we are having this issue in other places we are doing fee estimates?

gnardini:
> Doesn't this create a potential discrepancy between what we show users and what we actually do? For example, if someone has both cUSD and CELO, we will always show them the fee in CELO but then will always deduct their fee from their cUSD balance.
>
> If my understanding is correct, I think we need to get rid of this. It looks like you've set the default currency = cUSD for the param in the relevant function calls but I'd definitely test it out to be safe.

Before this change we used cUSD if there was enough balance and CELO otherwise.

The new parameter `preferredFeeCurrency` is passed as `GOLD` for the two cases where we would prefer to pay for gas with CELO (withdrawing and selling  CELO) and `DOLLAR` otherwise. In those two cases, the condition would evaluate to `false` and we would get a `feeCurrency` of `GOLD` so we'd pay for gas with CELO.

The condition is to maintain the logic we used to have for all other cases (where `preferredFeeCurrency` is `DOLLAR`). I just didn't want to risk forcing an operation to fail if the user has no cUSD if they have enough CELO to pay for gas. Maybe that's not what we want though, I'm just keeping what was already there for all cases except the two I mentioned.

packages/mobile/src/transactions/send.ts
tarikbellamine: Gotcha, I totally misread this the first time. Sounds good!

tarikbellamine (approved):
LGTM!

packages/mobile/src/components/CeloAmountInput.tsx
```diff
@@ -43,13 +55,20 @@ export default function CeloAmountInput({
...
+          <ActivityIndicator size="small" color={colors.goldUI} />
+        </View>
+      )}
     </TextInputWithButtons>
   )
 }
 
 const styles = StyleSheet.create({
   maxAmount: fontStyles.small600,
+  loadingContainer: {},
```
jeanregisser:
```suggestion
```

jeanregisser (approved):
Great! 👍

packages/mobile/src/exchange/WithdrawCeloScreen.tsx
gnardini: Missed this comment somehow. I changed the `amount` to a `string` since I saw it was being transformed to string right after anyways.

I checked the other places but didn't find something like this happening

Lss-Ankit:
Hi @jeanregisser , @gnardini  Verified task using latest Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found the following:

- CELO is cut as fee while withdrawing CELO and Selling CELO. 
- Observed that cUSD balance is not effected due to withdraw and Sell CELO. 
- Observed that Fee .001 is shown while withdrawing CELO but Fee is not cut as CELO 0.001 it slight different. we have it checked from https://explorer.celo.org. 
- Also, Observed that user is able to withdraw full amount by tapping on MAX button while withdrawing CELO. 


Can you please let us know 3rd and 4th point are expected and Also, let us know if we need to verify anything else. 

Thanks!!

jeanregisser:
All good thanks. The fee is probably rounded up for display.

PR #5347, opened 2020-10-15
Fix make-release args

nambrot:
### Description

- Pass `--dry_run` arg
- Fix `build_directory` arg

### Tested

Not tested


yorhodes (approved):
LGTM
should we document concurrently?

Issue #4891, opened 2020-08-29
Transaction should not be added to the feed if user does not enter pin


Lss-Ankit:
**Frequency:** 100%
**Repro App version:**  Test Flight build Valora app build  v1.0.0 (17) , Android Valora Mainnet build v1.0.0 (16)
**Repro on device:** iPhone 7+ (13.3.1), Google Pixel Xl (8.1) 
**Test Account:** Backup Key: (757) 699-9899. 
**Repro Steps:**
1. Launch the Application and signin with test account
2. Tap on Send button. 
3. Select any valora contact user. 
4. Enter amount and tap on Review button 
5. Tap on Send button. 
6. Tap on < arrow from Enter PIN screen. 
7. Observe. 

**Current Behavior:** “Confirming…” text is shown continuously for transaction
**Expected Behavior:**  Transaction activity should not be shown. 
**Investigation:** Transaction activity with confirming text disappeared after cold launching the app. 
**Impact:** Users may get confused / frustrated by seeing transaction activity with Confirming text. 
**Attachment:** [Confirmissue.mp4](https://drive.google.com/file/d/1K8AZ_ZUmguQ_IjUREsTNdBTuR78aYgwK/view?usp=sharing) (Bug observe at 18 sec)

Lss-Ankit:
Hi @tarikbellamine I have verified this issue using latest Android internal build v1.3.0(1004294316) and found issue is fixed.
User redirect to send page when tapping on < button on enter Pin page 
Repro on devices: Vivo V11 pro (10),iPhone 6+ (12.4.5)

PR #5396, opened 2020-10-16
[Wallet] Don't start transactions until the PIN is correctly input

### LINKED ISSUE #5276: Hitting "Back" on the PIN flow does not gracefully cancel the transaction
### Expected Behavior

Backing out of the PIN flow should cancel the current transaction (instead of proceeding with the transaction and then showing a tx failure)

### Current Behavior

cancelling from the pin navigates to the same screen as success, then shows a failed transaction (noticed in verification)

gnardini:
### Description

When sending cUSD or withdrawing CELO, if the user pressed the back button in the PIN input screen the user was redirected to the starting screen and the transaction still shows up on the feed but is never confirmed or shows as failed. With this change, if pressing the back button is redirected to the confirmation screen again so that they can chose whether they want to go through with the transaction or not.

### Tested

In the emulator, doing transactions and pressing the back button to see that the behaviour is the intended one.

### Related issues

- Fixes #5276


packages/mobile/src/exchange/WithdrawCeloReviewScreen.tsx
```diff
@@ -26,23 +26,21 @@ import { navigate, navigateBack } from 'src/navigator/NavigationService'
...
 
 type Props = StackScreenProps<StackParamList, Screens.WithdrawCeloReviewScreen>
 
 function WithdrawCeloReviewScreen({ route }: Props) {
   const { amount, recipientAddress } = route.params
   const { t } = useTranslation(Namespaces.exchangeFlow9)
-  // loading is never set to false, when the withdrawal is complete or after a short while,
-  // withdrawCelo saga will navigate to |ExchangeHomeScreen|.
-  const [loading, setLoading] = useState(false)
+  const isLoading = useSelector((state: RootState) => state.exchange.isLoading)
```
jeanregisser: nit: we have a typed `useSelector` in src/redux/useSelector`

packages/mobile/src/navigator/types.tsx
```diff
@@ -141,6 +141,7 @@ export type StackParamList = {
   [Screens.PincodeEnter]: {
     withVerification?: boolean
     onSuccess: (pin: string) => void
+    onCancel?: () => void
```
jeanregisser: Why not making this param mandatory? 
This would enforce that all places using `PincodeEnter` handle the cancel case,

packages/mobile/src/exchange/saga.ts
```diff
@@ -400,16 +402,23 @@ function* withdrawCelo(action: WithdrawCeloAction) {
...
     if (context?.id) {
       yield put(removeStandbyTransaction(context.id))
     }
 
-    yield put(showError(ErrorMessages.TRANSACTION_FAILED))
+    if (error.message === ErrorMessages.INCORRECT_PIN) {
+      yield put(showError(ErrorMessages.INCORRECT_PIN))
+    } else {
+      yield put(showError(ErrorMessages.TRANSACTION_FAILED))
+    }
```
jeanregisser: nit: ok to skip that now, but related to our previous discussion where we should migrate toward using events and avoid dispatching multiple actions sequentially, both `removeStandbyTransaction` and `showError` could be directly handled at the reducer level from the `withdrawCeloFailed` event (adding the necessary info to act on it of course).

jeanregisser (approved):
Looking good! 👍 

@i1skn I know you spent some time on this a couple of months ago, could you have a look too? 🙏

packages/mobile/src/exchange/saga.ts
gnardini: Yeah, you're right. I feel like this is a good time to start working on this, otherwise I'll probably forget about it next time I'm modifying something like this again. I changed it so that only one action is dispatched and that action is handled in the 3 reducers (the exchange one, the remove transaction one and the alerting one). Let me know what you think :)

Lss-Ankit:
@gnardini  @jeanregisser  Verified issue using Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found the following.
- "Failure sending payment" message is shown and user redirect to Sending screen when user taps on < back button from ENTER PIN screen. 
It is seems intended behavior to me as well because previously Confirming... text is shown when user taps on < back button from Enter PIN screen. 

FYI, we have already closed our logged issue  for the same. #4891

PR #5398, opened 2020-10-16
Add positive output to the verify-bytecodes script

m-chrzan:
### Description

Previously the script would just exit with code 0 and no stdout output about success. Added logs reporting which contracts have been checked and at which address on the network.

### Other changes

`getImplementationAddress` was previously returning pre-release 1 library addresses without a "0x" prefix, fixed this for consistency in logs.

Improved the documentation comment in `verify-bytecodes.ts`.

### Tested

`yarn verify-bytecodes -b rc1 -n rc1 -r -f` succeeds outputs the new logs.

### Related issues

- Part of #4812 

### Backwards compatibility

Only cosmetic script change.


ewilz:
Do we need to add the same flag inside `verify-release.sh` as well, or is that different?
In thinking about suppressing the logs for these scripts I was thinking of adding an optional `--verbose` flag. would having both `--quiet` and `--verbose` be confusing?

nambrot (approved):

Issue #2701, opened 2020-02-07
'Exchange & Security Fee' is shown as '$0.00' on the 'Buy/Sell Gold Confirmation' screen even though the dollar gets cut with the exchange fee. 


Lss-Ankit:
**Frequency:** 100%
**App version:** iOS Celo Alfajores Wallet Testflight  (1.6.0 (21) ), Internal Android Celo Alfa v1.6.0. 
**Repro on:** Google Pixel XL 2 (9.0), Samsung Galaxy A8+ (8.0.0), 
**Testing Account:** 7228953526/backup Key

**Repro Steps:** 
1) Launch the app and login with testing account.
2) Tap on “Gold” tab. 
3) Tap on “Buy” gold. 
4) Enter amount and tap on “Review” button. 
5) Observe Exchange fee shows 0.00. 
6) Tap on “Buy # gold” button. 
7) Observe Dollars cuts with exchange fee. 

**Current Behavior:** Exchange & Security Fee' is shown as '$0.00' 
**Impact:** User will be confused as amount is cut with exchange fee where as Exchange Fee shows 0 on "Buy/Sell confirmation" screen  
**Expected Behavior:** “Exchange & Security fee” should be shown. 
**Attachment:** [Android_exchange fee gets cut but not showning.mp4 ](https://drive.google.com/file/d/1TNK7q5K0IHjYnuxaBZf4hV39vdwWNm1p/view?usp=sharing)(Bug observe at 15 and 1:15 sec)

annakaz:
Hey @Lss-Ankit the video ends at 1:11 so I cannot see the bug you referenced at 1:15. Could you post another video? Thanks!

Lss-Ankit:
@annakaz Sorry for the inconvenience. I have re-verified this issue and observed that issue is reproducing partially now in Android & iOS. 

Verified on builds : iOS Test flight build : v1.6.3 (25) & Android Celo Alfa build : v1.6.2 (104794504)

Verified on Devices : iPhone SE (13.3) & Samsung Galaxy J7 Prime (6.0.1)

- **In Android,** Respective estimation & security fee is shown on 'Buy' gold confirmation screen. Issue reproduce on 'Sell gold' confirmation screen. '0' amount is shown for estimation fee in 'Sell Gold' confirmation screen. 
[...]

PR #5415, opened 2020-10-16
[Wallet] Allow up to 1% difference on exchange rate for exchanging CELO

### LINKED ISSUE #5279: Soften minBuyAmount restriction on CELO exchanges
### Expected Behavior
Users should be able to exchange without CELO for cUSD without having their transactions reverted due to shifts in the exchange rate.

### Current Behavior
When the exchange rate shifts too much in between exchange estimation and execution, the transaction is reverted. How we define "too much" is probably too aggressive given we've seen this error happen fairly frequently, it creates a confusing user experience to throw a failing error, and user expectations are probably closer to that of a market order than limit order.

https://celo-org.slack.com/archives/CL7BVQPHB/p1602003881149900?thread_ts=1601931476.130200&cid=CL7BVQPHB

https://celo-org.slack.com/archives/CE8KJ192M/p1602115985037800?thread_ts=1602072921.028900&cid=CE8KJ192M

https://sentry.io/organizations/celo/issues/1840948860/?project=1250733&query=is%3Aunresolved+Calculated+buyAmount+was+less+than+specified+minBuyAmount&statsPeriod=14d

gnardini:
### Description

Before this PR, if we wanted to buy X CELO we were accepting to receive at a minimum X-0.01. This caused several errors, especially with bigger numbers since the exchange rate can vary between the estimation and the execution of the request. This PR changes that to accept at least X*0.99 of the input amount.

### Tested

Adding a log for the limit and using different values to see that the minimum is 0.99 * the input amount.

### Related issues

- Fixes #5279


tarikbellamine (approved):
Nice!

Lss-Ankit:
Hi @gnardini We have verified task using Test flight build v1.3.0(27) and Android play store internal build v1.3.0(1004294316) and found the following: 
- 0.010 cUSD (Exchange fee)  is cut when user buy 0.01 CELO. 
- 0.020 cUSD (Exchange fee) is cut when user buy 0.02 CELO
- 1.94 cUSD (Exchange fee) is cut when user buy 1 CELO. 

Also, we have observed that Exchange fee is shown 0 on Buy CELO screen. 
Attachment: 
CELO_Exchangefee

can you please confirm is it expected behavior or we need to file a bug for the same? 
Also, please let us know if we need to verify any thing else ? 
Thanks!

gnardini:
Hi @Lss-Ankit 

This is a bit of a behind the scenes change. The only thing to verify is that the exchanges still go through successfully, the rest wasn't modified by this task.

Thanks for verifying!

Lss-Ankit:
Okay thanks @gnardini  for  the update. 
And FYI, for exchange fee 0 we have bug https://github.com/celo-org/celo-monorepo/issues/2701

PR #5418, opened 2020-10-16
Suppress dependent output in verify-bytecode/check-versions/verify-release

nambrot:
### Description

When running the verification scripts, output of dependent commands like build commands should not be shown to the user. This PR adds an `-l` flag to these scripts to specify a log file, defaulting to `/dev/null`.

### Tested

- #4812


packages/protocol/scripts/bash/check-versions.sh
```diff
@@ -50,3 +58,6 @@ fi
 # MultiSig contracts, and the ReleaseGold contract.
 CONTRACT_EXCLUSION_REGEX=".*Test|Mock.*|I[A-Z].*|.*Proxy|^LinkedList$|^SortedLinkedList$|^SortedLinkedListWithMedian$|MultiSig.*|ReleaseGold|FixidityLib|MetaTransactionWallet|SlasherUtil|UsingPrecompiles"
 yarn ts-node scripts/check-backward.ts sem_check --old_contracts $BUILD_DIR_1/contracts --new_contracts $BUILD_DIR_2/contracts --exclude $CONTRACT_EXCLUSION_REGEX $REPORT_FLAG
+
+echo " - Return to original git ref"
+git checkout $ORIGINAL_GIT_REF > $LOG_FILE
```
ewilz: no new line?? do we care ¯\_(ツ)_/¯

ewilz:
if we care to see the logs, it would be extremely convenient to print them in the console. still need to learn my way around bash but is that easy to do with the -l flag param?? Having the console be the default would be awesome, but I can just learn the command too. In my experience, `--verbose` is more common to see detailed output, but not a blocker.

packages/protocol/scripts/bash/check-versions.sh
nambrot: Sorry, what do mean by new line here?

nambrot:
Yup, one can still do it with the `-l` flag and just specify a log file. I think as developers you'll likely want to run with it, but for Celo stakeholders, the output is more often than not distracting IMO?

packages/protocol/scripts/bash/verify-release.sh
```diff
@@ -11,18 +11,21 @@ set -euo pipefail
...
+# -l: Path to a file to which logs should be appended
 
 BRANCH=""
 NETWORK=""
 PROPOSAL=""
 FORNO=""
+LOG_FILE="/dev/null"
 
-while getopts 'b:n:p:f' flag; do
+while getopts 'b:n:p:fl:' flag; do
```
yorhodes, line 22: `f:l`

packages/protocol/scripts/bash/check-versions.sh
yorhodes: file ends without new line

yorhodes (approved):
1 fix

packages/protocol/scripts/bash/check-versions.sh
nambrot: ohhh

packages/protocol/scripts/bash/verify-release.sh
yorhodes: jk

ewilz:
> Yup, one can still do it with the `-l` flag and just specify a log file. I think as developers you'll likely want to run with it, but for Celo stakeholders, the output is more often than not distracting IMO?

I guess what i"m saying is, defaulting the output to the console  (NOT a file) is really convenient. BUt I think I would just be able to use `&3` or something if I MUST specify...



EDIT: Actually I don't see why it's either: don't print logs OR print logs to console. Specifying a file instead of just having it print to the console seems like overkill to me. I'm being nit picky, don't want to hold this up, but just my thoughts on this.

PR #5419, opened 2020-10-16
Fix the make-release script

m-chrzan:
### Description

Some regressions sneaked their way into the make-release script.

* missing `;;` in `case` statement
* wrong artifacts directory passed (extra `/contracts` appended)
* the "all libraries have been relinked" check could fail on ignored contracts

This PR also makes it such that when running `make-release.sh`, it will return to the current branch to run `make-release.js`. Similar steps should be taken with `check-versions` (and potentially other scripts in the backwards compatibility suite?). 

### Other changes

There is a temporary hack to increase the gas limit when making the release. This should ideally be captured in a more global config?

### Tested

`yarn make-release -b celo-core-contracts-v1.rc1 -n development -p proposal.json -i example-initialize-data.json -r report.json` against a previously generated devchain with rc1 contracts, and a correctly generated `report.json` .

### Related issues

- Part of #4812

### Backwards compatibility

Should be good.


nambrot (approved):

packages/protocol/scripts/bash/make-release.sh
```diff
@@ -45,4 +45,14 @@ yarn build
 rm -rf $BUILD_DIR && mkdir -p $BUILD_DIR
 mv build/contracts $BUILD_DIR
 
-yarn run truffle exec ./scripts/truffle/make-release.js --network $NETWORK --build_directory $BUILD_DIR/contracts --report $REPORT --proposal $PROPOSAL --initialize_data $INITIALIZE_DATA $DRYRUN
+git checkout -
+yarn build
+
+# TODO: remove the truffle_override
+yarn run truffle exec ./scripts/truffle/make-release.js \
```
ewilz, line 52: yessss multil-line FTW!

PR #5422, opened 2020-10-16
Metadata crawler posts to Discord

### LINKED ISSUE #5379: Add Discord Bot Alerting for Metadata Crawler

timmoreton:
### Description

* Metadata crawler posts problems with Attestation Service to Discord, if an env var is specified
* Add this webhook URL to the encrypted env file for rc1, and extend blockscout Helm chart to pass it through to metadata crawler.

### Tested

Locally so far

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/5379

### Backwards compatibility

Yes


packages/metadata-crawler/src/crawler.ts
```diff
@@ -199,6 +209,39 @@ async function processAttestationServiceStatusForValidator(
...
+  if (discordWebhook && isElected) {
+    const currentValid =
+      status.state === AttestationServiceStatusState.NoAttestationSigner ||
+      status.state === AttestationServiceStatusState.UnreachableHealthz ||
+      status.state === AttestationServiceStatusState.Valid
+
+    if (!currentValid) {
+      const content =
+        `:no_mobile_phones: **Problem with Attestation Service!** ${status.name} \n` +
+        `For validator \`${validator.address}\` in group \`${validator.affiliation}\`\n` +
```
nambrot, line 223: Consider adding the name here

packages/celotool/src/lib/blockscout.ts
```diff
@@ -90,6 +90,7 @@ async function helmParameters(
       envVar.BLOCKSCOUT_SUBNETWORK_NAME,
       celoEnv
     )}"`,
+    `--set blockscout.discord_webhook_url=${fetchEnvOrFallback(envVar.METADATA_CRAWLER_DISCORD_WEBHOOK, '')}`,
```
nambrot: I know we still have this on the blockscout helm chart, but maybe separate it to something like metadata-crawler namespace?

nambrot (approved):
Sgtm, the only question is if it might get a bit spammy since we run this quite frequently?

packages/metadata-crawler/src/crawler.ts
aslawson: +1 to adding Validator name and group name if available -- sometimes the validator name doesn't make it immediately obvious which group it belongs to

aslawson (commented):
> the only question is if it might get a bit spammy since we run this quite frequently?
@nambrot what is the current interval that this crawler runs at?

aslawson (approved):

nambrot:
It currently runs every 30 minutes I believe

packages/metadata-crawler/src/crawler.ts
timmoreton: We already display the validator name with `${status.name}`.  Group name is a bit harder.

timmoreton: @nambrot are we getting discord name claims soon? would love to @ mention the handle in the metadata.

timmoreton:
> Sgtm, the only question is if it might get a bit spammy since we run this quite frequently?

I agree.. Let's see how it goes. We could reduce interval to 60mins. A better solution like recording status between runs is the point at which we break this thing off from the blockscout piece I think!

PR #5430, opened 2020-10-19
[Wallet] Prevent exiting force update screen

### LINKED ISSUE #5429: Fix "Force Update" mechanism in Valora
### Expected Behavior

When the app version is lower than the minimum upload, a non-dismissable Update screen should appear.

### Current Behavior

The update screen can be dismissed.

gnardini:
### Description

The Update screen could be dismissed before this PR. It shouldn't be dismissable, the intention is to force the update.

I had to add the min required version in Redux to force the screen when backgrounding and reopening the app since if the app was reloaded the screen could still be skipped. This happened specially on Android.

### Tested

Locally changing the app version on both platforms.

### Related issues

- Fixes #5429


jeanregisser (approved):
Thanks!! 🥇 

Just to be sure, this also prevents dismissing using the Android back button, or "swipe back" on iOS, right?

gnardini:
@jeanregisser I'm sorry, I updated the PR just before your comment :( You might want to take another look.

It does prevent swiping back and pressing the back button, yes!

packages/mobile/src/navigator/NavigatorWrapper.tsx
```diff
@@ -112,6 +119,15 @@ export const NavigatorWrapper = () => {
...
   }
 
+  let updateRequired = false
+  if (minRequiredVersion) {
+    const version = DeviceInfo.getVersion()
+    updateRequired = isVersionBelowMinimum(version, minRequiredVersion)
+    if (!updateRequired) {
+      dispatch(setRequiredVersion(null))
+    }
+  }
```
jeanregisser: We shouldn't dispatch during render.
This logic should live in a saga.

packages/mobile/src/app/actions.ts
```diff
@@ -26,6 +26,7 @@ export enum Actions {
   UNLOCK = 'APP/UNLOCK',
   SET_SESSION_ID = 'SET_SESSION_ID',
   OPEN_URL = 'APP/OPEN_URL',
+  SET_REQUIRED_VERSION = 'APP/SET_REQUIRED_VERSION',
```
jeanregisser: Would be nice to avoid introducing new setters and try to gradually migrate to events.
How about we use `MIN_APP_VERSION_DETERMINED` instead?

jeanregisser (changes requested):

packages/mobile/src/navigator/NavigatorWrapper.tsx
gnardini: This is run only once. Which saga do you think this belongs on? I understand where you're coming from, but this seemed like the most intuitive way to solve this and I don't see a big risk dispatching here.

packages/mobile/src/navigator/NavigatorWrapper.tsx
```diff
@@ -46,7 +51,21 @@ export const NavigatorWrapper = () => {
...
+    if (!minRequiredVersion) {
+      return false
+    }
+    const version = DeviceInfo.getVersion()
+    const versionBelowMinimum = isVersionBelowMinimum(version, minRequiredVersion)
+    if (!versionBelowMinimum) {
+      dispatch(minAppVersionDetermined(null))
+    }
+    return versionBelowMinimum
+  }, [minRequiredVersion])
```
jeanregisser: Still feels wrong, I don't understand why we need to call `dispatch` here.

This should be in a selector I think.

gnardini: Well, I just wanted to avoid doing this device check every time, if we already know the user updated the app I feel it would be better to have the `minRequiredVersion` as null but I guess we can just not dispatch anything and check against the latest minRequiredVersion found every time. 

Is something like this what you had in mind?

tarikbellamine (commented):
Question - how is the swipe and hardware back button being disabled?

My understanding is that swipes need to be disabled like this: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/navigator/Headers.tsx#L24

and hardware back button needs to be overwritten with an event listener like this: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/identity/PhoneNumberLookupQuotaScreen.tsx#L42

gnardini:
> Question - how is the swipe and hardware back button being disabled?

They were disabled with two mechanisms, depending on the situation:

- The first one is when going through the Firebase saga, which happens when the version mismatch is found. I changed the `navigate` by a `replace`, so the stack contains only the update screen. Swiping doesn't work and if the user presses the back button on android the app is just closed since it's the only screen in the stack.

- If the app is reloaded somehow (can happen if the app is backgrounded for a period of time or on old phones) then the redux state is used and the same thing that is used for the pincode on open is used. Basically the update screen is drawn on top of the home screen, so the user can't interact with anything and just sees the update screen. Swiping or pressing back doesn't work because it's not a new screen on the stack, it's just drawn on top of it.

The two cases you pointed out are for when there are other screens in the stack, but it's not the case here.

packages/mobile/src/navigator/NavigatorWrapper.tsx
jeanregisser: Yes I think so! Happy to chat about this tomorrow, I might have missed something here.

gnardini:
Changed the implementation so that the minVersion is read in real time from Firebase,

jeanregisser (approved):
Great! :shipit: 👍

Lss-Ankit:
Hey @gnardini i have verified tried to verify this task with taking play store v1.2 but not getting Force update screen. 
can you please let me know how I can check this scenarios with v1.3
Thanks!

PR #5436, opened 2020-10-19
KomenciKit

bowd:
### Description

KomenciKit is a wrapper for the Komenci service.
It will be consumed by Valora and used during fee-less onboarding.

### Other changes

I've extended the `BaseWrapper` with some convenience attributed:

###### eventTypes

This will essentially act as an `enum` of events defined in the contract.
This is useful because we're using the `getPastEvents` which wasn't strictly typed, it used a string as argument.
And we were using it with strings all over the place which is error prone.
The extension allows the compiler to verify that the event passed in exists on the contract.

###### methodIds

This is a mapping between method name and method ID.  It's used in Komenci to have a more ergonomic way to define meta transaction filters.


packages/komencikit/README.md
```diff
@@ -0,0 +1,29 @@
...
+- startSession: Initiate a Komenci session with a solved captcha
+- deployWallet: deploy a MTW for Valora's EOA
+- requestAttestations: request attestations fee-lessly 
+- submitMetaTransaction: submit arbitrary meta transactions via Komenci
+- selectIssuers: wrapper for selectIssuers as meta-tx
+- completeAttestation: wrapper for complete attestations as meta-tx
+
+## User Guide
+
+To start working with Komencikit you need a to pass in a ContractKit instance, the external account
```
codyborn: nit: incomplete sentence

packages/komencikit/README.md
```diff
@@ -0,0 +1,29 @@
...
+import { KomenciKit } from "@celo/komencikit"
+
+const contractkit = newKit('https://alfajores-forno.celo-testnet.org:8545')
+
+const komenciKit = new KomenciKit(contractkit, "0xaaaa...aaaa", {
+    url: 'api.komenci.celo.org' // I made this up
+})
+```
+
+See `examples/attestation-flow.ts` for an example of a full attestation flow executed via KomenciKit
```
codyborn:
```suggestion
See [examples/attestation-flow.ts](./examples/attestation-flow.ts) for an example of a full attestation flow executed via KomenciKit
```

packages/komencikit/examples/attestation-flow.ts
```diff
@@ -0,0 +1,157 @@
...
+  const requestAttestations = await komenciKit.requestAttestations(identifier, walletAddress, 3)
+  const statsAfter = await attestations.getAttestationStat(identifier, walletAddress)
+  console.log('After ===== ')
+  console.log(statsAfter)
+  console.log('RequestAttestations: ', requestAttestations)
+  if (!requestAttestations.ok) {
+    return
+  }
+
+  const events = await attestations.getPastEvents('AttestationsRequested', {
```
codyborn: Can we make these a constant enum defined somewhere? 
`AttestationsRequested` `AttestationIssuerSelected` `WalletDeployed`

packages/komencikit/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@celo/komencikit",
+  "version": "0.0.1-dev",
+  "description": "Celo's KomenciKit user to interact with the Komenci meta-tx server",
```
codyborn:
```suggestion
  "description": "Celo's KomenciKit client to interact with the Komenci meta-tx server",
```

packages/komencikit/src/actions.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export const StartSessionResp = t.type({
+  token: t.string,
+})
+
+export type StartSessionResp = t.TypeOf<typeof StartSessionResp>
+
+export const startSession = action<ActionTypes.StartSession, StartSessionPayload, StartSessionResp>(
+  ActionTypes.StartSession,
+  'POST',
```
codyborn: Can we define this as an enum type if it doesn't exist already?

packages/komencikit/src/client.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    })
+
+    describe('when the request encounters network errors', () => {
+      it('returns a wrapped error', async () => {
+        const act = testAction({})
+        fetchMock.mockAbort()
+
+        const res = await client.exec(act)
+        expect(res.ok).toBe(false)
+        if (res.ok === false) {
```
codyborn, line 39: line 38 will throw if this is not false, right?

packages/komencikit/src/client.ts
```diff
@@ -0,0 +1,86 @@
...
+        const decodedResp = action.codec.decode(payload)
+        if (isRight(decodedResp)) {
+          return Ok(decodedResp.right)
+        } else {
+          return Err(new ResponseDecodeError(payload))
+        }
+      } else if (resp.status === 404) {
+        return Err(new NotFoundError(this.url + action.path))
+      } else if (resp.status === 403) {
+        return Err(new Unauthorised())
```
codyborn, line 74: nit
```suggestion
        return Err(new Unauthorized())
```

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,378 @@
...
+   * @param txHash - the hash of the transaction to watch
+   * @private
+   */
+  private async waitForReceipt(txHash: string): Promise<Result<TransactionReceipt, TxError>> {
+    let receipt: TransactionReceipt | null = null
+    let waited = 0
+    while (receipt == null && waited < this.options.txRetryTimeoutMs) {
+      receipt = await this.contractKit.web3.eth.getTransactionReceipt(txHash)
+      if (receipt == null) {
+        await sleep(100)
```
codyborn: We can also make this a property under KomenciOptions: `txRetryPollingIntervalMs`

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,378 @@
...
+          const deployer = await this.contractKit.contracts.getMetaTransactionWalletDeployer(
+            resp.result.deployerAddress
+          )
+          const events = await deployer.getPastEvents('WalletDeployed', {
+            fromBlock: receipt.blockNumber,
+            toBlock: receipt.blockNumber,
+          })
+
+          const deployWalletLog = events.find(
+            (event) => normalizeAddressWith0x(event.returnValues.owner) === this.externalAccount
```
codyborn: We should also verify that the implementationAddress in the event matches the one sent to the API. This prevents a compromised Komenci node from deploying a malicious MTWProxy instance and Valora using it as its own.

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,378 @@
...
+    identifier: string,
+    walletAddress: string,
+    attestationsRequested: number
+  ): Promise<Result<TransactionReceipt, FetchError | TxError>> {
+    const wallet = await this.getWallet(walletAddress)
+    const nonce = await wallet.nonce()
+    const attestations = await this.contractKit.contracts.getAttestations()
+
+    const approveTx = await attestations.approveAttestationFee(attestationsRequested)
+    const approveTxSig = await wallet.signMetaTransaction(approveTx.txo, nonce)
```
codyborn: nit: it might be better to increment nonce `nonce++` here in case we add more txs in the future since we won't need to worry about ordering.

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,378 @@
...
+    bailOnErrorTypes: [
+      FetchErrorTypes.Unauthorised,
+      FetchErrorTypes.ServiceUnavailable,
+      TxErrorTypes.Revert,
+    ],
+    onRetry: (_args, error, attempt) => {
+      console.debug(`${TAG}/requestAttestations attempt#${attempt} error: `, error)
+    },
+  })
+  public async requestAttestations(
```
codyborn, line 211: We can add the DEK registration to this API as well.

packages/komencikit/README.md
```diff
@@ -0,0 +1,29 @@
+# KomenciKit
+
+The KomenciKit library is a wrapper for the Komenci Service API, which is used for fee-less onboarding by allowing 
+the consumer to execute fee-less attestations with the help of MetaTransactions.
+
+The main actions that KomenciKit exposes are:
+- startSession: Initiate a Komenci session with a solved captcha
+- deployWallet: deploy a MTW for Valora's EOA
+- requestAttestations: request attestations fee-lessly 
+- submitMetaTransaction: submit arbitrary meta transactions via Komenci
```
codyborn: Do we want to allow this? For security, each contract and method should be added manually and building an API to expose it is a good forcing function to perform session checks and input validation.

packages/komencikit/src/actions.ts
```diff
@@ -0,0 +1,144 @@
...
+  'v1/startSession',
+  StartSessionResp
+)
+
+interface GetDistributedBlindedPepperPayload {
+  e164Number: string
```
codyborn, line 78: We should be blinding and unblinding the phone number on the client before it makes its way to Komenci.

codyborn: Sorry just realized we were doing this _within_ Komenci and it needs to be moved out to this client to avoid giving Komenci the phonenumber <> on-chain id mapping. We can do that in a subsequent PR.

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,378 @@
...
+   * 2. Komenci saves the nonce in state, associated with the external account requesting it
+   * 3. KomenciKit signs a message containing the nonce and uses it to call startSession
+   *
+   * @param externalAccount
+   * @returns the signature of the login message
+   * @private
+   */
+  private async getLoginSignature(): Promise<Result<string, LoginSignatureError>> {
+    try {
+      const signature = await this.contractKit.web3.eth.sign(
```
codyborn: We should use EIP712 (signTypedData format) for these types of signatures.

codyborn: If it's significant work, we can roll it in later.

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,378 @@
...
+    return this._wallet
+  }
+
+  /**
+   * Used to create a signature for a login message that allows the server to verify that the
+   * externalAccount passed in is actually owned by the caller.
+   *
+   * TODO: In order to increase security and prevent replay attacks we should introduce a nonce
+   * Flow would be:
+   * 1. KomenciKit asks for a nonce from Komenci
```
codyborn: Some ideas for a replay attack prevention:
1. Sign the `captchaResponseToken` since this can only be used once
2. Have the client generate the unique sessionId and sign this

packages/komencikit/src/retry.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+        impl
+      )
+
+      await expect(instance.testMethod()).resolves.toEqual(Err(new BailError()))
+      expect(impl).toHaveBeenCalledTimes(2)
+    })
+  })
+
+  describe('options:onRetry', () => {
+    it('is called after each unsuccsseful try', async () => {
```
codyborn:
```suggestion
    it('is called after each unsuccessful try', async () => {
```

codyborn (changes requested):
Amazing work, added some questions and nit picks.

packages/komencikit/src/client.spec.ts
bowd: Yeah exactly, but the reason for the `if` is that the typescript compiler doesn't yet have jest expect comprehension sadly (it's a work in progress thing) so even if I add the expectation for false, the compiler doesn't yet discriminate between the `res` union types unless I do the explicit `if` then I can actually access `res.error`.

I was working on some custom jest matchers to help with our `Result` type but I found it a bit weird to type those so I gave up temporarily.

packages/komencikit/README.md
bowd: I get what you're saying. I guess to I saw this as a way to reduce duplication and simplify the logic where the meta-tx is simple and we don't need to add anything on top it could be passed in directly. This would be coupled with server-side validation to only filter allowed calldata and ensure that the wallet called is what is expected.

But I can also see the point of having a clearer KomenciKit API, even if the exposed functions rely on a generalised end-point on Komenci's side.

packages/komencikit/src/actions.ts
bowd: I actually had this on the back of my mind for a while and wanted to ask you about it, because I also had the impression that we had to blind the PN but then saw that komenci was expecting the real thing. Good to know that we need to blind it 👍.

packages/komencikit/src/kit.ts
bowd: Yeah I felt like the data-structure would be to thin to warrant an EIP712. Is there a more general convention about when to use personal message vs EIP712 or is it more like personal message is out we should always use 712?

packages/komencikit/src/kit.ts
bowd: Ah the `catpcahResponseToken` option is really clever love it ❤️ .

packages/komencikit/examples/attestation-flow.ts
bowd: Yeah it's one thing I wanted to look into. Theoretically they should be autogenerated somewhere. I was expecting to see them on the contract wrapper but either that's not expected or there's something weird going on in the solidity->generated types->contract->wrapper. But I'll look into this some more.

packages/komencikit/README.md
bowd: Or I'm not sure if you're referring to also exposing these individually on the Komenci API side as well or not. I guess it would potentially make it easier to validate if we do that.

packages/komencikit/README.md
bowd: I drifted off to sleep 😅

packages/komencikit/src/login.ts
```diff
@@ -0,0 +1,29 @@
+import { Address } from '@celo/contractkit'
+import { EIP712TypedData } from '@celo/utils/lib/sign-typed-data-utils'
+
+export const buildLoginTypedData = (
```
tarikbellamine, line 4: @gnardini is looking into incorporating DeviceCheck and SafetyNet. At least SafetyNet outputs a verifiable token but not sure about DeviceCheck yet. Do we want to accept these as inputs if they are available for the first iteration of Komenci?

packages/komencikit/src/verifyWallet.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const verifyWallet = async (
+  contractKit: ContractKit,
+  walletAddress: Address,
+  allowedImplementations: Address[],
+  expectedSigner: Address
+): Promise<Result<true, WalletValidationError>> => {
+  const code = await contractKit.web3.eth.getCode(walletAddress)
+  // XXX: I'm unsure whether this is safe or if we should store the
+  // bytecode as a constant in `mobile` and pass it into KomenciKit
```
tarikbellamine, line 21: Based on my convos with @asaj, I think we landed on needing a constant bytecode to be available for comparison to ensure the deployed wallet meets expectations. I can't really speak to what other guardrails might be in place though.

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,424 @@
...
+   * getDistributedBlindedPepper: uses the /v1/distributedBlindedPepper endpoint to
+   * request the identifier and pepper associated with a phone number in a fee-less scenario
+   *
+   * @param e164Number - phone number
+   * @param clientVersion
+   * @returns the identifier and the pepper
+   */
+  getDistributedBlindedPepper = async (
+    e164Number: string,
+    clientVersion: string
```
tarikbellamine, line 133: What's `clientVersion` and how will Valora know it?

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,424 @@
...
+  ): Promise<Result<GetDistributedBlindedPepperResp, FetchError>> => {
+    return this.client.exec(getDistributedBlindedPepper({ e164Number, clientVersion }))
+  }
+
+  /**
+   * deployWallet: uses the /v1/deployWallet endpoint to deploy a MetaTransactionWallet Proxy
+   * pointing it to the implementation passed as an argument
+   * The function takes care of waiting for retrying, waiting for receipt and log parsing
+   *
+   * @param implementationAddress the implementation address Valora requires
```
tarikbellamine, line 143: Nit: I think the description would be more clear if you referenced the fact that this will be Valora's EOA address

packages/komencikit/src/kit.ts
tarikbellamine: +1 to DEK registration but think that's now accounted for in the `setAccount` wrapper. I like that approach because it will allow Valora control over when to submit the transaction

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,424 @@
...
+    ],
+    onRetry: (_args, error, attempt) => {
+      console.debug(`${TAG}/requestAttestations attempt#${attempt} error: `, error)
+    },
+  })
+  public async requestAttestations(
+    identifier: string,
+    walletAddress: string,
+    attestationsRequested: number
+  ): Promise<Result<TransactionReceipt, FetchError | TxError>> {
```
tarikbellamine, line 215: It may be worth adding a separate error type to symbolize the "attestation request limit reached" error, which I believe we are initially setting to 10. Perhaps you can add as an error type within `KomenciErrorTypes`?

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,424 @@
...
+
+  /**
+   * Utility function used to instantiate a MetaTransactionWalletWrapper cached by address
+   *
+   * @param txHash - the hash of the transaction to watch
+   * @private
+   */
+  _wallet?: MetaTransactionWalletWrapper
+  private async getWallet(address: string): Promise<MetaTransactionWalletWrapper> {
+    if (this._wallet === undefined || this._wallet.address !== address) {
```
tarikbellamine: Nit: Consider using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,424 @@
...
+  public async requestAttestations(
+    identifier: string,
+    walletAddress: string,
+    attestationsRequested: number
+  ): Promise<Result<TransactionReceipt, FetchError | TxError>> {
+    const attestations = await this.contractKit.contracts.getAttestations()
+    const wallet = await this.getWallet(walletAddress)
+    let nonce = await wallet.nonce()
+
+    const approveTx = await attestations.approveAttestationFee(attestationsRequested)
```
tarikbellamine: Seems like there's a check you can do beforehand in the event that attestations have been requested but not revealed and save yourself from submitting these transactions. Example of how we are using it in the Valora flow: https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/identity/verification.ts#L481

packages/komencikit/src/kit.ts
```diff
@@ -0,0 +1,424 @@
...
+  public async setAccount(
+    name: string,
+    dataEncryptionKey: string,
+    walletAddress: Address,
+    proofOfPossession: Signature | null = null
+  ): Promise<Result<TransactionReceipt, FetchError | TxError>> {
+    const accounts = await this.contractKit.contracts.getAccounts()
+    return this.submitMetaTransaction(
+      walletAddress,
+      accounts.setAccount(name, dataEncryptionKey, walletAddress, proofOfPossession).txo
```
tarikbellamine, line 314: Don't you need to `await` this method call like you're doing elsewhere?

tarikbellamine (commented):
Nice work! Left a couple of comments and questions.

One thing that may make sense to think a little about now is analytics. We've got the Valora verification flow pretty well measured at this point and Komenci will be a relative black box. I'm not sure how much tracking makes sense for the first iteration but just wanted to put it on your radar as a means to collect data for debugging and performance measurement.

packages/komencikit/README.md
```diff
@@ -0,0 +1,31 @@
+# KomenciKit
+
+The KomenciKit library is a wrapper for the Komenci Service API, which is used for fee-less onboarding by allowing 
+the consumer to execute fee-less attestations with the help of MetaTransactions.
+
+The main actions that KomenciKit exposes are:
```
tarikbellamine, line 6: Do we need a wrapper for `escrow.withdraw` or are you thinking to just submit through the generic function? This that's the only onboarding on-chain transaction not represented in this list.

And in general, must a transaction have a wrapper in KomenciKit in order to be subsidized?

packages/komencikit/src/client.spec.ts
codyborn: Ah gotcha, makes sense.

codyborn (approved):
LGTM

packages/komencikit/src/kit.ts
bowd: @codyborn, keep me honest but AFAIK client version is the actual version of Valora, and it's something that Odis requires for tracking I guess. It's currently passed in as `DeviceInfo.getVersion()` in `identity/privateHashing.ts`.

packages/komencikit/src/verifyWallet.ts
bowd: Yeah I guess that was more about how to track the implementation's bytecode itself. We were between using bytecode digests or the actual address of the contract. I was inclining towards bytecode digest because I was under the wrong impression that it would be constant across networks and be easier to manage on Valora - it wouldn't need a different mapping per network. But realistically because we're using library linking it's gonna have different bytecodes therefore we're ok with Valora referring to a MTW implementation by the address, and Komenci verifying server side that address maps to a known implementation.

Here I'm actually verifying the Proxy contract bytecode to ensure that the Proxy itself is what we expect. I suspect we won't need to change the Proxy anytime soon so I think it's safe to extract that from the compilation output. But I'll ping the CAP team to figure out if there are holes in this approach.

packages/komencikit/src/login.ts
bowd: Yeah we can add them, not necessarily to the login struct because I think having the captcha token is enough for replay protection. But currently I stripped these from the client under the impression that we'll stick only with captcha. If we feel confident in adding those I can do that but I'd merge this as is in the mean time to have a starting point for the mobile work.

packages/komencikit/src/kit.ts
bowd: But it's not - but that's also more to your point 😅
This is the address of the MetaTransactionWallet implementation that Valora will have saved as a constant. 
Komenci (api) will verify that the implementation is a known implementation that we deployed, and will then deploy a Proxy and set the implementation of that proxy to this address.
Valora then needs to verify:

- That it's MTW is actually a Proxy (by verifying the bytecode)
- That it's pointing to the chosen implementation
- That Valora's EOA is the signer
- That the MTW is its own Owner

(these are all handled in KomenciKit)

packages/komencikit/src/kit.ts
bowd: I think this depends on how we think about KomenciKit. I would leave this logic in Valora and only think of the `requestAttestations` function here as a drop-in replacement for the transaction that Valora was sending before. When and if to request additional attestations should be up to Valora.

packages/komencikit/src/kit.ts
bowd: It depends on the implementation in the wrapper - these calls result in a `CeloTransactionObject` but some are async, some are not.

packages/komencikit/README.md
bowd: A transaction doesn't need to have a wrapper. I was actually inclined to remove some of the wrappers and rely on the `submitMetaTransaction` action that takes in an arbitrary MetaTransaction and pushes it. 
That will only work, however with transaction that we allow on the API side by filtering on contract address and methodID.

The only transaction that we actually _must_ wrap is the `Attestations.request` logic because there we're doing the actual subsidy batch and not just executing a simple meta transaction.

packages/komencikit/src/kit.ts
bowd: Yeah I didn't add that in because I wanted to do it in a separate PR when I implement it on the server side as well.

packages/komencikit/src/login.ts
tarikbellamine: Sounds good. Feel free to merge as is and we can always add those in once we have more clarity.

packages/komencikit/src/kit.ts
tarikbellamine: Ah gotcha that makes sense

packages/komencikit/src/kit.ts
tarikbellamine: Makes sense, I agree

packages/komencikit/src/kit.ts
tarikbellamine, line 314: The `walletAddress` expected by the `accounts.setAccount` method is actually the EOA not the MTW.

In general, the use of `walletAddress` in KomenciKit to refer to the MTW is confusing because in a broader context we are trying to move towards a world where `accountAddress === MTW proxy address` and `walletAddress === EOA address`. I understand if you want to keep the terminology as is for the rest of Komenci, but I'd suggest changing it in KomenciKit as it will be confusing for future Valora devs

tarikbellamine (changes requested):
Small request regarding terminology and a needed change to the `setAccount` method.

packages/komencikit/src/kit.ts
codyborn: That's right. It's purely for monitoring and lets us know on the ODIS side how many users are still using a given version. This can be helpful when deprecating old code paths.

packages/komencikit/src/kit.ts
```diff
@@ -193,11 +210,11 @@ export class KomenciKit {
   })
   public async requestAttestations(
     identifier: string,
-    walletAddress: string,
+    metaTxWalletAddress: string,
```
tarikbellamine, line 212: Nit: this is the only transaction wrapper that has `identifier` as the first argument and the address as the second. Can we swap these to avoid mistakes when calling?

tarikbellamine (approved):
Changes look good. Thanks!

Left one small optional request.

packages/komencikit/src/kit.ts
bowd: Yeah I actually thought of this as well I'll make all the function take the metaTxWalletAddress as the first argument.

PR #5438, opened 2020-10-19
[Wallet] Allow skipping verification if ODIS is down

### LINKED ISSUE #5437: Allow skipping verification if ODIS is down
### Expected Behavior

If ODIS happens to be down, the onboarding shouldn't get stuck in any screen.

### Current Behavior

When going through the onboarding, the verification screen gets stuck because there is a permanent loading screen and the skip verification modal doesn't show up when you press on the header button.

Note: This is blocking the wallet e2e tests.

gnardini:
### Description

ODIS is down in Alfajores right now and when doing the onboarding the app is stuck on the verification screen because there is a permanent loading screen and the skip verification modal doesn't show up. I copied the modal so that it's shown even if the screen is loading.

Note: This is blocking the wallet e2e tests.

### Tested

Manually and added a unit test.

### Related issues

- Fixes #5437


annakaz (commented):
Is this modal always going to be shown, or just when ODIS is down? Also has the e2e Android test been updated to use this skip modal?

gnardini:
@annakaz 
> Is this modal always going to be shown, or just when ODIS is down? Also has the e2e Android test been updated to use this skip modal?

The modal is now duplicated, once for the loading screen and once for the normal screen. So the modal should show up no matter if the loading screen is showing or not. It pops up when someone taps on the Skip button on the header.

The e2e tests should have no difference, they tap on the skip button and then press on the confirm button on the modal.

jeanregisser (approved):
LGTM 👍

Lss-Ankit:
Hi @gnardini can you please provide more details to verify this task. 
And AFAIK the latest builds are Android play store internal build v1.3.0(1004294316) & Test Flight build v1.3.0(27) connected with Mainet, 
So, can you please provide a build details which connect to Alfajore network to verify this task if we need to verify this task on Alfajore network.

gnardini:
Hi @Lss-Ankit 
It is a difficult task to verify. The way I could find is to test is the following:

- Restore an account that has funds. The screen after putting in the Account Key is the verification screen. If you try to press the "Skip" button while the screen is still loading, the confirmation modal should pop up. 

The screen should look like this:

Screen Shot 2020-10-26 at 10 31 53

This is the modal that should pop up when pressing the "Skip" button:
Screen Shot 2020-10-26 at 10 33 59


The problem used to be that the popup didn't show up when loading. Now it should show up even if the screen is still loading.

Lss-Ankit:
Hi @gnardini Verified issue on latest Test flight build v1.3.0(27) and Android play store internal build v1.3.0(1004294316) and observed that user is able to tap on Skip button when loading indicator is shown on Confirm screen. 

Verified devices: 
- Google pixel Xl (8.1) , iPhone SE 2 (14.1)

PR #5441, opened 2020-10-19
Add flag to verify-deployed

nambrot:
### Description

Missed it in the last PR


packages/protocol/scripts/bash/verify-deployed.sh
```diff
@@ -20,7 +20,7 @@ RELEASE_1=""
 FORNO=""
 LOG_FILE="/dev/null"
 
-while getopts 'b:n:rf' flag; do
+while getopts 'b:n:rfl:' flag; do
```
emaG3m, line 23:
```
➜  protocol git:(master) yarn verify-deployed -n $NETWORK -b $RELEASE -l ./text.txt  -r -f
yarn run v1.22.5
$ ./scripts/bash/verify-deployed.sh -n baklava -b celo-core-contracts-v1.rc1 -l ./text.txt -r -f
./scripts/bash/verify-deployed.sh: line 30: OPTARG: unbound variable
error Command failed with exit code 1.
```
I think it needs to be 'b:n:l:rf` since `l` takes an arg?? That's the only way it works for me

ewilz (approved):

PR #5445, opened 2020-10-20
Update Docs for Datadirectory on Testnets

YazzyYaz:
Based on datadirectory issues in https://discord.com/channels/600834479145353243/601046166620078100/767968101530468382 in #validator channel and following discussions


trianglesphere:
I think that we need to include the datadir flag for the baklava proxy as well (line 240 in running a validator on baklava).

jcortejoso (approved):
lgtm

YazzyYaz:
@trianglesphere good call! just added it to the proxy line. Please let me know if it's good to go now.

mcortesi:
added another one... pls check that everywhere we use `--baklava` we also use `--datadir`

PR #5481, opened 2020-10-20
Make initialization of the proxy part of the governance proposal

### LINKED ISSUE #5479: Initializing data should be part of the proposal during contract upgrade

nambrot:
### Description

This PR removes the initialization of the proxy in `make-release` and instead makes it part of the proposal. This allows the initialization be more verifiable instead of relying on the deployer of the proxy to have initialized it correctly.

For that, we have to skip the setting of the implementation on the proxy, as otherwise any caller to `initialize` would be able to make themselves owner. That also means that we are no longer transferring ownership with respect to the implementation contract functions in proxy storage.

A follow up PR should do the [verification of the initialization transaction](https://github.com/celo-org/celo-monorepo/issues/5484) as well as [verifying that the storage root is "pure"](https://github.com/celo-org/celo-monorepo/issues/5483)

This PR also fixes a grave issue where we are passing the hashed contract name to `Registry#setAddressFor` when in reality, it takes the plain contract name and hashes on-chain.

Fixes #5479

### DIFF
--- a/packages/protocol/lib/compatibility/verify-bytecode.ts
+++ b/packages/protocol/lib/compatibility/verify-bytecode.ts
@@ -34,7 +34,8 @@ interface VerificationContext {
 // Checks if the given transaction is a repointing of the Proxy for the given
 // contract.
 const isProxyRepointTransaction = (tx: ProposalTx, contract: string) =>
-  tx.contract === `${contract}Proxy` && (tx.function === '_setImplementation' || tx.function === '_setAndInitializeImplementation')
+  tx.contract === `${contract}Proxy` &&
+  (tx.function === '_setImplementation' || tx.function === 'initialize')
 
 const isImplementationChanged = (contract: string, context: VerificationContext): boolean =>
   context.proposal.some((tx: ProposalTx) => isProxyRepointTransaction(tx, contract))
@@ -127,7 +128,10 @@ const dfsStep = async (queue: string[], visited: Set<string>, context: Verificat
   } else {
     // tslint:disable-next-line: no-console
     console.log(
-      `${isLibrary(contract, context) ? 'Library' : 'Contract'} deployed at ${implementationAddress} matches ${contract}`)
+      `${
+        isLibrary(contract, context) ? 'Library' : 'Contract'
+      } deployed at ${implementationAddress} matches ${contract}`
+    )
   }
 
   // push unvisited libraries to DFS queue
@@ -148,7 +152,7 @@ export const verifyBytecodes = async (
   proposal: ProposalTx[],
   Proxy: Truffle.Contract<ProxyInstance>,
   web3: Web3,
-  isBeforeRelease1: boolean = false,
+  isBeforeRelease1: boolean = false
 ) => {
   const queue = contracts.filter((contract) => !ignoredContracts.includes(contract))
   const visited: Set<string> = new Set(queue)
--- a/packages/protocol/lib/compatibility/verify-bytecode.ts
+++ b/packages/protocol/lib/compatibility/verify-bytecode.ts
@@ -3,7 +3,7 @@ import {
   LibraryPositions,
   linkLibraries,
   stripMetadata,
-  verifyLibraryPrefix,
+  verifyLibraryPrefix
 } from '@celo/protocol/lib/bytecode'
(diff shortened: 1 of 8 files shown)


packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -241,13 +218,33 @@ module.exports = async (callback: (error?: any) => number) => {
...
+
+            const initializeAbi = (contract as any).abi.find(
+              (abi: any) => abi.type === 'function' && abi.name === 'initialize'
             )
+            console.log(`Adding initialization of ${contractName} to proposal`)
+            if (initializeAbi) {
+              const args = initializationData[contractName]
+              const callData = web3.eth.abi.encodeFunctionCall(initializeAbi, args)
+              console.log(
+                `Initializing ${contractName} with: ${args} (encoded as callData: ${callData})`
```
asaj: nit: this message implies that state changes are happening now rather than being added to the proposal

asaj: Maybe it's fine since the previous message explicitly states otherwise

asaj (approved):
Can you add more context to the PR description, including the motivation behind this change and any follow up work needed (e.g. checking the proxy storage root)

packages/protocol/scripts/truffle/make-release.ts
nambrot: Yeah that was my intention

yorhodes (approved):

packages/protocol/lib/compatibility/verify-bytecode.ts
```diff
@@ -34,7 +34,8 @@ interface VerificationContext {
 // Checks if the given transaction is a repointing of the Proxy for the given
 // contract.
 const isProxyRepointTransaction = (tx: ProposalTx, contract: string) =>
-  tx.contract === `${contract}Proxy` && (tx.function === '_setImplementation' || tx.function === '_setAndInitializeImplementation')
+  tx.contract === `${contract}Proxy` &&
```
nambrot: I don't think this will be relevant anymore, as this function is used for finding the implementation contract, which we would now do via the registry setting

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -233,15 +238,8 @@ module.exports = async (callback: (error?: any) => number) => {
...
-                args: [contract.address, callData],
-                value: '0',
-              })
-            } else {
-              proposal.push({
-                contract: `${contractName}Proxy`,
-                function: '_setImplementation',
-                args: [contract.address],
+                function: 'initialize',
+                args: [callData],
```
nambrot: Remind me, does this work? Since we actually call the initialize function, not via a delegate call?

yorhodes: it will resolve in the fallback function to a delegate call no?

nambrot: Yeah I think you are right

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -143,43 +141,25 @@ const deployProxy = async (
...
-        contract.address,
-        initializeAbi,
-        {},
-        ...args
-      )
-    }
-  } else {
+  // Only set implementation if there is no initialize on the implementation contract,
+  // as otherwise anyone could initialize via the proxy and become the owner
+  if (!initializeAbi) {
```
m-chrzan, line 150: I would suggest never doing this during the deployment, always setting implementation in governance proposal. That way the behavior is always the same (for every registry repoint in a proposal, there is an implementation setter as well), and will make the storage of each fresh Proxy exactly the same, making verifying this fact easier.

m-chrzan (commented):
How was this tested?

packages/protocol/scripts/bash/make-release.sh
```diff
@@ -11,22 +11,25 @@ set -euo pipefail
 # -i: Path to the data needed to initialize contracts.
 # -r: Path to the contract compatibility report.
 # -d: Whether to dry-run this deploy
+# -t: Whether to dry-run this deploy
```
ewilz, line 14: an after the fact comment lol: but confusing these have the same description [EDIT: actually looks lke a copy /paste error]

PR #5489, opened 2020-10-21
Update attestation bot for ODIS

timmoreton:
### Description

Update celotool verify commands (including Attestation Bot) for interface changes made when ODIS was deployed, and support use of ODIS to retrieve pepper for verification.

New parameters:

* `--salt`: Supply a salt and don't use ODIS
* `--context`: ODIS context: `mainnet`, `alfajores`, `alfajoresstaging` (ODIS isn't yet deployed on baklava)

### Tested

Locally

### Backwards compatibility

Yes


packages/celotool/src/cmds/bots/auto-verify.ts
```diff
@@ -70,6 +73,16 @@ export const builder = (yargs: Argv) => {
       description: 'The index of the account to use',
       default: 0,
     })
+    .option('salt', {
+      type: 'string',
+      description: 'The salt to use instead of getting a pepper from ODIS',
```
aslawson, line 78: interesting -- nice workaround to not have to rely on ODIS for testing :)

packages/celotool/src/cmds/bots/auto-verify.ts
```diff
@@ -70,6 +73,16 @@ export const builder = (yargs: Argv) => {
...
+      default: '',
+    })
+    .option('context', {
```
aslawson, line 81: can we not get this from the node we are connected to? or do we want the option for those to not be in sync?

if we don't already have a network ID --> human readable name, feels like we should.

aslawson (approved):
looks good

packages/celotool/src/cmds/bots/auto-verify.ts
nambrot: Yeah would be nice to get a better description of this arg

nambrot (approved):

nambrot:
`--context: Skip ODIS and use a pre-supplied salt (ODIS isn't yet deployed on baklava)`

Seems like the wrong description?

packages/celotool/src/cmds/bots/auto-verify.ts
timmoreton: this will be run in a Helm chart which can easily pass this in as a value (as soon as it's deployed not on baklava)

PR #5501, opened 2020-10-21
Add developer tools overview to docs

critesjosh:
### Description

Adds the developer tools listed [here](https://docs.google.com/document/d/1o6xkWbZvJFTOe5ZwYVpCIbmAM4FLJk6RK-Cib81i7EE/) to the Overview page in the "Developer Guide" sections of the docs.

### Tested

Linting completes with no errors.

### DIFF
--- a/packages/docs/developer-resources/overview.md
+++ b/packages/docs/developer-resources/overview.md
@@ -3,7 +3,13 @@
 This section contains information about some of the key tools and resources that will help developers start building applications
 on Celo.
 
+## Quick Start Guides
+
+View the [Developer Quick Start page](start.md) to get started using the Celo SDKs with guided coding exercises.
+
 ## Tools
+
+### SDKs
 
 - [ContractKit](contractkit/)
   - Javascript package of Celo blockchain utilities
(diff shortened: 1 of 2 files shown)


medhakothari (commented):
This seems really busy in comparison to the current page, which has 3 links on it; do you think it's worth it to find a way to divide it out into categories to make it easier to see where to go at-a-glance? If you think this is best though, I'm all for it!

packages/docs/developer-resources/overview.md
```diff
@@ -3,12 +3,45 @@
 This section contains information about some of the key tools and resources that will help developers start building applications
 on Celo.
 
-{% page-ref page="forno/" %}
+## Tools
 
-{% page-ref page="start.md" %}
```
medhakothari, line 8: noticed this link was removed altogether, was that intentional?

critesjosh: ah good catch, thanks. ill add it to the page

critesjosh:
> This seems really busy in comparison to the current page, which has 3 links on it; do you think it's worth it to find a way to divide it out into categories to make it easier to see where to go at-a-glance? If you think this is best though, I'm all for it!

I just broke the large list into smaller sections w/ headings to make it easier to parse.

medhakothari (approved):
LGTM!

PR #5513, opened 2020-10-22
Update Attestation Service on baklava to 1.1.0

timmoreton:
### Description

* Update Attestation Service on baklava to final 1.1.0 build
* Fix bug in Att Bot helm chart param


aslawson (approved):
By pre1.1.0 you just mean a build that contains the changes expected for 1.1.0?

I know you also intend to update this to try the new 8 digit flow.  Can that test be additive to the current testing (since the services should still support the old flow until Valora flips the feature switch to change its requests).

PR #5515, opened 2020-10-22
Allows wallet users to be able to withdraw CELO from Valora using a deep link

### LINKED ISSUE #5514: Wallet users SBAT send CELO from the wallet using a deep link
### Expected Behavior

Wallet users SBAT send CELO from the wallet using a deep link, similar to the `celo://wallet/pay` that we have for cUSD. Note: Requested by a community member. 

### Current Behavior

The current functionality only allows for deep links to be used to send cUSD, not CELO.

### LINKED ISSUE #5637: Payment Deeplinks SBAT work without currencyCode param
### Expected Behavior

In previous `celo://wallet/pay` deeplinks, an `amount` could be given without a `currencyCode`, and the user's local currency would be used by default and the user would be sent to the `SendConfirmation` Screen. After #5112 , the `currencyCode` param had to also be present if `amount` was present. Otherwise the user would be sent to the `SendAmount` screen, even if an `amount` was given. 

### Current Behavior

see above

medhakothari:
### Description

Allows wallet users to be able to withdraw CELO from Valora using a deep link. Uses the same link structure as for cUSD payments, but with an added "token" param: `celo://wallet/pay?address=0x...&amount=....&token=CELO`

### Other changes

- The param `token=cUSD` is optional, but works as well. If this param is not specified, it will default to cUSD.
- Allows allows deep links to work without the `currencyCode` param (see #5637 )

### Tested

Added tests, and manually tested the following: 

- [x]  Sent cUSD with only address in the deep link 
- [x]  Sent cUSD with  address and currencyCode in the deep link 
- [x]  Sent cUSD with  address , currencyCode, and amount in the deep link
- [x]  Sent cUSD with  address and amount in the deep link 
- [x]  Sent cUSD with only amount in the deep link (shouldn't work)
- [x]  Sent cUSD with  address ,  amount , token and  in the deep link 
- [x]  Sent cUSD with  address ,  amount , currencyCode, token and  in the deep link
- [x]  Sent CELO with only address and token in the deep link (shouldn't work)
- [x]  Sent CELO with  address, token, and amount in the deep link
- [x]  Sent CELO with  address, token, and amount in the deep link
- [x] Sent an unsupported token with  address,  token, and amount in the deep link (shouldn't work)

### Related issues

- Fixes #5514 
- Fixes #5637 

### Backwards compatibility

Yes


packages/mobile/src/send/utils.ts
```diff
@@ -210,15 +210,28 @@ export function* handleSendPaymentData(
...
+        feeEstimate: new BigNumber(0),
+      })
+    } else if (data.asset === 'cUSD' || !data.asset) {
+      const transactionData: TransactionDataInput = {
+        recipient,
+        amount,
+        reason: data.comment,
+        type: TokenTransactionType.PayPrefill,
+      }
+      navigate(Screens.SendConfirmation, { transactionData, isFromScan: true })
```
jeanregisser: Depending on what restriction we add on the schema for the `asset` field, we would need to handle the case where the asset is something unsupported (e.g. `BTC` 😛).
We could display our error banner `Unsupported assets XXX` for instance.

packages/mobile/src/qrcode/schema.ts
```diff
@@ -19,6 +19,7 @@ export const UriDataType = ioType({
   currencyCode: ioUnion([ioUndefined, ioKeyof(LocalCurrencyCode)]),
   amount: ioUnion([ioUndefined, ioString]),
   comment: ioUnion([ioUndefined, ioString]),
+  asset: ioUnion([ioUndefined, ioString]),
```
jeanregisser: Looks like we should be explicit and only allow `CELO` and `cUSD` here instead of any string.

jeanregisser: nit: What about using the term `token` instead `asset`?

jeanregisser (changes requested):
Thanks! 
Can you also add new tests covering the new use case?

packages/mobile/src/send/utils.ts
medhakothari: Gotcha! Right now, if the deep links don't work, we don't provide any error message to the user -- the app simply doesnt change screens. Should I leave it for consistency, or would you like me to add error banners for the other cases too? (for example `celo://wallet/pay?amount=.1` is invalid as well, and right now nothing is shown on the user's end).

packages/mobile/src/qrcode/schema.ts
medhakothari: I was leaving it vague in case we would have cEUR, etc, in the future but can limit it too! What do you think?

medhakothari: I like it! Changed :)

packages/mobile/src/send/utils.ts
```diff
@@ -184,28 +188,44 @@ export function* handleSendPaymentData(
...
+  if (data.amount) {
+    const currency = data.currencyCode
+      ? (data.currencyCode as LocalCurrencyCode)
+      : yield select(getLocalCurrencyCode)
     const exchangeRate: string = yield call(fetchExchangeRate, currency)
     const dollarAmount = convertLocalAmountToDollars(data.amount, exchangeRate)
     if (!dollarAmount) {
       Logger.warn(TAG, '@handleSendPaymentData null amount')
       return
     }
```
isabellewei: I think this portion could be put inside the `else if (data.token === 'cUSD' || !data.token)` clause? Since these variables aren't used if `data.token === 'CELO'`

packages/mobile/src/qrcode/schema.ts
jeanregisser: Yes we can leave like this if it's better to handle the unsupported values somewhere down in the processing of the data.

packages/mobile/src/send/utils.test.ts
```diff
@@ -128,10 +133,43 @@ describe('send/utils', () => {
...
         ...data,
         e164PhoneNumber: undefined,
+        token: undefined,
       }
       await expectSaga(handlePaymentDeeplink, deeplink)
         .provide([[matchers.call.fn(handleSendPaymentData), parsed]])
         .run()
     })
+
+    describe('deeplinks for withdrawing CELO', () => {
```
jeanregisser, line 198: Nice! Could you also add the scenario where we pay with cUSD?

packages/mobile/src/send/utils.ts
jeanregisser: I see yes, let's leave it at that for now, but could you open an issue to track this? 🙏 
Can be confusing for end users to expect something and nothing happens.

jeanregisser (commented):
Almost good!

packages/mobile/src/send/utils.ts
medhakothari: Agree! Made #5688

jeanregisser (approved):
Great! 💯

Lss-Ankit:
Hi @medhak1 @jeanregisser verified task using Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found the following 
- [X] Sent cUSD with only address in the deep link - 
- [ ] Sent cUSD with address and currencyCode in the deep link -Send screen shown and currency is USD. - can you please confirm is it expected behavior ? 
- [ ] Sent cUSD with address , currencyCode, and amount in the deep link - Sending screen is shown with PHP 0.10 amount.  - can you please confirm is it expected behavior ? 
- [ ] Sent cUSD with address and amount in the deep link - Sending screen is shown with PHP 0.10 amount.  - can you please confirm is it expected behavior ? 
- [x] Sent cUSD with only amount in the deep link 
- [x] Sent cUSD with address , amount , token and in the deep link - Currency is USD on Sending screen
- [x] Sent cUSD with address , amount , currencyCode, token and in the deep link - Currency is PHP on Sending screen
- [x] Sent CELO with only address and token in the deep link
- [X] Sent CELO with address, token, and amount in the deep link 
- [x] Sent CELO with address, token, currencyCode, and amount in the deep link 
- [x] Sent an unsupported token with address, token, and amount in the deep link 

**Need clarifications:**
A) User is redirect to feed screen when user taps on Edit button from Withdraw CELO Confirmation screen 
B) Confirming... text is shown continuously on CELO transaction feed when user taps on Withdraw button if user does not have enough CELO balance. 

**For Android devices,** Nothing happen when user taps on deep links and "Open this page to valora" pop up is also not shown when user copy & paste the given deep link into chrome browser. 

**can you please confirm should we need to file a bug for the above behaviors?** 

**Verified devices:** Google Pixel XL (8.1) , Galaxy M31 (10) , iPhone XR (13.2.2) 

**Thanks!!.**

medhakothari:
Hi! For your three unchecked boxes above, confirming that they are all expected behavior. As for the clarifications:
- A: This is a bug; it should go back to the WithdrawCELO screen 
- B: It should come on the txn feed, but when the app is refreshed it should be gone. If this is not the case, I would file a bug for it. The bigger issue is that the screen should check balance before allowing someone to hit the withdraw button, so filling a bug for that would be better. 
- Android devices: This is definitely a bug. I tried it on several android emulators and they seemed to work fine, if that's not the case then would file a bug for this.

jeanregisser:
Thanks @Lss-Ankit 

All look good except these:
> Sent cUSD with address and currencyCode in the deep link -Send screen shown and currency is USD.
> Sent cUSD with address and amount in the deep link - Sending screen is shown with PHP 0.10 amount.

Can you provide the links you used?

And to be clear no deeplinks worked for you on Android?

cc @medhak1

medhakothari:
@jeanregisser These were the two links used, in order: `celo://wallet/pay?address=0xd13eee43e38f65fe3839bef6297928b43de38a02&currencyCode=PHP` and `celo://wallet/pay?address=0xd13eee43e38f65fe3839bef6297928b43de38a02&amount=.1&currencyCode=PHP
`

medhakothari:
@jeanregisser Not sure, but would #5591 fix the android issue?

Lss-Ankit:
Hi @medhak1 @jeanregisser Verified again need clarification scenarios on IOS test flight build v1.5.0 (29) & Play store internal build v1.4.0 and  I have filed below new issues: 
1. https://github.com/celo-org/celo-monorepo/issues/5804
2. https://github.com/celo-org/celo-monorepo/issues/5803
3. https://github.com/celo-org/celo-monorepo/issues/5801
4. https://github.com/celo-org/celo-monorepo/issues/5802

PR #5518, opened 2020-10-22
ContractKit contract caching doesn't work for non-singleton contracts

bowd:
### Description

The contract caching logic doesn't take into account the explicit address passed in. The test case illustrates the problem.
If I try to instantiate two contracts that aren't singleton-ish, like MultiSig or MetaTransactionWallet, the cache will save the first one and always return it. 

### Other changes

None

### Tested

Added failing spec before fixing.

### Backwards compatibility

Yes, it will just never rely on the cache if an address is provided.


gastonponti (approved):
LGTM. Maybe you could add a TODO comment with the pair (contract, address) cache that we talked about, just to don't keep the track on that

bowd:
I actually manage to fix the typing thing and make it much nicer 😄

packages/contractkit/src/web3-contract-cache.ts
```diff
@@ -147,21 +147,24 @@ export class Web3ContractCache {
...
    * Get native web3 contract wrapper
    */
-  async getContract<C extends keyof typeof ContractFactories>(contract: C, address?: string) {
+  async getContract<
+    TContractType extends keyof typeof ContractFactories,
+    TContract = ReturnType<typeof ContractFactories[TContractType]>
+  >(contract: TContractType, address?: string): Promise<TContract> {
     if (this.cacheMap[contract] == null || address !== undefined) {
       debug('Initiating contract %s', contract)
       const createFn = ProxyContracts.includes(contract)
```
bowd: I don't understand this ternary if here -- it looks a bit weird, irrespective of me removing that unneeded typecast:

```
      const createFn = ProxyContracts.includes(contract)
        ? newProxy
        : ContractFactories[contract]
        ? ContractFactories[contract]
        : newProxy
```

So if ProxyContracts.includes(contract) is true, then use "newProxy", else if ContractFactories[contract] exists then use that otherwise use newProxy?

PR #5529, opened 2020-10-22
Add check for valid transactions to verify release

### LINKED ISSUE #5510: Proposal should only include transactions that can be verified to be not malicious
Right now a proposal could contain an arbitrary transaction and verify-release would not catch it

yorhodes:
### Description

- Validates that proposal transactions provided to `verify-release` are scoped to only release functionality

### Other changes

None

### Tested

Unit tested

### Related issues

- Fixes #5510 

### Backwards compatibility

Yes


nambrot (approved):
🙏

PR #5537, opened 2020-10-23
[Wallet] Fix parallel import of SMSs on Android

### LINKED ISSUE #5209: [Wallet] SMS gets auto-imported multiple times instead of only once
### Expected Behavior

When you open Inout codes screen for the N times and SMS is received app is only importing it once

### Current Behavior

When you open Inout codes screen for the N times and SMS is received app is importing it N times. 

Current assumption: this relates to the native module we use, where we do not unsubscribe properly from receiving SMSs

i1skn:
### Description

Essentially revert https://github.com/celo-org/celo-monorepo/pull/4990 and send attestations SMSs immediately on Android again.

### Tested

TBD

### Related issues

- Fixes https://github.com/celo-org/celo-labs/issues/578
- Fixes https://github.com/celo-org/celo-monorepo/issues/5209

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -42,7 +42,7 @@
     "@celo/client": "0.0.317",
     "@celo/contractkit": "0.4.16-dev",
     "@celo/react-components": "1.0.0",
-    "@celo/react-native-sms-retriever": "git+https://github.com/celo-org/react-native-sms-retriever#b88e502",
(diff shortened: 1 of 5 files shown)


i1skn:
I'm closing this for now, cause apparently changes in `react-native-sms-retriever` do not fix this issue.

Issue #5408, opened 2020-10-16
ContractKit Wrapper Updates for Release 1


ewilz:
As new changes to the Celo Core Contracts are being deployed to all the testnets + Mainnet, the contractKit wrappers need to be updated to be able to use these changes.

- [x] New `Exchange` functions
- [x] `Attestations` transferring
- [x] `Governance getVoteRecord` new signature

jmrossy:
@emaG3m Can you add some details here? I was about to open an issue for the Exchange wrapper being out of date with @tkporter's latest changes in #4580 but I guess this covers that?

nambrot:
@jmrossy This epic should indeed cover that. There are also some `Attestations` changes that we should add to the wrappers

PR #5542, opened 2020-10-23
Update contractkit wrappers to match release 1 contracts.

ewilz:
### Description
** Will Merge when `Contracts Release 1` CGP is officially passed **
Update contractkit wrappers to match release 1 contracts. Governance wrapper was already caught up. I excluded one attestations new function because it's only valuable to call it on-chain since it is merely a require statement. If reviewing please double check the release report to make sure I didn't miss any function changes in this PR.

Changes include:
- [x] New `Exchange` functions: (`sell()`/`buy()` instead of `exchange()`)
- [x] `Attestations` transferring: (`approveTransfer()` )
- [x] `Governance getVoteRecord` new signature (already added!)

### Other changes

added expectations to tests without expectations. 

### Tested

Did not test attestations function due to lack of sufficient test setup. I can do this if it's something we'd like to improve this sprint though.

### Related issues

- https://github.com/celo-org/celo-monorepo/issues/5408

### Backwards compatibility

yes


emaG3m:
Not ready for review, adding one more big update.

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -456,6 +456,28 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
+   * @param status The approval status
+   */
+  async approveTransfer(
+    identifier: string,
+    index: number,
+    from: Address,
+    to: Address,
+    status: boolean
+  ) {
+    return toTransactionObject(
```
ewilz: Is it possible we'd like to throw errors in this kit if we can predict an on-chain revert? Seems potentially useful, but I need to know how to access the address of `msg.sender` to do that here. Is that possible?

barbaraliau: Hm. I don't know if you can access `msg.sender` in the wrappers.

barbaraliau:
@ewilz Can you link the release report?

ewilz:
@barbaraliau  https://github.com/celo-org/celo-monorepo/releases/tag/celo-contracts-v1.rc1

packages/contractkit/src/wrappers/Exchange.ts
```diff
@@ -61,6 +61,64 @@ export class ExchangeWrapper extends BaseWrapper<Exchange> {
...
+   * DEPRECATED: use function sell
+   * Exchanges sellAmount of sellToken in exchange for at least minBuyAmount of buyToken
+   * Requires the sellAmount to have been approved to the exchange
+   * @param sellAmount The amount of sellToken the user is selling to the exchange
+   * @param minBuyAmount The minimum amount of buyToken the user has to receive for this
+   * transaction to succeed
+   * @param sellGold `true` if gold is the sell token
+   * @return The amount of buyToken that was transfered
+   */
+  exchange: (
```
ewilz, line 74: just relocated this function and put DEPRECATED in doc.

packages/contractkit/src/wrappers/Attestations.ts
nambrot: I think it's ok to not do these kind of runtime checks. At the latest, gas estimation will show that

nambrot: If you are just "exposing" the method on the wrapper, you can use the higher order function `proxySend` here as well

nambrot (approved):
Just one minor comment to use proxySend

PR #5567, opened 2020-10-23
Refactor CK explorer

yorhodes:
### Description

- Introduce `BaseExplorer` to deduplicate initialization logic across `BlockExplorer` and `LogExplorer`
- Allow `BaseExplorer.abiMapping` to be updated when the ABI at an address changes (ie governance change)

### Other changes

Move `mapFromPairs` to `base/collections`

### Tested

Tested with `governance:show` that utilizes `BlockExplorer` to decode governance proposals

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


packages/contractkit/src/explorer/base.ts
```diff
@@ -23,16 +31,43 @@ export const getContractDetailsFromContract = async (
...
-  pairs.forEach(([k, v]) => {
-    map.set(k, v)
-  })
-  return map
+const getAddressMappingFromDetails = (contractDetails: ContractDetails[], abiType: AbiType) =>
+  mapFromPairs(
+    contractDetails.map((cd) => [cd.address, getContractMappingFromDetails(cd, abiType)])
+  )
+
+export class BaseExplorer {
```
barbaraliau, line 60: I like this. Do you foresee a use for `getContractDetailsFromContract`, `obtainKitContractDetails` or the other methods to be used outside of classes? It's nice to have the option to only import a method and use it vs extending a class, but if it isn't used outside of the context, then this LGTM.

yorhodes: exported for any future use cases

barbaraliau (changes requested):
LGTM. Fix merge conflicts but then g2g.

PR #5568, opened 2020-10-23
Simplify registry address mapping fetch

yorhodes:
### Description

- Use string identifiers in registry `addressForString` call
- Rename `registry.allAddresses` to `registry.addressMapping`
- Parallelize address mapping fetch

### Other changes

### Tested

Tested with `packages/cli $./bin/run network:contracts`

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


AlexBHarley (approved):
Instead of adding a second `printValueMap` function, does it make sense for `printValueMap` to accept `Record` and `Map` types, then handle each one accordingly?

No strong feelings on this. Could also do `printValueRecord` and `printValueMap` as other options

Just wanting to give you something as I see this has been open for 7 days now.

PR #5570, opened 2020-10-23
adds komenci support to ODIS

alecps:
### Description

ODIS currently calculates quota without taking into consideration the balance or tx history of the account's walletAddress. With the addition of Komenci, MetaTransactionWallet based accounts will hold their balance and initiate transactions from an externally owned walletAddress. ODIS must take this into consideration when calculating the user's quota by summing the balances and tx counts of both the account address and the wallet address. 

### Other changes

None

### Tested

I've added some e2e tests but have not run them on staging yet. Note that this PR is against another branch (with monitoring related changes). I will make sure all tests pass on that branch before merging with master. The reason I am not running these tests now is that there will be merge conflicts with the other branch and efficiencies in testing both sets of changes at once, since testing requires deploying the changes to staging. 

### Related issues

- https://app.zenhub.com/workspaces/celo-5d94ef61d9b1e40001114c02/issues/celo-org/celo-monorepo/5401

### Backwards compatibility

Yes, because existing accounts that do not use the walletAddress feature should have the same quota


packages/phone-number-privacy/signer/src/signing/query-quota.ts
```diff
@@ -112,35 +114,56 @@ async function getQueryQuota(account: string, hashedPhoneNumber?: string) {
...
+      )
+    )
+  ).then((values) => values.reduce((a, b) => a.plus(b)))
 }
 
-export async function getCeloBalance(account: string): Promise<BigNumber> {
+export async function getWalletAddress(account: string): Promise<string> {
   return retryAsyncWithBackOff(
-    async () => (await getContractKit().contracts.getGoldToken()).balanceOf(account),
+    async () => (await getContractKit().contracts.getAccounts()).getWalletAddress(account),
```
codyborn, line 166: So if `account` is an EOA, this will return undefined? And undefined is acceptable in all current quota methods (`getTransactionCount`, `getDollarBalance`, `getCeloBalance`) which makes it okay to do? Perhaps you can add a note in the comments since this would be hard to follow if someone didn't understand the context.

PR #5571, opened 2020-10-23
ContractKit->0.4.17-dev, CLI->0.0.60-dev

medhakothari:
### Description

ContractKit->0.4.17-dev
CLI->0.0.60-dev


yorhodes (approved):
do we have to publish/bump base and utils?

medhakothari:
> do we have to publish/bump base and utils?

Hm, I didn't see any changes made in those packages. Is that not the case?

PR #5572, opened 2020-10-24
[Wallet] Use CELO to pay for gas when selling CELO

### LINKED ISSUE #5563: Transaction failure when exchanging CELO for cUSD with a very low cUSD balance
### Expected Behavior

Exchange CELO for cUSD should work with a very low cUSD balance (less than what is needed to pay for the required gas).

The `feeCurrency` should be set when sending the exchange TX on https://github.com/celo-org/celo-monorepo/blob/e93100513b1081a9fd6af89293b523246f08d6c5/packages/mobile/src/exchange/saga.ts#L309

### Current Behavior

We started seeing failed exchanges with the following error: `insufficient funds for gas * price + value + gatewayFee`
See convo in https://celo-org.slack.com/archives/C018VLA3YAK/p1603411667028300

Though https://github.com/celo-org/celo-monorepo/commit/e93100513b1081a9fd6af89293b523246f08d6c5 addressed some of this, and set the `feeCurrency` for the approval TX, @asaj spotted it didn't set it for the actual exchange TX created on https://github.com/celo-org/celo-monorepo/blob/e93100513b1081a9fd6af89293b523246f08d6c5/packages/mobile/src/exchange/saga.ts#L296

This would lead to a failed transaction when exchanging CELO for cUSD, if the cUSD was very low (less than what is needed to pay for the required gas).

gnardini:
### Description

A previous PR made it so we pay for gas with CELO rather than cUSD when selling CELO. However, we only did this for the approval tx, not the actual exchange one. Fixing that here.

### Tested

On the simulator

### Related issues

- Fixes #5563

### Backwards compatibility

N/A


annakaz (approved):
Looks good to me

packages/mobile/src/exchange/saga.test.ts
```diff
@@ -38,3 +43,37 @@ describe(doFetchTobinTax, () => {
...
+            inSymbol: CURRENCY_ENUM.GOLD,
+            inValue: SELL_AMOUNT.toString(),
+            outSymbol: CURRENCY_ENUM.DOLLAR,
+            outValue: (SELL_AMOUNT / 2).toString(),
+          },
+        },
+      })
+      .run()
+  })
+})
```
jeanregisser: Great that a test was added, however it doesn't catch the bug that this PR is fixing.
There are no assertions on the parameters passed to the `sendAndMonitorTransaction` function call.

gnardini: Yeah, sorry about that, just added it

jeanregisser (approved):
Great! 💯

Lss-Ankit:
@jeanregisser  @gnardini Verified with Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found that fees/gas amount is cut from only CELO while withdrawing CELO . 
We have confirmed it using https://explorer.celo.org site. 
- User is able to withdraw CELO even though user has 0 cUSD balance.

PR #5575, opened 2020-10-24
Introduce LibraryProxies

m-chrzan:
### Description

We tried proxying libraries using our standard Proxies, but that approach does not work - when we `delegatecall` to a library's Proxy, we're still using the original contract's storage so there's no way to dynamically read an implementation's address from a central library proxy.

This PR introduces a separate proxy type specifically for linked libraries. The design is inspired by [OpenZeppelin's Dispatcher](https://blog.openzeppelin.com/proxy-libraries-in-solidity-79fbe4b970fd/), but adapted to use the Celo Registry.

LibraryProxy implements a `_delegateToRegisteredLibrary(bytes32 libraryIdentifier)` function. This contract should be inherited from, and the inheriting contract's fallback function should call `_delegateToRegisteredLibrary` with the Registry identifier of the library being proxied to.

Any contract linking proxied libraries should inherit from `UsingProxiedLibraries` and call `setLibraryRegistry` with the Registry's address. This only needs to be done once per main contract, no matter how many linked libraries it uses, since all LibraryProxies will read the Registry's address from the same storage slot of the calling contract.

Before deployment of the calling contract, it should link `LibraryProxyShim` (a workaround library that allows us to delegatecall LibraryProxy functions) and the implementation library, *both* at the corresponding LibraryProxy's deployed address.

### Tested

Unit tests. Let me know if you see any potential edge cases that should still be tested.

### Related issues

- Related to #4812 since originally the first release was supposed to include proxied libraries. 

### Backwards compatibility

Doesn't touch any core contracts yet. Utilizing this for all linked libraries will require changes in contracts using them (as described above), as well as changes to migration, release, verification scripts, etc.

### DIFF
--- a/packages/protocol/contracts/common/LibraryProxy.sol
+++ b/packages/protocol/contracts/common/LibraryProxy.sol
@@ -9,8 +9,9 @@ import "./Registry.sol";
  * for calls to a linked library pointed to by a Registry.
  * @dev This contract should be inherited from (e.g. for a library called Math,
  * create `contract MathProxy is LibraryProxy`). The inheriting contract should
- * implement a fallback function that calls `_delegateToLibrary`, passing in a
- * Registry hashed identifier referencing the implementation library.
+ * implement a fallback function that calls `_delegateToRegisteredLibrary`,
+ * passing in a Registry hashed identifier referencing the implementation
+ * library.
  */
 contract LibraryProxy {
   // Used to store the address of the registry contract.
@@ -39,7 +40,7 @@ contract LibraryProxy {
     }
   }
 
-  function _delegateToLibrary(bytes32 libraryIdentifier) internal {
+  function _delegateToRegisteredLibrary(bytes32 libraryIdentifier) internal {
     Registry registry = Registry(_getRegistry());
 
     address implementationAddress = registry.getAddressForOrDie(libraryIdentifier);
--- a/packages/protocol/contracts/common/LibraryProxy.sol
+++ b/packages/protocol/contracts/common/LibraryProxy.sol
@@ -1,0 +1,85 @@
+pragma solidity ^0.5.13;
+/* solhint-disable no-inline-assembly, no-complex-fallback, avoid-low-level-calls */
+
+import "openzeppelin-solidity/contracts/utils/Address.sol";
+import "./Registry.sol";
+
+/**
+ * @title A library proxy utilizing the Unstructured Storage pattern, allowing
+ * for calls to a linked library pointed to by a Registry.
+ * @dev This contract should be inherited from (e.g. for a library called Math,
+ * create `contract MathProxy is LibraryProxy`). The inheriting contract should
+ * implement a fallback function that calls `_delegateToLibrary`, passing in a
+ * Registry hashed identifier referencing the implementation library.
+ */
+contract LibraryProxy {
+  // Used to store the address of the registry contract.
+  // NOTE: non-standard EIP1967 value to not clash with the calling contract's
+  // underlying Proxy.
+  bytes32 private constant REGISTRY_POSITION = bytes32(
+    uint256(keccak256("eip1967.proxy.registry")) - 1
+  );
+
+  function _setRegistry(address registryAddress) public {
+    bytes32 registryPosition = REGISTRY_POSITION;
+
+    require(Address.isContract(registryAddress), "Invalid contract address");
+
+    // Store the address of the implementation contract in an explicit storage slot.
+    assembly {
+      sstore(registryPosition, registryAddress)
+    }
+  }
+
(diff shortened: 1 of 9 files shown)


packages/protocol/contracts/common/LibraryProxy.sol
```diff
@@ -0,0 +1,86 @@
...
+
+  function _delegateToRegisteredLibrary(bytes32 libraryIdentifier) internal {
+    Registry registry = Registry(_getRegistry());
+
+    address implementationAddress = registry.getAddressForOrDie(libraryIdentifier);
+
+    // Avoid checking if address is a contract or executing delegated call when
+    // implementation address is 0x0
+    require(implementationAddress != address(0), "No Implementation set");
+    require(Address.isContract(implementationAddress), "Invalid contract address");
```
prestwich, line 51: is this check necessary? it protects only against malicious registry behavior

packages/protocol/contracts/common/LibraryProxy.sol
```diff
@@ -0,0 +1,86 @@
...
+ */
+contract LibraryProxy {
+  // Used to store the address of the registry contract.
+  // NOTE: non-standard EIP1967 value to not clash with the calling contract's
+  // underlying Proxy.
+  bytes32 private constant REGISTRY_POSITION = bytes32(
+    uint256(keccak256("eip1967.proxy.registry")) - 1
+  );
+
+  function _setRegistry(address registryAddress) public {
```
prestwich, line 24: why is this public?

prestwich:
Why introduce the `LibraryProxyShim` instead of having the target contract inherit from `LibraryProxy`? This is safe because the registry slot is static. Inheritance will prevent selector collision with between the target and any public methods on the `LibraryProxy` itself

If more conceptual separation is desired, introduce a `LibraryProxyStorageAccess` contract that both the proxy and target inherit which manages access to the fixed slot storage

m-chrzan:
> Why introduce the `LibraryProxyShim` instead of having the target contract inherit from `LibraryProxy`?

Didn't think of that, that's a much better approach. The other option (mentioned on slack) which gets rid of boilerplate in the calling contract altogether is to just hardcode the Registry's address in LibraryProxy as a constant (also getting rid of a storage read per library call).

prestwich:
> The other option (mentioned on slack) which gets rid of boilerplate in the calling contract altogether is to just hardcode the Registry's address in LibraryProxy as a constant (also getting rid of a storage read per library call).

If we anticipate the registry to never change, this is a good idea. Is the registry address consistent across networks in all cases?

packages/protocol/contracts/common/LibraryProxy.sol
m-chrzan: I'm pretty sure it *can* safely be public - LibraryProxy uses the original contract's storage, we don't care about the storage at the address where a LibraryProxy ends up deployed at.

With the shim approach, would this have even been callable by the calling contract if this were `internal`? Currently, a `delegatecall` attempting to call <this function's signature> would have been issued to this contract's deployed address, would this be reachable by that calling convention? 

Either way, will make this `internal` if we go with the inheritance approach or remove altogether if we go with hardcoded Registry address.

m-chrzan:
> Is the registry address consistent across networks in all cases?

Yes, its Proxy is actually predeployed in the genesis block to the `0xce10` address. We'd only ever change it if there were some critical reason that required the Registry's entire storage to change. Throughout the contracts package we've been keeping the convention that the registry is configurable (which I guess ends up being  nice for testing), but with how stable we are expecting the Registry's address to remain, I wouldn't mind moving away from this.

packages/protocol/contracts/common/LibraryProxy.sol
m-chrzan: True. We have the same sort of check in the regular Proxy, which also only really sanity checks Governance behavior. Is this check particularly costly?

ewilz: Maybe not so costly as an individual tx, but I'd say that since it guards the gateway to ALL libraries calls, it will get called A LOT, and that will add up.

prestwich: The `extcode` family of opcodes requires state access. So it is fairly high on the cost scale

ewilz: 🌈

packages/protocol/contracts/common/LibraryProxy.sol
prestwich: I agree that it's safe to make public, but it feels out of place. I don't have a particular reason for preferring it to be non-public, other than cleanliness

PR #5587, opened 2020-10-26
remove duplicate events and reinstated original criteria for firing `verification_reveal_all_attestations_complete`

aslawson:
### Description

- remove the duplicate `verification_reveal_all_attestations_start` & `verification_reveal_all_attestations_complete`
- move `verification_reveal_all_attestations_complete` event after all attestations are completed (where it had been originally)

### Other changes

NA

### Tested

NA

### Related issues

We are over-reporting the event `verification_reveal_all_attestations_complete` which is used to indicate user completion of the verification process.  Originally, once all 3 attestations were completed, we would fire a `verification_reveal_all_attestations_complete.` 

In a recent commit, we made changes to the logic to allow a user to instantly reveal if they started the verification process and already have actionable attestation issuers.  It then would request any more issuers if needed and reveal those.  This flow allows for the user to re-request to the same issuer or get new issuers.  The `verification_reveal_all_attestations_complete` was added right after both reveal request instance rather than after **all** reveal request resulted in successful attestation reveals.

This resulted in some attempts reporting this event twice.

Additionally, we were also treating this event as the terminating event in our metabase funnel, but have now added `verification_complete`

Screen Shot 2020-10-26 at 3 34 45 PM

### Backwards compatibility

Yes


packages/mobile/src/identity/verification.ts
```diff
@@ -346,6 +343,7 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
       }
     }
 
+    ValoraAnalytics.track(VerificationEvents.verification_reveal_all_attestations_complete)
```
nategraf: Moving it all the way down here is a bit misleading because IIUC, the complete attestations request above is another place where transactions need to be submitted. Moving this line 327 might be better. I might also support the idea of creating new events, even though we won't be able to correlate them with the one being sent today (changing the meaning of the events means we should compare `verification_reveal_all_attestations_complete` events before and after this change anyway.)

nategraf: Oh, also I think where this is here, it would give the same information as the `verification_complete` event. I'd have to verify that though.

aslawson: that's a good point -- i was trying to move it to where it was before, but we should probably just add `verification_complete` to the funnel as the true terminating state

aslawson:
Added `verification_complete` to funnel

Screen Shot 2020-10-27 at 2 02 29 PM

packages/mobile/src/identity/verification.ts
```diff
@@ -324,9 +322,9 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
             phoneHashDetails,
             attestations
           )
-          ValoraAnalytics.track(VerificationEvents.verification_reveal_all_attestations_complete)
         }
       }
+      ValoraAnalytics.track(VerificationEvents.verification_reveal_all_attestations_complete)
```
i1skn: I would put inside `if (!withoutRevealing)` block, cause otherwise we would fire `verification_reveal_all_attestations_complete` even when there was no `verification_reveal_all_attestations_start` before.

aslawson: oh good call 👍

i1skn (approved):
LGTM

PR #5588, opened 2020-10-27
Increase blocks collected for key rotation e2e test

kevjue:
### Description

Increases the number of blocks to collect for the key rotation e2e test.

### Tested

Ran e2e test locally on computer and circle CI.

### Backwards compatibility

N/A


trianglesphere (approved):
lgtm

PR #5591, opened 2020-10-27
[Wallet] Fix handling deep links

### LINKED ISSUE #4946: Sending screen is not shown by tapping on the “OPEN VALORA APP” button from the “Pay With” page if the application is not running in the background.
**Frequency:** 100%
**Repro App version:**  Test Flight build Valora app build  v1.0.0 (17) , Android Valora Mainnet build v1.0.0 (16)
**Repro on device:** iPhone 7+ (13.3.1), Google Pixel Xl (8.1) 
**Test Account:** Backup Key: (757) 699-9899. 
**URL:** https://grameen.beamandgo.com/celo 
**Pre-conditions:** App is not running in background. 

**Repro Steps:**
1. Launch the URL and login with a test account. 
2. Navigate to the Celo page to proceed a payment. 
3. Tap on OPEN VALORA APP from Pay with page. 
4. Observe Sending screen is not shown 

**Current Behavior:** Sending screen is not shown
**Expected Behavior:**  Sending screen should show. 
**Investigation:** Issue also repro when connecting… banner is shown in the app when the user taps on OPEN VALORA APP button. 
**Impact:** User may get frustrated as they blocked to proceed for the payment 
**Attachment:** [SendingPageIssue.mp4](https://drive.google.com/file/d/1v4-pgB0_wDh7XDhYP8SbqJnWPA5nW4XL/view?usp=sharing)

etuleu:
### Description

Handling deep links sometimes fails, depending on whether the app if foreground/background, and the process is running or not.

For ios the issue was that the deep link was not immediately available during app init, so using the url listener in the App component fixes it.

For android it was kind of the opposite problem. It only worked if the app was not running, because if you exit the app with the back button we would remove the url listener. When the activity was started again from the intent appInit was not getting called since the app was already running. So, checking for the url in the componentDidMount method fixes it.

### Other changes

Also wrote an e2e test to expose the issues and to make sure they don't break in the future.

### Tested

Used adb to send an intent with the deep link to the app in various states.

### Related issues

- Fixes #4946 

### Backwards compatibility


packages/mobile/android/app/src/main/java/org/celo/mobile/MainActivity.java
```diff
@@ -56,6 +57,14 @@ public void onPause() {
...
   }
 
+  @Override
+  public void onNewIntent(Intent intent) {
+    Boolean firebaseEnabled = Boolean.parseBoolean(BuildConfig.FIREBASE_ENABLED);
+    if (firebaseEnabled) {
+      super.onNewIntent(intent);
+    }
+  }
+
```
jeanregisser: Why is this needed?

packages/mobile/package.json
```diff
@@ -176,6 +176,7 @@
     "eslint": "^6.8.0",
     "jest-fetch-mock": "^2.1.2",
     "metro-react-native-babel-preset": "^0.58.0",
+    "mocha": "8.2.0",
```
jeanregisser: Detox needs this now?

packages/mobile/src/app/App.tsx
```diff
@@ -56,6 +56,11 @@ export class App extends React.Component<Props> {
 
     Linking.addEventListener('url', this.handleOpenURL)
 
+    const url: string | null = await Linking.getInitialURL()
```
jeanregisser:
```suggestion
    const url = await Linking.getInitialURL()
```

packages/mobile/android/build.gradle
```diff
@@ -38,7 +38,7 @@ allprojects {
             url("$rootDir/../../../node_modules/jsc-android/dist")
         }
         maven {
-            url "$rootDir/../../../node_modules/detox/Detox-android"
+            url "$rootDir/../node_modules/detox/Detox-android"
```
jeanregisser: Surprised by this path change, why does it install in `packages/mobile/node_modules` now?

jeanregisser (commented):
Looks mostly good!
e2e tests are both failing though.

packages/mobile/android/build.gradle
etuleu: I am actually having some trouble with detox because their device api seems to not work prioperly. Sometimes the app does not relaunch and so I tried upgrading detox to see if that fixes it but now I am getting other build problems. So, to not waste more time on this for now, I will revert the package upgrades, disable the e2e tests that don't pass and come back to it with a later PR. It would be really good to have e2e tests for this so it doesn't break in the future but for now I am going faster to fix more P1s.

packages/mobile/android/app/src/main/java/org/celo/mobile/MainActivity.java
etuleu: So, with FIREBASE_DISABLED this causes a crash. The better fix would be to have a test firebase running in CI. I would love to do that but I am not sure how long it would take so in the interest of time for this milestone this was necessary.

packages/mobile/package.json
etuleu: I will revert his along with the detox revert.

etuleu: It looks like I didn't commit all the files related to the detox upgrade anyway. Sorry about the mess! I knew I should not have tried to upgrade detox in the same PR. I regret doing it since now my build is messed up too. Starting with a clean repo and no upgrades to packages.

etuleu:
> Looks mostly good!
> e2e tests are both failing though.

Fixed the core e2e test, which is to start the app with the url. The other cases I skipped for now because on android they require having firebase enabled, and we will implement that later.

packages/mobile/android/app/src/main/java/org/celo/mobile/MainActivity.java
jeanregisser: oh ok, is this a crash on the native side or JS?

etuleu: The crash is on the native side. It looks like the react activity tries to call firebase dynamic links, which in turn complains about firebase app not being initialized.

packages/mobile/package.json
jeanregisser: Thanks, could you add a comment about this?

packages/mobile/android/app/src/main/java/org/celo/mobile/MainActivity.java
jeanregisser: Thanks, could you add a comment about this here?

jeanregisser (approved):
Great! 👍 
One last small comment, otherwise good to merge.

packages/mobile/package.json
etuleu: Oh this got reverted so no need to worry about this package.

Lss-Ankit:
@jeanregisser @etuleu  to verify this task, I need to sync with @evan for OTP to login into BeanAndGo web app.
will sync it and update the status here

Lss-Ankit:
@etuleu @jeanregisser Verified task using Test flight build v1.5.1 (30) and Play store internal build 1.5.1 (1004294319) and found the following: 

**Used devices:** iPhone SE 2 (14.1) , Google Pixel XL (8.1) 

**For IOS:**

- [x] App is Fore Ground: Verify Sending screen is shown when user taps on OPEN VALORA APP from B&G site. 
- [x] App is running in Back Ground: Verify Sending screen is shown when user taps on OPEN VALORA APP from B&G site. 
- [x] App is not running in Back Ground: Verify Sending screen is shown when user taps on OPEN VALORA APP from B&G site. 

**For Android:**
- [x] App is Fore Ground: Verify Sending screen is shown when user taps on OPEN VALORA APP from B&G site. 
- [x] App is running in Back Ground: Verify Sending screen is shown when user taps on OPEN VALORA APP from B&G site. 
- [x] App is not running in Back Ground: Verify Sending screen is shown when user taps on OPEN VALORA APP from B&G site. 

**Need Clarification:**
 
1.  "Could not calculate fee" message is shown on Sending screen after tapping on  OPEN VALORA APP from B&G site if user does not have enough balance. 
2. Also, Feed screen is shown for while before showing Sending screen when user taps on OPEN VALORA APP from B&G site if application is not running on back ground. 

Can you please let us know is it expected or we need to file a bug for the same. 
[Attachment: ](https://drive.google.com/file/d/11-U5BIHe4y5JFJotPduLFCamkO3BBdea/view?usp=sharing)

jeanregisser:
Thanks @Lss-Ankit, 
Could you open 2 issues for what you fund?
- the "Could not calculate fee" message could be confusing and should be changed to "Insufficient funds" or something similar.
- the delay in showing the Send Screen is not ideal. It's likely caused by the async request done to do the currency conversion.

Lss-Ankit:
Sure @jeanregisser i have raised ticket for both issues. below is the ticket for the same: 

1. https://github.com/celo-org/celo-monorepo/issues/5899
2. https://github.com/celo-org/celo-monorepo/issues/5898

PR #5599, opened 2020-10-27
[cli] Require --force with account:claim-attestation-service-url for non-TLS urls

timmoreton:
### Description

In the future we will be requiring TLS for Attestation Service URLs. This adds a check to the CLI command to require `--force` to add a URL that doesn't begin `https://`.
 
### Tested

Locally

### Backwards compatibility

In the unlikely event that users have scripted this, and do not have TLS deployed, scripts will need modification.

### DIFF
--- a/packages/cli/src/commands/account/claim-attestation-service-url.ts
+++ b/packages/cli/src/commands/account/claim-attestation-service-url.ts
@@ -1,4 +1,5 @@
 import { createAttestationServiceURLClaim } from '@celo/contractkit/lib/identity/claims/attestation-service-url'
+import { flags } from '@oclif/command'
 import { Flags } from '../../utils/command'
 import { ClaimCommand } from '../../utils/identity'
 export default class ClaimAttestationServiceUrl extends ClaimCommand {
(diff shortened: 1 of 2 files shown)


nambrot (approved):

medhakothari:
Hi @timmoreton! Could you update the [changelog](https://github.com/celo-org/celo-monorepo/blob/master/packages/cli/CHANGELOG.md) in this PR as well?

packages/cli/src/commands/account/claim-attestation-service-url.ts
```diff
@@ -8,17 +9,24 @@ export default class ClaimAttestationServiceUrl extends ClaimCommand {
     ...ClaimCommand.flags,
     url: Flags.url({
       required: true,
-      description: 'The url you want to claim',
+      description: 'The URL you want to claim. Should begin http://',
```
aslawson, line 12: you mean `https://`?

aslawson, line 12:
```suggestion
      description: 'The URL you want to claim. Should begin https://',
```

packages/docs/command-line-interface/account.md
```diff
@@ -147,12 +149,12 @@ OPTIONS
                                                      index addresses for local signing. Example --ledgerCustomAddresses
                                                      "[4,99]"
 
-  --url=https://www.celo.org                         (required) The url you want to claim
+  --url=https://www.celo.org                         (required) The URL you want to claim. Should begin http://
```
aslawson, line 152:
```suggestion
  --url=https://www.celo.org                         (required) The URL you want to claim. Should begin https://
```

PR #5600, opened 2020-10-27
Attestation Bot: Add check for expired attestation

timmoreton:
### Description

Attestation Bot was missing a check to see if an existing attestation had expired. Added it.

### Tested

On baklava. Previously observed:

`{"message":"Error: Gas estimation failed: The attestation request has expired\n    at ContractKit.<anonymous> (/Users/tim/Celo/celo-monorepo/packages/contractkit/src/kit.ts:343:15)\n`

Now successful.

### Backwards compatibility

Yes


nambrot (approved):

packages/celotool/src/lib/attestation.ts
```diff
@@ -45,7 +45,7 @@ export async function requestMoreAttestations(
   txParams: CeloTransactionParams = {}
 ) {
   const unselectedRequest = await attestations.getUnselectedRequest(phoneNumber, account)
-  if (unselectedRequest.blockNumber === 0) {
+  if (unselectedRequest.blockNumber === 0 || (await attestations.isAttestationExpired(unselectedRequest.blockNumber))) {
```
aslawson, line 48: does `getActionableAttestations` do what you want?

aslawson (approved):

packages/celotool/src/lib/attestation.ts
timmoreton: looks like it would, but I think keeping the flow in the bot the same as in Valora probably makes sense (Valora does the same right now, but had the expiration check)

PR #5608, opened 2020-10-27
AWS Oracle Infrastructure

### LINKED ISSUE #5382: Finish AWS oracle infrastructure
* Modify celotool to create appropriate AWS identities that have singing permission
* Assign AWS identity to individual pods
* Deploy on Baklava & Mainnet

tkporter:
### Description

This PR adds deployment tooling for deploying AWS oracles.

#### AWS HSM Oracle Deployment

[kube2iam](https://github.com/jtblin/kube2iam) is used for assigning AWS roles to specific pods. A pod can assume a role by setting the annotation `iam.amazonaws.com/role` to the ARN of the desired role. Celotool will allow grant the `AssumeRole` permission for the relevant roles to the underlying K8s nodes (ie VMs). Celotool also handles the creation of a role for each oracle, and the creation of a policy the role uses that allows an oracle role to use its HSM.

We still face the same issue as aad-pod-identity, where we have the oracles as a statefulset but each pod must have different pod metadata. The same approach is used that involves creating service accounts whose credentials are used in an init container to modify the metadata.

#### Code Structure

Previously, almost all deployment logic lived in `lib/oracles.ts`. Now, the deployment logic lives in the dir `lib/k8s-oracle`. There's an abstract base class, `BaseOracleDeployer`, that provides installation/upgrade/removal functions and specifies generic helm params. AKS HSM oracle deployment logic lives in `AksHsmOracleDeployer`, AWS HSM oracle deployment logic lives in `AwsHsmOracleDeployer`, and non-HSM (ie private keys from a string to perform in-memory signing) oracle deployment logic lives in `PrivateKeyOracleDeployer`. Both HSM deployers contain all relevant cloud identity creation & permission granting, and they both extend `RbacOracleDeployer`, which is another abstract subclass of  `BaseOracleDeployer` that will also create the necessary k8s service accounts that are used for changing oracle pod metadata in an init container.

### Other changes

* Instead of naming things with AWS or AKS in all caps, they're now stylized in CamelCase. This is mainly because AWSHSM is really ugly
* There are some .env changes that I do not plan on committing, they're just there for some testing right now

### Tested

A bunch of deployments

### Related issues

- Fixes #5382 

### Backwards compatibility

Yes


.env.alfajores
```diff
@@ -27,6 +27,19 @@ CELOSTATS_TRUSTED_ADDRESSES=""
 CELOSTATS_BANNED_ADDRESSES=""
 CELOSTATS_RESERVED_ADDRESSES=""
 
+CONTEXTS=azure-eastus
```
tkporter: Still playing around with some .env file stuff, I will clean this up soon

packages/celotool/src/lib/k8s-oracle/base.ts
```diff
@@ -0,0 +1,110 @@
...
+      ? getFornoUrl(this.celoEnv)
+      : getFullNodeHttpRpcInternalUrl(this.celoEnv)
+    const wsRpcProviderUrl = this.deploymentConfig.useForno
+      ? getFornoWebSocketUrl(this.celoEnv)
+      : getFullNodeWebSocketRpcInternalUrl(this.celoEnv)
+    return [
+      `--set environment.name=${this.celoEnv}`,
+      `--set image.repository=${fetchEnv(envVar.ORACLE_DOCKER_IMAGE_REPOSITORY)}`,
+      `--set image.tag=${fetchEnv(envVar.ORACLE_DOCKER_IMAGE_TAG)}`,
+      // TODO change this & the corresponding oracle env var
```
tkporter: Still doing this, waiting on my change in the oracle client to make the init retry logic not just for azure hsms. Will clean up soon

.env.oracledev
```diff
@@ -1,5 +1,7 @@
-ORACLE_DOCKER_IMAGE_REPOSITORY="celotestnet.azurecr.io/testnet/oracle"
-ORACLE_DOCKER_IMAGE_TAG="c7215fcbeccc0c61b306fbd0503a67bd0cf509de"
+ORACLE_DOCKER_IMAGE_REPOSITORY="tkporter/oracle"
+ORACLE_DOCKER_IMAGE_TAG="aws-test-ebbc438f36169b59a3397419d4b4dd6275ac09e6"
+
+# tkporter/oracle:aws-test-ebbc438f36169b59a3397419d4b4dd6275ac09e6
```
jcortejoso: delete?

packages/helm-charts/oracle/templates/statefulset.yaml
```diff
@@ -37,24 +37,34 @@ spec:
...
+      - name: set-metadata
         image: {{ .Values.kubectl.image.repository }}:{{ .Values.kubectl.image.tag }}
         command:
         - /bin/bash
         - -c
         args:
         - |
           RID=${POD_NAME##*-}
-          TOKEN_ENV_VAR_NAME="TOKEN_$RID"
+          TOKEN_ENV_VAR_NAME="TOKEN_$RID"\
```
jcortejoso: Is the `\` expected?

packages/celotool/src/lib/k8s-oracle/rbac.ts
```diff
@@ -0,0 +1,69 @@
+import { installGenericHelmChart, removeGenericHelmChart, upgradeGenericHelmChart } from 'src/lib/helm_deploy'
+import { execCmdWithExitOnFailure } from '../cmd-utils'
+import { BaseOracleDeployer } from './base'
+
+// Oracle RBAC------
+// We need the oracle pods to be able to change their label to accommodate
+// limitations in aad-pod-identity & statefulsets (see https://github.com/Azure/aad-pod-identity/issues/237#issuecomment-611672987)
+// To do this, we use an auth token that we get using the resources in the `oracle-rbac` chart
```
jcortejoso, line 8: If I understood correctly, the rbac chart is decoupled from oracle chart to create 1 service account each oracle replica, thus being able to pass the SA identities/tokens from the rbac chart as values to the oracle chart. 
And in the oracle chart all the pods from the oracle statefulset have all the tokens as env variables, but each pod uses a different token based on their $RID. Is that right?

jcortejoso (approved):
Awesome work combining the logic for Azure and AWS, and I think the result is cool and pretty similar from a logic point to both providers.

packages/celotool/src/lib/k8s-oracle/rbac.ts
tkporter: Yup! That is correct

PR #5614, opened 2020-10-28
Add generateDeterministicInviteCode

nambrot:
Create`generateDeterministicInviteCode` to allow the creation of a deterministic invite code with knowledge of the ODIS pepper


tarikbellamine (approved):
Thank you for doing this!

packages/utils/src/account.ts
```diff
@@ -73,6 +74,17 @@ export async function generateKeys(
   return generateKeysFromSeed(seed, changeIndex, addressIndex, derivationPath)
 }
 
+export function generateDeterministicInviteCode(
+  pepper: string,
```
tarikbellamine, line 77: Nit: can we rename to `recipientPepper` to eliminate confusion?

PR #5616, opened 2020-10-28
[Attestation Service] MessageBird provider

timmoreton:
### Description

* Add new `messagebird` provider for [MessageBird](https://messagebird.com/en/)

### Other changes

* Providers can now specify whether their callback for delivery status is `GET` or `POST`
* Minor typos in the example config

### Tested

Locally. Sent messages via MessageBird, confirmed delivery status receipts for both success and failure conditions.

### Backwards compatibility

Yes


packages/attestation-service/src/sms/index.ts
```diff
@@ -148,7 +157,7 @@ export function configuredSmsProviders() {
 }
 
 export function smsProvidersWithDeliveryStatus() {
-  return smsProviders.filter((provider) => provider.supportsDeliveryStatus())
+  return smsProviders.filter((provider) => provider.deliveryStatusMethod())
```
aslawson, line 160: what is the false state to this filter?  will an empty string result to false?

packages/attestation-service/src/sms/messagebird.ts
```diff
@@ -0,0 +1,88 @@
...
+  deliveryStatusURL: string | undefined
+
+  constructor(apiKey: string, unsupportedRegionCodes: string[]) {
+    super()
+    this.messagebird = initMB(apiKey)
+    this.unsupportedRegionCodes = unsupportedRegionCodes
+  }
+
+  async receiveDeliveryStatusReport(req: express.Request, logger: Logger) {
+    if (typeof req.query.reference === 'string' && typeof req.query.status === 'string') {
```
aslawson, line 31: didn't realize you could parse a GET query fields this way with typescript. cool :)

packages/attestation-service/src/sms/messagebird.ts
```diff
@@ -0,0 +1,88 @@
...
+      case 'sent':
+        return AttestationStatus.Upstream
+      case 'buffered':
+      case 'scheduled':
+        return AttestationStatus.Queued
+    }
+    return AttestationStatus.Other
+  }
+
+  deliveryStatusMethod = () => 'GET'
```
aslawson, line 58: weird that its a GET used to pass information...

aslawson (approved):

packages/attestation-service/src/sms/messagebird.ts
timmoreton: yep, weird

PR #5153, opened 2020-09-23
Base->0.0.4-dev, Utils->0.1.21-dev, & ContractKit->0.4.15-dev

medhakothari:
### Description

Base->0.0.4-dev
Utils->0.1.21-dev
ContractKit->0.4.15-dev

Related: Celocli --> #5125

Issue #5142, opened 2020-09-22
Handle transactiontimeouts due to user backgrounding the app 


aslawson:
During the request phase -- we make 3 onchain transaction with 45s timeouts.  About half of reported transactionTimeouts are actually due to users backgrounding the application during this time.  We should consider 

1. Investigate whether we can freeze the timers when the user is backgrounded or still handle the asynchronous results from the onchain request when they return to the app.
2. Report a different error that indicates the user backgrounded the app during this stage instead of transactionTimeout
3. Tell  user to not leave during the 2min timer

timmoreton:
I wonder if when we receive a timeout, we first check the receipt to see whether the tx was actually sent. In almost all cases, it will be.

aslawson:
> I wonder if when we receive a timeout, we first check the receipt to see whether the tx was actually sent. In almost all cases, it will be.

This does occur -- the timer is a race against an asynchronous request that is satisfied when it received the receipt, hash and confirmation

nategraf:
In #5620 I am adding a "grace period" to the transaction timeout which helps avoid triggering a timeout in cases where the app becomes suspended while tyhe transaction is pending with the network. In those cases, when the app returns to the foreground, and the initial timeout fires, the flow will enter the "grace period" which gives it time to reconnect to the network (if it lost connection) and retrieve transaction receipts. I've tested this on an iOS simulator and it seems to address the main failure mode we are looking at here.

I'm not confident we should call this issue resolved yet because if the app becomes suspended during gas estimation, or more generally before it is sent to the network, this will still fail. As a solution to this, and other transaction sending issues, it may be useful to add a transaction pool to the send mechanism that allows for retries of each stage of the transaction send flow and provides a flow of information to the transaction feed as well. (This would be a higher-level compliment to the Geth transaction pool, which is concerned with making sure any transaction handed off to it reaches the network)

etuleu:
Another suggestion from @jeanregisser, which I think would work well in addition to your fix is to ask iOS to give us 30s extra background time. That would make the experience seem "faster" because we can make progress on the transactions even if the app is backgrounded.

nategraf:
I didn't have context on this, but this article is helpful: https://docs.microsoft.com/en-us/xamarin/ios/app-fundamentals/backgrounding/ios-backgrounding-techniques/ios-backgrounding-with-tasks#background_tasks_in_iOS_7

aslawson:
[...]

PR #5620, opened 2020-10-28
Mitigate transaction timeout issues due to backgrounding and slow networks

### LINKED ISSUE #5353: [Valora] Extend the transactionTimeout Time
Some transactionTimeouts can be due to gas estimation latency or node connection issues.  In investigation, increasing this timeout to 60s should cover most cases.
https://celo-org.slack.com/archives/C017SK5RBFS/p1600110565219600

Band-aid solution

nategraf:
### Description

One of the most common errors we see in the verification flow at the moment is transaction timeout, which is added by the Valora app and currently specified to 45 seconds.

We have observed a number of cases where a transaction timeout will occur, and then the transaction will continue and be mined successfully not too long after. Looking at a number of these events, raising the timeout period would appear to increase the success rate of transactions, provided the user is willing to wait. (In cases where they do not want to wait, they may cancel the action they are taking and the transaction will not be monitored further.)

In this PR, the timeout is increased to 90s, which I have observed will handle most of the cases where the timeout triggers on transaction that eventually end up being mined. This should be a useful measure to increase transaction reliability until we can revisit the sending flow to incorporate additional improvements.

### Other changes

* Add some comments to the verifications flow.
* Added a grace period to after initial transaction timeout to mitigate #5142.
* Change view call for balance fetch to select from redux state to avoid expensive view call.
* Log full transaction details in the `GethNativeBridgeWallet` to improve debugging in Sentry.

### Tested

Tested on iOS simulator to confirm that without the changes, transaction report as timeout after being backgrounded for more than 90s. With the grace period, I observed that the timeout did not occur. On my full-speed WiFi, it got the transaction confirmation within 1 second of returning from suspension. On simulated 3G, it confirmed within 3 seconds of returning to the background.

### Related issues

- Fixed #5353
- Mitigates #5142

### Backwards compatibility

100%


gnardini (approved):

packages/mobile/src/transactions/send.ts
```diff
@@ -109,9 +115,7 @@ export function* sendTransactionPromises(
     `Going to send a transaction with id ${context.id}`
   )
 
-  const stableToken = yield getTokenContract(CURRENCY_ENUM.DOLLAR)
-  const stableTokenBalance = yield call([stableToken, stableToken.balanceOf], account)
-
+  const stableTokenBalance = new BigNumber(yield select(stableTokenBalanceSelector))
```
aslawson: hmm i wonder if there is any reason we were pulling this from onchain rather than the cached state before. are there any transactions that we might not have the this information in the state yet?  (I may be missing something in the context of how/when the state gets set)

packages/mobile/src/transactions/send.ts
```diff
@@ -189,21 +194,44 @@ export function* wrapSendTransactionWithRetry(
 ) {
   for (let i = 1; i <= TX_NUM_TRIES; i++) {
     try {
-      const { result, timeout, cancel } = yield race({
-        result: call(sendTxMethod),
-        timeout: delay(TX_TIMEOUT * i),
+      const task = yield fork(sendTxMethod)
+      let { result, timeout, cancelled } = yield race({
+        result: join(task),
```
aslawson, line 201: iiuc, the `join` is requiring that the task waits for the previous `sendTxMethod ` attempt to complete before another attempt? 

but seems the main benefit in the implementation is passing in the same task to this race condition and the one below with the grace period. That way, if its already completed, the second race condition will resolve almost immediately.

packages/mobile/src/transactions/send.ts
```diff
@@ -189,21 +194,44 @@ export function* wrapSendTransactionWithRetry(
...
-          cancel: take(cancelAction),
+          cancelled: take(cancelAction),
         }),
       })
 
+      // In some conditions (e.g. app backgrounding) the app may become suspended, preventing the
+      // send task from making progress. If this occurs for long enough that the timeout elapses, a
+      // timeout may be triggered even if the transaction send is complete. A second race is
+      // triggered here to handle these cases by giving the send task a grace period to return a
+      // result after the initial timeout fires.
```
aslawson, line 212: Very nice! Do you think that the time for network reconnection is the main cause of #5153 ?  How long could this take?

This grace period to check the completion of the task seems it would solve most of #5153 . 

Also have you tried this with a local build of the app? I'm wondering if the VerificationCountdown [TOTAL_TIME](https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/verify/VerificationCountdown.tsx#L9) may still expire during backgrounding and whether we'd still need to figure out how to pause this.

aslawson: cc @etuleu

nambrot:
When you say: 

> and then the transaction will continue and be mined successfully not too long after.

Is this because a confirmation by the light client can not be received by the app while it is suspended? But then when the app foregrounds, it will receive such confirmation?

aslawson (commented):
looks really good.  and nice work handling some of the race condition logic.

nategraf:
> Is this because a confirmation by the light client can not be received by the app while it is suspended? But then when the app foregrounds, it will receive such confirmation?

So in this section I am actually only referring to the timeout length as a whole. In theory, timeouts are useful for when you are pretty sure the operation you are doing is not going to succeed. (e.g. "If this transaction doesn't get mined within 90s, it's never going to be mined") Taking a quick look at the data we have, about 1.5% of transactions take longer than 45s and about 0.4% take long than 90s. (This does not account for the fact that transactions on slower networks are _much_ more likely to take more than 45s. The 60th percentile on Airtel UG is currently around 40s e2e)
https://metabase.celo-networks-dev.org/question/240/notebook

packages/mobile/src/transactions/send.ts
nategraf: It's possible, but we should really avoid this view call here if possible. In high latency environments view calls can take quite a while to complete. I'll look at writing a function to fetch this value if it's not available or expired, and use the cached value otherwise. (If you'd prefer we not tough this, I'm fine with leaving it for now)

packages/mobile/src/transactions/send.ts
nategraf: The main point here is indeed to ensure that the send is not cancelled automatically on timeout. If the Task completes, the `join` allows the `race` to exist immediately with the result. Otherwise the send will not be interrupted until the cancel statement below is reached.

nambrot:
Gotcha, thanks for the context. Great to see we are getting better insights into this, really wonder what's the cause of such huge confirmation times

packages/mobile/src/transactions/send.ts
nategraf: Even if the reconnection / receipt fetching were "instant" the timeout would still trigger when returning from suspension if the app checks the (expired) timer before the result of the send task. So the reconnection time is the reason that the grace period is 5 s rather than e.g. 100 ms, but the grace period would still be useful here in either case.

I agree this will solve the main issue we are seeing now, but I would assess that we still have a lot of work to do to handle suspension gracefully. Most notably, suspension at the wrong time can make any given network call fail, which in our current scheme means failing the whole transaction send even if the failure occurred after the tx was sent to the network. Jean reported a case of this on Slack this morning.

nategraf: It looks like `TOTAL_TIME` will indeed tick down while the user is away. It looks like `TOTAL_TIME` is only used to set expectations, will not cancel the process if it runs out, so I guess that's ok?

packages/mobile/src/transactions/send.ts
aslawson: i like the check the cache-first approach, seems it could only improve the flow. worst case, its not in the case and we fall to the same logic that already exists 👍

packages/mobile/src/transactions/send.ts
aslawson: oh gotcha -- i now see the cancel(task)

packages/mobile/src/transactions/send.ts
aslawson: interesting -- ya I think its ok from the functionality perspective. from the user experience, the typically user would return to the "enter message" screen while the logic to make all the transactions would resume progress, so i think pausing it may be more of an optimization.  there could be a decent delay if they background before making the `waitForSelectingIssuers`.

i think then we can close the ticket #5153 and open two more
1. around investigating general transaction suspension
2. optimization for the pausing the countdown timer (would need to sync with product on the priority of this)

aslawson (approved):
lgtm

packages/mobile/src/transactions/send.ts
nategraf: I spent about 30 minutes trying to figure out how to do this in a way that will actually reduce the number of calls in the common case, and could not think of a clean solution. I've reverted this change.

Lss-Ankit:
Hi @jeanregisser @nategraf Verified task using Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found the following: 
- Verification failed pop up is shown when app is running on background with slow networks. 
We have tried 3 times but for two times we got verification failed pop up and not able to get the attestation code if app is running on background. 

Please find the logs for the same. 
[celo_logs.txt](https://github.com/celo-org/celo-monorepo/files/5510757/celo_logs.txt)
Also, please let us know if we need to verify any thing else.

PR #5621, opened 2020-10-29
ODIS Client Update

### LINKED ISSUE #5627: ODIS ContractKit Client Update
- Separate phone number blinding from API request
- Support custom callback for auth header signing (allows us to sign the header in a separate service).

codyborn:
### Description

Some updates for Komenci:
1. Custom auth header signer. This allows us to separate out the signing from the request management so that the onboarding service can manage the request while the relayers simply perform the auth header signing.
2. Separate public method for passing in pre-blinded phone number. This allows Valora to pass the blinded phone number and for Komenci to use this new public method instead which expects the pre-blinded phone number.

### Tested

New tests included

### Backwards compatibility

Yes, no change to previous method interface.


jmrossy (commented):
I don't have a lot of context into Komenci but these changes all lgtm :)

packages/contractkit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celo/contractkit",
-  "version": "0.4.18-dev",
+  "version": "0.4.19-dev",
```
medhak1, line 3: I think it would make more sense to do this in a different PR, where you also change all the files that consume the dev version of the contractkit kit

medhak1:
Hey Cody! Could you also update the [changelog](https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/CHANGELOG.md) in this PR?

packages/contractkit/package.json
medhakothari: I think it would make more sense to do this in a different PR, where you also change all the files that consume the dev version of the contractkit kit

medhakothari (changes requested):
Also don't have Komenci context, but the logic LGTM! Thanks for adding tests
Could you also update the [changelog](https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/CHANGELOG.md) in this PR?

packages/contractkit/package.json
codyborn: Good point, thanks!

medhakothari (approved):
LGTM 🚀  Would also be good to have someone working on Komenci look at it too

packages/contractkit/src/identity/odis/phone-number-identifier.test.ts
```diff
@@ -43,18 +59,74 @@ describe(getPhoneNumberIdentifier, () => {
...
+        authSigner,
+        serviceContext,
+        base64BlindedMessage
+      )
+
+      const base64UnblindedSig = await blsBlindingClient.unblindAndVerifyMessage(base64BlindSig)
+      const sigBuf = Buffer.from(base64UnblindedSig, 'base64')
+
+      const pepper = getPepperFromThresholdSignature(sigBuf)
+      const phoneHash = getPhoneHash(sha3, mockE164Number, pepper)
```
bowd: Feels like this part gets a bit too "in the weeds" and it lacks a bit in ergonomics. I'm gonna leave the comment here but I read through the whole PR and understood the changes a bit. I would suggest a slightly different approach. Instead of extracting `performGetPhoneNumberIdentifier` from `getPhoneNumberIdentifier` maybe we should expose these functions:

```typescript
function getPhoneNumberIdentifier(...) {
  const base64BlindedMessage = getBlindedPhoneNumber(...);
  const base64BlindSig = getBlindedPhoneNumberSignature(...);
  return getPhoneNumberIdentifierFromSignature(...); 
}

function getBlindedPhoneNumber(...) {
  // blind the phone number
}

function getBlindedPhoneNumberSignature(...) {
  // construct body and query ODIS
}

function getPhoneNumberIdentifierFromSignature(...) {
  // construct pepper and identifier from signature  
}
``` 

This way I think we get a cleaner API that we can then compose to fit our needs. Basically we're splitting the function into 3 components `blind phone`->`get signature`->`construct identifier` because we want to `blind phone` in Valora, then `get signature` in Komenci and then give back the result to Valora to `construct identifier`.

While this split is a bit more granular it will keep the logic from bleeding out of the contractkit ODIS utils because in the current split it's basically just the `getBlindedPhoneNumberSignature` (named `performGetPhoneNumberIdentifier` - which at the minimum would need to change because it's actually getting the signature only) step that's extracted and callable by Komenci which leaves the other two to be implemented in Valora somehow and duplicate the logic that's already in `getPhoneNumberIdentifier` which might prove problematic in the future.

bowd (changes requested):

packages/contractkit/src/identity/odis/phone-number-identifier.test.ts
codyborn: Good call out. I was debating this as well and decided against it because the un/blinding logic is concise; but you're right, it's better to handle this in one place to minimize the impact of changes going fwd.

bowd (approved):

Issue #5611, opened 2020-10-28
“Purchase Look up” & “Skip” buttons are shown misaligned on the Look up Quota screen. 


Lss-Ankit:
**Frequency:** 100%
**Repro App version:** Test Flight Build v1.3.0 (27), Android play store internal build v1.3.0(1004294316)
**Repro on devices:** iPhone SE 2 (14.1) , iPhone XS Max (13.5.10, Samsung Galaxy A5 (7.0) 

Account Key: (757) 699-9899/Backup Key
Pre-condition: Unverified User should have at least $0.01 balance. 

**Repro Steps:** 
1. launch the app and Sign up with an unverified user. 
2. Tap on the Send button. 
3. Search 3 verified users from the send screen. 
4. Make sure the user receives the Look up Quota screen. 
5. Observed “Purchase Look up” & skip 

**Current Behavior:** “Purchase Look up” & “Skip” buttons are shown misaligned on the Look up Quota screen. 
**Expected Behavior:** “Purchase Look up” & “Skip” buttons should show  aligned on the Look up Quota screen. 
**Impact:** bad user experience. 
**Attachment:** LookUpScreen.png
![LookUpScreen](https://user-images.githubusercontent.com/55572027/97441502-d06ed500-194e-11eb-98f0-49bb521b8887.png)

Lss-Ankit:
Hi I have verified this issue using latest Android internal Build v1.3.0(1004294316) on lower end devices & observed that issue is still reproduced.
**Repro on devices:** Samsung Galaxy A3(5.0.2), Samsung Galaxy A5(7.0)

Lss-Ankit:
Verified issue on latest test flight build v1.5.5 (34) and Android play store internal build v1.5.5 (1004294323) and found that issue is still repro

Lss-Ankit:
Verified issue on latest App store build v1.5.5 and Android play store  build v1.5.5  and found that issue is still reproduced.
**Repro on devices:** Samsung Galaxy A5(7.0), iPhone 6+(12.4.5)

Lss-Ankit:
Hi @gnardini I have verified this issue using latest Android internal build v1.6.0(1004294326) & Test Flight Build v1.5.1(37) and observe issue is still reproduced.
**Repro on devices:** Samsung Galaxy A5(7.0) ,iPhone 7 Plus (14.1)

nityas:
@Lss-Ankit -- with feeless onboarding, how is this still reproduced?

[...]

PR #5622, opened 2020-10-29
Komenci <> Valora Integration: contact mapping + DEK/wallet registration and querying

### LINKED ISSUE #62: Sender/Requester of funds needs to link SCW to EOA

tarikbellamine:
### Description
This PR seeks to lay the initial foundation for supporting Komenci and the introduction of meta-transaction wallets (referred to as MTW) within Valora. A key structural shift that is occurring is Valora will now have two relevant addresses: the user's MTW address and the externally owned account (EOA) address, which has historically been the user's only account.

Initially, the MTW will serve as a means to feelessly onboard new users (via the Komenci service) but will eventually be able to unlock many more features (e.g., more user friendly account recovery, abstraction of seed phrase, etc.).

Two key changes are occurring in this PR:
- the MTW is being registered as the `accountAddress` while the EOA is being registered as the `walletAddress` (previously the EOA was both the wallet address and account address)
- the phone number lookup logic is looking up the `walletAddress` (previously we were looking up the `accountAddress` because the EOA was both `walletAddress` and `accountAddress` so the distinction didn't practically matter)

### Other changes
Put some placeholders for where we will store the MTW address, which was necessary to keep the build stable while development is ongoing.

### Tested
New tests added. Was also able to successfully lookup and send from an iOS and Android device.

### Related issues
- Closes https://github.com/celo-org/komenci/issues/62

### Backwards compatibility
Yes


packages/mobile/src/identity/contactMapping.ts
```diff
@@ -218,19 +221,44 @@ export function* fetchAddressesAndValidateSaga({
...
-  }
-
+function* getAccountAddresses(e164Number: string) {
+  const phoneHashDetails: PhoneNumberHashDetails = yield call(fetchPhoneHashPrivate, e164Number)
+  const phoneHash = phoneHashDetails.phoneHash
   const lookupResult: IdentifierLookupResult = yield call(lookupAttestationIdentifiers, [phoneHash])
   return getAddressesFromLookupResult(lookupResult, phoneHash) || []
 }
 
+function* getWalletAddressesAndUpdateCache(e164Number: string) {
```
tarikbellamine: @jmrossy can you please take a look at this and lmk if there are some edge cases I may be missing?

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -103,10 +126,53 @@ export function* registerAccountDek(account: string) {
...
+    // If MTW has been created, route the user's DEK/wallet registration through it
+    // because accountAddress is determined by msg.sender
+    if (mtwAddress) {
+      try {
+        const mtwWrapper: MetaTransactionWalletWrapper = yield call(
+          [contractKit.contracts, contractKit.contracts.getMetaTransactionWallet],
+          mtwAddress
+        )
+
+        const proofOfPossession: {
```
tarikbellamine: @nambrot does this look right to you?

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -103,10 +126,53 @@ export function* registerAccountDek(account: string) {
...
+        // is an active Komenci session (i.e., submit through KomenciKit or not)
+
+        setAccountTx = accountsWrapper.setAccount(
+          '',
+          publicDataKey,
+          walletAddress,
+          proofOfPossession
+        )
+
+        const setAccountTxViaMTW: CeloTransactionObject<string> = yield call(
```
tarikbellamine: @bowd does the above MTW instantiation + this transaction creation and submission look right to you?

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -218,19 +221,44 @@ export function* fetchAddressesAndValidateSaga({
...
+  walletAddresses.forEach((address, i) => {
+    const walletAddress = address.toLowerCase()
+    const accountAddress = accountAddresses[i]
+    // `getWalletAddress` returns 0x0 when there isn't a wallet registered
+    if (walletAddress !== '0x0') {
+      walletToAccountAddress[walletAddress] = accountAddress
+      possibleUserAddresses.push(walletAddress)
+    } else {
+      // NOTE: Only need this else block if we are not confident all wallets are registered
+      walletToAccountAddress[accountAddress] = accountAddress
```
jeanregisser, line 256: Is the `accountAddress` already in lowercase?

packages/mobile/src/web3/actions.ts
```diff
@@ -72,6 +79,13 @@ export const setAccount = (address: string): SetAccountAction => {
   }
 }
 
+export const setMtwAddress = (address: string): SetMtwAddressAction => {
```
jeanregisser, line 82: I don't see this function being referenced in this diff. Preparing for future PR?

tarikbellamine: Yes exactly, it will be set during the verification flow (next PR). Given we are referencing the empty state variable here to keep the master build stable, I wanted to also create the setter function in this PR so I don't forget to do it later

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: Yeah, [right here](https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/identity/contactMapping.ts#L271). Thanks for checking

packages/mobile/src/identity/matchmaking.ts
```diff
@@ -12,12 +12,16 @@ import { NumberToRecipient } from 'src/recipients/recipient'
...
+import { mtwAddressSelector } from 'src/web3/selectors'
 
 const TAG = 'identity/matchmaking'
 
 // Uses the phone number privacy service to find mutual matches between Celo users
 export function* fetchContactMatches(e164NumberToRecipients: NumberToRecipient) {
-  const account: string = yield call(getConnectedUnlockedAccount)
+  const walletAddress: string = yield call(getConnectedUnlockedAccount)
+  const mtwAddress: string | null = yield select(mtwAddressSelector)
+  const accountAddress: string = mtwAddress || walletAddress
```
jmrossy: I'd recommend a more descriptive name for these concepts, the difference between walletAddress / accountAddress isn't clear. It's long but explicitly labeling things as EOAs or MTWs may be the safest thing. 

Related, I'd also recommend you create a new module and isolate all address disambiguation to that one place. That way you don't need to have unrelated modules (like this one for fetching contact matches) be aware of the subtle differences. Maybe adding this logic to the `getConnected*` sagas makes sense.

packages/mobile/src/identity/reducer.ts
```diff
@@ -92,6 +96,7 @@ export interface State {
   addressToE164Number: AddressToE164NumberType
   // Note: Do not access values in this directly, use the `getAddressFromPhoneNumber` helper in contactMapping
   e164NumberToAddress: E164NumberToAddressType
+  walletToAccountAddress: WalletToAccountAddressType
```
jmrossy, line 101: A comment here around what this is for would probably be useful :)

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -218,19 +221,44 @@ export function* fetchAddressesAndValidateSaga({
...
+    accountAddresses.map((accountAddress) => call(accountsWrapper.getWalletAddress, accountAddress))
+  )
+
+  const possibleUserAddresses: string[] = []
+  const walletToAccountAddress: WalletToAccountAddressType = {}
+  walletAddresses.forEach((address, i) => {
+    const walletAddress = address.toLowerCase()
+    const accountAddress = accountAddresses[i]
+    // `getWalletAddress` returns 0x0 when there isn't a wallet registered
+    if (walletAddress !== '0x0') {
```
jmrossy: To be safe, might be good to handle more cases (e.g. `0x000000000000000000000000`). `BigNumber.from(walletAddress).isZero()` should do the trick

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -218,19 +221,44 @@ export function* fetchAddressesAndValidateSaga({
...
+      walletToAccountAddress[walletAddress] = accountAddress
+      possibleUserAddresses.push(walletAddress)
+    } else {
+      // NOTE: Only need this else block if we are not confident all wallets are registered
```
jmrossy, line 255: All what kind of wallets are registered, EOA or MTW? Many EOA will not be registered with the Accounts contract.

packages/mobile/src/identity/contactMapping.ts
jmrossy: I think the convention we use elsewhere is `fetch` which implies a cache update to. We haven't done it consistently though.

packages/mobile/src/identity/contactMapping.ts
jmrossy, line 256: I'm not sure I understand the intention of this mapping. When is it meaningful to say an account address maps to itself? What's a wallet in this context if not one of the `walletAddresses` from above? Can you help me understand this bit?

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -218,19 +221,44 @@ export function* fetchAddressesAndValidateSaga({
...
+
+  const possibleUserAddresses: string[] = []
+  const walletToAccountAddress: WalletToAccountAddressType = {}
+  walletAddresses.forEach((address, i) => {
+    const walletAddress = address.toLowerCase()
+    const accountAddress = accountAddresses[i]
+    // `getWalletAddress` returns 0x0 when there isn't a wallet registered
+    if (walletAddress !== '0x0') {
+      walletToAccountAddress[walletAddress] = accountAddress
+      possibleUserAddresses.push(walletAddress)
```
jmrossy, line 253: So the intention is to give precedence to MTWs instead of EOAs when we find one, is that right?

packages/mobile/src/identity/contactMapping.ts
jmrossy: Tough to say from just a PR review but can't think of anything obvious missing.

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -103,10 +126,53 @@ export function* registerAccountDek(account: string) {
     }
 
     // Generate and send a transaction to set the DEK on-chain.
-    const setAccountTx = accountsWrapper.setAccount('', publicDataKey, account)
+    let setAccountTx = accountsWrapper.setAccount('', publicDataKey, walletAddress)
     const context = newTransactionContext(TAG, 'Set wallet address & DEK')
-    yield call(sendTransaction, setAccountTx.txo, account, context)
+    // If MTW has been created, route the user's DEK/wallet registration through it
+    // because accountAddress is determined by msg.sender
+    if (mtwAddress) {
```
jmrossy: Nit consider breaking this out. Similar to my comment above, it probably makes sense to group MTW related complexity in one place. And/or maybe a new option to `sendTransaction` that handles this re-routing. Ultimately up to you, just thinking out loud!

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -103,10 +126,53 @@ export function* registerAccountDek(account: string) {
...
+          walletAddress,
+          proofOfPossession
+        )
+
+        const setAccountTxViaMTW: CeloTransactionObject<string> = yield call(
+          mtwWrapper.signAndExecuteMetaTransaction,
+          setAccountTx.txo
+        )
+        yield call(sendTransaction, setAccountTxViaMTW.txo, walletAddress, context)
+        yield put(updateWalletToAccountAddress({ [walletAddress.toLowerCase()]: mtwAddress }))
```
jmrossy: how about the action creator handles lowercasing?

jmrossy: Can you help me understand why this is here? contact mappings seem unrelated to DEK registration

packages/mobile/src/identity/matchmaking.ts
tarikbellamine: Good point about isolating the disambiguation. Probably too soon to integrate it into existing `getConnected*` functions because we don't want to do this everywhere just yet. I repeat this logic at least a few times (and will probably add more in subsequent PRs) so a new getter function makes a lot of sense.

As for terminology, what I was going for was consistency with how these are referred to on the contracts side where a user can have both an `accountAddress` and `walletAddress`. This is generally supposed to be a parent (account) child (wallet) relationship where the two addresses can be the same but don't have to be. Am I misunderstanding the existing terminology?

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: Oh that's a nifty trick. Will do that, thanks

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: I believe all `walletAddress` get registered at the same time as the DEK when we call the `setAccount` method on the Accounts contract during the `registerDek` saga.

I recently learned that when calling `setAccount` the sender is automatically registered as the `accountAddress` and the argument passed is registered as the `walletAddress`. So when we register the DEK on-chain with an EOA and also pass in the EOA address as an argument, that's how we end up with the `accountAddress` equaling the `walletAddress`.

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: With feeless attestations we are moving towards a world where the MTW is what gets verified instead of the EOA. That means when we check for addresses verified against a given identifier, we will be receiving back verified MTWs (for users who verified feelessly) and EOAs (for users who paid for their own verification).

What I am doing here is checking all the "verified" addresses and seeing if they have a `walletAddress` registered to them. This way, I can be sure that we will always be sending to EOAs (and never MTWs).

In the event they don't have a wallet registered (the `else` block), I am assuming that the "verified" address is an EOA. This mostly exists for the low probability edge case that a user has a verified account but has not yet registered their DEK on-chain (because wallet address gets registered at the same time as the DEK).

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: Adding to the mapping in this block is mostly for completeness rather than absolute necessity.

In the `doFetchDataEncryptionKey` saga, I need a way to go from EOA to MTW. I need this because moving forward the DEK will be registered to the MTW and the metadata included in transactions (for encryption) only includes the EOA. There is no easy way to go from `walletAddress` => `accountAddress` on-chain, which is why we need the local mapping. 

Adding to the mapping in this block basically just lets me use the mapping later without caring. As an alternative, when I use the mapping later I could say `if map lookup returns undefined, use EOA`. Do you think that would be more intuitive?

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: The `setAccount` method does 3 things:
- sets the `accountAddress = msg.sender`
- sets the `walletAddress = the address argument` passed to it (proof of possession of wallet is required when account != wallet)
- registers the public `DEK`to the `accountAddress`

Since this is the place the `walletAddress` is getting registered, I figured it also made sense to update our mapping here.

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: Yeah, we are kinda caught in a weird spot where we want to eventually move towards every user having a MTW but first introduce it only for a specific use case (feeless onboarding). That's why I've been holding off on abstracting this kind of logic into functions that are widely used like `sendTransaction`.

Your point still stands though - I will think about how I can better contain this logic.

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -169,31 +170,33 @@ export function* fetchAddressesAndValidateSaga({
...
     }
 
     const userAddress = yield select(currentAccountSelector)
+    const secureSendPossibleAddresses = [...walletAddresses]
     const secureSendPhoneNumberMapping = yield select(secureSendPhoneNumberMappingSelector)
     // If fetch is being done as part of a payment request from an unverified address,
     // the unverified address should be considered in the Secure Send check
-    if (requesterAddress && !addresses.includes(requesterAddress)) {
-      addresses.push(requesterAddress)
+    if (requesterAddress && !secureSendPossibleAddresses.includes(requesterAddress)) {
```
asaj, line 193: I'm not sure I fully understand what's happening here. If we received a payment request, why do we need to do secure send?

packages/mobile/src/identity/contactMapping.ts
asaj: +1, you can do `eqAddress(walletAddress, NULL_ADDRESS)`

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -218,19 +221,44 @@ export function* fetchAddressesAndValidateSaga({
...
+
+  const accountAddresses: string[] = yield call(getAccountAddresses, e164Number)
+  const walletAddresses: string[] = yield all(
+    accountAddresses.map((accountAddress) => call(accountsWrapper.getWalletAddress, accountAddress))
+  )
+
+  const possibleUserAddresses: string[] = []
+  const walletToAccountAddress: WalletToAccountAddressType = {}
+  walletAddresses.forEach((address, i) => {
+    const walletAddress = address.toLowerCase()
```
asaj: You can use `normalizeAddress` or `normalizeAddressWith0x`. Note that anywhere addresses are presented to the user we should use checksum addresses

packages/mobile/src/identity/contactMapping.ts
asaj: Are we caching the account address as well? If not, how will we look up e.g. the DEK?

IMO the best thing would probably be to have an `Account` object, which consists of an account address, DEK, wallet address, verification info, etc., and have a cached mapping for which phone numbers are the keys and `Account`s are the values (well, probably actually two mappings, one from phone number -> (pepper, identifier), and one from identifier -> Account). And various operations on the `Account` object that allow us to refresh some (or all) of the cached values.

packages/mobile/src/identity/matchmaking.ts
asaj: Yeah +1 for account and wallet address

packages/mobile/src/identity/matchmaking.ts
```diff
@@ -12,12 +12,16 @@ import { NumberToRecipient } from 'src/recipients/recipient'
...
 import { getAuthSignerForAccount } from 'src/web3/dataEncryptionKey'
 import { getConnectedUnlockedAccount } from 'src/web3/saga'
+import { mtwAddressSelector } from 'src/web3/selectors'
 
 const TAG = 'identity/matchmaking'
 
 // Uses the phone number privacy service to find mutual matches between Celo users
 export function* fetchContactMatches(e164NumberToRecipients: NumberToRecipient) {
-  const account: string = yield call(getConnectedUnlockedAccount)
+  const walletAddress: string = yield call(getConnectedUnlockedAccount)
```
asaj, line 20: Should these be of type `Address`?

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -153,7 +225,7 @@ export async function getRegisterDekTxGas(account: string, currency: CURRENCY_EN
   }
 }
 
-export function* getAuthSignerForAccount(account: string) {
+export function* getAuthSignerForAccount(accountAddress: string, walletAddress: string) {
```
asaj, line 232: `s/comment key/data encryption key` below

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -170,7 +242,8 @@ export function* getAuthSignerForAccount(account: string) {
       const upToDate: boolean = yield call(
         isAccountUpToDate,
         accountsWrapper,
-        account,
+        accountAddress,
```
asaj, line 249: I'm not totally clear when we make ODIS requests in Valora but just want to flag the additional performance cost of checking that our DEK is registered here

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -103,10 +126,53 @@ export function* registerAccountDek(account: string) {
     }
 
     // Generate and send a transaction to set the DEK on-chain.
-    const setAccountTx = accountsWrapper.setAccount('', publicDataKey, account)
+    let setAccountTx = accountsWrapper.setAccount('', publicDataKey, walletAddress)
```
asaj: If you just need to register the DEK, suggest calling `setAccountDataEncryptionKey`. If you just need to register the wallet address, suggest calling `setWalletAddress`. Only if you need to set both is it best to use `setAccount`

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -103,10 +126,53 @@ export function* registerAccountDek(account: string) {
...
+        const setAccountTxViaMTW: CeloTransactionObject<string> = yield call(
+          mtwWrapper.signAndExecuteMetaTransaction,
+          setAccountTx.txo
+        )
+        yield call(sendTransaction, setAccountTxViaMTW.txo, walletAddress, context)
+        yield put(updateWalletToAccountAddress({ [walletAddress.toLowerCase()]: mtwAddress }))
+      } catch (error) {
+        Logger.debug(
+          `${TAG}@registerAccountDek`,
+          'Failed to register via the MTW, will attempt with EOA. Error: ',
```
asaj: I don't think this is the right fallback behavior?

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: This is needed to cover an edge case where:
- someone requesting money from you has a verified account on-chain
- but they are making a request from a non-verified account that they own

What this does is just adds the unverified account to the list of possible accounts to disambiguate.

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: Yeah see the `updateWalletToAccountAddress` action below. We are storing a mapping of `walletAddress => accountAddress` specifically for DEK lookups.

I agree that there is probably a better way to organize and consolidate the disparate caches within Valora but I am inclined to kick that can down the road for now.

packages/mobile/src/identity/matchmaking.ts
tarikbellamine: We use `string` instead of `Address` in most places across Valora. I agree we should be better about using more specific typing.

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: Yup, I think we can remove this when we have more certainty that DEKs are being registered. I believe the reason we had this is because we placed DEK registration attempts in various spots across Valora to ensure the widest possible coverage (believe it happens on the first of invite redemption, verification, or user's first send tx) but also didn't want to charge the user for transactions they didn't need.

Good call out but don't feel super comfortable removing it at this point.

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: Yeah we need to set both. I will update the naming to reflect that in a subsequent PR.

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: You're right, not what we want to do here.

packages/mobile/src/identity/actions.ts
```diff
@@ -312,6 +320,24 @@ export const updateE164PhoneNumberAddresses = (
...
 })
 
+export const updateWalletToAccountAddress = (
+  walletToAccountAddress: WalletToAccountAddressType
+): UpdateWalletToAccountAddressAction => {
+  const newWalletToAccountAddresses: WalletToAccountAddressType = {}
+  const walletAddresses = Object.keys(walletToAccountAddress)
+
+  for (const walletAddress of walletAddresses) {
+    const newWalletAddress = walletAddress.toLowerCase()
```
nategraf: Similar to using `eqAddress` elsewhere, we likely want to use `normalizeAddress` from `@celo/base` here and anywhere else where our goal is to normalize the address.

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -216,15 +221,43 @@ export function* fetchAddressesAndValidateSaga({
...
+    // `getWalletAddress` returns a null address when there isn't a wallet registered
+    if (!eqAddress(walletAddress, NULL_ADDRESS)) {
+      walletToAccountAddress[walletAddress] = accountAddress
+      possibleUserAddresses.push(walletAddress)
+    } else {
+      // NOTE: Only need this else block if we are not confident all wallets are registered
+      walletToAccountAddress[accountAddress] = accountAddress
+      possibleUserAddresses.push(accountAddress)
+    }
+  })
```
nategraf: `.forEach` is a bit of an anti-pattern. Using `for ... of` with `.entries` should be preferred.
```suggestion
 for (const [i, walletAddress] of walletAddresses.entries()) {
    const accountAddress = accountAddresses[i]
    // `getWalletAddress` returns a null address when there isn't a wallet registered
    if (!eqAddress(walletAddress, NULL_ADDRESS)) {
      walletToAccountAddress[walletAddress] = accountAddress
      possibleUserAddresses.push(walletAddress)
    } else {
      // NOTE: Only need this else block if we are not confident all wallets are registered
      walletToAccountAddress[accountAddress] = accountAddress
      possibleUserAddresses.push(accountAddress)
    }
  }
```

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -44,8 +54,12 @@ export function* doFetchDataEncryptionKey(address: string) {
     contractKit.contracts,
     contractKit.contracts.getAccounts,
   ])
-  const dek: string = yield call(accountsWrapper.getDataEncryptionKey, address)
-  yield put(updateAddressDekMap(address, dek || null))
+  const walletToAccountAddress: WalletToAccountAddressType = yield select(
+    walletToAccountAddressSelector
+  )
+  const accountAddress = walletToAccountAddress[walletAddress] || walletAddress
```
nategraf: It may not be wise to default to `walletAddress` if the key is not found, because that indicates that mapping is not set, whereas if it is correctly set to the EOA the mapping will return the `walletAddress`. The default may need to be a fetch on the account synchronously instead. (Also, the `??` operator is more appropriate here than the `||` operator)

packages/mobile/src/web3/saga.ts
```diff
@@ -298,6 +298,16 @@ export function* getConnectedUnlockedAccount() {
...
 }
 
+// This will return MTW if there is one and the EOA if
+// there isn't. Eventually need to change naming convention
+// used elsewhere that errouneously refers to the EOA
+// as `account`
+export function* getAccountAddress() {
+  const walletAddress: string = yield call(getAccount)
+  const mtwAddress: string | null = yield select(mtwAddressSelector)
+  return mtwAddress || walletAddress
```
nategraf:
```suggestion
  return mtwAddress ?? walletAddress
```

nategraf (commented):
I don't currently see anywhere that wallet address registration on the account is a blocking requirement. It occurs in the verification flow and elsewhere, tied to the DEK registration, but is allowed to fail silently. With this PR we need to make sure it succeeds because if the phone number is verified without registering the wallet address on the account, a user will look up the phone number mapping, get the address of the MTW, and without a wallet address set will send the funds there assuming it is the EOA. Additionally, this value will be cached and may result in the error persisting even after the account owner fixes the issue.

I also had one other comment on an edge case in the registration.

nategraf (approved):
Approving to allow flexibility in how your address my review comments. (i.e. either in this PR or in a follow-up PR)

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: > It may not be wise to default to `walletAddress` if the key is not found, because that indicates that mapping is not set, whereas if it is correctly set to the EOA the mapping will return the `walletAddress`

I don't think that's necessarily true. The vast majority of users today have their DEK's registered and both their `accountAddress` and `walletAddress` set as their EOA address on the Accounts contract. That's why it seemed reasonable to me to default to checking the user's EOA if an `accountAddress` didn't exist in their locally stored mapping.

Ultimately, I don't see much downside to doing this because if a DEK is not found, the comment just won't be encrypted. The worst case scenario is that comments get encrypted with a DEK the user no longer owns. I don't think we have to deal with that until users can recover their account without their EOA (because the DEK is derived from their EOA mnemonic).

> The default may need to be a fetch on the account synchronously instead

Can you please expand on this?

nategraf: Ah, I see your point. LGTM 👍

packages/mobile/src/identity/contactMapping.ts
asaj: Agree it doesn't make sense to do now, just flagging for the future since it looks like this will be worth cleaning up a bit when we have time

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -59,9 +73,47 @@ export function* createAccountDek(mnemonic: string) {
...
+    const proofOfPossession: {
+      v: number
+      r: string
+      s: string
+    } = yield call(accountsWrapper.generateProofOfKeyPossession, accountAddress, walletAddress)
+
+    setAccountTx = accountsWrapper.setAccount('', publicDataKey, walletAddress, proofOfPossession)
+
+    const setAccountTxViaMTW: CeloTransactionObject<string> = yield call(
+      mtwWrapper.signAndExecuteMetaTransaction,
```
bowd: Not 100% sure but you might need to do `[mtwWrapper, mtwWrapper.signAndExecuteMetaTransaction]` here like you're doing above in order to have `this` bound in the function call.

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -59,9 +73,47 @@ export function* createAccountDek(mnemonic: string) {
...
+      setAccountTx.txo
+    )
+    yield call(sendTransaction, setAccountTxViaMTW.txo, walletAddress, context)
+    yield put(updateWalletToAccountAddress({ [walletAddress]: accountAddress }))
+  } else {
+    yield call(sendTransaction, setAccountTx.txo, walletAddress, context)
```
bowd, line 113: Don't you need to also update the mapping here?
`yield put(updateWalletToAccountAddress({ [walletAddress]: walletAddress }))`

bowd (approved):
Great work :+1:

Not due to something specific but overall it feel like we will need to take a step back and really think about these changes after GA because it's starting to feel like we're introducing more and more edge cases. The day of reckoning shall come :D

I know that this was more or less a rush job but it feels as solid as can be 🎸

packages/mobile/src/identity/contactMapping.ts
jmrossy: Thanks for clarifying. I think leaving the mapping undefined is slightly more clear but I'll leave that up to you :)

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: Yeah good call, I changed this on a different branch but this is the right one to do that

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: I default to using `walletAddress` when nothing is found in the mapping but I guess there is no harm in also writing to this mapping with your sugggestion

Lss-Ankit:
Hi @tarikbellamine Verified task on latest test flight build v1.5.5 (34) and Android play store internal build v1.5.5 (1004294323) and found the following 
- [x] Verify "Lookup Quota exceeded" screen is shown when non verified user search verified user more than 3 times. 
- [ ] Verify Look up Quota exceeded screen UI. 
- [x] Verify user is able to Skip the look quota screen by tapping on Skip button. 
- [x] Verify  "Your phone number lookup quota is used up" message is shown after tapping on Skip button. 
- [x] Verify spinning is shown for a while when user taps on Purchase Lookup (2) button. 
- [x] Verify user is redirect to Send Amount screen after tapping on Purchase Lookup (2) button. 
- [x] Verify "Lookup Quota Purchase" transaction with $0.01 amount is shown on transaction feed. 
- [x] Verify Lookup Quota Purchase screen is not shown now when user sending a payment to verified user. 

**Existing issue:**
- https://github.com/celo-org/celo-monorepo/issues/5611 

Please let me know if i need to verify thing else?

Issue #5426, opened 2020-10-19
Integration support for feeless attestation


jeanregisser:
### What's important about this?

Feeless attestation is a blocker for GA. We want to be sure the integration with the wallet is optimal.

### Definition of "done"

Feeless attestation is integrated and fully functional.

### What's involved in doing this work?

- Help review PR related to feeless attestation
- Give input when needed
- Help write unit/e2e tests needed
- Thorough test of each delivered PR related to feeless attestation 

### Open questions

- ?

gnardini:
The work on this is done for now :)

PR #5630, opened 2020-10-29
[Wallet] Add reCaptcha verification behind feature flag

gnardini:
### Description

Adds reCaptcha implementation behind a feature flag. If `KOMENCI` is `true`, after clicking on `Start` on the `VerificationEducationScreen` the user will be prompted with a reCaptcha challenge that they'll need to complete before being able to verify. It also adds [SafetyNet](https://developer.android.com/training/safetynet/attestation) which is a check for Android devices.

This is missing integration with KomenciKit

### Tested

By changing the value of the KOMENCI flag and seeing the captcha flow (captcha code is logged)

### Related issues

- Part of #5426


packages/mobile/android/app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gnardini, line 1: I know this file looks weird, it's because of `SafetyNet`. It depends on `react-native-securerandom` which doesn't support automatic linking, and when doing it manually it threw an error because this file was missing. 

This manual linking is the reason for the changes to the `settings.gradle` and the Podfile as well.

packages/mobile/src/transactions/__snapshots__/TransferConfirmationCard.test.tsx.snap
```diff
@@ -108,17 +108,26 @@ exports[`TransferConfirmationCard renders correctly for CELO withdrawal transact
                   </svg>
                 </View>
               </View>
-              <Text
+              <View
```
gnardini, line 111: This changes are due to the change to the implementation of `Expandable`. It changes nothing visually, it's just wrapping the `Text` with a `View`.

packages/mobile/locales/en-US/onboarding.json
```diff
@@ -43,9 +43,9 @@
     "loadingBody": "This can take up to 90 seconds"
   },
   "verificationEducation": {
-    "title": "Confirm",
-    "header": "Confirm phone number",
-    "body": "To make sure your number is really yours, we’re going to send you three messages that will cost about 0.05 cUSD each.\n\nConfirming your number takes about three minutes.",
+    "title": "Phone Number",
+    "header": "Connect your phone number",
+    "body": "Connecting takes about three minutes. To confirm your number, you’ll receive three messages.",
```
cmcewen: Do we need to keep this text for the fallback version? cc @tarikbellamine

packages/mobile/package.json
```diff
@@ -107,6 +107,7 @@
     "react-native-fs": "^2.16.6",
     "react-native-gesture-handler": "^1.6.1",
     "react-native-geth": "https://github.com/celo-org/react-native-geth#5864c09",
+    "react-native-google-safetynet": "https://github.com/celo-org/react-native-google-safetynet#e698d70",
```
cmcewen: not sure if you created a PR to get this upstreamed - but usually nice to do

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -47,6 +60,12 @@ function VerificationEducationScreen({ route, navigation }: Props) {
   const numberVerified = useSelector(numberVerifiedSelector)
   const partOfOnboarding = !route.params?.hideOnboardingStep
 
+  const country = useMemo(() => {
+    const regionCode = getRegionCode(e164PhoneNumber || '')
+    const countries = new Countries(i18n.language)
```
cmcewen, line 65: this looks like it could be expensive - but just a note that useMemo is not always necessary https://blog.logrocket.com/rethinking-hooks-memoization/

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -93,6 +119,26 @@ function VerificationEducationScreen({ route, navigation }: Props) {
     setShowLearnMoreDialog(false)
   }
 
+  const showCaptcha = async () => {
+    setIsCaptchaVisible(true)
+    const safetyNetAttestationResponse = await getSafetyNetAttestation()
+    Logger.info('SafetyNet attestation complete:', JSON.stringify(safetyNetAttestationResponse))
```
cmcewen, line 125: is this response big? do we need to log it all?

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -93,6 +119,26 @@ function VerificationEducationScreen({ route, navigation }: Props) {
...
+    setSafetyNetAttestation(safetyNetAttestationResponse)
+  }
+  const hideCaptcha = () => setIsCaptchaVisible(false)
+
+  const handleCaptchaResolved = (res: any) => {
+    hideCaptcha()
+    const captchaCode = res?.nativeEvent?.data
+    if (captchaCode !== 'cancel' && captchaCode !== 'error') {
+      Logger.info('Captcha code received: ', captchaCode)
+      // TODO: Add some brief loading indicator and send captcha somewhere before continuing.
```
cmcewen, line 135: given the feature flag this is okay, but normally i would request some changes 😛

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -162,12 +207,39 @@ function VerificationEducationScreen({ route, navigation }: Props) {
...
+          nationalPhoneNumber={getE164DisplayNumber(e164PhoneNumber || '')}
+          editable={false}
+        />
         {firstButton}
         <View style={styles.spacer} />
         <TextButton style={styles.doINeedToConfirmButton} onPress={onPressLearnMore}>
           {t('verificationEducation.doINeedToConfirm')}
         </TextButton>
       </ScrollView>
+      <Modal isVisible={isCaptchaVisible} style={styles.recaptchaModal}>
```
cmcewen, line 223: i mentioned this to cal, but we should just be mindful of modals from a perf perspective - can lead to overdraw https://developer.android.com/topic/performance/rendering/overdraw

packages/mobile/src/verify/safety/GoogleReCaptcha.tsx
```diff
@@ -0,0 +1,64 @@
...
+					window.ReactNativeWebView.postMessage("cancel"); 
+					document.getElementById('captcha').style.display = 'none';
+				}
+				var onDataExpiredCallback = function(error) {  window.ReactNativeWebView.postMessage("expired"); };  
+				var onDataErrorCallback = function(error) {  window.ReactNativeWebView.postMessage("error"); } 
+				</script> 
+			</head>
+			<body> 
+				<div id="captcha" style="height: 100vh;">
+          <div style="text-align: center; height: 100%; display: flex; flex-direction: column; justify-content: center; align-items: center;">
```
cmcewen: could you fix the indents here? (just kidding 😄 )

packages/mobile/src/verify/safety/GoogleReCaptcha.tsx
```diff
@@ -0,0 +1,64 @@
...
+              data-expired-callback="onDataExpiredCallback"  
+              data-error-callback="onDataErrorCallback">
+            </div>
+          </div>
+				</div>
+			</body>
+			</html>`
+  return (
+    <WebView
+      originWhitelist={['*']}
```
cmcewen: does this have to be *? it looks like we maybe want https? https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#originwhitelist

packages/mobile/android/app/src/main/res/values/strings.xml
cmcewen: Would love to fork/fix/upstream after wednesday as good citizens of the open source world

packages/mobile/locales/en-US/onboarding.json
tarikbellamine: Yeah good catch. We will want to keep all the old text because both feeless and non-feeless flows will be possible.

packages/mobile/android/app/src/main/res/values/strings.xml
gnardini: Yeah, I like doing that sort of stuff too, but the maintainers don't seem very active. The first reason I had to fork was to add support for AndroidX since it was failing for that reason. There's a PR that adds that from almost a year ago: https://github.com/rajivshah3/react-native-google-safetynet/pull/283 which has this message:

> Thanks for the PR! I can review and investigate the build failure in a few days when I have some time

(December 2019)

packages/mobile/src/verify/VerificationEducationScreen.tsx
gnardini: That was a great article, I've specially been guilty of `4. Using useMemo solely for referential equalities` in the past.

I didn't benchmark how expensive the call is tbh, but it's iterating and applying transformations over a list, if there are many re-renders it could get expensive I guess. I'll make sure not to abuse the use of `useMemo` :)

packages/mobile/src/verify/VerificationEducationScreen.tsx
gnardini: This is an example response:
```
{
"apkCertificateDigestSha256": ["+sYXRdwJA3hvue3mKpYrOZ9zSPC7b4mbgzJmdZEDO5w="], 
"apkDigestSha256": "5UVIjg1t6wE5CfIhgRMiXQAmB7PzUpSUyGuirP/qyOs=", 
"apkPackageName": "org.celo.mobile.alfajores.dev", 
"basicIntegrity": true, 
"ctsProfileMatch": true, 
"evaluationType": "BASIC", 
"nonce": "IdJQ46NEBT214jlGZ036i7vZq81j/dBiUOAR6HEM", 
"timestampMs": 1603991278194
}
```

Probably  not all fields are necessary, but I thought they might be useful.

packages/mobile/src/verify/VerificationEducationScreen.tsx
gnardini: Yeah, this PR by itself is not super useful, it's just laying the foundation for the next one when this is integrated with KomenciKit :D

packages/mobile/src/verify/safety/GoogleReCaptcha.tsx
gnardini: Run it through a formatter so it looks nicer :)

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -162,12 +207,39 @@ function VerificationEducationScreen({ route, navigation }: Props) {
...
+              height: headerHeight - insets.top,
+            },
+            styles.recaptchaClose,
+          ]}
+          title={t('global:close')}
+        />
+        <GoogleReCaptcha
+          siteKey={RECAPTCHA_SITE_KEY}
+          url={WEB_LINK}
+          languageCode="en"
```
tarikbellamine: Don't we want this to be whatever the app language is?

tarikbellamine (approved):
LGTM.

Please make sure you've thoroughly tested on both OS's and specifically the duplication of the phone number input part.

packages/mobile/src/verify/VerificationEducationScreen.tsx
jeanregisser: Yes, we do use these modals in a bunch a places already, we can address this separately.

yarn.lock
```diff
@@ -29300,6 +29318,20 @@ react-native-secure-randombytes@^3.0.2:
...
+  dependencies:
+    base64-js "*"
+
+react-native-securerandom@^1.0.0:
+  version "1.0.0"
+  resolved "https://registry.yarnpkg.com/react-native-securerandom/-/react-native-securerandom-1.0.0.tgz#1cff2f727c90c9ec3318b42dbf825a628b53b49b"
+  integrity sha512-lnhcsWloFzMN/HffyDBlh4VlqdhDH8uxEzUIH3aJPgC1PxV6OKZkvAk409EwsAhcmG/z3yZuVKegKpUr5IM9ug==
+  dependencies:
+    base64-js "*"
+
```
jeanregisser: We shouldn't add 2 versions of the same depdency here.

jeanregisser (approved):
Looking good! 👍  See small comment below.

PR #5632, opened 2020-10-29
Add commands for slashing down validators using DowntimeSlasherv2

### LINKED ISSUE #5508: Testnet nodes can be removed with DowntimeSlasher

yorhodes:
### Description

- Adds `validator:set-bitmaps` command
- Adds `validator:downtime-slash` command
- Modify `validator:signed-blocks` command to work well with downtime slasher

### Other changes

- Adds `intRangeArray` flag parser

### Tested

Tested on private test net
Slashed successfully on baklava
[downtime-slash-history.txt](https://github.com/celo-org/celo-monorepo/files/5478700/downtime-slash-history.txt)

### Related issues

- Fixes #5508

### Backwards compatibility


AlexBHarley:
I guess I don't understand the bitmap stuff, anywhere you can point me to some docs? At first glance I don't understand how slashing start and end periods are a range, but that's unrelated to this PR.

yorhodes:
> I guess I don't understand the bitmap stuff, anywhere you can point me to some docs? At first glance I don't understand how slashing start and end periods are a range, but that's unrelated to this PR.

The canonical validator set for a given epoch is stored in an ordered list. The signature bitmap tracks whether a validator has signed all blocks in the current epoch by OR-ing all of the block header signatures. The signature bitmaps can't span epoch boundaries because the signer indices in the list can and will change. As a result, we need to slice the intervals on epoch boundaries. The bitmap set is also very gas heavy as it is doing a storage write for each block in the range, so we need to segment the intervals such that the gas limit isn't exceeded. 

I refactored this code a bunch to work with an interval type which is hopefully clearer for consumers.

packages/celotool/src/e2e-tests/slashing_tests.ts
```diff
@@ -304,33 +304,17 @@ describe('slashing tests', function(this: any) {
...
-        slotSize,
-        kit
-      )
-
-      for (let i = 0; i < intervalArrays.startBlocks.length; i += 1) {
-        await slasher
-          .setBitmapForInterval(intervalArrays.startBlocks[i], intervalArrays.endBlocks[i])
-          .send({ from: user, gas: 5000000 })
+      for (const interval of intervals) {
+        await slasher.setBitmapForInterval(interval).send({ from: user, gas: 5000000 })
```
nambrot, line 311: Why do you have to set gas here?

nambrot (approved):
💪

PR #5653, opened 2020-10-30
[Attestation Service] configurable rate limit

timmoreton:
### Description

* Make the rate limit configurable with the `RATE_LIMIT_REQS_PER_MIN` parameter.
* Fix a bug where the rate limit window was not being specified per the comment 

### Tested

Locally

### Backwards compatibility

Yes. Default value `300`.


packages/attestation-service/src/index.ts
```diff
@@ -45,14 +45,18 @@ async function init() {
 
   await startPeriodicHealthCheck()
 
+  const rateLimitReqsPerMin = parseInt(fetchEnvOrDefault('RATE_LIMIT_REQS_PER_MIN', '300'), 10)
   const rateLimiter = rateLimit({
-    windowMs: 5 * 60 * 100, // 5 minutes
-    max: 100,
```
aslawson, line 50: looks like we are increasing rate limit default from 100-->300 and decreasing the window (though technically increasing because of the bug)?  so this is much higher rate then we intended it being before 100/5 vs 300/1, any reason for the bump?

aslawson (approved):

packages/attestation-service/src/index.ts
timmoreton: you're right, I think this new default is too high. In a minute with 3 concurrent attestation requests of say 3-6 POSTs+GETs each, plus a load balancer hitting /healthz every 5 secs, plus a few /status and /metrics calls, I can't see us needing more than 50 requests per minute.

PR #5656, opened 2020-10-31
[Wallet] Fix account timeout issue during verification

### LINKED ISSUE #3280: Ongoing Verification Issues -- Android,iOS : ‘Verification Failed’ & Verification Error’ popup is shown while verifying the phone number.
**Frequency:** 100%
**Android App version:** Android Celo Alfa v1.6.2 (104794504)
**Repro on Android Device:** Samsung Galaxy J7 Prime (6.0.1)

**Repro Steps:** 
1) Launch the app and Perform sign up flow
2) Navigate to verification flow 
3) Try to paste the code and Observe

**Current Behavior:** ‘Verification Error’ popup is shown while verifying verification code

**Expected Behavior:** ‘Verification Failed’ & ‘Verification Error’ popup should be not shown and account should be verified 

**Impact:** User is unable to verify the account 

**Attachment:**  [Android_Verification failed error message is shown.mp4](https://drive.google.com/file/d/1SNfjcVGuKTKvAiOZPerc8TuC-J-cRkZU/view?usp=sharing) (Bug observed between 1:22 to 2:16 mins)

i1skn:
### Description

* Increase account unlock timeout to 10 minutes to fix referenced issue.
* Re-unlock account before verification start, so give at least 10 min guaranteed to a user to verify. It will also do it on resend messages.

### Other changes

Fix issue, when last confirmed attestation code is become empty after it is confirmed.

### Tested

Verification flow done

### Related issues

- Fixes https://github.com/celo-org/celo-monorepo/issues/3280

### Backwards compatibility

Yes


packages/mobile/src/identity/verification.ts
```diff
@@ -83,14 +84,19 @@ export interface AttestationCode {
...
     const attestationsWrapper: AttestationsWrapper = yield call([
       contractKit.contracts,
       contractKit.contracts.getAttestations,
     ])
 
+    // Lock and ask user for a PIN because we
+    // want to reset timeout for unlocked account
+    clearPasswordCaches()
+    yield call(unlockAccount, account, true)
+
```
jeanregisser, line 107: This causes the user to be asked for their PIN when starting verification, right?

packages/mobile/src/identity/reducer.ts
```diff
@@ -200,7 +200,7 @@ export const reducer = (
     case Actions.COMPLETE_ATTESTATION_CODE:
       return {
         ...state,
-        ...completeCodeReducer(state, state.numCompleteAttestations + 1),
+        numCompleteAttestations: state.numCompleteAttestations + 1,
```
jeanregisser, line 203: Why is this change needed?

jeanregisser (commented):
Thanks!
I see no tests were updated for this, can you add some to make sure this doesn't break in the future?

packages/mobile/src/identity/reducer.ts
i1skn: this is needed because `completeCodeReducer` should be used only for the case when we set completed codes initially. We can not use it here, because on this line https://github.com/celo-org/celo-monorepo/blob/77df351699e03bd9df96d6ab13eaa3d04a9c1886/packages/mobile/src/identity/reducer.ts#L370 we would assign to `attestationCodes[i]` undefined yet `acceptedAttestationCodes[i]`, which will be defined after this reducer is finishes.

packages/mobile/src/identity/verification.ts
i1skn: yeap 🙂

i1skn:
@jeanregisser I have addressed the feedback! PTAL

jeanregisser (approved):
Great! 👍

Lss-Ankit:
@jeanregisser Verified task using Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found that user is able to complete phone number verification flow with out facing any verification failed pop up. 

Verified on : iPhone XR (13.2.2) and Google Pixel XL (8.1)

PR #5658, opened 2020-11-01
Extend KomenciKit with CheckSession and QuotaExceededError

bowd:
### Description

Added a new action that verifies that a session is still around and returns the current action quota for that session.
Treat 429 explicitly by returning FetchError.QuotaExceededErrors that the client can interpret.


tarikbellamine (approved):
Thanks for the quick turnaround!

PR #5659, opened 2020-11-02
KomenciKit - Client-side un/blinding

codyborn:
### Description

Keep phone number secure by performing un/blinding on the Valora-side. This keeps the phone number mapping private, preventing Komenci from viewing the association.

### Tested

Unit tests

### Backwards compatibility

Not backward compatible, but it's not currently in use.


packages/komencikit/src/actions.ts
```diff
@@ -96,10 +96,16 @@ export const startSession = action<ActionTypes.StartSession, StartSessionPayload
 )
 
 interface GetDistributedBlindedPepperPayload {
-  e164Number: string
+  base64BlindedMessage: string
```
bowd: On Komenci's side you called this `blindedPhoneNumber` - not sure which one makes more sense but I think we should keep only one 😅

packages/komencikit/src/kit.ts
```diff
@@ -154,9 +160,31 @@ export class KomenciKit {
   })
   public async getDistributedBlindedPepper(
     e164Number: string,
-    clientVersion: string
+    clientVersion: string,
+    blsBlindingClient: BlsBlindingClient
```
bowd, line 164: I'm wondering whether it's a bit cleaner from the Valora perspective to just pass in the ODIS_PUB_KEY as an option to the constructor and construct the blinding client inside of komencikit :-/

Taking it as a param to this function is also a bit more composable in a sense. In the sense that I imagine `BlsBlindingClient` as a thin-ish interface that can be implemented in multiple ways and passed in.

bowd (changes requested):

packages/komencikit/src/actions.ts
codyborn: Good catch!

packages/komencikit/src/kit.ts
codyborn: Yeah, that's possible however the same BlsBlindingClient instance needs to be re-used for the unblinding (it's potentially stateful) so we'd have to make KomenciKit stateful as well which introduces more complexity. As we discussed in the sync today, Valora uses the ReactBlsBlindingClient while vanilla typescript uses the WasmBlsBlindingClient so it's easier to have the caller decide which client to use.

bowd: Yup I was split from the get-go but after we talked about it during the sync it did make more sense to pass that in here :+1:

bowd (approved):

PR #5660, opened 2020-11-02
[Wallet] Change Invite Send flow and add invite to drawer

gnardini:
### Description

- Adds a new item to the `NavigationDrawer` that opens a popup which allows a user to invite their friends.
- On the send flow, if the user is inviting someone, the 30 cent invite fee is no longer added and the transaction to send those 30 extra cents doesn't happen.

Those two things will only be enabled when the `KOMENCI` feature flag is turned on (it's off right now).

Note: Some of the texts are still under revision and might change.

### Other changes

Added a small change to the `VerificationEducationScreen` that was missing from a previous PR.
 
### Tested

Going through the Send flow inviting someone and seeing that no extra fee gets charged and nothing else unexpected happens. Also pressing on the Invite item on the drawer and trying to invite someone.

### Related issues

- Part of #5426


packages/mobile/locales/en-US/inviteFlow11.json
```diff
@@ -17,5 +17,10 @@
   "inviteFriendsToCelo": "Invite Friends to {{appName}}",
   "inviteAnyone": "Send easily to your friends by adding your contacts",
   "inviteReceived": "Invite Received",
-  "escrowPaymentNotificationTitle": "Invited {{mobile}}"
+  "escrowPaymentNotificationTitle": "Invited {{mobile}}",
+  "inviteDialog": {
+    "title": "Share prosperity by inviting a friend",
+    "body": "Connect with your friends and family on Valora. Then send money to each other in seconds.",
```
jeanregisser:
```suggestion
    "body": "Connect with your friends and family on {{appName}}. Then send money to each other in seconds.",
```

packages/mobile/locales/en-US/sendFlow7.json
```diff
@@ -46,8 +46,8 @@
   "sentEscrowPayment": "Sent Escrow Payment",
   "mobileNumber": "Mobile #",
   "walletAddress": "Wallet Address",
-  "inviteSMS": "Welcome to {{appName}}! Get {{appName}} at {{link}} and enter your code to get started: {{code}}",
-  "inviteSMSWithEscrowedPayment": "Welcome to {{appName}}! I've sent you Celo Dollars. Go to {{link}} and enter your code to get started: {{code}}",
+  "inviteWithoutPayment": "Hi {{name}}! I’ve been using Valora to send money worldwide. It’s easy to try and I want to invite you to check it out with this link: {{link}}",
+  "inviteWithEscrowedPayment": "Hi {{name}}! I’ve been using Valora to send money worldwide. I just sent you {{amount}} Celo Dollars. Please download the Valora App and redeem your Celo Dollars using this link: {{link}}",
```
jeanregisser:
```suggestion
  "inviteWithEscrowedPayment": "Hi {{name}}! I’ve been using Valora to send money worldwide. I just sent you {{amount}} Celo Dollars. Please download the {{appName}} App and redeem your Celo Dollars using this link: {{link}}",
```

packages/mobile/src/analytics/Properties.tsx
```diff
@@ -324,11 +324,22 @@ interface InviteEventsProperties {
     escrowIncluded: boolean
     error: string
   }
+  [InviteEvents.invite_start]: {
+    escrowIncluded: boolean
```
jeanregisser, line 340: Should we also track the escrowed amount?

packages/mobile/src/app/actions.ts
```diff
@@ -27,6 +27,7 @@ export enum Actions {
   SET_SESSION_ID = 'SET_SESSION_ID',
   OPEN_URL = 'APP/OPEN_URL',
   MIN_APP_VERSION_DETERMINED = 'APP/MIN_APP_VERSION_DETERMINED',
+  INVITE_MODAL_TEXT = 'APP/INVITE_MODAL_TEXT',
```
jeanregisser: Can we use an event name, maybe `TOGGLE_INVITE_MODAL`?

packages/mobile/src/invite/saga.test.ts
```diff
@@ -192,6 +201,43 @@ describe(watchSendInvite, () => {
...
+      .put(storeInviteeData(mockInviteDetails))
+      .put(
+        updateE164PhoneNumberAddresses(
+          {},
+          { [mockAccount.toLowerCase()]: mockInviteDetails.e164Number }
+        )
+      )
+      .run()
+
+    expect(Share.share).toHaveBeenCalled()
```
jeanregisser: nit: Can we also check the share arguments? to make sure the message is there.

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -73,6 +79,10 @@ type CustomDrawerItemListProps = Omit<DrawerContentOptions, 'contentContainerSty
   protectedRoutes: string[]
 }
 
+interface DrawerItemParams {
+  onPress: () => void
```
jeanregisser:
```suggestion
  onPress?: () => void
```

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -216,6 +229,9 @@ function CustomDrawerContent(props: DrawerContentComponentProps<DrawerContentOpt
 export default function DrawerNavigator() {
   const { t } = useTranslation(Namespaces.global)
   const isCeloEducationComplete = useSelector((state) => state.goldToken.educationCompleted)
+  const name = useSelector(nameSelector)
+  const dispatch = useDispatch()
+  const inviteLink = useAsync(generateInviteLink, []).result
```
jeanregisser: Shouldn't this be generated only onPress?

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -263,6 +279,24 @@ export default function DrawerNavigator() {
...
+                toggleInviteModal(
+                  t('sendFlow7:inviteWithoutPayment', {
+                    name,
+                    link: inviteLink,
+                  })
+                )
+              ),
+          }}
+          options={{ title: t('invite'), drawerIcon: Invite }}
+        />
```
jeanregisser: Does this replace the main active screen under the drawer?

packages/mobile/src/send/SendConfirmation.test.tsx
```diff
@@ -216,3 +229,38 @@ describe('SendConfirmation', () => {
...
+      },
+    })
+
+    const tree = render(
+      <Provider store={store}>
+        <SendConfirmation {...mockInviteScreenProps} />
+      </Provider>
+    )
+
+    expect(tree).toMatchSnapshot()
```
jeanregisser, line 264: I'm not a big fan of only having a snapshot test, as we usually ignore big snapshots as they are hard to read.
Would much prefer to see assertions on the presence of text/buttons.

I know we're short on time here and no need to update this now though 😉

packages/mobile/src/verify/VerificationEducationScreen.test.tsx
```diff
@@ -2,11 +2,22 @@ import * as React from 'react'
...
+
+  beforeAll(() => {
+    features.KOMENCI = true
+  })
+
+  afterAll(() => {
+    features.KOMENCI = komenciEnabled
+  })
+
   it('shows the `skip` button when already verified', () => {
```
jeanregisser, line 21: nit: maybe we want to test the KOMENCI=false case too?

jeanregisser (approved):
Great work! 👍

packages/mobile/locales/en-US/sendFlow7.json
```diff
@@ -46,8 +46,8 @@
   "sentEscrowPayment": "Sent Escrow Payment",
   "mobileNumber": "Mobile #",
   "walletAddress": "Wallet Address",
-  "inviteSMS": "Welcome to {{appName}}! Get {{appName}} at {{link}} and enter your code to get started: {{code}}",
-  "inviteSMSWithEscrowedPayment": "Welcome to {{appName}}! I've sent you Celo Dollars. Go to {{link}} and enter your code to get started: {{code}}",
+  "inviteWithoutPayment": "Hi {{name}}! I’ve been using Valora to send money worldwide. It’s easy to try and I want to invite you to check it out with this link: {{link}}",
```
jeanregisser:
```suggestion
  "inviteWithoutPayment": "Hi {{name}}! I’ve been using {{appName}} to send money worldwide. It’s easy to try and I want to invite you to check it out with this link: {{link}}",
```

packages/mobile/locales/en-US/inviteFlow11.json
gnardini: Whoops, missed the `Valora` text in a few places :P

packages/mobile/src/navigator/DrawerNavigator.tsx
gnardini: No, it can be called from any screen that has the hamburguer menu and the dialog will pop up on top.

packages/mobile/src/verify/VerificationEducationScreen.test.tsx
gnardini: While doing this I discovered something that was missing:

If the user had 0 balance, we were still showing the `Insufficient balance` button instead of the `Continue` button since a balance is not necessary for Komenci.

packages/mobile/src/send/SendConfirmation.test.tsx
gnardini: I agree, I added some assertions for the modal, maybe we could have some for the fee as well

packages/mobile/src/navigator/DrawerNavigator.tsx
gnardini: Maybe, yeah, I did it this way because I need the `inviteLink` to create the message to share and that is needed to toggle the modal, but there's probably a better way to do that, I'll give it some more thought

gnardini: Changed the place of the generation of the link so it happens when the invite button is tapped :)

Lss-Ankit:
@gnardini @jeanregisser Verified task on latest test flight build v1.5.5 (34) and Android play store internal build v1.5.5 (1004294323) 

- Verify Invite option is shown on Hamburger menu. 
- Verified user is able to send an invitation to any non Valora user from Menu. 
- Verified that 0.30 cUSD is not cut when user sending an invitation to non Valora user from send screen. 

**Need Clarification:**
For IOS: 
- Invitation message is shown as subject as well when user sending an invitation via Gmail app. 
Issue does not repro for Android. 
[Attachment](https://drive.google.com/file/d/10rNWPi7cxZAefvYljGP7QcCWIH_lvyAW/view?usp=sharing)
Devices Used : Google Pixel Xl (8.1) , iPhone SE 2 (14.1)

gnardini:
That depends on how each OS handles inviting, so it looks OK to me

PR #5662, opened 2020-11-02
[Wallet] Don't allow phone numbers from sanctioned countries

### LINKED ISSUE #5661: Block access to phone numbers from sanctioned countries
### Expected Behavior

If someone tries to use a phone number from a sanctioned country, they get blocked and can't continue.

### Current Behavior

They can use phone numbers from those countries without any problem.

gnardini:
### Description

If someone uses a phone number from a sanctioned country, they get stuck with an error on the screen to input name & number and can't continue to use the app.

### Tested

On the simulator with phone numbers from each of the banned countries.

### Related issues

- Fixes #5661
- Part of #5426


packages/utils/src/countries.ts
```diff
@@ -33,6 +33,13 @@ const matchCountry = (country: LocalizedCountry, query: string) => {
   )
 }
 
+// Iran, Cuba, North Korea, Sudan, Syria
+const bannedCountryCodeAlpha2 = ['IR', 'CU', 'KP', 'SD', 'SY']
+
+export const isCountryBanned = (countryCodeAlpha2: string) => {
+  return bannedCountryCodeAlpha2.indexOf(countryCodeAlpha2) >= 0
+}
+
```
jeanregisser: nit: was wondering if this should really be in utils, or specific only to Valora. Thoughts?

jeanregisser (approved):
LGTM 👍

packages/utils/src/countries.ts
gnardini: Yeah, I think that makes more sense, moved!

jeanregisser: Cool ok in that case, we should use the [country feature flags](https://github.com/celo-org/celo-monorepo/blob/2ff4dfd0f15c7287995410c3006e75dd5bf06377/packages/mobile/src/flags.ts#L20-L26), happy to patch this myself to get that off your plate.

gnardini: Thanks for offering, but it looked like something quick and I think it was a good time for me to use these flags so that I get familiar with them to use in the future :)

Lss-Ankit:
Hi @gnardini , @jeanregisser  can we get banned countries list so we can verify this task. 
Thanks!

jeanregisser:
Yes current list is: Iran, Cuba, North Korea, Sudan, Syria

Lss-Ankit:
Hi @jeanregisser thanks for the country list. 
Verified task on latest build Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found the following: 
- "Valora is not available in your country" message is shown when user taps on Next button from Account info screen after selecting and entering phone number for countries - Iran, Cuba, North Korea, Sudan, Syria

Need clarification: 
- We are able to see North Korea but able to see below option from Korea. 
1.  "Valora is not available in your country" message is shown when user taps on Next button from Account info screen after selecting and entering phone number for "Korea, Democratic People's republic.... " with +850 code. 
2. Also, able  to see +82 code - Korea, Republic of option and here user is able to sign up with this option. 
So can you please let us know is it expected behavior? 

Thanks!!

jeanregisser:
Yes this is expected.

PR #5663, opened 2020-11-02
Komenci <> Valora Integration: feeless verification

### LINKED ISSUE #77: Use KomenciKit to execute fee-less attestations in Valora

tarikbellamine:
### Description
This PR creates an alternative "feeless verification" flow that leverages the Komenci service to pay for verification related fees. In order to communicate with Komenci, Valora imports KomenciKit, which is a library (created by Bogdan) that wraps all the API calls in a manner that's easier to digest.

There are several new actions that need to take place for users to verify feelessly:
- We must check that Komenci is on and ready
- The user must complete a captcha (included in a previous PR by Gonza)
- Komenci will take the captcha token to start a new Komenci session
- Once in an active Komenci session, the API will be called to deploy a new meta-transaction wallet (MTW) through which we will be able to verify the user
- Register the DEK/relevant addresses on the Accounts contract before moving forward with formal verification (because not being able to do so now constitutes a breaking failure of verification)

From there, the verification flow largely follows the same structure as the original flow. In this first iteration of feeless verification, state is kept completely separate for normal verification to reduce complexity.

We've made the design decision to always allow the user to try feeless onboarding if:
- they are not already verified
- komenci is live
- the user has not encountered more than 2 unexpected Komenci errors in the last 3 hours (this was a somewhat arbitrary restriction and can be adjusted)

Please note that this PR builds on another open (but approved) PR. As such, it is safe to disregard the changes made in the following files:
- privateHashing
- contactMapping
- dataEncryptionKey (except for the `registerWalletAndDekViaKomenci` saga)

### Other changes
No

### Tested
Tested on device. Holding off on more comprehensive testing given our desire to integrate this more completely with the original verification flow.

### Related issues
- Closes https://github.com/celo-org/komenci/issues/77

### Backwards compatibility
No - this will require a breaking upgrade.


packages/komencikit/src/verifyWallet.ts
```diff
@@ -1,8 +1,8 @@
 import { Address, normalizeAddress } from '@celo/base'
 import { Err, Ok, Result } from '@celo/base/lib/result'
 import { ContractKit } from '@celo/contractkit'
+import { ABI as Proxy } from '@celo/contractkit/lib/generated/Proxy'
```
bowd: This isn't right -- we either extract the bytecode in a constat or we need it from protocol. The ABI in contractkit doesn't contain the bytecode.

packages/contractkit/src/governance/proxy.ts
```diff
@@ -63,10 +89,41 @@ export const PROXY_SET_IMPLEMENTATION_SIGNATURE = SET_IMPLEMENTATION_ABI.signatu
 export const PROXY_SET_AND_INITIALIZE_IMPLEMENTATION_SIGNATURE =
   SET_AND_INITIALIZE_IMPLEMENTATION_ABI.signature
 
-export const getInitializeAbiOfImplementation = (proxyContractName: string) => {
-  const implementationABI = require(`../generated/${proxyContractName.replace('Proxy', '')}`)
-    .ABI as AbiItem[]
-  const initializeAbi = implementationABI.find((item) => item.name === 'initialize')
+const findInitializeAbi = (items: AbiItem[]) => items.find((item) => item.name === 'initialize')
+
+const initializeAbiMap = {
```
bowd: Curious why this change is related to this PR - do we need to remove dynamic requires?

packages/komencikit/src/verifyWallet.ts
```diff
@@ -2,32 +2,26 @@ import { Address, normalizeAddress } from '@celo/base'
 import { Err, Ok, Result } from '@celo/base/lib/result'
```
bowd: This is fixed here: https://github.com/celo-org/celo-monorepo/pull/5724

packages/mobile/src/identity/verification.ts
```diff
@@ -319,6 +330,8 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
       }
 
       yield put(setVerificationStatus(VerificationStatus.CompletingAttestations))
+      // NOTE: I think we want the DEK registration out of the race because
+      // we want that to happen no matter what
```
bowd: +1

bowd (commented):
Awesome to see this coming together! I'm finish the review temporarily with a comment. I still need to wrap my head around what's happening in `feelessVerification.ts`. I'm not super in the know with react-native and sagas and what-not but after reading through it once my first feeling is that it's a bit too complex compared to my intuition of what should be going on. But it also feels like we are forced into this complexity by the existing house of cards.

I'll take another look tomorrow after I let it digest for a while. 🚀

asaj:
> my first feeling is that it's a bit too complex compared to my intuition of what should be going on. But it also feels like we are forced into this complexity by the existing house of cards.

Would it simplify things to model verification as a state machine?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+  )
+}
+
+function* fetchVerifiedMtw(contractKit: ContractKit, walletAddress: string, e164Number: string) {
+  Logger.debug(TAG, '@fetchVerifiedMtw', 'Starting fetch')
+  const feelessVerificationState: FeelessVerificationState = yield select(
+    feelessVerificationStateSelector
+  )
+
+  const phonHashDetails: PhoneNumberHashDetails = yield call(
```
asaj: typo

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+          possibleMtwAddress,
+          ALLOWED_MTW_IMPLEMENTATIONS,
+          walletAddress
+        )
+      )
+    )
+  } catch (error) {
+    Logger.debug(TAG, '@fetchVerifiedMtw', 'Unable to validate existing verified accounts')
+    Logger.error(TAG, '@fetchVerifiedMtw', error)
+    return null
```
asaj: Does this mean we will return null if one of the `verifyWallet` calls throws? If one were to succeed, for example, we wouldn't want to return null because another threw an exception

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+  }
+
+  const verifiedMtwAddresses = possibleMtwAddresses.filter(
+    (address, i) => verificationResults[i].ok
+  )
+
+  if (verifiedMtwAddresses.length > 2) {
```
asaj: Why 2 and not 1?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    throw Error(
```
asaj, line 614: Will this go to segment?

packages/mobile/src/identity/feelessVerification.ts
asaj: Or why > and not >=?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      'More than one verified MTW with walletAddress as signer found. Should never happen'
+    )
+  }
+
+  const verifiedMtwAddress = verifiedMtwAddresses[0]
```
asaj, line 619: What will happen if the length is zero?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    return null
+  }
+
+  yield put(
+    feelessUpdateVerificationState({
+      ...feelessVerificationState,
+      status: {
+        isVerified: true,
+        numAttestationsRemaining: 0,
+        total: NUM_ATTESTATIONS_REQUIRED,
```
asaj: Note that just because a matching address was found in the attestations contract, the attestations state is not necessarily what is being recorded here.

For example, you could have found an address with 3/10 completed attestations. In this case, probably we would want to complete more. I'm not sure how valora currently treats attestations state, does it look at the number of requested attestations, or just the number of completed attestations?

Specifically, with moving to N/M, these numbers are likely to be wrong.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+
+const TAG = 'identity/feelessVerification'
+
+const KOMENCI_URL = 'https://komenci.celo-networks-dev.org'
+// const KOMENCI_URL = 'http://192.168.86.33:3000'
+// TODO: Populate this with expected implementation address
+const ALLOWED_MTW_IMPLEMENTATIONS: Address[] = ['0x88a2b9B8387A1823D821E406b4e951337fa1D46D']
+const CURRENT_MTW_IMPLEMENTATION_ADDRESS: Address = '0x88a2b9B8387A1823D821E406b4e951337fa1D46D'
+
+export function* feelessFetchVerificationState() {
```
asaj, line 109: +1 to this as the first action! To risk sounding like a broken record, I think the onboarding code could be simplified greatly, and made much more reliable, by modeling the attestations flow as a state machine.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      url: KOMENCI_URL,
+      token: feelessVerificationState.komenci.sessionToken,
+    })
+
+    try {
+      // Checks if a MTW was verified in a previous attempt and updates state.
+      // This check will likely fail because the pepper hasn't been cached yet
+      // but needs to happen for the edge case that a user has a cached pepper
+      // and Komenci is down
+      yield call(fetchVerifiedMtw, contractKit, walletAddress, e164Number)
```
asaj: I don't understand why this call is necessary here

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    const ownPepper = pepperCache[e164Number]
+
+    Logger.debug(
+      TAG,
+      '@fetchSessionState Session status:',
+      JSON.stringify(sessionStatusResult.result)
+    )
+
+    sessionActive = true
+    // Somtimes `metaTxWalletAddress` is returned as undefined for an active session.
```
asaj: typo. Also, in which cases will we have an unverifiedMtwAddress cached but have metaTxWalletAddress be undefined?

Separately, why do we have the notion of unverified vs verified metatxwallet address? Presumably we would just have one metatxwalletaddress in our state (which ofc could be undefined if we don't have one yet), and the verification state of that address would be stored elsewhere in the state data structure.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    } = sessionStatusResult.result
+    const ownPepper = pepperCache[e164Number]
+
+    Logger.debug(
+      TAG,
+      '@fetchSessionState Session status:',
+      JSON.stringify(sessionStatusResult.result)
+    )
+
+    sessionActive = true
```
asaj: Is this redundant? We set it to true above

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      Logger.debug(TAG, 'Unable to check it MTW is verified on first attempt')
+    }
+    // Throws error if Komenci is not ready for user, otherwise it updates state
+    yield call(fetchKomenciReadiness, komenciKit)
+
+    // Throws error if unable to retreive phone hash, otherwise it updates state
+    yield call(fetchPhoneHashDetailsFromCache, e164Number)
+
+    // Updates state with sessionStatus and the mtwAddress associated with the session
+    yield call(fetchSessionState, komenciKit, e164Number)
```
asaj: nit: `fetchKomenciSessionState`

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+
+    // Checks if a MTW was verified in a previous attempt and updates state
+    yield call(fetchVerifiedMtw, contractKit, walletAddress, e164Number)
```
asaj, line 155: Why do we need to to this if we already have a mtw address that we were able to recover from our cached state? These view calls are expensive

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      // but needs to happen for the edge case that a user has a cached pepper
+      // and Komenci is down
+      yield call(fetchVerifiedMtw, contractKit, walletAddress, e164Number)
+    } catch (error) {
+      Logger.debug(TAG, 'Unable to check it MTW is verified on first attempt')
+    }
+    // Throws error if Komenci is not ready for user, otherwise it updates state
+    yield call(fetchKomenciReadiness, komenciKit)
+
+    // Throws error if unable to retreive phone hash, otherwise it updates state
```
asaj, line 148: IIUC this throws if the pepper isn't cached? That seems like the wrong behavior if we're going to fetch it from Komenci later

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    yield call(fetchPhoneHashDetailsFromCache, e164Number)
+
+    // Updates state with sessionStatus and the mtwAddress associated with the session
+    yield call(fetchSessionState, komenciKit, e164Number)
+
+    // Checks if a MTW was verified in a previous attempt and updates state
+    yield call(fetchVerifiedMtw, contractKit, walletAddress, e164Number)
+
+    // Updates state with any attestation progress that's been made
+    const status: AttestationsStatus = yield call(fetchAttestationStatus, contractKit)
```
asaj, line 158: We should be able to skip this if we weren't able to find a meta transaction wallet address, either in our cache or in attestations contract. These view calls are also expensive

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    Logger.debug(TAG, 'Verification timed out')
+    yield put(showError(ErrorMessages.VERIFICATION_TIMEOUT))
+    yield put(feelessSetVerificationStatus(VerificationStatus.Failed))
+  }
+  Logger.debug(TAG, 'Done verification')
+
+  yield put(refreshAllBalances())
+}
+
+export function* feelessRestartableVerification(initialWithoutRevealing: boolean) {
```
asaj, line 201: What does `withoutRevealing` mean?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    let feelessVerificationState: FeelessVerificationState = yield select(
+      feelessVerificationStateSelector
+    )
+
+    const komenciKit = new KomenciKit(contractKit, walletAddress, {
+      url: KOMENCI_URL,
+      token: feelessVerificationState.komenci.sessionToken,
+    })
+
+    // Start by checking again to make sure Komenci is ready. Throws error if not
```
asaj: Didn't we just do this?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+  Logger.debug(TAG, '@startOrResumeKomenciSession', 'Starting session')
+  // Fetch session state to make sure we have the most up-to-date session info
+  yield call(fetchSessionState, komenciKit, e164Number)
+  const feelessVerificationState: FeelessVerificationState = yield select(
+    feelessVerificationStateSelector
+  )
+
+  const { sessionActive, captchaToken, sessionToken } = feelessVerificationState.komenci
+
+  // If there isn't an active session, start one. Need to include `sessionActive`
```
asaj, line 719: nit; Why did we cache sessionActive in the state earlier, rather than just interpret our remaining quota here to decide if we need to start a new session?

packages/mobile/src/identity/feelessVerification.ts
asaj: In fact, I don't understand why we need to do this at all? Why don't we cache it in the feeless verification state? If it's cached, we'll have it, if not, we won't. I don't see why we need to fetch it here

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      ...feelessVerificationState,
+      actionableAttestations,
+      status,
+    })
+  )
+
+  return status
+}
+
+function* startOrResumeKomenciSession(komenciKit: KomenciKit, e164Number: string) {
```
asaj, line 709: nit: `maybeStartKomenciSession`?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    // Start by checking again to make sure Komenci is ready. Throws error if not
+    yield call(fetchKomenciReadiness, komenciKit)
+
+    // There should be no instances where the first param is truethy but the second isn't.
+    // Mainly including the second param to satisfy typescript
+    if (
+      !feelessVerificationState.status.isVerified ||
+      !feelessVerificationState.komenci.unverifiedMtwAddress
+    ) {
+      // Start or resume a Komenci session and update state. Throws error if unable to do so
```
asaj: nit: // Start a new Komenci session if one doesn't exist or we've run out of quota

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      yield call(startOrResumeKomenciSession, komenciKit, e164Number)
+
+      // Adds phoneHash into verification state, will ping Komenci for it only if needed.
+      // Throws an error if unable to get phoneHash
+      yield call(fetchPhoneHashDetails, komenciKit, e164Number)
+
+      const startingPepperQuota = feelessVerificationState.komenci.pepperQuotaRemaining
+      // Now that we are guarnateed to have the phoneHash, check again to see if the
```
asaj, line 268: This feels like it should belong in `feelessFetchVerificationState`...

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+        yield call(endFeelessVerification)
+        return true
+      }
+
+      yield call(fetchOrDeployMtw, contractKit, komenciKit, walletAddress, e164Number)
+
+      // Registering the DEK and wallet before beginning verification to guarantee that every
+      // verified MTW address (i.e., accountAddress) has an EOA (i.e., walletAddress)
+      // registered to it. If that's not guaranteed, users are at risk of sending to MTWs
+      yield call(feelessDekAndWalletRegistration, komenciKit, walletAddress)
```
asaj, line 284: Are we blocking on this transaction succeeding? IMO there's no reason no to batch this with the attestation request. Would save us 5-10s if this call is blocking..

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+
+      // ====== Below this point feeless verification has the same structure as the original ====== //
+
+      feelessVerificationState = yield select(feelessVerificationStateSelector)
+      const { status, actionableAttestations, phoneHashDetails, komenci } = feelessVerificationState
+      const { phoneHash } = phoneHashDetails
+      const { unverifiedMtwAddress } = komenci
+
+      if (!unverifiedMtwAddress) {
+        throw Error('MTW not yet deploy. Should never happen')
```
asaj: nit: deployed

bowd:
> > my first feeling is that it's a bit too complex compared to my intuition of what should be going on. But it also feels like we are forced into this complexity by the existing house of cards.
> 
> Would it simplify things to model verification as a state machine?

Yeah I wanted to echo your  sentiment on state machines but I don't know how much of a departure that is from what's currently the pattern in Valora and I also don't know enough about how we manage state in Valora to have certainty in this. But indeed intuitively it feels like more rigid states and transitions could simplify things.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      )
+
+      if (!deployWalletResult.ok) {
+        Logger.debug(TAG, '@fetchOrDeployMtw', 'Unable to deploy MTW')
+        throw deployWalletResult.error
+      }
+
+      unverifiedMtwAddress = deployWalletResult.result
+    } catch (error) {
+      // Fetch session state now that we are sure to have a MTW
```
asaj: Why are we sure of this?

asaj:
Btw I understand that we're going with the "quick and dirty" way for now, with a follow-up refactor, so please feel free to punt some of these comments to that refactor

packages/contractkit/src/governance/proxy.ts
tarikbellamine: Yes. @yorhodes is merging a PR that fixes this and pushed a commit to my branch in the meantime.

yorhodes: its merged I think

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: I agree. Unfortunately, a larger structural change to verification felt too risky given the timeline and lack of resources but is hopefully something we can prioritize shortly after GA.

You can read more about my thought process here: https://celo-org.slack.com/archives/CL7BVQPHB/p1603991866447600

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: If the `fetchKomenciReadiness` check is first and Komenci is down, then this check will never happen. How else can we check if a user is already verified in the event of a Komenci outage?

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: > IIUC this throws if the pepper isn't cached? That seems like the wrong behavior if we're going to fetch it from Komenci later

Why does it seem like the wrong behavior? This is essentially just a helper function that checks if a user already has their own pepper cached. If they don't, and therefore don't know their own phone hash, we bail on the rest of the checks because we don't have the identifier needed to update any progress the user may or may not have.

> In fact, I don't understand why we need to do this at all? Why don't we cache it in the feeless verification state? If it's cached, we'll have it, if not, we won't. I don't see why we need to fetch it here

What if the user has their own pepper already from attempting a non-feeless verification? Why waste time and $$$ asking Komenci for something the user may already have?

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: At this point, we would only have a `unverifiedMtwAddress` if the user had an active Komenci session they were resuming. What about users that already verified but had their Komenci session expire?

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: See above comment. Not being able to recover a potentially `unverifiedMtwAddress` from an active Komenci session does not preclude a wallet being verified in a previous, inactive Komenci session or from a previous, non-feeless verification attempt.

IMO this check should happen as long as their is a `phoneHash` to check, which is the case right now.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: If this is `true` we will not do any automatic reveals (i.e., send SMS's) for the user as part of the flow. Instead, we will skip them ahead to a screen where they can input any verification codes they may have received, which also has a button at the bottom to request most SMS's.

This basically exists to allow users who have received > 0 codes to input them first before requesting more. Think of the extreme case where a user bails but later receives all 3 SMS's. When they come back to the flow, we first give them the opportunity to share their progress with us before we do anything else.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yeah, I'll remove this.

I was thinking about the instance where there was a long wait between the call that happens in `feelessFetchVerificationState` and the actual verification flow, which would mean the user left the "start verification" screen open for a long time without actually pressing "start".

Thinking more about it, it serves no purpose because the next call would just fail anyways if Komenci is down so there is no UX advantage here (as opposed to the previous one where we could divert users to non-feeless verification is Komenci was down).

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Are you suggesting we move the pepper query into `feelessVerificationState`? I agree that that feels like a better home for it from a "separation of concerns" perspective but practically speaking I think it makes less sense given how these functions are currently tied into the UI.

The user is waiting for `feelessFetchVerificationState` to finish before they can even press "start" on the "verification start" screen. For that reason I am reluctant to add a non-state related Komenci call that will further slow things down. While it may look expensive to be repeating function calls, remember that many of these will quickly bail out if the right conditions aren't being met.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: We are blocking on this transaction succeeding. This is what I mentioned in that thread with @nategraf about how I am attempting to guarantee that every verified MTW has a DEK and `walletAddress` registered to it in the Accounts contract.

I agree that there is an optimization opportunity around batching this tx, as there are with other transactions later in the flow (e.g., approve + request tx for requests). I am biased to keeping it as is for now and tackle optimizations within the context of the larger remodeling we'd like to do of this flow.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yep good catch

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: > Also, in which cases will we have an unverifiedMtwAddress cached but have metaTxWalletAddress be undefined?

If a user starts Session1, has a MTW deployed for them during Session1, then needs to start Session2 because they've run out of quota on Session1. In this case, the `mtwAddress` associated with Session2 is `undefined`.

It is important that we keep using the `mtwAddress` created in Session1 because attempting to deploy a new MTW in Session2 while using the same EOA as Session1 will cause a Komenci failure. @bowd can provide more background on this specific behavior.


> Separately, why do we have the notion of unverified vs verified metatxwallet address? 

I think my naming may be causing some confusion. I presume any mtw address to be an `unverifiedMtwAddress` until proven otherwise. Once we know an address to be verified, the flow immediately ends so there isn't really any notion of operating on a verified MTW address in this flow.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yup you're correct but it's only a temporary fix (as in I won't merge with it like this).

See the `TODO` up top. Once Bogdan's [bytecode PR](https://github.com/celo-org/celo-monorepo/pull/5724) merges then I will be removing this try/catch block.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Because I was tired :sweat_smile:

Good catch!

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yeah all verification errors that cause the flow to throw logged by the `VerificationEvents.verification_error` event (see `feelessStartVerification` generator)

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yeah all verification errors that cause the flow to throw are logged under the `VerificationEvents.verification_error` event (see `feelessStartVerification` generator)

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Then `verifiedMtwAddress` will be `undefined` and we will throw an error on the next line.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yeah that's a good call out. I will break this whole status object into a `VERIFIED_ATTESTATION_STATUS` constant with a note that it will need to change if our verification criteria ever does.

For now I don't think any change is needed because the `total` property is irrelevant for considerations of verification.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: I like my naming better :grin:

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Because there is value in knowing the state of your session before you're ready to start a new one.

For example, we have the first `fetchKomenciSessionState` in the `feelessFetchVerificationState` generator, which happens immediately when the user lands on the "start verification screen". If there is not one currently active, then the user must complete a captcha before we can start a new one for them. Given the need for user action (completing a captcha) between when a session is known to be inactive and when we are able to start a new one, it made most sense for me to keep these two functions separate.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Poorly worded comment on my part. Changed it to: `Fetch session state now that there should be no cases where we don't have a  MTW`.

As you can see below, I am checking if `unverifiedMtwAddress` is still undefined and throwing a fatal error if that's the case.

packages/komencikit/src/verifyWallet.ts
tarikbellamine: Sweet! Please ping me when it's merged

packages/mobile/src/identity/actions.ts
```diff
@@ -251,11 +342,23 @@ export const resetVerification = (): ResetVerificationAction => ({
...
+export const feelessResetVerification = (): FeelessResetVerificationAction => ({
+  type: Actions.FEELESS_RESET_VERIFICATION,
+})
+
 export const setVerificationStatus = (status: VerificationStatus): SetVerificationStatusAction => ({
   type: Actions.SET_VERIFICATION_STATUS,
   status,
 })
 
+// TODO: delete this if we dont end up using it
```
nategraf: It looks like this is indeed in use, so I suppose this comment is stale?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+import { setMtwAddress } from 'src/web3/actions'
+import { getContractKit } from 'src/web3/contracts'
+import { registerWalletAndDekViaKomenci } from 'src/web3/dataEncryptionKey'
+import { getConnectedAccount, getConnectedUnlockedAccount } from 'src/web3/saga'
+import { TransactionReceipt } from 'web3-eth'
+
+const TAG = 'identity/feelessVerification'
+
+const KOMENCI_URL = 'https://komenci.celo-networks-dev.org'
+// const KOMENCI_URL = 'http://192.168.86.33:3000'
```
nategraf: Cleanup local constants
```suggestion
```

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      call(getContractKit),
+      call(getConnectedUnlockedAccount),
+      select(feelessVerificationStateSelector),
+      select(e164NumberSelector),
+      put(feelessSetVerificationStatus(VerificationStatus.GettingStatus)),
+    ])
+
+    const komenciKit = new KomenciKit(contractKit, walletAddress, {
+      url: KOMENCI_URL,
+      token: feelessVerificationState.komenci.sessionToken,
```
nategraf, line 128: We may want to check that the token is set (and is valid).

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    })
+
+    try {
+      // Checks if a MTW was verified in a previous attempt and updates state.
+      // This check will likely fail because the pepper hasn't been cached yet
+      // but needs to happen for the edge case that a user has a cached pepper
+      // and Komenci is down
+      yield call(fetchVerifiedMtw, contractKit, walletAddress, e164Number)
+    } catch (error) {
+      Logger.debug(TAG, 'Unable to check it MTW is verified on first attempt')
```
nategraf: Typo "it MTW is" -> "if MTW is"
```suggestion
      Logger.debug(TAG, 'Unable to check if MTW is verified on first attempt')
```

packages/mobile/src/identity/feelessVerification.ts
nategraf: One the things I was confused about here was why, if the `fetchVerifiedMtw` call above failed, this one would ever succeed? IIUC correctly we don't fetch the pepper in this flow, so if it wasn't cached before, we still will not have it now. More generally, I understand the case of why we want to call `fetchVerifiedMtw` in this flow, but don't understand why we do it twice. It would be better to find some way to check at most once.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+
+    // Checks if a MTW was verified in a previous attempt and updates state
+    yield call(fetchVerifiedMtw, contractKit, walletAddress, e164Number)
+
+    // Updates state with any attestation progress that's been made
+    const status: AttestationsStatus = yield call(fetchAttestationStatus, contractKit)
+    ValoraAnalytics.track(VerificationEvents.verification_fetch_status_complete, {
+      ...status,
+    })
+    yield put(feelessSetVerificationStatus(VerificationStatus.Stopped))
```
nategraf: Recommend moving this call to a `finally` block.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+    return error.message
+  } finally {
+    receiveMessageTask?.cancel()
+    if (Platform.OS === 'android') {
+      autoRetrievalTask?.cancel()
+    }
+  }
+}
+
+function* fetchKomenciReadiness(komenciKit: KomenciKit) {
```
nategraf, line 427: IIUC this function is supposed to update the state about whether Komenci is available. Should it update the state to false if the service check fails?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      ...feelessVerificationState,
+      komenci: {
+        ...feelessVerificationState.komenci,
+        serviceAvailable: true,
+      },
+    })
+  )
+}
+
+function* fetchPhoneHashDetailsFromCache(e164Number: string) {
```
nategraf, line 477: Could this function be refactored as a selector? It seems like its return value is entirely dependent on that state tree, so it seems odd that we call a function to     take state from one part of the tree to another.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+
+  let sessionActive = true
+  let {
+    requestQuotaRemaining,
+    pepperQuotaRemaining,
+    unverifiedMtwAddress,
+  } = feelessVerificationState.komenci
+
+  // An inactive session is not fatal, it just means we will need to start one
+  if (!sessionStatusResult.ok) {
```
nategraf, line 528: We should check for inactive state vs other error types here.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+
+    // No pepper quota remaining is only bad if it's not already cached. Given Komenci will fetch
+    // a pepper for you once, a session could be invalid due to the pepper condition if a user
+    // fetched their pepper once this session then uninstalled without starting a new session
+    if (
+      (!ownPepper && !distributedBlindedPepper) ||
+      !requestSubsidisedAttestation ||
+      !submitMetaTransaction
+    ) {
+      sessionActive = false
```
nategraf, line 562: We should log some kind of error here.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+  Logger.debug(TAG, '@reFetchVerifiedMtw', 'Starting fetch')
+  const feelessVerificationState: FeelessVerificationState = yield select(
+    feelessVerificationStateSelector
+  )
+
+  const { pepperQuotaRemaining } = feelessVerificationState.komenci
+  // If Komenci was used to fetch their pepper, check if the they have a verified
+  // account. If user already had their pepper, then check would have happened
+  // in `feelessFetchVerificationState`
+  if (startingPepperQuota > pepperQuotaRemaining) {
```
nategraf: Instead of assuming a function was called previously, can we record somthing in the state to let us know for sure? This check just seems fragile.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+
+  const { sessionActive, captchaToken, sessionToken } = feelessVerificationState.komenci
+
+  // If there isn't an active session, start one. Need to include `sessionActive`
+  // because that's the only way we'll know if Komenci session is active but
+  // quota is used
+  if (!sessionActive || !sessionToken.length) {
+    // Should never get here without a captcha token
+    if (!captchaToken.length) {
+      Logger.debug(TAG, '@startOrResumeKomenciSession', 'No captcha found')
```
nategraf: We should probably log at error level here.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+        CURRENT_MTW_IMPLEMENTATION_ADDRESS
+      )
+
+      if (!deployWalletResult.ok) {
+        Logger.debug(TAG, '@fetchOrDeployMtw', 'Unable to deploy MTW')
+        throw deployWalletResult.error
+      }
+
+      unverifiedMtwAddress = deployWalletResult.result
+    } catch (error) {
```
nategraf, line 825: We should be able to check for the error type to make this really was an error because the MTW was already deployed.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+      mtwAddress,
+      numAttestationsRequestsNeeded
+    )
+
+    if (!approveTxResult.ok) {
+      Logger.debug(TAG, '@feelessRequestAttestations', 'Failed approve tx')
+      throw approveTxResult.error
+    }
+
+    ValoraAnalytics.track(VerificationEvents.verification_request_attestation_approve_tx_sent)
```
nategraf, line 962: We should note in these events if `komenci` is being used

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1061 @@
...
+  }
+
+  ValoraAnalytics.track(VerificationEvents.verification_reveal_attestation_complete, { issuer })
+
+  yield put(feelessCompleteAttestationCode(code))
+  Logger.debug(TAG + '@feelessCompleteAttestation', `Attestation for issuer ${issuer} completed`)
+}
+
+// Get the code from the store if it's already there, otherwise wait for it
+function* feelessWaitForAttestationCode(issuer: string) {
```
nategraf, line 1052: It seems odd to me that this one is separate for feeless. Could this simply take the `code` as an input instead of from the state and share the same function and acton with the other version of this?

packages/mobile/src/identity/feelessVerificationErrors.ts
```diff
@@ -0,0 +1,107 @@
...
+  const errorString = error.toString()
+
+  if (
+    Object.keys(TxErrorTypes).includes(errorString) ||
+    Object.keys(KomenciKitErrorTypes).includes(errorString) ||
+    Object.keys(FetchErrorTypes).includes(errorString) ||
+    Object.keys(FeelessVerificationErrors).includes(errorString)
+  ) {
+    unexpectedKomenciError = true
+  }
```
nategraf: Can you add some comment to explain what this check is doing? It's rather opaque

packages/mobile/src/identity/reducer.ts
```diff
@@ -378,11 +534,32 @@ const completeCodeReducer = (state: State, numCompleteAttestations: number) => {
...
 
+const feelessCompleteCodeReducer = (state: State, feelessNumCompleteAttestations: number) => {
+  const { feelessAttestationCodes, feelessAcceptedAttestationCodes } = state
+  // Ensure numCompleteAttestations many codes are filled
+  for (let i = 0; i < feelessNumCompleteAttestations; i++) {
+    feelessAttestationCodes[i] = feelessAcceptedAttestationCodes[i] || {
+      code: ATTESTATION_CODE_PLACEHOLDER,
+      issuer: ATTESTATION_ISSUER_PLACEHOLDER,
+    }
+  }
```
nategraf: This is a state mutation. Two ways to avoid this would be to 1. Rewrite this as a `map` call or 2. copy `feelessAttestationCodes` before modifying it.

packages/mobile/src/identity/reducer.ts
```diff
@@ -437,9 +617,16 @@ export const isVerificationStateExpiredSelector = (state: RootState) => {
...
-    !lastRevealAttempt ||
-    timeDeltaInSeconds(Date.now(), lastRevealAttempt) > ATTESTATION_REVEAL_TIMEOUT_SECONDS
-  )
+export const isFeelessVerificationStateExpiredSelector = (state: RootState) => {
+  const { lastFetch } = state.identity.feelessVerificationState
+  return !lastFetch || timeDeltaInSeconds(Date.now(), lastFetch) > VERIFICATION_STATE_EXPIRY_SECONDS
+}
+
+// Use this only as part of the notification display logic
+export const tryFeelessOnboardingSelector = ({
```
nategraf, line 653: Can you clarify this a bit more? Why is it unsuitable elsewhere?

packages/mobile/src/identity/verification.ts
```diff
@@ -100,13 +107,15 @@ export function* fetchVerificationState() {
     })
     if (timeout) {
       Logger.debug(TAG, '@fetchVerificationState', 'Token balances is null or undefined')
+      yield put(setVerificationStatus(VerificationStatus.Stopped))
```
nategraf: I would recommend putting this into a `finally` block to avoid needing it at every exit point in this function, and to handle `saga` cancellation.

packages/mobile/src/identity/verification.ts
```diff
@@ -319,6 +330,8 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
       }
 
       yield put(setVerificationStatus(VerificationStatus.CompletingAttestations))
+      // NOTE: I think we want the DEK registration out of the race because
```
nategraf: I definitely agree with this. Related to the edge case mentioned in #5622, wallet address registration should be a requirement before the attestation is complete. My recommendation would be to move this such that it occurs in parallel with the verification flow, and check here to make sure it is complete before finalizing the attestation,

packages/mobile/src/invite/saga.ts
```diff
@@ -387,7 +387,8 @@ export function* skipInvite() {
     yield put(setHasSeenVerificationNux(true))
     Logger.debug(TAG + '@skipInvite', 'Done skipping invite')
     ValoraAnalytics.track(OnboardingEvents.invite_redeem_skip_complete)
-    navigateHome()
+    // navigateHome()
+    navigate(Screens.VerificationEducationScreen)
```
nategraf, line 409: Should this be gated on the Komenci feature flag? (i.e. if they skip invite and aren't using Komenci, they should be sent to the Home screen?)

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -118,23 +189,129 @@ export function* registerAccountDek(account: string) {
...
 
+// Unlike normal DEK registration, registration via Komenci should be considered fatal. If there
+// is no on-chain mapping of accountAddresss => walletAddress, then senders will erroneously
+// send to MTW instead of EOA. A no-op if registration has already been done
+export function* registerWalletAndDekViaKomenci(
+  komenciKit: KomenciKit,
+  accountAddress: string,
+  walletAddress: string
+) {
+  ValoraAnalytics.track(OnboardingEvents.account_dek_register_start)
```
nategraf, line 204: Adding a property to this event noting that this is a Komenci assisted registration would be helpful

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -118,23 +189,129 @@ export function* registerAccountDek(account: string) {
...
+
+  Logger.debug(
+    `${TAG}@registerAccountDekViaKomenci`,
+    'Setting wallet address and public data encryption key'
+  )
+
+  // Don't think we need to unlock the account anymore because we do
+  // it at the start of the feeless verification flow
+  // yield call(getConnectedUnlockedAccount)
+  ValoraAnalytics.track(OnboardingEvents.account_dek_register_account_unlocked)
```
nategraf: We should ensure the account is unlocked here because if the unlock timeout has passed (e.g. the use backgrounded the app during the verification process and came back after > 10 minutes) then we either need to unlock here or bail.

packages/mobile/test/values.ts
```diff
@@ -19,9 +19,12 @@ import {
   RecipientWithQrCode,
 } from 'src/recipients/recipient'
 
+export const nullAddress = '0x0'
```
nategraf: Null address is usually 20 0 bytes. Although you could compare addresses as integers (i.e. discount any leading zeros) most applications don't tend to do that.

nategraf (commented):
Alright, I gave it all a look! I will mention that I had a bit of hard time understanding the feeless verification flow because so many of the functions would modify global state, but return nothing 😕 . I understand that we don't have time to restructure everything though.

* The `revokeVerification` saga should be updated to also revoke MTW attestations. (It does not need to use Komenci) It's used in the user-testing flow. https://github.com/celo-org/celo-monorepo/blob/cbc34f20c9ac7876c3c8d22dec94d040a846b6b1/packages/mobile/src/identity/revoke.ts#L22
* In my brief testing with Komenci, it seemed that the service may end up deploying more than one MTW per signer. IIUC, then this means it's possible to get partway through a verification flow (e.g. 2/3 attestations completed) and fail, then get a new MTW on the next session. In this case, the state variables in the redux tree will indicate that we have 2/3 attestations completed when we actually have 0 on the new MTW wallet address. Two options to button up this edge case would be 1. Ensure Komenci always returns the same address for a given signer or 2. Ensure we clear attestation state if/when we receive a new MTW.
* Maybe I missed it, but I couldn't work out whether the captcha token can expire, and if it does, what happens. Is that case handled gracefully?

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yeah, you are right. I can add logic so that this check only happens once. If Komenci is live, it will happen at the end of `feelessFetchVerificationState`. If Komenci is down, it will check it once before bailing out on the rest of the checks. But there is no reason why it should ever happen twice.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: @bowd can correct me if I am wrong but I don't think that's necessary because an empty or expired session token will just be swallowed without error. We make no assumptions about validity and ping Komenci for session status as the source of truth

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Will put a logger.debug for now and make sure to instrument this with analytics when I have time to do a pass for that

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yeah you're right. That's a good suggestion I'll do that. It will also allow me to delete those quota counters that I never really ended up needing due to the `checkSession` endpoint we ended up implementing recently.

packages/mobile/src/identity/feelessVerificationErrors.ts
tarikbellamine: Gotcha, I added this comment:  `Any errors that of these types are unexpected Komenci errors and are likely indicative of a service failure that needs remediation`. 

I also edited the comment explaining the function to read:
`If an error occurs during the feeless verification flow that is "unexpected" and likely due to a Komenci service failure, add the timestamp of it's occurrence to state. If the user encounters more errors than we feel comfortable with during a given window, we won't allow them to attempt feeless verification until a certain amount of time has passed in order to give cLabs time to remediate the issue`.

Make sense?

packages/mobile/src/identity/reducer.ts
tarikbellamine: That's an old comment I forgot to delete, thanks for flagging. We can use this elsewhere.

packages/mobile/src/web3/dataEncryptionKey.ts
tarikbellamine: Yeah makes sense. After looking at the function, I now see that it does nothing if the account is already unlocked so I guess there is no harm in including it.

packages/mobile/src/identity/feelessVerification.ts
asaj: Batching the approve and request tx *should* be a requirement enforced by Komenci... I agree that batching the DEK/wallet registration is a future optimization

asaj: If Komenci isn't batching them then we have a big problem, but I would be very surprised if it's not

packages/mobile/src/invite/saga.ts
tarikbellamine: Yeah this is just a placeholder to allow us to test feeless verification. I will revert this before merging; my subsequent PR for escrow transactions will include the appropriate logic here.

packages/mobile/src/identity/reducer.ts
tarikbellamine: Great catch. I copied this from elsewhere without checking it. Will correct them in both places.

packages/mobile/src/identity/verification.ts
tarikbellamine: I've moved it out of the race condition but don't believe any further changes are necessary for now. I think the edge case in #5622 pertains only to feeless verification. If DEK/wallet registration fails for non-feeless verification, then our new logic will correctly assume that `accountAddress === walletAddress`.

nategraf:
One thing I just noticed while going through the app onboarding is that when Koemenci is down, it still takes the user to the verification education screen, showing a "Skip for now" button instead of "Start". It would be better to skip the screen entirely if possible.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: The reason why it's not modeled as a selector is because we want it to throw an error in some instances. Of course, we could follow the selector call with a throw condition if `selectorResult === null || undefined` but that's basically all we are doing in this generator anyways

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: That would be more efficient, though any "fatal" failure that occurred here would also occur in the else block immediately following this.

@bowd what type of error gets returned when there is simply no session associated with a given `sessionToken`?

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: I don't think the error type is properly categorized by Komenci/KomenciKit but I made this check more explicit by comparing `storedAddress` with `deployedAddress`.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: That's probably true. Given our desire to better integrate feeless and non-feeless verification shortly, I chose to prioritize a clear separation of state & actions over code maintainability and dryness. In places where I thought it was safe to reuse code, I went for it. When I was unsure, I defaulted to the "safe" choice of duplicating because I believe testing the use case of switching between feeless and non-feeless will be difficult.

tarikbellamine (commented):
> I will mention that I had a bit of hard time understanding the feeless verification flow because so many of the functions would modify global state, but return nothing 

Dang, I'm sorry to hear this was confusing. I first tried to keep the functions pure and only reassign variables in the main flow but it quickly got unwieldy and difficult to keep track of all the resassignments. I was hoping that following a model where each function operated on the latest state then updated it + keeping the state updates MECE (i.e., no 2 functions can update the same state variable) would be less confusing than the alternative.

> The revokeVerification saga should be updated to also revoke MTW attestations.

Good call out. Had it noted but forgot to do it so thanks for the reminder.

> In my brief testing with Komenci, it seemed that the service may end up deploying more than one MTW per signer.....Two options to button up this edge case would be 1. Ensure Komenci always returns the same address for a given signer or 2. Ensure we clear attestation state if/when we receive a new MTW.

My understanding was that Komenci would not allow this to happen. In the spirit of making Valora resilient to Komenci changes, I went ahead and did #2 as I don't see any downside.

> Maybe I missed it, but I couldn't work out whether the captcha token can expire, and if it does, what happens.

In the event the captcha token expires before it's used to start a Komenci session, the session would simply fail to start. Regardless of why sessions fail or expire, we will force them through the captcha flow again.

> One thing I just noticed while going through the app onboarding is that when Koemenci is down, it still takes the user to the verification education screen, showing a "Skip for now" button instead of "Start". It would be better to skip the screen entirely if possible.

Right now the rule is if they receive more than 2 Komenci failures (including service being down) within 3 hours, that banner will disappear. Unfortunately, the check to determine if the service is down happens on the VerificationEducation screen. Longer term we can think about placing that availability check elsewhere to be more proactive about hiding the banner.

packages/mobile/src/identity/feelessVerification.ts
nategraf: Ok, this makes sense! I see now that it would simply fail a bit further down when it checks the session status 👍

packages/mobile/src/identity/feelessVerification.ts
nategraf: Ok. Nothing about this looks incorrect, so not really worth debating now 👍

packages/mobile/src/identity/feelessVerification.ts
nategraf: That's true. If there is a "fatal" error in checking the state there is not much harm in trying to start a new session anyway. Worst case scenario is that the error is a timeout due to unresponsive service, in which case we either would have failed at the service readiness check or will fail to start the new session, only delaying feedback to the user, but not causing a flow to fail that would have otherwise succeeded.

packages/mobile/src/identity/verification.ts
```diff
@@ -319,17 +330,24 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
...
+      // we want that to happen no matter what
```
nategraf: Suggest removing this comment based on the discussion above.
```suggestion
```

packages/mobile/src/identity/verification.ts
```diff
@@ -319,17 +330,24 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
...
+          phoneHashDetails,
+          attestations
+        ),
         // This is needed, because we can have more actionableAttestations than NUM_ATTESTATIONS_REQUIRED
         requiredAttestationsCompleted: call(requiredAttestationsCompleted),
       })
 
+      // Set acccount and data encryption key in Accounts contract
+      // This is done in other places too, intentionally keeping it for more coverage
+      yield call(registerAccountDek, account)
```
nategraf: Suggest making this use `spawn` because we don't want to wait for it (we are not dependent on the results) and we will not raise an error if it fails.
```suggestion
      yield spawn(registerAccountDek, account)
```

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1088 @@
...
+    try {
+      // Throws error if Komenci is not ready for user, otherwise it updates state
+      yield call(fetchKomenciReadiness, komenciKit)
+    } catch (error) {
+      // Checks if a MTW was verified in a previous attempt and updates state.
+      // This check will likely fail because the pepper hasn't been cached yet
+      // but needs to happen for the edge case that a user has a cached pepper
+      // and Komenci is down
+      try {
+        yield call(fetchVerifiedMtw, contractKit, walletAddress, e164Number)
```
nategraf, line 140: IIUC, we want to return after this call if it yields a verified MTW address, because the state will have been set to verified. Is that correct? Otherwise, this will throw below no matter what this call does and, as a result, abort the flow.

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -0,0 +1,1088 @@
...
+
+      if (!deployWalletResult.ok) {
+        Logger.debug(TAG, '@fetchOrDeployMtw', 'Unable to deploy MTW')
+        throw deployWalletResult.error
+      }
+
+      deployedUnverifiedMtwAddress = deployWalletResult.result
+    } catch (error) {
+      // Fetch session state now that there should be no cases where we don't have a  MTW
+      yield call(fetchKomenciSessionState, komenciKit, e164Number)
```
nategraf, line 827: IIUC, this can result in receiving an MTW address from Komenci, in which case we want to continue without error, using that address. I'm guessing that was unintentionally removed and the `deployedUnverifiedMtwAddress` should be updated here.

packages/mobile/src/identity/verification.ts
```diff
@@ -100,13 +107,15 @@ export function* fetchVerificationState() {
...
       Logger.debug(TAG, '@fetchVerificationState', 'Token balances is null or undefined')
+      yield put(setVerificationStatus(VerificationStatus.Stopped))
       return
     }
     const isBalanceSufficientForSigRetrieval = yield select(
       isBalanceSufficientForSigRetrievalSelector
     )
     if (!isBalanceSufficientForSigRetrieval) {
       Logger.debug(TAG, '@fetchVerificationState', 'Insufficient balance for sig retrieval')
+      yield put(setVerificationStatus(VerificationStatus.Stopped))
```
nategraf: Because we have the `fianally` block, I don't think we need these here.
```suggestion
      return
    }
    const isBalanceSufficientForSigRetrieval = yield select(
      isBalanceSufficientForSigRetrievalSelector
    )
    if (!isBalanceSufficientForSigRetrieval) {
      Logger.debug(TAG, '@fetchVerificationState', 'Insufficient balance for sig retrieval')
```

packages/mobile/src/web3/actions.ts
```diff
@@ -72,6 +79,13 @@ export const setAccount = (address: string): SetAccountAction => {
   }
 }
 
+export const setMtwAddress = (address: string | null): SetMtwAddressAction => {
+  return {
+    type: Actions.SET_MTW_ADDRESS,
+    address: address?.toLowerCase() ?? address,
```
nategraf, line 85: `address?.toLowerCase()` will return `null` if address is `null` so the `??` should be unnecessary. 
```suggestion
    address: address?.toLowerCase(),
```

nategraf:
I believe I've spotted a couple of places where the latest changes had unintended side effects. Otherwise, I don't see any issues. Thanks for addressing all my comments!

nategraf (approved):

packages/mobile/src/web3/actions.ts
tarikbellamine: I think it actually returns `undefined` in the failure case so will leave it as is to maintain continuity of using `null` to symbolize no MTW

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yeah great catch. I think that's exactly what we want to do.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Spot on. Thanks!

packages/mobile/src/identity/verification.ts
```diff
@@ -320,16 +354,21 @@ export function* doVerificationFlow(withoutRevealing: boolean = false) {
...
+          phoneHashDetails,
+          attestations
+        ),
         // This is needed, because we can have more actionableAttestations than NUM_ATTESTATIONS_REQUIRED
         requiredAttestationsCompleted: call(requiredAttestationsCompleted),
       })
 
+      // Set acccount and data encryption key in Accounts contract
+      // This is done in other places too, intentionally keeping it for more coverage
+      yield spawn(registerAccountDek, account)
```
i1skn, line 370: why move `registerAccountDek` out of `race` above?

tarikbellamine: IIUC we want DEK registration no matter what happens first. Previously DEK registration would only be happening if `actionableAttestationCompleted` won the race but not if `requiredAttestationsCompleted` was first. This way, DEK registration is attempted no matter what.

i1skn: that make sense, what do you think of the following then: `all(registerAccountDek, race(actionableAttestationCompleted, requiredAttestationsCompleted))`. This way we run `registerAccountDek` in parallel and guarantee that it was called. Timewise it should not be too different, but we avoid running detach task.

tarikbellamine: Sounds good to me, will do

Lss-Ankit:
Hi @tarikbellamine Verified task on latest test flight build v1.5.5 (34) and Android play store internal build v1.5.5 (1004294323) and updated the status in below sheet

**Test plan sheet:** https://docs.google.com/spreadsheets/d/1phHo7MUMQNLjuJOic1TlZwoHM2T0iecT07VLoKtiqMg/edit#gid=539420021

 
**Existing issue:**

- #5166 

Please let me know if i need to verify thing else?

PR #5665, opened 2020-11-02
Fixed blockscout upgrade for releases in inconsistent state

carterqw2:
### Description

When someone deleted a deployment from a release, it is impossible to run `celotool` to restore the missing deployment. Removes the requirement to run Terraform to get the internal TX node load balancer IP address, and relies on `gcloud` CLI instead, removing the requirement to have service account credentials locally.

### Tested

Tested on Baklava.


jcortejoso (approved):
lgtm

PR #5674, opened 2020-11-02
Don't dereference env name for testnets

trianglesphere:
### Description

Does a literal comparison between env name rather than trying to dereference it in bash land (it already gets templated in).

### Other changes

### Tested

Testnets deploy and validators don't get stuck in crash loops.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


jcortejoso (approved):
lgtm

Issue #5424, opened 2020-10-19
Implement Cash Out and add more Cash In/Out Providers


jeanregisser:
### What's important about this? 

Users should have an easy way of getting value into and out of Valora. Valora currently only supports Cash In with Moonpay and has no Cash Out providers. This epic includes the work required to Cash Out, which will involve sending a cUSD transaction to the Cash Out provider. It also includes adding additional providers (Simplex, and Flexa and Wyre if they are ready) once the integrations are ready. 

### Definition of done

**Cash out flow:** The wallet supports Cash Out with cUSD using a provider's Webview. This will involve sending a cUSD transaction to the Cash Out provider. It is unclear exactly what this flow (need design) will be - whether it will require leaving the webview to approve a transaction - more detail to come from Simplex. 

**New providers:** As new providers like Simplex (expected by October 31) and others are added, their webview is added to the Cash In/Out screen. Just adding them in the same way Moonpay is displayed - no additional changes to this flow. 

Out of scope: working with providers to support integration (that work falls under DevEx this milestone), labeling transactions from providers in the transaction feed, local currency support in the Cash In/Out flow, cUSD/CELO automatic conversions if only CELO available

Open question: is CELO Cash Out support needed, or can it be cUSD only? 

Blocker for GA: we need at least 1 Cash-In provider for cUSD and 1 Cash-Out provider for cUSD

annakaz:
### What's important about this? 

Users should have an easy way of getting value into and out of Valora. Valora currently only supports Cash In with Moonpay and has no Cash Out providers. This epic includes the work required to Cash Out, which will involve sending a cUSD transaction to the Cash Out provider. It also includes adding additional providers (Simplex, and Flexa and Wyre if they are ready) once the integrations are ready. 

### Definition of done

**Cash out flow:** The wallet supports Cash Out with cUSD using a provider's Webview. This will involve sending a cUSD transaction to the Cash Out provider. It is unclear exactly what this flow will be - whether it will require leaving the webview to approve a transaction - more detail to come from Simplex. 

**New providers:** As new providers like Simplex (expected by October 31) and others are added, their webview is added to the Cash In/Out screen. Just adding them in the same way Moonpay is displayed - no additional changes to this flow. 

Out of scope: working with providers to support integration (that work falls under DevEx this milestone), labeling transactions from providers in the transaction feed, local currency support in the Cash In/Out flow, cUSD/CELO automatic conversions if only CELO available

[...]

PR #5678, opened 2020-11-03
Simplex cUSD integration in Valora

annakaz:
### Description

Add Simplex to allow users to buy cUSD. 

Simplex does not allow for an address to be passed as a parameter, so there is a banner with a user's address to copy within the Simplex webview

Update Cash In flow to go directly to the provider selection, as there is only one cUSD provider (Simplex) and one CELO provider (Moonpay) for now, so it is not necessary for users to enter an amount so we can display fees to them.

This UI will soon be replaced with the new Cash In / Out flow but wanted to land a minimal Simplex integration so we could allow our users to buy cUSD and to start testing Simplex as soon as possible. 

Screen Shot 2020-11-02 at 9 32 00 PM
Screen Shot 2020-11-02 at 9 32 09 PM
Screen Shot 2020-11-02 at 9 32 17 PM

### Tested

The integration team from Simplex tested a mainnet APK with the production integration to confirm that the purchase of cUSD works as expected
On an emulator to ensure flow displays as expected

### Related issues

- Part of #5424 

### Backwards compatibility

Yes


packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -136,9 +133,15 @@ function FiatExchangeOptions({ route, navigation }: Props) {
       },
     ],
     addFunds: [
+      {
+        image: <Image source={simplexLogo} style={styles.simplexLogo} resizeMode={'contain'} />,
+        screen: Screens.Simplex,
+        supportedCurrenciesNote: 'cUSD only',
```
jeanregisser: This needs to be translated.

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -136,9 +133,15 @@ function FiatExchangeOptions({ route, navigation }: Props) {
...
+      },
       {
         image: <Image source={moonpayLogo} style={styles.moonpayLogo} resizeMode={'contain'} />,
         screen: Screens.MoonPay,
+        supportedCurrenciesNote: 'CELO only',
```
jeanregisser: This needs to be translated.

packages/mobile/src/fiatExchanges/Simplex.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+function Simplex() {
+  const account = useSelector(currentAccountSelector)
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.addressBanner}>
+        <Text
+          style={[styles.tapToCopy, fontStyles.regular]}
+        >{`Tap to copy your Valora CUSD address`}</Text>
```
jeanregisser: This needs to be translated.

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
```diff
@@ -184,6 +194,24 @@ const styles = StyleSheet.create({
     height: 30,
     width: 104,
   },
+  simplexLogo: {
+    height: 59,
+    width: 111,
+  },
```
jeanregisser, line 200: nit: No need to address this now, but would be nice to use a standardized height for providers logo and adjust the asset instead. The simplex logo has some padding on the left which causes it to not be aligned with the MoonPay logo on the left.
Also we should provide the @1.5x, @2x, @3x, @4x sizes so it's crisp on all devices.

packages/mobile/src/fiatExchanges/Simplex.tsx
```diff
@@ -0,0 +1,66 @@
+import colors from '@celo/react-components/styles/colors'
+import fontStyles from '@celo/react-components/styles/fonts'
+import * as React from 'react'
+import { useTranslation } from 'react-i18next'
+import { StyleSheet, Text, View } from 'react-native'
+import { WebView } from 'react-native-webview'
+import { useSelector } from 'react-redux'
+import AccountNumber from 'src/components/AccountNumber'
+import { SIMPLEX_URI } from 'src/config'
```
jeanregisser, line 9: AFAIK, this new config wasn't committed, causing the build to fail.

jeanregisser (approved):
Great! :+1:
I went ahead and translated the new strings.

packages/mobile/src/fiatExchanges/Simplex.tsx
annakaz: Updated, thanks for catching this

packages/mobile/src/fiatExchanges/FiatExchangeOptions.tsx
annakaz: Sounds good, will merge for now as it as we'll be updating the UI with the new flow

packages/mobile/locales/en-US/fiatExchangeFlow.json
```diff
@@ -6,6 +6,9 @@
   "cashOutComingSoon": "Cash Out (Coming Soon)",
   "celoDollarsAt": "Celo Dollars @ <1></1>",
   "amountCelo": "Amount (CELO)",
+  "onlyCeloDollars": "cUSD only",
+  "onlyCelo": "CELO only",
+  "tapToCopyCeloDollarsAddress": "Tap to copy your {{appName}} CUSD address",
```
gnardini, line 11: Is the cap on `CUSD` (as opposed to `cUSD`) intended?

gnardini (approved):
Awesome to have another provider 😄

Lss-Ankit:
Hi @annakaz , @jeanregisser  Started verifying task using v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found below new issue: 
- https://github.com/celo-org/celo-monorepo/issues/5743. 

Also, in Android [Randomly] Something went wrong error message screen is shown after tapping on Simplex option from Add and Withdraw option. 
- Not getting verification code from mail id while adding cUSD from Simplex option. 
Note: 
- Checked spam mail as well but we are not getting mails there as well. 
Please let us know if we need to file a bug as we are not getting verification code mail 
Hence we are blocked to performed testing on this task can you please do the needful.

PR #5679, opened 2020-11-03
Use hard-coded gas values in the verifications flow

nategraf:
### Description

Especially on high latency and slow network connections, gas estimation can regularly take a very long time, often 10-20 seconds, and represents a point of failure if network conditions are degraded.

In this PR, gas values are set statically based on the gas usage of representative transactions. This eliminates the need for gas estimation and will increase performance and reliability of the verifications flow. I've chosen to use 1.5x the highest observed gas usage in representative transactions.

One risk of this change is that it makes the flow fragile to significant future changes to the Attestations flow. If for some reason in the future, the Attestations smart contract requires a significant amount more gas to run the same functions, old client will not be compatible. I am proposing this change now because I see that changes that increase the gas cost by more than 1.5x, without breaking interface, are fairly unlikely and will certainly not be done regularly.

### Tested

Tested by running through the verifications flow on Alfajores. Including on my WiFi and on simulated 3G connectivity.

Note:
When testing with simulated 3G connectivity, the flow was completed up to having selected issuers in a little over 1 minute (compared to about 45 seconds on WiFi) but failed on first try and took about 2 minutes 15 seconds on the second try when using `master`. 

### Backwards compatibility

100%


packages/mobile/src/identity/verification.ts
```diff
@@ -509,11 +509,15 @@ function* requestAttestations(
...
+    // Using hard-coded gas value to avoid running gas estimation.
+    // Note: This is fragile and needs be updated if there are significant
+    // changes to the contract implementation.
     yield call(
       sendTransaction,
       approveTx.txo,
       account,
-      newTransactionContext(TAG, 'Approve attestations')
+      newTransactionContext(TAG, 'Approve attestations'),
+      150000
```
gnardini: Thanks for this! How would you feel about having these 4 numbers as constants so that it's easier to tell what they represent at a glance?

nategraf: Good idea! I changed to do this in 9c74575

gnardini (approved):
Looks great 😁

jeanregisser (approved):
Great! 👍 

I'm noticing this doesn't require update to the test.
Do you think we should cover this in some way?

Lss-Ankit:
Hi @jeanregisser , @nategraf verified task using Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found the verified failed pop up is shown when user try to verify phone number using 3G connection mobile data. 

Verified device: iPhone SE (13.5.1)	Samsung Galaxy A5(7.0). 

Please find the logs for the same which may help to fix the issue. 
[celo_logs.txt](https://github.com/celo-org/celo-monorepo/files/5511243/celo_logs.txt)

PR #5683, opened 2020-11-03
MTWD remove state mapping and permissions

bowd:
### Description

Based on multiple discussions we came to the conclusion that we should remove the on-chain mapping of EOA->MTW from the MetaTransactionWalletDeployer. This mapping can grow stale when recovery enables a MTW to switch signer.

By removing that state it also makes sense to make the wallet deployment permissionless as there's no state.


packages/protocol/contracts/common/MetaTransactionWalletDeployer.sol
```diff
@@ -44,36 +31,9 @@ contract MetaTransactionWalletDeployer is
 
   /**
      * @notice Used in place of the constructor to allow the contract to be upgradable via proxy.
-     * @param initialDeployers a list of addresses that are allowed to deploy wallets
      */
-  function initialize(address[] calldata initialDeployers) external initializer {
+  function initialize() external initializer {
```
asaj: Does this contract still need to be ownable? Does it need to be initializable?

bowd: Nope good callout 👍

packages/protocol/contracts/common/MetaTransactionWalletDeployer.sol
```diff
@@ -10,29 +10,11 @@ import "./ExternalCall.sol";
 import "./Initializable.sol";
 import "./MetaTransactionWallet.sol";
 
-contract MetaTransactionWalletDeployer is
-  IMetaTransactionWalletDeployer,
-  ICeloVersionedContract,
-  Initializable,
-  Ownable
-{
+contract MetaTransactionWalletDeployer is IMetaTransactionWalletDeployer, ICeloVersionedContract {
```
asaj, line 9: Please remove unused imports above

asaj (approved):
Contract changes LGTM

PR #5687, opened 2020-11-03
Use migrationsConfig from release branch

### LINKED ISSUE #5590: Use linked libraries config from release branch
### Expected Behavior

`make-release` uses versioned linked libraries

### Current Behavior

`make-release` uses canonical linked libraries

nambrot:
### Description

Adjusts `make-release` to use the migrationsConfig from the git reference we are deploying

### Related issues

- Fixes #5590 

### Backwards compatibility

Question: Just wanted to check that using the migrations config from celo-core-contracts-v1.rc1 is ok to do, even though it specifies FixidityLib as a dependency. As far as I can tell, it’s just gonna result in useless FixidityLib links or will it prevent an inline?


packages/protocol/scripts/bash/make-release.sh
```diff
@@ -47,14 +47,18 @@ rm -rf build/contracts
 yarn build
 rm -rf $BUILD_DIR && mkdir -p $BUILD_DIR
 mv build/contracts $BUILD_DIR
-
+cp migrationsConfig.js $BUILD_DIR/
 git checkout -
+cp $BUILD_DIR/migrationsConfig.js ./
 yarn build:ts
 
+
```
m-chrzan: Extraneous newline?

m-chrzan (approved):
> As far as I can tell, it’s just gonna result in useless FixidityLib links or will it prevent an inline?

That's correct! Inlining happens at compile time, so the linker won't have any references in the bytecode to which to link at deploy time.

PR #5690, opened 2020-11-03
[Wallet] Add shake to Contact screen

### LINKED ISSUE #5689: Add shake to send logs functionality to the wallet
### Expected Behavior

When shaking the device, the user is directed to the Contact screen so that they can share logs.

### Current Behavior

Shaking does nothing.

gnardini:
### Description

When shaking the device, the user is now directed to the Contact screen so that they can share logs.

### Tested

On physical Android and iOS devices by shaking them.

### Related issues

- Fixes #5689


gnardini:
Open question: 
The current implementation redirects users to the contact screen always when shaking. Is that desirable or would it better for this to happen only during the onboarding flow?

jeanregisser:
> Open question:
> The current implementation redirects users to the contact screen always when shaking. Is that desirable or would it better for this to happen only during the onboarding flow?

I think that's fine!

packages/mobile/src/navigator/NavigatorWrapper.tsx
```diff
@@ -97,6 +99,15 @@ export const NavigatorWrapper = () => {
...
   }, [isReady])
 
+  React.useEffect(() => {
+    RNShake.addEventListener('ShakeEvent', () => {
+      navigate(Screens.SupportContact)
+    })
+    return () => {
+      RNShake.removeEventListener('ShakeEvent')
+    }
+  })
```
jeanregisser:
```suggestion
  }, [])
```

Otherwise it's gonna subscribe/unscribe on each render.

jeanregisser (changes requested):
Awesome 💯 
Almost good to merge.

packages/mobile/src/navigator/NavigatorWrapper.tsx
gnardini: Whoops, that's really bad from me :(

jeanregisser: No worry, it's easy to miss, I think there's a lint rule we could add to make sure it's present.

jeanregisser (approved):
Ship it! :shipit:

Lss-Ankit:
Hi @gnardini , @jeanregisser  Verified task using Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found the following: 

- User is redirect to Contact with text field and attach log screen. 
- Also, observed that user is not redirected to Contact screen when user shaking the phone if user is on Enter PIN screen. 
Issue is repro only when user is on Enter PIN screen from Account key screen. 
Issue does not repro if user is on Enter PIN screen while sending payment. 

[Attachment: ](https://drive.google.com/file/d/1A7QEVCBuBRpNnS0uawSKyfyP91awffNQ/view?usp=sharing)
Can you please confirm should we file a bug for the same?

gnardini:
I think it's not necessary to file a bug for this since the main goal was for it to be used while on the onboarding flow.

Lss-Ankit:
ohk thanks for the confirmation!!

PR #5691, opened 2020-11-03
Proxy docs 

kevjue:
### Description

Proxy docs subsection for the Validator guid docs.


packages/docs/operations-manual/proxy.md
```diff
@@ -0,0 +1,22 @@
+# Validator proxy
+
+Validator uptime is essential for the health of the Celo blockchain.  To help with validator uptime, operators can use the proxy node, which will provide added security for the validator.  It allows the validator to run within a private network, and to communicate to the rest of the Celo network via the proxy.
+
+Also, starting from the Celo client 1.2 release, we will support assigning multiple proxies per validator.  This provides better uptime for the validator for the case of a proxy going down.  Also, it will help with making each proxy enode URL less public by only sharing it with a subset of the other validators.
+
+**NOTES**
+The communication protocol between the validator and it's proxies implemented in release 1.2 is NOT backwards compatible to the pre-1.2 protocol.  So if the proxy or validator is being upgraded to 1.2, then both needs to be upgraded to that version.  Note that validators and proxies using release 1.2 are still compatible with remote nodes.
```
trianglesphere: I've also got this warning at https://github.com/celo-org/celo-monorepo/blob/trianglesphere/replica-e2e/packages/docs/operations-manual/node-upgrades.md#upgrading-proxy-nodes. It's probably good to have it both places. One thing that I used was the little card thing rather than putting it inline.

trianglesphere (approved):
Someone who knows docs better should make sure that this page gets deployed and included on the sidebar.

packages/docs/operations-manual/proxy.md
kevjue: I changed it to use your style so that we're consistent.

trianglesphere:
I think the link you added is broken and should be `../getting-started/running-a-validator-in-mainnet.md#deploy-a-validator-machine).` instead

PR #5693, opened 2020-11-03
Only release contracts with Proxies

### LINKED ISSUE #5474: `make-release` only releases contracts with proxies

nambrot:
### Description

This adds an additional check to only deploy contracts (or consider deploying) that have an accompanying proxy contract

### Tested

- Dry-ran it on Alfajores

### Related issues

- Fixes #5474 

### Backwards compatibility

Should be


packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -72,6 +72,19 @@ class ContractAddresses {
 
 const REGISTRY_ADDRESS = '0x000000000000000000000000000000000000ce10'
 
+const isCoreContract = async (contractName: string) => {
```
m-chrzan: Would it be better to call this `isProxiedContracts`? Is it the case that there are no (and there won't be) contracts with a corresponding Proxy that wouldn't be considered core Celo contracts? 

Would using `CeloContractName` from `lib/registry-utils.ts` be more appropriate?

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -210,7 +223,9 @@ module.exports = async (callback: (error?: any) => number) => {
       }
     }
     for (const contractName of contracts) {
-      await release(contractName)
+      if (await isCoreContract(contractName)) {
```
m-chrzan: Will this not skip all linked libraries?

packages/protocol/scripts/truffle/make-release.ts
nambrot: I did both!

packages/protocol/scripts/truffle/make-release.ts
nambrot: I believe each `release` will deploy the linked library separately

packages/protocol/scripts/truffle/make-release.ts
```diff
@@ -72,6 +73,22 @@ class ContractAddresses {
 
 const REGISTRY_ADDRESS = '0x000000000000000000000000000000000000ce10'
 
+const isProxiedContract = async (contractName: string) => {
+  if (contractName.endsWith('Proxy')) {
+    return false
+  }
+
+  try {
+    await artifacts.require(`${contractName}Proxy`)
```
m-chrzan: I don't think you need to `await` on `artifacts.require`, so this whole function doesn't need to be async.

m-chrzan (approved):

PR #5697, opened 2020-11-03
Upgrade our bip39 to TS version

cmcewen:
### Description

Library updates include bip39 + randombytes


yarn.lock
```diff
@@ -25641,6 +25641,35 @@ node-libs-react-native@1.0.3, node-libs-react-native@^1.0.3:
     url "^0.11.0"
     util "^0.10.3"
 
+node-libs-react-native@^1.2.0:
```
jeanregisser: We should use only one version of `node-libs-react-native`, right?

packages/mobile/package.json
```diff
@@ -229,7 +229,7 @@
       },
       "ios.sim.debug": {
         "binaryPath": "ios/build/Build/Products/Debug-iphonesimulator/celo.app",
-        "build": "set -o pipefail && xcodebuild -workspace ios/celo.xcworkspace -scheme celo-test -configuration Debug -sdk iphonesimulator -derivedDataPath ios/build | bundle exec xcpretty",
+        "build": "set -o pipefail && xcodebuild -workspace ios/celo.xcworkspace -scheme celo-test -configuration Debug -sdk iphonesimulator -derivedDataPath ios/build  | bundle exec xcpretty",
```
jeanregisser: This needs to be reverted.

jeanregisser (approved):
Great! 👍 🥇 
Could you update the description of the PR with more context about why it's needed (for posterity)?

PR #5698, opened 2020-11-03
[Attestation Service] supply Nexmo callback URL so multiple services can share account, use application only to filter numbers

timmoreton:
### Description

* Supply the callback parameter when sending SMS via Nexmo. This means that it no longer needs to be specified in the Nexmo dashboard, meaning that multiple Attestation Services can share an account without configuring applications.
* As @zviadm pointed out, JWT-based auth for Nexmo applications doesn't work or appear to have any effect. Nowhere in the Nexmo dashboard can you see logs/analytics filtered by application. 
* Number selection was not filtering by application id. So we now make the effect of supplying `NEXMO_APPLICATION` that it filters to select only numbers linked to the application. This doesn't seem to require JWT-based auth to that application.
* Updated docs. 

### Tested

* See comments below.
* Removed callback URL from general settings and Nexmo app in turn to verify successful auth with each.
* Configured Nexmo with and without applications and observed success sending from numbers linked to a supplied application. 

### Backwards compatibility

Yes


zviadm:
@timmoreton I am not sure if providing PrivateKey will actually work for messages.sendSms calls anyways. Looking at the nexmo sdk, that stuff just seems to use what they call "legacy" httpClient which doesn't use applicationId information. 
I think applicationId thing is simply not supported right now for sending SMSes through that SDK. 

Were you able to confirm that it was indeed picking up the applicationId when testing with the private key and stuff?

timmoreton:
> @timmoreton I am not sure if providing PrivateKey will actually work for messages.sendSms calls anyways. Looking at the nexmo sdk, that stuff just seems to use what they call "legacy" httpClient which doesn't use applicationId information.
> I think applicationId thing is simply not supported right now for sending SMSes through that SDK.
> 
> Were you able to confirm that it was indeed picking up the applicationId when testing with the private key and stuff?

You're right, it appears the notion of applications are half-implemented in Nexmo and have little effect. There is no way you can see in the logs or analytics in the dashboard what application a SMS was sent via. There also appears to be no access control, so linking a number to an application doesn't mean that you have to auth via that route.  I just tested a couple of additional things:
* with the applicationId and privateKey specified, and callback not specified in sendSms or in Nexmo dashboard default, it fails to pick up the callback URL correctly from the application settings.
* with the applicationId supplied but not the privatekey, you can anyway select a number linked to that application.

Luckily, by providing the callback in the sendSms function, we can sidestep the need for applications anyway. I will leave them as a convenience for being able to select a specific group of numbers.

aslawson (commented):
just so i understand all that's going on, it seems that:

1. we originally added nexmo applications because we thought they were necessary for specifying a callback url in the nexmo dashboard.  it turns out we can actually pass the url as a callback in the `sendSMS` request
2. turns out the request were not actually being routed to the applications, but rather pulling from the global pool.  this was resolved with passing the applicationId as an option and 'has_application' = true.
3. even with the applicationId included, nexmo failed to use the callback url specified in the nexmo dashboard for that application.

since we identified that we can pass the deliveryStatusURL as a parameter, there is really no need for multiple applications unless someone wants to split their phone numbers among the services.

is this understanding correct?

timmoreton:
> 1. we originally added nexmo applications because we thought they were necessary for specifying a callback url in the nexmo dashboard.  it turns out we can actually pass the url as a callback in the `sendSMS` request

correct

> 2. turns out the request were not actually being routed to the applications, but rather pulling from the global pool.  this was resolved with passing the applicationId as an option and 'has_application' = true.

it's very unclear from Nexmo what it would mean for "the request to be routed to the applications".  It doesn't seem to make any difference in terms of billing, logs, analytics, etc. We were not selecting numbers from the application, and so if you now specify an applicationId, we now do that -- as the only point of having an application configured.

> 3. even with the applicationId included, nexmo failed to use the callback url specified in the nexmo dashboard for that application.

even supplying an applicationId and private key, correct.

> since we identified that we can pass the deliveryStatusURL as a parameter, there is really no need for multiple applications unless someone wants to split their phone numbers among the services.

correct - the docs should now make that clear - LMK if you feel they don't!

aslawson:
> correct - the docs should now make that clear - LMK if you feel they don't!
@timmoreton yes they seem clear -- I just wanted to understand the series of learnings to lead to that conclusion 👍

packages/attestation-service/config/.env.development
```diff
@@ -45,6 +45,7 @@ TWILIO_UNSUPPORTED_REGIONS=CU,SY,KP,IR,SD
 
 # MessageBird parameters (fill in values for 'x')
 MESSAGEBIRD_API_KEY=x
+MESSAGEBIRD_UNSUPPORTED_REGIONS=CU,SY,KP,IR,SD
```
aslawson, line 48: do we want to add messagebird to `SMS_PROVIDERS` or wait until we feel more confident about it as an option?

aslawson (approved):
LGTM

packages/attestation-service/config/.env.development
timmoreton: mm this is a release process edge case. Because the docs reference this config file from master branch, we don't want to include anything that is not backwards compatible with the current released version. Because MessageBird is not yet in a released version, adding it to `SMS_PROVIDERS` would break things.  We should add it, but only just as we release the next version I guess.

PR #5699, opened 2020-11-04
[ContractKit] Add retry and backoff around fetch metadata

timmoreton:
### Description

Add retry and backoff around fetching metadata. Seen a few occasions where the services hosting validator metadata will return transient errors, such as connection resets, in which case an attestation attempt from Valora or a status check from the metadata crawler will fail and not retry. 

### Tested

Locally

### Backwards compatibility

Yes


packages/contractkit/src/identity/metadata.ts
```diff
@@ -38,11 +39,17 @@ export class IdentityMetadataWrapper {
...
   static async fetchFromURL(kit: ContractKit, url: string) {
-    const resp = await fetch(url)
-    if (!resp.ok) {
-      throw new Error(`Request failed with status ${resp.status}`)
-    }
-    return this.fromRawString(kit, await resp.text())
+    return retryAsyncWithBackOff(
+      async () => {
+        const resp = await fetch(url)
+        if (!resp.ok) {
```
AlexBHarley, line 45: Thoughts on a circuit breaker for 404? `resp.ok` will be false on a 404, but we don't want to retry.

nambrot:
Can we do this at a calling site? This to me doesn't seem like the library should make that choice

timmoreton:
> Can we do this at a calling site? This to me doesn't seem like the library should make that choice

I think it complicates the calling service (for example CK uses this to figure out if a set of issuers are valid -- and the calling service would then have to retry ALL of them, if one metadata fetch failed).  Plus I feel that the library is a fine place to work around transient failures. That's exactly what the the blockchain tx pool does, for instance.  Looks like alchemy have extended that philosophy to [web3](https://docs.alchemyapi.io/documentation/alchemy-web3) -- would be interesting to consider that.

packages/contractkit/src/identity/metadata.ts
timmoreton: great point!

nambrot:
> for example CK uses this to figure out if a set of issuers are valid -- and the calling service would then have to retry ALL of them, if one metadata fetch failed

I'm not sure how this applies? Retrying with asyncWithBackoff seems quite easy to use?

Fwiw, not a strong opinion on my side

timmoreton:
> I'm not sure how this applies? Retrying with asyncWithBackoff seems quite easy to use?

It's easy, but if I have requested say 3 attestations and I wrap say [`getActionableAttestations`](https://github.com/celo-org/celo-monorepo/blob/master/packages/contractkit/src/wrappers/Attestations.ts#L270) with `asyncWithBackoff` from the calling client, then if 1 call to  `fetchFromURL` fails, then I need to re-make more view calls and all 3 calls to `fetchFromURL`. Does that make sense?

packages/contractkit/src/identity/metadata.ts
AlexBHarley: Looking at the fetch docs I see any `4XX` code will make `resp.ok` false. Worth noting

nambrot:
That makes sense. How about an optional parameter specifying the retry logic of network calls within?

nambrot (approved):

PR #5703, opened 2020-11-04
Add ability to specify signing algorithm for Azure HSM

bowd:
### Description

Azure HSM gives us the ability to create curves that support signing algorithms which we can use for managing Celo accounts. This feature has undergone some changes recently:

- Curve `SECP256K1` is now named `P-256K`
- Signing algorithm `ECDSA256` is now named `ES256K`

[Source](https://github.com/Azure/azure-sdk-for-node/issues/4603)

But when interacting with the signing API you can only use `ECDSA256` with `SECP256K1` curves and must use `ES256K` with `P-256K` curves, annoyingly. 

We currently have `SECP256K1` curves in Azure and have recently created `P-256K` for Komenci, but our KeyVaultClient was referencing `ECDSA256` as a constant. Thus I was seeing this error when trying to interact with the keys:

```
"message": "Key and signing algorithm are incompatible. Key https://mainnet-komenci-eastus.vault.azure.net/keys/mainnet-komenci-funder-eastus/0c6493f97acb41e7ba35546488b7f14a uses curve 'P-256K', and algorithm 'ECDSA256' can only be used with curve 'P-256' or 'SECP256K1'."
```

This PR adds the ability to specify an optional signing algorithm. This is intended as a temporary workaround in order to unblock Komenci work and for me to test the assumption that this is actually the issue.

I would suggest improving on this PR by deriving the signing algorithm to be used from the curve type of the key.

### DIFF
--- a/packages/sdk/wallets/wallet-hsm-azure/src/azure-hsm-wallet.ts
+++ b/packages/sdk/wallets/wallet-hsm-azure/src/azure-hsm-wallet.ts
@@ -3,23 +3,24 @@ import { Address, publicKeyToAddress } from '@celo/utils/lib/address'
 import { RemoteWallet } from '@celo/wallet-remote'
 import debugFactory from 'debug'
 import { AzureHSMSigner } from './azure-hsm-signer'
-import { AzureKeyVaultClient } from './azure-key-vault-client'
+import { AzureKeyVaultClient, AzureKeyVaultSigningAlgorithm } from './azure-key-vault-client'
 
 const debug = debugFactory('kit:wallet:aws-hsm-wallet')
 
 // Azure Key Vault implementation of a RemoteWallet
 export class AzureHSMWallet extends RemoteWallet<AzureHSMSigner> implements ReadOnlyWallet {
-  private readonly vaultName: string
   private keyVaultClient: AzureKeyVaultClient | undefined
 
-  constructor(vaultName: string) {
+  constructor(
+    private readonly vaultName: string,
+    private readonly signingAlgorithm?: AzureKeyVaultSigningAlgorithm
+  ) {
     super()
-    this.vaultName = vaultName
   }
 
   protected async loadAccountSigners(): Promise<Map<Address, AzureHSMSigner>> {
     if (!this.keyVaultClient) {
-      this.keyVaultClient = this.generateNewKeyVaultClient(this.vaultName)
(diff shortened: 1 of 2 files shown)


AlexBHarley (approved):

AlexBHarley:
Actually would be nice if we ran the e2e tests just to verify we're not breaking anything.

bowd:
Agreed, but like I said I don't even thing this is the best approach for the problem - I think we can solve this internally but I'm using this branch temporarily in Komenci to make something work. But I wouldn't suggest we merge this until we (a) settle on a long term solution (b) add more testing around it.

But FYI I validated that it does indeed work on the new curves 🇨🇼

packages/contractkit/src/utils/azure-key-vault-client.ts
```diff
@@ -5,6 +5,13 @@ import { BigNumber } from 'bignumber.js'
 import { bigNumberToBuffer, bufferToBigNumber, makeCanonical, Signature } from './signature-utils'
 import { publicKeyPrefix, recoverKeyIndex, sixtyFour, thirtyTwo } from './signing-utils'
 
+export enum AzureKeyVaultSigningAlgorithm {
+  ECDSA256 = 'ECDSA256',
+  ES256K = 'ES256K',
```
codyborn: Let's change the default to `ES256K` since `ECDSA256` isn't compatible with the HSMs anymore

codyborn: Or does this only affect keys created after a certain date?

codyborn: > Keys created with the preview curve name SECP256K1 can only be signed with the preview algorithm name, which happens to be ECDSA256 (not ES256K). It's just a name change. Internally they do the exact same thing.

I see.. can we instead fallback to the old curve if the new one fails?

bowd: Yeah that's one possibility. I was thinking that the cleanest way would be to use the API to get the name of the curve for that specific key from Azure and then based on that determine which algo name to use. I'll investigate this a bit today to see if there's a clean solution. I feel like fallback to other name would be some legacy weird behaviour. 
Querying the curve type would also allow us to improve DevEx by throwing more appropriate errors if the curve is not what we expect.

bowd:
I rebased on master to be in line with the refactoring but I also attempted another implementation here: #6010
It's a bit cleaner but I have to test it out a bit and make sure that everything is as expected.

PR #5704, opened 2020-11-04
chore: relax helm version to v3 instead of exactly v3.3

AlexBHarley:
### Description

Helm v3.3 is too restrictive, I installed the latest and got v3.4


tkporter (approved):
Thanks!

PR #5709, opened 2020-11-04
Add CUSD transfer to MTW

asaj:
### Description

This PR adds the ability for anyone to transfer CUSD out of a MetaTransactionWallet to its signer. This will allow CUSD that is accidentally sent to the MTW instead of the EOA to be automatically transferred to the EOA without any action from the user. This functionality can eventually be deprecated by upgrading the MTW.

 
### Other changes

None

### Tested

Unit tests in progress.

### Related issues

- None
### Backwards compatibility

Yes


packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -260,6 +262,21 @@ contract MetaTransactionWallet is
...
+   * @notice Transfers all cUSD to the `signer`.
+   * @dev Valora v1.2 ignores the `walletAddress` of the user, which may result in cUSD being
+   *   sent to users MetaTransactionWallets instead of their EOA. This function allows that cUSD
+   *   to be transferred to the EOA without requiring action by the user.
+   * @return Whether or not the cUSD transfer succeeded.
+   */
+  function transferCeloDollarsToSigner() external returns (bool) {
+    bytes32 STABLE_TOKEN_REGISTRY_ID = keccak256(abi.encodePacked("StableToken"));
+    IRegistry registry = IRegistry(0x000000000000000000000000000000000000ce10);
+    IERC20 stable = IERC20(registry.getAddressForOrDie(STABLE_TOKEN_REGISTRY_ID));
```
m-chrzan: Why not use UsingRegistry for this?

packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -591,4 +599,30 @@ contract('MetaTransactionWallet', (accounts: string[]) => {
...
+    const value = 100
+    let mockStableToken: MockStableTokenInstance
+    beforeEach(async () => {
+      const Registry: RegistryContract = artifacts.require('Registry')
+      const registry: RegistryInstance = await Registry.at(
+        '0x000000000000000000000000000000000000ce10'
+      )
+      // Problem is this is owned by governance...
+      console.log(await registry.owner())
+      console.log(accounts)
```
m-chrzan: Assume these are debugging logs/comments that should be removed from final version?

packages/protocol/test/common/metatransactionwallet.ts
```diff
@@ -591,4 +599,30 @@ contract('MetaTransactionWallet', (accounts: string[]) => {
...
+      const MockStableToken: MockStableTokenContract = artifacts.require('MockStableToken')
+      mockStableToken = await MockStableToken.new()
+      await registry.setAddressFor(CeloContractName.StableToken, mockStableToken.address)
+      await mockStableToken.mint(signer, value)
```
m-chrzan: Why do we mint to `signer`, not `wallet`?

packages/protocol/contracts/common/MetaTransactionWallet.sol
asaj: I wanted to avoid adding a storage variable for something that might be deprecated in the near future.

packages/protocol/test/common/metatransactionwallet.ts
asaj: Yeah tests are still WIP

asaj: Namely, I can't register the mock token b/c the registry is owned by governance

packages/protocol/test/common/metatransactionwallet.ts
asaj: Oops

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -260,6 +262,21 @@ contract MetaTransactionWallet is
     return (returnValues, returnLengths);
   }
 
+  /**
```
nambrot, line 264: Can we add a note that this will possibly be deprecated?

packages/protocol/contracts/common/MetaTransactionWallet.sol
```diff
@@ -260,6 +262,21 @@ contract MetaTransactionWallet is
...
+   * @notice Transfers all cUSD to the `signer`.
+   * @dev Valora v1.2 ignores the `walletAddress` of the user, which may result in cUSD being
```
nambrot: Nit: Valora <v1.2

packages/protocol/contracts/common/MetaTransactionWallet.sol
m-chrzan: Makes sense. Make the StableToken Registry ID a constant rather than computing it on each call?

bowd: Why not make this `transferERC20ToSigner(address)`? So we can pass in arbitrary ERC20 contracts and get celo support as well.

asaj: Yeah great suggestion

PR #5712, opened 2020-11-04
add back alfajores developer wallet links

aaronmgdr:
### Description

Adds links to the Celo Alfajores Developer wallet
### Other changes

changes copy to better reflect this is a test net only wallet,
fixes some alignment issues

### Tested
https://dev.celo.org/developers/wallet

Screen Shot 2020-11-04 at 10 07 19 AM


packages/web/public/static/locales/en/download.json
```diff
@@ -1,11 +1,11 @@
 {
-  "coverTitle": "Celo Test Wallet",
-  "coverSubTitle": "Temporarily Not Available",
+  "coverTitle": "Celo Developer Wallet",
+  "coverSubTitle": "A testing wallet.\nDo not send real funds.",
```
critesjosh: actually can we update this to be more descriptive?

```suggestion
  "coverSubTitle": "An Alfajores Testnet wallet.\nDo not send real funds.",
```

isabellewei (approved):

PR #5715, opened 2020-11-04
Check if contracts need to relink libraries

### LINKED ISSUE #5598: check-versions should check linked libraries
- It should be well documented how changes in dependencies are expected to be reflected in implementing contracts

m-chrzan:
### Description

The check-versions script should identify all contracts that need to be redeployed by signaling a version number change.

When a contract's linked library dependency is updated, the contract itself will have to be redeployed to update its linking to the new library version's address. This was previously not checked, this PR adds a check for this.

### Other changes

Moved the `Change` class and its children/visitor to a separate file to avoid circular imports.

### Tested

Unit tests.

### Related issues

- Fixes #5598


nambrot (approved):
Approving to get the fix in, test PR coming after

PR #5722, opened 2020-11-05
s/--celostats/--ethstats

keefertaylor:
### Description

Switch flag names: `--celostats` doesn't exist in blockchain 1.1.0.


timmoreton (approved):
@kevjue can you make a note to revert this change after 1.2.0 is promoted to mainnet ?

PR #5724, opened 2020-11-05
Add bytecode SHA3 for integrity check

bowd:
### Overview

We need to remove the dependency on the protocol package and validate the Proxy integrity by storing a constant SHA3 of the Proxy bytecode.


asaj:
How was this tested?

asaj (approved):

PR #5732, opened 2020-11-05
Switch Valora escrow expiry to 1 hour

### LINKED ISSUE #5731: Reduce sender escrow time from 24 hours to 1 hour
### Expected Behavior

Users can reclaim funds in escrow if verification fails after 1 hour

Important for the GA use case as it allows people to reclaim funds in case the receiver is not able to verify or verifies a different number

### Current Behavior

Can't reclaim funds for 24 hours

annakaz:
### Description

Reduce expiry time to one hour instead of 24 hours. This is to mitigate the effect of a verification failure preventing a GA user from receiving their funds

### Tested

Just a parameter change, and we have a test that checks that the escrow payment expiry works as expected

### Related issues

- Fixes #5731 

### Backwards compatibility

This will only affect escrow payments sent moving forward


jeanregisser (approved):
LGTM 👍

PR #5734, opened 2020-11-05
[Wallet] Remove force unlock from initial onboarding flow

i1skn:
### Description

https://github.com/celo-org/celo-monorepo/pull/5656 has introduced mandatory enter of PIN code while onboarding flow to prevent account unlock timeout issue.

This PR would disable it specifically to the initial onboarding (we hope that 10 minutes since user sets the PIN would be enough to finish onboarding initially) and keep it if user comes back to verification later. 

### Caveat
The only one I see is when user goes to verification from Home screen, enters PIN, press Start/Resume, press back, user again needs to enter a PIN code. Not sure how often this is happening and to which extent it would ruin the UX.

### Tested

Initial onboarding with verification and return to verification flow later.

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/e2e/src/usecases/RestoreAccountOnboarding.js
+++ b/packages/mobile/e2e/src/usecases/RestoreAccountOnboarding.js
@@ -78,11 +78,6 @@ export default RestoreAccountOnboarding = () => {
     await sleep(5000)
   })
 
-  it('Pin', async () => {
-    // Enter pin before verification starts
-    await enterPinUi()
-  })
-
   it('VerifyEducation', async () => {
     await waitForElementId('VerificationEducationSkipHeader')
 
--- a/packages/mobile/src/identity/actions.ts
+++ b/packages/mobile/src/identity/actions.ts
@@ -182,6 +182,7 @@ export interface UpdateAddressDekMapAction {
 
 export interface FetchVerificationState {
   type: Actions.FETCH_VERIFICATION_STATE
+  forceUnlockAccount: boolean
 }
 
 export interface UpdateVerificationState {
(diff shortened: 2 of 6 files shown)


jeanregisser (approved):
Thanks! 💯 👍

Lss-Ankit:
@jeanregisser Verified task using Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317) and found that user is able to complete phone number verification flow with out facing any verification failed pop up. 

Verified on : iPhone XR (13.2.2) and Google Pixel XL (8.1)

Lss-Ankit:
Hi @i1skn @jeanregisser  Verified task using Test flight build v1.4.0 (28) and Android play store (internal build) v1.4.0 (1004294317)  and found the following: 

- Verify Phone Number during onboard user: 
1. [ ] Enter PIN is asked when user taps on Start button from Connect Your Phone number screen. after that  User is able to sign up successfully.   
Note: It is happening randomly Sometimes Enter PIN is asked or some times it is not asked. 
We have verified this scenarios three time where 1st time we got the  Enter PIN screen. 
2. [ ] Enter PIN is not asked when user taps on Resume button from Connect Your Phone number screen if user navigate back from the Confirm screen (Paste Three code screen).  after that  User is able to sign up successfully.  
3. [ ] Enter PIN is asked when user taps on Resend button while onboarding the user. 

Skip the Phone Number verification during on boarding and Perform verification flow on Home screen. 
1. [ ] Enter PIN screen is shown when user taps on Confirm Now button from CTA on home screen. 
2. [ ] Enter PIN screen is shown when user taps on Confirm your phone number option from Setting screen. 
3. [ ] Enter PIN screen is not asked when user taps on Start button from Connect your phone number screen. 
4. [ ] Enter PIN screen is shown when user taps on Cancel button from Confirming screen with timer. 
5. [ ] After Step 4, Enter PIN screen is not shown when user taps on Start button from Connect your Phone number screen. 
6. [ ] Enter PIN screen is shown when user taps on < button from Confirm screen. 
Due to issue #5738  we are not able to test further. 

Can you please let us know if we need to verify any thing else?

jeanregisser:
Thanks!

>  Enter PIN screen is not asked when user taps on Start button from Connect your phone number screen.
Can you clarify this?

@i1skn what do you think of these?

PR #5737, opened 2020-11-05
Update full-nodes helm chart

trianglesphere:
### Description

I ran into a couple issues trying to deploy a full node set to my own testing cluster/testnet. These are the minimal changes to make it work.

One thing that I have less context around is if we should completely remove the rolling updates section. Those rolling update options are for (I believe) deployments and not StatefulSet.  

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


tkporter (approved):
Surprised this never bit us earlier, thanks!

PR #5741, opened 2020-11-06
Pass through dynamic DB vars

codyborn:
### Description

Use external managed postgres instance. Pass through new variables from environment variables. Still TODO is to migrate the password to a K8s secret (temporarily made it a regular env var to get it working).

### Other changes

Add env vars for new region: West Europe

### Tested

Staging deployment


aaronmgdr:
stale

PR #5742, opened 2020-11-06
Komenci <> Valora Integration: invites and escrow

tarikbellamine:
### Description
This PR builds on the previous two PRs of the series ([DEK/wallet registration](#5622) and [feeless verification](#5663) to provide an escrow experience that does not use "invite codes". 

This PR includes unmerged changes included in previous PRs. The most relevant files for this PR:
- escrow/saga
- invite/saga

The pertinent UI changes:
1. Elimination of the "Enter Invite Code" screen as there is no longer an invite code to redeem
2. Copy changes to prepopulated invite messages we form for users on WhatsApp or SMS

The transaction changes:
1. Unique escrow `paymentIds` and the private key required to redeeem are now both formed using a hash of the recipient's pepper
2. Added support for redemption of multiple escrow transctions

### Other changes
N/A

### Tested
Not yet

### Backwards compatibility
No


packages/mobile/src/escrow/saga.ts
```diff
@@ -102,6 +134,74 @@ function* transferStableTokenToEscrow(action: EscrowTransferPaymentAction) {
...
+      phoneHash
+    )
+
+    // Need to increment the derivation path of the paymentId based on
+    // how many are pending to avoid collisions with existing escrow txs
+    const addressIndex = escrowPaymentIds.length
+    const { publicKey, privateKey } = generateDeterministicInviteCode(pepper, addressIndex)
+    console.log('ESCROW DATA: ')
+    console.log(addressIndex, pepper)
+    console.log(publicKey, privateKey)
```
nambrot: Just highlighting to remove this

packages/mobile/src/escrow/saga.ts
```diff
@@ -102,6 +134,74 @@ function* transferStableTokenToEscrow(action: EscrowTransferPaymentAction) {
...
+  Logger.debug(TAG + '@transferToEscrowWithoutCode', 'Begin transfer to escrow')
+  try {
+    ValoraAnalytics.track(EscrowEvents.escrow_transfer_start)
+    const { phoneHashDetails, amount, context } = action
+    const { phoneHash, pepper } = phoneHashDetails
+    const [contractKit, walletAddress]: [
+      ContractKit,
+      string,
+      string,
+      FeelessVerificationState
```
nategraf: It looks like this a typo in copying from elsewhere.
```suggestion
```

packages/mobile/src/escrow/saga.ts
```diff
@@ -102,6 +134,74 @@ function* transferStableTokenToEscrow(action: EscrowTransferPaymentAction) {
...
+
+    const escrowPaymentIds: string[] = yield call(
+      [escrowWrapper, escrowWrapper.getReceivedPaymentIds],
+      phoneHash
+    )
+
+    // Need to increment the derivation path of the paymentId based on
+    // how many are pending to avoid collisions with existing escrow txs
+    const addressIndex = escrowPaymentIds.length
+    const { publicKey, privateKey } = generateDeterministicInviteCode(pepper, addressIndex)
```
nategraf: Whenever a payment is redeemed, it is removed from the received array. As a result if a user receives an escrow payment with this methods, redeems it, and then received another the payment ID will be repeated. (In theory, the signature form the previous transaction can then be reused to avoid the need to know the recipients pepper.)
More immediately problematic though is that if the recipient had two pending payments, redeemed the first, then received a third, the second and third would have the same payment ID as the second and as a result would fail.

packages/mobile/src/escrow/saga.ts
```diff
@@ -117,6 +217,125 @@ function* registerStandbyTransaction(context: TransactionContext, value: string,
...
+
+  const { r, s, v } = splitSignature(contractKit, signature)
+  const withdrawTx = escrowWrapper.withdraw(paymentId, v, r, s)
+  return withdrawTx
+}
+
+function* withdrawFromEscrowUsingPepper(komenciActive: boolean = false) {
+  try {
+    ValoraAnalytics.track(OnboardingEvents.escrow_redeem_start)
+    Logger.debug(TAG + '@withdrawFromEscrowViaKomenci', 'Withdrawing escrowed payment')
```
nategraf: Should the tag in this logging statement match the function name?
```suggestion
    Logger.debug(TAG + '@ withdrawFromEscrowUsingPepper', 'Withdrawing escrowed payment')
```

packages/mobile/src/escrow/saga.ts
```diff
@@ -117,6 +217,125 @@ function* registerStandbyTransaction(context: TransactionContext, value: string,
...
+    const [contractKit, walletAddress, mtwAddress, feelessVerificationState]: [
+      ContractKit,
+      string,
+      string,
+      FeelessVerificationState
+    ] = yield all([
+      call(getContractKit),
+      call(getConnectedUnlockedAccount),
+      select(mtwAddressSelector),
+      select(feelessVerificationStateSelector),
```
nategraf, line 265: Should this also be getting the EOA verification state in case that's how they were verified, or does this flow not support that case? If the latter, could you add a note to the function?

packages/mobile/src/escrow/saga.ts
```diff
@@ -117,6 +217,125 @@ function* registerStandbyTransaction(context: TransactionContext, value: string,
...
+      contractKit.contracts,
+      contractKit.contracts.getEscrow,
+    ])
+
+    const escrowPaymentIds: string[] = yield call(
+      [escrowWrapper, escrowWrapper.getReceivedPaymentIds],
+      phoneHash
+    )
+
+    const msgHash = contractKit.web3.utils.soliditySha3({ type: 'address', value: walletAddress })
```
nategraf: Can we move this hashing call into `formEscrowWithdrawTxWithNoCode`? It would have helped me understand this flow on first pass.

packages/mobile/src/escrow/saga.ts
```diff
@@ -117,6 +217,125 @@ function* registerStandbyTransaction(context: TransactionContext, value: string,
...
+    const context = newTransactionContext(TAG, 'Withdraw from escrow')
+    // TODO: Batch the tranasctions and submit them together via `executeTransactions`
+    // method on an instance of the MTW then submitting like usual
+    const withdrawTxSuccess: boolean[] = []
+    for (let i = 0; i < escrowPaymentIds.length; i += 1) {
+      const withdrawTx: CeloTransactionObject<boolean> = yield formEscrowWithdrawTxWithNoCode(
```
nategraf: As mentioned above, the payment index can move if a payment is redeemed. With the current code, that would break our attempt to withdraw.

packages/mobile/src/escrow/saga.ts
```diff
@@ -117,6 +217,125 @@ function* registerStandbyTransaction(context: TransactionContext, value: string,
...
+        escrowWrapper,
+        escrowPaymentIds[i],
+        pepper,
+        msgHash,
+        i
+      )
+
+      try {
+        if (!komenciActive) {
+          yield call(sendTransaction, withdrawTx.txo, walletAddress, context)
```
nategraf: Should we have a balance check before getting to this point? In order to redeem the escrow without Komenci, they will need funds in their account already. The effect of this is that this transaction would fail at gas estimation.

packages/mobile/src/escrow/saga.ts
```diff
@@ -117,6 +217,125 @@ function* registerStandbyTransaction(context: TransactionContext, value: string,
...
+        } else {
+          const withdrawTxResult: Result<TransactionReceipt, FetchError | TxError> = yield call(
```
nategraf, line 339: We should probably ensure we have an active session here before we try this call? If a session is not open, can open a new session here?
One case I can imagine here is that the user closed the application after verification (e.g. left their phone on the table while they wait) and came back to it 15 minutes later when their session with Komenci is no longer valid.

packages/mobile/src/escrow/saga.ts
```diff
@@ -321,11 +539,21 @@ export function* watchFetchSentPayments() {
 export function* watchVerificationEnd() {
   while (true) {
     const update: SetVerificationStatusAction = yield take(IdentityActions.SET_VERIFICATION_STATUS)
+    const feelessUpdate: FeelessSetVerificationStatusAction = yield take(
+      IdentityActions.FEELESS_SET_VERIFICATION_STATUS
+    )
```
nategraf: You probably want a `race` between these two. Otherwise it will wait for _both_ and I imagine only one or the other will happen.

packages/mobile/src/invite/saga.ts
```diff
@@ -174,6 +176,7 @@ export function* sendInvite(
     const inviteCode = createInviteCode(temporaryWalletAccount.privateKey)
```
nategraf, line 181: If we are inviting without a code, we'd like to avoid generating the temporary account.

packages/mobile/src/invite/saga.ts
```diff
@@ -182,6 +185,12 @@ export function* sendInvite(
       }
     )
 
+    if (features.ESCROW_WITHOUT_CODE && amount) {
```
nategraf: If `ESCROW_WITHOUT_CODE` is enabled, but `amount` is `0` or `undefined` we should set the user up to send the invite text, the exit the flow. If we continue, it will transfer 25 cents to a random address, which is less   than ideal.

packages/utils/src/account.ts
```diff
@@ -75,12 +75,12 @@ export async function generateKeys(
 }
 
 export function generateDeterministicInviteCode(
```
nategraf, line 77: Once ARKE is available we plan to have users generate their own pepper (e.g. outside of ODIS). If we want this to be forward compatible, we should include in the hash the `phoneHash` to ensure that if two users pick the same pepper (potentially on purpose) the keys will still be unique.

nategraf (commented):
I reviewed the invite and escrow sagas and have some comments:
* I noticed an issue with how we derive the payment IDs that could result in collisions in an edge case. I haven't thought of a solution yet, but I'll spend some time noodling on it.
* I noticed that the only way to enter the flow to withdraw from escrow flow is immediately following verification. If so, this means that if a user goes through verification successfully, but fails escrow withdrawal and closes the app, the will have to reinstall the app and start over to withdraw from escrow. I wanted to check if you have the same understanding. It looks like this may be a preexisting issue that is more relevant now, and if so we see what the rest of the team thinks about what to do.
* IIUC users with `ESCROW_WITHOUT_CODE` enabled will not attempt to redeem escrow payments sent to them by users in the previous method. If at all possible, we should ensure that users can redeem payments sent to them by older versions of the app. (It is true that this would not result in loss of funds however because the sender can revoke the payment after 24 hours)

packages/mobile/src/escrow/saga.ts
nategraf: Ok, so I have an idea for what we can do here.

When we are creating a payment, we can simply try derivation indices repeatedly, starting at 0 and incrementing, until we hit one that is not used. So the process here would be to:
1. Retrieve all the `escrowPaymentIds` for a recipient and turn them into a set.
2. Set `index = 0`
3. Generate `paymentId = generateDeterministicInviteCode(pepper, index)`
4. If `paymentId` exists in `escrowPaymentIds`, increment `index` and repeat step 3

Withdrawals can be accomplished with a similar process to find the private key for a given payment. In order to prevent looping forever in case there is a payment not generated by this process in the list, some cap such as `1000` should be used for the index.

packages/mobile/src/escrow/saga.ts
tarikbellamine: We are only pulling in the `feelessVerificationState` because we need to instantiate a new KomenciKit instance. For a non-Komenci redemption (which is supported in the `if` block of the try/catch in this generator), there is no info needed from the regular `verificationState`.

What this doesn't support is redemption via the old method, which is a decision I'll address in another comment.

packages/mobile/src/escrow/saga.ts
tarikbellamine: Yep my bad

packages/mobile/src/escrow/saga.ts
tarikbellamine: I believe the assumption is that if they have enough funds needed to complete verification, then they will have enough remaining to pay for gas for this tx.

I agree that this is not a resilient approach. Given the breadth of changes, my goal was to achieve UX parity so, while it is a valid concern, I'm inclined to tackle this in the larger rework needed for the escrow redemption process within Valora.

packages/mobile/src/escrow/saga.ts
tarikbellamine: Good point

packages/mobile/src/invite/saga.ts
tarikbellamine: Good call

packages/utils/src/account.ts
tarikbellamine: I'm not super familiar with this so could you please double check I'm hashing correctly?

packages/mobile/src/escrow/saga.ts
tarikbellamine: That would be a more complete approach but I'm reluctant to do that until we've built a way for users to initiate a redemption outside of the automatic one done during the onboarding flow. While the edge case you pointed out is possible, in practice that would be very rare because Komenci has very lax session expiration criteria (I think it's the first of one week passing or >20 metatx's subsidized)

packages/mobile/src/escrow/saga.ts
tarikbellamine: Astute observation. This wasn't on my radar because the way escrow currently works (redeem all at once and then never again) doesn't lead to this issue. However, there is no guarantee things will remain this way (in fact, we really should make the escrow experience better) so definitely better to get in front of this.

You definitely saved future me a lot of debugging time here!

packages/mobile/src/escrow/saga.ts
tarikbellamine: Good point. Think I've addressed this now

tarikbellamine (commented):
> I noticed that the only way to enter the flow to withdraw from escrow flow is immediately following verification. If so, this means that if a user goes through verification successfully, but fails escrow withdrawal and closes the app, the will have to reinstall the app and start over to withdraw from escrow. I wanted to check if you have the same understanding. It looks like this may be a preexisting issue that is more relevant now, and if so we see what the rest of the team thinks about what to do.

Yeah, good observation. The UX of escrow redemption definitely needs work but Product is aware of this and does not consider it a GA blocker.

> IIUC users with ESCROW_WITHOUT_CODE enabled will not attempt to redeem escrow payments sent to them by users in the previous method. If at all possible, we should ensure that users can redeem payments sent to them by older versions of the app.

Another good observation. There was a conscious choice by Product to cut this when shedding scope to make GA feasible. Not ideal for existing users with funds out in escrow, but as you noted they can always reclaim and reissue.

packages/utils/src/account.ts
nategraf: Looks good to me!

packages/mobile/src/escrow/saga.ts
```diff
@@ -278,19 +281,28 @@ function* withdrawFromEscrowUsingPepper(komenciActive: boolean = false) {
...
     // method on an instance of the MTW then submitting like usual
     const withdrawTxSuccess: boolean[] = []
-    for (let i = 0; i < escrowPaymentIds.length; i += 1) {
+    // Using an upper bound of 1000 to be sure this doesn't run forever
+    for (let i = 0; i < 1000; i += 1) {
+      const { paymentId, privateKey } = generateEscrowPaymentIdAndPk(phoneHash, pepper, i)
+      if (!paymentIdSet.has(paymentId)) {
+        continue
+      }
+
```
nategraf: A quick suggestion to avoid looping 1000 times in the common case that all payments are processed.
```suggestion
    for (let i = 0; i < 1000 && paymentIdSet.size > 0; i += 1) {
      const { paymentId, privateKey } = generateEscrowPaymentIdAndPk(phoneHash, pepper, i)
      if (!paymentIdSet.has(paymentId)) {
        continue
      }
      paymentIdSet.delete(paymentId)

```

nategraf (approved):
Looks _great_ to me! 🙌

packages/mobile/src/escrow/saga.ts
```diff
@@ -117,6 +222,167 @@ function* registerStandbyTransaction(context: TransactionContext, value: string,
...
+        paymentId,
+        privateKey,
+        walletAddress,
+        mtwAddress,
+        value
+      )
+
+      try {
+        if (!komenciActive) {
+          yield call(sendTransaction, withdrawTx.txo, walletAddress, context)
```
asaj: IIUC this will result a call to `Escrow.withdraw` from the EOA, but it is only the MTW that has permission to withdraw, so this will fail.

asaj (dismissed):

asaj:
Also note that after withdrawing from escrow we need to then transfer the CUSD from the MTW to the EOA

asaj:
I pushed a fix to cmcewen/alf-build that I think should work

packages/mobile/src/escrow/saga.ts
tarikbellamine: I think this is already being done `formEscrowWithdrawTxWithNoCode`

asaj: For posterity @tarikbellamine is right, this would work b/c we are batching the txs via MTW

packages/mobile/src/escrow/saga.ts
```diff
@@ -117,6 +220,182 @@ function* registerStandbyTransaction(context: TransactionContext, value: string,
...
+  walletAddress: string,
+  metaTxWalletAddress: string,
+  value: BigNumber
+) {
+  const msgHash = contractKit.web3.utils.soliditySha3({
+    type: 'address',
+    value: metaTxWalletAddress,
+  })
+
+  const signature: string = (await contractKit.web3.eth.accounts.sign(msgHash, privateKey))
```
asaj: The object returned by this function has the following fields: 
```
{ message, messageHash, v, r, s, signature }
```

Why do we only take the `signature` field, just to split it up into r, s, and v? Suggest we just read those directly

packages/mobile/locales/en-US/sendFlow7.json
```diff
@@ -48,6 +48,7 @@
   "walletAddress": "Wallet Address",
   "inviteWithoutPayment": "Hi {{name}}! I’ve been using {{appName}} to send money worldwide. It’s easy to try and I want to invite you to check it out with this link: {{link}}",
   "inviteWithEscrowedPayment": "Hi {{name}}! I’ve been using {{appName}} to send money worldwide. I just sent you {{amount}} Celo Dollars. Please download the {{appName}} App and redeem your Celo Dollars using this link: {{link}}",
+  "inviteWithEscrowAndNoCode": "Welcome to Valora, {{name}}! I've sent you Celo Dollars. Go to {{link}} to get started",
```
gnardini: Is there a reason to use this and not `inviteWithEscrowedPayment`? The wording on the other string is the one on [this doc](https://docs.google.com/document/d/1DftgFh0PShTWjynMAbGzaoDOknh0MGSexfVCuJyPgc4/edit#heading=h.yrdzg1i0xig3). Also, the `{{name}}` doesn't actually go here, since we're using the name of the user using the app, which doesn't make a lot of sense. It should be removed from the other strings in master.

packages/mobile/src/invite/saga.ts
```diff
@@ -182,15 +183,32 @@ export function* sendInvite(
...
-      {
-        name,
-        amount: amount?.toString(),
-        link,
+    const link = features.ESCROW_WITHOUT_CODE
+      ? VALORA_DOMAIN
+      : yield call(generateInviteLink, inviteCode)
+
+    let messageProp
+    if (features.ESCROW_WITHOUT_CODE && amount) {
```
gnardini: I don't think the `features.ESCROW_WITHOUT_CODE` should affect the string used, we should still use one of the other two.

packages/mobile/src/config.ts
```diff
@@ -135,3 +135,4 @@ export const SIMPLEX_URI = 'https://valoraapp.com/simplex'
 
 export const APP_STORE_ID = Config.APP_STORE_ID
 export const DYNAMIC_LINK_DOMAIN = Config.DYNAMIC_LINK_DOMAIN
+export const VALORA_DOMAIN = 'http://valoraapp.com'
```
gnardini: Use https?

packages/mobile/src/invite/saga.ts
```diff
@@ -182,15 +183,32 @@ export function* sendInvite(
...
+      amount: amount?.toString(),
+      link,
+    })
+
+    if (features.ESCROW_WITHOUT_CODE) {
+      if (amount) {
+        yield call(initiateEscrowTransfer, e164Number, amount)
       }
-    )
+      yield call(navigateToInviteMessageApp, e164Number, inviteMode, message)
```
gnardini: This `navigateToInviteMessageApp` uses SMS by default, not the native share. We can just do `yield call(Share.share, { message })` to open the native share dialog and let users choose how to share.

gnardini (commented):
Looking good! A few small things regarding sharing :)

packages/mobile/src/config.ts
jeanregisser: We already have a config for this: `WEB_LINK` in our [brandingConfig](https://github.com/clabs-co/valora-app-branding/blob/4c00eee947ee419709dcf76a50af982cb988cd16/src/brandingConfig.ts#L2).

packages/mobile/src/escrow/utils.ts
```diff
@@ -11,11 +11,35 @@ export const splitSignature = (contractKit: ContractKit, signature: string) => {
...
+
+export const generateUniquePaymentId = (
+  existingPaymentIds: string[],
+  phoneHash: string,
+  pepper: string
+) => {
+  const paymentIdSet: Set<string> = new Set()
+  for (const paymentId of existingPaymentIds) {
+    paymentIdSet.add(paymentId)
+  }
```
jeanregisser:
```suggestion
  const paymentIdSet: Set<string> = new Set(existingPaymentIds)
```

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -1029,6 +1036,13 @@ export function* feelessCompleteAttestation(
 
   Logger.debug(TAG + '@feelessCompleteAttestation', `Completing code for issuer: ${code.issuer}`)
 
+  // Needed to ensure that codes inputted in quick succession are not
+  // assigned the same nonce by Komenci
+  while (yield select(feelessProcessingInputCodeSelector)) {
+    yield delay(100)
```
jeanregisser, line 1058: Is the delay actually important or are we just interested in reacting to redux state changes?
in that case we could replace the delay by `yield take()` which would wait for any dispatched action.

packages/mobile/locales/en-US/sendFlow7.json
jeanregisser: Yes this was changed in master today.

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -72,17 +76,24 @@ function VerificationEducationScreen({ route, navigation }: Props) {
...
   useEffect(() => {
-    if (verificationState.status.isVerified || feelessVerificationState.status.isVerified) {
+    dispatch(initializeAccount())
+  }, [])
+
+  useEffect(() => {
+    if (status.isVerified) {
       dispatch(setNumberVerified(true))
     }
   }, [verificationState.status.isVerified, feelessVerificationState.status.isVerified])
```
jeanregisser, line 87:
```suggestion
  }, [status.isVerified])
```

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -124,7 +130,7 @@ export function* feelessFetchVerificationState() {
     ])
 
     const komenciKit = new KomenciKit(contractKit, walletAddress, {
-      url: KOMENCI_URL,
+      url: feelessVerificationState.komenci.callbackUrl || KOMENCI_URL,
```
jeanregisser: Do we still need the fallback?

packages/mobile/src/identity/feelessVerification.ts
```diff
@@ -92,7 +98,7 @@ import { TransactionReceipt } from 'web3-eth'
 
 const TAG = 'identity/feelessVerification'
 
-const KOMENCI_URL = 'https://komenci.celo-networks-dev.org'
+export const KOMENCI_URL = 'https://staging-komenci.azurefd.net'
```
jeanregisser: This is only temporary right?

packages/mobile/src/escrow/saga.ts
tarikbellamine: Good call. It was a blind copy/paste.

packages/mobile/src/send/SendAmount.tsx
```diff
@@ -189,6 +181,29 @@ function SendAmount(props: Props) {
...
+            maxAmount: maxAmount?.toFixed(2),
+            symbol: localCurrencySymbol,
+          })
+        )
+      } else {
+        isOutgoingPaymentRequest ? onRequest() : onSend()
+      }
+      setReviewButtonPressed(false)
+    }
+  }, [reviewButtonPressed, recipientVerificationStatus])
```
jeanregisser, line 205: There's a slight bug here since `localCurrencyExchangeRate` and `localCurrencySymbol` are not the in dependencies list of `useEffect`. Though in practice those values won't change at this point.
Adding them to the dependencies list is not enough though, we'd also need to make sure the action is not dispatched twice. Which would happen if other values change.

jeanregisser (commented):
Looking great! 👍 🚀 
Mostly small things commented.

packages/mobile/src/config.ts
```diff
@@ -135,3 +136,19 @@ export const SIMPLEX_URI = 'https://valoraapp.com/simplex'
...
+
+export const ALLOWED_MTW_IMPLEMENTATIONS: Address[] =
+  DEFAULT_TESTNET === 'mainnet'
+    ? ['0x6511FB5DBfe95859d8759AdAd5503D656E2555d7']
+    : ['0x5C9a6E3c3E862eD306E2E3348EBC8b8310A99e5A', '0x88a2b9B8387A1823D821E406b4e951337fa1D46D']
+
+export const CURRENT_MTW_IMPLEMENTATION_ADDRESS: Address =
+  DEFAULT_TESTNET === 'mainnet'
+    ? '0x6511FB5DBfe95859d8759AdAd5503D656E2555d7'
+    : '0x5C9a6E3c3E862eD306E2E3348EBC8b8310A99e5A'
```
jeanregisser: nit: these network specific configs should be moved to https://github.com/celo-org/celo-monorepo/blob/master/packages/mobile/src/geth/networkConfig.ts

packages/mobile/src/escrow/utils.ts
tarikbellamine: oh nice, i cant believe i didnt know this!

packages/mobile/src/verify/VerificationEducationScreen.tsx
tarikbellamine: I think we want to leave this as is. There is an edge case where a user is allowed to `tryFeeless` but they've already verified in the non-feeless flow.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: Yes, this has to do with data center routing.

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: The placement was temporary, the url is not. That will be our staging URL moving forward.

packages/mobile/src/send/SendAmount.tsx
tarikbellamine: True, but I feel pretty comfortable leaving it as is. Nothing bad happens if the user is over the limit here; it's an artificial constraint we are placing as an added security measure. There is also no way to change `localCurrency` on the screen.

Is there a use case you had in mind where this could be problematic?

packages/mobile/src/identity/feelessVerification.ts
tarikbellamine: It may not be as simple as `yield take` given we want them to be taken one-by-one and there may be 2 codes "waiting". I've left a `TODO` to look into this later.

packages/mobile/src/verify/VerificationEducationScreen.tsx
jeanregisser: I understand we might not want to change this at the last minute but the dependencies listed should only be values that are used in the useEffect block.

packages/mobile/src/send/SendAmount.tsx
jeanregisser: Again it's harmless, `localCurrency` doesn't change but `localCurrencyExchangeRate` can change.
The point is that it's not entirely right and introduces a subtle bug. Which might or might not bite us in the future ;)

Lss-Ankit:
Hi @tarikbellamine Verified task on latest test flight build v1.5.5 (34) and Android play store internal build v1.5.5 (1004294323) and observe the followings
- User A is able to invite the userB with escrow payment
- User B redirected to app store OR play-store when tapping on received URL
However, user is not redirected to latest v1.5.5 builds even app in installed 
- user get blocked to install the latest build v1.5.5 by tapping on received invite URL link

Can you please provide more information about this task.

Lss-Ankit:
Hi @tarikbellamine Verified task on latest test flight build v1.5.5 (34) and Android play store internal build v1.5.5 (1004294323) and observe the followings
- User A is able to invite the userB with escrow payment
- User B redirected to app store OR play-store when tapping on received URL
However, user is not redirected to latest v1.5.5 builds even app in installed 
- user get blocked to install the latest build v1.5.5 by tapping on received invite URL link

Can you please provide more information about this task.

PR #5754, opened 2020-11-06
Switch away from release-v2

nambrot:
### Description

We deleted the tag for v2 since it's likely going to be a no-op, so let's switch back to v1


nambrot:
The `protocol-test-release` test will fail right now until https://github.com/celo-org/celo-monorepo/pull/5715 properly tells us to release implementing contracts when underlying dependencies change.

m-chrzan (approved):
Looks good

PR #5759, opened 2020-11-07
Update to new Baklava network id

k0kk0k:
Update to new Baklava network id


trianglesphere:
It looks like there was a recent docs update that removed the network ID. Thanks for flagging this.

PR #5760, opened 2020-11-07
upgrade celo blockchain to v1.1.0

alchemydc:
### Description

Upgrades celo blockchain image to v1.1.0.

### Other changes

Removed references to genesis.json and bootnodes, as these are now dynamically derived from the docker images themselves.

Improve attestation service docs

### Tested

On mainnet with elected validator.

### Related issues

n/a

### Backwards compatibility

n/a


packages/terraform-modules-public/gcp/example/README.md
```diff
@@ -299,6 +299,176 @@
...
+    Now, on a system which has access to the attestation_signer private key, generate a proof of possession for that key as follows:
+
+    ```bash
+    #!/bin/bash
+    set -x
+
+    ######
+    # use this script on an attestation signer tx-node to generate a proof of possession, needed for key rotation
+
+    CELO_IMAGE=us.gcr.io/celo-org/celo-node:f4bda6d027f0af6c68ae4d146abe791739ae6cba
```
tkporter: What do you think of `us.gcr.io/celo-org/geth:1.1.0`?

packages/terraform-modules-public/gcp/testnet/modules/proxy/startup.sh
```diff
@@ -347,8 +347,8 @@ IN_MEMORY_DISCOVERY_TABLE_FLAG=""
 
 # Load configuration to files
 mkdir -p $DATA_DIR/account
-echo -n '${genesis_content_base64}' | base64 -d > $DATA_DIR/genesis.json
-echo -n '${bootnodes_base64}' | base64 -d > $DATA_DIR/bootnodes
+#echo -n '${genesis_content_base64}' | base64 -d > $DATA_DIR/genesis.json
```
tkporter, line 352: delete if not needed? or maybe a comment explaining why these comments & the others are still included if there's a reason

tkporter (approved):
Thanks! Couple tiny things

packages/terraform-modules-public/gcp/testnet/modules/proxy/startup.sh
alchemydc: Left these in here because handling of genesis block and bootnodes changed significantly in 1.1.0, and there's some ongoing disagreement about how to handle it going forward.  I clarified the reasoning in a comment.

packages/terraform-modules-public/gcp/example/README.md
alchemydc: Right.  Of course.

Issue #5435, opened 2020-10-19
Scope Cash In/Out integrations for local providers


jeanregisser:
- We need to see how to integrate them. maybe they should do this by directly contributing  in a open source way
- Do they need a license? proven in some way.
- Country-specific flags to show integrations

PR #5764, opened 2020-11-09
[Wallet] Add local cashout providers behind remote feature flag

annakaz:
### Description

Add local providers
- Ponto (cUSD -> GCash in Philippines)
- Kotani (cUSD -> M-Pesa in Kenya) 
which are behind a remote feature flag

Remote feature flag is configured from firebase the same way we've set up minAppVersion. This allows us to remotely toggle on these providers once they are live.

Note that both provider URIs that are linked are still being completed by the providers

### Tested

Toggled feature flag in alfajores firebase project 

### Related issues

- Part of #5435 

### Backwards compatibility

Yes


packages/mobile/src/app/actions.ts
```diff
@@ -99,6 +101,16 @@ interface MinAppVersionDeterminedAction {
...
+interface PontoFeatureFlagSetAction {
+  type: Actions.SET_PONTO_FEATURE_FLAG
+  enabled: boolean
+}
+
+interface KotaniFeatureFlagSetAction {
+  type: Actions.SET_KOTANI_FEATURE_FLAG
+  enabled: boolean
+}
+
```
jeanregisser, line 113: Fine for now given the short timeline we have, but I think this could be more generic and handle all remote feature flags in a single "REMOTE_FEATURE_FLAGS_UPDATED" action.

packages/mobile/src/app/reducers.ts
```diff
@@ -16,6 +16,8 @@ export interface State {
   lastTimeBackgrounded: number
   sessionId: string
   minVersion: string | null
+  pontoEnabled: boolean
+  kotaniEnabled: boolean
```
jeanregisser, line 20: No need to do anything now, we'll probably store remote feature flags in a dedicated reducer.

jeanregisser (approved):
Awesome! 👍

i1skn (approved):
LGTM 👍

PR #5779, opened 2020-11-10
[cli] Transfers cmd: Add verification using an initial small Tx

gastonponti:
### Description

Transfer celo and dollars commands, have a new flag that allows the user to verify with a small tx (< 1 CELO, < 1 cUSD), that the user received the transaction. The amount of the 1st Tx, will be asked, and that input will allow to send the rest of the value of the command

### Tested

In alfajores, manually

### Backwards compatibility

yes


packages/cli/src/commands/transfer/celo.ts
```diff
@@ -35,6 +42,40 @@ export default class TransferCelo extends BaseCommand {
       .hasEnoughCelo(from, value)
       .runChecks()
 
+    if (res.flags.withTxVerification && value.isGreaterThanOrEqualTo(new BigNumber('200e18'))) {
+      const firstTxValue = BigNumber.random(18).multipliedBy('1e18')
+
+      cli.action.start(`Sending Verifiying Transaction...`)
+      const txResult = await celoToken
+        .transferWithComment(to, firstTxValue.toFixed(), 'Verificable Tx')
```
barbaraliau, line 50: Should the third arg be 'Verifiable Tx' or 'Verification Tx'?

packages/cli/src/commands/transfer/celo.ts
```diff
@@ -35,6 +42,40 @@ export default class TransferCelo extends BaseCommand {
...
+        .send()
+
+      await txResult.waitReceipt()
+      cli.action.stop()
+
+      let retries = 3
+      let preMessage = ''
+      do {
```
barbaraliau, line 58: I'm looking at this on github so it might just be weird, but is the spacing off for this?

packages/cli/src/commands/transfer/dollars.ts
```diff
@@ -53,6 +75,36 @@ export default class TransferDollars extends BaseCommand {
       )
       .runChecks()
 
+    if (requireTwoTx) {
+      cli.action.start(`Sending Verifiying Transaction...`)
```
barbaraliau, line 79: typo: 'Verifying"

packages/cli/src/commands/transfer/dollars.ts
```diff
@@ -53,6 +75,36 @@ export default class TransferDollars extends BaseCommand {
...
+        .send()
+
+      await txResult.waitReceipt()
+      cli.action.stop()
+
+      let retries = 3
+      let preMessage = ''
+      do {
+        const response: string = await cli.prompt(
+          `${preMessage}Insert the value in wei of the transaction used as verifier`,
```
barbaraliau, line 91: Should this be wei if they are sending dollars?

packages/cli/src/commands/transfer/celo.ts
gastonponti: you mean the `preMessage`?
the second `preMessage` (the one for the retries) has it

packages/cli/src/commands/transfer/dollars.ts
gastonponti: well, even if they are dollars, 1cUSD = 1x(10^18)
Is this particular case, I was not completely sure about the usage of this. If you are sending this to a Valora, for example, the user will see it as 0.xx (even trimming the other numbers, I don't know). But If you are checking this from the `cli` of the explorer you can get that number in wei.

But we could do, is to reduce the random number from 0.01 to 1 (this will even make it easier for the user to text it), and ask the user for the cUSD amount as you said before. Same with the CELO command, something between 1 and 0.01 CELO.
We will be losing a LOT of possibilities, but I don't think that this command targets the technical user (who could always check the explorer and retrieve the number without asking the other user)

gastonponti: What I would like is to unify both. Don't want to have one in cUSD and the other in wei. WDYT? Should I reduce the random and ask for a 1e18 ether/celo representation? I think that it will be easier for the common user

barbaraliau: I think the target is non-technical as you mentioned, and if it is for verification then keeping it as either cUSD or wei is a lot easier for the user. It would be nice to have this be a small transaction, so we want to have more than two decimal places. 1e18 ether/celo makes sense.

PR #5786, opened 2020-11-10
Komenci Deployer

codyborn:
### Description

Merging @eruizgar91's changes into master. 
This PR introduces Komenci deployment capabilities into Celotool. Includes configuration for 3 datacenters (EastUS, WestEurope, BrazilSouth). Uses KeyVault for secret management.

### Tested

Used for deployment to staging and mainnet. Tested deploying Oracles.

### Backwards compatibility

No changes have been made to other service deployments.


.env.komenci
```diff
@@ -0,0 +1,207 @@
+KOMENCI_DOCKER_IMAGE_REPOSITORY="celotestnet.azurecr.io/komenci/komenci"
+KOMENCI_DOCKER_IMAGE_TAG="8276e69c1e80eeab8e89e296c0dc42c920e90cb7"
```
bowd: I keep thinking that we should have this per cluster because it gets weird if we're deploying different versions to each cluster. And we also don't have a clear record. I'm not sure, for example, what version is up in brasil right now.

.env.alfajores
```diff
@@ -110,3 +118,48 @@ MOBILE_WALLET_PLAYSTORE_LINK="https://play.google.com/apps/internaltest/47009904
 
 PROMTOSD_SCRAPE_INTERVAL="5m"
 PROMTOSD_EXPORT_INTERVAL="5m"
+
+# each context should have its own environment variables, generally of the form
+# <context in all caps and underscores>_*
+CONTEXTS=azure-eastus,azure-westeuAZURE_EASTUS_KOMENCI_KUBERNETES_CLUSTER_NAME
```
bowd: Looks like some editing mishap, the `AZURE_EASTUS_KOMENCI_KUBERNETES_CLUSTER_NAME` should be it's own thing?

.env.alfajores
```diff
@@ -110,3 +118,48 @@ MOBILE_WALLET_PLAYSTORE_LINK="https://play.google.com/apps/internaltest/47009904
...
+
+# --- Komenci ---
+
+KOMENCI_DOCKER_IMAGE_REPOSITORY="celotestnet.azurecr.io/komenci/komenci"
+KOMENCI_DOCKER_IMAGE_TAG="8276e69c1e80eeab8e89e296c0dc42c920e90cb7"
```
bowd: These are also defined in the `.env.komenci` file, not sure how the system works but it seams like one would override the other unless the env variables are prefixed.

.env.komenci
```diff
@@ -0,0 +1,207 @@
...
+
+ORACLE_UNUSED_ORACLE_ADDRESSES=
```
bowd: Remove?

.env.komenci
```diff
@@ -0,0 +1,207 @@
+KOMENCI_DOCKER_IMAGE_REPOSITORY="celotestnet.azurecr.io/komenci/komenci"
```
bowd: Still a bit confused about what should live in .env.komenci vs waht should live in .env.rc1 / .env.alfajores

.env.komenci
```diff
@@ -0,0 +1,207 @@
...
+GCP_TEST_FULL_NODES_COUNT=1
+GCP_TEST_FULL_NODES_DISK_SIZE=10
+
+GCP_TEST_ASIA_GCP_PROJECT_NAME=celo-testnet
+GCP_TEST_ASIA_GCP_ZONE=asia-northeast2-a
+GCP_TEST_ASIA_KUBERNETES_CLUSTER_NAME=federated-dev-asia-northeast2-a
+GCP_TEST_ASIA_FULL_NODES_COUNT=1
+GCP_TEST_ASIA_FULL_NODES_DISK_SIZE=10
+
+# ---- General ----
```
bowd: Everything from here down seams dangerous in the sense that I'm not sure why it lives here and if it clashes with some other env vars in some other files and if they override each other.

bowd (commented):
Left a few comments bellow on parts I don't fully understand and I also want to take this opportunity to voice two qualms I have with our approach to infra - but these are general, not really this PR specific:

1. I think that we should decouple regions from environments, we're making a mental tie between EASTUS / WESTEU to staging (alfajores) and SOUTHBR to production (mainnet) which I thought was a bit weird from the beginning. What if I want a mainnet deployment in WESTEU I need to extend my naming strategy.

2. I think that the celotool deployment helper is interesting but I feel that it's grown a bit too complex and it might be doing more than it should. For example I think it's saner / easier to manage if we actually had different `values.yaml` files for each target cluster / env you want to deploy to instead of having that data in ENV vars and then overriding that one file in the chart. Especially because there are ways to tie k8s secret management to Azure Key Vaults to eliminate the need to download secrets during deployment entirely. Also the different things that need to be setup in the cluster feel like they might be better suited for terraform.

.env.alfajores
codyborn: thanks, I must have pasted it by mistake

.env.komenci
codyborn: Sorry for the confusion, this file is being deprecated in favor of the per environment one. In this case all of clusters in the same environment (alfajores | rc1) will have the same image.

bowd (approved):

PR #5790, opened 2020-11-10
Protect against beneficiaries that are smart contracts

### LINKED ISSUE #5452: ReleaseGold script should check that the beneficiary is not a smart contract

nambrot:
### Description

beneficiaries that cannot call functions on the `ReleaseGold` smart contracts might cause loss of funds. This adds a check in a script to highlight if a beneficiary is a smart contract itself which likely means that a wrong address was copied

### Tested

- Not tested

### Related issues

- Fixes #5452


packages/protocol/scripts/truffle/deploy_release_contracts.ts
```diff
@@ -89,6 +89,20 @@ async function handleGrant(config: ReleaseGoldConfig, currGrant: number) {
...
+  const bytecode = await web3.eth.getCode(config.beneficiary)
+  if (bytecode !== '0x') {
+    const response = await prompts({
+      type: 'confirm',
+      name: 'confirmation',
+      message: `Beneficiary ${config.beneficiary} is a smart contract which might cause loss of funds if not properly configured. Are you sure you want to continue? (y/n)`,
+    })
+
+    if (!response.confirmation) {
+      console.info(chalk.yellow('Skipping grant due to user response'))
```
asaj, line 101: Can we log the identifier as part of this? And/or write it (them) to a file?

asaj (approved):

packages/protocol/scripts/truffle/deploy_release_contracts.ts
nambrot: It's part of the previous message (which I just copied from the other prompt below)?

PR #5795, opened 2020-11-10
Adjust migrations for Testnet contract deploys

### LINKED ISSUE #5727: contracts migrations should work with v1

nambrot:
### Description

Over time, our migrationsConfig have become out of sync and thus the current state does not yield a satisfactory experience for deploying contracts on a real testnet (i.e. not ganache). As I've been working more on env-test, I also adjusted the default migration overrides to make some special keys more accessible on these testnets.

As a result, testnet contract deploys should work out of the box, and an upcoming PR for env-tests will make running these tests work out of the box as well.

### Tested

- On my own testnets

Fixes #5727


packages/protocol/migrationsConfig.js
```diff
@@ -506,10 +430,139 @@ const NetworkConfigs = {
       votesRatioOfLastVsFirstGroup: 1.0,
     },
   },
+  rc1: {
```
nambrot, line 434: This is copied for the most part from the old Default config

yorhodes (approved):
Exciting!

PR #5797, opened 2020-11-10
Add Oracle/Reserve/Attestation to env-tests

### LINKED ISSUE #5728: env-tests should cover attestations

nambrot:
### Description

This PR adds tests for Oracle reporting, reserve spending and requesting/completing attestations

### Tested

- Can be run with `CONFIG=../../.env.mnemonic yarn test`

### Related issues

- Fixes #5728


ewilz:
What is the scope of the tests in this package? Back-end integration tests of all contract services?

nambrot:
Indeed. When looking for testing whether a release is successful on a testnet, I realized that none of our testing is able to easily expand to this use case. So I sought out to create a test suite that can be run against environments. I hope that it can eventually be run in the same context as the e2e-tests in celotool, or as nightly tests on public testnets

packages/env-tests/src/shared/attestation.ts
```diff
@@ -0,0 +1,256 @@
...
+export async function requestMoreAttestations(
+  attestations: AttestationsWrapper,
+  phoneNumber: string,
+  attestationsRequested: number,
+  account: Address,
+  txParams: CeloTransactionParams = {}
+) {
+  const unselectedRequest = await attestations.getUnselectedRequest(phoneNumber, account)
+  if (
+    unselectedRequest.blockNumber === 0 ||
```
yorhodes, line 59: what does this mean?

packages/env-tests/src/shared/attestation.ts
```diff
@@ -0,0 +1,256 @@
...
+    (await attestations.isAttestationExpired(unselectedRequest.blockNumber))
+  ) {
+    await attestations
+      .approveAttestationFee(attestationsRequested)
+      .then((txo) => txo.sendAndWaitForReceipt(txParams))
```
yorhodes, line 64: nit: don't mix .then with async/await syntax

packages/env-tests/src/shared/attestation.ts
```diff
@@ -0,0 +1,256 @@
...
+        return {
+          status: response.status,
+          text: await response.text(),
+          issuer: attestation.issuer,
+          name: attestation.name,
+          known: true,
+        }
+      }
+
+      return
```
yorhodes, line 105: this scoping is confusing

packages/env-tests/src/shared/attestation.ts
```diff
@@ -0,0 +1,256 @@
...
+// Could mark already checked messages in the future
+async function findValidCode(
+  attestations: AttestationsWrapper,
+  messages: string[],
+  identifier: string,
+  attestationsToComplete: ActionableAttestation[],
+  account: string
+) {
+  for (const message of messages) {
+    try {
```
yorhodes, line 137: it's not clear to me what this function does and the `try { if (!x) continue } catch { continue }` smells at a glance

packages/env-tests/src/shared/attestation.ts
```diff
@@ -0,0 +1,256 @@
...
+  return usableNumber
+}
+
+async function fetchLatestMessagesFromToday(client: Twilio, phoneNumber: string, count: number) {
+  return client.messages.list({
+    to: phoneNumber,
+    pageSize: count,
+    // Twilio keeps track of dates in UTC so it could be yesterday too
+    dateSentAfter: moment()
+      .subtract(2, 'day')
```
yorhodes, line 207: yesterday is 2 days ago?

packages/env-tests/src/shared/attestation.ts
```diff
@@ -0,0 +1,256 @@
...
+  phoneNumber: string,
+  identifier: string,
+  account: Address,
+  attestationsToComplete: ActionableAttestation[],
+  txParams: CeloTransactionParams = {},
+  logger: Logger,
+  timeToPollForTextMessages: number
+) {
+  const startDate = moment()
+  logger.info({ pollingWait: POLLING_WAIT }, 'Poll for the attestation code')
```
yorhodes, line 224: this is a constant log?

packages/env-tests/src/shared/attestation.ts
```diff
@@ -0,0 +1,256 @@
...
+      account
+    )
+
+    if (!res) {
+      await sleep(POLLING_WAIT)
+      continue
+    }
+
+    logger.info(
+      { waitingTime: moment.duration(moment().diff(startDate)).asSeconds() },
```
yorhodes, line 245: probably not useful but shameless plug for `secondsToDurationString`

packages/env-tests/src/shared/attestation.ts
```diff
@@ -0,0 +1,256 @@
...
+      'Received valid code'
+    )
+
+    const completeTx = await attestations.complete(identifier, account, res.issuer, res.code)
+
+    await completeTx.sendAndWaitForReceipt(txParams)
+
+    logger.info({ issuer: res.issuer }, 'Completed attestation')
```
yorhodes: should we log info about specific attestation?

packages/env-tests/src/tests/attestation.ts
```diff
@@ -0,0 +1,137 @@
...
+  getIdentifierAndPepper,
+  getPhoneNumber,
+  pollForMessagesAndCompleteAttestations,
+  reportErrors,
+  requestAttestationsFromIssuers,
+  requestMoreAttestations,
+} from '../shared/attestation'
+
+async function fundAttestationAccount(
+  context: Context,
```
yorhodes: if we have context type `Context`, which imo should have a better name, we don't also need stableTokenWrapper as a parameter

packages/env-tests/src/tests/attestation.ts
```diff
@@ -0,0 +1,137 @@
...
+  stableToken: StableTokenWrapper,
+  root: string,
+  from: string,
+  logger: Logger
+) {
+  context.kit.defaultFeeCurrency = stableToken.address
+
+  const rootTransfer = stableToken.transfer(from, ONE.toString())
+  const rootTransferReceipt = await rootTransfer.sendAndWaitForReceipt({ from: root })
```
yorhodes, line 30: what's the difference between `from` and `root`
I'm confused

packages/env-tests/src/tests/attestation.ts
```diff
@@ -0,0 +1,137 @@
...
+  describe('Attestation Test', () => {
+    const logger = context.logger.child({ test: 'attestation' })
+
+    // TODO: Check for TWILIO ENV vars
+    test('receive attestations for a phone number', async () => {
+      const root = await getKey(context.mnemonic, TestAccounts.Root)
+      context.kit.addAccount(root.privateKey)
+
+      const mnemonic = await generateMnemonic()
+      const from = await generateKeys(mnemonic)
```
yorhodes: this is slow
do we need a new one everytime?

packages/env-tests/src/tests/attestation.ts
```diff
@@ -0,0 +1,137 @@
...
+
+      const attestationsToComplete = await attestations.getActionableAttestations(
+        identifier,
+        fromAddress
+      )
+
+      logger.info({ attestationsToComplete }, 'Reveal to issuers')
+
+      // Wait for attestation services to sync
+      await sleep(5000)
```
yorhodes, line 100: this feels fragile

packages/env-tests/src/tests/attestation.ts
```diff
@@ -0,0 +1,137 @@
...
+        fetchEnv(envVar.TWILIO_ACCOUNT_AUTH_TOKEN)
+      )
+
+      const phoneNumber = await getPhoneNumber(twilioClient, fetchEnv(envVar.TWILIO_ADDRESS_SID))
+
+      const { identifier, pepper } = await getIdentifierAndPepper(
+        context.kit,
+        odisContext,
+        fromAddress,
+        phoneNumber.phoneNumber,
```
yorhodes, line 75: so phoneNumber is not actually phoneNumber?

packages/env-tests/src/tests/attestation.ts
```diff
@@ -0,0 +1,137 @@
...
+        fromAddress,
+        attestationsToComplete,
+        {},
+        logger,
+        attetationsToRequest
+      )
+
+      const attestationStat = await attestations.getAttestationStat(identifier, fromAddress)
+
+      expect(attestationStat.total).toEqual(attetationsToRequest)
```
yorhodes: typo: attetations => attestations

packages/env-tests/src/tests/attestation.ts
```diff
@@ -0,0 +1,137 @@
...
+      expect(attestationStat.completed).toEqual(attetationsToRequest)
+
+      // move back funds to root
+      const stableBalanceRoot = await stableToken.balanceOf(fromAddress)
+      const transferBackTx = stableToken.transfer(
+        root.address,
+        stableBalanceRoot
+          .times(0.95)
```
yorhodes, line 137: is this to ensure enough funds?

packages/env-tests/src/tests/oracle.ts
```diff
@@ -0,0 +1,60 @@
...
+import { fundAccount, getKey, ONE, TestAccounts } from '../scaffold'
+
+export function runOracleTest(context: Context) {
+  describe('Oracle Test', () => {
+    const logger = context.logger.child({ test: 'exchange' })
+    beforeAll(async () => {
+      await fundAccount(context, TestAccounts.Exchange, ONE.times(2))
+    })
+
+    // TODO: Check if oracle account is authorized
```
yorhodes, line 14: is this done with `isOracle`?

packages/env-tests/src/tests/oracle.ts
```diff
@@ -0,0 +1,60 @@
...
+          },
+          'no existing rate, using the median'
+        )
+      } else {
+        rateToReport = ourRate.rate
+        logger.debug({ rate: ourRate.rate.toString() }, 'fetched existing oracle report')
+      }
+
+      // Move the rate in one direction or another
+      rateToReport = Math.random() < 0.5 ? rateToReport.times(1.05) : rateToReport.times(0.95)
```
yorhodes:
```suggestion
      rateToReport = rateToReport.times(0.95 + Math.random() * 0.10)
```

packages/env-tests/src/tests/oracle.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      const reportTx = await oracles.report(CeloContract.StableToken, rateToReport, from.address)
+      const reportTxReceipt = await reportTx.sendAndWaitForReceipt({ from: from.address })
+      logger.debug({ receipt: reportTxReceipt }, 'rate reported')
+
+      const newOracleRates = await oracles.getReports(CeloContract.StableToken)
+      const ourNewRate = newOracleRates.find((_) => _.address === from.address)
```
yorhodes, line 54: this seems pretty obvious
should we add a `getOracleRate(oracle, token)` to wrapper?

packages/env-tests/src/tests/reserve.ts
```diff
@@ -0,0 +1,69 @@
...
+        .times(transferRatio)
+        .integerValue(BigNumber.ROUND_DOWN)
+      const spenderTx = reserve.transferGold(custodian.address, valueToTransfer.toString())
+      const multiSigTx = await spenderMultiSig.submitOrConfirmTransaction(
+        reserve.address,
+        spenderTx.txo
+      )
+      logger.debug(
+        {
+          data: spenderTx.txo.encodeABI(),
```
yorhodes, line 50: is the data more useful than the tx in the logs?

yorhodes (commented):
Love it 🚀

packages/env-tests/src/shared/attestation.ts
nambrot: Copied this from celotools attestation bot. This basically means if there is no unselected request

packages/env-tests/src/shared/attestation.ts
nambrot: Also copied, but IMO i prefer this over a double await or a "unnecessary variable assignment", wdyt?

packages/env-tests/src/shared/attestation.ts
nambrot: Would you mind elaborating?

packages/env-tests/src/shared/attestation.ts
nambrot: So this code tries to find a valid message by iterating over the messages and applying various "filters". Ideally a `map/filter` is probably "better", but that can't be done lazily, so this seemed preferrable

packages/env-tests/src/shared/attestation.ts
nambrot: Copied from celotool again, but I believe yesterday is the day after the day two days ago :P

packages/env-tests/src/shared/attestation.ts
nambrot: Yeah, just to indicate to the log viewer what the constant value is

packages/env-tests/src/shared/attestation.ts
nambrot: I think that is great for human output, but this is meant as structured logging :)

packages/env-tests/src/tests/attestation.ts
nambrot: Do you know how slow SMS sends are :P

Since the effects of the attestation tests persists between invocations, I wanted to avoid a scenario in which a identifier/account combo already exhausted all issuers. Thus the easiest thing is to just create a new account each time.

packages/env-tests/src/tests/attestation.ts
nambrot: It's a twilio phone number object

packages/env-tests/src/tests/attestation.ts
nambrot: Indeed :(
Added a todo

packages/env-tests/src/tests/attestation.ts
nambrot: Practically yes, theoretically no. I guess it would be nice to have functions on the wrapper method which gives us a "good" number given current gas prices

packages/env-tests/src/tests/oracle.ts
nambrot: This could be yes

packages/env-tests/src/tests/reserve.ts
nambrot: I personally find it, what do you think?

yorhodes (approved):
lets followup on some of these CK improvements

PR #5799, opened 2020-11-10
[Feature] added prometheus self-scrape job

yourbuddyconner:
### Description

Added a job to the prometheus helm chart, allowing prometheus to scrape itself.

### Other changes

N/A 

### Tested

This is just boilerplate prometheus config. 

### Backwards compatibility

This affects no existing scrape configs.


tkporter (approved):
Curious why 20s? Feels pretty frequent but I'll defer to you on that

yourbuddyconner:
> Curious why 20s? Feels pretty frequent but I'll defer to you on that

Yeah I was just splitting the difference between 5s and 60s, I figure we can turn these off after I get the data I need.

PR #5800, opened 2020-11-11
Fix gas price cache to handle multiple currencies

nategraf:
### Description

Most significantly, this PR includes a fix to the gas price cache in the mobile app to allow for multiple currencies to be used.

Additionally, this PR contains some other useful changes implemented during our last hackathon.

### Other changes

* Use a `Lock` for getting contractkit instance to avoid reliance on polling and flag-based locking.
* Don't clear the transaction feed when the call to the blockchain API fails, such as when the device is offline.
* Allow specifying the address when getting core smart contracts to allow for external caching or testing non-canonical versions of the contract.
* Add `networkConnectedSelector` to access the network connectivity state reported by the device.
* Allow for specification of `staticGas` when calling `sendAndMonitorTransaction`.

### Tested

Existing unit tests.


nategraf:
This PR is now merged into https://github.com/celo-org/celo-monorepo/pull/6217

PR #5810, opened 2020-11-11
Get Ganache's PID from lsof

m-chrzan:
### Description

In the original version of the script, `$!` apparently wasn't pointing to the ganache process that would end up running in the background, binding port 8545. This PR grabs the process's PID with `lsof` instead.

Going further, this PR now introduces a bash function that will start a command in the background, but block the current process until that command outputs a particular string. In this case, we start Ganache and wait for it to output "Ganache STARTED", so we know that it has bound to port 8545 and is ready to receive RPCs.

### Tested

Kills ganache.


nambrot (approved):

packages/protocol/scripts/bash/release-on-devchain.sh
```diff
@@ -39,8 +39,6 @@ mv build/contracts $BUILD_DIR
 echo "- Run local network"
 # yarn devchain generate-tar devchain.tar.gz
 yarn devchain run-tar devchain.tar.gz >> $LOG_FILE &
-GANACHE_PID=$!
```
nambrot, line 42: Can we get the PID here and print it out, in case the script fails and the user can kill the process manually?

nambrot (approved):

packages/protocol/scripts/bash/release-on-devchain.sh
m-chrzan: We'd have to give ourselves enough time so we're pretty sure ganache has bound the port. I guess right now we're assuming that `yarn build` below takes enough time to not need a `sleep`, so we could put `lsof` after that, but that would leave us with the edge case of not printing it out after a failed build.

I guess that could be acceptable (this first build is of an old branch, which should have been tested and building correctly for a while now).

nambrot (approved):

PR #5824, opened 2020-11-11
Make sure account is unlocked when revoking

cmcewen:
### Description

Revoke was failing because we don't make sure the account is unlocked


annakaz (approved):
👍

cmcewen:
this was an old version problem 😞

PR #5825, opened 2020-11-11
[Wallet] Restore mnemonics in any language if default fails

### LINKED ISSUE #5204: Autodetect mnemonic language
### Expected Behavior

When importing a mnemonic, it works if it's valid regardless of the language the mnemonic and app are in.

### Current Behavior

If the app is in Spanish and we try to import an English mnemonic, we get an error.

gnardini:
### Description

Use the words in the mnemonic to tell which language the mnemonic is actually in and use that to restore the account instead of using the app's language.

### Tested

Manually and there's a unit test as well

### Related issues

- Fixes #5204

### Backwards compatibility

N/A


packages/utils/src/account.test.ts
```diff
@@ -205,5 +213,7 @@ describe('Mnemonic validation', () => {
       const password = await generateKeys(mnemonics[i], 'password')
       expect({ derivation0, derivation1, password }).toEqual(expectedPrivateKeys[i])
     }
+    // Spanish mnemonic is detected even if we pass english.
+    expect(validateMnemonic(spanishMnemonic, MnemonicLanguages.english)).toBeTruthy()
```
nategraf: As written, it would be better if this were to fail. I think we should have this be the case that if someone _does not_ specify a language, it will try any language, but if they do it should only try that one language. It's true that this may mean trying two languages more often on validation, but the computational cost is trivial.

packages/utils/src/account.ts
```diff
@@ -145,6 +157,19 @@ function getWordList(language?: MnemonicLanguages) {
...
+    MnemonicLanguages.chinese_traditional,
+    MnemonicLanguages.english,
+    MnemonicLanguages.french,
+    MnemonicLanguages.italian,
+    MnemonicLanguages.japanese,
+    MnemonicLanguages.korean,
+    MnemonicLanguages.spanish,
+  ]
+}
+
```
nategraf, line 170: Are you sure we ship with all the word lists? IIRC, we only ship with English and Spanish. If we do, this looks good, but I'd recommend double checking. One trick you could do is to check this at runtime with `Object.keys`, possible with a static list of all possible languages to filter out any object keys that are not languages.

nategraf (commented):
Really glad you put this together. 🙌  I had a couple of minor comments.

(Note: I'll be gone until next Wednesday, so feel free to merge this without a second look from me)

etuleu:
I have no idea how this works but why is the language relevant? Can we treat the string as an array of bytes?

packages/utils/src/account.ts
gnardini: Even if we don't have the word lists, having it like this will future-proof the function, if we don't have the word list it will just fail now but if we ever add it in the future we don't have to remember to add it here as well.

I had a types issue using `Object.keys` and other similar options because it detected them as strings rather than the enum value, couldn't find a way of getting all of the values of an enum by googling :(

gnardini:
@etuleu 
> I have no idea how this works but why is the language relevant? Can we treat the string as an array of bytes?

I'm not sure I follow. The account key words can be in many languages, and previously we were only comparing the words input by the user with the words in the user's app language. With this change we compare them against all languages so that they can recover their account even if their account key language doesn't match the language of their app.

jeanregisser (approved):
Awesome! This is gonna make our user's experience more predictable 💯

PR #5847, opened 2020-11-12
Extract governance:build-proposal command

### LINKED ISSUE #5341: Governance Proposal construction SBAT use scientific notation for integer values
### Expected Behavior

You can use scientific notation

### Current Behavior

You have to specify the number fully spelt out which for larger numbers is error-prone

### LINKED ISSUE #5342: Governance Proposal Construction should output a JSON file instead of proposing
### Expected Behavior

The command should output a JSON file that can be proposed directly, but also embedded in a CGP

### Current Behavior

`celocli governance:propose --interactive` allows one to interactively create a proposal, but submits it immediately.

nambrot:
### Description

Extracts the interactive building so that one can build a proposal without actually proposing it

### Related issues

- Fixes #5342 
- Fixes #5341


yorhodes (approved):
sweet

PR #5854, opened 2020-11-12
Fix check of previously verified accounts when in feeless flow

### LINKED ISSUE #5853: Feeless verification is requesting SMS's even though user is already verified
### Expected Behavior
Feeless flow should correctly detect when a user is already verified.

### Current Behavior
It is attempting to verify users that are already verified.

tarikbellamine:
### Description
The feeless verification flow was incorrectly using the `lookupIdentifiers` method on the Attestations contract when it should have been using `lookupAccountsForIdentifier`. This was causing us to check if EOAs were verified instead of MTWs, which is why it was never detecting any previously verified accounts.

### Other changes
Moving `dekAndWalletRegistration` into the `endVerification` generator to ensure that it's always called.

### Tested
Yes

### Related issues
- Fixes #5853

### Backwards compatibility
Yes


jeanregisser (approved):
Awesome thanks for fixing this! 👍 

Note for later, would probably be good that this part be covered by a unit test to minimize the chance of breaking it in the future.

Lss-Ankit:
Hi @tarikbellamine i Have verified this issue latest test flight build v1.5.5 (34) and Android play store internal build v1.5.5 (1004294323) and observe the following

- Feeless flow is not shown when a user is already verified.

Devices: iPhone 6+ (12.4.5) , Vivo v11 pro (10)

Can you please let us know if we need to test anything else.

PR #5859, opened 2020-11-13
[Attestation Bot] Fix choice of number, and US number creation

timmoreton:
### Description

Attestation Bot fixes:
* Pick only numbers that can receive incoming SMS (have a dummy URL specified for Twilio to POST to)
* Create numbers in the US correctly  
* Add explicit Twilio dependency for celotool

Baklava deployment fixes:
* deploy latest to baklava
* no longer require Twilio messaging service ID for Attestation Bot (API calls that used it had no effect)
* update helm charts

### Tested

Locally and by deploying to baklava

### Backwards compatibility

Yes


packages/celotool/src/lib/attestation.ts
```diff
@@ -255,9 +274,8 @@ export async function createPhoneNumber(
 
     await twilioClient.incomingPhoneNumbers.create({
       phoneNumber: usableNumber,
-      addressSid,
```
nambrot, line 258: Remind me again why wouldn't we try to get phone numbers in other jurisdictions that require the addressSid?

packages/celotool/src/lib/attestation.ts
```diff
@@ -187,7 +194,13 @@ export async function chooseFromAvailablePhoneNumbers(
   maximumNumberOfAttestations: number,
   salt: string,
 ) {
-  const availableNumbers = await twilioClient.incomingPhoneNumbers.list()
+  const availableNumbers = (await twilioClient.incomingPhoneNumbers.list()).filter(
+    (number) => number.smsUrl === DUMMY_SMS_URL
+  )
+  if (!availableNumbers?.length) {
+    return undefined
+  }
```
aslawson, line 202: Wouldn't findSuitableNumber handle the case where the availableNumbers is an empty list due to filtering?

packages/celotool/src/lib/attestation.ts
aslawson: i'm unclear what this was originally used for?

aslawson (approved):
LGTM

packages/celotool/src/lib/attestation.ts
timmoreton: From trying the API, this parameter has no effect here

packages/celotool/src/lib/attestation.ts
timmoreton: that was what I originally figured, but it didn't seem to 🤷

packages/celotool/src/lib/attestation.ts
timmoreton: There seems to be no way AFAICT of creating a number inside a messaging service.  Passing the address in here via `messagingServiceSid` here results in `address not found` for me.

timmoreton: So I resolved we would not use a messaging service for the bot, as previously, and just make sure we only select numbers to get attested that definitely can accept incoming messages

PR #5872, opened 2020-11-13
Reduce the number of validators in the validator order e2e test.

oneeman:
This is an attempt to reduce its failures on CI, which don't also happen locally. The hypothesis is that 10 validators is too much for the CI machines and is leading to frequent round changes in consensus.

### Description

The e2e are noticeably flakier on CircleCI than when run locally on a dev machine.  This test in particular was failing persistently for me in CI but not failing at all locally.  I hypothesized that it's because 10 validators is too much for the CI executors.  This PR changes it to use 5 validators.  This has no impact on the test's logic, because 5 validators is enough for the validator ordering in each epoch to be meaningful.  There are no other hard-coded constants that have to be changed because of it.

### Tested

The test consistently passes locally.  I created a branch on celo-blockchain to use this PR's branch, and the test (which was consistently failing on CI before) passed the first time: https://app.circleci.com/pipelines/github/celo-org/celo-blockchain?branch=oneeman%2Fe2e-ci-test

### Backwards compatibility

No impact on production, and the new test setting works regardless of the celo-blockchain version being tested.


trianglesphere (approved):
lgtm

PR #5875, opened 2020-11-13
Support absolute paths for proposal

nambrot:
### Description

Proposals were only specifiable by relative path


asaj (approved):

yorhodes:
`realpath` is not available on osx by default cc @nambrot

PR #5877, opened 2020-11-13
[Wallet] Change the method used to get account addresses from phone numbers

### LINKED ISSUE #5876: [Wallet] Change the method used to get account addresses from phone numbers
### Expected Behavior

We get the account address when querying for it.

### Current Behavior

The method responsible for getting the account address is actually fetching the wallet address.

More info: https://celo-org.slack.com/archives/CL7BVQPHB/p1605282150297100

gnardini:
### Description

The method responsible for getting the account address is actually fetching the wallet address.

More info: https://celo-org.slack.com/archives/CL7BVQPHB/p1605282150297100

### Tested

Manually by verifying and sending stuff with comments and there are also tests.

### Related issues

- Fixes #5876

### Backwards compatibility

Should be

### DIFF
--- a/packages/mobile/src/identity/commentEncryption.ts
+++ b/packages/mobile/src/identity/commentEncryption.ts
@@ -240,12 +240,12 @@ function* verifyIdentityMetadata(data: IdentityMetadataInTx[]) {
     const accountAddresses: Address[] = yield call(lookupAccountAddressesForIdentifier, phoneHash)
 
     // Check that there are verified addresses.
-    const onChainAddresses: Address[] = yield call(
+    const verifiedAccountAddresses: Address[] = yield call(
       filterNonVerifiedAddresses,
       accountAddresses,
       phoneHash
     )
-    if (!onChainAddresses || !onChainAddresses.length) {
+    if (verifiedAccountAddresses.length === 0) {
       Logger.warn(
         TAG + 'verifyIdentityMetadata',
         `Phone number and/or salt claimed by address ${metadata.address} is not verified. Values are incorrect or sender is impersonating another number`
@@ -253,11 +253,11 @@ function* verifyIdentityMetadata(data: IdentityMetadataInTx[]) {
       continue
     }
     const walletAddresses: Address[] = yield all(
-      onChainAddresses.map((accountAddress) =>
+      verifiedAccountAddresses.map((accountAddress) =>
         call(accountsWrapper.getWalletAddress, accountAddress)
       )
     )
-    if (!walletAddresses.find((a) => eqAddress(a, metadata.address))) {
+    if (!walletAddresses.some((walletAddress) => eqAddress(walletAddress, metadata.address))) {
       Logger.warn(
         TAG + 'verifyIdentityMetadata',
         `Phone number and/or salt claimed by address ${metadata.address} does not match any on-chain addresses. Values are incorrect or sender is impersonating another number`
--- a/packages/mobile/src/identity/commentEncryption.ts
+++ b/packages/mobile/src/identity/commentEncryption.ts
@@ -3,7 +3,6 @@
 // because these manage comment metadata
 
 import { PhoneNumberHashDetails } from '@celo/contractkit/lib/identity/odis/phone-number-identifier'
-import { IdentifierLookupResult } from '@celo/contractkit/lib/wrappers/Attestations'
 import { eqAddress, hexToBuffer } from '@celo/utils/src/address'
 import {
   decryptComment as decryptCommentRaw,
@@ -225,38 +224,35 @@ function* verifyIdentityMetadata(data: IdentityMetadataInTx[]) {
     return []
   }
 
-  const phoneHashes = new Set<string>()
-  data.map((d) => {
-    const phoneHash = getPhoneHash(d.e164Number, d.salt)
-    phoneHashes.add(phoneHash)
-    d.phoneHash = phoneHash
-  })
+  const verifiedTx = []
 
-  const lookupResult: IdentifierLookupResult = yield call(
-    lookupAttestationIdentifiers,
-    Array.from(phoneHashes)
-  )
+  for (const metadata of data) {
+    const phoneHash = getPhoneHash(metadata.e164Number, metadata.salt)
+    metadata.phoneHash = phoneHash
+    const lookupResult: string[] = yield call(lookupAttestationIdentifiers, phoneHash)
 
-  return data.filter((d) => {
-    const onChainAddresses = getAddressesFromLookupResult(lookupResult, d.phoneHash!)
+    // Check that there are verified addresses.
+    const onChainAddresses = yield call(getAddressesFromLookupResult, lookupResult, phoneHash)
     if (!onChainAddresses || !onChainAddresses.length) {
       Logger.warn(
         TAG + 'verifyIdentityMetadata',
         `Phone number and/or salt claimed by address ${d.address} is not verified. Values are incorrect or sender is impersonating another number`
       )
-      return false
+      continue
     }
 
-    if (!onChainAddresses.find((a) => eqAddress(a, d.address))) {
+    if (!onChainAddresses.find((a) => eqAddress(a, metadata.address))) {
       Logger.warn(
         TAG + 'verifyIdentityMetadata',
-        `Phone number and/or salt claimed by address ${d.address} does not match any on-chain addresses. Values are incorrect or sender is impersonating another number`
+        `Phone number and/or salt claimed by address ${metadata.address} does not match any on-chain addresses. Values are incorrect or sender is impersonating another number`
       )
-      return false
+      continue
     }
 
-    return true
-  })
+    verifiedTx.push(metadata)
+  }
+
+  return verifiedTx
 }
 
 // Dispatch updates to store with the new information
--- a/packages/mobile/src/identity/commentEncryption.ts
+++ b/packages/mobile/src/identity/commentEncryption.ts
@@ -2,7 +2,9 @@
 // Use these instead of the functions in @celo/utils/src/commentEncryption
 // because these manage comment metadata
 
+import { Address } from '@celo/base'
 import { PhoneNumberHashDetails } from '@celo/contractkit/lib/identity/odis/phone-number-identifier'
+import { AccountsWrapper } from '@celo/contractkit/lib/wrappers/Accounts'
 import { eqAddress, hexToBuffer } from '@celo/utils/src/address'
 import {
   decryptComment as decryptCommentRaw,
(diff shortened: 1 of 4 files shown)


packages/mobile/src/identity/contactMapping.ts
```diff
@@ -261,36 +258,43 @@ function* fetchWalletAddresses(e164Number: string) {
...
+  ])
+
   const verifiedAddresses: string[] = []
-  for (const address of Object.keys(addressToStats)) {
+  for (const address of lookupResult) {
     if (!isValidNon0Address(address)) {
       continue
     }
+    // Get stats for the address
+    const stats: AttestationStat = yield call(
```
gnardini, line 292: Doing a call for each account received. This could be an issue when called from `verifyIdentityMetadata` on `packages/mobile/src/identity/commentEncryption.ts`, but my understanding is that it's not run often, so I'm unsure of what the performance hit will actually be. Any thoughts on this? Should I call another batch method to get stats to better cover that case?

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -225,38 +224,39 @@ function* verifyIdentityMetadata(data: IdentityMetadataInTx[]) {
...
         TAG + 'verifyIdentityMetadata',
-        `Phone number and/or salt claimed by address ${d.address} is not verified. Values are incorrect or sender is impersonating another number`
+        `Phone number and/or salt claimed by address ${metadata.address} is not verified. Values are incorrect or sender is impersonating another number`
       )
-      return false
+      continue
     }
 
-    if (!onChainAddresses.find((a) => eqAddress(a, d.address))) {
+    if (!onChainAddresses.find((a) => eqAddress(a, metadata.address))) {
```
asaj: metadata.address is the wallet or account address? IIUC it should be the account address. Should we then be checking that the wallet address for that account address was the sender of the tx?

I guess I don't see how this verifies the transaction, probably I'm missing something somewhere. What's stopping me from sending a tx that I claim is from someone else?

packages/mobile/src/identity/contactMapping.ts
tarikbellamine: If a batched request can be made to the contract, I think it's worth it. Otherwise, you can break this loop up to make it more performant by:
1. filtering invalid addresses
2. doing a `yield all` with a `call` for the attestation state of every valid address
3. filtering for verified addresses

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -261,36 +258,43 @@ function* fetchWalletAddresses(e164Number: string) {
...
 
 // Deconstruct the lookup result and return
 // any addresess that are considered verified
-export function getAddressesFromLookupResult(
-  lookupResult: IdentifierLookupResult,
-  phoneHash: string
-) {
-  if (!lookupResult || !lookupResult[phoneHash]) {
-    return null
+export function* getAddressesFromLookupResult(lookupResult: string[], phoneHash: string) {
```
tarikbellamine: Can we use a more specific type than `string[]`? Is this still of type `IdentifierLookupResult[]`?

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -301,7 +305,7 @@ export function getAddressesFromLookupResult(
     verifiedAddresses.push(address.toLowerCase())
   }
 
-  return verifiedAddresses.length ? verifiedAddresses : null
+  return verifiedAddresses
```
tarikbellamine: Can we rename this to `verifiedAccountAddresses`? I think it would be helpful to be as clear as possible about the addresses we are getting given the confusion over when we are dealing `accountAddress` vs `walletAddress`

packages/mobile/src/identity/commentEncryption.ts
gnardini: Yeah, there's something I'm missing here. The comments on payments are shown fine and I see the phone number as the sender but this equality fails and I see the message below in the logs. This method is used to update phone numbers to salt and phone number to address mappings. I'm not sure how those are used, I guess it makes some of our caches useless? I feel like even if we fix this here it will have repercussions in other places that I don't fully understand.

In any case we'll need to get the wallet address here, I guess that requires a call to the accounts contract? I'll check if we're requesting it somewhere else.

tarikbellamine: It's the wallet address. We basically want to do what you suggested but in reverse.
1. Use `phoneNumber` and `salt` from the metatx to compute `phoneHash`
2. Use `phoneHash` to lookup the `identifiers`
3. Use `identifiers` to get the `verifiedAccounts`
4. Use `verifiedAccounts` to get the `walletAddresses` registered
5. Confirm that at least one of the `walletAddresses` is equal to the address the transaction was sent from

I think we are missing step 4 right now.

packages/mobile/src/identity/contactMapping.ts
```diff
@@ -224,8 +221,8 @@ export function* fetchAddressesAndValidateSaga({
 function* getAccountAddresses(e164Number: string) {
   const phoneHashDetails: PhoneNumberHashDetails = yield call(fetchPhoneHashPrivate, e164Number)
   const phoneHash = phoneHashDetails.phoneHash
-  const lookupResult: IdentifierLookupResult = yield call(lookupAttestationIdentifiers, [phoneHash])
-  return getAddressesFromLookupResult(lookupResult, phoneHash) || []
+  const lookupResult: string[] = yield call(lookupAttestationIdentifiers, phoneHash)
+  return yield call(getAddressesFromLookupResult, lookupResult, phoneHash) || []
```
tarikbellamine: Correct me if I am wrong but I don't think you need that `|| []` anymore because you've changed the output of `getAddressesFromLookupResult` to be an empty array instead of `null` when it's empty

packages/mobile/src/identity/commentEncryption.ts
tarikbellamine: Alternatively, we could also use the `lookupIdentifiers` as that returns the `walletAddresses` straight away, though there is a bit of trade-off with the logic not being explicit as we've seen

packages/mobile/src/identity/contactMapping.ts
gnardini: It's not, `IdentifierLookupResult` has stats information, this is just a list of addresses. Do we have a type for addresses?

packages/mobile/src/identity/commentEncryption.ts
asaj: @tarikbellamine +1 to your first approach. Will all of this be done via view calls or is there cached information we will be able to rely on for this?

packages/mobile/src/identity/contactMapping.ts
asaj: We do! In @celo/base. I don't think Valora uses it much yet, but we should move in that direction

packages/mobile/src/identity/commentEncryption.ts
tarikbellamine: I think view calls for now but there should be no reason we need to look up this info more than once for the vast majority of cases. I'm hesitant to throw more caches in the mix because, as you've noted, our caching situation right now is kinda messy and should be better organized.

If you feel strongly about the performance hit, it's probably something we could tack on. Otherwise, it may be better to wait until we take a more comprehensive look at how and when we are caching across Valora. Will leave the call up to you and @gnardini

gnardini: This particular function is not called very often, so I'm nor sure it would add much value. I do think that having a unified cache like @tarikbellamine suggested would work best, but we have to rethink our caches app-wide which we could hopefully tackle soon.

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -221,42 +223,52 @@ function* verifyIdentityMetadata(data: IdentityMetadataInTx[]) {
...
-  return data.filter((d) => {
-    const onChainAddresses = getAddressesFromLookupResult(lookupResult, d.phoneHash!)
+  const verifiedTx = []
+  for (const metadata of data) {
+    const phoneHash = getPhoneHash(metadata.e164Number, metadata.salt)
+    metadata.phoneHash = phoneHash
+    const accountAddresses: Address[] = yield call(lookupAccountAddressesForIdentifier, phoneHash)
+
+    // Check that there are verified addresses.
+    const onChainAddresses: Address[] = yield call(
```
tarikbellamine: Nit: Can we rename this to `verifiedAccountAddresses`? I don't think `onChainAddresses` means anything in specific

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -221,42 +223,52 @@ function* verifyIdentityMetadata(data: IdentityMetadataInTx[]) {
...
+      filterNonVerifiedAddresses,
+      accountAddresses,
+      phoneHash
+    )
     if (!onChainAddresses || !onChainAddresses.length) {
```
tarikbellamine: I think this will always be an array so the first part of the check is not necessary

packages/mobile/src/identity/commentEncryption.ts
```diff
@@ -221,42 +223,52 @@ function* verifyIdentityMetadata(data: IdentityMetadataInTx[]) {
...
+      continue
     }
-
-    if (!onChainAddresses.find((a) => eqAddress(a, d.address))) {
+    const walletAddresses: Address[] = yield all(
+      onChainAddresses.map((accountAddress) =>
+        call(accountsWrapper.getWalletAddress, accountAddress)
+      )
+    )
+    if (!walletAddresses.find((a) => eqAddress(a, metadata.address))) {
```
tarikbellamine: Nit: I think using the `some` method is a little more appropriate

tarikbellamine (approved):
Left a couple of nits but looks good to me. Nice job!

packages/mobile/src/identity/commentEncryption.ts
asaj: Definitely do not feel strongly about the performance hit and agree with both of you that this is the sort of thing better left to a refactor, just wanted to flag

Lss-Ankit:
@gnardini  can you please provide more details to verify this task.

gnardini:
@Lss-Ankit I think this is tested already by other things in the test plan, like sending to a phone number with a comment and seeing that the funds are received correctly and the comment is visible

Lss-Ankit:
Hi @gnardini I have re-verified this issue using latest Android internal build v1.5.591004294323) & Test flight build v1.5.5(34)  & observed following.
- User able to see all fund details, account address, comment in both Sent & Received activity. 

**Devices:** Samsung Galaxy A5(7.0) ,iPhone SE(13.5.1)

PR #5880, opened 2020-11-13
[Attestation Service] Config to randomize provider selection, tune health check

timmoreton:
### Description

* `SMS_PROVIDERS_RANDOMIZED` option, off by default, that randomizes provider order on a per-attestation basis when no per-country override list is specified. Once set, the provider order is written into the DB and subsequent retries do not re-shuffle.
*  `MAX_AGE_LATEST_BLOCK_SECS` option allows the maximum age of the latest block at which the health check fails to be configurable. Previously, we saw issues in baklava where slow blocks were causing spewing of Attestation Service alerts.

### Other changes

* Update Twilio dependency
* Docs updates

### Tested

Locally

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes


aslawson (approved):
looks good :)

PR #5884, opened 2020-11-16
Remove unnecesary postgres in komenci deployment

eruizgar91:
### Description

In the deployment of komenci there is not need it the postgres deployment, because we are using one that is deployed already for all the environments.

### Other changes

### Tested

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


bowd (approved):

.env.alfajores
```diff
@@ -140,7 +140,7 @@ CONTEXTS=azure-eastus,azure-westeu
 # --- Komenci ---
 
 KOMENCI_DOCKER_IMAGE_REPOSITORY="celotestnet.azurecr.io/komenci/komenci"
-KOMENCI_DOCKER_IMAGE_TAG="e855372cc885b9f483c38de5d2a4ef0b2fc1618f"
+KOMENCI_DOCKER_IMAGE_TAG="e8bfdebcd91061a0190cf1f40005bf0d6c567dee"
```
codyborn, line 143: Is this the latest image?

codyborn (approved):
LGTM

PR #5889, opened 2020-11-16
Use Set instead of Array for Secure Send + don't allow re-verification if an EOA is verified

### LINKED ISSUE #5887: Users with already verified EOA are asked to verify again using MTW after reinstalling and restoring account
### Expected Behavior

Users already verified with their EOA shouldn't need to verify again when restoring their account with Valora v1.5 and feeless attestation.

### Current Behavior

Steps to reproduce:
- have an account already verified with Valora < 1.5
- upgrade to v1.5.4
- delete the app, reinstall app
- restore account
- app asks to verify again and deploy a MTW for the user
- observe that secure send is now required when sending to that user's phone number

See thread in https://celo-org.slack.com/archives/CL7BVQPHB/p1605534658343100?thread_ts=1605527173.336900&cid=CL7BVQPHB

tarikbellamine:
### Description
There is an edge case in which a user may have both an EOA and MTW verified, with the EOA assigned as the `walletAddress` for both. In this event, Secure Send was triggering because we were using an array to store the `walletAddresses` instead of a set. Now instead of Secure Send triggering when there is more than one wallet address, it will only trigger if there is more than one unique wallet address. 

### Other changes
No

### Tested
Yes

### Related issues

- Closes #5887 

### Backwards compatibility
Yes


packages/mobile/src/identity/contactMapping.ts
```diff
@@ -248,15 +248,15 @@ function* fetchWalletAddresses(e164Number: string) {
...
     } else {
       // NOTE: Only need this else block if we are not confident all wallets are registered
       walletToAccountAddress[accountAddress] = accountAddress
-      possibleUserAddresses.push(accountAddress)
+      possibleUserAddresses.add(accountAddress)
     }
   }
   yield put(updateWalletToAccountAddress(walletToAccountAddress))
-  return possibleUserAddresses
+  return [...possibleUserAddresses]
```
gnardini: I would just add a small comment here explaining that this is to convert the set into an array. It looks a bit weird otherwise and someone might accidentally remove it.

gnardini (approved):
LGTM 🎉

packages/mobile/src/identity/contactMapping.ts
gnardini: Another option I think is to do `Array.from(possibleUserAddresses)` which wouldn't need the comment imo.

tarikbellamine: Good point, will follow your suggestion so it's more clear why I am doing it

Lss-Ankit:
Hi @tarikbellamine i Have verified this issue latest test flight build v1.5.5 (34) and Android play store internal build v1.5.5 (1004294323) and observe the followings

- Send secure flow is not trigger when sending to that user's phone number

Devices: iPhone 6+ (12.4.5) , Vivo v11 pro (10)

Can you please let us know if we need to test anything else

PR #5903, opened 2020-11-17
Cusd reward events parsing

ewilz:
I’ve tried to make the tooling as generic as possible. What we're doing is fetching events, and parsing them to ultimately get rewards amounts. For convenience, I added an `events` command to `celocli` (though I understand this may not be the right home for the code) that has `events:fetch` for fetching events, and `events:parse` for parsing the events data into the information we need. The commands below show how to use the generic tooling to reach a time-weighted average for attested accounts (though I understand we may not take this approach to rewards). Each command below will write to a json file that will be useful in a future command. There is a `batchSize` parameter because if we are fetching too many events, the forno server terminates our connection.

The following commands get time-weighted average balances from (arbitrary) block 3000001 - 3500000 for addresses with 3 attestations
```bash
celocli events:fetch:attestationcompleted --fromBlock 0 --toBlock 3000000 --batchSize 10000
celocli events:parse:attestedaccounts --minimum 3 --eventsJson attestation-completed-events.json  --outputJson

celocli events:fetch:transfercusd --fromBlock 0 --toBlock 3000000 --batchSize 10000
celocli events:parse:balances --eventsJson transferevents.json --filter attested-accounts.json

celocli events:fetch:transfercusd --fromBlock 3000001 --toBlock 3500000  --batchSize 10000
celocli events:parse:balancestwa --eventsJson events.json --fromBlock 3000001 --toBlock 3500000  --initialBalances balances.json 
```

to use the arbitrary block numbers above....you may need to run this first 😅:
```
export NODE_OPTIONS="--max-old-space-size=8192"
```


ewilz:
outsourcing this to other repo

PR #5905, opened 2020-11-17
Do not update Valora balance if balance query was unsuccessful

### LINKED ISSUE #5285: Valora intermittently displays extremely large and wrong user balances
### Expected Behavior
Valora fetches the correct user balance

### Current Behavior
Sometimes a very large number is incorrectly shown as the user's balance. This is messing up analytics in addition to confusing the user. For whatever reason, the incorrect balance appears to consistently be 3.96xE57 or xE75

Added context and potential solutions:
- https://celo-org.slack.com/archives/CABQPRW0Z/p1602641615093500

Examples:
- https://groups.google.com/a/clabs.co/g/support/c/iAM-J6AL4TY/m/aFI_nWJPBAAJ
- https://celo-org.slack.com/archives/CNCUN1R0D/p1598427957038500

tarikbellamine:
### Description
When view calls to the token balance contracts revert or fail, they often return a very large number instead of an error message. This has been causing us to intermittently show some users balances of over 10M and logging it in the analytics, which corrupts our data and distorts our graphs.

This PR implements a reasonable upper bound of 10M. Any query that returns a balance over 10M is now assumed to be a failed query.

### Other changes
No

### Tested
Yes

### Related issues

- Fixes #5285 
### Backwards compatibility
Yes


packages/mobile/src/tokens/saga.ts
```diff
@@ -73,15 +77,19 @@ export function tokenFetchFactory({ actionName, token, actionCreator, tag }: Tok
...
-      ValoraAnalytics.track(
-        AppEvents.fetch_balance,
-        token === CURRENCY_ENUM.DOLLAR
-          ? {
-              dollarBalance: balance.toString(),
-            }
-          : { goldBalance: balance.toString() }
-      )
+
       yield put(actionCreator(balance.toString()))
```
annakaz: Should setting the balance also be skipped when it is too large, instead of hiding it?

The concern with hiding it with the above check with formattedValue is that there are other checks that may incorrectly succeed (whether someone has sufficient balance in the Send flow, for example)

tarikbellamine: Good point. I guess that implies that we would continue to show the last fetched balance instead of the erroneous one, which is a more elegant solution.

I'll make this change, thanks!

packages/mobile/src/tokens/saga.ts
```diff
@@ -18,6 +18,10 @@ import * as utf8 from 'utf8'
 
 const TAG = 'tokens/saga'
 
+// We can safely assume that any balance query returning a number
+// higher than this is incorrect (currently set to 10M)
+export const VALORA_BALANCE_UPPER_BOUND = new BigNumber('1e10')
```
jeanregisser: nit: we shouldn't use the Valora naming  in the open source code.

```suggestion
export const BALANCE_UPPER_BOUND = new BigNumber('1e10')
```

packages/mobile/src/exchange/ExchangeConfirmationCard.test.tsx
```diff
@@ -24,7 +24,19 @@ it('renders correctly with no exchange rate', () => {
...
+it('renders correctly with larger numbers', () => {
+  const tree = renderer.create(
+    <Provider store={store}>
+      <ExchangeConfirmationCard
+        makerAmount={{ value: '240000.00', currencyCode: 'cUSD', localAmount }}
+        takerAmount={{ value: '180000000', currencyCode: 'cGLD', localAmount }}
+      />
+    </Provider>
+  )
+  expect(tree).toMatchSnapshot()
```
jeanregisser: Only checking with a snapshot doesn't tell much as we usually don't pay too much attention to them.
Would be better to also check the amount we expect is displayed.

Something like this:
```tsx
expect(getElementText(tree.getByTestId('SomeTestId'))).toEqual('$240,000')
```

packages/mobile/src/tokens/saga.ts
jeanregisser: Should we move this to `src/config.ts`?

jeanregisser (commented):
Great! 👍 
Left some small comments.

packages/mobile/src/components/CurrencyDisplay.tsx
```diff
@@ -180,7 +181,9 @@ export default function CurrencyDisplay({
   const sign = value?.isNegative() ? '-' : showExplicitPositiveSign ? '+' : ''
   const formatAmount = getFormatFunction(formatType)
   const formattedValue =
-    value && displayCurrency ? formatAmount(value.absoluteValue(), displayCurrency) : '-'
+    value && displayCurrency && value.lt(WALLET_BALANCE_UPPER_BOUND)
+      ? formatAmount(value.absoluteValue(), displayCurrency)
+      : '-'
```
jeanregisser: I would rather not have this check here.
I think this is too defensive and could be surprising in the future, given that incorrect high values are handled in the tokens saga (where they can happen).

packages/mobile/src/tokens/saga.ts
```diff
@@ -73,15 +74,20 @@ export function tokenFetchFactory({ actionName, token, actionCreator, tag }: Tok
...
           : { goldBalance: balance.toString() }
-      )
-      yield put(actionCreator(balance.toString()))
+
+      // Only update balances when it's less than the upper bound
+      if (balance.lt(WALLET_BALANCE_UPPER_BOUND)) {
+        yield put(actionCreator(balance.toString()))
+        ValoraAnalytics.track(AppEvents.fetch_balance, balanceLogObject)
+      } else {
+        ValoraAnalytics.track(AppEvents.fetch_balance_error, balanceLogObject)
```
jeanregisser, line 89: I think the `fetch_balance_error` event name is too generic and should be specific about this incorrect high value error.
If someone looks at this from the analytics dashboard it probably won't be clear that it's about this specific case.

jeanregisser (commented):
Nice! 
Could you also add a test covering this?

packages/mobile/src/tokens/saga.ts
tarikbellamine: Do you have something else in mind?

We generally do `event_name` and `event_name_error` to capture errors related to specific events, then we use the event params to differentiate between the various causes of this error. In this case, we are passing very large numbers as params, which should be a clear indicator for what the error was. If you think it would be helpful, I could add another param (e.g., `largeBalance = true`) to make this more explicit?

gnardini: Assuming this is the only source of errors for fetching a balance it looks fine to me. If we had other sources of errors for fetching balance I'd vote for a field explaining what type of error like `errorType` or something. Not sure if that's necessary here.

gnardini (approved):
LGTM

packages/mobile/src/config.ts
```diff
@@ -64,6 +64,10 @@ export const VERIFICATION_STATE_EXPIRY_SECONDS = 30
 
 export const ATTESTATION_REVEAL_TIMEOUT_SECONDS = 60 // 1 minute
 
+// We can safely assume that any balance query returning a number
+// higher than this is incorrect (currently set to 10M)
+export const WALLET_BALANCE_UPPER_BOUND = new BigNumber('1e10')
```
nategraf, line 69: If at some point a user files a bug report that "Balance is not displayed correctly on my account with greater than $10M" I indeed agree that will be an interesting day. 😆

PR #5906, opened 2020-11-18
Fixed wrong syntax for docker ports on the docs

martinvol:
Also made it consistent with the rest of the docs, changing `localhost` for `127.0.0.1`


tkporter (approved):
nice catch!

PR #5913, opened 2020-11-18
CIP10 contracts changes

AlexBHarley:
### Description

For more context take a look at [CIP10](https://github.com/celo-org/celo-proposals/blob/master/CIPs/0010.md). There's so many commits as there was a lot of back and forth on implementation details.

### Other changes

N/A.

### Tested

Added an extensive matrix of tests to test this for backwards compatibility.

### Related issues

- Fixes [Finalise CIP10 implementation](https://app.zenhub.com/workspaces/cap-sprint-board-60059934ed9c6f000e5594f6/issues/celo-org/celo-monorepo/5486)

### Other changes

`Accounts.sol` is now `InitializableV2`.

### Backwards compatibility

These changes are backwards compatible.


packages/protocol/test/common/accounts.ts
```diff
@@ -436,148 +411,283 @@ contract('Accounts', (accounts: string[]) => {
...
 
-  Object.keys(authorizationTestDescriptions).forEach((key) => {
-    describe('authorization tests:', () => {
-      let authorizationTest: any
-      beforeEach(async () => {
-        authorizationTest = authorizationTests[key]
-        await accountsInstance.createAccount()
-      })
+  describe('general authorization', () => {
+    const authorized = accounts[1]
```
nambrot: Wdyt about using `signer` instead of `authorized` which sounds to me more a state than an address

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -35,12 +35,18 @@ contract Accounts is
...
+    bool deleted;
+  }
+
   struct Account {
     bool exists;
     // Each account may authorize signing keys to use for voting, valdiating or attestation.
     // These keys may not be keys of other accounts, and may not be authorized by any other
     // account for any purpose.
     Signers signers;
+    mapping(string => mapping(address => SignerAuthorization)) signerAuthorizations;
```
nambrot: Use string instead of hash

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -59,15 +66,23 @@ contract Accounts is
   event AttestationSignerAuthorized(address indexed account, address signer);
   event VoteSignerAuthorized(address indexed account, address signer);
   event ValidatorSignerAuthorized(address indexed account, address signer);
+  event SignerAuthorized(address indexed account, address signer, string role);
```
yorhodes: I think role should be `indexed`

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -59,15 +66,23 @@ contract Accounts is
...
   event AttestationSignerRemoved(address indexed account, address oldSigner);
   event VoteSignerRemoved(address indexed account, address oldSigner);
   event ValidatorSignerRemoved(address indexed account, address oldSigner);
+  event DefaultSignerSet(address indexed account, address signer, string role);
+  event DefaultSignerRemoved(address indexed account, address oldSigner, string role);
+  event SignerRemoved(address indexed account, address oldSigner, string role);
```
yorhodes: ditto

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -180,6 +195,59 @@ contract Accounts is
...
+   * @param role the role to register a default signer for
+   */
+  function setDefaultSigner(address signer, string memory role) public {
+    require(
+      isGenericSigner(msg.sender, signer, role),
+      "Must finish authorization before setting default signer"
+    );
+
+    Account storage account = accounts[msg.sender];
+    if (keccak256(abi.encodePacked(role)) == keccak256(abi.encodePacked(VoteSigner))) {
```
yorhodes: if we are going to be computing hashes of `role`, shouldn't we store them as `bytes32`?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -59,15 +66,23 @@ contract Accounts is
...
+  event SignerRemoved(address indexed account, address oldSigner, string role);
   event AccountDataEncryptionKeySet(address indexed account, bytes dataEncryptionKey);
   event AccountNameSet(address indexed account, string name);
   event AccountMetadataURLSet(address indexed account, string metadataURL);
   event AccountWalletAddressSet(address indexed account, address walletAddress);
   event AccountCreated(address indexed account);
 
+  string constant ValidatorSigner = "celo.org/core/validator";
+  string constant AttestationSigner = "celo.org/core/attestation";
+  string constant VoteSigner = "celo.org/core/vote";
```
yorhodes: do we really want these to be strings?

packages/protocol/contracts/common/Accounts.sol
yorhodes: the rightside of these `==` comparisons  can be a `constant` either way

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -180,6 +195,59 @@ contract Accounts is
...
+   */
+  function setDefaultSigner(address signer, string memory role) public {
+    require(
+      isGenericSigner(msg.sender, signer, role),
+      "Must finish authorization before setting default signer"
+    );
+
+    Account storage account = accounts[msg.sender];
+    if (keccak256(abi.encodePacked(role)) == keccak256(abi.encodePacked(VoteSigner))) {
+      account.signers.vote = signer;
```
yorhodes, line 252: if we are deprecating, why still assign to `signers`?

yorhodes: backwards compatibility?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -180,6 +195,59 @@ contract Accounts is
...
+    bytes32 r,
+    bytes32 s
+  ) public {
+    authorize(signer, v, r, s);
+
+    Account storage account = accounts[msg.sender];
+    account.signerAuthorizations[role][signer] = SignerAuthorization({
+      started: true,
+      completed: true
+    });
```
yorhodes: I think these 2 booleans are packed into 1 storage slot so it only counts as 1 `SSTORE`

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+   * @param account The address of account that authorized signing.
+   * @param signer The address of the signer.
+   * @param role The role that has been authorized.
+   */
+  function isGenericSigner(address account, address signer, string memory role)
+    internal
+    view
+    returns (bool)
+  {
+    require(isAccount(account), "Unknown account");
```
yorhodes: I'm not sure we should have a `require` on a view 
This is also redundant in `isDefaultSigner`

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+  }
+
+  /**
+   * @notice Whether or not the signer has been registered as a signer for role
+   * @param account The address of account that authorized signing.
+   * @param signer The address of the signer.
+   * @param role The role that has been authorized.
+   */
+  function isGenericSigner(address account, address signer, string memory role)
+    internal
```
yorhodes: views should usually be `public`

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+   * @param _account The address of account that authorized signing.
+   * @param signer The address of the signer.
+   * @param role The role that has been authorized.
+   */
+  function isDefaultSigner(address _account, address signer, string memory role)
+    internal
+    view
+    returns (bool)
+  {
+    require(isAccount(_account), "Unknown account");
```
yorhodes: ditto

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+    Account storage account = accounts[_account];
+
+    if (
+      keccak256(abi.encodePacked(role)) == keccak256(abi.encodePacked(ValidatorSigner)) &&
```
yorhodes: we can avoid gas here and elsewhere by storing the `bytes32` values

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+  }
+
+  /**
+   * @notice Removes the currently authorized signer for a specific role.
+   * @param signer The address of the signer.
+   * @param role The role that has been authorized.
+   */
+  function removeSigner(address signer, string memory role) public {
+    address defaultSigner = getDefaultSigner(msg.sender, role);
+    if (defaultSigner == signer) {
```
yorhodes: `defaultSigner` is an "unnecessary variable"

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+
+    if (keccak256(abi.encodePacked(role)) == keccak256(abi.encodePacked(ValidatorSigner))) {
+      account.signers.validator = address(0);
+    } else if (
+      keccak256(abi.encodePacked(role)) == keccak256(abi.encodePacked(AttestationSigner))
+    ) {
+      account.signers.attestation = address(0);
+    } else if (keccak256(abi.encodePacked(role)) == keccak256(abi.encodePacked(VoteSigner))) {
+      account.signers.vote = address(0);
+    }
```
yorhodes: ditto

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+
+    emit DefaultSignerRemoved(msg.sender, account.defaultSigners[role], role);
+    account.defaultSigners[role] = address(0);
```
yorhodes: I think we usually emit events after storage writes

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+   */
+  function authorizeSigner(address signer, string memory role) public {
+    require(isAccount(msg.sender), "Unknown account");
+    require(
+      isNotAccount(signer) && isNotAuthorizedSignerForAnotherAccount(signer),
+      "Cannot re-authorize address signer"
+    );
+
+    accounts[msg.sender].signerAuthorizations[role][signer] = SignerAuthorization({
+      started: true,
```
yorhodes, line 424: started is never assigned `false` other than empty storage
do we need it?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+
+    emit SignerAuthorized(msg.sender, signer, role);
+  }
+
+  /**
+   * @notice Finish the process of authorizing an address to sign on behalf of the account. 
+   * @param _account The address of account that authorized signing.
+   * @param role The role to finish authorizing for.
+   */
+  function completeSignerAuthorization(address _account, string memory role) public {
```
yorhodes: maybe I'm not familiar enough with CIP10 but why separate `authorize` and `complete`, and why limit this to only `signer == msg.sender`?
what is the downside to someone authorizing and completing you as a signer on your behalf?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -396,15 +590,42 @@ contract Accounts is
...
+   * @param _account The address of the account.
+   * @param role The role of the signer.
+   */
+  function getDefaultSigner(address _account, string memory role) public view returns (address) {
+    require(isAccount(_account), "Unknown account");
+
+    Account storage account = accounts[_account];
+
+    address signer;
+    if (keccak256(abi.encodePacked(role)) == keccak256(abi.encodePacked(ValidatorSigner))) {
```
yorhodes: there is a lot of redundancy in these if/elses (switch)
can we abstract?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -424,9 +643,14 @@ contract Accounts is
...
    */
   function getAttestationSigner(address account) public view returns (address) {
-    require(isAccount(account), "Unknown account");
-    address signer = accounts[account].signers.attestation;
-    return signer == address(0) ? account : signer;
+    return getDefaultSigner(account, AttestationSigner);
+  }
+
+  function hasDefaultSigner(address account, string memory role) public view returns (bool) {
+    require(isAccount(account));
```
yorhodes: ditto

yorhodes: `eth_call` reverting isn't super useful imo

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -424,9 +643,14 @@ contract Accounts is
...
+
+    address defaultSigner = getDefaultSigner(account, role);
```
yorhodes: `defaultSigner` is an "unnecessary variable"

yorhodes (changes requested):
haven't been able to look at the tests yet but in the spirit of throughput initiatives, lets try and make some gas optimizations first

packages/protocol/contracts/common/Accounts.sol
nambrot: Agree on storing as `bytes32` and offering versions of methods that pass the role as the hash already

packages/protocol/contracts/common/Accounts.sol
nambrot: Yup

nambrot: though I'm wondering if we should really force everyone to make these comparisons and instead just indicate it legacy roles with an extra parameter or different function?

packages/protocol/contracts/common/Accounts.sol
nambrot: yup

packages/protocol/contracts/common/Accounts.sol
nambrot: I believe in completing the authorization in the method below

packages/protocol/contracts/common/Accounts.sol
nambrot: It's a double opt-in. Like we have for our existing roles. You don't really want to allow the "burning" of an address in this scheme without their consent

packages/protocol/contracts/common/Accounts.sol
nambrot: Why shouldn't we have requires on a view? If it is to simplify certain blocks

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
+    return accounts[account].signerAuthorizations[role][signer].completed;
+  }
+
+  /**
+   * @notice Whether or not the signer has been registered as the default signer for role
+   * @param _account The address of account that authorized signing.
+   * @param signer The address of the signer.
+   * @param role The role that has been authorized.
+   */
+  function isDefaultSigner(address _account, address signer, string memory role)
```
nambrot: Same question of whether we should have separate functions to be more gas-efficient

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,196 @@ contract Accounts is
...
   function removeAttestationSigner() public {
-    Account storage account = accounts[msg.sender];
-    emit AttestationSignerRemoved(msg.sender, account.signers.attestation);
-    account.signers.attestation = address(0);
+    address signer = getDefaultSigner(msg.sender, AttestationSigner);
+    removeSigner(signer, AttestationSigner);
+    emit AttestationSignerRemoved(msg.sender, signer);
+  }
+
+  function defaultSignerToAccount(address signer, string memory role)
```
nambrot: Can't this be just a simple `signerToAccount`?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -396,15 +590,42 @@ contract Accounts is
     }
   }
 
+  /**
+   * @notice Returns the signer for the specified account and role.
+   * @param _account The address of the account.
+   * @param role The role of the signer.
+   */
+  function getDefaultSigner(address _account, string memory role) public view returns (address) {
```
nambrot: diito on separating for gas efficiency

packages/protocol/contracts/common/Accounts.sol
nambrot: What if the view is used on a non-view function tho?

packages/protocol/test/common/accounts.ts
```diff
@@ -14,21 +14,6 @@ import {
 const Accounts: AccountsContract = artifacts.require('Accounts')
 const Registry: RegistryContract = artifacts.require('Registry')
 const MockValidators: MockValidatorsContract = artifacts.require('MockValidators')
-const authorizationTests: any = {}
```
nambrot, line 17: Why remove these? I.e. we should keep them to test the backwards compatability, no?

nambrot (commented):
🙏

packages/protocol/test/common/accounts.ts
AlexBHarley: Sure was just following previous tests

packages/protocol/test/common/accounts.ts
AlexBHarley: I assume you saw that these are used further down.

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: Makes sense

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: Changed to bytes32

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: Agree that it's a little confusing. I guess there's two things at play here

- the act of proving you own an address
- setting that address as an authorised signer for an account.

Our system treats those two as one step currently. One question to ask is why have `authorizeSignerWithSignature` at all? The answer being we want to prove proof of possession before saying XXX was authorised by YYY. 

 Having `completeSignerAuthorization` is just the proof of possession for SC addresses.

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: TBH I don't see how. We have this switch block in four places but I'm not sure what an extracted function would look like as we're doing different things each time.

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: So we are of course making this backwards compatible by continuing to write to `signers`. That means all code that looks at `signers` will continue to work moving forward.

New code that's written can keep doing that if they only ever expect to deal with these roles, however it'd be recommended to check `isSigner` with a given address and role _or_ read the `signerAuthorizations` struct to verify.

packages/protocol/contracts/common/Accounts.sol
yorhodes: I suppose it is a question of what is the expected failure mode
Do developers expect to encounter a revert on an `eth_call` or a "default value"?
My point was also that we are performing the same `require` twice in one `isSigner` request

yorhodes: I guess other contracts can consume the `revert` reason from a `call` but I haven't seen many examples of this in practice

yorhodes: `call`, `delegatecall`, and `staticcall` return `true` if the account address is non-existent in the EVM
I see the value of `require` and `revert` as rolling back state changes which shouldn't be necessary in a `view`

AlexBHarley: Fair enough in this case we can remove the redundant `require`'s. 

In general though we do `require` on view calls a lot, if we now feel it's not the best practice I'm happy to not do it moving forward.

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: You mean an `isLegacySigner` function?

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: This is just extracting the common logic for each old `voteSignerToAccount`, `validatorSignerToAccount` and `attestationSignerToAccount` where we:

- perform signer to account
- verify the signer is for vote, validator or attestation respectively
- verify the signer is not an account.

Here I'm just keeping backwards compatibility

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: We do actually have a test that verifies this call reverts, would we be OK removing that?

packages/protocol/contracts/common/Accounts.sol
nambrot: I think so?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -83,12 +83,18 @@ contract Accounts is
...
+   */
+  constructor(bool test) public InitializableV2(test) {}
+
   /**
    * @notice Returns the storage, major, minor, and patch version of the contract.
    * @return The storage, major, minor, and patch version of the contract.
    */
   function getVersionNumber() external pure returns (uint256, uint256, uint256, uint256) {
-    return (1, 1, 1, 1);
+    return (2, 0, 0, 0);
```
nambrot: We have to update the storage version?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -35,12 +35,19 @@ contract Accounts is
...
     bool exists;
-    // Each account may authorize signing keys to use for voting, valdiating or attestation.
-    // These keys may not be keys of other accounts, and may not be authorized by any other
-    // account for any purpose.
+    // [Deprecated] Each account may authorize signing keys to use for voting,
+    // validating or attestation. These keys may not be keys of other accounts,
+    // and may not be authorized by any other account for any purpose.
     Signers signers;
+    mapping(bytes32 => address) defaultSigners;
+    mapping(bytes32 => mapping(address => SignerAuthorization)) signerAuthorizations;
```
nategraf: If these new fields are moved to the end of the `struct` I don't believe this needs to be a storage version upgrade, which is much more painful as discussed below.

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -59,21 +66,35 @@ contract Accounts is
...
+   */
+  constructor(bool test) public InitializableV2(test) {}
+
   /**
    * @notice Returns the storage, major, minor, and patch version of the contract.
    * @return The storage, major, minor, and patch version of the contract.
    */
   function getVersionNumber() external pure returns (uint256, uint256, uint256, uint256) {
-    return (1, 1, 1, 1);
+    return (2, 0, 0, 0);
```
nategraf: As long as this PR is a breaking storage layout change, I don't think it works as intended. Currently, the release process will require deployment of a new storage proxy that will have a clean-slate for storage. As a consequence of that none of the legacy account information will be available, which would break everything. This could be implemented such that the new contract calls into the old contract for values it is missing, but that is not how this code is currently written, and it would be very painful to do so. I don't actually think this needs to be a breaking storage version change as long as we move the new fields to the end of the `struct` they are being added to. We should do that and redefine this as a minor release, as it is otherwise only adding new functionality.

packages/protocol/contracts/common/Accounts.sol
nategraf: Based on the fact that we will never be able to fully deprecate the `signers` mapping (which would require every active account to call into this contract to set a default signer or to implement a migration function that we run), I would actually suggest the opposite of this comment and _only_ write to `signers`. Because `accounts` is `internal` I wouldn't anticipate this to cause any issues (but maybe I am missing something).

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -180,6 +201,53 @@ contract Accounts is
...
+   * @param role The role to authorize signing for.
+   * @param v The recovery id of the incoming ECDSA signature.
+   * @param r Output value r of the ECDSA signature.
+   * @param s Output value s of the ECDSA signature.
+   * @dev v, r, s constitute `signer`'s signature on `msg.sender`.
+   */
+  function authorizeSignerWithSignature(address signer, bytes32 role, uint8 v, bytes32 r, bytes32 s)
+    public
+  {
+    authorize(signer, v, r, s);
```
nategraf: I'm surprised the signature is not, and has not been, over the `role` as well as the account address. Under this implementation, a signature from the signer completing the authorization for a particular role can be repurposed to complete authorization of _any_ role at the account owners discretion. We could argue that this is an issue, that the signer fully trusts the account owner, but this at the very least feels like a nasty sharp edge.

nategraf: One additional comment on signatures is that the CIP-10 spec mentions this being upgraded to use EIP-712 signatures. Is this still happening?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,174 @@ contract Accounts is
...
+      isNotAccount(signer) && isNotAuthorizedSignerForAnotherAccount(signer),
+      "Cannot re-authorize address signer"
+    );
+
+    accounts[msg.sender].signerAuthorizations[role][signer] = SignerAuthorization({
+      started: true,
+      completed: false
+    });
+
+    emit SignerAuthorized(msg.sender, signer, role);
```
nategraf: `SignerAuthorized` is currently used in both `authorizeSigner` and `completeSignerAuthorization` without distinction. We should provide some way to distinguish between the initiation and completion of an authorization, either by adding a field to the event or using two different events.

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,174 @@ contract Accounts is
...
+  function completeSignerAuthorization(address _account, bytes32 role) public {
+    Account storage account = accounts[_account];
+    require(
+      account.signerAuthorizations[role][msg.sender].started == true,
+      "Signer authorization not started"
+    );
+    require(
+      account.signerAuthorizations[role][msg.sender].completed == false,
+      "Signer already authorized"
+    );
```
nategraf: Why not just allow this to be idempotent by removing this check?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,174 @@ contract Accounts is
...
+    emit SignerAuthorized(_account, msg.sender, role);
+  }
+
+  /**
+   * @notice Whether or not the signer has been registered as a signer for role
+   * @param account The address of account that authorized signing.
+   * @param signer The address of the signer.
+   * @param role The role that has been authorized.
+   */
+  function isGenericSigner(address account, address signer, bytes32 role)
```
nategraf: I find `isGenericSigner` to be a confusing name out-of-context of the CIP-10 proposal. (My first though when reading it was that it was a signer for all roles. I also don't quite understand the purpose of this function, which is dicuss more below.

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -284,40 +352,174 @@ contract Accounts is
...
+  }
+
+  /**
+   * @notice Whether or not the signer has been registered as a signer for role
+   * @param account The address of account that authorized signing.
+   * @param signer The address of the signer.
+   * @param role The role that has been authorized.
+   */
+  function isSigner(address account, address signer, bytes32 role) public view returns (bool) {
+    return isGenericSigner(account, signer, role) || isDefaultSigner(account, signer, role);
```
nategraf: Why check `isDefaultSigner` after `isGenericSigner`? My understanding of the intent is that the default signer is a special designation for a signer, but a default signer must as be by definition a signer. Is that not accurate? If not, it might be helpful to include some more clarifying comments about the relationship or change the code such that this is true.

nategraf (commented):
Stopping in on my way to understand CIP-10 for account recovery purposes 👋  I have some comments as a bit of an outsider to this work and hopefully its helpful.

packages/protocol/contracts/common/Accounts.sol
nategraf: I found this naming confusing as it does not actually depend on whether the signer is marked as the default. I'd suggest a name change, and based on the fact that this is intended to consolidate internal logic, a visibility change as well.

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: Great callout that `accounts` is `internal`. With that in mind we should be good to exclude `vote`, `validator` and `attestation` signers from the `defaultSigners` mapping.

packages/protocol/contracts/common/Accounts.sol
nambrot: Ohh yeah agree on both of these great points!

packages/protocol/contracts/common/Accounts.sol
nambrot: While I think it's gonna be a while, I do think we'll eventually be able to migrate everyone over or am I underestimating the cost of doing so?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -396,15 +596,39 @@ contract Accounts is
     }
   }
 
+  /**
+   * @notice Returns the signer for the specified account and role. If no 
+   * signer has been specified it will return the account itself.
+   * @param _account The address of the account.
+   * @param role The role of the signer.
+   */
+  function getDefaultSigner(address _account, bytes32 role) public view returns (address) {
```
nambrot: I don't know if this got lost/already addressed in the comments, but wdyt about separating out a function that does the legacy checks and one that doesn't? I.e. if I'm using a non-legacy role, I know that I don't have to check for these and save gas that way

packages/protocol/test/common/accounts.ts
```diff
@@ -432,148 +379,469 @@ contract('Accounts', (accounts: string[]) => {
...
+        sig.r,
+        sig.s
+      )
+      expect(signer).to.be.equal(recoveredSigner)
+    })
 
-      describe(`#authorize${upperFirst(authorizationTestDescriptions[key].subject)}()`, () => {
-        const authorized = accounts[1]
-        let sig
+    it('should set the authorized signer in two steps', async () => {
```
nambrot: I think you might want to have a separate test to check that one `completeSignerAuthorization` reverts when `authorizeSigner` wasn't called yet? In general I'm wondering if you want to split this group into the conventional `describe('#method')` groups that test reversion and event emission separately for readability?

packages/protocol/contracts/common/Accounts.sol
AlexBHarley: This has now been done. Makes a lot of sense

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -462,6 +462,25 @@ contract Accounts is
...
+    view
+    returns (bool)
+  {
+    return accounts[account].defaultSigners[role] == signer;
+  }
+
+  function isIndexedSigner(address account, address signer, bytes32 role)
+    public
+    view
+    returns (bool)
```
yorhodes: can you add some natspec comments here?

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -597,14 +629,24 @@ contract Accounts is
...
 
   /**
-   * @notice Returns the signer for the specified account and role. If no 
-   * signer has been specified it will return the account itself.
+   * @notice Checks whether the role is one of Vote, Validator or 
+   * Attestation
+   * @param role The role to check
+   */
+  function isLegacyRole(bytes32 role) public view returns (bool) {
+    return role == ValidatorSigner || role == AttestationSigner || role == VoteSigner;
```
yorhodes: this is slightly pedantic but I would order these with `VoteSigner` first for more short-circuits for normal users

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -83,6 +113,26 @@ contract Accounts is
   function initialize(address registryAddress) external initializer {
     _transferOwnership(msg.sender);
     setRegistry(registryAddress);
+    setEip712DomainSeparator();
```
yorhodes: this won't be called as this contract can't be reinitialized (without a storage upgrade)

packages/protocol/contracts/common/Accounts.sol
```diff
@@ -83,6 +113,26 @@ contract Accounts is
...
+  }
+
+  function setEip712DomainSeparator() private {
```
yorhodes: we should make this `public onlyOwner` and call it in a subsequent governance proposal

yorhodes:
I see that the `test-release` is still indicating a bump to the storage version
https://app.circleci.com/pipelines/github/celo-org/celo-monorepo/34832/workflows/ca20cd8d-d820-4453-a89f-66d023ec0888/jobs/450663
Any idea why?

yorhodes (approved):

PR #5914, opened 2020-11-18
Fix typing annoyance of send/sendAndWaitForReceipt

zviadm:
### Description
Fix mis typing by making parameter overrides more permissive.
This avoids having to do "any" casting like this:
```
.sendAndWaitForReceipt({from: xxx} as any)
```

### Other changes

n/a

### Tested

testing with pull request

### Related issues
n/a

### Backwards compatibility
Makes types more permissive so it should be backwards compatible.


zviadm:
Looks like this requires re-generating contractkit docs. I really don't want to set this up locally to generate the docs, so would appreciate if someone can take over this diff and merged it with new generated docs too.

gastonponti (approved):
LGTM, Thanks!

We are going to address a big PR in the following weeks that will fix a lot of the typing issues that we have today. But It's great to have this change right now!
(also I've updated your docs)

PR #5917, opened 2020-11-18
fix logic bug in wasValidatorDownForIntervals

zviadm:
if you want to use .some, then return value should have been negated. but using .every is way more intuitive.

### Description

Fixes simple logic bug.


eelanagaraj (approved):
LGTM

PR #5918, opened 2020-11-18
Baklava forno 2.0

tkporter:
### Description

Sets up forno 2.0 for baklava at `https://baklava-forno-k8s.celo-testnet.org`. Currently is only in a cluster in europe-west1-- the existing baklava cluster does not have VPC-native pod IPs enabled so it is not compatible with the new forno.

### Other changes

Added some env vars to preserve backward compatibility of deployed mainnet fullnodes now that the context names have been changed.

### Tested

Deployed

### Related issues

n/a

### Backwards compatibility

n/a


.env.baklava
```diff
@@ -71,6 +73,30 @@ AZURE_ORACLE_CENTRALUS_ADDRESS_AZURE_KEY_VAULTS=0x412ebe7859e9aa71ff5ce4038596f6
 AZURE_ORACLE_CENTRALUS_FULL_NODES_COUNT=2
 AZURE_ORACLE_CENTRALUS_FULL_NODES_DISK_SIZE=30
 
+# ---- Forno ----
+
+# ---- General Forno ----
+
+# A list of every context that forno will use.
```
trianglesphere, line 80: One question for my own curiosity, k8s forno works independently of the vm testnet b/c it relies on full nodes deployments and not tx/tx-private?

trianglesphere (approved):
lgtm

.env.baklava
tkporter: Exactly

PR #5925, opened 2020-11-19
[Wallet] Add support for 8 digits verification codes

### LINKED ISSUE #5384: [Valora] Valora processes 8 digit numeric digital codes
Feature switch so we can do it when all validators are upgraded.

### LINKED ISSUE #5394: Valora process 8-digit numerical codes

i1skn:
### Description

This PR implements short verification codes. See the demo below.
To enable this feature we need all the validators to support `https`, which is not the case atm, so this feature are disabled by feature flag.

### Short code
Essentially, short code, which user receives consist of two parts: prefix and security code (e.g. `12345678`: `1` - prefix, `2345678` - security code).

We use prefix to identify which issuer has sent the code. Because we can not store persistent mapping of prefix<>issuer on the client side (will be gone after reinstall), we are doing the following:
* Set prefix to issuer's address mod 10 ([0,9]) and pass it with reveal request.
* When we receive a short code with a prefix back, we create a list of validators which map to the same prefix.
* We send security code to obtain attestation code to all issuers from the previous step
* One of them should return a full attestation code, all others should fail.

This approach allows us to scale for more than 10 actionable attestations at the same time and to avoid changing `Attestation.sol` contract to store mapping prefix<>issuer in it.

#### Caveat
We do occasionally redundant requests with security code to wrong validators and those exposing the code itself and increasing waiting time for the user. For 3 SMSs the probability of this happening is 30%.

For more information see this discussion https://discord.com/channels/600834479145353243/732977310101405736/778336853505081344
### Demo
screen-20201119-212100~2

### Tested

With short codes enabled/disabled.

### Related issues

- Fixes #5384
- Fixes #5394

### Backwards compatibility

Yes


packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -67,6 +67,16 @@ export interface AttesationServiceRevealRequest {
   salt?: string
   smsRetrieverAppSig?: string
   language?: string
+  securityCodePrefix?: string
+}
+
+export interface AttesationServiceCodeForSecurityRequest {
```
codyborn: nit
```suggestion
export interface AttestationServiceCodeForSecurityRequest {
```

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -583,6 +575,19 @@ export class AttestationsWrapper extends BaseWrapper<Attestations> {
...
   }
 
+  /**
+   * Returns attestation code for provided security code from validator's attestation service
+   * @param serviceURL: validator's attestation service URL
+   * @param body: onServiceCodeForSecurityRequest
+   */
+  getAttestationForSecurityCode(serviceURL: string, body: AttesationServiceCodeForSecurityRequest) {
+    const urlParams = new URLSearchParams({ ...body })
+    return fetch(appendPath(serviceURL, 'get_attestations') + '?' + urlParams, {
```
codyborn, line 584: Does this include any retry logic? We should add this (with exponential backoff) if not.

packages/mobile/src/identity/utils.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  })
+})
+
+describe(extractShortSecurityCodeMessage, () => {
+  it('should extract with correct message', () => {
+    expect(extractShortSecurityCodeMessage('<#> Celo security code: 51365977 5yaJvJcZt2P')).toEqual(
+      '51365977'
+    )
+  })
+  it('should NOT extract with incorrect message', () => {
```
codyborn: Sorry, I don't follow. Why is this an incorrect message when the one above is correct?

packages/mobile/src/identity/verification.ts
```diff
@@ -618,13 +623,55 @@ export function attestationCodeReceiver(
...
-      if (!code) {
+      if (features.SHORT_VERIFICATION_CODES) {
+        securityCode = extractShortSecurityCodeMessage(action.message)
+        if (securityCode) {
+          const lookupHash = parseInt(securityCode[0], 10)
+          const lookupAttestations = attestations.filter(
+            (attestation) => hashAddressToSingleDigit(attestation.issuer) === lookupHash
+          )
+
+          for (const attestation of lookupAttestations) {
```
codyborn: Perhaps we can wrap the code creation and verification in a library/class. For protocols like this, it helps to not expose these details to the caller since a future code change can break it if the author doesn't understand how the complete picture.

packages/mobile/src/identity/verification.ts
```diff
@@ -618,13 +623,55 @@ export function attestationCodeReceiver(
...
+            // Try to recover the full attestation code from the matching issuer's attestation service
+            try {
+              const message = yield call(
+                getAttestationCodeFromSecurityCode,
+                attestationsWrapper,
+                account,
+                phoneHashDetails,
+                attestation,
+                securityCode.substr(1)
```
codyborn: Why is it removing the first char?

packages/mobile/src/identity/verification.ts
```diff
@@ -618,13 +623,55 @@ export function attestationCodeReceiver(
...
+    const allIssuers = attestations.map((a) => a.issuer)
+    let securityCode: string | null = null
+    let attestationCode
+
     try {
-      const code = extractAttestationCodeFromMessage(action.message)
-      if (!code) {
+      if (features.SHORT_VERIFICATION_CODES) {
+        securityCode = extractShortSecurityCodeMessage(action.message)
+        if (securityCode) {
```
codyborn: Should we log in case securityCode is undefined?

packages/mobile/src/identity/verification.ts
```diff
@@ -618,13 +623,55 @@ export function attestationCodeReceiver(
...
+                TAG + '@attestationCodeReceiver',
+                'Getting attestation code for security code error: ',
+                e
+              )
+              continue
+            }
+          }
+        }
+      } else {
+        attestationCode = extractAttestationCodeFromMessage(action.message)
```
codyborn: It might be cleaner if we run this out of the else statement regardless and just change what we pass into it.
```
let message = action.message
if (features.SHORT_VERIFICATION_CODES) {
  ...
  message = yield call(...)
}
attestationCode = extractAttestationCodeFromMessage(message)
```

codyborn (changes requested):
Looks good! Added some suggestions about organization.

packages/mobile/src/identity/utils.test.ts
i1skn: The security code consist of two parts: first digit is the prefix we use to identify the validator, when we receive the code and the rest 7 digits is the code itself. I will a comment there with the explanation as well 👍

packages/mobile/src/identity/verification.ts
i1skn: See the explanation above

packages/contractkit/src/wrappers/Attestations.ts
i1skn: Because we have potentially multiple requests for the same security code and some requests we expect to fail I'm not sure that this is a good idea. WDYT?

i1skn: I'm having a hard time to understand when do we want to do that. Basically the fact that we are not 100% sure that current validator is the one who sent the code brings some complexity here. Possible cases I see:

* Timeout - is it really worth trying it once again considering that it might not be the one we need.
* `4xx` - we expect this to happen for some validators.
* `5xx` - something went really wrong on the attestation service side. Probably, it make sense to try it later, but how much time we want to spend on each validator in total?

Any thoughts? cc: @aslawson @timmoreton @codyborn

codyborn: Perhaps we can parallelize the requests.. if we get a 200 response from one of them we return immediately. Otherwise, for each 500 or timeout we retry a couple times before giving up.

codyborn (approved):
Nice work! There's still the open question on the retry logic.. let me know what you think.

packages/contractkit/src/wrappers/Attestations.ts
aslawson: This would be an optimization. We do have manual retry by the user.  If there is a failure and user selects "resend", it will try the existing actionable issuer first again. We allow the user to do this after 1 min, but if we receive either a timeout or a failure (other than expected failure on "already sent too many times" or "wrong code") we could resend the request within that minute.

packages/contractkit/src/wrappers/Attestations.ts
```diff
@@ -67,6 +67,16 @@ export interface AttesationServiceRevealRequest {
   salt?: string
   smsRetrieverAppSig?: string
   language?: string
+  securityCodePrefix?: string
+}
+
+export interface AttestationServiceSecurityCodeRequest {
```
aslawson: it would probably make sense to use the same `GetAttestationRequestType` in utils that is used by attestation service. could also update the `AttesationServiceRevealRequest` to the `AttestationRequestType`

packages/env-tests/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "license": "MIT",
   "dependencies": {
-    "@celo/contractkit": "0.4.18",
+    "@celo/contractkit": "0.4.19-dev",
```
aslawson: was this a published version rather than dev intentionally? will be need to change it back after new contractkit version is published?

packages/mobile/src/identity/securityCode.ts
```diff
@@ -0,0 +1,103 @@
...
+  attestationsWrapper: AttestationsWrapper,
+  phoneHashDetails: PhoneNumberHashDetails,
+  account: string,
+  attestations: ActionableAttestation[],
+  securityCodeWithPrefix: string
+) {
+  const securityCodePrefix = parseInt(securityCodeWithPrefix[0], 10)
+  const lookupAttestations = attestations.filter(
+    (attestation) => hashAddressToSingleDigit(attestation.issuer) === securityCodePrefix
+  )
```
aslawson: any thoughts on caching this at the same time that cache ActionableAttestations rather than hashing+filtering all attestations each time someone wants to send a security code to a service?

considering the number of actionable attestations will be low, the time savings it probably not super significant. but seems like the most efficient way to implement.

packages/mobile/src/identity/securityCode.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  for (const attestation of lookupAttestations) {
```
aslawson: could consider parallelizing this.

pro - could save time
con - sequential request could reduce the number or requests needed (eg. the match is earlier in the list)

packages/contractkit/src/wrappers/Attestations.ts
aslawson: actually i suppose that's for the initial request for get the security code, so really only "wrong code" would be the acceptable error

packages/mobile/src/identity/securityCode.ts
```diff
@@ -0,0 +1,103 @@
...
+        attestation,
+        securityCodeWithPrefix.substr(1) // remove prefix
+      )
+      return message
+    } catch (e) {
+      Logger.warn(
+        TAG + '@getAttestationCodeForSecurityCode',
+        'Getting attestation code for security code error: ',
+        e
+      )
```
aslawson: wonder if we should differentiate between an expected error "wrong code" and others which we should log as Logger.error?

aslawson (approved):
looks really good. and love the demo -- so beautiful!

just a couple comments/ideas, but nothing blocking

packages/env-tests/package.json
i1skn: @nambrot can you help us here? Should it be non-dev by intention?

i1skn: Ok, I've reverted the change, so `env-tests` still depends on the released `0.4.18`

i1skn:
@aslawson @codyborn I've addressed all the feedback. Also, decided to make requests to obtain attestation code in parallel to all matching issuers in case of a collision, but without a retry.

packages/env-tests/package.json
nambrot: ~~Yeah I was trying to use a published version intentionally to make moving out of the monorepo easier~~

Actually looking at this again, I'd appreciate changing this to the dev-version and making the adjustment on env-tests if you already done so

packages/mobile/src/identity/securityCode.ts
```diff
@@ -0,0 +1,119 @@
...
+
+const TAG = 'identity/securityCode'
+
+const cache: { [key: string]: string } = {}
+
+function hashAddressToSingleDigit(address: Address): number {
+  return new BigNumber(address.toLowerCase()).modulo(10).toNumber()
+}
+
+export function getSecurityCodePrefix(issuerAddress: Address) {
```
nambrot: Should this code not live in ContractKit or Utils? I.e. other wallets will need to do this too right?

i1skn: Well, if other wallet would be implementing the short code verification, than yes. We can move it to `Utils` I think.

packages/mobile/src/identity/securityCode.ts
```diff
@@ -1,21 +1,26 @@
...
 function hashAddressToSingleDigit(address: Address): number {
   return new BigNumber(address.toLowerCase()).modulo(10).toNumber()
 }
 
 export function getSecurityCodePrefix(issuerAddress: Address) {
-  return `${hashAddressToSingleDigit(issuerAddress)}`
+  if (!cache[issuerAddress]) {
+    cache[issuerAddress] = `${hashAddressToSingleDigit(issuerAddress)}`
+  }
+  return cache[issuerAddress]
```
aslawson: I don't actually think this is providing that much gain, hashAddressToSingleDigit doesn't seem too computationally complex. I was thinking we could have a cache like you originally suggested that was `key:digit --> List<IssuerAddresses>`. That way when you receive a code, you can just check the cache for the list of issuers to send to rather than iterating through each issuer. Though, it would require more maintenance -- eg. removing entries whenever you remove entries from the existing ActiveAttestations cache

packages/mobile/src/identity/securityCode.ts
aslawson: I think a lot of the current implementation exists in the Attestation Wrapper (though not saying that's the most appropriate place for it...)

packages/mobile/src/identity/securityCode.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+  return null
+}
+
+// We look for the case, where all the promises fail or the first one to succeed.
+// Promise.all looks for all the values or the first error, so we can switch roles
+// for reject and resolve to achieve what we need
+async function raceUntilSuccess<T>(promises: Array<Promise<T>>) {
+  try {
+    const errors = await Promise.all(
```
codyborn, line 25: We can use `Promise.any` and not have to switch the definitions

i1skn: Unfortunately it's not a part of https://github.com/microsoft/TypeScript/blob/master/lib/lib.es2015.promise.d.ts

packages/mobile/src/identity/securityCode.ts
i1skn: I would not include it in the current PR. We can handle it later as an improvement to the current solution.

i1skn:
I've addressed all the feedback. Lmk if there anything else, otherwise I would like to merge it tomorrow before 1.6.0 cut. Thanks!

aslawson (approved):
shipit 🐑

packages/contractkit/src/wrappers/Attestations.ts
timmoreton: I think it would be nice if we can retry with backoff except on cases where we get 4xx -- especially 403 (security code invalid, or attempts exceeded) or 404 (attestation not found -- i.e it was a hash collision and this is the wrong service).  However I feel it is an optimization if time is tight

